PR #190, opened 2026-05-02
feat: execute Batch B — CLI surface and consent gate

Geoffe-Ga:
Replaces the four CLI stubs (ingest/classify/link/review) with real
engine wiring, gates `process` and `ingest` on per-source consent, and
fixes two pipeline correctness bugs the launch was leaking through.

* BUG-003 — `Pipeline._run_classification` now invokes the LLM only
  when the rule classifier left the fragment unclassified or below
  `ClassificationConfig.confidence_threshold`. Sources in
  `human_review_sources` skip the LLM entirely and route to the review
  queue.

* BUG-004 — `Pipeline._run_redaction` is fail-loud: when matches exist
  and `redaction.dry_run` is false the pipeline raises
  `RedactionRequiredError` and refuses to ingest. Documented in
  `docs/redaction.md`. Pinned by an e2e test that previously xfailed.

* INC-001 — `creek ingest --type --input --vault` resolves the
  registry, runs the matching ingestor, and writes idempotently via
  `VaultWriter`. `creek classify` and `creek link` load fragments from
  `<vault>/01-Fragments/`, dispatch on `--method`, and persist back to
  frontmatter (new engines: `creek/classify/classify_engine.py`,
  `creek/link/link_engine.py`).

* INC-002 — `creek review` walks the queue with an
  accept/override/defer/quit prompt, persisting decisions as
  `classification_method: manual` plus a `classified_at` timestamp
  (new runner: `creek/classify/review_runner.py`).

* INC-010 — `creek process` and `creek ingest` instantiate a
  `ConsentManager` rooted at `<vault>/00-Creek-Meta/Processing-Log/`
  and prompt for first-time consent. `--yes` skips the prompt and is
  recorded in the consent log. Non-interactive callers without
  `--yes` and prior consent abort exit 1 with a clear message.

* INC-011 — `creek classify --force` overrides
  `classification_method: manual`; `creek link --rebuild` deletes the
  cached embeddings archive at `<vault>/00-Creek-Meta/embeddings.npz`
  before recomputing.

Tests: 21 new engine tests + new CLI integration tests for ingest,
classify, link, review, consent gating, and redaction abort. The
previously-xfailing redaction e2e test is now strict and passing.

https://claude.ai/code/session_01Fxqhm2226mKAUJsFZDXYoE

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -17,7 +17,6 @@ from typing import TYPE_CHECKING
 
 import frontmatter
 import yaml
-from pydantic import ValidationError
 
 from creek.classify.constants import (
     CLASSIFICATION_METHOD_KEY,
@@ -29,6 +28,7 @@ from creek.classify.llm import LLMClassifier
 from creek.classify.rules import RuleClassifier
 from creek.ingest.base import LA_TZ
 from creek.models import Fragment, Frequency
+from creek.vault.reader import try_load_fragment
 
 if TYPE_CHECKING:
     from creek.config import CreekConfig
@@ -239,7 +239,15 @@ def _classify_one(
 
 
 def _read_fragment(md_file: Path) -> tuple[Fragment, str, dict[str, object]] | None:
-    """Load a fragment file's metadata, body, and raw frontmatter dict.
+    """Thin alias for :func:`creek.vault.reader.try_load_fragment`.
+
+    Kept as a private name so existing patches in
+    ``tests/test_classify_engine.py`` continue to target the engine
+    module rather than the shared reader. The semantics are
+    identical: ``None`` means "not a Creek fragment, silently skip",
+    while real I/O failures propagate so the engine's outer
+    ``except`` block can record them on
+    :attr:`ClassifySummary.errors`.
 
     Args:
         md_file: Markdown file to load.
@@ -247,25 +255,14 @@ def _read_fragment(md_file: Path) -> tuple[Fragment, str, dict[str, object]] | N
     Returns:
         ``(fragment, body, raw_metadata)`` for valid fragments, or
         ``None`` when the file is well-formed YAML but is **not** a
-        Creek fragment (no ``type: fragment`` key, or a schema
-        mismatch). Real I/O failures are propagated to the caller so
-        they can be recorded on :attr:`ClassifySummary.errors`.
+        Creek fragment.
 
     Raises:
         OSError: When the file cannot be opened.
         ValueError: When the YAML cannot be parsed.
         yaml.YAMLError: When the YAML parser rejects the document.
     """
-    post = frontmatter.load(str(md_file))
-    metadata = dict(post.metadata)
-    if metadata.get("type") != "fragment":
-        return None
-    try:
-        fragment = Fragment.model_validate(metadata)
-    except ValidationError:
-        logger.debug("Skipping invalid fragment frontmatter: %s", md_file)
-        return None
-    return fragment, str(post.content), metadata
+    return try_load_fragment(md_file)
 
 
 def _write_fragment(
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 import logging
 from dataclasses import dataclass
 from datetime import datetime
-from pathlib import Path  # noqa: TC003 — runtime use in dataclass field
+from pathlib import Path  # noqa: TC003  # no issue: runtime dataclass field
 from typing import TYPE_CHECKING
 
 import frontmatter
@@ -176,7 +176,7 @@ def _classify_one(
         if confidence >= confidence_threshold:
             return rule_result, True
 
-    if llm is None:  # pragma: no cover — guarded by ``method == "llm"``
+    if llm is None:  # pragma: no cover  # no issue: defensive guard, unreachable
         msg = "LLM classifier required when method='llm'"
         raise RuntimeError(msg)
     return llm.classify(rule_result, content=body), False
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -13,12 +13,17 @@ import logging
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path  # noqa: TC003 — runtime use in dataclass field
-from typing import TYPE_CHECKING, Final
+from typing import TYPE_CHECKING
 
 import frontmatter
 import yaml
 from pydantic import ValidationError
 
+from creek.classify.constants import (
+    CLASSIFICATION_METHOD_KEY,
+    CLASSIFIED_AT_KEY,
+    MANUAL_METHOD,
+)
 from creek.classify.llm import LLMClassifier
 from creek.classify.rules import RuleClassifier
 from creek.ingest.base import LA_TZ
@@ -28,14 +33,6 @@ if TYPE_CHECKING:
     from creek.config import CreekConfig
 
 logger = logging.getLogger(__name__)
-
-_CLASSIFICATION_METHOD_KEY: Final[str] = "classification_method"
-"""Frontmatter key carrying ``rules | llm | manual``."""
-
-_CLASSIFIED_AT_KEY: Final[str] = "classified_at"
-"""Frontmatter key carrying the ISO-8601 classification timestamp."""
-
-_MANUAL_METHOD: Final[str] = "manual"
 
 
 @dataclass(frozen=True)
@@ -66,7 +63,6 @@ def run_classify(
     vault_path: Path,
     config: CreekConfig,
     method: str,
-    batch_size: int,
     force: bool,
 ) -> ClassifySummary:
     """Classify every fragment in *vault_path* using the chosen method.
@@ -75,9 +71,6 @@ def run_classify(
         vault_path: Vault root.
         config: Loaded Creek configuration.
         method: ``"rules"`` or ``"llm"``.
-        batch_size: Reserved; the current LLM classifier processes
-            fragments individually because the per-vault scan is
-            already streaming.
         force: When ``True``, overwrite ``classification_method:
             manual`` decisions.
 
@@ -99,13 +92,19 @@ def run_classify(
 
     for md_file in sorted(fragments_root.rglob("*.md")):
         total += 1
-        record = _read_fragment(md_file)
+        try:
+            record = _read_fragment(md_file)
+        except (OSError, ValueError, yaml.YAMLError) as exc:
+            errors.append(f"unreadable fragment {md_file}: {exc}")
+            continue
         if record is None:
-            errors.append(f"unreadable fragment: {md_file}")
+            # File parsed cleanly but is not a Creek fragment (no
+            # ``type: fragment`` field, or schema mismatch). Silently
+            # skip — markdown notes coexist with fragments in the vault.
             continue
         fragment, body, raw = record
 
-        if not force and raw.get(_CLASSIFICATION_METHOD_KEY) == _MANUAL_METHOD:
+        if not force and raw.get(CLASSIFICATION_METHOD_KEY) == MANUAL_METHOD:
             preserved += 1
             continue
 
@@ -190,14 +189,18 @@ def _read_fragment(md_file: Path) -> tuple[Fragment, str, dict[str, object]] | N
         md_file: Markdown file to load.
 
     Returns:
-        ``(fragment, body, raw_metadata)`` or ``None`` when the file is
-        not a valid fragment record.
+        ``(fragment, body, raw_metadata)`` for valid fragments, or
+        ``None`` when the file is well-formed YAML but is **not** a
+        Creek fragment (no ``type: fragment`` key, or a schema
+        mismatch). Real I/O failures are propagated to the caller so
+        they can be recorded on :attr:`ClassifySummary.errors`.
+
+    Raises:
+        OSError: When the file cannot be opened.
+        ValueError: When the YAML cannot be parsed.
+        yaml.YAMLError: When the YAML parser rejects the document.
     """
-    try:
-        post = frontmatter.load(str(md_file))
-    except (OSError, ValueError, yaml.YAMLError):
-        logger.debug("Skipping unreadable markdown file: %s", md_file)
-        return None
+    post = frontmatter.load(str(md_file))
     metadata = dict(post.metadata)
     if metadata.get("type") != "fragment":
         return None
@@ -233,8 +236,8 @@ def _write_fragment(
     """
     new_metadata = dict(raw)
     new_metadata.update(fragment.model_dump(mode="json"))
-    new_metadata[_CLASSIFICATION_METHOD_KEY] = method
-    new_metadata[_CLASSIFIED_AT_KEY] = datetime.now(tz=LA_TZ).isoformat()
+    new_metadata[CLASSIFICATION_METHOD_KEY] = method
+    new_metadata[CLASSIFIED_AT_KEY] = datetime.now(tz=LA_TZ).isoformat()
 
     post = frontmatter.Post(content=body, **new_metadata)
     md_file.write_text(frontmatter.dumps(post), encoding="utf-8")
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -10,7 +10,7 @@ and ``classified_at`` timestamp on each fragment so that subsequent
 from __future__ import annotations
 
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path  # noqa: TC003  # no issue: runtime dataclass field
 from typing import TYPE_CHECKING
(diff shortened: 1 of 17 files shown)


Geoffe-Ga:
Thanks for the careful review. Addressed in 734712c:

**Blockers**
- 🔴 **Duplicated constants** — extracted `CLASSIFICATION_METHOD_KEY`, `CLASSIFIED_AT_KEY`, and `MANUAL_METHOD` to a new `creek/classify/constants.py`. Both `classify_engine.py` and `review_runner.py` now import from there. A rename in either writer can no longer silently break the manual-fragment filter.
- 🔴 **Misleading test name** — split into two tests:
  - `test_run_classify_skips_non_fragment_files` (silent skip; asserts `errors == []`).
  - `test_run_classify_unreadable_file_records_error` (real error path; patches `_write_fragment` to raise `OSError` and asserts `len(errors) == 1`).
  - Refactored `_read_fragment` to propagate `OSError`/`YAMLError` so the engine records them on `ClassifySummary.errors` instead of conflating them with "not a Creek fragment" silent skips.

**Code quality**
- `import sys` moved to module top in `cli.py`.
- Dropped unused `batch_size` from `run_classify` and unused `config` from `_gate_consent`. The CLI's `--batch-size` flag is gone; concurrency lives in `llm.max_concurrent`.
- Extracted the per-test fragment writer helper to `tests/helpers.py` (`write_fragment_file`); the three engine test files now share it.
- Strengthened `test_run_link_uses_existing_cache`: instead of the `mtime >= mtime` tautology, the test patches `EmbeddingLinker.generate_embeddings` and asserts the cache hit by checking every fragment id is passed via `existing_ids`, plus asserts the on-disk bytes are identical.

**Security (LOW accepted)**
- `_operator_identity()` now strips metacharacters (`r"[^\w.@+\- ]"`), truncates to 64 chars, and falls back to `"cli"` for an empty result. Three new tests cover the strip, truncate, and fallback paths.
- The `--yes`-in-CI note is left as documented operational guidance per your "operational concern rather than a code defect" framing.

All 9 `check-all.sh` gates green; 2759 unit + 17 e2e tests pass; coverage 93.94%.

@claude please re-review.

---

Geoffe-Ga:
Two reviews landed concurrently on `734712c`: one LGTM (re-review trigger) and one CHANGES_REQUESTED (push trigger from a different workflow run) flagging CLAUDE.md §1.3 `noqa`/`pragma` format compliance.

Addressed in 39c2bda by switching the new modules to the reviewer's `# no issue: <reason>` shortcut:

| Location | Before | After |
|---|---|---|
| `creek/classify/classify_engine.py:15` | `# noqa: TC003 — runtime use in dataclass field` | `# noqa: TC003  # no issue: runtime dataclass field` |
| `creek/classify/classify_engine.py:179` | `# pragma: no cover — guarded by method == "llm"` | `# pragma: no cover  # no issue: defensive guard, unreachable` |
| `creek/classify/review_runner.py:18` | em-dash form | `# no issue: runtime dataclass field` |
| `creek/link/link_engine.py:14` | em-dash form | `# no issue: runtime dataclass field` |
| `tests/helpers.py:13` | em-dash form | `# no issue: runtime helper signature` |

**Out-of-scope, by design:** `creek/consent.py:25` and `creek/redact/scanner.py:30` predate this PR and use the same em-dash convention. Updating those would be a project-wide sweep that should land in its own PR alongside a CLAUDE.md update reconciling the prescribed format with the actual codebase pattern. Happy to do that as follow-up if you'd like — just say the word.

All 9 `check-all.sh` gates green.

@claude please re-review.

---

Geoffe-Ga:
Two reviews landed concurrently on `39c2bda`: the re-review trigger run gave LGTM; the push-trigger fresh-look run flagged two new blockers. Addressing both in 24cedc3, with one push-back.

### Fixed

🔴 **Blocker B — unguarded `save_embeddings` in link engine.** Correct catch — load path was already guarded, save path wasn't. Wrapped `linker.save_embeddings(merged, cache_path)` in `try/except OSError`, logged warning, returned `merged` unchanged so linking still succeeds. Lost cache costs a recompute on the next run. New regression test `test_run_link_embeddings_save_oserror_does_not_crash` patches `EmbeddingLinker.save_embeddings` to raise `OSError("disk full")` and asserts `fragment_count == 1`. (`creek/link/link_engine.py:180-191`)

🟢 **Cleanup — duplicate `from unittest.mock import patch`** inside `test_run_classify_unreadable_file_records_error` removed; already imported at module top. (Residual nit from the 00:52 LGTM.)

### Pushing back

🔴 **Blocker A — `KeyError` caught in `_run_ingest`.** This catch is **deliberate**, not a smell. `creek/ingest/base.py:295-303` (`assemble_ingested_fragment`) explicitly documents:

> Raises:
>     KeyError: If ``parsed.metadata`` is missing the ``frontmatter`` or ``markdown`` keys; this signals an ingestor contract violation rather than a recoverable parse error.

`KeyError` is the documented domain signal for an ingestor failing to populate the required metadata keys. The pre-existing `Pipeline._run_ingestion` (`creek/pipeline.py:251`, predates this PR) catches the same `(KeyError, ValueError)` pair for the same reason — it's the codebase convention for batch ingest. Removing `KeyError` would propagate the exception out of `_run_ingest` and crash the entire batch the moment one ingestor misbehaves; per-file errors are exactly the right granularity here.

The second `KeyError` catch (`writer.write_fragment`) defends against a missing `_PLATFORM_SUBFOLDER` mapping entry, with a totality unit test on the mapping itself — same per-file recovery rationale.

Happy to introduce a domain-specific `IngestorContractError` if you want a tighter type, but that should land as its own PR with the corresponding refactor in `assemble_ingested_fragment` and the pre-existing `Pipeline._run_ingestion` site.

### Non-blocking deferrals

🟡 **HIGH — `_gate_consent` piped-stdin hang.** The function already short-circuits on non-TTY input via `_is_interactive()` (`creek/cli.py:160-174`) — `typer.prompt` is unreachable from a non-interactive caller without prior consent because the function exits 1 first. The "if somehow `typer.prompt()` is reached" branch in the review is theoretical given the guard. Tracking as a future hardening note rather than a code change.

🟢 The three other LOW security observations (symlink traversal, Unicode lookalikes, single-writer race) are reasonable future-PR items but neither introduced by this PR nor blocking; preserving the existing assumption.

All 9 `check-all.sh` gates green; 2761 unit + 17 e2e tests pass.

@claude please re-review.

---

Geoffe-Ga:
Two reviews on `24cedc3`: the re-review trigger gave LGTM and accepted the `KeyError` push-back; the push-trigger fresh-look run found three new items. Addressing two-of-three plus the LOW nit in 2154736.

### Fixed

🔴 **BLOCKING — `classify` CLI dropped `summary.errors`.** Confirmed; `run_classify` accumulates errors but the CLI never printed them. Now mirrors the `ingest` block:

```python
if summary.errors:
    console.print(f"[yellow]Errors: {len(summary.errors)}[/yellow]")
    for err in summary.errors:
        console.print(f"  [dim]{err}[/dim]")
```

🟡 **HIGH — `--rebuild` deleted the cache regardless of `--method`.** Confirmed; cache unlink ran before the method dispatch. Gated on `method == "embeddings"`. New regression test `test_run_link_rebuild_is_noop_for_temporal` writes a sentinel into the cache, runs `method="temporal" rebuild=True`, and asserts the bytes survive.

🟢 **LOW — `_override_frequency` case-sensitivity.** Previous code worked for the documented casing (`F1`..`F10`, `unclassified`) but `Unclassified`/`UNCLASSIFIED` fell through. Replaced with `_parse_frequency_input(value)` that does a single case-insensitive scan across the enum members. Three new unit tests cover the casing variants; a new CLI-level test (`test_run_interactive_override_accepts_uppercase_unclassified`) confirms `UNCLASSIFIED` now overrides correctly end-to-end.

### Pushing back

🟡 **HIGH — Fragment-loading triplication** (`classify_engine._read_fragment`, `review_runner._read_entry`, `link_engine._load_fragments`). The reviewer themselves wrote:

> Open a GH issue rather than blocking this PR.

Agreed. Tracking it as future work; the consolidation belongs in its own PR (likely a `creek.vault.reader.load_vault_fragments()` helper that yields `(Path, Fragment, body, raw_metadata)` tuples) rather than expanding the scope of Batch B. I'll open the issue after this lands.

### Other low-severity items deferred

- **`_OPERATOR_IDENTITY_ALLOWED_CHARS` admits `\w` Unicode** — the docstring already documents that we accept unicode usernames intentionally. Noted, no code change.
- **Already-consented sources skip `source_path.exists()`** — minor surprise but downstream ingestors fail with their own error; not a security gap.
- **`_format_summary` doesn't validate negative size** — practically impossible (`Path.stat().st_size >= 0`).

All 9 `check-all.sh` gates green.

@claude please re-review.

---

Geoffe-Ga:
Sixth review pass on `2154736` flagged two real bugs. Both correct, both fixed in 1e496b6.

### Fixed

🔴 **BLOCKER 1 — `--method llm` short-circuit silently dropped writes.** Confirmed: the loop did `if was_skipped: skipped += 1; continue` without writing, so a high-confidence rule result never made it to disk. Fragment stays unstamped and re-enters every subsequent classify run.

Fix:
- Always write the fragment.
- Stamp `classification_method: rules` (not `llm`) when the rule classifier is what produced the answer — truthful provenance over CLI flag echo.
- `skipped_high_confidence` now means "LLM not invoked", not "write skipped". A fragment in this state counts toward both `classified` and `skipped_high_confidence`.

Existing `test_run_classify_llm_skips_high_confidence` strengthened: now reloads the file from disk and asserts `reloaded["classification_method"] == "rules"` and `summary.classified == summary.skipped_high_confidence`. Without this assertion the bug was invisible.

🔴 **BLOCKER 2 — `total` counted non-Creek markdown files.** Confirmed: `total += 1` ran before `_read_fragment` had a chance to reject non-fragment files. The CLI's "Classified 2 of 5" reading was misleading whenever the vault had unrelated notes in `01-Fragments/`.

Fix: increment `total` only after `_read_fragment` returns a confirmed fragment. Updated `test_run_classify_skips_non_fragment_files` to assert `total == 0` for a non-fragment input (the previous `total == 1` assertion was pinned to the buggy behaviour). New `test_run_classify_total_only_counts_creek_fragments` seeds 2 fragments + 3 unrelated notes and asserts `total == 2`.

### Refactor

`run_classify`'s cyclomatic complexity bumped over the project's B-rank cap after the fixes. Extracted the per-file dispatch into `_process_file`, threaded mutable counters through a private `_RunCounts` dataclass. No behaviour change; the function is now leaner and the inner branches all live in one place.

### Other items (no action)

The remaining notes from the review (LA_TZ leakage, atomic writes, consent gate stale path, fragment-loader DRY, hardcoded LA timezone) are valid future-PR items but explicitly non-blocking. Tracking them as follow-ups.

All 9 `check-all.sh` gates green; 2765 unit + 17 e2e tests pass; coverage 93.91%.

@claude please re-review.

---

Geoffe-Ga:
Seventh review pass on `1e496b6`: the re-review trigger gave LGTM on both blocker fixes; the push-trigger fresh look found one new blocker that I missed.

### Fixed

🔴 **BLOCKING — `_persist_manual` had no `OSError` handler.** Confirmed and addressed in a51414c. The classify engine's `_process_file` already wrapped `_write_fragment` in `try/except OSError` and appended to `counts.errors`, but `review_runner._persist_manual` was the same write call without the guard. A disk-full mid `creek review` session would have crashed with a raw traceback.

Fix:
- Added `ReviewSummary.errors: list[str]` (mirrors `ClassifySummary.errors`).
- New `ReviewQueueRunner._save` wraps `_persist_manual` in `try/except OSError`, records to `summary.errors`, and prints a red error to the console immediately so the operator can react.
- `run_interactive` only increments `accepted`/`overridden` on successful saves — failed writes don't pollute the counters.
- The CLI's `review` command now surfaces `summary.errors` the same way `classify` does.

New regression test `test_run_interactive_persist_oserror_records_error` patches `_persist_manual` to raise `OSError("disk full")` and asserts `summary.accepted == 0`, `len(summary.errors) == 1`, and `"disk full" in summary.errors[0]`.

### Other items (deferred)

The non-blocking nits from the same review (`tests/test_cli.py` not using `write_fragment_file`, `LA_TZ` location coupling, `--batch-size` removal documentation, `link` CLI silent skip on parse errors) are all reasonable future-PR items. Tracking as follow-ups; not expanding Batch B scope further.

All 9 `check-all.sh` gates green.

@claude please re-review.

---

Geoffe-Ga:
Eighth review pass on `a51414c`: the re-review trigger gave LGTM; the push-trigger fresh look raised two new BLOCKING items. Both addressed in 5beb0f7.

### Fixed

🔴 **BLOCKER 1 — Missing CLI coverage for consent cache hit + per-source isolation.** Legitimate gap. The pipeline-level `TestPipelineConsent::test_consent_granted_runs_ingestion` covered it indirectly, but the CLI surface didn't. Added two new tests in `tests/test_cli.py`:

- `test_process_skips_prompt_when_consent_already_recorded` — invokes `--yes` once, then runs again without `--yes` and asserts exit 0 and no "First time processing" in output.
- `test_process_consent_is_per_source` — grants consent for `src_a`, runs against `src_b` without `--yes`, asserts exit 1 with a consent message. Pins per-source isolation.

🔴 **BLOCKER 2 — Fragment-loading triplicated.** Two prior reviewers framed this as non-blocking; this one escalated it. Rather than ping-pong on reviewer disagreement, doing the consolidation:

New module `creek/vault/reader.py`:
- `try_load_fragment(md_file)` — single source of truth. Returns `(fragment, body, raw_metadata) | None`. Real I/O failures propagate; non-fragment markdown silently returns None.
- `iter_vault_fragments(fragments_root)` — convenience wrapper for callers that don't need to surface I/O errors.

Refactored call sites:
- `classify_engine._read_fragment` — now a one-line alias (kept the private name so existing `mock.patch(...)` test patterns keep working).
- `review_runner._read_entry` — calls `try_load_fragment`, wraps in `ReviewEntry`.
- `link_engine._load_fragments` — collapses to a single call + list comprehension; `Fragment` moved under `TYPE_CHECKING`.

Schema changes to `Fragment` or rename of the `type` sentinel now need to update one place, not three.

New `tests/test_vault_reader.py` adds six unit tests covering the valid/non-fragment/schema-mismatch/missing-file/mixed-directory/missing-root cases.

### Other items deferred

The non-blocking nits from this and prior reviews (`tests/test_cli.py` not consistently using `write_fragment_file`, `LA_TZ` location coupling, `--batch-size` removal documentation, link CLI silent skip on parse errors, retry on bad frequency input, atomic write-then-rename) remain reasonable future-PR items. Tracking as follow-ups.

All 9 `check-all.sh` gates green; 2772 unit + 17 e2e tests pass; coverage 93.96%.

@claude please re-review.

---

Geoffe-Ga:
Two reviews on `5beb0f7`: re-review trigger gave LGTM and confirmed both prior blockers resolved; push-trigger fresh look raised three new items. Addressing one, pushing back on two.

### Fixed

🔴 **BLOCKING — `ClassifySummary(frozen=True)` with mutable `list[str]`.** Legitimate point: `frozen=True` blocks attribute reassignment but not in-place `.append()`, so the annotation was misleading. Fixed in d26d875:

- `errors: list[str]` → `errors: tuple[str, ...]`
- `run_classify` snapshots via `tuple(counts.errors)` at construction
- Mutable accumulation still lives on the private `_RunCounts` dataclass
- New regression test `test_classify_summary_errors_is_an_immutable_tuple` pins both `isinstance(summary.errors, tuple)` and that `.append(...)` raises `AttributeError`
- Updated existing assertion `summary.errors == []` → `== ()`

### Pushing back

🟡 **`KeyError` caught in `_run_ingest`.** This is the *third* time this exact item has surfaced. The history:

- **Round 4** (`4362593317`): same reviewer raised it as BLOCKING, I pushed back with the rationale below.
- **Round 5** (`4362647957`): a separate reviewer accepted the push-back: *"The reviewer's concern doesn't apply because the `KeyError`s that can reach these handlers are all explicitly raised by library code with diagnostic messages. The push-back stands."*
- **Round 9** (now): re-raised again, downgraded to MEDIUM.

The argument hasn't changed:
- `creek/ingest/base.py:295-303`'s `assemble_ingested_fragment` *explicitly documents and raises* `KeyError` as its ingestor-contract-violation signal — this is a designed exception, not Python's implicit dict miss.
- `creek/pipeline.py:313`'s `Pipeline._run_ingestion` (predates this PR) catches the same `(KeyError, ValueError)` pair for the same per-file recovery reason.
- Removing `KeyError` from the catch would crash the entire ingest the moment one ingestor misbehaves — strictly worse UX for batch processing.

A separate PR introducing a domain-specific `IngestorContractError` and updating both call sites in lock-step would be a cleaner long-term answer; that's not Batch B's scope.

🟡 **`iter_vault_fragments` naming.** The very same reviewer wrote in the LGTM comment two minutes earlier (`4364064716`):

> `iter_vault_fragments` is named like a generator but returns a `list`. `load_vault_fragments` would be a more accurate name, but the current name isn't wrong and changing it would touch all three call sites. **Not worth the churn.**

Holding to that judgment. The `iter_*` naming is mildly imprecise but harmless and the LGTM-side analysis is correct: cross-module rename for stylistic gain isn't worth shipping a tenth round of review. If we open the follow-up issue for `LinkSummary.errors` surfacing (also raised in this review), the rename can ride along.

All 9 `check-all.sh` gates green; 70 tests pass in the impacted suites.

@claude please re-review.

---

PR #194, opened 2026-05-02
perf(batch-e): vault writer + dedup + voice scaling fixes (PERF-001/003/004, BUG-006)

Geoffe-Ga:
Executes [Batch E — Vault writer + dedup performance](plans/prompts/2026-04-28/batch-E-vault-performance.md). Five commits, four scaling/race fixes plus a `slow` marker so the benchmarks they add never bloat the default unit pass.

## Changes

| Commit | Issue(s) | Summary |
|---|---|---|
| `4910e4f` | infra | Add `slow` pytest marker; `./scripts/test.sh` excludes it from unit, `--all` opts in. |
| `9064a63` | **PERF-001 + BUG-006** | `VaultWriter` per-directory `.id-index.jsonl` (O(1) lookup), `O_CREAT \| O_EXCL` atomic file creation with counter retry, `provenance.jsonl` append, `threading.Lock` around the index + provenance updates. |
| `e55ea67` | **PERF-003** | Vectorised `find_duplicates` via L2-normalise + `X @ X.T`; optional FAISS `IndexFlatIP.range_search` behind `DeduplicationConfig.use_faiss`, lazy-imported with graceful fallback. |
| `81fe51a` | **PERF-004** | `VoiceProfileGenerator.generate_all_profiles` streams qualifying fragments to per-register JSONL accumulators in a `tempfile.TemporaryDirectory`; bodies dropped per-iteration so peak memory is one register, not the whole vault. |
| `dd2f17a` | review | Address Claude PR review feedback (see below). |

## Review feedback addressed (commit `dd2f17a`)

- **🟡 writer**: `_atomic_write_text` switched from a fixed `.{name}.tmp` sidecar to `tempfile.mkstemp(dir=target_dir)` so concurrent writers cannot collide on the temp name. Regression test stresses 8 threads × 50 writes against the same path.
- **🟡 voice**: `_JsonlWriter.append` now flushes per line so an interrupted run leaves a coherent prefix instead of half a JSON object. Regression test asserts an out-of-band reader sees each entry immediately.
- **🟢 writer**: `_atomic_create` retry loop capped at `_MAX_FILENAME_COLLISION_RETRIES = 10_000` with `RuntimeError`; replaces the prior unbounded `while True`. Test monkey-patches `os.open` to always collide and verifies the cap fires.
- **🟢 dedup**: removed redundant `np.fill_diagonal(-inf)` (`np.triu(k=1)` already excludes the diagonal, and configured thresholds are always > 0).
- **🟢 dedup**: narrowed `ImportError` swallowing — `_pair_indices` only falls back to matmul when the *top-level* `faiss` import fails (via the new `_faiss_available()` probe). Submodule failures inside FAISS surface as real bugs.

Skipped (non-blocking, would require restructuring with no test gain):
- Lock-scope optimisation in `_write_model` (perf hint; current code is correct, single in-process serialisation is the documented trade-off).
- `_find_existing` deprecation note (purely cosmetic; method is `_`-prefixed already).
- Second opinion on the explicit `del` in `_stream_into` (already commented; subjective).

## Test plan

- [x] `./scripts/test.sh --unit` → 2 795 passed, 19 deselected.
- [x] `python -m pytest -m slow` → 3 passed (10k-fragment dedup < 30 s, 2k-fragment vault writer constant-time, voice memory < 200 MB).
- [x] `./scripts/typecheck.sh` strict clean.
- [x] `./scripts/lint.sh` + `./scripts/format.sh --check` clean.
- [x] Aggregate coverage 93.67% (≥ 90% gate); per-file gate green.
- [x] Pylint 9.75/10; complexity all A.

The single failing local check is the pre-existing `pyjwt 2.7.0 → CVE-2026-32597` from the system Python environment — verified to fail on `main` too — and is resolved by CI's clean install from `requirements-dev.txt`.

https://claude.ai/code/session_01NgnhBebeSwdT8ReRd3bKtA

### DIFF
--- a/creek-tools/creek/clean/semantic_dedup.py
+++ b/creek-tools/creek/clean/semantic_dedup.py
@@ -228,9 +228,11 @@ def _matmul_pair_indices(
     """
     import numpy as np
 
-    sim = matrix @ matrix.T
-    np.fill_diagonal(sim, -np.inf)
-    sim_upper = np.triu(sim, k=1)
+    # ``triu(k=1)`` already zeroes the diagonal and lower triangle, and
+    # configured thresholds are always > 0 (enforced in
+    # ``SemanticDeduplicator.__init__``), so the zeroed entries cannot
+    # masquerade as real matches.
+    sim_upper = np.triu(matrix @ matrix.T, k=1)
     rows, cols = np.where(sim_upper >= threshold)
     sims = sim_upper[rows, cols].astype(np.float32, copy=False)
     return rows.astype(np.int64), cols.astype(np.int64), sims
@@ -452,17 +454,28 @@ class SemanticDeduplicator:
         """Return ``(rows, cols, sims)`` for pairs above resonance threshold.
 
         Honors :attr:`DeduplicationConfig.use_faiss` and gracefully
-        degrades to the dense matmul if FAISS is not importable.
+        degrades to the dense matmul only if the top-level ``faiss``
+        import itself fails. ``ImportError`` raised from inside FAISS
+        (e.g. a missing submodule) is **not** swallowed so internal
+        FAISS bugs surface as real failures rather than silent
+        fallbacks.
         """
-        if self.config.use_faiss:
-            try:
-                return _faiss_pair_indices(normalised, self.resonance_threshold)
-            except ImportError:
-                logger.warning(
-                    "use_faiss=True but faiss is not installed — "
-                    "falling back to dense matmul.",
-                )
+        if self.config.use_faiss and self._faiss_available():
+            return _faiss_pair_indices(normalised, self.resonance_threshold)
         return _matmul_pair_indices(normalised, self.resonance_threshold)
+
+    @staticmethod
+    def _faiss_available() -> bool:
+        """Return whether the optional ``faiss`` package is importable."""
+        try:
+            import faiss  # noqa: F401
+        except ImportError:
+            logger.warning(
+                "use_faiss=True but faiss is not installed — "
+                "falling back to dense matmul.",
+            )
+            return False
+        return True
 
     def check_fragment(
         self,
--- a/creek-tools/creek/clean/semantic_dedup.py
+++ b/creek-tools/creek/clean/semantic_dedup.py
@@ -222,20 +222,27 @@ def _matmul_pair_indices(
     """Return upper-triangle (i, j, sim) triples above *threshold* via matmul.
 
     Caller is responsible for ensuring *matrix* has been L2-normalised.
-    Memory peaks at ``N²`` float32 entries (≈ 400 MB at N = 10 000) —
-    acceptable up to that scale and the documented FAISS path takes
-    over for larger corpora.
+
+    Memory: only **one** ``N x N`` float32 matrix is live at a time
+    (≈ 400 MB at N = 10 000). The previous implementation used
+    ``np.triu`` which materialised a second N x N copy alongside the
+    matmul result, doubling peak memory; the current code uses
+    ``np.triu_indices`` to extract the strictly upper-triangle
+    coordinates without copying the matrix.
+
+    The documented FAISS path is the right cutover for larger corpora.
     """
     import numpy as np
 
-    # ``triu(k=1)`` already zeroes the diagonal and lower triangle, and
-    # configured thresholds are always > 0 (enforced in
-    # ``SemanticDeduplicator.__init__``), so the zeroed entries cannot
-    # masquerade as real matches.
-    sim_upper = np.triu(matrix @ matrix.T, k=1)
-    rows, cols = np.where(sim_upper >= threshold)
-    sims = sim_upper[rows, cols].astype(np.float32, copy=False)
-    return rows.astype(np.int64), cols.astype(np.int64), sims
+    sim = matrix @ matrix.T
+    rows, cols = np.triu_indices(sim.shape[0], k=1)
+    upper = sim[rows, cols]
+    mask = upper >= threshold
+    return (
+        rows[mask].astype(np.int64, copy=False),
+        cols[mask].astype(np.int64, copy=False),
+        upper[mask].astype(np.float32, copy=False),
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -313,6 +320,10 @@ class SemanticDeduplicator:
         self.resonance_threshold = config.resonance_threshold
         self._index: dict[str, np.ndarray] = {}
         self._dimension: int | None = None
+        # Memoised result of the ``import faiss`` probe in
+        # ``_faiss_available``. ``None`` = not yet probed; ``True`` /
+        # ``False`` = importable / not.
+        self._faiss_probe_cache: bool | None = None
 
     def _validate_embedding(
         self,
@@ -464,9 +475,18 @@ class SemanticDeduplicator:
             return _faiss_pair_indices(normalised, self.resonance_threshold)
         return _matmul_pair_indices(normalised, self.resonance_threshold)
 
-    @staticmethod
-    def _faiss_available() -> bool:
-        """Return whether the optional ``faiss`` package is importable."""
+    def _faiss_available(self) -> bool:
+        """Return whether the optional ``faiss`` package is importable.
+
+        The probe result is cached on the instance so a long-running
+        pipeline that calls :meth:`find_duplicates` repeatedly with
+        ``use_faiss=True`` against a FAISS-less environment emits at
+        most one fallback warning per deduplicator instead of one per
+        call.
+        """
+        cached = self._faiss_probe_cache
+        if cached is not None:
+            return cached
         try:
             import faiss  # noqa: F401
         except ImportError:
@@ -474,7 +494,9 @@ class SemanticDeduplicator:
                 "use_faiss=True but faiss is not installed — "
                 "falling back to dense matmul.",
             )
+            self._faiss_probe_cache = False
             return False
+        self._faiss_probe_cache = True
         return True
 
     def check_fragment(
--- a/creek-tools/creek/clean/semantic_dedup.py
+++ b/creek-tools/creek/clean/semantic_dedup.py
@@ -21,16 +21,30 @@ fragment vs. existing index) processing modes.
 Embedding vectors must all share the same dimensionality.  Mismatched
 dimensions will raise a ``ValueError`` with context about the offending
 fragment and expected size.
+
+The batch path uses a vectorised matmul (``X @ X.T`` after L2
+normalisation) so a 10 000-fragment vault is dominated by BLAS, not Python
+loops. A FAISS-backed approximate-nearest-neighbour fallback is available
+behind :class:`DeduplicationConfig.use_faiss`; FAISS is an optional
+dependency and the code degrades gracefully to the dense matmul when it
+is not installed.
 """
 
 from __future__ import annotations
 
+import logging
 from typing import TYPE_CHECKING, Literal
 
 from pydantic import BaseModel, Field
 
 if TYPE_CHECKING:
     import numpy as np
+    from numpy.typing import NDArray
+
+    FloatArray = NDArray[np.float32]
+
+
+logger = logging.getLogger(__name__)
 
 
 def _cosine_similarity(vec_a: np.ndarray, vec_b: np.ndarray) -> float:
@@ -56,6 +70,36 @@ def _cosine_similarity(vec_a: np.ndarray, vec_b: np.ndarray) -> float:
         return 0.0
 
     return float(np.dot(vec_a, vec_b) / (norm_a * norm_b))
+
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+
+class DeduplicationConfig(BaseModel):
+    """Tunable knobs for :class:`SemanticDeduplicator`.
+
+    Threshold ranges are enforced inside :class:`SemanticDeduplicator`'s
+    constructor (not by Pydantic) so the validation error message stays
+    consistent across the keyword-argument and config-object code paths.
+
+    Attributes:
+        duplicate_threshold: Cosine similarity at or above which a pair
+            is classified as a duplicate.
+        resonance_threshold: Cosine similarity at or above which (but
+            below ``duplicate_threshold``) a pair is classified as a
+            resonance.
+        use_faiss: When ``True`` and the optional ``faiss`` package is
+            installed, use an approximate-nearest-neighbour index for
+            the batch all-pairs query. Falls back to the dense matmul
+            transparently when ``faiss`` is unavailable so this knob
+            never becomes a hard dependency.
+    """
+
+    duplicate_threshold: float = Field(default=0.95)
+    resonance_threshold: float = Field(default=0.75)
+    use_faiss: bool = False
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


Geoffe-Ga:
Thanks for the three reviews. Addressed the substantive items in `dd2f17a` and updated the PR title/description to reflect that this bundles PERF-001/003/004 + BUG-006, not just the marker change.

| Item | Verdict comment | Status | Where |
|---|---|---|---|
| PR title misleading | review #2 🔴 | **Updated** title to `perf(batch-e): vault writer + dedup + voice scaling fixes …` and rewrote the description with a per-commit table | PR meta |
| `_atomic_write_text` fixed temp name → cross-process collision | review #2 🟡 | Switched to `tempfile.mkstemp(dir=target_dir)`; added 8-thread × 50-write regression test | `creek/vault/writer.py` `_atomic_write_text` |
| `_JsonlWriter.append` never flushes | review #2 🟡 | Added per-append `flush()`; regression test reads file out-of-band between appends | `creek/generate/voice.py` `_JsonlWriter.append` |
| `_matmul_pair_indices` redundant `fill_diagonal` | review #2 🟢 | Removed; relies on `np.triu(k=1)` + the threshold > 0 invariant | `creek/clean/semantic_dedup.py` `_matmul_pair_indices` |
| `_pair_indices` swallows internal FAISS `ImportError` | review #2 🟢 | Split out `_faiss_available()` so only the top-level `faiss` import determines fallback; submodule errors propagate | `creek/clean/semantic_dedup.py` `_pair_indices` / `_faiss_available` |
| `_atomic_create` unbounded `while True` | review #3 🟢 | Capped at `_MAX_FILENAME_COLLISION_RETRIES = 10_000` with `RuntimeError`; regression test monkey-patches `os.open` to always collide | `creek/vault/writer.py` `_atomic_create` |

Skipped, with rationale:

- **Lock-scope optimisation in `_write_model`** (review #3 🟢): the current single-lock-around-the-whole-write is the documented serialisation contract; releasing mid-write would re-introduce the TOCTOU between dup-check and atomic create. Worth a follow-up issue with proper benchmarks if it surfaces as a real bottleneck.
- **`_find_existing` deprecation note** (review #1): the method is `_`-prefixed already and is exercised directly by tests; renaming is cosmetic.
- **`del body, fragment, loaded` in `_stream_into`** (review #2 🟢): explicit memory hint, already commented; no consensus that removing it is an improvement.

Local gates: `./scripts/test.sh --unit` 2 795 passed; `pytest -m slow` 3 passed; mypy strict clean; ruff clean; coverage 93.67%; pylint 9.75. Pushed as `dd2f17a`.

@claude please re-review

---

Geoffe-Ga:
Addressed the new round of feedback (reviews #4–#6) in `8effb2d`.

| Item | Verdict comment | Status | Where |
|---|---|---|---|
| `test_temp_jsonl_is_cleaned_up` is vacuous (scans `tmp_path` but `TemporaryDirectory` creates under OS temp root) | review #4 / #6 🔴 | **Rewrote** — monkey-patches `voice_mod.tempfile.TemporaryDirectory` to inject `dir=tmp_path`, captures the allocated path, asserts it is actually deleted after `generate_all_profiles` returns. The test now fails if the `with` clause is removed from `_streaming_accumulator`. | `tests/test_voice_profiles.py::test_temp_jsonl_is_cleaned_up` |
| `_matmul_pair_indices` docstring memory estimate 2× off | review #4 / #6 🟡 | **Fixed both code and docstring** — switched from `np.triu(matmul, k=1)` (which materialised a second N×N copy, peaking at ~800 MB) to `np.triu_indices` so only one N×N float32 matrix is live at a time; peak at N=10 000 is now actually ~400 MB. Docstring rewritten to match. | `creek/clean/semantic_dedup.py::_matmul_pair_indices` |
| `test_matches_pairwise_loop` float32/float64 dtype mismatch (seed-sensitive flake) | review #4 / #6 🟡 | **Aligned dtype** — reference loop now uses `np.asarray(..., dtype=np.float32)` end-to-end, matching the matmul path; comparison is bit-identical to the production code. | `tests/test_semantic_dedup.py::test_matches_pairwise_loop` |
| `_faiss_available()` logs fallback warning on every `find_duplicates` call | review #6 🟡 | **Cached** — added `_faiss_probe_cache: bool \| None` initialised in `__init__`; the probe runs once per deduplicator. New regression test runs 5 `find_duplicates` calls and asserts exactly 1 warning record. | `creek/clean/semantic_dedup.py::_faiss_available` + `tests/test_semantic_dedup.py::test_faiss_probe_cached_across_calls` |

### Skipped, with rationale

- **File mode `0o644` on index/provenance** (review #6 🟢 LOW): consistent with the surrounding `.md` files in the vault; switching just the metadata sidecars to `0o600` would be inconsistent and surprising. Worth a separate hardening pass that covers the whole vault uniformly.
- **`DeduplicationConfig` model_validator for inverted thresholds** (review #5 🟢 LOW): validation already happens at the `SemanticDeduplicator.__init__` boundary with a custom message that's covered by `TestThresholdValidation`; adding a Pydantic validator would change the exception type and break those tests for no behavioural gain.
- **`check_fragment` O(N) memory note** (review #6 🟢 LOW): informational; will land with the next docs sweep.
- **`_generate_filename` deprecation/removal** (review #6 🟢 LOW): production path uses `_atomic_create` directly; the legacy method is `_`-prefixed and only used by tests. Marking as deprecated would be cosmetic.
- **`_write_model` whole-section lock** (review #6 🟢 LOW + earlier reviews): same rationale as before — releasing mid-write reopens the TOCTOU between dup-check and atomic-create. Worth a follow-up benchmark-driven issue if it surfaces as a bottleneck.
- **Conventional-commit type for the previous review-fix commit** (review #4 minor): the previous commit used `review(...)` which is non-standard; this commit uses `fix(batch-e):` which is.

### Local gates

- `./scripts/test.sh --unit` → 2 796 passed (one new regression test added).
- `pytest -m slow` → 3 passed (10k dedup, 2k vault writer, voice memory).
- mypy strict, ruff lint+format, pylint 9.75 — all clean.
- Aggregate coverage 93.67%; per-file gate green.

Pushed as `8effb2d`. @claude please re-review.

---

Geoffe-Ga:
Addressed the latest CHANGES_REQUESTED in `f46285b`.

| Item | Severity | Status | Where |
|---|---|---|---|
| No migration from `provenance.json` → `provenance.jsonl` (existing vaults silently orphan history) | 🔴 BLOCKING | **Migrated.** `_log_provenance_locked` now calls `_migrate_legacy_provenance_locked` on first write — reads the legacy JSON array, appends each entry to the new JSONL log, and `unlink`s the legacy file so the migration is idempotent. Three regression tests: happy-path migration, idempotence across two writer instances, and corrupt-legacy tolerance. | `creek/vault/writer.py::_migrate_legacy_provenance_locked` + `tests/test_vault_writer.py::TestProvenanceLegacyMigration` |
| `_append_index_entry` POSIX `O_APPEND` atomicity assumption (only ≤ `PIPE_BUF`) | 🟡 MODERATE | **Bounded.** Added `_PIPE_BUF_BYTES = 4096` with a module-level docstring explaining why; both `_append_index_entry` and the new `_append_provenance_entry` reject lines that exceed it with a `ValueError("exceeds PIPE_BUF")`. Regression test feeds a 5 000-byte ID and asserts the error fires. | `creek/vault/writer.py::_append_index_entry` / `_append_provenance_entry` + `tests/test_vault_writer.py::test_oversized_index_entry_rejected` |
| Merge conflict risk with PR #193 on `_log_provenance` | 🔴 BLOCKING | **Out of scope here** — I cannot modify PR #193's branch. PR #193 (Batch C, PERF-002) rewrites `_log_provenance` to route through the new `AuditLog` primitive (hash-chained); this PR (Batch E) needed the JSONL O(1) append to make the constant-time scaling test pass. The two implementations should reconcile when PR #193 lands; whichever PR merges second will need to swap its `_log_provenance` body to call `AuditLog`. The merge order is a user decision; flagging in the PR description for the maintainer. |

### Skipped LOW notes (with rationale)

- `_streaming_accumulator` cleanup-comment note (review #7 🟢 LOW): cleanup is provided by `tempfile.TemporaryDirectory.__exit__`, which is the language contract — additional commenting would restate Python documentation.
- `_find_existing` "backwards compatibility" framing (review #7 / #8 🟢 LOW): purely cosmetic; the method is `_`-prefixed and exercised only by tests.
- `check_fragment` O(N) per-call rebuild note (review #8 🟢 LOW): worth a separate docs pass; doesn't affect this PR's goal.
- `FloatArray` TYPE_CHECKING comment (review #8 🟢 LOW): consistent with the existing `numpy` import pattern in the file; deferring would not change behaviour.
- `test_concurrent_distinct_writes_no_loss` JSONL integrity assertion (review #7 🟡): the existing `_read_provenance` already raises on malformed lines, so the regression *is* caught — only the diagnostic message could be friendlier. Worth a follow-up touch-up; not blocking.
- Lock-scope optimisation in `_write_model` (multiple reviews 🟡 / 🟢): same answer as before — releasing mid-write reopens the TOCTOU between dup-check and atomic-create. Real performance issue, but it requires a redesigned critical section, which deserves its own benchmark-driven follow-up rather than a stealth change inside this PR.

### Local gates

- `./scripts/test.sh --unit` → 2 800 passed, 19 deselected.
- `pytest -m slow` → 3 passed.
- mypy strict, ruff lint+format clean. Coverage 93.63% aggregate; per-file gate green.

Pushed as `f46285b`. @claude please re-review.

---

Geoffe-Ga:
All four LGTMs from the previous review round noted small follow-up items. Folded the substantive ones into `560e4e8`.

| Item | Severity | Status | Where |
|---|---|---|---|
| `ValueError` from `_append_provenance_entry` can escape the migration loop and block vault upgrade | 🟡 HIGH (non-blocking) | **Fixed.** Split `_migrate_legacy_provenance_locked` into a header (always unlinks the legacy file in `finally`, sets a `_legacy_provenance_migrated` flag) and a body (`_migrate_legacy_provenance_body`) that catches `ValueError` per-entry. A single oversized legacy row can no longer wedge the vault. Regression test seeds a 5 000-byte legacy `id` alongside a well-formed entry and asserts the well-formed one survives, the user write succeeds, and `provenance.json` is gone. | `creek/vault/writer.py::_migrate_legacy_provenance_locked` + `_migrate_legacy_provenance_body` + `tests/test_vault_writer.py::test_migration_drops_oversized_entry_without_blocking_writes` |
| `_migrate_legacy_provenance_locked` stats `provenance.json` on every write | 🟢 LOW | **Cached.** New `_legacy_provenance_migrated: bool` flag on the writer instance, set after the first attempt (success or graceful fallback). Steady-state per-write cost is one boolean check. | `creek/vault/writer.py::__init__` |
| `_PIPE_BUF_BYTES` docstring overstates POSIX portability | 🟢 LOW | **Rewrote.** POSIX's PIPE_BUF atomicity contract applies to pipes/FIFOs only; the regular-file behaviour we rely on is a Linux VFS implementation detail, not a portable POSIX guarantee. The 4 096-byte cap is now framed as "conservative across platforms". | `creek/vault/writer.py::_PIPE_BUF_BYTES` |
| `_find_existing` / `_generate_filename` "backwards compatibility" framing misleading | 🟢 LOW | **Reframed** to "retained for tests and tooling that introspect the writer" with a pointer to `_atomic_create` for new callers. Both methods are `_`-prefixed; "backwards compatibility" was inaccurate. | `creek/vault/writer.py::_find_existing` / `_generate_filename` |

### Skipped, with rationale

- **`os.write` partial-write loop** (review 06:34:46Z 🟢 LOW): blast radius is a dangling empty file, not silent data loss — the JSONL index entry is appended *after* `_atomic_create` returns, so a partial fragment write leaves no orphaned index pointer. Linux local FS effectively guarantees full writes. Worth revisiting when NFS support becomes a goal.
- **Implicit invariant in `_write_model`** (review 07:24:58Z 🟢 LOW): `_load_index_locked` is the only path that populates the cache and is called by `_find_existing_locked` immediately above the assignment. The invariant is real but already enforced by control flow; an `assert` would be cosmetic.
- **Direct FAISS unit test** (review 06:34:46Z 🟢 minor): would only run in environments that install the optional FAISS dep; the existing fallback test via monkey-patched `__import__` is sufficient for our test stack.
- **"Duplicate Path import in voice.py"** (review 06:34:46Z 🟢 minor): false alarm — ruff is clean, `Path` is imported once at module top.
- **macOS / NFS portability of `_PIPE_BUF_BYTES`** (review 06:31:34Z 🟢 LOW): noted in the new docstring; concrete platform support beyond Linux is out of scope for Batch E.

### Local gates

- `./scripts/test.sh --unit` → 2 801 passed, 19 deselected.
- `pytest -m slow` → 3 passed.
- mypy strict, ruff lint+format clean. Coverage 93.67%; per-file gate green.

Pushed as `560e4e8`. @claude please re-review.

---

Geoffe-Ga:
Addressed the three required items in `daf04ba`.

| Item | Severity | Status | Where |
|---|---|---|---|
| Implicit coupling: `self._dir_indexes[target_dir]` access in `_write_model` is fragile | 🔴 BLOCKING | **Fixed.** Replaced with `self._load_index_locked(target_dir)` (idempotent — cache hit returns immediately, so the steady-state cost is unchanged). A future refactor that skips the dup-check branch can no longer trip a `KeyError`. | `creek/vault/writer.py::_write_model` |
| Wide lock scope undocumented | 🔴 BLOCKING | **Documented.** Added a 7-line block comment on the `with self._lock:` line explaining that the wide critical section (dup-check + file create + index append + provenance append) is intentional, that narrowing it would re-introduce the BUG-006 TOCTOU, and pointing future contributors at the benchmark-driven follow-up issue rather than letting them silently shrink the lock. | `creek/vault/writer.py::_write_model` |
| `.id-index.json` typo in module docstring | 🟢 LOW | **Fixed.** Module-level docstring now says `.id-index.jsonl`, matching the `INDEX_FILENAME` constant. | `creek/vault/writer.py` (module docstring, line 9) |

### Skipped, with rationale

- **Path traversal hardening on `_atomic_create`** (🟢 LOW): defence-in-depth. `_sanitize_title` already strips `/` and unsafe characters; an `assert candidate.parent == target_dir` would be belt-and-suspenders. Worth filing as a follow-up issue but doesn't change behaviour today.
- **`0o644` mode configurability** (🟢 LOW): per-vault file-mode policy is its own design conversation; matches the existing `.md` files' permissions in practice.
- **`FloatArray` outside `TYPE_CHECKING`** (🟢 LOW): functions annotated with it use `from __future__ import annotations`, so runtime `get_type_hints()` would `NameError` only if a caller asks for resolved hints. No production caller does. Moving the alias would be a style preference; not blocking.
- **`_JsonlWriter` context-manager interface** (🟢 LOW): the only caller is `_streaming_accumulator`, which already wraps it in a `try/finally`. A second caller would justify the refactor; today's call site is self-correcting.
- **`_atomic_create` called via `self.` despite being `@staticmethod`** (🟢 LOW): consistent with the rest of the codebase's style; ruff/pylint don't object.
- **`find_duplicates` with `len == 1` test** (🟢 LOW): covered by `test_single_fragment` in `TestBatchMode`. Not a gap.
- **`_faiss_pair_indices` "Falls back" wording** (🟢 LOW): minor wording; the docstring's intent (FAISS path is exact, not approximate) is conveyed.
- **`tracemalloc` Python-heap-only note** (🟢 LOW): true, but the test guards against the regression we care about (re-introducing the load-everything pattern). Worth a docstring note, not blocking.

### Local gates

- `./scripts/test.sh --unit` → 2 801 passed, 19 deselected.
- `pytest -m slow` → 3 passed.
- mypy strict, ruff lint+format clean. Coverage 93.67%; per-file gate green.

Pushed as `daf04ba`. @claude please re-review.

---

Issue #270, opened 2026-05-23
FEAT-029: Dynamic voice-register switching during a CrawDad session (FEAT-015 deferral)


Geoffe-Ga:
## Context

PR #234 (FEAT-015, CrawDad v1.0 composer + agent loop) explicitly deferred runtime voice-register switching to FEAT-016:

&gt; Dynamic register switching via `activate_register` intent is deferred (the AC requires only session-start loading; FEAT-016 territory).

PR #237 (FEAT-016 slash commands) **did not** ship this capability — backlog-grooming verification (2026-05-23):

- `grep -rn "activate_register" creek-tools crawdad` → zero hits.
- `crawdad/crawdad/cli.py:92-95` still says, in present-tense future, *"FEAT-015 adds the composer + voice-skill stack to the wiring. The skill stack is loaded once at session start from `/creek-skills/` per the FEAT-015 plan; FEAT-016 **will** introduce dynamic register switching."*
- The only "register" hits in tests are `extra_registers=(...)` (session-start loading), not runtime switching.

So dynamic register switching is unimplemented and has no tracking issue. Filing this so it's not lost.

## Goal

Let the operator change CrawDad's active voice register mid-session without restarting the bot. Example: starting in `confessional` and switching to `analytic` when the conversation shifts mode.

## Requirements

- New intent type `activate_register` in `crawdad/crawdad/intents.py`, with a `register: str` field naming the target register skill.
- Router (`crawdad/crawdad/router.py`) recognises the user-message patterns that mean "switch register" (Haiku-side classification, similar to other intents).
- Dispatcher (`crawdad/crawdad/dispatcher.py`) handles the intent by re-loading the voice-skill stack from `/creek-skills/` with the new register name, replacing the active stack in place.
- Slash-command surface: optionally expose `/crawdad register ` for explicit switching alongside the conversational form.
- Allowlist: only registers that exist as skill files under `/creek-skills/register/` are accepted; unknown names produce a soft error (no crash).
- Persistence is **not** required — register switches are session-scoped.
- Update the docstring in `crawdad/crawdad/cli.py:92-95` to past-tense once shipped.

## Acceptance Criteria

- [x] `activate_register` intent recognised by the router for natural-language phrases ("switch to analytic register", "use confessional voice", etc.).
- [x] Dispatcher re-loads `skill_loader.load_skills_for_session(...)` with the new register name; subsequent composer turns use the new stack.
[...]

PR #280, opened 2026-05-23
feat(crawdad): Discord attachment ingest gate with creek.redact.scan (FEAT-027)

### LINKED ISSUE #274: FEAT-033: CrawDad accepts Discord file attachments and routes them through the ingest pipeline
> _**Renumbered 2026-05-23 (backlog grooming):** was FEAT-027, now FEAT-033. Bumped to resolve a namespace collision with #264 (FEAT-027 cross-source aggregation in the atomize pipeline). The classification/atomize-epic issue (#264) retains FEAT-027; the CrawDad-epic issue is bumped into the FEAT-030+ range. The issue number (#274) is unchanged. Note: merged PR #280 retains its original "FEAT-027" title as historical record._

## Context
CrawDad's value proposition is *"Discord is the door you already use."* Today that promise has a hard gap: CrawDad does not handle Discord message attachments. The full creek pipeline is already exposed through the MCP surface (`creek.ingest`, `creek.classify`, `creek.link`, `creek.compile`, `creek.mine`, `creek.draft`, `creek.save`, ...) and the Haiku→Sonnet agent loop can dispatch to any of it, but the bot code does not read `message.attachments`. So dropping a file into chat ("here's my journal export — ingest these notes") silently does nothing.

This forces the user back to the terminal for the one workflow they most expect to do conversationally: *adding source material to their vault*.

## Requirements
- **Read attachments**: when a message arrives in an allowlisted channel from an allowlisted user, the bot inspects `message.attachments` for any uploaded files.
- **Staging path**: each attachment downloads to a deterministic staging directory under the vault — e.g. `<vault>/00-Creek-Meta/Inbound/<channel-id>/<message-id>/`. Never directly into `01-Fragments/`; the safety pass + ingestor decide what (if anything) flows into the vault proper.
- **Safety pass first, always**: before any ingest, run `creek redact --scan` over the staging directory and surface the report inline in the Discord reply. No `--apply` and no ingest without explicit user consent (a follow-up Discord message or button click).
- **Intent inference**: the Haiku router takes both the user's text and the attachment list, and may choose `creek.ingest` with an inferred `--type` (chat export, document, markdown, image, etc., per the existing INGESTOR_REGISTRY). When the type is ambiguous, the bot asks inline rather than guessing.
- **Idempotency**: the same file uploaded twice does not re-ingest. Deterministic fragment IDs make this safe; the bot should also recognise and reply "already ingested" rather than churning.
- **Multiple attachments**: a single message with N attachments is treated as one ingestion batch when types align (e.g. five HTML files of the same export), or as N separate ingestions when they don't.
- **Size and type limits**: configurable max attachment size (default 25 MB — Discord's limit); a configurable allow/deny extension list (deny by default for binaries the system can't handle).
- **Privacy posture**: every downloaded attachment carries a privacy tier inherited from the channel or message context (configurable per channel in `crawdad.yaml`).
- **Errors**: any failure (download timeout, ingestor unavailable, safety pass blocks) returns a clear, calm Discord message — never silent.

## Why this matters
- **Discord-as-door becomes real.** Without attachment handling, users still have to open a terminal to add anything to the vault. With it, the conversational layer covers the whole pipeline end-to-end.
- **The safety pass becomes embedded in the everyday loop**, not a thing the user has to remember to run separately.
- **Lowers the floor of contribution.** Phones, tablets, anywhere Discord runs — anything can become a fragment source by dropping a file in chat.

## Acceptance Criteria
- [ ] The bot reads `message.attachments` on every allowlisted message.
- [ ] Each attachment downloads to a deterministic staging path under `<vault>/00-Creek-Meta/Inbound/`.
- [ ] A `creek redact --scan` report is included in the Discord reply before any apply or ingest happens.
- [ ] Ingestion only proceeds on explicit user consent (reply / button).
- [ ] Idempotent re-uploads return a "already ingested" message rather than re-running.
- [ ] Per-attachment size and extension limits are enforced and configurable.
- [ ] Tests cover: single attachment, multiple attachments same type, multiple attachments mixed types, oversized, denied extension, redaction findings present, redaction findings absent.
- [ ] Tests with >=90% branch coverage.
- [ ] MyPy strict, Ruff zero violations.

## Out of scope
- Voice/audio attachments — file a follow-up that pairs an audio ingestor with this gate.
- Discord thread replies as ingestion conversations — file separately.
- Cross-server channel handling — single server for v1.

Geoffe-Ga:
## Summary

Closes #274 (FEAT-027) — CrawDad accepts Discord file attachments and routes them through the ingest pipeline.

CrawDad's promise — *"Discord is the door you already use"* — required closing a hard gap: the bot ignored `message.attachments` entirely, so dropping a file in chat silently did nothing. This PR wires the attachment path end-to-end with a deterministic safety pass.

**CrawDad bot side (`crawdad/`)**
- New `crawdad.attachments` module: per-channel / per-message staging path under `00-Creek-Meta/Inbound/`, configurable size + extension limits, filename sanitisation, content-hash idempotency (re-upload of identical bytes returns "already staged"), and extension → ingestor-type inference matching `INGESTOR_REGISTRY`.
- `CrawDadConfig.attachments` (Pydantic) carries the limits and optional per-channel privacy tier overrides; refuses absolute `staging_subpath` so attachments cannot escape the vault.
- `handle_message` short-circuits the agent loop on turns that carry attachments: download → `creek.redact.scan` via MCP → reply with download summary, scan report, and an explicit "I did **not** ingest anything" consent prompt.

**MCP tool side (`creek-tools/`)**
- New `creek.redact.scan` MCP tool wrapping `RedactionScanner.scan_batch` over a vault-relative directory. Read-only — never invokes the redactor, never returns matched text (only salted SHA-256 hashes + locations + severity).
- `_resolve_within_vault` validates the resolved path (symlinks followed) sits inside the vault before reading anything.
- Registered in `server.py` and `EXPECTED_TOOLS`; full audit logging via `MCPAuditLog`.

## Acceptance criteria (all met)

- [x] Reads `message.attachments` on every allowlisted message.
- [x] Each attachment lands at the deterministic staging path under `/00-Creek-Meta/Inbound/<channel_id>/<message_id>/`.
- [x] `creek redact --scan` report is included in the Discord reply before any apply or ingest happens.
- [x] Ingestion only proceeds on explicit user consent (the bot never auto-ingests; the reply explicitly says "I did **not** ingest anything").
- [x] Idempotent re-uploads return "already staged" rather than re-running the scan.
- [x] Per-attachment size and extension limits are configurable in `crawdad.yaml`.
- [x] Tests cover: single attachment, multiple same-type, multiple mixed-type, oversized, denied extension, redaction findings present, redaction findings absent.
- [x] ≥90% branch coverage (crawdad 93.5%, creek-tools 93.5%).
- [x] MyPy strict + Ruff clean across both packages; Bandit zero medium+ findings.

## Test plan

- [ ] CI green on both `crawdad/` and `creek-tools/` jobs.
- [ ] Verify `crawdad/scripts/check-all.sh` exits 0 in the configured venv (mypy, ruff, security, complexity, docstrings, tests + coverage).
- [ ] Verify `creek-tools/scripts/check-all.sh` exits 0 (lint, format, typecheck, security, complexity, tests, per-file coverage gate).
- [ ] Manual smoke: drop a `.md` and a `.exe` into an allowlisted channel — the `.md` lands at the staging path with a scan report; the `.exe` is rejected with "extension not allowed".
- [ ] Manual smoke: re-upload the same `.md` — bot replies "already staged" without re-scanning.

## Out of scope (per the issue)

- Voice/audio attachments (separate follow-up).
- Conversational thread-replies as ingestion conversations.
- Cross-server channel handling.

The "Reply with `ingest`" consent flow today routes through the existing agent loop without pending-batch awareness. The reply mentions both the conversational and CLI paths so users have a clear escape hatch. Full conversational consent flow is intentionally a follow-up.

https://claude.ai/code/session_01RHdi5Cep2Hne77Mka1sTEv

---
_Generated by [Claude Code](https://claude.ai/code/session_01RHdi5Cep2Hne77Mka1sTEv)_


Geoffe-Ga:
## Review feedback addressed in 4cb809e

Thanks for the thorough three-round review — consolidated the actionable items and addressed every blocker and high. Summary of what changed:

### 🔴 Blockers (all fixed)

| # | Issue | Fix |
|---|---|---|
| 1 | `session_state is None` gate fired before the attachment branch — users dropping files during degraded state got "session unavailable" instead of staging | Reordered `handle_message`: allowlist → attachment branch → session-state → loop deps. Renamed `_passes_preflight` to a pure `_passes_allowlist` (no side effects). Regression test: `test_attachment_path_runs_when_session_state_is_none` |
| 2 | `_channel_tier` re-extracted `channel_id` from the filesystem path | `channel_id: int` is now threaded explicitly from `_handle_attachments` → `_run_safety_scan` → `_channel_tier`. Path-string extraction gone. Test: `test_channel_tier_default_falls_back_to_personal` |
| 3 | Non-`MCPUnavailableError` exceptions escaped `_run_safety_scan` | Added a broad `except Exception` with `_LOGGER.exception` (operator sees full traceback) + soft error reply. Test: `test_run_safety_scan_swallows_unexpected_exceptions` |
| 4 | `asyncio.run()` inside a sync test | Converted to `async def` matching every other async test in the file |

### 🟡 Highs (all fixed)

- **Size gate trusted Discord-reported metadata** → added post-download `len(data) > max_size_bytes` check in `_process_one`. Test: `test_post_download_size_check_rejects_underreported_size`.
- **Raw MCP response body forwarded to Discord** → new `_format_scan_section` parses dict responses and extracts `report_markdown`, falls back to the code-fenced raw section for plain text. Tests: `test_run_safety_scan_extracts_report_markdown_from_dict_response`, `test_run_safety_scan_falls_back_for_unparseable_dict_response`.
- **Consent flow had no follow-up issue** → filed [#281 (FEAT-028)](https://github.com/Geoffe-Ga/Creek-Vault/issues/281) and linked it in an in-code TODO near `_INGEST_CONSENT_PROMPT`.

### 🟢 Quality nits (fixed)

- `channel_privacy_tiers` now validates tier strings via `field_validator` (refuses `"superopen"` etc.). Tests: `test_attachment_config_rejects_unknown_privacy_tier`, `test_attachment_config_accepts_all_four_tier_values`.
- `ProcessedAttachments.accepted` and `rejected` now have symmetric `default_factory=tuple` defaults.
- Inline `from crawdad.attachments import ProcessedAttachments` removed from test functions; consolidated at the top.
- `assert session_state is not None` is gone — the gate reorder makes the type narrowing fall out naturally.

### Pushed back on (with justification)

- **`.hidden` filename extension preservation** — cosmetic; preserving leading-dot files would complicate `sanitize_filename` for marginal benefit. The fallback to `attachment.bin` is safe.
- **Per-message attachment count limit** — Discord caps at 10 by default; no clear product reason to cap below that. Total bytes are bounded by `10 × max_size_bytes` which is the same as 10 separate messages.
- **`load_config()` per `redact_scan_tool` invocation** — consistent with every other tool in `creek_mcp.tools.*`. Optimization-only follow-up.
- **Duplicate `_FakeAttachment` across test files** — kept for test isolation. The shapes differ slightly because each test file exercises a different surface.
- **Discord-metadata size check looking like a redundant pre-filter** — kept as the cheap pre-filter so we don't even initiate the download for obvious oversize cases. The post-download check is the authoritative gate.

### Quality gates

- crawdad: 186 tests, 93.82% branch coverage, 7/7 gates ✅
- creek-tools: 3667 tests, 93.52% branch coverage, 12/12 gates ✅

@claude please re-review.

---

Geoffe-Ga:
## All review nits addressed in 244cb66

Thanks for the second-round COMMENTS review. Both the LGTM and COMMENTS verdicts came in for the previous HEAD (4cb809e) within minutes of each other — addressing every actionable item from both, plus the docstring nit from the LGTM:

### Fixed

| Source | Item | Fix |
|---|---|---|
| COMMENTS #1 | Misleading `**Attachments staged at**` header on all-rejected uploads | Header is now `**Attachments** (would stage to ...)` when nothing was accepted. Test asserts the new copy. |
| COMMENTS #3 | `staging_subpath` validator missed `..` parent traversal | Added a parts-based ``..`` check with its own clear error message. Test: `test_config_refuses_parent_traversal_in_staging_subpath`. |
| COMMENTS #4 | Inline `from crawdad.mcp_client import MCPUnavailableError` in `test_handle_subprocess_unavailable_replies_gracefully` | Consolidated to top-level imports in `test_bot.py`. |
| LGTM docstring nit | `channel_privacy_tiers` docstring listed only 3 of 4 valid tiers | Added `all` and clarified that the validator runs at config-parse time. |

### Filed as follow-up

- **COMMENTS #2 — Extension-only gating, no MIME / magic-byte check** → [#285 (FEAT-029)](https://github.com/Geoffe-Ga/Creek-Vault/issues/285). Acceptable for v1 given the consent-first + safety-pass guarantees, but tracked for v2 so the gap doesn't rot.

### Pushed back on / kept as-is

- **`_FakeAttachment` duplicated across `test_attachments.py` and `test_bot.py`** — kept for test isolation. The shapes differ (one has a `fail` field for download-error scenarios) and consolidation into `conftest.py` would couple unrelated test surfaces.
- **`_EXTENSION_TO_INGESTOR_TYPE` ↔ `_DEFAULT_ALLOWED_EXTENSIONS` drift** — the reviewer correctly flagged that adding to one without the other yields "unknown type" silently. Acceptable behaviour (bot stages the file, replies with "unknown type", user picks). Hard-coupling them would prevent legitimate cases like staging `.csv` without an ingestor mapping (`spreadsheet` exists today but a new ingestor type might be wanted later).
- **`_format_scan_section` JSON heuristic** — brace-detection is fragile in principle but harmless in practice: the `json.JSONDecodeError` catch + `isinstance(..., str)` guard handle every non-conforming case correctly. Both edges are unit-tested.

### Quality gates

- crawdad: 187 tests, 93.85% branch coverage, 7/7 gates ✅
- creek-tools: unchanged (3667 tests passing, 93.52% coverage)

@claude please re-review.

---

Geoffe-Ga:
## HIGH consent-prompt truncation fixed in e7702a6

Thanks — that was a real gap. The safety invariant held (the router never fires on attachment turns, so no auto-ingest could happen), but you're right that the user-facing trust signal could disappear silently.

### Fix

`_handle_attachments` now sends the consent prompt as a separate, never-truncated Discord message:

```python
# Before:
reply = "\n\n".join((summary, scan_section, _INGEST_CONSENT_PROMPT))
await message.channel.send(_truncate_for_discord(reply))

# After:
body = "\n\n".join((summary, scan_section))
await message.channel.send(_truncate_for_discord(body))
await message.channel.send(_INGEST_CONSENT_PROMPT)
```

The body is still truncation-capable (for very long scan outputs), but the short consent string is always delivered on its own send.

### Regression test

Added `test_attachment_reply_consent_prompt_survives_long_scan_output` that stubs the MCP response body at 3000+ chars and asserts `"did **not** ingest anything"` survives in the channel output. Verified the test fails on the old code (consent gets truncated away) and passes on the new code.

### Updated existing tests

The seven attachment-path tests that asserted `len(channel.sent) == 1` now expect 2 messages (body + consent) and verify the consent string is present in the second send. Single-message paths (all-rejected, all-already-staged, non-attachment agent-loop) are unchanged — they still send exactly one message.

### Quality gates

- crawdad: 188 tests, 93.85% branch coverage, 7/7 gates ✅
- creek-tools: unchanged

@claude please re-review.

---

Geoffe-Ga:
## Merged `main` into the branch in 22d216b

`main` moved during the review iteration: FEAT-029 (dynamic voice-register switching, #270 / #286) landed and touched `crawdad/crawdad/bot.py` at the same location as my `_handle_attachments` / `_run_safety_scan` helpers. Resolved the conflict by keeping both — main's `_resolve_skills` helper goes after my attachment helpers.

### Numbering note

My earlier follow-up issue (#285, originally labelled FEAT-029 for MIME-type verification) collided with main's FEAT-029. I renamed it to **FEAT-030** to clear the conflict. The in-code TODO near `_INGEST_CONSENT_PROMPT` still references issue #281 (FEAT-028 conversational consent flow), which is unchanged.

### Verification

After the merge:

- crawdad: 210 tests, 94.21% branch coverage, 7/7 gates ✅ (the suite picked up the 22 new tests that came in with main's FEAT-029)
- creek-tools: all 12 gates ✅

Per the merge gate, the merge commit supersedes the previous LGTM — `@claude please re-review`.

---

Issue #255, opened 2026-05-23
FEAT-024: Hierarchy-aware linking — suppress trivial parent/child and adjacent-sibling resonances


Geoffe-Ga:
## Context
With FEAT-020 (#251, hierarchy fields) and FEAT-021/022/023 (#252/#253/#254, the re-atomization operators and orchestrator) in place, the embedding linker will start seeing parent fragments AND their children in the same vault. Today the linker treats every fragment as a peer and emits a `resonance` edge whenever cosine similarity ≥ threshold. After re-atomization that becomes noise:

- A parent fragment and its own child are guaranteed to be highly similar — that's not a resonance, it's the same text at two granularities.
- Adjacent children of the same parent (sentence-after-sentence in a paragraph, exchange-after-exchange in a session) are often topically continuous — also not a real resonance.
- Cross-level matches (a sentence-level child and an exchange-level parent in a different source) may be useful, but should carry the level information so downstream consumers can decide.

This issue makes the linker hierarchy-aware so resonances stay meaningful. Referenced as the deferred follow-up in FEAT-020's out-of-scope: *"Linker awareness of parent/child relationships (siblings shouldn't trivially resonate) — to be filed once FEAT-020..023 land."*

## Requirements
- `EmbeddingLinker` (and the `temporal` / `eddies` linkers as relevant) skip pairs where:
  - One fragment is an ancestor of the other (via `parent_id` chain).
  - Both fragments share a `parent_id` and are within K positions of each other in the parent's `child_ids` (K configurable, default 2).
- Resonance edges record the levels of both endpoints (e.g. `from_level: sentence`, `to_level: exchange`) so synchronicity and mining strategies can filter by level.
- New `LinkingConfig` knob: `hierarchy_sibling_skip_window: int` (default 2) — how many adjacent siblings to suppress on either side.
- Synchronicity detection (`creek report --type synchronicity`) gains an additional criterion: source-different is already required; add level-asymmetry-aware ranking so cross-level synchronicities surface deliberately, not as accidents.
- Existing flat-fragment behaviour is unchanged: when no hierarchy exists, the linker behaves exactly as it does today.

## Acceptance Criteria
- [ ] Resonance edges between ancestor/descendant pairs are not produced.
- [ ] Resonance edges between adjacent siblings (within `hierarchy_sibling_skip_window`) are not produced.
- [ ] Resonance edges record `from_level` and `to_level`.
[...]

PR #289, opened 2026-05-23
feat(link): hierarchy-aware resonance suppression (FEAT-024) (#255)

Geoffe-Ga:
Make the embedding linker hierarchy-aware so re-atomized vaults no
longer emit trivial parent/child or adjacent-sibling resonances. Each
resonance now carries the structural level of both endpoints, and the
synchronicity detector ranks cross-level pairs ahead of same-level
pairs so level-asymmetric matches surface deliberately.

- New ``LinkingConfig.hierarchy_sibling_skip_window`` (default 2).
- New ``Resonance`` Pydantic model with ``from_level`` / ``to_level``.
- ``EmbeddingLinker.find_resonances`` accepts a fragment lookup and
  skips ancestor/descendant pairs plus adjacent-sibling pairs within
  the configured window. Flat-fragment vaults are unchanged.
- ``Synchronicity`` gains ``level_a`` / ``level_b``; detector orders
  cross-level pairs first, then by similarity descending.
- Synchronicity detector accepts both ``Resonance`` and the legacy
  3-tuple format for transitional compatibility.

https://claude.ai/code/session_01TqnfRx7cXbE1HYq7qXZHXQ

### DIFF
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -117,6 +117,20 @@ class LinkingConfig(BaseModel):
     same ``session`` by the FEAT-022 aggregator. Inclusive boundary.
     """
 
+    hierarchy_sibling_skip_window: int = Field(default=2, ge=0)
+    """FEAT-024 sibling-suppression window for hierarchy-aware linking.
+
+    With re-atomization (FEAT-020..023) every parent fragment shares the
+    vault with its own children, and adjacent children of one parent are
+    usually topically continuous — so naive cosine similarity emits noise
+    instead of meaningful resonances. The embedding linker drops pairs
+    that are ancestor/descendant unconditionally, and pairs that share a
+    parent and sit within this many positions of each other in the
+    parent's ``child_ids`` list. ``0`` disables sibling suppression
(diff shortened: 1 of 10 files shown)


Geoffe-Ga:
Addressed in 718f005:

- **BLOCKING (type-ignore in test helpers)** — `_hier_fragment` and `_level_fragment` now take `level: FragmentLevel`. The two `# type: ignore[arg-type]` suppressions are gone; `./scripts/typecheck.sh` is clean.
- **HIGH (`| object` in `_normalise_resonance`)** — signature is now `Resonance | tuple[str, str, float]`. The fall-through branch logs a `WARNING` instead of swallowing silently, and `TestUnrecognisedResonanceShape.test_alien_shape_is_skipped_with_warning` covers that runtime path (no `type: ignore` — uses `cast` to thread an alien object through the now-strict signature).

Code-quality nits (warning log was bundled with HIGH; the `sibling_skip_window` default duplication and `fragments_by_id` duplication notes I'd like to defer to a follow-up issue — both are pure refactors with no behaviour change). All 12 `check-all.sh` gates green.

@claude please re-review

---

Issue #264, opened 2026-05-23
FEAT-027: Cross-source aggregation in the atomize pipeline


Geoffe-Ga:
## Context

PR #261 (FEAT-022 conversational aggregator) explicitly scoped cross-source aggregation as **out of scope** for v1:

> Out of scope — Cross-source aggregation (only fragments sharing a `source` are grouped in v1).

Today `creek/atomize/aggregate.py` groups fragments only when they share the same `source`. In practice a single line of thought often spans several sources — a Discord thread that continues into a ChatGPT conversation, a journal entry whose follow-up landed on Substack, etc. The current operator cannot stitch those into a shared parent, even when content and time would justify it.

This issue tracks the v2 extension that lets the aggregator consider cross-source fragments when their content and timing warrant it.

## Requirements

- Extend `aggregate(fragments, *, level, config)` to accept a `cross_source: bool` knob (default `false`).
- When `cross_source=true`:
  - The exchange/burst/session boundary heuristics use the existing temporal and similarity thresholds, but the source-identity gate is dropped.
  - The resulting parent fragment records all contributing sources in its metadata (e.g. `sources: ["discord:#general", "chatgpt:conversation-42"]`).
  - Provenance is preserved: each child still points back to its origin source via the existing fragment fields.
- New config knob `linking.cross_source_aggregation: bool` (default `false`) flips the behaviour for whole-pipeline runs.
- The existing single-source path remains the default and is unchanged.

## Acceptance Criteria

- [ ] `aggregate(..., cross_source=False)` behaviour matches today.
- [ ] `aggregate(..., cross_source=True)` produces parents that span multiple sources when the temporal and similarity thresholds permit.
- [ ] Parent metadata records every contributing source.
- [ ] Idempotent: re-running with the same flag yields identical parents.
- [ ] Property test: enabling `cross_source` never produces fewer parents than the single-source mode on the same input (it only merges further).
- [ ] Tests with >=90% branch coverage on the new branch.
- [ ] MyPy strict, Ruff zero violations.

## Out of scope

- Changing the default to `true`.
- Re-classifying existing single-source parents in the vault.

## References

- PR #261 (FEAT-022) — original aggregator and the explicit out-of-scope clause.
- `creek-tools/creek/atomize/aggregate.py`.

## Surfaced by

Backlog grooming session 2026-05-23. Originally flagged as out-of-scope in PR #261.

Geoffe-Ga:
[...]

PR #291, opened 2026-05-23
feat(atomize): cross-source aggregation knob (FEAT-027, #264)

Geoffe-Ga:
Extend the conversational aggregator with a ``cross_source: bool``
parameter (default False). When True, the exchange/burst/session
boundary heuristics drop the source-identity gate so a single parent
can span multiple platforms/channels; the parent records every
contributing source key as a ``source/<key>`` tag while child fragments
retain their original ``FragmentSource`` pointers. Cross-source parent
IDs derive from a stable ``cross-source:`` prefix joining every
distinct source key so re-runs stay idempotent.

Adds ``LinkingConfig.cross_source_aggregation`` (default False) so the
behaviour can be flipped from ``creek_config.yaml`` for whole-pipeline
runs. The single-source path remains the unchanged default.

Covered by new unit tests, idempotency tests, helper tests, and a
parametrised property test asserting that at session level
(gap-only, no speaker / similarity gate) cross-source mode never
produces more parents than single-source mode on the same input.

https://claude.ai/code/session_013wQEnMZJfRRrkAZbG7SxUp

### DIFF
--- a/creek-tools/creek/atomize/aggregate.py
+++ b/creek-tools/creek/atomize/aggregate.py
@@ -80,6 +80,7 @@ def aggregate(
     *,
     level: AggregateLevel,
     config: AggregationConfig,
+    cross_source: bool = False,
 ) -> list[Fragment]:
     """Aggregate fragments upward to the requested ``level``.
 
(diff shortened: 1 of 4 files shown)


Geoffe-Ga:
@claude Can I get a review, please?

Geoffe-Ga:
Thanks for the review!

**Issue 2 (stale `_inherit_source` docstring)** — fixed in 7f26a1a. The new docstring distinguishes single-source (FEAT-022, callers guarantee a shared source key) from cross-source (FEAT-027, first child's source is a representative anchor while full provenance is captured in `source/<key>` tags and per-child `FragmentSource` pointers).

**Issue 1 (dead config field — `cross_source_aggregation` never read by the pipeline)** — taking your Option B and deferring. This PR's scope is the atomize-layer surface only (`creek/atomize/aggregate.py`, `creek/config.py`, tests); the call-site wiring at `creek/classify/reatomize.py:487` is outside that scope and will land in a follow-up (FEAT-027 E2: "wire `cross_source_aggregation` config flag through reatomize → aggregate"). The knob is intentionally exposed in `LinkingConfig` first so the follow-up only needs to add the call-site argument — no further schema churn. I'll file the follow-up issue alongside the merge.

The two non-blocking observations (singleton source tags, exchange-vs-cross sort tie-breaks) are accurate; I've left them as-is to keep blast radius small. If either turns out to matter to downstream consumers, happy to address in the follow-up.

---

PR #211, opened 2026-05-09
feat(query): compile-layer-first routing for mine/draft (FEAT-004)

Geoffe-Ga:
## Summary

Implements [FEAT-004](../blob/main/plans/git-issues/FEAT-004-compiled-layer-query-routing.md): `creek mine` and `creek draft` now route through the compiled layer (`02-Threads/`, `03-Eddies/`, `06-Frequencies/`) before falling back to fragments — operationalising the compile-then-query discipline that ADOPT-001 made structural in FEAT-003.

- New `creek/generate/compile_routing.py` module loads the compiled-page index keyed by `(target_kind, target_id)` and exposes `log_compile_gap()` which appends to `00-Creek-Meta/Processing-Log/compile-gaps.jsonl`.
- `mine_thread_terminus` derives `source_fragments` from the compiled thread's provenance; `mine_liminal_cross_eddy` compares against the compiled eddy body; `mine_wavelength_windows` gates fragments through compiled frequency-index provenance.
- `draft.gather_source_material` reads compiled thread/eddy bodies for the prompt; falls back to frontmatter descriptions only when the compiled page is missing.
- `--bypass-compiled` flag added to `creek mine` and `creek draft`. It short-circuits the index, emits a stderr warning, and is documented as a diagnostic escape hatch.
- New `00-Creek-Meta/Skills/query.SKILL.md` (~128 lines, ≤1500 tokens). `AGENTS.md`'s inline 3-rule query contract is replaced with a pointer to it — closing the FEAT-001 asymmetry where compile/lint/save had dedicated skills but query was inline.
- `creek-tools/docs/generation.md` documents the routing change and the `--bypass-compiled` flag.

## Acceptance criteria verified

- ✅ `creek mine` and `creek draft` both check compiled pages before fragments (see `mine_thread_terminus`, `mine_liminal_cross_eddy`, `mine_wavelength_windows`, `gather_source_material`).
- ✅ `--bypass-compiled` flag exists on both commands and is documented (`creek/cli.py`, `docs/generation.md`).
- ✅ Missing compiled pages log to `00-Creek-Meta/Processing-Log/compile-gaps.jsonl` (`compile_routing.log_compile_gap`).
- ✅ Regression test verifies `creek draft` routes through the compiled layer in the absence of `--bypass-compiled` (`TestGatherSourceMaterialCompileFirst::test_compiled_thread_body_is_used_when_present` and `…test_bypass_compiled_pulls_descriptions_and_skips_log`).
[...]

PR #292, opened 2026-05-23
chore(compile): FEAT-004 follow-ups — shared factory, perf, lazy load, tests

### LINKED ISSUE #215: FEAT-004 follow-ups: test helper consolidation, perf nits, doc trust assumption, branch tests
## Summary

Six minor follow-ups surfaced in the LGTM re-reviews on PR #211 (FEAT-004). The reviewer marked each as non-blocking and "worth a follow-up issue." Filing them here so they don't get lost — none block any current PR; pick them up with the next compile-layer work or as a one-off `chore(compile)` PR.

Source reviews:
- https://github.com/Geoffe-Ga/Creek-Vault/pull/211#issuecomment-4412865388 (full re-review)
- https://github.com/Geoffe-Ga/Creek-Vault/pull/211#issuecomment-4411446317 (initial review)

### 1. 🟡 Consolidate duplicated test helpers across compile-routing test files

**Files:** `creek-tools/tests/test_mining.py` (`_write_compiled_thread_page`, `_write_compiled_eddy_page`), `creek-tools/tests/test_drafts.py` (same names, slightly different defaults — `fragment_ids=()` vs required), `creek-tools/tests/test_compile_routing.py` (similar but standalone).

The three helpers diverge in subtle ways (default args, target-path subfolders) and are open to drift when the `CompiledPage` schema evolves.

**Fix:** extract a shared `tests/factories/compiled.py` (or `conftest.py` fixture) returning a `CompiledPage` plus an optional `write_to(vault, target_id, target_kind, ...)` method. Migrate all three test modules to consume the factory.

Acceptance: a single source of truth for compiled-page test fixtures; `grep _write_compiled_thread_page tests/` returns one definition.

### 2. 🟢 `fragment_ids_for` dedup — O(n²) → O(n) via `dict.fromkeys`

**File:** `creek-tools/creek/generate/compile_routing.py:79–86`

```python
seen: list[str] = []
for entry in page.provenance:
    for fid in entry.fragment_ids:
        if fid and fid not in seen:   # O(n) membership check per iteration
            seen.append(fid)
return tuple(seen)
```

For a personal vault this is fine, but `if fid not in seen` on a list is O(n) per insertion. `dict.fromkeys` preserves insertion order in O(n) total:

```python
return tuple(dict.fromkeys(
    fid for entry in page.provenance for fid in entry.fragment_ids if fid
))
```

Acceptance: replacement keeps insertion order (regression test exists: `TestLoadCompiledPages::test_fragment_ids_for_dedupes_provenance`); no behaviour change on existing tests.

### 3. 🟢 Skip unconditional `_load_threads_by_id` / `_load_eddies_by_id` on full-cache hit

**File:** `creek-tools/creek/generate/drafts.py` — `_compose_thread_section` / `_compose_eddy_section`.

Both helpers call `_load_threads_by_id` / `_load_eddies_by_id` even when every requested ID hits the compiled-page cache and the fallback path is never taken. For a fully-compiled vault this is wasted disk I/O.

**Fix:** lazy-load. Only scan the frontmatter directory when at least one ID misses the compiled cache. Pseudo-code:

```python
threads: dict[str, Thread] = {}
for tid in idea.threads:
    page = compiled.thread(tid)
    if page is not None:
        entries.append(_render_compiled_entry(tid, page))
        continue
    if not threads:  # lazy: only load on first miss
        threads = _load_threads_by_id(vault_path / _THREADS_SUBDIR)
    ...
```

Acceptance: existing draft tests still pass; a new test proving `_load_threads_by_id` isn't called when every thread hits the compiled cache (e.g., monkeypatched spy or `Path.rglob` patch).

### 4. 🟢 Targeted unit tests for `_render_compiled_entry`

**File:** `creek-tools/creek/generate/drafts.py:633–645` (the heading-strip heuristic)

`_render_compiled_entry` is currently exercised only indirectly through `gather_source_material`. The `lstrip("# ")` heuristic has three branches worth pinning:

```python
def test_render_compiled_entry_strips_matching_title_header():
    """`# {title}` first line is stripped when it matches page.title."""

def test_render_compiled_entry_preserves_non_title_first_line():
    """First line that does NOT match page.title is kept verbatim."""

def test_render_compiled_entry_empty_body_placeholder():
    """Empty body after stripping renders as the fallback string."""
```

Acceptance: three unit tests added to `tests/test_drafts.py`; coverage on `drafts.py:633–645` reaches 100%.

### 5. 🟢 Test for `yaml.YAMLError` branch in `_load_pages`

**File:** `creek-tools/creek/generate/compile_routing.py:128–132`

The `except (OSError, ValueError, yaml.YAMLError)` catch in `_load_pages` is uncovered (lines 130–132 in coverage report).

**Fix:**
```python
def test_skips_malformed_yaml(tmp_path):
    target = tmp_path / "02-Threads" / "bad.md"
    target.parent.mkdir(parents=True, exist_ok=True)
    target.write_text("---\nkey: [\n---\nbody", encoding="utf-8")
    result = _load_pages(target.parent, "thread")
    assert result == {}
```

Acceptance: test covers the YAMLError branch; per-file coverage on `compile_routing.py` rises closer to 100%.

### 6. 🟢 Remove noisy inline comment on `_warn_bypass_compiled`

**File:** `creek-tools/creek/cli.py:174` (the `# noqa: T201  # operator-facing CLI warning` comment).

The function name and docstring already say "operator-facing CLI warning"; the trailing `#` comment after `noqa: T201` adds noise without context. Keep `# noqa: T201` (the lint exemption is required) but drop the trailing prose.

**Fix:** change `print(  # noqa: T201  # operator-facing CLI warning` → `print(  # noqa: T201`.

Acceptance: lint still passes; comment density drops by one line.

### 7. 🟢 Document the trust assumption on `compile_routing.py`'s scanned directories

**File:** `creek-tools/creek/generate/compile_routing.py` module docstring.

`_load_pages` uses `rglob("*.md")` which follows symlinks. The `type: compiled_page` sentinel + Pydantic validation make it safe in practice (planted non-pages are skipped), but the trust assumption isn't called out.

**Fix:** add a one-liner to the module docstring stating that `02-Threads/`, `03-Eddies/`, and `06-Frequencies/` are assumed to contain only trusted, well-formed files written by `creek compile` (FEAT-003).

Acceptance: the trust assumption is documented; future maintainers reading the module docstring know not to point this code at untrusted directories.

## Out of scope

None of these block a current PR. All seven can land in a single `chore(compile)` PR or be folded into the next compile-layer feature (e.g., FEAT-008 lint will likely touch `compile-gaps.jsonl` consumption and is a natural home for items 1, 4, and 5).

## Acceptance criteria (overall)

- [ ] Item 1: shared compiled-page test factory; one definition per helper.
- [ ] Item 2: `fragment_ids_for` uses `dict.fromkeys`; existing tests pass.
- [ ] Item 3: full-cache hit skips frontmatter scan; new spy-test pins the optimisation.
- [ ] Item 4: three `_render_compiled_entry` tests added; line coverage at 100%.
- [ ] Item 5: malformed-YAML test added; covers the `yaml.YAMLError` branch.
- [ ] Item 6: trailing comment dropped; lint passes.
- [ ] Item 7: trust assumption documented in module docstring.

## References

- PR #211 (FEAT-004): https://github.com/Geoffe-Ga/Creek-Vault/pull/211
- FEAT-004 plan: `plans/git-issues/done/FEAT-004-compiled-layer-query-routing.md` (after backlog grooming).

Geoffe-Ga:
Closes #215.

## Summary

Seven minor follow-ups surfaced in the PR #211 LGTM re-review:

1. **Test-helper consolidation.** Three test modules each defined their own `_write_compiled_*_page` helpers with subtly different signatures and target paths. They're now imported from a new shared `tests/factories/compiled.py` (`build_compiled_page`, `write_compiled_page`, `write_compiled_thread_page`, `write_compiled_eddy_page`, `write_compiled_frequency_page`). `grep _write_compiled_thread_page tests/` now returns one definition.
2. **`fragment_ids_for` perf.** Replaces the `if fid not in seen: seen.append(fid)` O(n²) dedup with `dict.fromkeys(...)` (O(n) total, preserves insertion order). Pinned by the existing `test_fragment_ids_for_dedupes_provenance` regression test.
3. **Lazy frontmatter load.** `_compose_thread_section` and `_compose_eddy_section` no longer call `_load_threads_by_id` / `_load_eddies_by_id` when every requested ID hits the compiled cache. New monkeypatched-spy tests pin the optimisation for both kinds, plus a "loads-on-first-miss" variant that proves the scan happens exactly once and is reused.
4. **`_render_compiled_entry` unit tests.** Three new tests pin the heading-strip heuristic (matching title stripped, non-title preserved, empty-body placeholder). Function coverage is now 100% lines / 100% branches.
5. **`yaml.YAMLError` branch.** New `test_skips_malformed_yaml` covers the previously-uncovered `yaml.YAMLError` arm in `_load_pages` via a deliberately-broken YAML frontmatter file.
6. **Drop noisy comment.** `print(  # operator-facing CLI warning` → `print(`. The function name and docstring already say it; the inline comment was just noise.
7. **Document trust assumption.** `compile_routing.py` module docstring now states that `02-Threads/`, `03-Eddies/`, and `06-Frequencies/` are trusted output written by `creek compile` — `_load_pages` uses `rglob` which follows symlinks, and that's safe only because the routing layer is never pointed at untrusted content.

## Quality gates

All 12 `./scripts/check-all.sh` gates green:
- 3793 tests passing, 93.72% branch coverage
- Ruff lint + format clean, MyPy strict clean
- Per-file coverage gate green (134 ≥ 80%, 2 waivered ≥ 65%)

## Test plan

- [x] `./scripts/check-all.sh` exit 0 locally
- [ ] CI green
- [ ] LGTM

---
_Generated by [Claude Code](https://claude.ai/code/session_01JFvvez73NgeKgGyLyZM856)_

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -1478,7 +1478,7 @@ def _warn_bypass_compiled(verb: str) -> None:
     routing introduced in FEAT-004. Warning loudly is the contract:
     bypass should be a deliberate, visible choice, not silent.
     """
-    print(  # operator-facing CLI warning
+    print(
         f"[creek {verb}] WARNING: --bypass-compiled is set; "
         "the compiled-layer-first contract is being side-stepped.",
         file=sys.stderr,
--- a/creek-tools/creek/generate/compile_routing.py
+++ b/creek-tools/creek/generate/compile_routing.py
@@ -10,6 +10,14 @@ compile-gaps.jsonl`` so ``creek lint`` can surface the gap later.
 This module is the single read-side surface for the compiled layer.
 Mining and drafting both consume it; the CLI's ``--bypass-compiled``
 escape hatch short-circuits the index load entirely.
+
+Trust assumption: the scanned directories ``02-Threads/``,
+``03-Eddies/``, and ``06-Frequencies/`` are treated as trusted output
+written by ``creek compile`` (FEAT-003). ``_load_pages`` uses
+``rglob("*.md")``, which follows symlinks; the ``type: compiled_page``
+sentinel plus Pydantic validation make planted non-pages a no-op in
+practice, but the routing layer should not be pointed at untrusted
+content.
 """
 
 from __future__ import annotations
@@ -77,13 +85,18 @@ class CompiledPageIndex:
         return self.frequency_indexes.get(target_id)
 
     def fragment_ids_for(self, page: CompiledPage) -> tuple[str, ...]:
-        """Return the unique fragment IDs traced by *page*'s provenance."""
-        seen: list[str] = []
-        for entry in page.provenance:
-            for fid in entry.fragment_ids:
-                if fid and fid not in seen:
-                    seen.append(fid)
-        return tuple(seen)
+        """Return the unique fragment IDs traced by *page*'s provenance.
+
+        ``dict.fromkeys`` preserves insertion order in O(n) total — the
+        previous list-based dedup paid an O(n) membership check per
+        insertion. The :class:`TestLoadCompiledPages` suite pins the
+        ordering invariant.
+        """
+        return tuple(
+            dict.fromkeys(
+                fid for entry in page.provenance for fid in entry.fragment_ids if fid
+            ),
+        )
 
 
 def empty_index(*, bypassed: bool = False) -> CompiledPageIndex:
--- a/creek-tools/creek/generate/compile_routing.py
+++ b/creek-tools/creek/generate/compile_routing.py
@@ -87,10 +87,8 @@ class CompiledPageIndex:
     def fragment_ids_for(self, page: CompiledPage) -> tuple[str, ...]:
         """Return the unique fragment IDs traced by *page*'s provenance.
 
-        ``dict.fromkeys`` preserves insertion order in O(n) total — the
-        previous list-based dedup paid an O(n) membership check per
-        insertion. The :class:`TestLoadCompiledPages` suite pins the
-        ordering invariant.
+        Uses ``dict.fromkeys`` for O(n) deduplication; insertion order
+        is guaranteed by the Python 3.7+ language spec.
         """
         return tuple(
             dict.fromkeys(
--- a/creek-tools/creek/generate/drafts.py
+++ b/creek-tools/creek/generate/drafts.py
@@ -462,10 +462,15 @@ class DraftGenerator:
         vault_path: Path,
         compiled: CompiledPageIndex,
     ) -> str:
-        """Render the ``## Threads`` block, compile-first."""
+        """Render the ``## Threads`` block, compile-first.
+
+        The frontmatter scan ``_load_threads_by_id`` is deferred until a
+        cache miss actually needs it, so fully-compiled vaults skip the
+        disk walk entirely.
+        """
         if not idea.threads:
             return ""
-        threads = _load_threads_by_id(vault_path / _THREADS_SUBDIR)
+        threads: dict[str, Thread] | None = None
         entries: list[str] = []
         for tid in idea.threads:
             page = compiled.thread(tid)
@@ -480,6 +485,8 @@ class DraftGenerator:
                     surfaced_by="draft",
                     reason="missing",
                 )
+            if threads is None:
+                threads = _load_threads_by_id(vault_path / _THREADS_SUBDIR)
             thread = threads.get(tid)
             if thread is not None:
                 desc = thread.description.strip() or "(no description)"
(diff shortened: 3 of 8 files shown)


Geoffe-Ga:
Addressed in fe7904a.

| # | Item | Status |
|---|------|--------|
| 1 | `# noqa: TC003` in `tests/factories/compiled.py:22` | ✅ Fixed — `Path` moved into the `TYPE_CHECKING` block. Annotations are lazy under `from __future__ import annotations`, so the runtime import was indeed unnecessary. |
| 2 | Merge split `from tests.factories.compiled import …` blocks in `test_drafts.py` / `test_mining.py` | 🔁 No-op — ruff's isort restores the split form on every auto-format because each `from X import Y as Z` aliased entry lands in its own import block under the project's default configuration. I merged them manually; the formatter reverted them. Happy to revisit if there's a config tweak you'd prefer. |
| 3 | `fragment_ids_for` docstring references `TestLoadCompiledPages` | ✅ Fixed — rephrased to name the language-spec guarantee ("insertion order is guaranteed by the Python 3.7+ language spec") instead of the test class. |
| 4 | Minor fixture data drift (`excerpt-{i}` → `excerpt {i}`; default `body` for frequency pages) | ↪️ Acknowledged but no change — the strings are opaque test fixtures (no assertion depends on them) and the factory is now the single source of truth. Documenting that explicitly in the factory docstring would help, but I'd rather not bloat the module with "these are not load-bearing" disclaimers; the docstring already says "deterministic provenance" which I think carries the same intent. |

All 12 `./scripts/check-all.sh` gates remain green; 3793 tests pass.

@claude please re-review.

---

Issue #285, opened 2026-05-23
FEAT-035: MIME-type / magic-byte verification for Discord attachments


Geoffe-Ga:
> _**Renumbered 2026-05-24 (backlog grooming):** was FEAT-030, now FEAT-035. The FEAT-030 slot was already taken by #262 (Substack-aware ingestion, closed) — retconned during the 2026-05-23 grooming pass. This issue was filed against an apparently-free slot before that bump propagated. The issue number (#285) is unchanged. Body references to FEAT-027 (the CrawDad attachments work in #274/PR #280) have been updated to FEAT-033 to match the canonical numbering._

## Context

[FEAT-033 (PR #280)](https://github.com/Geoffe-Ga/Creek-Vault/pull/280) ships per-attachment **extension-only** gating in `crawdad/crawdad/attachments.py` (`_extension_allowed`). A reviewer flagged this as a v1 limitation: a renamed executable (`evil.exe` renamed to `evil.md`) passes the allow list and reaches the staging directory.

This is acceptable for v1 because:
- The bot **never** auto-ingests — every attachment requires explicit user consent (FEAT-033 § Consent gate).
- The downstream `creek redact --scan` runs before any ingest.
- Bot deployments are personal-tool scoped with an allowlisted user/channel pair (FEAT-013 §personal-use scoping).

But "the user explicitly typed `ingest`" is not the same as "the user knew what was actually in the file". This issue tracks adding a second, content-based check.

## Requirements

- **Magic-byte verification**: after download, sniff the first N bytes of the file and verify the detected MIME type matches what the extension claims. Mismatches do not auto-reject — they surface in the safety report alongside the redaction findings.
- **Optional reject mode**: a `crawdad.yaml` knob (e.g. `attachments.reject_on_mime_mismatch: bool = False`) flips the soft warning into a hard rejection at the same gate `_process_one` uses today.
- **Library choice**: prefer `python-magic` (bundles libmagic) or the lighter `filetype` package; document the trade-off in an ADR. Pure-stdlib `mimetypes.guess_type` is extension-only and does not help here.
- **Allow list narrowing**: with content-type verification in place, narrow the default `allowed_extensions` to only those whose magic-byte signatures we can verify (text files don't have signatures — treat them as a special case where content sampling is acceptable).
- **Audit trail**: log every mismatch to the existing `MCPAuditLog` via the scan tool, so a polyglot file (zip-disguised-as-pdf, etc.) leaves a clear paper trail.

[...]

PR #306, opened 2026-05-24
feat(crawdad): MIME-type verification for Discord attachments (FEAT-035, #285)

Geoffe-Ga:
Add magic-byte / content-sample content-type verification on top of the
extension allow list shipped in FEAT-033. After download, every
attachment is sniffed with ``filetype`` (binary signatures) or with a
UTF-8 + NUL-byte content sample (text formats), and the detected MIME
is compared against the extension's claim. Mismatches surface as a
soft-warning ⚠ line in the Discord safety report; the new
``attachments.reject_on_mime_mismatch`` knob in ``crawdad.yaml`` flips
the warning into a hard rejection at the same gate ``_process_one``
uses today.

Narrows the default ``allowed_extensions`` to the set whose content
type can be verified: legacy Office formats (``.doc``/``.xls``/``.ppt``)
drop out because ``filetype`` cannot reliably detect OLE compound
documents and the formats carry macro-execution risk. Operators who
need them can re-add the extensions in ``crawdad.yaml``.

Adds ADR-0002 documenting the library choice (``filetype`` over
``python-magic``: no libmagic system dep, smaller surface, fits the
single-user / pip-installable deployment target).

Tests cover the AC scenarios: matched binary, matched text, polyglot
(zip-disguised-as-PDF), executable-disguised-as-markdown, unknown
extension, the configurable reject mode, and the summary surfacing.

### DIFF
--- a/crawdad/crawdad/attachments.py
+++ b/crawdad/crawdad/attachments.py
@@ -81,56 +81,69 @@ _EXTENSION_TO_INGESTOR_TYPE: dict[str, str] = {
     ".webp": "image",
 }
 
-# FEAT-035: extensions whose declared content type can be verified by
-# ``filetype.guess`` against the first kilobyte of the downloaded body.
-# Each entry lists every MIME string the library may legitimately return
-# for that extension — OOXML containers, for example, are zip files at
-# the byte level so ``filetype`` correctly reports ``application/zip``
-# unless the runtime libmagic version has dedicated OOXML detection.
-_EXTENSION_TO_ACCEPTABLE_MIMES: dict[str, frozenset[str]] = {
-    ".pdf": frozenset({"application/pdf"}),
-    ".png": frozenset({"image/png"}),
-    ".jpg": frozenset({"image/jpeg"}),
-    ".jpeg": frozenset({"image/jpeg"}),
-    ".gif": frozenset({"image/gif"}),
-    ".webp": frozenset({"image/webp"}),
-    ".docx": frozenset(
-        {
-            "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-            "application/zip",
-        }
-    ),
-    ".xlsx": frozenset(
-        {
-            "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
-            "application/zip",
-        }
-    ),
-    ".pptx": frozenset(
-        {
-            "application/vnd.openxmlformats-officedocument.presentationml.presentation",
-            "application/zip",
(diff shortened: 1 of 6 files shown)


Geoffe-Ga:
Addressed in `222f6c1`.

**Blockers**

| # | Fix | Where |
|---|---|---|
| 1 | Dual-dict → single `_EXTENSION_MIME_SPECS` keyed by frozen `_ExtensionMimeSpec` (canonical + acceptable). Drift impossible by construction. Structural test pins `canonical ∈ acceptable` for every spec. | `crawdad/attachments.py`, `tests/test_attachments.py::test_extension_mime_specs_canonical_is_member_of_acceptable` |
| 2 | `test_process_attachment_unknown_extension_still_accepts` fixture now uses `frozenset({".toml"})` — pins only the MIME gate, leaves the empty-allow-list semantics covered by `test_empty_allowed_extensions_passes_anything_not_denied`. | `tests/test_attachments.py` |

**Comments addressed**

- Text-extension matches now report per-extension MIMEs (`text/markdown`, `application/json`, `text/html`, `text/csv`) instead of a blanket `text/plain`. New test `test_verify_mime_type_text_match_uses_per_extension_mime` covers it.
- Added YAML round-trip tests for `attachments.reject_on_mime_mismatch` — both the explicit `true` override and the absent-default case (`tests/test_config.py`).

**Pushing back (with rationale)**

- *"Add a heuristic-verification note to the summary for text-extension files"* — the heuristic limitation is documented in ADR-0002 §Negative and in the `_TEXT_EXTENSION_TO_MIME` block comment. Adding a per-file disclosure to every text attachment in the Discord reply would add summary noise on every reply (most attachments are text) without changing the security posture, since the FEAT-027 user-consent gate remains the authoritative refusal point. Happy to revisit if a stricter deployment context arises.
- *"`filetype>=` floor pin"* — CrawDad doesn't ship a separate lockfile today; the floor matches the pattern the rest of the deps use. Tightening the constraint when CrawDad picks up a lockfile is a clean follow-up.
- *"`verify_mime_type` implicit public surface"* — left as-is. It's exported intentionally for callers who want to verify outside the download path, and the module is small enough that an `__all__` would be redundant.
- *"ADR location under `creek-tools/`"* — left as-is. The repo only has one ADR tree today; if `crawdad/` ever grows its own, the file can move then.

Local: `./scripts/check-all.sh` → 7/7 green. 232 tests pass, 94.55% branch coverage.

@claude please re-review.

---

PR #330, opened 2026-05-24
fix(classify): exit non-zero when --method llm provider is unavailable

### LINKED ISSUE #317: bug(classify): `--method llm` reports "Classified N of N" and exits 0 when LLM provider is unavailable
## Symptom

Running `creek classify --method llm` with the Anthropic provider configured but **without** `CREEK_ANTHROPIC_CONSENT=1` set produces this output:

```
WARNING: Cloud classification enabled. Fragment content will be sent to Anthropic's servers.
Anthropic provider not available: Anthropic cloud classification requires explicit consent. Set CREEK_ANTHROPIC_CONSENT=1 to confirm that fragment content may be sent to Anthropic's servers.
LLM provider unavailable — returning fragment 'X' unchanged
... (repeated ~173 times) ...
Classified 188 of 188 fragment(s) (0 manual preserved, 17 skipped).
```

Exit code is **0**. Nothing was actually classified, but the summary claims success and the shell sees a clean exit.

## Expected

- Exit code should be **non-zero** when `--method llm` was requested but the provider was unavailable and produced zero classifications.
- The summary should not say "Classified 188 of 188" when in fact 0 fragments received an LLM classification.

## Reproduction

1. `creek init --vault /tmp/v`
2. Edit `/tmp/v/00-Creek-Meta/creek_config.yaml`: set `llm.provider: anthropic`
3. Have `ANTHROPIC_API_KEY` exported but **not** `CREEK_ANTHROPIC_CONSENT`
4. Ingest any source
5. `creek classify --vault /tmp/v --method llm`

## Root cause hint

The summary counter appears to increment for every fragment the orchestrator iterates over, even when the LLM call returned `unchanged` due to provider unavailability. The summary should distinguish "successfully classified" from "passed through unchanged."

## Discovered during

A `creek-walkthrough` session with a real first-time user ingesting 188 Substack essays on 2026-05-24.

Geoffe-Ga:
## Summary

- Adds `LLMProviderUnavailableError` raised by `run_classify` before the per-fragment loop when the configured LLM provider fails its availability check, so the CLI can exit non-zero with a clear remediation hint instead of pretending "Classified 188 of 188" succeeded.
- CLI prints `Classification aborted: LLM provider 'anthropic' is unavailable: set ANTHROPIC_API_KEY and CREEK_ANTHROPIC_CONSENT=1 ...` and exits with code 1.
- MCP `classify_tool` wrapper catches the same exception and returns the standard `status: refused` payload so MCP clients keep their stable response shape.
- Fragments are left completely untouched on this failure path, preserving the OPS-001 resume contract (no bogus `classification_method: llm` stamps when the LLM never ran).

## Test plan

- [x] New unit tests in `tests/test_classify_engine.py` covering the raise path, provider name in the error, no progress-log artefacts, and that `--method rules` never constructs the LLM.
- [x] New CLI tests in `tests/test_cli.py` asserting non-zero exit code, that "Classified" no longer appears, and that fragments stay unstamped on disk.
- [x] New MCP test in `tests/test_mcp_write_tools.py` confirming the refusal payload shape.
- [x] Existing OPS-001 resume tests updated to bypass the new gate via an autouse fixture (they mock `classify_with_reasoning` so the gate would otherwise short-circuit them).
- [x] `./scripts/check-all.sh` exits 0 (12/12 checks pass, coverage 93.69%).
- [x] Manual e2e per the issue recipe: `creek classify --vault /tmp/v --method llm` with `CREEK_ANTHROPIC_CONSENT` unset now prints the abort message and exits with code 1; the seeded fragment on disk is untouched.

Closes #317

https://claude.ai/code/session_011e4q4ZQcwtP9zVEp21T6ki

---
_Generated by [Claude Code](https://claude.ai/code/session_011e4q4ZQcwtP9zVEp21T6ki)_

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -78,6 +78,41 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+class LLMProviderUnavailableError(RuntimeError):
+    """Raised when ``creek classify --method llm`` cannot reach its provider.
+
+    Surfaced by :func:`run_classify` *before* any fragment is rewritten
+    so that:
+
+    1. The CLI can exit non-zero instead of pretending the run
+       succeeded with zero work done.
+    2. The vault stays clean — no fragments end up stamped with
+       ``classification_method: llm`` when the LLM never actually ran.
+
+    The configured provider name is preserved on the exception so the
+    CLI message can name it ("anthropic" / "ollama") and point the
+    operator at the right environment variable to fix.
+
+    Attributes:
+        provider: The configured ``llm.provider`` (e.g. ``"anthropic"``).
+    """
+
+    def __init__(self, provider: str, detail: str) -> None:
+        """Build the exception with a CLI-ready message.
+
+        Args:
+            provider: The configured LLM provider name.
+            detail: Provider-specific reason (env var missing, health
+                check failed, etc.) — embedded in the message so the
+                operator does not have to scroll back through warnings
+                to see what went wrong.
+        """
+        self.provider = provider
+        super().__init__(
+            f"LLM provider {provider!r} is unavailable: {detail}",
+        )
+
+
 @dataclass(frozen=True)
 class ClassifySummary:
     """Counts produced by a single ``creek classify`` run.
@@ -158,6 +193,13 @@ def run_classify(
 
     Returns:
         A :class:`ClassifySummary` reporting per-method counts.
+
+    Raises:
+        LLMProviderUnavailableError: When ``method == "llm"`` and the
+            configured provider fails its availability check. Raised
+            *before* any fragment is processed so the vault is not
+            polluted with bogus ``classification_method: llm`` stamps
+            when the LLM never ran.
     """
     fragments_root = vault_path / "01-Fragments"
     if not fragments_root.exists():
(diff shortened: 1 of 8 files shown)


Geoffe-Ga:
Thanks for the review. Addressed in d49f06b:

**BLOCKING items**
- Removed every `# Issue #317` reference from production code and tests (`classify_engine.py`, `cli.py`, `creek_mcp/tools/classify.py`, plus the three test files); rewrote the comments to state the invariant (refuse-to-iterate, keep the vault clean) without the issue number.
- Replaced `import pytest as _pytest` + `_pytest.raises(...)` inside test function bodies with the existing module-level `pytest` import and `pytest.raises(...)`. Pulled the unused `TYPE_CHECKING` block in `test_classify_few_shot.py` since `pytest` is now imported at runtime.

**HIGH items**
- Swapped every `new_callable=lambda: property(lambda _self: ...)` patch (10 sites across `test_classify_engine.py`, `test_classify_engine_resume.py`, `test_cli.py`, `test_mcp_write_tools.py`) for the idiomatic `patch.object(LLMClassifier, "available", new_callable=PropertyMock, return_value=...)`.
- Added three new unit tests for `_describe_llm_unavailability` covering: Anthropic (asserts both `ANTHROPIC_API_KEY` and `CREEK_ANTHROPIC_CONSENT` appear), Ollama (asserts `Ollama` daemon + `llm.url` config key appear), and the generic fallback (asserts `creek_config.yaml` pointer appears and that provider-specific env-var names do NOT leak).

**LOW items deferred**
- `_seed_single_fragment` vs `_write_fragment` duplication: added a docstring on `_seed_single_fragment` documenting that the engine-test helper exists and explaining why the CLI test keeps a local seeder (deliberately-minimal body for output assertions). Out of scope to extract to `conftest.py` in this PR.
- Empty-vault ordering: the existing `test_run_classify_returns_zeroes_when_no_fragments_dir` test pins the early-return behaviour; leaving as-is per the reviewer's "may be intentional" note.

`./scripts/check-all.sh` exits 0 (12/12 checks pass, 4063 tests passed, coverage 93.70%).

@claude please re-review

---

PR #337, opened 2026-05-25
fix(purge): actually remove embedding rows on purge, audit honestly (GAP-001)

Geoffe-Ga:
The purge engine's audit log advertised `embeddings_removed` but the
field was a fabrication — it mirrored `fragments_deleted` while the
embeddings parquet cache was never touched. Right-to-be-forgotten on
intimate-tier content silently left the partially-invertible vectors
on disk, contradicting both the README's RTBF guarantee and the audit
log's role as the compliance record.

- Add `purge_fragment_ids_from_cache` and `delete_embeddings_cache`
  helpers in `creek.link.embeddings` to scrub the parquet by ID and
  to remove the file outright for vault-level purge.
- Wire those helpers into every file-deleting purge operation
  (`fragment`, `source`, `source-path`, `daterange`, `vault`),
  respecting dry-run.
- Promote `embeddings_removed` to a real `PurgeResult` field so the
  audit entry reports the actual row delta — zero when the cache
  had not been built yet, the exact count otherwise.
- Update the threat model and `cleaning-and-purge` docs so the
  contract matches the implementation.

Co-authored-by: geoff@creekmasons.com

### DIFF
--- a/creek-tools/creek/link/embeddings.py
+++ b/creek-tools/creek/link/embeddings.py
@@ -28,7 +28,7 @@ from creek.models import (
 )
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping
+    from collections.abc import Iterable, Mapping
     from pathlib import Path
 
     from sentence_transformers import SentenceTransformer as SentenceTransformerType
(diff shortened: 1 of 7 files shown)


Geoffe-Ga:
Addressed the HIGH blocker from the previous review:

- **HIGH: missing `purge_source_path` GAP-001 integration test** — added `test_source_path_purge_removes_embedding_rows` in `tests/test_purge.py` (commit `1f06f79`). The test seeds a parquet cache with three fragments, runs `purge_source_path("/exports/", match="substring")`, and asserts both `result.embeddings_removed == 2` and that the audit entry records the same on the `source-path` operation.

Non-blocking items, briefly:

- **LOW: non-atomic parquet rewrite** — acknowledged. Single-writer CLI tool, acceptable for now; can be hardened with a temp-file-rename in a follow-up if a multi-process scenario emerges.
- **LOW: full table loaded into memory** — same posture; expected vault sizes keep this comfortable.
- **Code Quality: `pc.is_in` alternative** — deliberately avoided `pyarrow.compute` because its stubs are incomplete and the symbols (`is_in`, `sum`, `invert`, `cast`) trip `mypy --strict`. The pure-Python mask is correct and well within budget for the expected row counts.
- **Code Quality: collapse `keep_mask`/`keep_array`** — kept as two named bindings to make the bool-list → `pa.array` step explicit; reads more clearly than a single nested expression.
- **Code Quality: `GAP-001` tags in docstrings** — happy to strip these in a follow-up grooming pass once the gap is closed in main; they're useful trail for now.

`./scripts/check-all.sh` still exits 0 (all 12 gates green).

@claude please re-review.

---

PR #341, opened 2026-05-25
fix(purge): emit pre-destruction intent + post-destruction outcome audit pair (GAP-002, GAP-010)

Geoffe-Ga:
The purge engine used to write its audit line *after* the wipe loop
completed. A SIGKILL, OSError, OOM kill, or Ctrl-C mid-loop left the
filesystem in a partial state with zero audit trail: the operator had
no record of what was being attempted, let alone how far it got.

This change splits every purge operation into a pair of audit entries
sharing a UUID4 `operation_id`:

- An **intent** entry is written *before* the first destructive op,
  capturing the planned scope (operation, criteria, dry-run flag).
  Pure pre-destruction forensic trail — even if the process dies
  between the intent write and the first unlink, the trail exists.
- An **outcome** entry is written *after* the body runs. On success
  it carries `status="complete"` with the real counts; on any
  exception it carries `status="partial"` with the partial counts and
  a `failure_reason` field naming the exception type and message,
  then the original exception propagates.

The pattern wraps every purge entry point — `purge_fragment`,
`purge_source`, `purge_source_path`, `purge_daterange`,
`purge_classifications`, and `purge_vault` — via the shared
`_run_audited(result, body)` helper, closing the GAP-002/GAP-010
"after-destruction audit" pattern across the engine in one PR.

The engine is **not** transactional: there is no staging-directory
rename pattern, so a crash partway through `_wipe_folder_contents`
still leaves the filesystem half-deleted. The intent + outcome pair
is the recovery contract, documented explicitly in the module
docstring and in `docs/cleaning-and-purge.md`.

Backward compatibility: pre-GAP-002 audit entries (no `phase` field)
read back as `phase="outcome"` with `operation_id=""` and
`status=None`. The legacy `_write_audit(result)` shim is preserved
for the small number of direct callers in the test suite.

### DIFF
--- a/creek-tools/creek/purge/audit.py
+++ b/creek-tools/creek/purge/audit.py
@@ -6,6 +6,15 @@ entry captures the timestamp, operation, criteria dict, affected
 fragment IDs, deletion counts, references scrubbed, embeddings removed,
 operator, and dry-run flag.
 
+Phase pairing (GAP-002): every purge operation emits **two** entries —
+an ``intent`` line written *before* the first destructive op, then an
+``outcome`` line written *after* the destructive section completes
+(``status="complete"``) or after an exception aborts it
+(``status="partial"``). Both entries share a UUID4 ``operation_id`` so
+a recovery tool can pair them. Pre-GAP-002 entries have no ``phase``
+field; they read back as ``phase="outcome"`` with an empty
+``operation_id`` and ``status=None``.
+
 Timezone (BUG-002): purge-audit entries deliberately stamp ``UTC``
 rather than ``America/Los_Angeles`` (the rest of the pipeline). The
 audit chain is forensic infrastructure — its consumers are operators
@@ -30,7 +39,7 @@ import json
 import logging
 from datetime import UTC, datetime
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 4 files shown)


Geoffe-Ga:
Addressed the HIGH blocker:

- **HIGH: `type: ignore[arg-type]` on `status=status`** — fixed in commit `9b9c4f4`. Imported `PurgeOutcomeStatus` (the `Literal["complete", "partial"]` alias from `creek.purge.audit`) and changed the `_write_outcome_audit` signature from `status: str` to `status: PurgeOutcomeStatus`. Both call sites already pass the string literals, so the change is one line plus the import. The `# type: ignore` is gone; `./scripts/check-all.sh` exits 0 (all 12 gates).

Non-blocking Code Quality nits, briefly:

- **`_write_audit` shim removal** — happy to file a follow-up issue if you want to track it separately. Today's only direct caller is `test_write_audit_known_operations_use_explicit_allowlist`, which deliberately exercises the legacy validate-and-write path; once that test migrates we can delete the shim.
- **`_write_outcome_audit` lacks `_validate_operation`** — it actually does: see `deletes_files = self._validate_operation(result.operation)` on the first non-docstring line of the function (`engine.py`). Looks like the reviewer scanned for an early-return validate call and missed the inline form; nothing to change.
- **`_entries_for_run` masking migration entries** — fair point; defending against accidental cross-talk would be useful. Logging as a non-blocking test-hygiene follow-up rather than touching every GAP-002 test in this PR.

@claude please re-review.

---

PR #346, opened 2026-05-25
fix(mine): surface per-strategy diagnostics and expose mining knobs (#340)

### LINKED ISSUE #340: bug(mine): `creek mine` returns zero seeds and no diagnostic on a 188-fragment classified vault
## Symptom

`creek mine` returns `No idea seeds surfaced.` on a vault with:

- 188 classified fragments (frequency populated on 185, wavelength populated on the same)
- Phase distribution across the corpus:

  ```
  peaking:        63
  restoration:    57
  rising:         45
  withdrawal:     11
  bottoming_out:   7
  unclassified:    5
  ```
- 1 compiled eddy page
- A fully-populated `embeddings.parquet` (1,212 pairwise similarity edges)

Both the default invocation and `--bypass-compiled --phase peaking` return zero seeds. No `compile-gaps.jsonl` file is written either, so the miner didn't even attempt the compiled-layer-first fallback path.

The skill description acknowledges *"Nothing surfaces → a young or small vault simply hasn't pooled enough yet."* That calibration is reasonable for tens of fragments. With 188 classified essays and 63 candidates in the current phase, the miner returning zero is more likely a real defect than honest emptiness.

## Expected

On a corpus of this size with this much classification + linking infrastructure, the wavelength-phase-window strategy alone should be able to surface at least 1–3 seeds (since 63 fragments match the current phase). The resonance-chain strategy should also find something across 1,212 similarity edges.

Even if no seeds *score* high enough to display, the miner should:
- write a `compile-gaps.jsonl` entry for each strategy that fell back due to missing compiled pages
- print a one-line diagnostic indicating which strategies ran and what scores they considered
- not silently report "No idea seeds surfaced" without surfacing *why*

## Reproduction

1. Ingest ~100+ classified fragments (e.g., 188 Substack essays)
2. Run `creek link` for embeddings, temporal, eddies
3. Compile at least one eddy
4. `creek mine --vault <vault> --phase peaking --limit 10`
5. Output: `No idea seeds surfaced.`
6. Re-run with `--bypass-compiled`: same output
7. Confirm `00-Creek-Meta/compile-gaps.jsonl` does not exist

## Hypothesis

Three possible causes (any may be true):

1. **Strategies are all gated on infrastructure we don't have.** Thread terminus needs threads (we have 0). Liminal cross-eddy needs compost/unnamed in liminal (we have 0 in either physically — see #339). Resonance chain may need specific edge weights or chain lengths the corpus doesn't have. Wavelength-phase window may need extra signal beyond `phase` to score (e.g., recency, which is broken for us since `authored_at: null` on many fragments).

2. **The score thresholds are calibrated for a different corpus shape** — perhaps small-N personal vaults with daily journal entries — and quietly filter out everything from a long-form essay corpus.

3. **Bugs in any of the four strategies** — the miner runs all of them and de-duplicates; a silent error in one could explain a partial empty, but all four returning empty in unison points to either a shared dependency or calibration.

The compiled-layer-first contract also doesn't seem to engage — the `compile-gaps.jsonl` file the help text says should be appended to on fallback is never created.

## Workaround during walkthrough

`creek draft --seed-*` (FEAT-032 manual seeding flags) lets the user bypass mining entirely and draft from a hand-chosen topic. Walkthroughs can recover by switching to manual seeding, but the mining step itself produces nothing the user can act on.

## Discovered during

A `creek-walkthrough` session on 2026-05-25. The user had a fully-prepared vault and asked the system *what wants to be written?* — and got `No idea seeds surfaced.` with no diagnostic. The next pool (drafting) was reachable only via manual seeding.

Geoffe-Ga:
## Summary

Closes #340.

`creek mine` on a 188-fragment classified vault returned `No idea seeds surfaced.` with no breadcrumb explaining why — operators had no way to tell whether they were close, miles off, or hitting a fallback path that silently produced nothing. This PR addresses both parts called out in the issue: honest diagnostics for every run, and YAML-tunable thresholds for small-N corpora.

### Part A — Honest diagnostics (always)

- New `StrategyDiagnostic` dataclass records `candidates_considered`, `candidates_kept`, `top_score`, `threshold`, and an optional `fallback_reason` for each of the four mining strategies.
- New `IdeaMiner.mine_all_with_report() -> MiningRunReport` returns the deduped seed list together with one diagnostic per strategy. `mine_all` is now a thin shim around it; existing callers (`creek_mcp/tools/mine.py`, `phase_filtered_seeds`, tests) are unchanged.
- Each strategy emits an INFO-level log line:
  ```
  [mine] <strategy>: <n> considered, <m> kept (top_score=... threshold=...)[ reason=...]
  ```
  Operators can grep for `[mine]` to triage a zero-seed run from pipeline logs.
- Resonance chain now appends a `compile-needed` entry to `compile-gaps.jsonl` when no synchronicity records exist on disk — the exact silent-zero case that bit the user, since `embeddings.parquet` edges do not contribute until they are written as synchronicity notes under `10-Liminal/Synchronicities/`.
- CLI replaces the bare `No idea seeds surfaced.` with:
  > `No idea seeds surfaced. Ran 4 strategies; top candidate scored 0.21 (threshold 0.40). See <vault>/00-Creek-Meta/Processing-Log/compile-gaps.jsonl for fallback reasons.`

### Part B — Calibration knobs in `creek_config.yaml`

New `MiningConfig` section exposes the four previously hard-coded thresholds:

```yaml
mining:
  min_thread_fragments: 10
  min_chain_length: 3
  similarity_liminal: 0.35
  similarity_resonance: 0.6
```

`creek mine` reads the section via `_load_config_for_vault(...)` and forwards every knob to `IdeaMiner`, so operators can calibrate against a small-N corpus without monkey-patching defaults.

### Threshold defaults — unchanged

I did NOT change any default threshold values. The library constants (`DEFAULT_MIN_THREAD_FRAGMENTS=10`, `DEFAULT_MIN_CHAIN_LENGTH=3`, `DEFAULT_SIMILARITY_LIMINAL=0.35`, `DEFAULT_SIMILARITY_RESONANCE=0.6`) and the new `MiningConfig` defaults are byte-for-byte identical to the pre-PR behaviour. The issue's "do not change defaults without justification" guidance is honoured — what was clearly wrong wasn't the numbers, it was that they were invisible and unreachable. With the new diagnostics, an operator can now see precisely which threshold their corpus is failing to clear and reach for the knob, rather than guessing.

## Test plan

- [x] `./scripts/check-all.sh` — all 12 gates green
- [x] `./scripts/test.sh --all` — 4151 passed, 1 skipped (Ollama-gated)
- [x] Coverage 93.68% (>= 90% gate)
- [x] Per-file coverage gate: 146 files >= 80%, 2 waivered files >= 65%
- [x] Interrogate docstring coverage 99.6% (>= 95% gate)
- [x] Cyclomatic complexity: all new functions <= 8 (gate is <= 10)
- [x] New tests:
  - `TestMiningRunReport` — diagnostic shape, top_score/threshold semantics, per-strategy presence
  - `TestMiningDiagnosticLogLines` — INFO log line emitted for each strategy and names the threshold
  - `TestMiningFallbackGapLog` — resonance-chain fallback appends to `compile-gaps.jsonl`, bypass mode suppresses
  - `TestIdeaMinerConfigKnobs` — constructor accepts threshold overrides
  - `TestMiningConfig` — `MiningConfig` defaults match library, Pydantic validation rejects out-of-range
  - `test_mine_no_seeds_message_names_top_score_and_threshold` — CLI message names score, threshold, gap-log path
  - `test_mine_respects_creek_config_mining_knobs` — YAML knobs flow through to `IdeaMiner`

https://claude.ai/code/session_015jFVbqmaYeuDQi8fBetdRo

---
_Generated by [Claude Code](https://claude.ai/code/session_015jFVbqmaYeuDQi8fBetdRo)_

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -1691,7 +1691,11 @@ def mine(
     from creek.generate.mining import IdeaMiner
 
     config = _load_config_for_vault(vault)
-    vault_path = vault if vault is not None else config.vault_path
+    # Route the resolved vault through ``_resolve_vault`` so future
+    # symlink/existence/type-coercion hooks added to that helper apply
+    # uniformly across every CLI subcommand, regardless of whether the
+    # path came from ``--vault`` or the auto-discovered config.
+    vault_path = _resolve_vault(vault if vault is not None else config.vault_path)
     current_phase = _parse_phase(phase)
     override = _parse_include_tier(include_tier)
     if bypass_compiled:
@@ -1740,10 +1744,12 @@ def _print_no_seeds_diagnostic(
     to clear, then points the operator at ``compile-gaps.jsonl`` for
     fallback breadcrumbs.
     """
+    from creek.generate.mining import COMPILE_GAPS_LOG_RELPATH
+
     n_strategies = len(report.diagnostics)
     top_score = report.top_score
     threshold = report.top_threshold
-    gaps_path = vault_path / "00-Creek-Meta/Processing-Log/compile-gaps.jsonl"
+    gaps_path = vault_path / COMPILE_GAPS_LOG_RELPATH
     console.print(
         f"[yellow]No idea seeds surfaced. Ran {n_strategies} strategies; "
         f"top candidate scored {top_score:.2f} (threshold {threshold:.2f}). "
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -1691,10 +1691,7 @@ def mine(
     from creek.generate.mining import IdeaMiner
 
     config = _load_config_for_vault(vault)
-    # Route the resolved vault through ``_resolve_vault`` so future
-    # symlink/existence/type-coercion hooks added to that helper apply
-    # uniformly across every CLI subcommand, regardless of whether the
-    # path came from ``--vault`` or the auto-discovered config.
+    # Re-resolve so future _resolve_vault hooks apply regardless of source.
     vault_path = _resolve_vault(vault if vault is not None else config.vault_path)
     current_phase = _parse_phase(phase)
     override = _parse_include_tier(include_tier)
@@ -1738,23 +1735,32 @@ def _print_no_seeds_diagnostic(
     *,
     vault_path: Path,
 ) -> None:
-    """Print the issue #340 zero-seed diagnostic message.
+    """Print a per-strategy breakdown when no seeds surfaced.
 
-    The message names the highest score and the threshold it failed
-    to clear, then points the operator at ``compile-gaps.jsonl`` for
-    fallback breadcrumbs.
+    Each strategy reports its own score/threshold in its own units
+    (fragment counts, component sizes, Jaccard similarity, binary
+    gates) so the operator can read across them without conversion.
+    Points at ``compile-gaps.jsonl`` for fallback breadcrumbs.
     """
     from creek.generate.mining import COMPILE_GAPS_LOG_RELPATH
 
     n_strategies = len(report.diagnostics)
-    top_score = report.top_score
-    threshold = report.top_threshold
     gaps_path = vault_path / COMPILE_GAPS_LOG_RELPATH
-    console.print(
-        f"[yellow]No idea seeds surfaced. Ran {n_strategies} strategies; "
-        f"top candidate scored {top_score:.2f} (threshold {threshold:.2f}). "
-        f"See {gaps_path} for fallback reasons.[/yellow]",
-    )
+    lines = [
+        f"[yellow]No idea seeds surfaced. Per-strategy breakdown "
+        f"({n_strategies} strategies):[/yellow]",
+    ]
+    for diag in report.diagnostics:
+        suffix = f" — {diag.fallback_reason}" if diag.fallback_reason else ""
+        lines.append(
+            f"  [yellow]{diag.strategy.value:<20}"
+            f" {diag.candidates_considered} considered,"
+            f" {diag.candidates_kept} kept"
+            f" (top score {diag.top_score:.2f} /"
+            f" threshold {diag.threshold:.2f}){suffix}[/yellow]",
+        )
+    lines.append(f"[yellow]See {gaps_path} for fallback breadcrumbs.[/yellow]")
+    console.print("\n".join(lines))
 
 
 def _read_voice_core(path: Path | None) -> str:
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -36,6 +36,7 @@ if TYPE_CHECKING:
     from creek.generate.compost_embedding import CompostExemplar
     from creek.generate.compost_verifier import SupportsVerifyCompost
     from creek.generate.drafts import DraftLLM, SeedSpec
+    from creek.generate.mining import MiningRunReport
     from creek.ingest.base import Ingestor
     from creek.models import CompileTargetKind, Frequency, Mode, Phase
     from creek.purge import PurgeEngine, PurgeResult
@@ -1689,18 +1690,24 @@ def mine(
     """
     from creek.generate.mining import IdeaMiner
 
-    vault_path = _resolve_vault(vault)
+    config = _load_config_for_vault(vault)
+    vault_path = vault if vault is not None else config.vault_path
     current_phase = _parse_phase(phase)
     override = _parse_include_tier(include_tier)
     if bypass_compiled:
         _warn_bypass_compiled("mine")
-    seeds = IdeaMiner(
+    report = IdeaMiner(
         privacy_override=override,
         bypass_compiled=bypass_compiled,
-    ).mine_all(
+        min_thread_fragments=config.mining.min_thread_fragments,
+        min_chain_length=config.mining.min_chain_length,
+        similarity_liminal=config.mining.similarity_liminal,
+        similarity_resonance=config.mining.similarity_resonance,
+    ).mine_all_with_report(
         vault_path,
         current_phase=current_phase,
     )
+    seeds = list(report.seeds)
     fragment_ids = sorted({fid for seed in seeds for fid in seed.source_fragments})
     _audit_privacy_override_if_needed(
         vault_path=vault_path,
@@ -1709,7 +1716,7 @@ def mine(
         fragment_ids=fragment_ids,
     )
     if not seeds:
-        console.print("[yellow]No idea seeds surfaced.[/yellow]")
+        _print_no_seeds_diagnostic(report, vault_path=vault_path)
         return
 
     display = seeds if limit <= 0 else seeds[:limit]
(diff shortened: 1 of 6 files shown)


Geoffe-Ga:
Addressed in `f1dbc65`.

**Blocker 1 (wavelength `top_score`)** — fixed in `creek/generate/mining.py::_mine_wavelength_windows_with_diagnostic`. Now reports `float(len(phase_matches))` (pre-threshold peak) instead of `float(len(seeds))` (post-filter count). This matches the docstring and the pattern used by the other three strategies, and resolves the ambiguity where "10 phase matches all dropped by the praxis-explicit gate" rendered identically to "no phase matches at all" in the diagnostic.

**Blocker 2 (vault path bypasses `_resolve_vault()`)** — fixed in `creek/cli.py::mine`. The path now routes through `_resolve_vault(vault if vault is not None else config.vault_path)`, so both the explicit `--vault` and the config-fallback case go through the same helper. Functionally equivalent today, but future symlink/existence/type-coercion hooks added to `_resolve_vault` will apply uniformly across every subcommand.

**New test (TDD)** — added `TestMiningRunReport::test_wavelength_top_score_reports_phase_matches_not_seeds` in `tests/test_mining.py`. Seeds 10 `LATENT` phase-matching fragments, asserts `diag.candidates_considered == 10`, `diag.candidates_kept == 0`, `diag.top_score == 10.0`. Verified Red against the prior commit before applying the Blocker 1 fix.

**Minor 1 (magic path string)** — extracted `COMPILE_GAPS_LOG_RELPATH` in `creek/generate/mining.py` aliased to the canonical `compile_routing.COMPILE_GAPS_RELPATH`. `_print_no_seeds_diagnostic` now imports and reuses it instead of hardcoding the literal vault-relative path. Single source of truth.

**Minor 2 (useless alias)** — dropped the inner `import pytest as _pytest` in `test_mine_respects_creek_config_mining_knobs`. Promoted the module-level `pytest` import out of `TYPE_CHECKING` so `pytest.approx` is available at runtime without re-import.

**`_with_diagnostic` indirection** — read the note; leaving as-is per your guidance.

Local: `./scripts/check-all.sh` exits 0 (4133 passed, coverage 93.68%, all 12 quality gates green).

@claude please re-review.

---

Geoffe-Ga:
Pushed 0e82b71 to fix the failing `test_mine_command` / `test_mine_with_phase` CI tests (the ones from the log).

**Root cause**
Those two tests pre-date this PR and used `/fake/vault` as a "doesn't matter" placeholder because the original `mine` body only *read* from the vault and gracefully returned empty when missing. This PR added the resonance-chain `compile-needed` log, which writes through `log_compile_gap` on every zero-synchronicity run; that writer does `log_path.parent.mkdir(parents=True, exist_ok=True)` on `<vault>/00-Creek-Meta/Processing-Log/`. With `vault=/fake/vault`, the mkdir fails with `PermissionError: [Errno 13] Permission denied` on Ubuntu CI runners (which can't create `/fake/`) and the test exits 1. Locally the same call returns `FileNotFoundError` and is masked by the writer's own error handling, so the tests pass on a developer box but fail on every CI runner — a real CI/local skew.

**Fix**
Use `tmp_path` as the vault in both tests so the mine command can exercise its new write side against a real writeable directory. The third test (`test_mine_with_unknown_phase_errors`) intentionally keeps `/fake/vault` — Typer's argument validation rejects `--phase nonsense` with `exit_code == 2` before the mine body ever touches the path.

Local `./scripts/check-all.sh` exits 0: 4133 passed, 93.68% branch coverage, all 12 gates green.

---

Geoffe-Ga:
Addressed the CHANGES_REQUESTED blocker (and the two minor reviewer notes) in 3d5fba5.

**Blocker — `MiningRunReport.top_score` mixes incompatible units**

Both reviewers flagged that `max(diag.top_score for diag in diagnostics)` aggregates fragment counts, component sizes, Jaccard similarity ∈ [0, 1], and a binary 1.0 gate — meaningless. The failure mode they cited (vault with 12 phase-matching `LATENT`-praxis fragments, CLI reads `top candidate scored 12.00 (threshold 1.00)` while seeds=0) is the exact ambiguity issue #340 set out to eliminate.

The fix is structural rather than cosmetic:

1. **Removed `MiningRunReport.top_score` / `top_threshold` properties.** The class docstring now explicitly says callers must iterate `diagnostics` and surface per-strategy units side-by-side.
2. **`cli._print_no_seeds_diagnostic` rewritten** to emit one row per strategy with name, considered/kept counts, the strategy's own `top_score` / `threshold` pair in its own units, and any `fallback_reason`. Aggregate scalars are gone.
3. **Wavelength now sets `fallback_reason = _NO_EXPLICIT_PRAXIS`** when phase matches exist but all are dropped by the praxis gate (new constant: `"phase matches found but none had praxis_potential=EXPLICIT"`). Previously this case left `fallback_reason = None`, which is what silenced the failure mode in the original review.

**Minors picked up from the second review**

- 4-line `_resolve_vault` rationale comment in `cli.mine` shrunk to a single line per CLAUDE.md's "no multi-line comment blocks" convention.
- `_print_no_seeds_diagnostic` docstring's `issue #340` reference removed per CLAUDE.md's "issue refs in docstrings will rot" rule.

**`_best_eddy_match` dead-code note** (low-severity nit from both reviews): leaving for a follow-up — it's a one-line wrapper around `_best_eddy_match_unfiltered` and the call-site refactor was deliberate; can be cleaned up in a separate hygiene PR without bundling it here.

**Tests**

- `test_wavelength_top_score_reports_phase_matches_not_seeds` extended to assert `fallback_reason` names the praxis gate.
- `test_report_top_score_reflects_best_candidate` replaced with `test_empty_vault_yields_per_strategy_diagnostics_and_no_seeds` (aggregate property gone).
- `test_mine_no_seeds_message_names_top_score_and_threshold` renamed to `test_mine_no_seeds_message_lists_per_strategy_breakdown` and extended to assert every strategy name appears on its own row alongside score / threshold / considered / kept.

Local: `./scripts/check-all.sh` exits 0 (4133 passed, 93.70% branch coverage, all 12 quality gates green).

@claude please re-review.

---

Geoffe-Ga:
Pushed 355891a addressing both 🟡 items from the second re-review.

**Item 1 — Resonance-chain `fallback_reason=None` for short components**

This is the symmetric case to the wavelength praxis-gate-drop closed in 3d5fba5. When synchronicities exist on disk but every connected component is shorter than `min_chain_length`, the per-strategy row used to read `top_score=N / threshold=M` with no indication that the chain-length floor was the blocking gate — exactly the "almost kept" blind spot the reviewer flagged.

Added `_NO_QUALIFYING_CHAIN` constant. `_resonance_fallback_reason` signature widened to take `seeds_kept` + `largest_component` and now decides between three honest states:

- **no synchronicities at all** → `_NO_SYNCHRONICITIES` (and compile-gap log unless bypass)
- **synchronicities exist, every component < `min_chain_length`** → `_NO_QUALIFYING_CHAIN` formatted as `"synchronicities exist but largest component has 2 fragments, below min_chain_length=3"`
- **at least one chain kept** → `None`

Docstring rewritten to describe all three cases — the previous `"Log + return"` wording was inaccurate even before this change (bypass mode already suppressed the log). That ties off the third 🟢 LOW the first reviewer also called out.

**Item 2 — `_best_eddy_match` dead-code wrapper**

Definitive answer: deleted. A repo-wide grep (`grep -rn '_best_eddy_match\b' creek/ tests/`) confirms zero callers besides the wrapper's own definition. The only post-refactor call site (`_mine_liminal_cross_eddy_with_diagnostic`) already uses `_best_eddy_match_unfiltered` directly and applies `similarity_liminal` inline.

**TDD test added**

`test_resonance_chain_short_components_records_fallback_reason` builds a 2-fragment component (below default `min_chain_length=3`) linked by one synchronicity edge and asserts `candidates_kept=0`, `top_score=2.0`, and that `fallback_reason` names both `min_chain_length` and the actual largest-component size.

Local: `./scripts/check-all.sh` exits 0 (4133 passed, all 12 gates green).

Note on the two simultaneous verdicts on `3d5fba5`: I followed the later (CHANGES_REQUESTED at 22:29:42Z) per the address-feedback rule that the most-recent verdict wins. Both reviewers' lower-severity items are now closed.

@claude please re-review.

---

PR #361, opened 2026-05-27
fix(purge): scrub fragment-ID references vault-wide, not just in 02/03 (GAP-004)

Geoffe-Ga:
`creek purge` advertised "scrubbing every reference along the way" but
only walked 02-Threads and 03-Eddies via `_decrement_counts`, plus
wiki-link removal vault-wide via `_scrub_wikilinks`. Derived content
under 04-Praxis, 05-Wavelength, 07-Voice/Drafts, 08-Decisions,
09-Reference, 10-Liminal and the deployed Skills tree silently kept
references to purged fragments — in particular YAML provenance lists
like `source_fragments: [frag-…]` in draft files, and bare ID mentions
in body text. A user purging an intimate-tier fragment would still
find its synthesized draft (with both YAML provenance and prose body
copied from the original fragment) on disk after purge — `grep -r
<frag_id>` lit it up.

This change adds `_scrub_provenance(fragment_id)`: a word-boundary
regex pass over every `.md` file in the vault that replaces matching
fragment-ID mentions with the literal `[purged]` placeholder. One
pass catches both YAML list entries and prose mentions because the
substitution is text-level. Wiki-link removal (by title) stays on
`_scrub_wikilinks` since it matches by name not ID.

The compliance audit log at `00-Creek-Meta/audit/purge.jsonl` is
JSONL, not Markdown, so the recursive `.md` walk excludes it
naturally — `affected_fragments` in audit entries keeps the real ID
for forensic reconstruction. A new test pins this invariant.

Wires into:
- `purge_fragment` — direct call
- `_purge_single` (used by `purge_source`, `purge_source_path`,
  `purge_daterange`) — call per matched fragment

`PurgeResult.provenance_scrubbed` and `PurgeAuditEntry.provenance_scrubbed`
record the count. Pre-GAP-004 audit entries default to 0 via Pydantic
and the legacy coercer.

Docs (`cleaning-and-purge.md`) and CLI `--help` for `purge fragment`
/ `purge source` now enumerate the scrubbed folders explicitly.

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -2638,7 +2638,14 @@ def purge_fragment(
         help="Skip interactive confirmation",
     ),
 ) -> None:
-    """Delete a fragment and scrub every reference to it."""
+    """Delete a fragment and scrub every reference to it.
+
+    Scrubbing covers (GAP-004): wiki-links by title removed from every
+    `.md` file in the vault, plus YAML provenance lists and bare body
+    mentions of the fragment ID replaced with ``[purged]`` across
+    01-Fragments through 10-Liminal and the deployed Skills tree.
+    Embedding-cache rows for the fragment are also dropped (GAP-001).
+    """
     engine = _build_engine(vault, dry_run=dry_run)
     if not dry_run and not _confirm(
         f"Purge fragment {fragment_id!r} and all references?",
@@ -2705,7 +2712,10 @@ def purge_source(
       a configurable ``--match`` mode (INC-008): ``exact`` (default),
       ``substring``, or ``regex``.
 
-    The two are mutually exclusive; pass exactly one.
+    The two are mutually exclusive; pass exactly one. Reference
+    scrubbing (GAP-004) runs for every matched fragment: wiki-links
+    by title plus word-boundary ID mentions replaced with
+    ``[purged]`` across the whole vault.
     """
     if (source_type is None) == (source_path is None):
         console.print(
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
@claude please review.

(The previous `claude-review` workflow run completed `success` for HEAD `b2047bb` without posting a verdict comment — re-requesting so the merge gate has something to evaluate.)

---

Geoffe-Ga:
Two verdicts on this PR — Reviewer 1 LGTM, Reviewer 2 COMMENTS (latest). Both non-blocking; no CHANGES_REQUESTED. CI fully green.

Per the COMMENTS protocol, every actionable item is captured as a follow-up issue so nothing is lost when this merges:

- **#370** — Document YAML flow-list side-effect of `[purged]` marker at the code level (Reviewer 2 HIGH — the one item Reviewer 2 flagged as "worth addressing before merge"; tracking as a docstring/audit follow-up rather than blocking the safety fix from landing).
- **#371** — Tighten `_scrub_provenance` word-boundary regex for hyphen-suffixed IDs (Reviewer 1 Finding 1).
- **#372** — Share vault `.md` walk across `_scrub_wikilinks` and `_scrub_provenance` (Reviewer 1 Finding 4 / Reviewer 2 perf note).
- **#373** — GAP-004 polish bundle: lift `import re` to module level, comment the `isinstance(frag_id, str)` guard, relax `== 4` to `>= 4`, clarify `cleaning-and-purge.md` walk-scope wording.

Merging now.

---

PR #384, opened 2026-05-27
feat(aggregate): bubble UP weighted classifications when zooming out (#369)

Geoffe-Ga:
## Summary

The zoom-out twin of #368. When FEAT-022's `aggregate()` stitches `document`-level chats into `exchange` / `burst` / `session` parents, this PR wires the holonic combiner (#367) into `_build_parent` so each aggregated parent carries a `Fragment.weighted` derived from its children's weighted profiles — conviction × confidence weighted, per the #367 contract. Length is nowhere in the call path.

## Stacked on #383 (#368)

Branch base is `claude/issue-368-split-bubble-up` (PR #383 ← #382 ← #379 ← #378 ← PR #359). #369 calls `holonic.combine` (introduced in #367, available since #382 landed in the stack); no overlap with #368's split-path changes, so the two are siblings under #367 in terms of code dependency but stacked here for review-order simplicity.

## What ships

- **`AggregationConfig.weighted_fill_floor: float = 0.5`** — structural gate. When the fraction of children with non-`None` `Fragment.weighted` falls below this floor, the parent's `weighted` is `None` (the vault is too sparsely classified for this aggregation to be meaningful). Confidence-based dampening is left to the combiner; this gate only decides whether to *call* it.
- **`_combine_children_weighted(children, fill_floor)`** helper: filters `None`-weighted children, applies the fill-floor, calls `holonic.combine` with **no** `weights` override (the default per-child weight is each child's `overall_confidence`).
- **Wiring in `_build_parent`**: the aggregated parent's `Fragment.weighted` is now populated by the helper above. All other behaviour (deterministic ID, `child_ids` order, joiner-concatenated titles, source inheritance, tags) is unchanged.

## Tests (9 new, all green)

- Five high-confidence F2 chats aggregate into an F2-dominant exchange.
- Five confident-but-divergent F1/F3/F5/F7/F9 messages dampen confidence below 0.9 (divergence penalty fires).
- **Conviction-over-length**: 1 confident-short F3 message beats 4 hedged-long F5 messages in the exchange's top frequency.
- Fill-floor gate: 3-of-5 unweighted → `parent.weighted is None`.
- Fill-floor gate: 3-of-5 weighted → `parent.weighted is not None`.
- All-unweighted → no fabrication (`parent.weighted is None`).
- Zero-confidence pass-through: every child weighted at `overall_confidence=0` yields a non-`None` parent profile that is empty (honest "no signal").
- Single-child aggregation: combiner idempotence holds end-to-end.
[...]

Issue #369, opened 2026-05-27
feat(aggregate): bubble UP weighted classifications when zooming out (#364)


Geoffe-Ga:
## Context

The zoom-out twin of #368. The user's example:

> with smaller fragments that need to be combined with their neighbors, like text messages that don't say much by themselves but when taken with the two or five or however many preceding them deliver a clear insight that is clearly classifiable in the ontology, the math for determining parts vs wholes scores (or holonic scores) is necessary to pay careful intentional attention to.

When FEAT-022's `aggregate()` stitches `document`-level chats into an `exchange`, then exchanges into a `burst`, then bursts into a `session`, each new parent should carry a combined weighted classification computed from its children — via the holonic combiner (#367), with each child's influence equal to its `overall_confidence`. A child whose classifier failed soft (confidence ≈ 0) contributes nothing — chat messages that don't say much by themselves don't pretend to.

Length plays no role in the default path. A confidently-classified short message dominates a long-but-hedged one.

Parent: epic #364
Depends on: #366 (weighted classifier — so children have non-`None` `weighted` fields), #367 (combiner)

## Goal

In `creek-tools/creek/atomize/aggregate.py`, extend `_build_parent` to populate the parent's `weighted` field from its children's tuples:

1. Collect children with a non-`None` `weighted` field.
2. Apply the `weighted_fill_floor` gate (see below) — if too few children carry weighted tuples at all, set parent `weighted = None` (we don't have enough signal to fabricate one).
3. Call `holonic.combine(kept_children)` with **no** `weights` override, so each child's mass = its own `overall_confidence`. Children whose `overall_confidence` is exactly 0 contribute nothing (per #367's zero-confidence pass-through invariant) and naturally drop out without special-casing.
4. Set the parent's `weighted` field to the combined result.

### The `weighted_fill_floor` gate

`AggregationConfig.weighted_fill_floor: float = 0.5` (default). The gate is **structural, not signal-based** — it answers "do enough of my children have any classification at all?", not "are the classifications confident?".

- If `kept_fraction = len(children_with_weighted) / len(all_children) < weighted_fill_floor`, the parent's `weighted` is `None`. The vault is too sparsely classified for this aggregation to be meaningful.
- If the gate passes, the combiner takes over and **confidence does the rest of the work** — there is no separate "confidence floor" because #367's math already produces a low parent `overall_confidence` when children are dim, and downstream consumers already know to ignore low-confidence classifications.

### Cross-source caveats

[...]

PR #396, opened 2026-05-29
fix(clean): break clean.dedup ↔ ingest.base circular import

Geoffe-Ga:
## Summary

A cold-start `import creek.consent` (or any first-import of `creek.ingest.base`) currently raises on `main`:

```
ImportError: cannot import name 'generate_fragment_id' from partially
initialized module 'creek.ingest.base'
```

This is a cross-subsystem circular import that only trips when `ingest.base` is imported **first** in a clean interpreter. The in-process pytest session masks it (earlier-collected modules warm `sys.modules` in a safe order), which is why it slipped through — but a cold CLI start or `python -c "import creek.consent"` hits it every time.

## Root cause

```
ingest.base
  -> models                       (base.py:29)
  -> classify.weighted            (models.py late import, Fragment.weighted forward-ref)
  -> classify.__init__ -> llm -> llm.prompts
  -> generate.indexes -> generate.__init__ -> generate.state
  -> clean.__init__ -> clean.dedup
  -> ingest.base                  (dedup.py:26 — re-entry while partially initialised)
```

`clean/dedup.py` imported `generate_fragment_id` at module top. When the chain re-enters `ingest.base`, only its first 29 lines have run — `generate_fragment_id` (defined at `base.py:322`) doesn't exist yet → `ImportError`.

This cycle was introduced incrementally as the `generate → clean` and `classify → generate` edges were added (the recent draft-composition PRs wired `generate.state → clean.hygiene`), combining with the pre-existing `models → classify.weighted` late import.

## Fix

Move the single `generate_fragment_id` import into the one method that uses it (`Deduplicator._register`, `dedup.py:180`), breaking the back-edge of the cycle. `generate_fragment_id` is referenced exactly once, so the lazy import is a minimal, behaviour-preserving change. The pattern matches the late-import already used in `creek/models.py` for the same family of reasons.

## Regression guard

`tests/test_import_cycles.py` spawns a **fresh subprocess** per cross-subsystem entry-point module and asserts each imports cleanly. A subprocess is required because the in-process session cannot reproduce a cold import order once `sys.modules` is warm. Parametrised over `creek.consent`, `creek.clean.dedup`, `creek.ingest.base`, `creek.generate.state`, `creek.classify.llm.prompts` — two were red before this fix (`creek.consent`, `creek.ingest.base`), all five green after.

## Why now

This blocks PR #384 (the last of the holonic-weighted stack, #369): its merge of current `main` exposes the cycle in the test-collection order. Filing the fix as a standalone bugfix PR (per the maintainer's call) keeps it cleanly separated from the #369 feature; #384 will rebase on top once this lands.

## Test plan

- [x] `uv run pytest tests/test_import_cycles.py -q` → 5 passed (2 red pre-fix).
- [x] `uv run pytest tests/` (wide terminal) → 4541 passed, 1 skipped (Ollama); the pre-existing `test_purge.py` sandbox-`COLUMNS` artifact deselected — unrelated, green in CI.
- [x] `uv run mypy creek/` → 0 issues, 136 files.
- [x] Ruff lint + format, pylint (9.88/10 on `dedup.py`), xenon, refurb, tryceratops, bandit, interrogate (100%) — all clean.

https://claude.ai/code/session_01Rmvpqhr5ktg7abxiHMRdTj

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rmvpqhr5ktg7abxiHMRdTj)_

### DIFF
--- a/creek-tools/creek/clean/dedup.py
+++ b/creek-tools/creek/clean/dedup.py
@@ -175,10 +175,7 @@ class Deduplicator:
         if result.is_duplicate:
             return result
 
-        # Imported lazily to break the import cycle clean.dedup ->
-        # ingest.base -> models -> classify -> generate -> clean.dedup:
-        # a module-level import re-enters ingest.base before
-        # generate_fragment_id is defined.
+        # Lazy import breaks the dedup -> ingest.base cycle (test_import_cycles.py).
         from creek.ingest.base import generate_fragment_id
 
         fragment_id = generate_fragment_id(source, timestamp, content)
--- a/creek-tools/creek/clean/dedup.py
+++ b/creek-tools/creek/clean/dedup.py
@@ -23,8 +23,6 @@ from typing import Literal
 
 from pydantic import BaseModel
 
-from creek.ingest.base import generate_fragment_id
-
 _MANIFEST_RELATIVE = Path("00-Creek-Meta") / "dedup-manifest.json"
 """Relative path from vault root to the deduplication manifest file."""
 
@@ -176,6 +174,12 @@ class Deduplicator:
         result = self._check_hashes(exact_hash, normalized_hash)
         if result.is_duplicate:
             return result
+
+        # Imported lazily to break the import cycle clean.dedup ->
+        # ingest.base -> models -> classify -> generate -> clean.dedup:
+        # a module-level import re-enters ingest.base before
+        # generate_fragment_id is defined.
+        from creek.ingest.base import generate_fragment_id
 
         fragment_id = generate_fragment_id(source, timestamp, content)
         self._exact_index[exact_hash] = fragment_id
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
Folded in both non-blocking notes from the reviews in `b60f2d7`:

- ✅ Added `timeout=30` to the `subprocess.run` in `test_import_cycles.py` — a future side-effectful import that blocks now fails fast as `TimeoutExpired` instead of hanging CI.
- ✅ Condensed the 4-line lazy-import comment in `dedup.py` to a single line per CLAUDE.md; the full import-chain narrative lives in the PR body and the test module docstring.

No behaviour change. Local: 5/5 import-cycle tests + 43 dedup/import tests green; ruff, mypy strict (137 files), pylint 9.88 clean.

The pre-existing `_compute_exact_hash` / `generate_fragment_id` hash-formula duplication one reviewer noted is genuinely out of scope for this circular-import fix (collapsing it needs the larger refactor the cycle currently prevents) — leaving it untouched.

@claude please re-review.

---

Issue #393, opened 2026-05-29
Strip issue-number references from epic #349 docstrings (CLAUDE.md hygiene)


Geoffe-Ga:
Follow-up from #389 (PR for #354) — https://github.com/Geoffe-Ga/Creek-Vault/pull/389#issuecomment-4566109832

## Reviewer note (verbatim)

> `CLAUDE.md` is explicit: "Don't reference the current task, fix, or callers… since those belong in the PR description and rot as the codebase evolves." `outline.py`'s module docstring alone has six "(issue #354)" references; `drafts.py` has them throughout method docstrings.

## Requested change

Sweep the epic #349 surface (`creek/generate/outline.py`, `creek/generate/twist.py`, `creek/generate/grounding.py`, `creek/generate/dimensional_retrieval.py`, `creek/classify/prompt.py`, and the touched parts of `creek/generate/drafts.py` / `creek/cli.py`) and replace `(issue #NNN)` / `(#NNN)` docstring references with an invariant-focused sentence where the *why* is non-obvious, or drop them where the code is self-explanatory. PR descriptions and `git blame` already carry the issue linkage.

## Test idea

A lightweight repo check (or interrogate config) that flags `issue #` / `(#\d+)` in docstrings under `creek/` could prevent regression — optional.

## Notes

Epic-wide chore: the issue-ref pattern is present across several already-merged PRs in epic #349, so a single cleanup pass is more consistent than fixing one file. The reviewer marked this as the one ❌ CLAUDE.md-adherence item on #389 (non-blocking).

Geoffe-Ga:
Closing as complete — the docstring issue-ref sweep across the epic #349 surface landed in three merged PRs (split to keep file ownership conflict-free during the parallel batch):

- #400 — `outline.py`, `twist.py`, `dimensional_retrieval.py`, `classify/prompt.py`
- #404 — `generate/drafts.py` and the draft-CLI docstrings
- #405 — `generate/grounding.py`

`rg 'issue #|\(#[0-9]+\)'` over those files now returns no docstring hits. PR descriptions and `git blame` carry the issue linkage instead.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #380, opened 2026-05-27
feat(draft): bidirectional grounding guard (#355)

Geoffe-Ga:
## Summary

After a draft is generated, the new guard runs paragraph-level cosine similarity against the source fragments that fed the prompt and surfaces two opposite failure modes called out by issue #355:

- **Too derivative** — at least one draft paragraph paraphrases a single source paragraph closely (max similarity above `draft.derivative_upper`, default `0.85`). Voice-true paraphrase that adds nothing new.
- **Too ungrounded** — fewer than `draft.grounding_lower` (default `0.30`) fraction of draft paragraphs have any meaningful overlap with a source paragraph. Voice-true on the surface, conceptually invented.

The guard's verdict is surfaced everywhere the issue asks for:

- **stderr** — one-line `grounding guard: derivative=… (upper …), grounding=… (lower …) — within bounds|flagged: too derivative + too ungrounded`.
- **Frontmatter** — `derivative_score`, `grounding_score`, plus per-paragraph annotations under `paragraph_grounding`.
- **Lint** — new deterministic `draft-grounding` check audits saved drafts whose stored scores fall outside the configured bounds; pre-guard drafts without scores are skipped.

## Architecture

- New `creek/generate/grounding.py` does the paragraph-split + cosine-scoring math with a swappable `EmbeddingFn` callable so tests stay offline and deterministic. `GroundingThresholds.from_config` resolves the YAML knobs; `default_embedding_fn` builds the production callable lazily so the sentence-transformer never imports at unit-test time.
- New `creek/lint/checks/draft_grounding.py` reads each saved draft's frontmatter and emits one finding per failure with the offending metric + threshold in the message.
- `creek/config.py` gains `DraftConfig` (`derivative_upper`, `grounding_lower`, both clamped to `[0.0, 1.0]`) under the `draft:` YAML section.
- `creek/generate/drafts.py` diff is the minimum needed to wire the guard: append `derivative_score`/`grounding_score`/`paragraph_grounding` fields to `Draft` (defaults preserve every existing test), accept an optional `embedding_fn` + `grounding_thresholds` on `DraftGenerator`, call the guard from both `generate_draft` and `generate_seeded_draft`, and stamp the scores into `save_draft`'s frontmatter. No existing logic mutated — the guard short-circuits when no embedding callable is wired so the historical test suite keeps passing without dragging in `sentence-transformers`.

[...]

Issue #355, opened 2026-05-26
feat(draft): bidirectional grounding guard — flag drafts that are too derivative OR too ungrounded (#349)


Geoffe-Ga:
## Context

A drafting system has two failure modes:

- **Too derivative.** Output paraphrases a single source fragment closely. Voice-true but says nothing new. (The walkthrough's first draft.)
- **Too ungrounded.** Output uses none of the user's actual musings. Voice-true on the surface but conceptually invented — not what the user has thought. (The risk of pure signature-only composition.)

Both are failure modes, on opposite sides of the same axis. The desired band is the middle: real source material, recombined and re-twisted.

Parent: voice-faithful synthesis epic (linked in the issue tracker)

## Goal

After a draft is generated, compute two metrics and surface them:

1. **Derivative score:** maximum normalised semantic similarity between any contiguous N-gram (or paragraph) of the draft and any contiguous N-gram of any source fragment used. High = paraphrase.
2. **Grounding score:** for each conceptually-anchored claim in the draft (extracted via the same provenance pipeline used by `creek compile`), is there a source fragment in the user's vault that motivated it? Low = invented material.

Configurable thresholds in `creek_config.yaml` under `draft:`:

- `derivative_upper`: e.g. `0.85` — above this, a paragraph is treated as paraphrase and flagged
- `grounding_lower`: e.g. `0.30` — below this fraction of claims grounded in real sources, the draft is flagged as too invented

Surface both in:

- Stderr during `creek draft` (one-line summary)
- The draft's frontmatter (`derivative_score`, `grounding_score`, plus per-paragraph annotations)
- Optionally as a `creek lint` finding for drafts that fail either bound

## Scope

**In scope:**
- Two scoring functions in a new `creek/draft/grounding.py`
- Thresholds in `creek_config.yaml`
- Frontmatter additions
- Lint integration (new check name `draft-grounding`)
- Tests covering: clearly-derivative draft, clearly-invented draft, well-balanced draft

**Out of scope:**
- Automatic redraft on failure (the system surfaces the score; the human decides)
- N-gram detection beyond paragraph-level semantic similarity (could be future work)

## Acceptance criteria

- [ ] A draft known to be paraphrase-heavy (e.g., the walkthrough's first draft) scores high on derivative and gets flagged
- [ ] A draft known to be invented (synthetic test fixture: voice signatures only, no real source) scores low on grounding and gets flagged
- [ ] A well-balanced draft passes both thresholds
[...]

PR #405, opened 2026-05-29
feat(grounding): harden cosine guard and split grounding thresholds

### LINKED ISSUE #387: Split DraftConfig.grounding_lower into per-paragraph floor and grounded-fraction minimum
Follow-up from #380 (PR for #355) — https://github.com/Geoffe-Ga/Creek-Vault/pull/380#issuecomment-4559202063

## Reviewer note (verbatim)

> `grounding_lower` controls two distinct things: (1) whether an individual paragraph "counts as grounded" and (2) the minimum acceptable _fraction_ of grounded paragraphs. These are intentionally coupled (documented in the docstring), but an operator who wants a strict grounding threshold per-paragraph (say 0.4) with a lenient fraction requirement (say 0.2 of paragraphs) cannot express that with the current schema.

## Requested change

Split the single `draft.grounding_lower` knob in `creek/config.py` (`DraftConfig`) into two independent thresholds:

- `grounding_lower` — the per-paragraph cosine-similarity floor for "is this paragraph grounded at all"
- `grounding_fraction_lower` (new) — the minimum acceptable fraction of grounded paragraphs across the whole draft

`creek/generate/grounding.py` (`score_draft` / `GroundingReport.is_flagged_grounding`) and `creek/lint/checks/draft_grounding.py` both read the single value today; thread the second knob through both.

## Test idea

Add a config-plumbing test: a draft where every paragraph clears the per-paragraph floor but the grounded *fraction* is below `grounding_fraction_lower` should flag, and vice-versa — proving the two knobs are independent.

## Notes

This is a documented design trade-off in the current PR, not a bug. Defaults stay lenient (`0.30` for both) so the split is backwards-compatible. Best done once a calibration set exists.

### LINKED ISSUE #388: Harden cosine_similarity against near-zero (not just exactly-zero) embedding norms
Follow-up from #380 (PR for #355) — https://github.com/Geoffe-Ga/Creek-Vault/pull/380#issuecomment-4559202063

## Reviewer note (verbatim)

> `cosine_similarity` zero-norm check uses float equality (`if 0.0 in (norm_a, norm_b)`), which is mathematically correct only for an exact all-zeros vector. A near-zero vector (e.g. sum-of-squares ≈ 1e-30) would not be caught and would produce a very large but finite cosine.

## Requested change

In `creek/generate/grounding.py`, replace the exact zero-norm guard in `cosine_similarity` with an epsilon floor (e.g. `if norm_a < 1e-12 or norm_b < 1e-12: return 0.0`) so a degenerate near-zero embedding collapses to "no resonance" rather than producing a noisy large cosine.

## Test idea

Add a `cosine_similarity` test with a vector whose norm is ~1e-30 against a normal unit vector; assert the result is `0.0` rather than a large finite value.

## Notes

Non-blocking and low-probability in practice — production sentence-transformer output is never near-zero unless explicitly constructed. Worth hardening since `EmbeddingFn` is a public injection point. The reviewer also flagged that a dimension-mismatch `ValueError` propagates uncaught through `_score_one_paragraph → score_draft → _build_guard_report` to an unformatted traceback; consider catching it at the `_build_guard_report` boundary and surfacing a clean message in the same change.

Geoffe-Ga:
## Summary

Hardens the `creek draft` grounding guard and decouples its two grounding knobs.

### #388 — cosine guard hardening (`creek/generate/grounding.py`)
- Replaced the exact zero-norm guard (`if 0.0 in (norm_a, norm_b)`) with a `1e-12` epsilon floor (`if norm_a < 1e-12 or norm_b < 1e-12`) so a degenerate near-zero embedding (norm ~1e-30) collapses to `0.0` instead of dividing into a large-but-finite, meaningless cosine.
- A dimension-mismatch `ValueError` previously propagated uncaught through `_score_one_paragraph → score_draft → _build_guard_report` to a raw traceback. `cosine_similarity` now raises a dedicated `GroundingDimensionError` (a `ValueError` subclass, so the documented contract holds) and `score_draft` re-raises it with an operator-actionable message at the guard boundary.

### #387 — split grounding thresholds (`creek/config.py`, `creek/generate/grounding.py`, `creek/lint/checks/draft_grounding.py`)
- Kept `grounding_lower` as the per-paragraph cosine floor and added an independent `grounding_fraction_lower` (minimum fraction of grounded paragraphs, default `0.30`), placed immediately after `grounding_lower` in `DraftConfig`.
- Threaded the new knob through `GroundingThresholds`, `GroundingReport.is_flagged_grounding`, the stderr summary line, and the `draft-grounding` lint check (which now compares the stored fraction against `grounding_fraction_lower`).
- Defaults stay `0.30 / 0.30`, so behaviour is backwards-compatible.

### #393 (grounding.py portion)
- Stripped `(issue #355)` / `PR #380` references from `grounding.py` docstrings in favour of invariant-focused wording.

## Tests
- Near-zero (norm ~1e-30) cosine collapses to `0.0`.
- Mismatched embedding callable raises a clean `GroundingDimensionError`.
- Two-knob independence at both layers: a draft where every grounded paragraph clears the per-paragraph floor but the grounded fraction is below `grounding_fraction_lower` flags (and the reverse passes); the lint check exercises the same independence.

`./scripts/check-all.sh` passes (exit 0): 12/12 checks green, 4558 tests pass, coverage 93.52%.

Closes #387
Closes #388
Part of #393

https://claude.ai/code/session_011uEVKaHvU9qMDkZiGguDsU

---
_Generated by [Claude Code](https://claude.ai/code/session_011uEVKaHvU9qMDkZiGguDsU)_

### DIFF
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -720,14 +720,22 @@ class DraftConfig(BaseModel):
     * **Too derivative** — at least one draft paragraph paraphrases a
       source-fragment paragraph closely (cosine similarity above
       :attr:`derivative_upper`).
-    * **Too ungrounded** — fewer than :attr:`grounding_lower` fraction of
-      draft paragraphs share *any* similarity with a source paragraph
-      above the same lower bound, i.e. the draft is conceptually
-      invented.
+    * **Too ungrounded** — fewer than :attr:`grounding_fraction_lower` of
+      the draft's paragraphs clear the per-paragraph :attr:`grounding_lower`
+      similarity floor against *any* source paragraph, i.e. the draft is
+      conceptually invented.
 
-    Defaults are deliberately lenient (0.85 / 0.30); operators tighten
-    once they have a calibration set for their own corpus. Both
-    thresholds are cosine similarities and so must live in ``[0.0, 1.0]``.
+    The two grounding knobs are independent: :attr:`grounding_lower`
+    decides whether a single paragraph counts as grounded, while
+    :attr:`grounding_fraction_lower` decides what share of the draft's
+    paragraphs must be grounded for the draft to pass. An operator can
+    demand a strict per-paragraph anchor (e.g. 0.4) while tolerating a
+    lenient grounded fraction (e.g. 0.2), or the reverse.
+
+    Defaults are deliberately lenient (0.85 / 0.30 / 0.30); operators
+    tighten once they have a calibration set for their own corpus. All
+    three knobs are cosine similarities or fractions and so must live in
+    ``[0.0, 1.0]``.
     """
 
     derivative_upper: float = Field(default=0.85, ge=0.0, le=1.0)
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
@claude please review — the review job completed but no verdict comment was posted on this PR. Requesting a fresh review so the verdict can be recorded before merge.

---

Issue #419, opened 2026-05-31
FEAT-040.2: Idiolect / voice-fingerprint profiler — measure how *this* user actually writes


Geoffe-Ga:
Parent epic: #417

## Context
The heart of the reframe (FEAT-040). Produces the per-vault `VoiceFingerprint`: a quantitative measurement of the user's real writing habits, used as the false-positive authority and the distance baseline for every detector, and as the source of the idiolect prompt preamble. Complements the qualitative exemplars already harvested by the Voice Skill Tree.

## Scope
- `creek/generate/ai_style/fingerprint.py`:
  - `VoiceFingerprint` (typed, serialisable): for each measurable feature, the user's rate/distribution — AI-vocab per-word rates (so we know which "AI words" the user genuinely uses), em-dash density, curly-quote/apostrophe usage, copula (`is`/`are`/`has`) vs marketing-verb (`serves as`/`boasts`/`features`) ratio, sentence-length distribution (mean/spread), paragraph length, transition-opener rate (`Additionally,`…), rule-of-three frequency, negative-parallelism rate, heading-case habit (sentence vs title), hedging/disclaimer rate, and overall lexical-diversity/type-token stats. Store sample sizes + confidence per feature.
  - `build_fingerprint(vault_path, config) -> VoiceFingerprint` and persistence to `00-Creek-Meta/voice-fingerprint.json` (refreshable; versioned).
- **Authorship filter (critical correctness requirement):** include only genuinely user-authored text. Concretely: `source.author == self`; for `platform in {chatgpt, claude}` conversation fragments use **only the user-turn portion** (or exclude them if the body cannot be split cleanly), since the assistant half is AI text that would poison the baseline; weight `journal`/`markdown` and `substack` highest. Make the inclusion rule explicit and configurable.
- Confidence/thresholding: expose `fragment_count` and per-feature support so issue 01 can soften flagging when the corpus is thin.

## Out of scope
Detection, scoring, repair, prompts (other issues). The qualitative exemplar harvest (already done by `creek skills generate` / `creek report --type voice`).

## Design constraints
- **Do not fingerprint AI text.** A test must prove that assistant-turn content from a ChatGPT export does not enter the fingerprint (e.g. construct a fragment whose user-turn says "I reckon" and assistant-turn is full of "tapestry/underscore", and assert the AI-vocab rates reflect only the user-turn).
- Deterministic, offline, frontmatter-safe (read bodies only).
[...]

Issue #417, opened 2026-05-31
FEAT-040 (epic): Make generated text sound like *this user*, by measuring their voice from the vault


Geoffe-Ga:
## Summary
The goal is not "make output undetectable as AI." It is: **make everything the generation pipeline emits sound like the specific person whose writing is in this vault.** AI has a strong, recognisable accent; even with the full Voice Skill Tree and grounded source material (FEAT-032), generic AI tropes still leak into `creek draft` output. This epic measures the user's actual idiolect from their own ingested writing and uses that measurement to (a) steer generation toward their voice and (b) detect and close the gap where the AI accent has crept in.

The Wikipedia field guide *"Signs of AI writing"* (WP:AISIGNS) is the **prior** — the candidate set of features worth looking at. The user's vault is the **posterior** — the authority on which of those features are actually foreign to *this* writer. A "tell" is therefore redefined: it is not "a word on a list," it is **a measurable divergence between the draft and the user's own writing.**

## The reframe (read this first — it changes every child issue)
Previously the design flagged features against a generic human baseline with generic false-positive caveats. That is wrong for our goal. Two consequences:

1. **False-positive caveats are derived from the vault, not hand-written.** If the user demonstrably uses em-dashes, curly quotes, the rule of three, or the word "tapestry" at rate R in their own authored fragments, then the draft using them at ≤ R is *voice*, not a tell — suppress it. The candidate tell list only tells us *what to measure*; the vault tells us *what to flag*.
2. **The score is a distance, not a count.** We score a draft by how far its feature profile sits from the user's measured feature profile (the "voice fingerprint"), in both directions: over-using something the user avoids, AND under-using something the user does (e.g., the user writes short declarative `is`/`has` sentences but the draft is wall-to-wall `serves as`/`stands as`). Closing the gap — not zeroing a checklist — is the objective.

## The one critical data-hygiene constraint
The vault is built by ingesting the user's sources — and several of those are **not the user's own prose**:
- ChatGPT/Claude conversation fragments are *turn pairs*; the assistant half is AI text. Fingerprinting it would poison the baseline with the exact accent we're trying to remove.
- Substack posts are the user's published (often edited) voice — high signal.
- Journals / personal markdown are the purest signal.

[...]

PR #429, opened 2026-05-31
feat(ai-style): voice-fidelity detector framework (FEAT-040.1)

### LINKED ISSUE #418: FEAT-040.1: Detector framework, distance-to-voice scoring, and calibrate-against-own-writing harness
Parent epic: #417

## Context
Backbone for FEAT-040. Establishes the typed catalog, the scan engine, the **distance** score model (draft profile vs the user's voice fingerprint), the config, and a calibration harness whose negative set is *the user's own writing*. Every later detector registers into this framework; the framework consumes the `VoiceFingerprint` produced by issue 02.

## Scope
- `tells.py` — typed `Tell`: `id`, `category: Literal["mechanical","lexical","rhetorical","discourse","citation"]`, `feature_key` (the measurable feature this tell corresponds to in the fingerprint, e.g. `"em_dash_density"`, `"ai_vocab.tapestry"`, `"rule_of_three_rate"`), `handling: Literal["autofix","prevent","surface"]`, `description`, `measure(text) -> float` (the draft's rate for this feature), and `detect(text, *, fingerprint, config, context) -> list[Span]`. A `TELL_REGISTRY` with `register()`.
- `model.py` — `Finding` (tell id, span, line, excerpt, `draft_rate`, `user_rate`, `direction: "over"|"under"`) and `ScanReport` (findings, per-feature draft-vs-user deltas, and a scalar `voice_distance`).
- **Distance score** (the crux): `voice_distance(draft_profile, fingerprint, weights)` = weighted, length-normalised divergence between the draft's measured feature vector and the user's. Penalise **both** over-use (draft rate ≫ user rate on avoid-features) and under-use (draft rate ≪ user rate on signature features). A feature where draft ≈ user contributes ~0 regardless of its generic "AI-ness." Weights live in config.
- `scanner.py` — `scan(text, *, fingerprint, config, context) -> ScanReport`; each tell fires only when the draft's `measure()` diverges from the fingerprint's stored rate for that `feature_key` beyond a configurable margin. **If the vault has no measurement for a feature** (sparse corpus), fall back to a conservative generic band (documented) rather than flagging aggressively.
- `AIStyleConfig` in `creek/config.py`: per-feature weights, divergence margins, `voice_distance_upper` threshold, category toggles, `min_fingerprint_fragments` (below which the system warns the fingerprint is unreliable and softens flagging), `enabled`.
- **Calibration harness** `creek/generate/ai_style/calibration/`: instead of a hand-labelled fixture, the negative set is the **user's own authored fragments** (via the 02 profiler's authorship filter) and the positive set is the guide's quoted AI examples (shipped as data). `calibrate(vault) -> CalibrationReport` reports: false-positive rate on the user's real writing (must be near zero — this is the headline metric), detection rate on the AI examples, and per-feature deltas. Mirror `creek/classify/calibration.py` + `DEFAULT_FLOORS`.

Seed with one trivial tell (literal `2025-xx-xx`) to prove the engine.

## Out of scope
The fingerprint computation itself (02). Any repair, prompt injection, guard, lint (later). Real catalogs (03–07).

## Design constraints
- Score is **length-normalised** and **vault-relative**; one stray word must never dominate.
- Graceful degradation when the fingerprint is thin (small/new vault): soften toward the generic prior, and surface a "fingerprint based on only N fragments" caveat rather than over-flagging.
- Pure/deterministic/no-network.

## Files to touch
- new: `creek/generate/ai_style/{__init__,tells,model,scanner}.py`, `calibration/*`
- edit: `creek/config.py` (`AIStyleConfig`), `generate_default_config`
- tests: `tests/generate/ai_style/test_scanner.py`, `test_distance.py`, `test_calibration.py`

## Acceptance criteria
- `voice_distance` is ~0 when a draft's feature vector equals the fingerprint, rises with divergence in either direction, and is documented.
- `calibrate()` reports near-zero false positives on the user's own fragments and non-zero detection on the AI examples; fails loudly if the user's writing flags above a configurable floor.
- Thin-fingerprint fallback exercised by a test. check-all.sh green; ≥90% cov.

## Est. LOC
~550–700. Depends on: 02 (interface; can be co-developed, but 02's `VoiceFingerprint` type should land first or be stubbed). Blocks: 03–10.

Geoffe-Ga:
## Summary
- Adds the FEAT-040 AI-style / voice-fidelity backbone under `creek/generate/ai_style/`: typed model (`Finding`, `ScanReport`, `VoiceFingerprint`), a registrable `Tell` catalog with one seed tell (placeholder dates), a **bounded, vault-relative `voice_distance`** score (penalises over-use of avoided features and under-use of signature features; a feature matching the user contributes ~0), the `scan()` engine, and a calibrate-against-own-writing harness.
- Wires `AIStyleConfig` into `CreekConfig` (category/feature weights, divergence margins, thin-fingerprint softening, calibration FP floor).
- The `VoiceFingerprint` type lands here so issue #419 (profiler) can populate it; later issues register the real catalogs and add the sanitizer/guard/lint.

## Design notes
- **Vault-relative by construction**: tells only *measure*; whether a divergence matters is decided against the user's stored rate, so a user who genuinely writes "tapestry" or uses em-dashes is not flagged.
- **Thin-corpus safety**: below `min_fingerprint_fragments` the distance is softened and reports carry `thin_fingerprint`, rather than over-flagging a sparse vault.
- Pure, deterministic, no network.

## Test plan
- `./scripts/check-all.sh` → all 12 checks pass (ruff, mypy strict, complexity, bandit, interrogate, full suite).
- 41 new tests; **100% coverage** of every new module; aggregate 93.73%.
- Headline test: a fingerprint where the user uses a feature suppresses the flag; the seed tell still detects the shipped AI examples.

Closes #418
Refs #417

### DIFF
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -766,6 +766,97 @@ class DraftConfig(BaseModel):
     is flagged ``truncated`` in frontmatter and warned about on stderr."""
 
 
+class AIStyleConfig(BaseModel):
+    """Configuration for the FEAT-040 AI-style / voice-fidelity subsystem.
+
+    The subsystem measures how far generated text diverges from the
+    user's own writing (their *voice fingerprint*) rather than from a
+    generic baseline. A "tell" fires for a feature only when the draft's
+    measured rate differs from the user's stored rate by more than
+    :attr:`default_margin` (or a per-feature override). The scalar
+    ``voice_distance`` aggregates those divergences, weighted by
+    category. Everything here is intentionally conservative: the headline
+    correctness property is that the user's *own* writing must not flag.
+    """
+
+    enabled: bool = True
+    """Master switch. When ``False`` the scanner returns an empty report."""
+
+    voice_distance_upper: float = Field(default=0.35, ge=0.0)
+    """Distance above which a draft is considered to diverge from the
+    user's voice enough to warrant a rewrite pass / lint finding."""
+
+    default_margin: float = Field(default=0.5, ge=0.0)
+    """Default divergence margin, in absolute rate units, a feature must
+    exceed before it produces a :class:`~creek.generate.ai_style.model.Finding`.
+    Rates are per-1000-words, so ``0.5`` means "half an occurrence per
+    thousand words above the user's own rate"."""
+
+    min_fingerprint_fragments: int = Field(default=5, ge=1)
+    """Below this many user-authored fragments the fingerprint is treated
+    as thin: divergence contributions are softened toward the generic
+    prior and reports carry a ``thin_fingerprint`` flag rather than
+    flagging aggressively."""
+
+    thin_fingerprint_softening: float = Field(default=0.25, ge=0.0, le=1.0)
+    """Multiplier applied to distance contributions when the fingerprint
+    is thin (see :attr:`min_fingerprint_fragments`). ``0.25`` keeps a
+    quarter of the signal so obvious tells still surface while the
+    measurement is unreliable."""
+
+    calibration_fp_floor: float = Field(default=0.05, ge=0.0, le=1.0)
+    """Maximum acceptable false-positive rate on the user's own writing
+    during calibration. Above this, :func:`calibrate` reports failure."""
+
+    enabled_categories: list[str] = Field(
+        default_factory=lambda: [
+            "mechanical",
+            "lexical",
+            "rhetorical",
+            "discourse",
+            "citation",
+        ],
+    )
+    """Tell categories the scanner runs. Trimming this disables a whole
+    family of detectors without touching the registry."""
+
+    category_weights: dict[str, float] = Field(
+        default_factory=lambda: {
+            "mechanical": 0.5,
+            "lexical": 1.0,
+            "rhetorical": 1.0,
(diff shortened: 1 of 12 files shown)


Geoffe-Ga:
Thanks — fixed the blocking `# type: ignore[arg-type]`. `_findings_for` now takes `direction: Direction` and the `scan()` caller annotates the narrowed literal, so mypy strict passes with no ignore. Also added the one-line docstring note that `under`-use findings are doc-level by design (per the non-blocking nit). `check-all.sh` is 12/12 green. The other notes are noted as future/educational. Pushed as $(git rev-parse --short HEAD).

Geoffe-Ga:
Addressed all three review suggestions in $(git rev-parse --short HEAD): `le=1.0` on `voice_distance_upper`; `enabled_categories`/`category_weights` typed against the shared `AIStyleCategory` literal (kept in `config` to avoid an import cycle, re-exported as `Category`); and a `summary()` FAIL-branch assertion. `check-all.sh` 12/12 green. The `MappingProxyType` immutability note I'll fold into #419 (the profiler) since that's where a live dict could be passed in.

PR #434, opened 2026-06-01
feat(ai-style): vault-relative rhetorical detectors (FEAT-040.6)

### LINKED ISSUE #423: FEAT-040.6: Rhetorical detectors — significance/legacy/trend, superficial analysis, puffery, weasel attribution (vault-relative)
Parent epic: #417

## Context
Detection-only for the *substantive* tells. These point at deeper problems (ungrounded synthesis, padding, opinion-as-fact) and are never auto-stripped — they feed prevention (08), targeted rewrite (09), and surfacing (10). Suppression is fingerprint-driven: a user who genuinely writes in a grand, legacy-minded register should not be flagged for doing so.

## Scope (per the guide)
- **Undue emphasis on significance / legacy / broader trends** (WP:AILEGACY/WP:AITREND): `marking a pivotal moment in the evolution of…`, `represents a significant shift`, `contributing to the broader…`, `stands as a testament`, `enduring legacy`, the biology over-framing variant, and hedged-preamble-then-puff.
- **Superficial analyses** (WP:SUPERFICIAL): trailing `-ing` clauses (`…, highlighting/underscoring/reflecting…`), incl. the RAG "Roger Ebert highlighted the lasting influence" variant; "has generated debate about X, Y, and Z".
- **Promotional / peacock language** (WP:AIPUFFERY): `boasts`, `vibrant`, `nestled`, `in the heart of`, travel-guide/press-release register; the cultural-heritage-importance reminder subtype.
- **Vague attribution & overgeneralisation** (WP:AIWEASEL): `Observers have cited`, `Experts argue`, `several sources` while citing one; one source presented as widely held.

## Out of scope
Auto-removal. Parallelisms/rule-of-three/structure (07). Lexical density (05).

## Design constraints
- **Surface, don't strip**; output the deeper-problem hint (e.g. "trailing `-ing` clause = likely unattributed synthesis → reground or cut").
- **Vault-relative suppression:** compare the draft's rate of these constructions to the user's measured rate (the fingerprint tracks `superficial_ing_rate`, `significance_phrase_rate`, etc.). A user whose own essays run grand does not get flagged for grandeur; the signal is the *delta*.
- Down-weight when a real inline citation is attached (compose with 11 when available); until then, low severity + caveat.
- Bias to precision: over-flagging substantive prose erodes trust. Negative set = the user's own fragments.

## Files to touch
- new: `creek/generate/ai_style/rhetorical.py`, optional phrase data in `data/`
- tests: `tests/generate/ai_style/test_rhetorical.py` (Idescat, Douéra, McAllen-Temple, Cadillac-Sollei AI positives; user's own feature-prose fragments as negatives that must not flag)

## Acceptance criteria
- AI examples flag with `rhetorical` + deeper-problem hint; the user's genuine register (per a fixture fingerprint) does not flag above baseline; sourced "critics argued [cite]" down-weighted.
- Precision recorded on the user's corpus. check-all.sh green; ≥90% cov.

## Est. LOC
~550–700. Depends on: 01, 02.

Geoffe-Ga:
## Summary
Four **surface-only** `rhetorical` tells — the *substantive* AI tells (significance/legacy puffery, superficial trailing `-ing` analysis, promotional peacockery, vague attribution). They point at deeper problems (ungrounded synthesis, padding, opinion-as-fact), so they are flagged for review/rewrite and **never auto-stripped**.

All four are **vault-relative**: their extractors are fingerprinted in `features.py`, so a writer whose own prose runs grand or promotional has a high baseline and isn't flagged. Each tell reuses the shared compiled pattern + extractor (DRY), so the locator highlights exactly what the measurer counted.

## Test plan
- `./scripts/check-all.sh` → 12/12; new module at 100% coverage.
- **Two-branch tests** for each tell (fires for a plain-user fingerprint, suppressed for an ornate-user one); a single significance phrase in 1000 words does not false-positive; findings carry a span + a "surface, don't strip" hint.

Closes #423
Refs #417


Geoffe-Ga:
Addressed both review items in d0cb044:

- **🔴 BLOCKING — `sparr?\w+` typo:** replaced with `spark\w*` so the "sparked/sparking debate about" family is actually matched. Added `test_superficial_ing_matches_spark_family` covering `sparked` and `generated`.
- **🟡 HIGH — missing `\b` in `PEACOCK_RE`:** anchored `in the heart of`, `rich cultural heritage`, and `natural beauty` with word boundaries. Added `test_peacock_phrase_respects_word_boundary` asserting `supernatural beauty` no longer false-positives while `natural beauty` still fires.

The non-blocking notes (magic-number rationale on `_PHRASE_MARGIN`, span-assertion quality) are already documented inline; happy to spin those into a follow-up if you'd prefer.

`./scripts/check-all.sh` green: 12/12, 4732 passed, 93.72% coverage.

Geoffe-Ga:
Addressed in 1623466 — you were right on both counts; an errant `git add -A` had swept runtime artifacts and unrelated infra into this branch.

**🔴 Blocking — resolved:**
- Removed `.claude/scheduled_tasks.lock` from tracking and added it to `.gitignore` (it's a live harness PID/session lock, never source).
- Removed `scripts/ralph/state.json` (mutable runtime state) from this PR.

**Unbundled (the noisy extras):** the Ralph loop infra (`scripts/ralph/*`, `.claude/commands/ralph-tick.md`) and the backlog/epic plan files are out of this branch too. They remain on disk and will land via their own dedicated change — this PR is now strictly FEAT-040.6.

**Non-blocking — done:** documented `_PHRASE_MARGIN` as additive, in phrases per 1000 words.

The PR diff vs `main` is now the 5 feature files + the one `.gitignore` line. `./scripts/check-all.sh` green: 12/12, 93.72% coverage.

The sparse-fingerprint test gap you noted (older fingerprint missing the new `feature_key`s) is a good catch — I'll file it as a follow-up rather than expand scope here.

Issue #464, opened 2026-06-02
feat(crawdad): /crawdad ask + /crawdad draft routing + ai-as-user save


Geoffe-Ga:
## Role

You are a senior Python engineer working in the `crawdad/` sibling repo, fluent in its
`discord.py` client, the FEAT-015 two-LLM loop, the async MCP stdio client, and the FEAT-016 slash
commands.

## Goal

Route CrawDad's `/crawdad draft <topic>` and a new `/crawdad ask <question>` (research medium) to
the MCP `author` verb, and save kept AI output to `11-Other-Authors/ai-as-user/<slug>/` with
`representativeness: endorsed`, `voice_weight: 0.0`.

## Context

- **Parent epic:** #452
- **Predecessor issue(s):** #460 (real MCP verb).
- **SPEC section:** §10 (CrawDad wiring), §7 (`ai-as-user` save target + attribution).
- **Files involved:**
  - `crawdad/crawdad/` — slash-command handlers + MCP dispatch.
  - `crawdad/` tests.
- **Prior decisions:** `/crawdad ask` uses the research medium; `/crawdad draft` uses the draft path. Saving AI output is an explicit user action (e.g. a confirm), filed via the existing `save` path into `11-Other-Authors/ai-as-user/`. Reuse the FEAT-015 loop for trivial chat (unchanged).
- **State of the world:** CrawDad has six slash commands; `/crawdad draft` exists but composes inline rather than via the desk; there is no `/crawdad ask`.

## Output Format

A single PR containing:

- [ ] `/crawdad ask` + `/crawdad draft` routed to the MCP `author` verb.
- [ ] Kept AI output saved to `11-Other-Authors/ai-as-user/` with correct attribution.
- [ ] Tests: a research question returns a cited reply; a saved draft lands with `author=ai`, `representativeness=endorsed`, `voice_weight=0.0`.

## Examples

```
/crawdad ask What distinguishes F6 medicine from its toxic dose?
→ (cited, voiced answer; "save" reaction files it to 11-Other-Authors/ai-as-user/)
```

## Constraints

**Scope fence:** Do not implement allowlist/degradation here (ISSUE_04). Do not modify the desk
internals (EPIC_02).

**Anti-bypass (verbatim, non-negotiable):**

> No `noqa`, `# type: ignore`, `pylint: disable`, `eslint-disable`, or equivalent
> linter/type-checker silencers. Fix the root cause. The only exception is the documented 4-line
[...]

Issue #452, opened 2026-06-02
epic: MCP author verb + CrawDad surfacing (FEAT-041)


Geoffe-Ga:
## Epic Summary

Surface the Creek Writing Desk beyond the CLI: expose an `author` verb over the creek-tools MCP
surface, and wire CrawDad's slash commands (`/crawdad ask`, `/crawdad draft`) to call it so the
Discord side answers research questions and drafts via the same desk. AI-authored outputs the
owner keeps are saved back to `11-Other-Authors/ai-as-user/` with correct attribution.
Implements SPEC §4.3 (adapters) and §10 (MCP + CrawDad).

## Scope

**In scope:**
- New `author` MCP tool mirroring the `creek author` CLI args, returning draft + provenance + reflection verdict over stdio.
- CrawDad `/crawdad ask <question>` (research medium) and `/crawdad draft <topic>` routed to the MCP `author` verb.
- Saving kept AI output to `11-Other-Authors/ai-as-user/<slug>/` with `representativeness: endorsed`, `voice_weight: 0.0`.
- User + channel allowlist enforcement and graceful degradation when the desk / creek-tools is unreachable.

**Out of scope:**
- The desk internals (EPIC_02, #451).
- Mediums other than research/draft (EPIC_04, #453).

## Success Criteria

The epic is done when:

- [ ] An MCP client calls the `author` tool over stdio and receives a well-typed draft + provenance + verdict.
- [ ] `/crawdad ask` answers a research question about APTITUDE/Wavelength end-to-end in Discord.
- [ ] A kept AI draft is filed to `11-Other-Authors/ai-as-user/` with correct attribution and exits the desk's privacy gates.
- [ ] Non-allowlisted callers get no response; an unreachable desk yields a soft error, not a crash.
- [ ] All child issues closed; both `creek-tools` and `crawdad` quality gates green on `main`.

## Child Issues

- [ ] #456 — Skeleton: MCP `author` verb stub returning the output shape
- [ ] #460 — Core: wire the MCP `author` verb to the real desk
- [ ] #464 — Core: CrawDad `/crawdad ask` + `/crawdad draft` routing + ai-as-user save
- [ ] #468 — Edges: allowlist + graceful degradation

## Sequencing Notes

- **Blocked by:** EPIC_02 (#451, the desk must exist).
- **Parallel-safe:** EPIC_04 (#453).

## SPEC Reference

[...]

PR #495, opened 2026-06-02
feat(crawdad): allowlist + graceful degradation for author routes (#468)

### LINKED ISSUE #468: feat(crawdad): allowlist enforcement + graceful degradation for author routes
## Role

You are a senior Python engineer working in `crawdad/`, fluent in its user+channel allowlist and
its graceful-degradation behavior when creek-tools is unreachable.

## Goal

Ensure the new author routes obey the security and resilience contract: non-allowlisted callers get
no response; an unreachable desk / creek-tools subprocess yields a soft error reply, never a crash.

## Context

- **Parent epic:** #452
- **Predecessor issue(s):** #464 (routes exist).
- **SPEC section:** §10 (CrawDad), and CrawDad's existing security model (allowlist; `crawdad.yaml` trusted input; no exit on MCP failure).
- **Files involved:**
  - `crawdad/crawdad/` — apply allowlist + degradation to `/crawdad ask` and `/crawdad draft`.
  - `crawdad/` tests.
- **Prior decisions:** Allowlist is enforced before any desk call. On MCP/desk failure, reply with a soft error and stay alive. `crawdad.yaml` is trusted input.
- **State of the world:** Allowlist + degradation exist for current commands; the new routes must inherit them.

## Output Format

A single PR containing:

- [ ] Allowlist enforced on the author routes (non-allowlisted → no response).
- [ ] Soft-error reply when the desk/creek-tools is unreachable; process stays up.
- [ ] Tests: non-allowlisted user gets nothing; simulated MCP failure yields a soft error, not an exception.

## Examples

```python
# Non-allowlisted caller:
assert handle_ask(non_allowlisted_ctx, "F6?") is None     # no response

# Desk unreachable:
reply = handle_ask(allowlisted_ctx_with_dead_mcp, "F6?")
assert "temporarily unavailable" in reply.lower()         # soft error, no crash
```

## Constraints

**Scope fence:** No new features — only security + resilience for the author routes. Do not change
the desk or MCP verb.

> linter/type-checker silencers. Fix the root cause. The only exception is the documented 4-line
> escape hatch (third-party library bug / language-version compatibility / benchmarked
> performance necessity / generated code) — and it must include the reason, a reference URL, an
> alternative considered, and a review date. See the `max-quality-no-shortcuts` skill.

**Tracer-code invariant:** CrawDad remains demoable and crash-resistant; security parity with
existing commands.

- [ ] All new and existing tests pass (CrawDad's quality gate).
- [ ] Pre-commit clean for the `crawdad/` repo.
- [ ] Negative-path tests (allowlist + degradation) included.
- [ ] PR body includes `Refs #452` and `Closes` this issue.
- [ ] Latest Claude reviewer `Verdict:` on HEAD is `LGTM`.

## Labels

`spec-decomposition`, `edges`, `crawdad`

Geoffe-Ga:
## Summary

Makes the new `/crawdad ask` and `/crawdad draft` routes (#464) obey CrawDad's security + resilience contract.

- **Allowlist (the real gap):** slash-command Discord callbacks bypass the free-text `_passes_allowlist` gate in `bot.py`, so the author callbacks were reachable by non-allowlisted users. New `_interaction_allowed(interaction, config)` — delegating to the single `CrawDadConfig.is_allowed` AND-gate — is called at the top of the `ask`/`draft` callbacks **before `defer()`**: a non-allowlisted user/channel is **silently ignored** (no defer, no reply), matching `bot._passes_allowlist`. `config` is threaded through `register()` and the `CrawDadClient` call site.
- **Graceful degradation:** both author handlers now route through a shared `_run_and_reply` helper that catches `MCPUnavailableError` and replies with the documented soft error (`"creek-tools is unreachable; try again in a moment."` — the same wording as `loop._MCP_UNAVAILABLE_REPLY`) instead of letting a stack trace reach Discord. The agent loop already converts MCP failures to a soft string; this is defense-in-depth at the route boundary, mirroring `_reply_workflow_run`.

### Scope note (follow-up)
The issue assumed existing commands already enforce the allowlist at the slash layer — they don't: **all** slash callbacks (reflect/checkin/surface/save/register/workflow) share this pre-existing callback-level gap. Per the scope fence ("only … the author routes"), this PR gates **only** `ask`+`draft`; I'll file a follow-up to extend `_interaction_allowed` to the rest. The `register(config=...)` plumbing to do so is now in place.

## Test plan
`cd crawdad && ./scripts/check-all.sh` → **7/7 green** (426 passed, slash_commands.py 99.03%). New negative-path tests:
- `_interaction_allowed` parametrized by user AND channel (the four AND-gate combinations).
- Author callbacks **deny non-allowlisted silently** (assert `defer` not called and nothing sent) and **allow allowlisted** (defer + one reply).
- `handle_ask`/`handle_draft` **degrade softly** when the runner raises `MCPUnavailableError` (one soft reply containing "unreachable", no exception propagates) and **relay** a soft-error string returned by the runner.

Closes #468
Refs #452

### DIFF
--- a/crawdad/crawdad/bot.py
+++ b/crawdad/crawdad/bot.py
@@ -810,6 +810,7 @@ class CrawDadClient(discord.Client):
             # ``test_register_wires_every_command_onto_tree`` test.
             register_slash_commands(
                 cast("Any", self.tree),
+                config=config,
                 loop_runner=loop_runner,
                 register_switcher=register_switcher,
                 workflow_lister=workflow_lister,
--- a/crawdad/crawdad/slash_commands.py
+++ b/crawdad/crawdad/slash_commands.py
@@ -41,8 +41,8 @@ if TYPE_CHECKING:
 
 _LOGGER = logging.getLogger("crawdad.slash_commands")
 
-# Issue #468: the documented soft-error shown when the desk / creek-tools
-# MCP surface is unreachable. Mirrors ``crawdad.loop._MCP_UNAVAILABLE_REPLY``
+# The documented soft-error shown when the desk / creek-tools MCP surface is
+# unreachable. Mirrors ``crawdad.loop._MCP_UNAVAILABLE_REPLY``
 # so the author routes degrade with the same wording as the agent loop.
 _MCP_UNAVAILABLE_REPLY = "creek-tools is unreachable; try again in a moment."
 
@@ -179,8 +179,8 @@ async def _run_and_reply(
 ) -> None:
     """Drive the loop runner and reply, degrading softly on MCP failure.
 
-    Issue #468: the author routes (`ask`/`draft`) own a resilience
-    contract — if the injected ``loop_runner`` ever lets an
+    The author routes (`ask`/`draft`) own a resilience contract — if the
+    injected ``loop_runner`` ever lets an
     :class:`~crawdad.mcp_client.MCPUnavailableError` escape (the loop
     normally converts it to a soft-reply string, but defense-in-depth
     keeps Discord from ever seeing a stack trace), reply with the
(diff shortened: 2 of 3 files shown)


Geoffe-Ga:
Addressed the verdict: stripped all 7 `Issue #468` references from comments and docstrings (slash_commands.py + the test section header), keeping the semantic *why* in each. check-all 7/7 green.

On the two low-priority notes: the duplicate `_MCP_UNAVAILABLE_REPLY` is a deliberate mirror of `loop._MCP_UNAVAILABLE_REPLY` (importing it would couple slash_commands → loop); and a `None` `channel_id` already fails closed via `is_allowed` (denied), which is the safe default. Happy to fold either into a follow-up if you'd prefer.

Issue #487, opened 2026-06-02
refactor(author): de-dupe by_title collisions + clearer walk loop var in agents.py


Geoffe-Ga:
Non-blocking cleanups the reviewer flagged on #463 (PR #485, both rounds LGTM). Tracked so they don't get lost; not auto-picked by ralph (no \`spec-decomposition\` label).

## 1. \`by_title\` collision in \`_build_link_graph\`
\`creek/author/agents.py\` builds \`by_title = {fragment.title: fragment.id for ...}\`. Two fragments sharing a title silently collapse to one id, so a \`[[wikilink]]\` resolved by title could point at the wrong fragment. Realistic Creek vaults rarely collide titles, hence non-blocking — but a defensive fix (prefer id resolution, and on title collision either skip or keep a list) would be more correct.

## 2. \`current\` loop-var naming in \`_bounded_walk\`
The BFS loop variable \`current\` actually holds the *depth index* (1..depth), not the current node. Rename to \`depth_index\`/\`hop\` for readability.

Both are readability/robustness only — no correctness regression for realistic vaults (per the #485 verdict).

Geoffe-Ga:
The `by_title` duplicate-title item here is now done in #486 (PR #511): `_build_link_graph`'s last-wins behavior is documented and pinned by a test. Remaining scope for this issue is just the cosmetic `_bounded_walk` loop-variable rename (the `current` name actually holds the depth index).

Issue #463, opened 2026-06-02
feat(author): real Graph + Retrieval specialist agents


Geoffe-Ga:
## Role

You are a senior Python engineer working in `creek/author/agents.py`, fluent in the compiled-layer
query path, the embeddings/link modules, and Obsidian backlink structure.

## Goal

Replace the **Graph** and **Retrieval** specialist stubs with real implementations: the Graph agent
navigates the compiled layer + Obsidian backlinks + frequency/wavelength indexes (bounded
breadth/depth walk); the Retrieval agent does semantic retrieval over `01-Fragments/`, `09-Reference/`,
and `11-Other-Authors/`. Both return structured, provenance-tracked `EvidenceBundle`s.

## Context

- **Parent epic:** #451
- **Predecessor issue(s):** #459 (medium contract drives specialist weighting).
- **SPEC section:** §4.1 (Graph/Retrieval rows), §6 (knowledge-graph navigation), §13 open question #4 (bounds).
- **Files involved:**
  - `creek-tools/creek/author/agents.py` — real Graph + Retrieval agents.
  - reuse `creek/link/embeddings.py`, the `creek query` path, backlink parsing.
  - `creek-tools/tests/` — agent tests on a fixture vault.
- **Prior decisions:** Open question #4 default — backlink walk **breadth 25 / depth 2**, relevance-pruned; make these config-bounded, not hard-coded. Evidence is structured (claims + `source_fragments`), never prose. `11-Other-Authors/` evidence carries its author attribution for downstream citation.
- **State of the world:** Stubs from ISSUE_01 return mock evidence. Embeddings + compiled-layer query already exist and should be reused, not reinvented.

## Output Format

A single PR containing:

- [ ] Real Graph agent: seed → bounded backlink walk → structured evidence with provenance.
- [ ] Real Retrieval agent: semantic retrieval across raw + reference + `11-Other-Authors/`.
- [ ] Tests: deterministic evidence bundles on the fixture; walk respects breadth/depth bounds; other-author evidence retains attribution.

## Examples

```python
ev = graph_agent.gather(seed="F6-Pluralism", contract=research, vault=fixture)
assert ev.claims and all(c.source_fragments for c in ev.claims)
assert ev.walk_stats.max_depth <= contract.graph_depth_bound   # default 2
```

## Constraints

**Scope fence:** Do not implement the Ontology agent (ISSUE_04), synthesis/voicing (ISSUE_05), or
[...]

Issue #451, opened 2026-06-02
epic: Creek Writing Desk — research medium e2e (FEAT-041)


Geoffe-Ga:
## Epic Summary

Build the multi-agent **Creek Writing Desk** and deliver the **research / encyclopedic medium**
end-to-end: `creek author --medium research --query "..."` produces a cited, voiced,
reflection-gated answer about the APTITUDE frequency framework and the Archetypal Wavelength.
Implemented on the Anthropic SDK managed-agents pattern — a Conductor (supervisor) routing to
Graph / Retrieval / Ontology specialist agents, a Voice agent, and a Reflection node
(LLM-as-judge) with bounded retries and human escalation. Implements SPEC §4, §5 (research
contract), §6, §8, §10 (CLI).

## Scope

**In scope:**
- `creek/author/` package: Conductor, specialist agents, Voice agent, Reflection node.
- `creek author` CLI verb with `--medium research`, `--query`, `--dry-run`, `--max-rounds`, `--include-tier`.
- `MediumContract` model + the `research` medium contract (deployed to `00-Creek-Meta/Skills/mediums/`).
- Graph-native retrieval (backlink walk, frequency/wavelength index seeds) returning structured, provenance-tracked evidence.
- Conductor synthesis (claims → fragment IDs), Voice agent (voice-skill tree), Reflection node (research rubric) with `max_author_rounds` retries + ESCALATE.
- Anthropic SDK wiring with prompt caching + model tiering.

**Out of scope:**
- MCP verb / CrawDad routing (EPIC_03, #452).
- Mediums other than `research` (EPIC_04, #453).
- Creating `11-Other-Authors/` or the attribution model (EPIC_01, #450 — consumed here, not built here).

## Success Criteria

The epic is done when:

- [ ] `creek author --medium research --query "..."` produces a cited answer on a fixture vault, every substantive claim carrying `provenance` to real fragment IDs / the ontology spec.
- [ ] `--dry-run` prints the plan + evidence bundle without synthesizing.
- [ ] A deliberately bad draft triggers REVISE then improves; exhausted retries ESCALATE rather than ship.
- [ ] Taxonomy in output is canonical (no aliases; INC-019 respected).
- [ ] Prompt-cache hits asserted on the static context.
- [ ] All child issues closed; `./scripts/check-all.sh` green on `main`.

## Child Issues

- [ ] #455 — Skeleton: `creek author` + `creek/author/` wired end-to-end with stubs
[...]

PR #511, opened 2026-06-02
perf(author): reuse EmbeddingLinker + embeddings cache in Retrieval agent (#486)

### LINKED ISSUE #486: perf(author): reuse EmbeddingLinker + embeddings cache in Retrieval agent
## Context

Follow-up from PR #485 (#463 — real Graph + Retrieval specialists). The real agents are correct and bounded, but the Retrieval path has two known performance/robustness gaps deferred to keep #463 focused:

## Items

1. **EmbeddingLinker / model reuse across `gather()` calls.** `RetrievalSpecialist.gather` creates a fresh `EmbeddingLinker` each call (`_rank_fragments`), so the sentence-transformer model is re-loaded once per author run. For interactive desk use (repeated `creek author` / MCP `author` calls in a session) this is noticeable latency. Reuse a process-cached linker (or inject one) **and** read fragment vectors from the persisted embeddings cache (`load_cache` / `embeddings_cache_path`) instead of re-embedding the whole corpus per query. Mind test isolation — a module-level cache must not leak a loaded (mock) model across tests.

2. **`by_title` duplicate-title override.** `_build_link_graph` maps `title -> id`; duplicate titles silently overwrite. If the vault treats duplicate titles as an error (or we want last/first-wins documented), make that explicit.

## Acceptance

Retrieval reuses the model + the embeddings cache across calls (measured improvement on a warm process), and the title-collision behavior is explicit + tested.

Refs #451

Geoffe-Ga:
## Summary

Follow-up to #463. `RetrievalSpecialist.gather` re-loaded the sentence-transformer model and re-embedded the whole corpus on every call; both are now amortised, with results **byte-for-byte identical** to before (the cache is a pure optimisation).

- **Linker reuse (instance-level, no module global):** the `EmbeddingLinker` is held lazily on `self._linker`, so a long-lived specialist reused across `gather()` calls loads the model once — while a fresh `RetrievalSpecialist()` per test starts cold, so the conftest model mock **cannot leak across tests**. `__init__(*, linker=None)` lets a caller inject/share one.
- **Embeddings cache for fragment vectors:** `_fragment_vector` draws from `load_cache(embeddings_cache_path(vault))`, trusting a cached vector only when the id is present AND its `content_hash` equals `content_hash_for_text(fragment_embedding_text(fragment))` (the cache writer's freshness key); a miss or stale hash falls back to a live embed. The query is embedded once. Absent cache file → exactly the prior behaviour.
- **`by_title` made explicit:** `_build_link_graph` documents duplicate-title handling as **last-wins** (ids always preferred over titles for wikilink resolution), pinned by a test. **Supersedes the `by_title` item of #487.**

## Test plan
`cd creek-tools && ./scripts/check-all.sh` → **12/12 green** (4997 tests). New tests in `tests/test_real_agents.py`:
- Model loaded **once** across two `gather()` calls on one instance.
- A warm parquet cache → **zero per-fragment embeds** (only the query embedded).
- **Cold-vs-warm bundle parity** (cache changes nothing observable).
- A deliberately-wrong `content_hash` forces a **live recompute** (freshness validation).
- Graceful degradation (`EmbeddingModelUnavailableError` → empty bundle) preserved.
- `by_title` last-wins edge.

Closes #486
Refs #451

### DIFF
--- a/creek-tools/creek/author/agents.py
+++ b/creek-tools/creek/author/agents.py
@@ -189,6 +189,8 @@ class RetrievalSpecialist:
                 first :meth:`gather` and cached on the instance.
         """
         self._linker = linker
+        self._cache: dict[str, CachedEmbedding] | None = None
+        self._cache_vault: Path | None = None
 
     def _get_linker(self, config: CreekConfig) -> EmbeddingLinker:
         """Return the instance's linker, creating and caching it on first use.
@@ -203,6 +205,33 @@ class RetrievalSpecialist:
             self._linker = EmbeddingLinker(config.embeddings)
         return self._linker
 
+    def _get_cache(
+        self,
+        linker: EmbeddingLinker,
+        vault: Path,
+    ) -> dict[str, CachedEmbedding]:
+        """Return the persisted embeddings cache, read once per instance + vault.
+
+        The parquet is loaded on first use for a given vault and reused across
+        later :meth:`gather` calls so the desk does not re-read it every call.
+        Staleness is harmless: ``_fragment_vector`` validates each entry's
+        ``content_hash`` against the fragment's current text, so a fragment
+        re-embedded into the parquet after this instance cached the dict simply
+        falls back to a live embed — never a wrong vector. A different *vault*
+        reloads.
+
+        Args:
+            linker: The reused linker whose ``load_cache`` reads the parquet.
+            vault: The vault whose embeddings cache is read.
+
+        Returns:
+            The fragment-id keyed embedding cache (empty when absent).
+        """
+        if self._cache is None or self._cache_vault != vault:
+            self._cache = linker.load_cache(embeddings_cache_path(vault))
+            self._cache_vault = vault
+        return self._cache
+
     def gather(self, query: str, vault: Path) -> EvidenceBundle:
         """Return the top-``retrieval_top_k`` fragments most relevant to *query*.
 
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
Both blocking items addressed:

1. **Cache re-read per call:** the parquet is now loaded once per instance + vault (`_get_cache`) and reused across `gather()` calls, matching the linker amortisation. Correctness is preserved even if the parquet changes afterward — `_fragment_vector`'s `content_hash` freshness check falls a stale/absent entry back to a live embed (never a wrong vector); a different vault reloads. New test `test_retrieval_loads_cache_once_across_gather_calls` asserts `load_cache` fires exactly once across two gathers on one instance.
2. **Mock fixture dependency:** `_seed_embeddings_cache` now documents that its vectors come from the autouse `mock_sentence_transformer` (deterministic mock, no real model) — so these are unit tests and correctly need no `@pytest.mark.integration` marker.

check-all 12/12 green.

Geoffe-Ga:
Fixed the brittle assertion: `test_retrieval_stale_cache_entry_is_recomputed` now asserts the exact embedded list `["alpha", fragment_embedding_text(by_id["frag-a"])]` (the query plus the one stale fragment, re-embedded live) — computed via the same helper, so it survives a `fragment_embedding_text` shape change and still proves the stale cache entry was not served. check-all 12/12 green.

Issue #471, opened 2026-06-02
feat(author): Conductor synthesis + Voice agent rendering


Geoffe-Ga:
## Role

You are a senior Python engineer working in `creek/author/conductor.py` and `creek/author/voice.py`,
fluent in the FEAT-040 voice-skill tree.

## Goal

Make the Conductor's **synthesis** real (merge the three specialists' evidence into a structured
draft where every substantive claim maps to `source_fragments`), and make the **Voice agent** real
(render the synthesized draft in the vault owner's voice using the voice-skill stack, medium- and
phase-aware).

## Context

- **Parent epic:** #451
- **Predecessor issue(s):** #467 (all three specialists are real).
- **SPEC section:** §4.2 (synthesis + Voice agent), §5 (research contract structure).
- **Files involved:**
  - `creek-tools/creek/author/conductor.py` — evidence merge → grounded draft.
  - `creek-tools/creek/author/voice.py` — load `<vault>/creek-skills/` voice stack and render.
  - `creek-tools/tests/` — synthesis + voice tests.
- **Prior decisions:** Synthesis stays grounded — no claim without provenance. Voice rendering reuses the existing voice stack; it does NOT pull from `11-Other-Authors/` for voice (only ideas, already attributed). For research medium, voicing is light (clarity over flourish per the contract).
- **State of the world:** Conductor synthesis + Voice are still stubs; the three specialists now return real evidence.

## Output Format

A single PR containing:

- [ ] Real synthesis: structured draft with per-claim provenance, honoring the research contract structure.
- [ ] Real Voice agent rendering via the voice-skill tree.
- [ ] Tests: every claim in the draft carries provenance to a real fragment ID; voice rendering applies on the fixture; a claim lacking evidence is dropped or flagged, never fabricated.

## Examples

```python
draft = run_author(medium="research", query="What is F6 Pluralism?", vault=fixture)
assert all(claim.source_fragments for claim in draft.claims)   # grounded
assert draft.rendered_text                                     # voiced
```

## Constraints

**Scope fence:** Do not implement the Reflection node / retries (ISSUE_06) or prompt-caching polish
(ISSUE_07). Voice must not train on or quote `11-Other-Authors/` as the owner's own words.

**Anti-bypass (verbatim, non-negotiable):**

> No `noqa`, `# type: ignore`, `pylint: disable`, `eslint-disable`, or equivalent
> linter/type-checker silencers. Fix the root cause. The only exception is the documented 4-line
[...]

PR #514, opened 2026-06-02
feat(author): wire voice-fidelity into the reflection node (#506)

### LINKED ISSUE #506: feat(author): wire voice-fidelity into the reflection node's primary path
Follow-up from #473 (PR #505 review). The reflection node's `check_voice_fidelity` (reuses the FEAT-040.x AI-style `scan`) is fully implemented and unit-tested, but the Conductor calls `reflection.review(body, evidence, rubric, contract=, vault=)` WITHOUT `fingerprint=`/`ai_style_config=`, so the check always skips through the primary path.

Wiring it requires the Conductor to obtain:
- the owner's `VoiceFingerprint` (a generated artifact — `creek report --type fingerprint`; decide whether to load a persisted fingerprint from the vault or compute on demand / cache), and
- the `AIStyleConfig` from `CreekConfig`.
then thread both into `review(...)`. Add an end-to-end test that a voice-divergent draft yields a `voice_fidelity` finding through `run()` (not just the unit test on `check_voice_fidelity`).

Until then `check_voice_fidelity` is a tested-but-dormant check (breadcrumb left at the `review` call site in `conductor.run`). Given `spec-decomposition` so ralph picks it up.

Geoffe-Ga:
## Summary

Follow-up to #471/#473 — the **last** writing-desk follow-up. The reflection node's `check_voice_fidelity` was implemented + unit-tested but **dormant**: the Conductor never passed `fingerprint=`/`ai_style_config=` into `review`, so the check always skipped.

- **New `Conductor._load_voice_inputs(vault)`** loads the vault's `AIStyleConfig` (`config.ai_style`) and the owner's persisted `VoiceFingerprint` (`load_fingerprint`, reading `00-Creek-Meta/voice-fingerprint.json`). Config loading is **best-effort** (any error → `(None, None)`, so the desk never crashes on a thin/offline vault), and an **empty fingerprint** (`fragment_count == 0`, none built) is returned as `None` so voice-fidelity stays **dormant exactly as before** on a vault without a built fingerprint.
- **`run()`** computes the voice inputs once before the voice/reflect loop and threads them into every `reflection.review(...)` call. Behaviour is unchanged when no fingerprint exists; otherwise the check now fires.

## Test plan
`cd creek-tools && ./scripts/check-all.sh` → **12/12 green**. New `tests/test_voice_fidelity_wiring.py`:
- **End-to-end (DoD):** a run with a **real persisted fingerprint + the real AI-style scanner** surfaces a `voice_fidelity` finding on `draft.findings` — proving `review` now receives a non-`None` fingerprint+config and the check fires through `run()`.
- **Dormant when unbuilt:** a vault with no `voice-fingerprint.json` produces no voice-fidelity finding and doesn't crash.
- **`_load_voice_inputs`** returns the fingerprint only when a non-empty one is persisted.

No existing tests modified.

Closes #506
Refs #451

### DIFF
--- a/creek-tools/creek/author/conductor.py
+++ b/creek-tools/creek/author/conductor.py
@@ -36,6 +36,8 @@ if TYPE_CHECKING:
     from pathlib import Path
 
     from creek.author.client import AuthorLLMClient
+    from creek.config import AIStyleConfig
+    from creek.generate.ai_style.model import VoiceFingerprint
     from creek.models import MediumContract
 
 logger = logging.getLogger(__name__)
@@ -86,6 +88,8 @@ class Reflector(Protocol):
         *,
         contract: MediumContract | None = None,
         vault: Path | None = None,
+        fingerprint: VoiceFingerprint | None = None,
+        ai_style_config: AIStyleConfig | None = None,
     ) -> ReflectionResult:
         """Return a structured result for *body* given *evidence* and *rubric*."""
         ...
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
Fixed the blocking crash: `_load_voice_inputs` now catches `pydantic.ValidationError` (not a `ValueError` subclass in v2) alongside `OSError`/`ValueError`, so a malformed `creek_config.yaml` degrades voice-fidelity to dormant `(None, None)` instead of crashing the conductor — restoring the never-crashes contract. Added `test_load_voice_inputs_degrades_on_malformed_config` (an out-of-range `max_author_rounds` → `(None, None)`, no raise), pinning `CREEK_CONFIG` so a leaked env var can't shadow it.

On the optional notes: `last_usage` is actually live (read at `conductor.run` → `AuthoredDraft.usage`), so I left it. check-all 12/12 green.

PR #523, opened 2026-06-03
fix(ai-style): detect bare "not X but Y" antithesis tic

### LINKED ISSUE #516: fix(ai-style): detect bare "not X, but Y" antithesis tic
## Summary
The `negative_parallelism` voice tell does not catch the bare `not X, but Y`
antithesis construction, which is the dominant AI tell saturating generated
drafts. Detect it so the voice-fidelity scan can flag (and the rewrite guard
can sand) it.

## Reproduction
`07-Voice/Drafts/2026-06-03-christ.md` is dense with the pattern, none of which
the current regex flags:
- *"Not metaphorically. Functionally."*
- *"The parables weren't instruction manuals. They were field reports…"*
- *"Not through bliss, necessarily, but through outrage…"*
- *"The work isn't to save everyone. The work is to be findable…"*
- *"…doesn't make you easier to manipulate—it makes you ungovernable."*

## Root cause
`creek/generate/ai_style/features.py:267-272` (`NEGATIVE_PARALLELISM_RE`) matches
only: `not only X but [also] Y`; contracted `it's not X, it's Y` (requires the
comma-then-`it's`); and `no X, no Y, just/but`. It misses bare `not X, but Y`,
`not X but Y`, and uncontracted `it is not … it is …`. The tell is also
vault-relative + soft: margin `_PHRASE_MARGIN=1.5`/1k words, `handling=surface`,
MID finding only (`discourse.py:42,50-64`).

## Fix strategy
1. Broaden `NEGATIVE_PARALLELISM_RE` to also match bare `not X, but Y`,
   `not X but Y`, uncontracted `it is not … it is …`, and
   `it's not about … it's about …`. The existing
   `negative_parallelism_density` extractor (`features.py:324-326`) and the
   `FINGERPRINT_FEATURES` registry entry already wire it through.
2. Consider adding/raising it toward a generic-prior tell (like
   `knowledge_cutoff`, `margin=0`) so a heavy antithesis cluster flags even when
   the owner's own baseline antithesis rate is nonzero — but keep it MID, since
   the owner *does* use deliberate antithesis occasionally; we want to catch
   *saturation*, not the occasional rhetorical turn.

## Acceptance criteria
- [ ] New failing tests first: each reproduction sentence above is matched by
      the broadened regex; the owner's occasional single antithesis in a real
      fragment does NOT over-flag (calibrate margin with a fixture).
- [ ] `negative_parallelism_density` rises measurably on the christ-draft body
      vs. a clean control (pin with a test).
- [ ] No regression in existing `tests/test_ai_style*` discourse-tell tests.
- [ ] `./scripts/check-all.sh` exit 0.

Refs #417

Geoffe-Ga:
## Summary
- Broaden `NEGATIVE_PARALLELISM_RE` to detect the dominant AI antithesis tell the old pattern missed: bare parallel-prepositional `not <prep> X but <prep> Y`, uncontracted `it is not … it is …` / `it is not about … it is about …`, cross-sentence subject restatement (`X weren't Y. They were Z`), and em-dash antithesis (`it doesn't … — it …`).
- Retain every existing match (`not only … but [also]`, contracted `it's not X, it's Y`, `no X, no Y, just/but`).
- Keep the tell MID/`surface` with the `_PHRASE_MARGIN=1.5`/1k-word gate — no escalation to a hard gate; saturation (a cluster) fires, a single rhetorical turn does not.

## Calibration against the owner's deliberate antithesis
The bare `not … but …` alternation requires a **shared preposition** (`not through X but through Y`), so a parallel rhetorical antithesis matches while a casual concessive ("Not my best work but it is honest") does not. This keeps the calibration harness (`test_calibration.py`, which uses that exact concessive as a clean-user fixture) green. Because the density feature is vault-relative and surface-only, one deliberate antithesis in a clean paragraph stays under the 1.5/1k-word margin (asserted) while a saturated paragraph clears it (asserted) — we catch saturation, not the occasional turn. The minimal change (broadened regex + the existing margin) was sufficient; no `antithesis_rate` generic prior was needed.

## Test plan
- New `tests/generate/ai_style/test_features.py::TestNegativeParallelismRegex` — each issue #516 reproduction sentence now matches; legacy constructions still match; an ordinary sentence and a bare negation do not.
- New `TestNegativeParallelismDensity` — saturated paragraph scores measurably higher than a clean control; clean control is zero.
- New `tests/generate/ai_style/test_discourse.py` calibration tests — one antithesis below the surface margin does **not** fire; a saturated cluster **does** fire for a plain user.
- No regression in existing `tests/test_ai_style*` / `tests/generate/ai_style/*` discourse-tell tests.
- `cd creek-tools && ./scripts/check-all.sh` → exit 0 (coverage 93.78%, all 12 gates pass, 5047 tests).

Closes #516
Refs #417

### DIFF
--- a/creek-tools/creek/generate/ai_style/features.py
+++ b/creek-tools/creek/generate/ai_style/features.py
@@ -277,6 +277,11 @@ def vague_attribution_density(text: str) -> float:
 #   4. Cross-sentence subject restatement ``X isn't/weren't Y. <subject>
 #      is/are/was/were/makes Z`` — the negated claim then its affirming echo,
 #      e.g. "The parables weren't instruction manuals. They were field reports."
+#      The echoed subject must be a *genuine* restatement: either a pronoun
+#      (they/it/he/she/we) or the SAME noun repeated (``the <noun>`` …
+#      ``the <noun>`` via the ``xnoun`` backreference). A fresh ``The <word>``
+#      with a different noun is not antithesis and must not match, e.g.
+#      "The algorithm doesn't converge. The teacher is patient."
 #   5. Em-dash antithesis ``it doesn't … — it …``.
 #   6. ``no X, no Y, just/but``.
 # Calibration: every alternation requires a contrasting second clause (a bare
@@ -288,8 +293,9 @@ NEGATIVE_PARALLELISM_RE = re.compile(
     r"[^.!?\n]{1,80}?,?\s*\bbut (?P=prep)\b"
     r"|\bit'?s not (?:just |merely |only )?[^.!?\n]{1,60}?,\s*it'?s\b"
     r"|\bit is not (?:just |merely |only |about )?[^.!?\n]{1,60}?,?\s*it is\b"
-    r"|\b(?:is|are|was|were|do|does)n'?t\b[^.!?\n]{1,80}?[.!?]\s+"
-    r"(?:they|it|he|she|we|the [a-z]+)\b[^.!?\n]{0,12}?"
+    r"|\b(?:the (?P<xnoun>[a-z]+) )?"
+    r"(?:is|are|was|were|do|does)n'?t\b[^.!?\n]{1,80}?[.!?]\s+"
+    r"(?:(?:they|it|he|she|we)|the (?P=xnoun))\b[^.!?\n]{0,12}?"
     r"\b(?:is|are|was|were|do|does|make|makes)\b"
     r"|\b(?:is|are|was|were|do|does)n'?t\b[^.!?\n]{1,60}?—"
     r"\s*it\b[^.!?\n]{0,40}?\b(?:is|makes?|does)\b"
--- a/creek-tools/creek/generate/ai_style/features.py
+++ b/creek-tools/creek/generate/ai_style/features.py
@@ -264,9 +264,35 @@ def vague_attribution_density(text: str) -> float:
 # Each pattern matches the offending phrase as group 0 (no lead-in capture)
 # unless noted, so a plain ``finditer`` span is the thing to surface.
 
+# Antithesis / negative-parallelism. The alternations, in order:
+#   1. ``not only X but [also] Y``.
+#   2. Bare ``not <prep> X[,] but <prep> Y`` — antithesis built from parallel
+#      prepositional phrases, the dominant AI tell from issue #516, e.g.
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
Addressed all four review findings (commit `c7bb436`):

1. **🟡 Tightened the cross-sentence arm (precision).** Pattern 4's subject clause no longer accepts any new `The <word>`. The negated first clause now optionally captures `the (?P<xnoun>[a-z]+)`, and the affirming echo must be either a pronoun (`they|it|he|she|we`) **or** the same noun via backreference. New tightened sub-pattern:

   ```
   |\b(?:the (?P<xnoun>[a-z]+) )?
   (?:is|are|was|were|do|does)n'?t\b[^.!?\n]{1,80}?[.!?]\s+
   (?:(?:they|it|he|she|we)|the (?P=xnoun))\b[^.!?\n]{0,12}?
   \b(?:is|are|was|were|do|does|make|makes)\b
   ```

   - `"The algorithm doesn't converge. The teacher is patient."` -> **None** (was a false match).
   - `"The work isn't to save everyone. The work is to be findable."` -> still matches (repeated noun).
   - `"The parables weren't instruction manuals. They were field reports."` -> still matches (pronoun echo).

   Added a test-first failing case for the algorithm/teacher pair (watched it fail on the loose regex, pass after tightening).

2. **🟢 Relabelled the legacy case.** `"It's not about the money, it's about the principle."` moved from `_NEW` to `_LEGACY` (the original regex's `it's not X, it's Y` arm already matched it). Still matches.

3. **🟢 Added the calibration false-positive test.** `NEGATIVE_PARALLELISM_RE.search("Not my best work but it is honest") is None` — pins the casual-concessive boundary. Passes (the shared-preposition design keeps it None; the pattern-4 tightening doesn't affect it).

4. **🟢 Density floor.** `test_density_higher_on_saturated_than_clean` now also asserts `saturated > 30.0`, with a comment noting the saturated fixture scores ~76.9 per 1000 words (5 matches / 65 words), leaving generous headroom while catching a regression that loses several matches.

`cd creek-tools && ./scripts/check-all.sh` -> exit 0 (`5049 passed, 20 deselected`; coverage 93.78%; all 12 quality checks passed). Thresholds held.

Issue #538, opened 2026-06-03
fix(crawdad): skill_loader misses voice-core at meta/voice-core.SKILL.md (only checks voice-core/SKILL.md)


Geoffe-Ga:
## Summary

CrawDad's `skill_loader` looks for the voice-core skill **only** at `creek-skills/voice-core/SKILL.md`, but the canonical generator (`creek skills generate`) writes it to `creek-skills/meta/voice-core.SKILL.md`. As a result, on a vault built by the official generator, CrawDad's conversational Sonnet composer silently runs **without** the "always loaded, load-bearing" voice-core layer — even though the file exists — so plain-chat replies (`reflect`/`checkin`/`surface`/free-text) are less voice-faithful than they should be.

## Evidence

- `creek skills generate` emits the voice-core to `meta/voice-core.SKILL.md` (confirmed on a generated demo vault: `creek-skills/meta/voice-core.SKILL.md`, ~1.8 KB; there is no `creek-skills/voice-core/` dir).
- `crawdad/crawdad/skill_loader.py:123` hardcodes the single location:
  ```python
  voice_core = root / "voice-core" / "SKILL.md"
  _append_if_present(collected, voice_core, name="voice-core")
  ```
  `_append_if_present` skips silently when the file is absent, so the miss is invisible.
- The creek-tools side already solves this. `creek/author/skills.py` (the writing desk's voice agent) resolves voice-core against **both** locations, in order:
  ```python
  _VOICE_CORE_CANDIDATES = (
      ("voice-core", "SKILL.md"),
      ("meta", "voice-core.SKILL.md"),
  )
  ```
  with a docstring noting: "The native generator writes the voice-core to `creek-skills/meta/voice-core.SKILL.md`; this loader also accepts the crawdad-style `creek-skills/voice-core/SKILL.md` so a tree authored either way resolves."

So `creek.author` (and therefore CrawDad's `/crawdad draft` and `/ask`, which route through it) finds the voice-core; only CrawDad's own conversational `skill_loader` misses it.

## Fix direction

Mirror `creek/author/skills.py`'s two-candidate resolution in `crawdad/crawdad/skill_loader.py`: check `voice-core/SKILL.md` first, then fall back to `meta/voice-core.SKILL.md`, and append the first that exists. Keep the "always loaded, first in the stack" ordering. Add a test asserting that a tree with only `meta/voice-core.SKILL.md` (the generator's layout) still loads voice-core into the stack.

## Severity

[...]

PR #546, opened 2026-06-05
feat(skills): Add crawdad-launch skill to start the bot for any vault

Geoffe-Ga:
## Summary

Adds a `crawdad-launch` Claude Code skill that brings the CrawDad Discord bot
online against a **user-chosen** Creek vault. The skill always asks for the
vault directory first, then runs a preflight script that fixes every known
launch failure mode before starting the bot in the background.

## Why

Launching CrawDad by hand means remembering a pile of non-obvious gotchas
(absolute `vault_path`, `State/latest.md`, the #538 voice-core mirroring, env
secrets, non-empty allowlists). This skill encodes all of them so the bot comes
up correctly against any vault, first try.

## Changes

- `.claude/skills/crawdad-launch/SKILL.md` — triggers on "launch CrawDad / start
  the bot / run CrawDad"; mandates asking for the vault, documents the
  background-launch + restart-to-update model.
- `.claude/skills/crawdad-launch/launch.sh` — preflight + config generator:
  - requires `DISCORD_BOT_TOKEN` + `ANTHROPIC_API_KEY`
  - rewrites a relative `vault_path` to absolute (else MCP tools read the wrong dir)
  - generates `State/latest.md` via `creek state` if missing
  - mirrors `meta/voice-core.SKILL.md` -> `voice-core/SKILL.md` (#538)
  - preserves Discord allowlists, writes `crawdad/crawdad.yaml`
  - derives repo / creek-tools / crawdad paths from its own location (portable)

## Test Plan

- [x] `shellcheck launch.sh` — clean
- [x] Ran preflight against `creek-demo-2026-05-30`: passed, auto-mirrored
      voice-core, wrote a correct `crawdad.yaml`
- [x] Launched the bot: `crawdad connected as CrawDad#1656`, `/crawdad`
      commands synced, MCP tools advertised

### DIFF
--- a/.claude/skills/crawdad-launch/SKILL.md
+++ b/.claude/skills/crawdad-launch/SKILL.md
@@ -9,6 +9,9 @@ description: >-
   secrets, allowlists) and starts the bot in the background. Do NOT use for
   driving the `creek` CLI / writing essays (use creek-cli) or editing
   crawdad/creek-tools source (use stay-green/work-issue).
+metadata:
+  author: Geoff
+  version: 1.0.0
 ---
 
 # crawdad-launch
@@ -19,8 +22,8 @@ per turn and runs a Haiku-router → MCP → Sonnet-composer agent loop.
 
 ## Step 1 — Ask which vault (always)
 
-Do **not** assume the vault. Ask the user for the vault directory. Offer the
-known options as a sensible default:
+Do **not** assume the vault. Ask the user for the vault directory. On this
+machine the known options (examples — substitute the user's actual vaults) are:
 
 - Demo / prod: `/Users/geoffgallinger/Documents/creek-demo-2026-05-30`
 - Primary: `/Users/geoffgallinger/Documents/creek`
@@ -30,10 +33,11 @@ Run unattended against the **demo** vault unless told otherwise.
 ## Step 2 — Preflight + write the launch config
 
 Run the bundled script with the chosen vault. It validates the vault, ensures
-the env secrets, fixes the known gotchas, and writes `crawdad/crawdad.yaml`:
+the env secrets, fixes the known gotchas, and writes `crawdad/crawdad.yaml`.
+Resolve the script via the repo root so this works on any checkout:
 
 ```bash
-bash /Users/geoffgallinger/Projects/creek-tools/.claude/skills/crawdad-launch/launch.sh <VAULT>
+bash "$(git rev-parse --show-toplevel)/.claude/skills/crawdad-launch/launch.sh" <VAULT>
 ```
 
 What it checks/fixes (each is a real launch failure mode):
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
Thanks — addressed all of it in b26e704:

**Blockers**
- 🔴 **Hardcoded paths in SKILL.md** — Step 2 now resolves the script via `bash "$(git rev-parse --show-toplevel)/.claude/skills/crawdad-launch/launch.sh"`; Step 3 instructs running the fully-resolved launch command the script prints under `Launch with:` (repo-derived absolute paths), so no machine-specific paths remain in the shared file.
- 🔴 **`sed -i ''` macOS-only** — switched to `sed -i.bak "…" "$VAULT_CONFIG" && rm -f "${VAULT_CONFIG}.bak"`, which works on GNU (Linux/CI) and BSD/macOS. The `.bak` also gives the pre-edit backup flagged under Code Quality.

**Should-fix**
- 🟡 **YAML injection** — the generated `crawdad.yaml` now single-quotes every path value, so a path with `:`/`#`/`&`/`*` can't corrupt the config or inject into the executed `mcp_server_command` argv.

**Code Quality**
- **Multi-entry allowlists** — `read_list()` now extracts *all* numeric ids per sequence (was first-only); verified against a 2-user / 3-channel fixture.
- **ID validation** — each extracted id is gated by `=~ ^[0-9]+$` before it enters the config.
- **Backup before edit** — covered by the `sed -i.bak` change above.
- **Frontmatter** — added the `metadata: author/version` block to match convention.
- **Discord defaults / vault paths** — flagged the default snowflakes as the author's own (replace-with-your-own) and the Step 1 vault paths as machine-specific examples.

Verified: shellcheck clean on bash 3.2, and the quoted config round-trips through `crawdad.config.load_config`.

Geoffe-Ga:
@claude please re-review

Geoffe-Ga:
Addressed all three blockers plus the noted nits in 53c1cf9:

**Blockers**
- 🔴 **Hardcoded vault paths in SKILL.md** — removed. Step 1 now discovers vaults via `find ~ -path '*/00-Creek-Meta/creek_config.yaml'` and offers the vault recorded in the existing `crawdad.yaml` (the last one launched) as the default. No machine-specific paths remain in the committed file.
- 🔴 **Step 1 / Step 2 contradiction** — resolved. The behaviour is now unambiguous: *always ask the user*, with the last-used vault as the default when they have no preference. The "run unattended against demo" line is gone.
- 🟡 HIGH **Hardcoded Discord IDs** — removed from source. Defaults now come from `CRAWDAD_DEFAULT_USER` / `CRAWDAD_DEFAULT_CHANNEL`; a fresh vault with no prior config and no env defaults fails with an actionable message rather than silently allowing anyone.

**Also fixed**
- 🔴 **`.bak` created then deleted** — replaced with an atomic `mktemp` + `mv` rewrite; the original survives if `sed` fails, and there's no misleading backup claim.
- 🟢 **Single-quote injection** — the YAML emitter now doubles embedded single quotes, so `Geoff's Creek` produces valid YAML. Verified by round-tripping a path containing `'`, `:`, `#`, `&` through PyYAML.

**On the remaining nits**
- *`creek state` stderr* — stdout goes to `/dev/null` but stderr is intentionally left attached, so the actual error message already surfaces on failure.
- *bats/unit tests* — agreed these would harden the awk/sed logic; I'd prefer to file that as a follow-up rather than add a test harness in this skill PR. Happy to do so if you'd rather it block.

Verified: shellcheck clean on bash 3.2; allowlist env-fallback (unset→fail, valid→ok, non-numeric→fail); demo config still round-trips through `crawdad.config.load_config`.

Geoffe-Ga:
@claude please re-review

PR #547, opened 2026-06-05
fix(skills): crawdad-launch MCP wrapper — forward ANTHROPIC_API_KEY, fix uv-spawn flakiness

Geoffe-Ga:
## Summary

Fixes two real CrawDad launch failures observed on the demo deployment, both
rooted in how CrawDad spawns the MCP server. `crawdad/crawdad/mcp_client.py:138`
builds `StdioServerParameters` with **no `env=`**, so the MCP stdio SDK launches
the subprocess with a scrubbed environment and a bare `uv run` command.

**Symptom 1 — degraded voice/cloud tools.** The scrubbed env drops
`ANTHROPIC_API_KEY`, so in-MCP cloud tools (`creek.draft`/`author`/`mine`/
`classify`) silently fall back to non-cloud behaviour and log *"Anthropic
provider not available"*. CrawDad's own composer has the key; the spawned tool
server doesn't.

**Symptom 2 — `mcp_unavailable` replies.** Spawning `uv run … creek-tools-mcp`
every turn is slow and contends on uv's project lock when the agent loop runs
several rounds, surfacing as `could not start or connect to the MCP subprocess
('uv')` → the user sees *"creek-tools is unreachable; try again in a moment."*

## Fix

The `crawdad-launch` skill now generates `mcp_server_command` as a **login-shell
wrapper**:

```yaml
mcp_server_command:
  - '/bin/zsh'
  - '-lc'
  - 'export CREEK_ANTHROPIC_CONSENT=1; exec ''<creek-tools>/.venv/bin/creek-tools-mcp'' --config ''<vault>/00-Creek-Meta/creek_config.yaml'''
```

- **Login shell** re-establishes the user's exported `ANTHROPIC_API_KEY` from
  their profile, surviving the SDK's env scrub (no secret is written to any file).
- **`exec`s the venv entry point directly** instead of `uv run`: deterministic,
  no lock contention, and `exec` hands stdio straight to the server so the
  JSON-RPC stream stays clean. Falls back to `uv run` if the venv binary isn't
  built yet.
- **`CREEK_ANTHROPIC_CONSENT=1`** enables the cloud tools; documented that this
  egresses vault content to Anthropic on the user's own key (intended use).

Skill bumped to 1.1.0; SKILL.md documents the wrapper and the consent trade-off.

## Test Plan

- [x] `shellcheck launch.sh` — clean
- [x] Generated config round-trips through `crawdad.config.load_config` (argv
      parses to the 3-element `zsh -lc …` wrapper)
- [x] Under a scrubbed env (`env -i HOME=… PATH=/usr/bin:/bin`), the wrapper
      restores `ANTHROPIC_API_KEY` (True) and `CREEK_ANTHROPIC_CONSENT` (1)
- [x] Real server boots under the scrubbed env with **0 bytes on stdout** (no
      JSON-RPC corruption)
- [x] Live bot restarted on the new config: `MCP tools advertised: …` (handshake
      via the wrapper succeeded) and `connected as CrawDad#1656`

### DIFF
--- a/.claude/skills/crawdad-launch/SKILL.md
+++ b/.claude/skills/crawdad-launch/SKILL.md
@@ -79,6 +79,12 @@ What it checks/fixes (each is a real launch failure mode):
   `uv run` if the venv binary isn't built yet (`uv sync` in `creek-tools`).
   **`CREEK_ANTHROPIC_CONSENT=1` means vault content egresses to Anthropic on the
   user's key** when those tools run — intended for the user's own vault.
+  Because the login shell only sources *login* profiles (`~/.zprofile`,
+  `~/.zshenv`, `~/.bash_profile`, `~/.profile`) and **not** interactive rc files
+  (`~/.zshrc`, `~/.bashrc`), `ANTHROPIC_API_KEY` must be exported in a login
+  file. The script probes for this under a scrubbed env and prints a `[warn]`
+  (with the fix) if the key is only in an rc file — heed it or the cloud tools
+  degrade exactly as before.
 
 If the script prints `ERROR:`, stop and fix what it reports before launching.
 
--- a/.claude/skills/crawdad-launch/SKILL.md
+++ b/.claude/skills/crawdad-launch/SKILL.md
@@ -96,9 +96,11 @@ cd <REPO>/crawdad && uv run crawdad run --config <REPO>/crawdad/crawdad.yaml
 
 Use `run_in_background: true`. Then read the background output to confirm it
 connected to Discord (look for the gateway/ready log line) and report the
-channel it's listening in. Both `crawdad` and `creek-tools-mcp` run from source
-via `uv run`, so **to pick up merged/edited code just restart the bot** — no
-reinstall needed.
+channel it's listening in. `crawdad` runs from source via `uv run`, and
+`creek-tools-mcp` runs from its pre-built editable venv binary (or `uv run` in
+the fallback) — both reflect source edits, so **to pick up merged/edited code
+just restart the bot**, no reinstall needed. Only a *dependency* change needs a
+`uv sync` in `creek-tools` first.
 
 ## Notes
 
--- a/.claude/skills/crawdad-launch/SKILL.md
+++ b/.claude/skills/crawdad-launch/SKILL.md
@@ -11,7 +11,7 @@ description: >-
   crawdad/creek-tools source (use stay-green/work-issue).
 metadata:
   author: Geoff
-  version: 1.0.0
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
Thanks for the thorough pass — addressed all four Code Quality items in fd34e5a (none deferred):

1. **Stale Step 3 note** — reworded: `crawdad` runs via `uv run`; `creek-tools-mcp` runs from its editable venv binary (still reflects source edits; only a *dependency* change needs `uv sync`). Restart-to-update still holds.
2. **`/bin/bash` fallback** — `LOGIN_SHELL` now defaults to `/bin/bash` (universally present) instead of `/bin/zsh`.
3. **fish edge case** — added the guard: a fish `$SHELL` is coerced to `/bin/bash` with a `[warn]`, since fish can't run `export VAR=val; exec …`.
4. **`sq()` comment** — added a one-liner mirroring the `shq()` comment so the two helpers aren't confused.

shellcheck still clean; the generated config round-trips. Requesting a quick re-review to confirm the final HEAD before merge.

Geoffe-Ga:
@claude please re-review

Geoffe-Ga:
Addressed the re-review items in 2349d75:

- **Login-profile / rc-file gap (highest-priority)** — the script now *probes* it: under a scrubbed env (`env -i …`, mimicking the MCP SDK's env scrub) it runs `$LOGIN_SHELL -lc 'printenv ANTHROPIC_API_KEY'` and prints `[ok]` if the login profile exposes the key, or a `[warn]` telling the user to move the export from `~/.zshrc`/`~/.bashrc` into a login file (`~/.zprofile`/`~/.zshenv`/`~/.bash_profile`). On this machine it reports `[ok]`, so the wrapper fix is now verified live rather than assumed. Also documented in SKILL.md.
- **Consent preflight log** — added `[ok] CREEK_ANTHROPIC_CONSENT=1 in MCP wrapper — cloud tools … egress vault content to Anthropic on your key`.
- **Comment placement** — folded the `/bin/bash` line into the block comment above `LOGIN_SHELL`.

On the optional "hoist `shq`/`sq` to a functions section": leaving as-is for now since the script is still short and the helpers sit right where they're used; happy to file a follow-up if you'd prefer.

shellcheck clean; config still round-trips.

Geoffe-Ga:
@claude please re-review

Geoffe-Ga:
Merged `main` (PR #548, pip>=26.1.2 / PYSEC-2026-196) into this branch to unblock the pip-audit gate — no skill changes since the LGTM. @claude please re-review the refreshed HEAD.

Issue #610, opened 2026-06-09
[#7] CrawDad OpenAI + Gemini providers + config selection


Geoffe-Ga:
**Epic:** #603 · **Sequence:** 7/8 · **Risk:** medium–high · **Depends:** #6

## Role
Python engineer in `crawdad/` (async), security lens, max-quality bar.

## Goal
Make CrawDad's backend selectable via `CRAWDAD_PROVIDER`, with OpenAI and Gemini async providers alongside Anthropic, keys discovered from env.

## Context
- After #6, `crawdad/crawdad/llm/` holds the `AsyncLLMProvider` Protocol, an Anthropic impl, and `build_async_provider(config)`.
- `crawdad/crawdad/config.py` today: eagerly requires `ANTHROPIC_API_KEY`, stores it on the frozen `CrawDadConfig`, and defines `DEFAULT_ROUTER_MODEL` / `DEFAULT_COMPOSER_MODEL` (env-overridable) — the only model literals in the package, enforced by `tests/test_no_model_literals.py`.
- `load_config` merges `crawdad.yaml` + env secrets.

## Deliverables
1. `config.py`:
   - Add `llm_provider: str` from `CRAWDAD_PROVIDER` (default `anthropic`).
   - Replace the unconditional `ANTHROPIC_API_KEY` check with a **provider-conditional** one: require the env key matching `llm_provider` (`ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `GOOGLE_API_KEY`), with a clear error naming the missing var.
   - **Stop storing the key on `CrawDadConfig`** — let the SDK read it from env (removes a secret from a frozen, widely-passed object).
   - Per-provider router/composer model defaults (Haiku-tier vs Sonnet-tier mapping for OpenAI/Gemini), all confined to `config.py`, env-overridable as today.
2. `crawdad/crawdad/llm/openai.py`, `gemini.py`: async providers mirroring the sync ones from creek-tools #4/#5 (same response/usage/error-redaction mapping; async SDK clients reading env keys).
3. `build_async_provider`: register `openai` + `gemini`.
4. `cli.py`: already provider-agnostic after #6 — verify it passes `config.llm_provider` through.
5. `pyproject.toml`: add `openai` + `google-genai` (match the package's dependency style — direct or optional extras). Regenerate the lock if present; update dev/test installs so mocks import.
[...]

Issue #603, opened 2026-06-09
EPIC: Swappable LLM providers (Anthropic / OpenAI / Gemini), keys from env


Geoffe-Ga:
## Summary

Make the cloud LLM backend selectable — Anthropic, OpenAI, or Google Gemini — for both `creek-tools` and CrawDad, via one config string per package. API keys are discovered from the environment by each provider's SDK and never written into the repo, config files, logs, or errors. Ollama (local) stays the `creek-tools` default.

## Why

Today the provider is effectively hard-wired:
- `creek-tools` has a real but two-way dispatch (`anthropic` vs Ollama).
- CrawDad is hard-wired to `anthropic.AsyncAnthropic`.

Operators should be able to swap providers without code changes and without ever storing a key in the repo.

## Architectural decision (load-bearing)

**Keep `creek-tools` and CrawDad decoupled.** They are siblings (FEAT-013): CrawDad imports nothing from `creek-tools`, talking to it only over MCP stdio. We do **not** introduce a shared `creek.llm` package. Each package gets its own provider abstraction — **sync** in `creek-tools`, **async** in CrawDad — mirroring each other's structure (a normalized `Completion` type + a `complete()` contract) without sharing a module. DRY holds within each package; the ~30–40-line per-provider wrapper overlap is the accepted price of the clean boundary. Revisit only if the duplicated bodies grow real logic.

## Constraints (apply to every child issue)

1. No `api_key` field on any config model — SDKs read env (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GOOGLE_API_KEY`).
2. Re-raise native SDK errors as `RuntimeError(type(exc).__name__)` — never leak request state.
3. Cloud-egress consent is provider-neutral; Ollama exempt.
4. No model-ID literals outside config modules.
5. Quality bar unchanged (coverage ≥90%, docstrings ≥95%, complexity ≤10, mypy strict, ruff/bandit clean).

## Children (sequenced, tracer-code; each ships green)

- [x] #604 — Normalize `Completion` + `LLMProvider` Protocol (creek-tools) ✅ merged in PR #612
- [ ] #605 — Provider factory; route orchestrator + author client through it
- [ ] #606 — Generalise the cloud-consent gate (provider-neutral)
- [ ] #607 — `OpenAIProvider` (sync) + optional dep + tests
- [ ] #608 — `GeminiProvider` (sync) + optional dep + tests
[...]

PR #626, opened 2026-06-10
fix(llm): fail-fast provider validation at config-load in both packages (#620)

### LINKED ISSUE #620: bug(llm): validate provider string at config-load (fail-fast) in both packages
**Follow-up to the merged provider epic #603.** Bug — silent late failure.

## Problem
`LLMConfig.provider` (`creek-tools/creek/config.py:95`) is a free `str = "ollama"` with no `Literal` constraint and no validator. A typo like `provider: anthropics` in `creek_config.yaml` parses cleanly and only surfaces as a `ValueError` from `build_provider` (`providers.py:1042-1047`) **when classification actually runs** — potentially long after startup.

CrawDad is worse in one spot: `router_model_for` / `composer_model_for` (`crawdad/crawdad/config.py:118-150`) resolve an unknown provider via `.get(provider, <anthropic default>)`, so a mistyped `CRAWDAD_PROVIDER` **silently falls back to Claude models** instead of erroring. (`build_async_provider` does raise, but the model-tier resolution masks the typo first.)

## Why it matters
A config typo should fail fast and loudly at load, not silently pick the wrong backend or crash mid-run after the operator thinks the bot booted fine.

## Proposed fix
- **creek-tools**: add a `field_validator("provider")` (or `Literal`) on `LLMConfig` bound to the registered keys (`_PROVIDER_REGISTRY` in `providers.py`). Keep the registry the single source — the validator reads its keys rather than re-listing them, to avoid drift.
- **CrawDad**: validate `CRAWDAD_PROVIDER` against `_PROVIDER_BUILDERS` keys at config-load (`load_config`), raising a clear error naming the unknown value and the valid set. Remove the silent anthropic fallback in `router_model_for` / `composer_model_for` (an unknown provider should already be impossible after validation).

## Acceptance
- An unknown provider string raises a clear error at config-load in **both** packages, naming the bad value and the valid set.
- The valid set is derived from the registry/builders, not a hand-maintained second list.
- Tests cover: each valid provider loads; an unknown provider raises at load (not at first call); CrawDad no longer silently downgrades an unknown provider to Claude.
- `./scripts/check-all.sh` green in both packages.

Geoffe-Ga:
## Summary

Follow-up bug fix to epic #603: a mistyped provider should **fail fast and loudly at config-load**, not silently pick the wrong backend or crash mid-run.

- **creek-tools**: adds a `field_validator("provider")` on `LLMConfig` that rejects an unregistered provider **at construction** (so `provider: anthropics` in `creek_config.yaml` raises at parse time, not when classification first runs). The valid set is read from a new `providers.known_providers()` — the **single source**, derived from `_PROVIDER_REGISTRY`, so it can't drift from what `build_provider` dispatches on. `build_provider` keeps its own `ValueError` as defense-in-depth.
- **CrawDad**: `router_model_for` / `composer_model_for` no longer **silently fall back to the Anthropic tier** for an unknown provider — they raise a clear `ValueError` (factored into `_resolve_model`). `load_config` already validates `CRAWDAD_PROVIDER` (from #610); a new regression test pins `_PROVIDER_KEY_ENV` / `_ROUTER_MODEL_DEFAULTS` / `_COMPOSER_MODEL_DEFAULTS` and `llm._PROVIDER_BUILDERS` to the **same provider set** so the lists can't drift (the anti-drift guarantee in place of a hand-maintained second list).

## Test plan

- **creek-tools**: every registered provider loads; `LLMConfig(provider="anthropics")` raises `ValidationError` at construction (the new fail-fast); `build_provider`'s guard is still exercised via `model_construct` (defense-in-depth).
- **CrawDad**: `router_model_for("frobnicate")` / `composer_model_for(...)` raise instead of returning a Claude model; the provider-registry sync test ties all four provider maps together; existing load-fail-fast test (#610) still green.
- `./scripts/check-all.sh` → exit 0 in **both** packages (creek-tools 12 gates; crawdad 7 gates).

## Acceptance

- ✅ Unknown provider raises a clear error at config-load in both packages, naming the bad value and the valid set.
- ✅ Valid set derived from the registry/builders (creek: `known_providers()`; crawdad: drift-guard test), not a hand-maintained second list.
- ✅ CrawDad no longer silently downgrades an unknown provider to Claude.

Closes #620
Refs #603

### DIFF
--- a/crawdad/crawdad/config.py
+++ b/crawdad/crawdad/config.py
@@ -118,9 +118,9 @@ _COMPOSER_MODEL_DEFAULTS: dict[str, str] = {
 def _resolve_model(provider: str, override_env: str, defaults: dict[str, str]) -> str:
     """Resolve a model tier for *provider*, honoring an env override.
 
-    Raises on an unknown provider rather than silently falling back (#620) —
-    `load_config` validates the provider, so an unknown value here is a
-    programmer error, not an operator typo to paper over.
+    Raises on an unknown provider rather than silently falling back to the
+    Anthropic default — the load-time validator catches operator typos, so an
+    unknown value here is a programmer error, not something to paper over.
 
     Args:
         provider: The selected backend.
@@ -141,6 +141,7 @@ def _resolve_model(provider: str, override_env: str, defaults: dict[str, str]) -
     except KeyError:
         known = ", ".join(sorted(defaults))
         msg = f"unknown provider {provider!r}; expected one of: {known}"
+        # The KeyError adds no caller-useful context; suppress its chain.
         raise ValueError(msg) from None
 
 
--- a/crawdad/crawdad/config.py
+++ b/crawdad/crawdad/config.py
@@ -115,6 +115,35 @@ _COMPOSER_MODEL_DEFAULTS: dict[str, str] = {
 }
 
 
+def _resolve_model(provider: str, override_env: str, defaults: dict[str, str]) -> str:
+    """Resolve a model tier for *provider*, honoring an env override.
+
+    Raises on an unknown provider rather than silently falling back (#620) —
+    `load_config` validates the provider, so an unknown value here is a
+    programmer error, not an operator typo to paper over.
+
+    Args:
+        provider: The selected backend.
+        override_env: The env var that overrides the default tier when set.
+        defaults: The per-provider default model map.
+
+    Returns:
+        The override when set, else the provider's default tier.
+
+    Raises:
+        ValueError: When *provider* names no known backend.
+    """
+    override = os.environ.get(override_env)
+    if override:
+        return override
+    try:
+        return defaults[provider]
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
Addressed the review feedback in 59121c4:

1. ✅ Removed all three issue-620 references from production source docstrings (`crawdad/crawdad/config.py`, `creek-tools/creek/classify/llm/providers.py`, `creek-tools/creek/config.py`), using the suggested rewrite for `_resolve_model`.
2. ✅ Added a comment in `test_provider_registries_stay_in_sync` acknowledging the deliberate private-name imports.
3. ✅ Annotated the intentional `from None` chain suppression in `_resolve_model`.

Both packages' `check-all.sh` pass locally (creek-tools 12/12, crawdad 7/7).

@claude please re-review

PR #629, opened 2026-06-10
test(llm): add opt-in live smoke test per provider for model onboarding (#623)

### LINKED ISSUE #623: test(llm): add opt-in live smoke test per provider for model onboarding
**Follow-up to the merged provider epic #603.** Testing gap.

## Problem
Every provider test in both packages mocks the vendor SDK (`tests/test_classify.py`, `tests/test_openai_provider.py`, `tests/test_gemini_provider.py`, and the CrawDad equivalents). Nothing ever makes a real API call. That's correct for unit coverage, but it means **onboarding a specific model id is unverified until the first live run** — and the things most likely to drift are exactly what mocks paper over: valid model ids, `finish_reason`/stop-reason enums, and `usage` field names per vendor.

## Why it matters
"Readiness for other LLM models" can't be claimed from green mocks alone. When you point at a new model, you want a one-command smoke that proves the round-trip (auth → request → normalized `Completion`) against the live API.

## Proposed fix
- Add an opt-in, `@pytest.mark.integration` smoke test per provider that makes a single tiny real `complete()` call and asserts a non-empty normalized `Completion` (text + a recognized stop reason + usage present).
- Gate it on the provider's key being present in the env; skip cleanly when absent so default CI stays mock-only and offline.
- Document the one-liner to run it when onboarding a new model (e.g. `./scripts/test.sh --integration -k openai`).
- Keep it out of the default `not integration and not e2e` selection so it never runs (or bills) in normal CI.

## Acceptance
- One integration-marked smoke per provider exists in both packages, skipped when the key is absent.
- A documented command runs the smoke for a chosen provider/model.
- Default `./scripts/check-all.sh` / CI behavior is unchanged (smoke not selected).

Geoffe-Ga:
## Summary
- Adds one `integration`-marked live smoke test per provider in **both packages**: creek-tools (`tests/test_live_smoke.py` — anthropic / openai / gemini / ollama) and CrawDad (`tests/test_live_smoke.py` — anthropic / openai / gemini, async). Each makes a single tiny real `complete()` call and asserts the normalized `Completion` round-trip (non-empty text, recognized stop reason, per-provider usage semantics). Every smoke skips cleanly when its provider's key is absent (Ollama gates on endpoint reachability instead).
- Makes the smoke a one-command model-onboarding check, documented in both READMEs: `CREEK_SMOKE_MODEL=<id> ./scripts/test.sh --integration -k <provider>` (creek-tools) and `CRAWDAD_COMPOSER_MODEL=<id> ./scripts/test.sh -m integration --no-cov` (CrawDad, via the existing env override).
- Keeps default runs and CI live-call-free: CrawDad's `test.sh` now deselects the `integration` marker on bare invocation (previously bare `pytest` would have selected it), and creek-tools' `--integration`/`--e2e` selections pass `--no-cov` — fixing a pre-existing breakage where the project-wide `--cov-fail-under=90` from pyproject `addopts` made any marker-only run fail unconditionally. `test.sh` also gains `-k` passthrough, which the issue's own proposed command assumed.

## Test plan
- With all vendor keys explicitly blanked: creek-tools `./scripts/test.sh --integration -k live_smoke` → 4 skipped, 5467 deselected; CrawDad `./scripts/test.sh -m integration --no-cov -k live_smoke` → 3 skipped.
- CrawDad bare `./scripts/test.sh` → 481 passed, 3 deselected, coverage 94.87% (default behavior preserved, smokes excluded).
- `./scripts/check-all.sh` exits 0 in both packages (12/12 and 7/7).
- Live-call paths exercised by design only when an operator opts in with keys present (no keys available in this session — see #623's thread for the first verification target).

Closes #623
Refs #603

### DIFF
--- a/crawdad/README.md
+++ b/crawdad/README.md
@@ -89,9 +89,11 @@ CrawDad's provider abstraction is intentionally **decoupled** from creek-tools' 
 **Live smoke test (model onboarding).** Unit tests mock every vendor SDK, so a model tier is only proven by a real call. With the provider's key in the env, one command makes a single tiny live request and asserts the normalized round-trip:
 
 ```bash
-./scripts/test.sh -m integration --no-cov -k openai                                       # smoke the composer's default tier
-CRAWDAD_COMPOSER_MODEL=some-new-model ./scripts/test.sh -m integration --no-cov -k gemini # smoke a candidate model id
+./scripts/test.sh -m integration -k openai                                       # smoke the composer's default tier
+CRAWDAD_COMPOSER_MODEL=some-new-model ./scripts/test.sh -m integration -k gemini # smoke a candidate model id
 ```
+
+(`test.sh` injects `--no-cov` for integration runs so the project-wide coverage gate doesn't bury the smoke output.)
 
 Each smoke skips cleanly when its key is absent; the default `./scripts/test.sh` run deselects the `integration` marker, so CI never makes (or bills) a live call.
 
--- a/crawdad/README.md
+++ b/crawdad/README.md
@@ -86,6 +86,15 @@ Starting the bot with the selected provider's key unset fails fast with a messag
 
 CrawDad's provider abstraction is intentionally **decoupled** from creek-tools' (sibling packages, no shared module) — see [ADR-0003](../creek-tools/docs/architecture/ADR/0003-decoupled-provider-abstractions.md).
 
+**Live smoke test (model onboarding).** Unit tests mock every vendor SDK, so a model tier is only proven by a real call. With the provider's key in the env, one command makes a single tiny live request and asserts the normalized round-trip:
+
+```bash
+./scripts/test.sh -m integration --no-cov -k openai                                       # smoke the composer's default tier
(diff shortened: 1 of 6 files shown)


Geoffe-Ga:
Addressed the review in 2caeb23 (chose to fix in-PR rather than file follow-ups, since the items were small):

1. ✅ **(HIGH)** `crawdad/scripts/test.sh` now auto-injects `--no-cov` whenever the args target the `integration` marker — with one adjustment to the suggested snippet: a bare `*integration*` glob would also match the default `-m "not integration"` selection and silently disable the coverage gate on every bare run, so the match excludes the negated form. Verified both ways: `-m integration` runs coverage-free; bare run keeps the 90% gate (481 passed, 94.87%).
2. ✅ `_messages()` helper replaced with a `_MESSAGES` module constant.
3. ✅ CrawDad smokes now assert `usage is not None` (all three providers extract usage; all three vendors report it on successful completions), with a comment stating the contract.
4. ✅ Each live call is wrapped in `asyncio.wait_for(..., timeout=30)` so a hung endpoint cannot stall the suite — no new dependency.

READMEs/docstrings updated to drop the now-unneeded explicit `--no-cov`. `check-all.sh` green in both packages (12/12, 7/7).

@claude please re-review

Issue #715, opened 2026-06-28
epic: Resonance discovery at scale


Geoffe-Ga:
# epic: Resonance discovery at scale

**SPEC:** `creek-tools/plans/2026-06-28_SPEC_vault-prod-readiness.md` (gap #6, "stretch")

## Outcome

Resonance/thread linking scales past ~35k fragments **without the O(n²) blowup**
— `creek link --method embeddings` completes in predictable time/memory on a
large vault by computing **top-k nearest neighbours per fragment** instead of all
above-threshold pairs.

## Why

`EmbeddingLinker.find_resonances` is memory-blocked (#596) but still all-pairs:
35k fragments = ~1.2B comparisons + an unbounded edge set, which is the slow step
and the historical "killed at 35k" cause. Embedding *generation* is already fine
(local, cached, incremental) — only the pairwise discovery needs the scale fix.

## Success criteria

- A top-k / ANN path bounds output to O(n·k) edges and bounds compute, behind a
  config toggle; the existing exact all-pairs path stays available + correct.
- Linking on a 35k-fragment vault completes within a documented time/memory
  budget; results are stable and the resonance quality is validated against the
  exact path on a fixture.

## Child Issues
- [ ] #722 — top-k / ANN resonance discovery (FAISS/hnswlib/sklearn), config-gated

## Constraints

Local-only (no new cloud dependency). The exact path remains the default until
the ANN path is validated. Standard 4-gate workflow.

Geoffe-Ga:
Sole child #722 merged (PR #735): a config-gated top-k resonance path (`resonance_method: topk`, `resonance_top_k`) bounds resonance discovery to O(n·k) via exact blocked numpy.argpartition — no new dependency, subset-of-exact, exact path stays the default. Resonance discovery at scale is delivered. Closing the epic.

Issue #596, opened 2026-06-08
perf(link): embeddings resonance is O(n²) and does not scale past ~10k fragments


Geoffe-Ga:
> SPEC: plans/github-issues/2026-06-07_SPEC_ingest-realworld-exports-and-link-scale.md

## Goal
Make `creek link --method embeddings` resonance discovery scale to large vaults (≥50k fragments) instead of brute-force O(n²) pairwise comparison.

## Context / root cause
- `creek/link/embeddings.py:~717` → `_find_resonances` iterates `itertools.combinations(range(len(ids)), 2)` and cosine-compares every pair. The code already comments: *"OPS-004: pairwise loop is O(N²); on a 10k-fragment vault this …"*.
- On a 35k-fragment vault (~600M pairs) the resonance pass was **killed mid-run** during a real bootstrap (2026-06-07). The 30k short Discord chat fragments dominated the corpus and drove the blowup.
- Embedding **generation** + the `embeddings.parquet` cache (O(n)) complete fine, and `--method temporal` (~O(n log n), 7.6k links on the same vault) is fine — only the pairwise resonance step fails to scale.

## Output format
- Replace the brute-force pairwise loop with a scalable nearest-neighbour approach: per-fragment **top-k** via a vector index (FAISS / hnswlib) or chunked thresholded matrix ops (numpy), optionally blocked by metadata (source/time). Threshold semantics (`similarity_threshold`, default 0.75) preserved.
- Keep the O(n) cache-build path unchanged (retrieval/mine/draft/author depend on `embeddings.parquet`).
- Benchmark/test asserting tractable runtime + bounded memory at ≥50k fragments (synthetic vectors), and that high-similarity pairs are still found (parity with the brute-force result on a small set).

## Constraints
- Don't regress resonance recall on small vaults (parity test).
- Any new dependency (FAISS/hnswlib) must be optional/guarded with a numpy fallback.
- No `# noqa`/`# type: ignore` to pass a gate.

## Done-Done
- [ ] Tests first: small-vault parity + large-vault tractability.
- [ ] `./scripts/check-all.sh` green (12/12).
- [ ] PR; reviewer Verdict LGTM.

PR #735, opened 2026-06-29
perf(link): top-k/ANN resonance discovery for scale (#722)

### LINKED ISSUE #722: perf(link): top-k/ANN resonance discovery for scale
## Role
Senior Python engineer in this repo's `creek/link/embeddings.py` +
`creek/config.py` (EmbeddingsConfig).

## Goal
Add a **top-k nearest-neighbour** resonance-discovery path so
`EmbeddingLinker.find_resonances` scales past ~35k fragments without the O(n²)
all-pairs blowup — bounding both compute and the emitted edge count — while
keeping the exact all-pairs path as the validated default.

## Context
- Parent epic: #715 (SPEC gap #6, "stretch").
- `find_resonances` (`creek/link/embeddings.py:657`) is already **memory**-blocked
  (#596 — block matmul, never the full NxN matrix) but still computes all pairs
  and emits every above-threshold edge. At 35k that is ~1.2B comparisons and a
  potentially huge edge set — the slow step and the historical "killed at 35k".
- Embedding **generation** is already fine (local sentence-transformers, parquet
  cache, incremental) — DO NOT touch it.
- Hierarchy suppression (`_is_hierarchy_suppressed`, FEAT-024) and the
  `Resonance` output shape must be preserved.

## Output Format
A single PR:
- [ ] A top-k path: for each fragment, keep its top-k most-similar neighbours
  above threshold (k from a new `EmbeddingsConfig` field, default e.g. 10),
  bounding output to O(n·k). Use an ANN index (FAISS / hnswlib) or an exact
  blocked top-k (numpy argpartition per row block) — pick per a documented
  trade-off; ANN behind an optional extra if it adds a heavy dep.
- [ ] Config-gated: the exact all-pairs path remains the default until validated;
  a config toggle selects top-k/ANN.
- [ ] Hierarchy suppression + `Resonance` records identical to the exact path for
  the pairs both emit.
- [ ] Tests: top-k returns each fragment's correct top-k above threshold on a
  fixture; output is a subset of the exact path's edges; a perf/scale test
  (skippable/marked) shows bounded behaviour at a large synthetic N.

## Examples
```yaml
embeddings:
  model: all-MiniLM-L6-v2
  similarity_threshold: 0.75
  resonance_method: topk        # exact (default) | topk
  resonance_top_k: 10
```

## Constraints
**Scope fence:** the resonance-discovery path only. No change to embedding
generation, the cache format, or what a resonance means. Local-only — if an ANN
lib is added, it goes behind an optional extra, not the base install.

> linter/type-checker silencers. Fix the root cause. The only exception is the
> documented 4-line escape hatch (third-party library bug / language-version
> compatibility / benchmarked performance necessity / generated code) — and it
> must include the reason, a reference URL, an alternative considered, and a
> review date. See the `max-quality-no-shortcuts` skill.

- [ ] `./scripts/check-all.sh` green (exit 0).
- [ ] Coverage on changed lines ≥90%.
- [ ] PR body includes `Refs #715` and `Closes #<this-issue>`.
- [ ] Latest Claude `Verdict:` on HEAD is `LGTM`.

## Labels
`spec-decomposition`, `prod-readiness`, `performance`

Geoffe-Ga:
## Summary
Adds a config-gated **top-k nearest-neighbour** resonance path so `EmbeddingLinker.find_resonances` scales past ~35k fragments without the O(n²) all-pairs edge blowup — while keeping the exact all-pairs path as the validated default.

- **Config:** `EmbeddingsConfig` gains `resonance_method` (`exact` | `topk`, default `exact`) and `resonance_top_k` (default 10). The exact path is unchanged and stays the default until `topk` is validated on a real vault.
- **Algorithm:** `topk` keeps each fragment's *k* most-similar above-threshold neighbours via an **exact blocked `numpy.argpartition` per row block** — no new dependency, reusing the #596 block-matmul so peak memory stays `block_rows × N` (never the full N×N matrix). Each directed (i → neighbour) pick is folded to its undirected `(i < j)` form and de-duplicated, so the output is always a **subset** of the exact path's edges, emitted in the same ascending `(i, j)` order. Bounds the edge set to **O(n·k)**.
- **Invariants preserved:** FEAT-024 hierarchy suppression and the `Resonance` record shape are identical to the exact path for the pairs both emit.
- **Refactor:** the exact inner loop is extracted to `_resonances_exact` alongside the new `_resonances_topk`; `_top_k_neighbours` selects per row. Behaviour-preserving — all 257 existing embeddings/link/config tests pass unchanged.

I chose **exact blocked top-k over an ANN library (FAISS/hnswlib)**: it adds no heavy optional dependency, stays exact (no recall loss), and reuses the existing memory-bounded block-matmul. ANN remains a future option behind an optional extra if k-NN recall ever needs sub-linear scaling.

## Test plan
New `tests/test_topk_resonance.py` (deterministic angular fan of unit vectors — no similarity ties — so each fragment's top-k is hand-verifiable):
- `exact` is the default method; the exact path emits all six pairwise edges;
- k=1 keeps each fragment's single best neighbour (`{(F0,F1),(F1,F2),(F2,F3)}`);
- top-k output is a strict **subset** of the exact edges (k=2 drops only the weakest, F0↔F3);
- top-k preserves ascending `(i, j)` order;
- `@pytest.mark.slow` scale test: n=800, k=5, threshold 0 (every pair qualifies) → edge count bounded by **n·k** (vs ~n²/2 for exact).

`./scripts/check-all.sh`: formatting/lint/refurb/mypy-strict/complexity all green; the only failing unit tests are the pre-existing author/compost/mcp environmental cluster (operator's live Ollama + creds), proven identical on clean `origin/main` — they pass on CI.

## Scope fence
Resonance-discovery path only — no change to embedding generation, the cache format, or what a resonance means. Local-only; no new dependency.

Refs #715
Closes #722

### DIFF
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -294,6 +294,19 @@ class EmbeddingsConfig(BaseModel):
     similarity_threshold: float = 0.75
     """Minimum cosine similarity for linking fragments."""
 
+    resonance_method: Literal["exact", "topk"] = "exact"
+    """Resonance-discovery strategy (#722).
+
+    ``exact`` (default) emits every above-``similarity_threshold`` pair — the
+    validated, exhaustive path. ``topk`` keeps only each fragment's
+    ``resonance_top_k`` most-similar above-threshold neighbours, bounding the
+    edge set to O(n·k) for large vaults; its output is always a subset of the
+    exact path's edges.
+    """
+
+    resonance_top_k: int = Field(default=10, ge=1)
+    """Per-fragment neighbour cap for ``resonance_method = "topk"`` (#722)."""
+
     cache_dir: str | None = None
     """Local directory for caching downloaded models."""
 
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -295,7 +295,7 @@ class EmbeddingsConfig(BaseModel):
     """Minimum cosine similarity for linking fragments."""
 
     resonance_method: Literal["exact", "topk"] = "exact"
-    """Resonance-discovery strategy (#722).
+    """Resonance-discovery strategy.
 
     ``exact`` (default) emits every above-``similarity_threshold`` pair — the
     validated, exhaustive path. ``topk`` keeps only each fragment's
@@ -305,7 +305,7 @@ class EmbeddingsConfig(BaseModel):
     """
 
     resonance_top_k: int = Field(default=10, ge=1)
-    """Per-fragment neighbour cap for ``resonance_method = "topk"`` (#722)."""
+    """Per-fragment neighbour cap when ``resonance_method = "topk"``."""
 
     cache_dir: str | None = None
     """Local directory for caching downloaded models."""
--- a/creek-tools/creek/link/embeddings.py
+++ b/creek-tools/creek/link/embeddings.py
@@ -757,8 +757,8 @@ class EmbeddingLinker:
 
         # ``exact`` (default) emits every above-threshold pair; ``topk`` keeps
         # only each fragment's k best neighbours. Both compute similarities one
-        # row-block at a time via vectorized matmul (never the full NxN matrix;
-        # OPS-004 / #596) and emit pairs in ascending ``(i, j)`` index order.
+        # row-block at a time via vectorized matmul (never the full NxN matrix)
+        # and emit pairs in ascending ``(i, j)`` index order.
         if self.config.resonance_method == "topk":
             resonances, suppressed = self._resonances_topk(
                 ids=ids,
--- a/creek-tools/creek/link/embeddings.py
+++ b/creek-tools/creek/link/embeddings.py
@@ -28,6 +28,8 @@ if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
     from pathlib import Path
 
+    import numpy as np
+    from numpy.typing import NDArray
     from sentence_transformers import SentenceTransformer as SentenceTransformerType
 
     from creek.config import EmbeddingsConfig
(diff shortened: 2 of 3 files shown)


Geoffe-Ga:
Addressed the non-blocking COMMENTS in `6ed3c57` (plus the earlier empty commit was only to re-trigger the review bot, which had failed to post a verdict twice — squashed away on merge):

- **Hierarchy suppression now tested in the top-k path** — `test_topk_suppresses_hierarchy_trivial_pairs` exercises the `suppressed` branch in `_resonances_topk` (a parent/child pair at cosine 1.0 is dropped).
- **O(n·k) bound is now CI-visible** — added `test_topk_edge_count_is_bounded_small` (n=60, k=3, no `slow` marker) so the core guarantee runs in the default suite; the large `@pytest.mark.slow` test stays for scale.
- **Tie-breaking documented** — `_top_k_neighbours` now notes that `argpartition` is not stable, so which tied neighbour survives the cap is implementation-defined.
- **Scrubbed `(#722)` from code** — removed the issue refs from the config docstrings + the `find_resonances` comment (CLAUDE.md).
- **Tightened** the test helper return annotation to `list[tuple[str, str]]`.

Deferred item #5 (warn when `resonance_top_k` is set under `exact`) as the explicitly-optional followup-issue nit. @claude please re-review.

Geoffe-Ga:
Addressed the round-2 nits in `2bf7355`:

- **Finished the `#722` scrub** — removed it from `test_topk_resonance.py`'s module docstring (the prior commit cleaned config.py + embeddings.py but missed the test file). `grep -rn '#722' creek/ tests/test_topk_resonance.py` is now clean.
- **Dropped the inherited `OPS-004 / #596` ref** from the `find_resonances` routing comment I rewrote.
- **Fixed the "every pair qualifies" comments** in both bound tests — `standard_normal` vectors produce negative cosines, so a 0.0 threshold admits only non-negatively-similar pairs (the bound holds either way); the comments now say so.
- **Documented the `slow` marker** semantics on the scale test (the small n=60 test is the always-on CI gate; the marker is descriptive, not auto-deselected).

Note for context: the review bot posted three verdicts on this PR (COMMENTS → LGTM → COMMENTS) across runs after an earlier infra hiccup where two runs completed without posting at all; all three flagged only non-blocking items. @claude please re-review.

Issue #736, opened 2026-06-29
test(fill): verify creek fill idempotency on a real vault + quality-aware overwrite dialog


Geoffe-Ga:
## Role
Senior Python engineer in this repo's `creek/cli.py` (`fill`, `_build_fill_steps`,
`_format_fill_summary`) and `creek/classify/` (`classify_engine.py`,
`constants.py`, the `ModelRouter` chokepoint).

## Goal
Two things, in one PR:
1. **Prove `creek fill` is genuinely idempotent on a real, un-mocked vault** — a
   second `creek fill` must change nothing on disk (byte-stable across every
   populated folder). Today's tests mock every step, so orchestration is covered
   but real end-to-end byte-stability is not.
2. **Add a quality-aware overwrite dialog** to `fill`: when an artifact already
   exists but a *higher-quality generation method* is now available/configured
   than the one that produced it, `fill` surfaces that and offers to regenerate
   (overwrite) rather than silently treating the existing artifact as final. The
   motivating ladder is classification fidelity:
   **rules (offline, lowest) < local LLM (qwen3 via Ollama) < Haiku/cloud
   (highest, per-tier)**. The non-interactive default must remain a safe no-op:
   never silently overwrite, never silently egress to a cloud provider.

## Context
- `creek fill` was added in #720 / merged PR #734. Its plan is built by
  `_build_fill_steps` (`creek/cli.py:2198`): link stages
  (embeddings → temporal → eddies → threads), the deterministic reports
  (decisions, unnamed, paradox, synchronicity, mode-profiles, wavelength), then
  index regeneration; `--with-compost` appends the compost overview. The
  command's own docstring claims "Every underlying step is idempotent, so a
  second `creek fill` is a clean no-op" — that claim is currently **unverified
  end-to-end**. **Classification and embedding are preconditions, not run by
  `fill`** — `fill` contains no classify step today.
- **Idempotency tests today are fully mocked.** `tests/test_cli_fill.py`
  (`_install_recorders`) monkeypatches `run_link`, every `_report_*`,
  `IndexGenerator`, and `CompostTracker` with label-recording fakes. They assert
  dependency order, non-fatal failure handling, dry-run, the compost append, and
  the summary line — but **nothing writes real bytes**, so a second `fill` being
  byte-stable is unproven.
- **Provenance already exists, partially.** `creek/classify/constants.py` defines
  `CLASSIFICATION_METHOD_KEY = "classification_method"` carrying
[...]

Issue #720, opened 2026-06-28
feat(cli): `creek fill` umbrella for the deterministic fill sequence


Geoffe-Ga:
## Role
Senior Python engineer in this repo's `creek/cli.py`.

## Goal
Add a **`creek fill`** umbrella command that runs the deterministic
vault-population sequence in dependency order, idempotently, on an
already-classified + already-embedded vault — so an operator gets a prod-ready
vault from one command instead of memorising six.

## Context
- Parent epic: #713. **This is the capstone — it lands LAST**, after
  the per-generator issues (#711, EPIC_01_ISSUE_03/04/05) are merged so every
  step it calls is real.
- The sequence (validated by hand on a 35k vault): `link --method embeddings`
  (resonances + thread links) → `link --method temporal` → `link --method eddies`
  → bulk thread-pages → `report --type decisions` → `report --type unnamed` →
  `report --type paradox` → `report --type synchronicity` → `report --type
  mode-profiles` → `report --type wavelength` → `index`.
- Each step is deterministic + local (only Compost, if included, touches an LLM —
  keep Compost opt-in via a flag, off by default).

## Output Format
A single PR:
- [ ] `creek fill --vault <path>` runs the sequence in order; each step is
  non-fatal (a step failure logs + continues so one bad step doesn't abort the
  rest) and the command prints a final per-folder population summary.
- [ ] `--dry-run` prints the plan without running. A `--with-compost` flag opts
  into the LLM compost pass (off by default).
- [ ] Idempotent: a second `creek fill` is a clean no-op across all folders.
- [ ] Tests: orchestration calls each underlying step (mocked) in order; a step
  failure does not abort the rest; `--dry-run` runs nothing; the summary reports
  real folder counts.

## Examples
```text
$ creek fill --vault V
[fill] link/embeddings … link/eddies … report/decisions … report/paradox … index
[fill] done. 02-Threads 184 | 03-Eddies 57 | 08-Decisions 22 | 10-Liminal 96
```

## Constraints
**Scope fence:** ORCHESTRATION only — call the existing step commands/functions;
implement no generator logic here. Respect dependencies (link before
threads/eddies; classify+embed are preconditions, not run by `fill`).

**Anti-bypass (verbatim, non-negotiable):**
> No `noqa`, `# type: ignore`, `pylint: disable`, `eslint-disable`, or equivalent
> linter/type-checker silencers. Fix the root cause. The only exception is the
[...]

PR #742, opened 2026-06-29
fix(fill): prove creek fill idempotency on a real vault + fix unnamed-digest non-idempotency (#736)

Geoffe-Ga:
## Summary
Delivers **#736's headline deliverable**: a real-vault, **un-mocked** byte-stability audit of `creek fill`. `test_cli_fill.py` mocks every step, so the docstring claim *"a second `creek fill` is a clean no-op"* was never proven end-to-end. The new audit builds a real scaffolded vault with classified fragments, runs the actual pipeline steps twice, and asserts every markdown note is byte-for-byte identical on the second run.

**The audit surfaced a genuine non-idempotency** in the unnamed weekly digest — a second fill rewrote `10-Liminal/Unnamed/Digests/<week>.md` and its history log. Three causes, all fixed in `creek/generate/unnamed.py`:

1. **Digest `generated` stamp** was a fresh wall-clock timestamp every run → `_stable_digest_generated` now preserves the prior stamp when the body + week bounds are unchanged.
2. **Self-referential growth section** — the digest read the unnamed-history log the *same run* had just updated, so run 2 saw its own entry as the "previous week". `_load_history` now takes the current week and returns only a strictly-earlier one.
3. **History log** appended a duplicate row each run and stamped a fresh per-row `generated_at` → now upserts by week and preserves `generated_at` when the week's data is unchanged. Extracted `_read_history_entries` + `_upsert_history_entry` (DRYs the load/append paths and keeps `_append_history` under the complexity gate).

## Test plan
- `tests/test_cli_fill_idempotency.py::test_second_creek_fill_is_byte_stable` — real vault, `creek fill` twice, asserts every `.md` note is byte-identical (sha256 snapshot). This is the headline.
- `tests/test_unnamed_digest.py::test_regenerating_same_week_is_byte_idempotent` — generator-level lock: the digest file and the history JSON are both byte-stable on a re-run, and the history keeps exactly one row per week.
- All 40 existing unnamed-digest tests stay green.

`./scripts/check-all.sh`: formatting/lint/mypy-strict/complexity/refurb/tryceratops all green; the only failing unit tests are the pre-existing author/compost/mcp environmental cluster (operator's live Ollama + creds), identical on clean `origin/main`; they pass on CI.

## Scope
This is **deliverable 1 of #736** (the idempotency audit + the bug it surfaced). **Deliverable 2 — the quality-aware overwrite dialog** (method-fidelity ladder, provenance extension, `fill` overwrite decision + `--upgrade` flag, Intimate-never-cloud preserved) — remains #736's tracked scope; see the [split note on #736](https://github.com/Geoffe-Ga/Creek-Vault/issues/736). Refs (does not close) #736.

Refs #720
Refs #736

### DIFF
--- a/creek-tools/creek/generate/unnamed.py
+++ b/creek-tools/creek/generate/unnamed.py
@@ -518,8 +518,7 @@ def _upsert_history_entry(
         prior.get(key) == new_entry.get(key) for key in _HISTORY_DATA_KEYS
     ):
         # Copy rather than mutate the caller's dict (no surprising side-effect).
-        new_entry = {
-            **new_entry,
+        new_entry = new_entry | {
             "generated_at": prior.get("generated_at", new_entry["generated_at"]),
         }
     merged = [e for e in entries if e.get("week_start") != week]
--- a/creek-tools/creek/generate/unnamed.py
+++ b/creek-tools/creek/generate/unnamed.py
@@ -43,7 +43,6 @@ from creek.models import Fragment
 from creek.time import effective_authored_at, effective_authored_date
 
 if TYPE_CHECKING:
-    from datetime import date
     from pathlib import Path
 
     from creek.link.embeddings import EmbeddingLinker
@@ -83,6 +82,10 @@ class _HistoryEntry:
         Returns:
             Dictionary with all history entry fields.
         """
+        # ``generated_at`` is stamped live here (not stored on the dataclass) so
+        # a genuinely new/changed week records when it was written; an unchanged
+        # week's stamp is preserved by ``_upsert_history_entry`` to stay
+        # byte-idempotent.
         return {
             "week_start": self.week_start,
             "fragment_count": self.fragment_count,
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
Addressed the non-blocking COMMENTS in `4859ec7`:
- **Duplicate `date` import** — removed the dead `from datetime import date` in the `TYPE_CHECKING` block (it's imported unconditionally for runtime use).
- **`_upsert_history_entry` mutation** — now builds a copy (`{**new_entry, "generated_at": ...}`) instead of mutating the caller's dict.
- **`_HistoryEntry.to_dict` live stamp** — added a comment explaining the deliberate live `generated_at` (idempotency handled in `_upsert_history_entry`).
- **`_stable_digest_generated` direct tests** — added four: no prior file, unchanged content preserves the stamp, changed body re-stamps, and an undecodable prior digest is treated as absent (no crash).

Deferred the `_scaffold` → `creek init` suggestion (you flagged it as a long-term follow-up; the manual scaffold is fine for now). @claude please re-review.

Geoffe-Ga:
CI's refurb (newer than my local v2.3.1) flagged `FURB173` on the `{**new_entry, ...}` copy from the last commit — modernised to `new_entry | {...}` (same semantics). Fixed in `edc3c0f`; the first push was green and this restores it.

Issue #744, opened 2026-06-30
fix(classify): failed LLM call is falsely stamped `classification_method: llm`


Geoffe-Ga:
## Role
Senior Python engineer in `creek/classify/classify_engine.py` (the per-fragment
write path) and `creek/classify/llm/orchestrator.py` (the retry loop).

## Goal
Stop `creek classify --method llm` from stamping `classification_method: llm`
(with a fresh `classified_at`) on fragments whose LLM call **failed** and was
returned unchanged. A failed classification must leave the fragment honestly
unclassified (or fall back to the rule result), never wear an `llm` label.

## Context (reproduced on a real vault)
A real `creek classify --method llm` run on `creek-demo-2026-06-07` hit
`Anthropic API call failed: BadRequestError` on **every** fragment (the account
was over budget — Anthropic returns a 400 `invalid_request_error` for "credit
balance too low"). The retry loop exhausted all 3 attempts per fragment and
logged `All 3 retries exhausted … returning unchanged`. Yet **33,351 fragments
on disk now read `classification_method: llm`** while their classification is
still empty. Verified on `frag-32b5e52317f5`
(`01-Fragments/Messages/2026-01-11-messages-5.md`):

```yaml
classification_method: llm
classified_at: '2026-06-29T18:42:03.850337-07:00'   # this failing run
frequency:
  primary: unclassified                              # never actually classified
  phase: unclassified
```

Root cause: `_finalise_fragment_write` derives
`write_method = RULES_METHOD if was_skipped else method`. `was_skipped` covers
the *confidence short-circuit* (rules were good enough), **not** an LLM
**failure**. When the LLM call errors and the orchestrator returns the fragment
unchanged, `was_skipped` is `False`, so `write_method` stays `llm` and the
fragment is stamped as if the LLM succeeded.

Practical harm: the resume-skip and the new `creek fill --upgrade` (#736) both
treat `llm` as final, so after the operator tops up their budget a plain
`classify --method llm` (no `--force`) will **skip** all 33,351 falsely-stamped
fragments — they can never be classified without `--force` or a manual reset.
This is also a "lying flag" of exactly the kind the #732 de-slop audit hunts.

## Output Format
A single PR:
- [ ] The orchestrator/engine distinguishes a **failed/unchanged** LLM result
      from a successful one (e.g. surface a `failed`/`changed` signal from the
      retry loop, or detect "returned unchanged after exhausted retries").
[...]

PR #747, opened 2026-06-30
fix(classify): surface the LLM provider's error cause, not a bare type (#745)

### LINKED ISSUE #745: fix(classify): surface the LLM provider's error body, not a bare `BadRequestError`
## Role
Senior Python engineer in `creek/classify/llm/orchestrator.py` and the provider
wrappers in `creek/classify/llm/providers.py`.

## Goal
When an LLM classification call fails, log the **provider's actual error
message** (e.g. Anthropic's "Your credit balance is too low …") instead of the
bare exception class name. Today the operator sees only `BadRequestError` and
cannot tell a budget/credit problem from a bad model id, a context-length
overflow, or a transient network error.

## Context (reproduced on a real vault)
A real `creek classify --method llm` run logged, for ~18,690 fragments:

```text
[fragment=… provider=anthropic] Classify attempt 1/3 failed for 'messages':
    Anthropic API call failed: BadRequestError
…
[fragment=… provider=anthropic] All 3 retries exhausted for 'messages' — returning unchanged
```

The underlying cause was an **over-budget 400** (Anthropic surfaces "credit
balance too low" as a `BadRequestError`/`invalid_request_error`), but the message
body — which says exactly that — was dropped. The operator spent a long run and a
debugging session before identifying it as a billing issue rather than a code
bug. The retry log keys off the exception **type**, not its message/`str(exc)`.

## Output Format
A single PR:
- [ ] The failure log includes the provider error's message/body
      (`str(exc)` and, where the SDK exposes it, the structured
      `error.type` / `error.message`), truncated to a sane length — not just the
      class name. Applies to the Anthropic, OpenAI, and Gemini wrappers
      symmetrically.
- [ ] The final "retries exhausted" line names the last error reason so a tail
      of the log is self-explanatory.
- [ ] No secrets/PII leak: log the API error message, never the request body /
      fragment content / API key.

## Examples
```text
# Today:
Classify attempt 1/3 failed for 'messages': Anthropic API call failed: BadRequestError
# Wanted:
Classify attempt 1/3 failed for 'messages': Anthropic 400 invalid_request_error:
    "Your credit balance is too low to access the Anthropic API …"
```

## Constraints
**Scope fence:** logging/observability of the failure only. Do not change the
retry count, the backoff, or the return-unchanged behaviour (the #<744> stamp
fix is separate). Keep the messages provider-neutral and secret-safe.

> linter/type-checker silencers. Fix the root cause. The only exception is the
> documented 4-line escape hatch (third-party library bug / language-version
> compatibility / benchmarked performance necessity / generated code) — with the
> reason, a reference URL, an alternative considered, and a review date.

- [ ] `./scripts/check-all.sh` green (exit 0).
- [ ] Test: a provider raising an error with a known message has that message
      appear in the failure log (caplog), and the API key / request body do not.
- [ ] PR body includes `Closes #<this-issue>`.
- [ ] Latest Claude `Verdict:` on HEAD is `LGTM`.

Geoffe-Ga:
## Summary

A failed cloud classification wrapped the SDK error as a `RuntimeError` carrying only the exception **type** — `BadRequestError` — which hid the real cause. Anthropic delivers an over-budget failure as a **400** whose message is *"Your credit balance is too low to access the Anthropic API…"*, identical in class to a malformed-request 400. So a whole vault's worth of budget failures read as an opaque wall of `BadRequestError` with no hint that the fix was *billing*.

This is the companion fix to #744 (which stopped stamping `llm` on a failed call). Together: a budget failure now both **names its cause** and **stays eligible for re-classify**.

## What changed

`creek/classify/llm/providers.py` — `_error_detail(exc)` now appends a redaction-safe detail keyed on HTTP status, and the three cloud `complete()` error blocks (Anthropic / OpenAI / Gemini) use it and re-raise `from exc`:

| Status | Surfaced |
|---|---|
| **4xx** client error | `(HTTP <s>): <API message>` — the API-side description of what's wrong with the request or account (billing, auth, model id, params), whitespace-collapsed and capped at 500 chars |
| **5xx / unknown** | status-only (`(HTTP <s>)`) or nothing — server bodies are opaque and stay redacted |

### Why the 4xx/5xx split (and not "show everything")

The providers **deliberately redact** the SDK error body (`test_complete_redacts_sdk_errors`) because a response body *can* echo request state. The split preserves that intent:

- A **4xx** `message` is an API-side description of the request/account — it is not your prompt or fragment content, and the API key lives only in request headers, never in the error message.
- A **5xx** body is opaque server internals → stays redacted.

This also keeps the existing redaction tests green: OpenAI's injects a status-less base error (→ status-only) and Gemini's injects a **503** (→ status-only), so neither surfaces its `"secret request payload"` probe.

## Example

Before:
```
Anthropic API call failed: BadRequestError
```
After:
```
Anthropic API call failed: BadRequestError (HTTP 400): Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits.
```

## Test plan

- New `tests/test_provider_error_detail.py` (6 cases): 4xx surfaces the API message; Gemini's `code` attribute path; 5xx withholds the message (status-only); unknown status → empty (type-only) form; over-long message capped with `…`; non-string `message` → status-only.
- All existing provider/classify redaction tests pass unchanged (248 passed locally across `test_provider_error_detail`, `test_gemini_provider`, `test_openai_provider`, `test_classify`).
- `ruff` + `mypy --strict` clean.

> Note: 8 author-desk/mcp/compost tests fail **only** on the author's local machine (a pre-existing local-environment artifact — verified identical at clean `origin/main` HEAD `1088d88`, whose CI/CD Pipeline run is green). They are independent of this change and pass in CI.

Closes #745

### DIFF
--- a/creek-tools/creek/classify/llm/providers.py
+++ b/creek-tools/creek/classify/llm/providers.py
@@ -206,6 +206,27 @@ def _error_detail(exc: Exception) -> str:
     return f" (HTTP {status}): {collapsed}"
 
 
+def _chain_is_safe(exc: Exception) -> bool:
+    """Whether the SDK error may remain as the wrapped ``RuntimeError.__cause__``.
+
+    True only for 4xx client errors — the same tier whose ``message``
+    :func:`_error_detail` already surfaces, so the chain exposes nothing the
+    message does not. For 5xx / unknown errors the body is deliberately
+    withheld, so the caller suppresses the chain (``raise … from None``) to keep
+    that redaction airtight even if a future caller logs with
+    ``logging.exception`` / ``exc_info=True`` (which would otherwise render
+    ``__cause__``).
+
+    Args:
+        exc: The vendor SDK exception.
+
+    Returns:
+        ``True`` when *exc* carries a 4xx status, else ``False``.
+    """
+    status = _error_status(exc)
+    return status is not None and _CLIENT_ERROR_MIN <= status < _SERVER_ERROR_MIN
+
+
 # Per-provider default model IDs. This matrix is the ONLY place model literals
 # live in the package — provider classes read their ``DEFAULT_MODEL`` from it
 # and other modules read the resolved value, never a literal (enforced by
@@ -346,9 +367,11 @@ class AnthropicProvider:
             The concatenated text content of the API response.
 
         Raises:
-            RuntimeError: If the SDK raises any ``AnthropicError``; the
-                original exception is re-raised as ``RuntimeError`` with
-                its type name to avoid leaking sensitive request state.
+            RuntimeError: If the SDK raises any ``AnthropicError``; re-raised
+                as ``RuntimeError`` with the exception type name and, for 4xx
+                client errors, the API's own error message (see
+                :func:`_error_detail`). 5xx / unknown bodies are withheld and
+                their exception chain suppressed to avoid leaking request state.
         """
         return self.call_with_metadata(prompt).text
 
@@ -416,9 +439,11 @@ class AnthropicProvider:
             and the SDK's token :attr:`~AnthropicCompletion.usage` when present.
 
         Raises:
-            RuntimeError: If the SDK raises any ``AnthropicError``; the
-                original exception is re-raised as ``RuntimeError`` with
-                its type name to avoid leaking sensitive request state.
+            RuntimeError: If the SDK raises any ``AnthropicError``; re-raised
+                as ``RuntimeError`` with the exception type name and, for 4xx
+                client errors, the API's own error message (see
+                :func:`_error_detail`). 5xx / unknown bodies are withheld and
+                their exception chain suppressed to avoid leaking request state.
         """
         import anthropic
 
@@ -433,7 +458,7 @@ class AnthropicProvider:
         except anthropic.AnthropicError as exc:
             detail = _error_detail(exc)
             msg = f"Anthropic API call failed: {type(exc).__name__}{detail}"
-            raise RuntimeError(msg) from exc
+            raise RuntimeError(msg) from (exc if _chain_is_safe(exc) else None)
         stop_reason = getattr(response, "stop_reason", None) or "end_turn"
         return AnthropicCompletion(
             text=_extract_anthropic_text(response),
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
Thanks for the thorough review. Addressed all three non-blocking items in `3fd265c` rather than deferring (two of them tighten the security guarantee this PR is about):

**1. `__cause__` chain leak (LOW security).** Good catch — `from exc` did make the withheld 5xx/unknown body reachable via `RuntimeError.__cause__` under a future `logging.exception`. Rather than just document it, I made the chain track the redaction tier: new `_chain_is_safe(exc)` returns `True` only for 4xx (whose message we already surface), so those keep the chain; 5xx/unknown now use `raise … from None`. The guarantee is now airtight even under `exc_info=True`. Covered by `test_call_5xx_withholds_body_and_suppresses_chain` (asserts `__cause__ is None`) and `_chain_is_safe` unit tests.

**2. Stale `Raises:` docstrings.** Updated all four (Anthropic ×2, OpenAI, Gemini) to describe the 4xx-message / 5xx-redacted behavior.

**3. Missing integration test for the motivating case.** Added `test_call_4xx_surfaces_api_message_not_just_type` — a real `anthropic.BadRequestError(400, "Your credit balance is too low…")` through `AnthropicProvider.call()`, asserting the message + `HTTP 400` surface and the key never does. Closes the loop on the exact billing scenario.

Local: 5783 passed, ruff + mypy clean. (The 8 author-desk/mcp failures are the pre-existing local-environment artifact noted in the PR body — green in CI.)

Geoffe-Ga:
Round-2 follow-ups addressed in `6191c85` (both flagged non-blocking):

- **Parallel OpenAI/Gemini integration tests** — added `test_complete_4xx_surfaces_api_message` + `test_complete_5xx_withholds_body_and_suppresses_chain` to both `test_openai_provider.py` and `test_gemini_provider.py`, mirroring the Anthropic pair. The #745 behavior (4xx surfaces the API message; 5xx withholds body + `__cause__ is None`) is now locked in for all three backends.
- **Gemini `.message` question** — verified empirically: `errors.APIError(400, {"error": {"message": …}})` exposes both `.message` and `.code`, and `_error_detail` returns `(HTTP 400): <message>`. So the 4xx-surfacing fires correctly for Gemini; the new `test_complete_4xx_surfaces_api_message` asserts it end-to-end.

Local: 255 provider/classify tests pass, ruff + refurb + mypy clean.

Issue #753, opened 2026-06-30
feat(mcp): care guardrail in the reflect/voice path (no med advice; crisis → human)


Geoffe-Ga:
## Goal

Ensure Creek's voice-generating tools never advise on psychiatric medication and never substitute for human help in acute distress — the care boundary Adepthood depends on (NORTH-STAR §10; mirrors adepthood#887/#890).

## Context

- Voice path: `creek.reflect` (this epic), `creek.draft`, `creek.author`, `creek.compile` — wherever the LLM writes in the user's voice.
- Adepthood runs an acute-distress pre-check before calling out (`geoffe-ga/adepthood#888`/#889), but Creek must also be safe by construction since other consumers (CrawDad) call the same tools.

## Acceptance Criteria

- [ ] A shared guardrail (one constant/policy) is applied to every voice-generating prompt: never advise reducing/stopping/changing medication; defer that to the user and their prescriber.
- [ ] When generated content would touch acute distress, the tool surfaces a pointer to human + professional support rather than a chatbot-only reply (return a structured care signal the client can render); it never dead-ends.
- [ ] Anti-guru tone preserved; no behavior change for ordinary content.
- [ ] Detection (if any) targets *acute* distress and does not pathologize ordinary dark/reflective writing.

## Constraints / DoD

stay-green TDD asserting the guardrail text/policy is present in the built prompts and that a distress fixture yields the care signal, not advice. `pre-commit` clean; coverage met. PR: `Refs #748`, `Closes #<this>`.

Refs #748
Refs geoffe-ga/adepthood#887

Issue #748, opened 2026-06-30
epic: Adepthood MCP handshake — expose the vault to the Adepthood app


Geoffe-Ga:
## Epic Summary

[`geoffe-ga/adepthood`](https://github.com/geoffe-ga/adepthood) (a journal-first PKM app) will use Creek as its private vault: the corpus store, the ontology classification (frequency + archetypal phase), and the Higher Self voice. Most of what Adepthood needs **already exists** in `creek-tools-mcp` (FEAT-010/011): `ingest`, `classify`, `save`, `link`, `report`, `draft`/`author`/`mine`, the `TierCeiling` (OPEN/PERSONAL/INTIMATE) privacy enforcement on every tool, local-first Ollama classification, and the hash-chained audit. This epic does **not** rebuild any of that — it adds the **handshake surface** Adepthood calls and the few capabilities it needs that don't exist yet.

The contract is owned jointly: Adepthood publishes its required side in `geoffe-ga/adepthood#950`; this epic implements Creek's side and keeps the two in sync. The cross-repo boundary is recorded in `geoffe-ga/adepthood#927`.

## Grounded current state

- MCP server: `creek-tools/creek_mcp/server.py` (`build_server`, `FastMCP`, **stdio** transport, `SERVER_NAME = "creek-tools-mcp"`). Per-consumer identity via `CREEK_MCP_CONSUMER`; every tool takes a `privacy_tier_ceiling`.
- Tier model: `creek-tools/creek_mcp/tier_ceiling.py` (`TierCeiling` OPEN/PERSONAL/INTIMATE) — already the privacy gate.
- Voice/generation today is essay-shaped (`creek.draft`, `creek.author`, `creek.mine`); there is **no** single-entry "margin notes" reflection tool.
- Transport is **stdio** with local consumers (CrawDad, Claude Code); a remote Adepthood backend/phone cannot reach it — the one architectural fork.

## Child Issues

- [x] #755 (decision — ratified & implemented; closed) Remote transport + consumer auth for a remote Adepthood client
- [x] #749 (docs) Versioned Adepthood↔Creek contract mirroring adepthood#950
- [x] #750 `creek.handshake` — capabilities + contract/ontology version + tier model
- [x] #751 `creek.reflect` — anchored Higher-Self margin notes on a single entry
- [x] #752 `creek.wheel` — per-frequency balance read for the Map
- [x] #753 Care guardrail in the reflect/voice path (no med advice; crisis → human)
[...]

PR #765, opened 2026-06-30
feat(mcp): creek.reflect — anchored Higher-Self margin notes (#751)

### LINKED ISSUE #751: feat(mcp): creek.reflect — anchored Higher-Self margin notes on a single entry
## Goal

Add a `creek.reflect` MCP tool that takes one journal entry and returns **anchored Higher-Self margin notes** (and optionally a short essay), grounded in retrieval over the user's own corpus — the reflection surface Adepthood's journal needs. This is distinct from the essay-shaped `draft`/`author`/`mine`.

## Context

- `creek-tools/creek_mcp/server.py` — register alongside `creek.draft`/`creek.author`; reuse the LLM factory pattern (`_build_draft_llm`, lazy so the server boots without a provider) and the `TierCeiling` gate.
- Voice grounding: the Voice Skill Tree + `link`/embeddings retrieval already in the pipeline (`07-Voice/`, `creek.link`).
- Contract: `creek.reflect` should return verbatim quote spans so the **client re-anchors** them to character offsets itself (model-supplied indices are never trusted) — mirror adepthood's `domain/resonance.py` anchoring contract in `geoffe-ga/adepthood#953`.

## Acceptance Criteria

- [ ] `creek.reflect({ content | entry_ref, privacy_tier_ceiling })` returns `{ notes: [{ quote, kind, note }], essay? }` where each `quote` is a verbatim substring of the input and `note` is warm, second-person, anti-guru (mirrors the user's own wisdom, never advises from above).
- [ ] Retrieval is grounded in the user's corpus (uses linked/related fragments), not generic advice.
- [ ] Honors `TierCeiling`: INTIMATE content is processed locally only and never egressed to a cloud model.
- [ ] Local-first: works on Ollama; degrades cleanly when no provider is available (clear error, server still boots).
- [ ] Audit-logged; read-only with respect to the corpus (does not create fragments — that's the ingest path).

## Constraints / DoD

stay-green TDD with an injected LLM (no live calls in tests). Honor the care boundary (coordinate with the care-guardrail issue). `pre-commit` clean; coverage threshold met; no bypasses. PR: `Refs #748`, `Closes #<this>`.

Refs #748
Refs geoffe-ga/adepthood#953

Geoffe-Ga:
## Summary

Adds a read-only **`creek.reflect`** MCP tool — the reflection surface Adepthood's journal needs. It takes one journal entry (raw `content` or an `entry_ref` fragment id) + a privacy-tier ceiling and returns `{notes: [{quote, kind, note}], essay?}`: warm, second-person, anti-guru notes that mirror the user's own wisdom, grounded in retrieval over their corpus. Distinct from the essay-shaped `draft`/`author`/`mine`.

Part of epic #748.

## Three guarantees (the security-critical core)

1. **INTIMATE never egresses.** The LLM is obtained from a *tier-keyed factory* `llm_factory(tier)`. The production factory (`_build_reflect_llm_factory`) resolves the `generation` stage through the config's `ModelRouter`, whose `_enforce_local_for_intimate` chokepoint forces an INTIMATE call onto the local `default` — or raises `IntimateRoutingError`, which the tool turns into a structured refusal rather than a crash or a cloud call. The tool **derives the routing tier from the ceiling, failing closed to INTIMATE** (`ALL → INTIMATE`); it never hand-picks a provider or re-checks the tier. Tested with a recording factory asserting the exact tier per ceiling, plus the refusal path.
2. **Quotes are verbatim.** Every returned `quote` is validated as a whitespace-normalised substring of the entry; model-invented spans are dropped. So the client can re-anchor verbatim quotes to character offsets itself (mirrors `adepthood#953`) and a hallucinated span never reaches it.
3. **Care boundary (#753).** An injectable `care_guard` seam escalates (and skips the model) when it flags the entry — wired now; the real guard lands in #753 per "coordinate with the care-guardrail issue".

Also: read-only wrt the corpus, audit-logged like every tool, and degrades to a clean refusal when no provider is available (server still boots).

## Files

- `creek_mcp/tools/reflect.py` — the tool: tier derivation, ceiling→override for retrieval admission, verbatim-quote validation, fenced-JSON/YAML parse with safe degradation, care seam, audit, refusals.
- `creek_mcp/server.py` — registers `creek.reflect` + the lazy production tier-keyed factory; new `reflect_llm_factory` build_server param for injection.
- `creek_mcp/tools/__init__.py` — export.

## Test plan (TDD, no live calls)

- `tests/test_mcp_reflect.py` (16): tier-per-ceiling routing (OPEN/INTIMATE/ALL→INTIMATE), `IntimateRoutingError`→refused, verbatim drop vs keep + empty-on-all-hallucinated, note shape, grounding query+override, audit w/ consumer, care escalation skips the LLM, empty-content refusal, read-only, `entry_ref` fragment resolution, fenced-JSON + YAML-fallback + unparseable parsing.
- `tests/test_mcp_server.py`: `creek.reflect` in `EXPECTED_TOOLS`; `call_tool` end-to-end returning verbatim notes; `_build_reflect_llm_factory` route-then-degrade (router + provider stubbed).
- Coverage: `reflect.py` 89.8%, `server.py` 86.3% (both > 80% gate). ruff / refurb / mypy --strict clean.

Closes #751
Refs #748

### DIFF
--- a/creek-tools/creek_mcp/server.py
+++ b/creek-tools/creek_mcp/server.py
@@ -37,6 +37,7 @@ from creek_mcp.tools import (
     purge_source_tool,
     purge_vault_tool,
     redact_scan_tool,
+    reflect_tool,
     report_tool,
     save_tool,
     skills_refresh_tool,
@@ -47,6 +48,9 @@ from creek_mcp.tools.draft import draft_tool
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+
+    from creek.models import PrivacyTier
+    from creek_mcp.tools.reflect import _LLM, _LLMFactory
 
 SERVER_NAME = "creek-tools-mcp"
 
@@ -97,11 +101,40 @@ def _build_compile_llm() -> Callable[[str], str]:
     return _engine_default_llm(load_config().llm)
 
 
+def _build_reflect_llm_factory() -> _LLMFactory:
+    """Return a tier-keyed LLM factory for ``creek.reflect``.
+
+    The returned ``factory(tier)`` resolves the ``generation`` stage through the
+    config's :class:`~creek.classify.llm.router.ModelRouter` for *tier*, so an
+    INTIMATE reflection is forced onto the local ``default`` model (or the router
+    raises ``IntimateRoutingError`` rather than egressing). Lazy imports keep the
+    server bootable without a provider — only a ``creek.reflect`` call then fails,
+    surfacing as a structured refusal.
+    """
+    from creek.classify.llm.providers import build_provider
+
+    router = load_config().model_router
+
+    def _factory(tier: PrivacyTier) -> _LLM:
+        cfg = router.resolve("generation", tier)
+        provider = build_provider(cfg)
+        if not provider.available:
+            msg = (
+                "LLM provider unavailable for reflection. "
+                "Check Ollama or ANTHROPIC_API_KEY configuration."
+            )
+            raise RuntimeError(msg)
+        return lambda prompt: provider.complete(prompt).text
+
+    return _factory
+
+
 def build_server(
     *,
     vault_path: Path | None = None,
     draft_llm_factory: Callable[[], Callable[[str], str]] | None = None,
     compile_llm_factory: Callable[[], Callable[[str], str]] | None = None,
+    reflect_llm_factory: Callable[[], _LLMFactory] | None = None,
 ) -> FastMCP:
     """Construct a :class:`FastMCP` instance with all FEAT-010/011 tools.
 
@@ -115,12 +148,16 @@ def build_server(
         compile_llm_factory: Optional factory for the compile LLM.
             Invoked lazily so only ``creek.compile`` needs an LLM
             provider.
+        reflect_llm_factory: Optional thunk returning the tier-keyed LLM
+            factory for ``creek.reflect``. Invoked lazily per call so only
+            ``creek.reflect`` needs an LLM provider.
     """
     server: FastMCP = FastMCP(SERVER_NAME)
     vault = _resolve_vault(vault_path)
     consumer = _consumer_from_env()
     factory = draft_llm_factory or _build_draft_llm
     compile_factory = compile_llm_factory or _build_compile_llm
+    reflect_factory = reflect_llm_factory or _build_reflect_llm_factory
 
     @server.tool(name="creek.handshake")
     async def _handshake(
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
Excellent review — both blocking issues were real and are fixed in `55e485d`.

**1. Security: routing tier now follows the entry's actual classification.** You're right that deriving the tier from the ceiling let an `entry_ref` to an intimate fragment route to cloud under a low ceiling. Fixed exactly along the `_content_tier`/#661 pattern you cited:
- `_resolve_entry` now also reads the fragment's persisted `privacy_tier` (`_fragment_tier`, fail-closed to INTIMATE like `tier_of`).
- `_routing_tier(ceiling, entry_tier)` routes at the **more sensitive** of the two — an INTIMATE fragment routes local even under `ceiling="open"`.
- Raw inline `content` has no classification, so there the bound is the declared ceiling — now stated explicitly in the module docstring (guarantee #1 is accurate again).
- New test `test_entry_ref_intimate_fragment_routes_local_despite_open_ceiling` pins it (fragment frontmatter `privacy_tier: intimate` + `ceiling=open` → factory asked for `INTIMATE`).

**2. Crash: `_parse_notes` now catches `yaml.YAMLError`.** Correct — `YAMLError` isn't a `ValueError`, so malformed (non-JSON, non-YAML) output would have propagated across the MCP boundary. Now caught; `test_malformed_yaml_response_does_not_crash` uses a genuinely malformed flow mapping (the prior test's fixture was valid YAML, as you noted).

**Nits also addressed:** deleted the dead `_Retrieve`/`_CareGuard` strings; `entry_ref`-not-found now refuses with the distinct reason `"entry_ref not found"`. (The per-call `rglob` I've left as-is for now — noted as a scale follow-up.)

Local: `./scripts/check-all.sh` green except the 8 pre-existing author-desk env tests; reflect.py coverage 91.6%; ruff/refurb/mypy clean.

Geoffe-Ga:
Fixed in the latest push: narrowed `except (IntimateRoutingError, RuntimeError)` to `except RuntimeError` (the routing error is a `RuntimeError` subclass, so it's still caught) with a comment documenting that, and dropped the now-unused import.

One note for the record: this didn't actually trip `ruff` — `B014` only fires when the subclass relationship is statically visible, and `IntimateRoutingError` lives in another module, so `./scripts/check-all.sh` was green on it. But you're right that it was redundant, so it's gone. 39 tests pass; ruff/refurb/mypy clean.

Geoffe-Ga:
Thanks — the adversarial re-verification is reassuring. Addressed the two cheap non-blocking items in the latest push:

- **Unvalidated `essay`**: rather than only a docstring note, the response now carries a machine-readable `essay_grounded: false` field (documented in `Returns`), so the Adepthood UI can avoid rendering `essay` with the same trust as the verbatim-checked `notes`.
- **Missing PERSONAL test**: added `test_personal_ceiling_requests_the_personal_tier`, so a silent mis-mapping of the middle ceiling is now caught.

The remaining items I've left as the follow-ups you flagged: `care_guard` wiring is literally #753's scope, and the per-call `rglob` over `01-Fragments` is a perf follow-up (fine at current vault scale). 40 tests pass; ruff/refurb/mypy clean.

PR #767, opened 2026-06-30
feat(care): care guardrail across the voice path — no med advice; crisis → human (#753)

### LINKED ISSUE #753: feat(mcp): care guardrail in the reflect/voice path (no med advice; crisis → human)

Geoffe-Ga:
## Summary

Makes Creek's voice-generating tools safe **by construction** — not only when a careful client (Adepthood) runs its own pre-check — because other consumers (CrawDad) call the same surface. This is the care boundary `creek.reflect`'s seam (#751) was built to receive.

Part of epic #748. Mirrors `adepthood#887/#890`.

## The shared care module — `creek/care/guardrail.py`

Placed in `creek/` (not `creek_mcp/`) so the core generators import it without a dependency inversion.

- **`CARE_POLICY`** — prose woven into every voice prompt: never advise starting/stopping/reducing psychiatric medication (defer to the user + their prescriber); on acute distress, point to human + professional support rather than counsel as a chatbot; otherwise keep the ordinary non-directive voice and **do not pathologize** honest, dark, grieving, or reflective writing.
- **`acute_distress_guard`** — a deliberately **narrow, acute-only** detector. It matches a small curated set of first-person imminent self-harm / suicidal *intent* phrasings, and **never** affect/grief/reflection-on-death ("I feel hopeless", "thinking about death") or hyperbole ("the deadline is killing me"). Returns a stable reason code, never the matched text. **Biased toward false negatives** so reflective journaling is never flagged — Adepthood runs the primary pre-check; this is the safe-by-construction backstop.
- **`CARE_SIGNAL`** — the structured "reach a human" envelope returned in place of model prose; it never dead-ends.

## Wiring (all four voice prompts + the reflect detector)

- `CARE_POLICY` woven into: `reflect._build_prompt`, draft `_compose_prompt` + `_compose_bare_section`, author voice `static` (cached system prompt), compile `_build_prompt`. (No shared prompt builder exists, so four sites.)
- `acute_distress_guard` threaded into the registered `creek.reflect` (the live-journal-entry surface); the escalate response now carries `care_signal`. *(draft/author/compile generate from the already-classified corpus, not a fresh user signal, so the policy text is their appropriate protection; reflect is the surface that needs the acute escalation.)*

## Test plan (TDD)

- `tests/test_care_guardrail.py` (5): policy covers medication/prescriber + non-pathologizing; acute fires on imminent intent; **ordinary dark writing + hyperbole + empty → None** (the key safety property); care-signal shape.
- `tests/test_care_guardrail_weave.py` (3): `CARE_POLICY` present in the reflect, compile, and author (system) prompts; draft covered by an added assertion in `test_drafts.py`.
- `tests/test_mcp_reflect.py`: escalation carries the structured `care_signal` (never dead-ends).
- `tests/test_mcp_server.py`: the registered `creek.reflect` escalates on acute content with the care signal and never reaches the (injected) LLM — proves `server.py` threads the real guard.
- Coverage: `guardrail.py` 100%. ruff / refurb / mypy --strict clean. No import cycles.

> Note on scope: the detector is intentionally minimal (a backstop). The medication policy in every prompt is the primary, always-on guardrail; the acute detector adds the human-pointer escalation on the one surface that takes a raw live entry.

Closes #753
Refs #748

### DIFF
--- a/creek-tools/creek/author/voice.py
+++ b/creek-tools/creek/author/voice.py
@@ -25,6 +25,7 @@ from creek.author.skills import (
     find_voice_core,
     read_skill,
 )
+from creek.care.guardrail import CARE_POLICY
 from creek.generate.ontology_glossary import GLOSS_STEER
 
 if TYPE_CHECKING:
@@ -155,7 +156,9 @@ def _split_voice_prompt(
         (``""`` when the vault carries none); ``dynamic`` is the evidence block
         followed by the ask block.
     """
-    static = "\n\n".join(part for part in _skill_sections(vault, evidence) if part)
+    static = "\n\n".join(
+        [CARE_POLICY, *(part for part in _skill_sections(vault, evidence) if part)]
+    )
     dynamic = "\n\n".join(
         part
         for part in (_evidence_section(evidence), _ask_section(query, medium, contract))
--- a/creek-tools/creek/care/__init__.py
+++ b/creek-tools/creek/care/__init__.py
@@ -1,0 +1,11 @@
+"""Cross-cutting care-boundary policy for Creek's voice-generating tools (#753).
+
+Houses the shared medication/care policy woven into every voice prompt and the
+conservative acute-distress detector that backs the reflect care seam. Lives in
+``creek/`` (not ``creek_mcp/``) so the core generators (draft / author / compile)
+can import it without depending on the MCP layer.
+"""
+
+from creek.care.guardrail import CARE_POLICY, CARE_SIGNAL, acute_distress_guard
+
+__all__ = ["CARE_POLICY", "CARE_SIGNAL", "acute_distress_guard"]
--- a/creek-tools/creek/care/guardrail.py
+++ b/creek-tools/creek/care/guardrail.py
@@ -43,6 +43,11 @@ CARE_POLICY: Final[str] = (
 # ("the deadline is killing me"), so ordinary dark/reflective journaling does
 # not trip the guard. Biased toward false negatives by design — Adepthood runs
 # the primary acute pre-check; this is Creek's safe-by-construction backstop.
+#
+# This matches ONLY these literal patterns — it is NOT a comprehensive classifier.
+# Paraphrase ("gonna kill myself"), typos, and non-English phrasing slip through
+# by design; do not mistake this backstop for the primary defense. Broadening it
+# is fine, but the safety guarantee rests on the primary pre-check, not this set.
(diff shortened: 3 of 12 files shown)


Geoffe-Ga:
Thanks — the safety-posture read is exactly the intent. Addressed all four non-blocking items in `dc59e45`:

1. **Stale docstring** (`reflect.py:30`): refreshed — it no longer forward-references #753 as future; it now names the production `acute_distress_guard` threaded in by `build_server`.
2. **Top-level import**: moved `acute_distress_guard` to `server.py`'s module top (it's stdlib-light `re`, unlike the genuinely-heavy lazy SDK imports it sat beside — you were right that the lazy placement looked accidental).
3. **`_compose_bare_section` weave**: added `test_bare_section_prompt_weaves_care_policy`, so the one previously-unpinned weave site is now directly asserted.
4. **Detector scope comment**: added a comment on `_ACUTE_PATTERNS` making explicit it matches only the literal patterns — paraphrase/typos/non-English slip through by design; it's a backstop, not a classifier, and the guarantee rests on the primary pre-check.

184 affected tests pass; ruff/mypy clean; no import cycles. CI confirmed green on the prior push.

PR #768, opened 2026-07-01
feat(mcp): creek.journal — Adepthood entry → fragment via shared ingest pipeline (#754)

### LINKED ISSUE #754: feat(mcp): Adepthood-journal ingest path (entry → fragment)
## Goal

Let an Adepthood journal entry flow into the vault as a fragment (so it is classified, linked, and available to `reflect`/`wheel`), using the existing ingest/save machinery with a journal-appropriate source identity and tier.

## Context

- Existing entry points: `creek.ingest(source_type, input_path)` and `creek.save(target, body, tier, source_kind, source_id, ...)` in `creek-tools/creek_mcp/server.py`. The 11 ingestors live in `creek/ingest/`.
- Deterministic fragment IDs are hashed from `(source, timestamp, content)` — re-sending the same entry must be idempotent.
- Adepthood write path: `geoffe-ga/adepthood#952` (sends non-intimate entries; intimate stays on the adepthood side / local).

## Acceptance Criteria

- [ ] An Adepthood journal entry (content + timestamp + tier + a stable external id) can be ingested as a fragment with `source`/`source_kind` identifying it as the Adepthood journal.
- [ ] Re-sending the same entry is idempotent (no duplicate fragment); an edited entry updates in place or supersedes cleanly.
- [ ] Tier is honored end-to-end (an entry sent as INTIMATE is stored under intimate handling; INTIMATE never egresses downstream).
- [ ] Reuses existing ingest/save/classify rather than adding a parallel pipeline.

## Constraints / DoD

stay-green TDD. Decide and document whether this rides `creek.save`, `creek.ingest`, or a thin journal wrapper — least new surface wins. `pre-commit` clean; coverage met. PR: `Refs #748`, `Closes #<this>`.

Refs #748
Refs geoffe-ga/adepthood#952

Geoffe-Ga:
## Summary

Lets an Adepthood journal entry (content + timestamp + tier + a **stable external id**) flow into the vault as a fragment — classified, linked, available to `reflect`/`wheel` — by **reusing** the ledger-backed markdown ingest, not a parallel pipeline. This is the last drainable child of epic #748.

## Approach: lift a shared module (per the chosen design)

The idempotency + edit-in-place engine (ledger + `write_fragment_idempotent` + tomb machinery) lived only in `creek/cli.py`, which the MCP layer must not import. So:

- **New `creek/ingest/pipeline.py`** holds that machinery, lifted verbatim and decoupled — it raises plain exceptions and returns a structured `IngestRunResult`, no typer/console/MCP coupling. Single source of truth.
- **`creek/cli.py`'s `_run_ingest` is now a thin wrapper** (typer.Exit + summary print) delegating to `pipeline.run_ingest`. **Behavior is identical:** all 47 ledger/idempotency/tomb/incremental tests pass unchanged, and the `from creek.cli import _run_ingest` test seam is preserved.

## The tool

- **Idempotent + edit-in-place, for free.** The entry is staged as markdown at a stable path derived from the external id (`00-Creek-Meta/adepthood/journal/<id>.md`). The `journal/` segment makes the ingestor classify it as `SourcePlatform.JOURNAL`; the stable path is the ledger source-key. Re-sending the same external id is a no-op; an **edited** entry updates in place (same fragment id, no orphan). Verified end-to-end against a real vault, not mocked.
- **Tier honored end-to-end.** The tier rides the entry frontmatter onto the fragment; the write-side ceiling **refuses** an entry whose tier exceeds the caller's ceiling — an INTIMATE entry is stored intimate, never downgraded.
- Audit-logged; returns `{status, fragment_id, action, tier, external_id}` (`action` ∈ created/updated/unchanged).

## Test plan (TDD)

- `tests/test_mcp_journal.py` (7, real vault + real `run_ingest`): entry → JOURNAL fragment carrying its tier; **idempotent re-send** (same id, one fragment); **edit-in-place** (same id, body updated, no duplicate); INTIMATE refused under an OPEN ceiling; INTIMATE stored intimate under an intimate ceiling; empty refused; audit-logged.
- `tests/test_mcp_server.py`: `creek.journal` in `EXPECTED_TOOLS`; `call_tool` idempotency end-to-end (re-send → same fragment id, one file).
- **Regression:** the full `test_ingest_*` / ledger suite passes unchanged (the CLI now delegates to the lifted module) — this is the key safety check for the refactor.
- Coverage: `pipeline.py` 86.1% / `journal.py` 85.9% (> 80% gate). ruff / refurb / mypy --strict clean; no import cycles.

Closes #754
Refs #748

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -53,13 +53,11 @@ if TYPE_CHECKING:
         SeedSpec,
     )
     from creek.generate.mining import MiningRunReport
-    from creek.ingest.base import Ingestor, ParsedFragment
+    from creek.ingest.base import Ingestor
     from creek.ingest.discord_dispatch import DiscordMode
-    from creek.ingest.ledger import LedgerRecord, SourceLedger
     from creek.link.link_engine import LinkSummary
     from creek.models import CompileTargetKind, Fragment, Frequency, Mode, Phase
     from creek.purge import PurgeEngine, PurgeResult
-    from creek.vault.writer import VaultWriter
 
 
 _INCLUDE_TIER_HELP = (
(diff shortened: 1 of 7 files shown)


Geoffe-Ga:
Thanks — the audit-completeness catch is a good one, and you're right it matters here specifically. Fixed in the latest push:

- **Complete write-side audit**: `creek.journal` now mirrors `save_tool`/`ingest_tool`. Malformed calls (blank content/id, unknown tier) refuse without an audit entry; the **tier-ceiling refusal** audits with the attempted `tier`; **success** audits with `tier` + `created_path` (`01-Fragments/Journal`) + `created_tier` + `affected_fragment_ids=[fragment_id]`. So `mcp.jsonl` can now reconstruct that a journal call wrote (or attempted) an INTIMATE fragment, and which fragment id resulted.
- **Validation-order fixed**: the audit call moved out of the top of the function to the refusal/success split, so a malformed call is no longer logged identically to a legitimate write.
- **Regression guards added**: `test_journal_success_audit_records_tier_and_fragment` and `test_refused_intimate_attempt_is_audited_with_tier` assert the audit *content*, not just `tool`/`consumer`.

On the two smaller notes: the double `SourceLedger.load()` I left as-is (correct, and negligible for a single-entry tool, as you noted); the new inbox dir inherits the vault's own filesystem permissions (created via `mkdir` under the vault root, same as every other Creek-Meta subdir). 31 journal/server tests pass; ruff/refurb/mypy clean.

Geoffe-Ga:
Both worth fixing since they touch the feature's core invariants — done in `e289b83`:

- **Idempotency-key collision** (real bug): `_safe_stem` now appends a hash of the *raw* external id, so `"a/b"` and `"a-b"` (which both slug to `a-b`) get distinct staged paths → distinct source-keys → distinct fragments. Still deterministic. Pinned by `test_slug_colliding_external_ids_stay_distinct`.
- **Unaudited ingest failure**: a failure inside `run_ingest` (past the tier check, content already staged) now audits the attempt with its `tier`/`created_tier` before refusing — so, per this PR's own goal, no tier-allowed write attempt is untraceable. Added an injectable `run` seam and `test_ingest_failure_is_refused_and_audited`.

On item 3 (JOURNAL-routing untested): `test_entry_becomes_a_journal_fragment` already asserts `source.platform == "journal"` on the written fragment, so the `journal/`-path classification is pinned — happy to add a more direct assertion if you'd prefer.

33 journal/server tests pass; ruff/refurb/mypy clean.

Geoffe-Ga:
Fixed in the latest push: rather than annotate the two `# type: ignore[arg-type]` comments, I removed the need for them — the idempotency test now uses a small typed `_send()` closure with explicit kwargs instead of `**args`, so mypy passes with zero suppressions (§6.2 satisfied). `grep type: ignore` on the file is now empty; mypy + ruff clean.

CI: the earlier push was green across the board; this test-only change is re-running now — I'll merge once it lands green.

Issue #757, opened 2026-06-30
epic: Confidential per-user hosting — VM, TEE-attested GPU voice, user-held keys, BYOK


Geoffe-Ga:
## Epic Summary

Implements the ratified hosting + key-custody + routing decision (#755, mirrored in `geoffe-ga/adepthood#927`): run Creek as a **persistent per-user VM** (ephemeral compute + a durable, encrypted, **user-owned** volume), reachable by a remote client (Adepthood) over **network MCP**, with **user-held keys + confidential compute (TEE)** and the existing `ModelRouter` (#642) handling tier-gated routing. This is the infrastructure the Adepthood handshake epic (#748) depends on; it is not the handshake itself.

## Decisions this implements (see #755)

- **Hosting:** per-user VM; ephemeral compute + durable encrypted volume.
- **Key custody:** user-held passphrase-derived key (Argon2id); released into the enclave only after **remote attestation**. Operator never holds it.
- **Recovery:** Obsidian-grade (option E) — no operator escrow/reset; a generated **recovery key shown once at setup**. No social recovery, no platform-keychain, no escrow.
- **Voice:** **GPU confidential computing (H100+)** in scope; the attested enclave is `is_cloud=False` and gives even INTIMATE content good private voice.
- **BYOK:** user's own Anthropic/OpenAI/Gemini key for OPEN/PERSONAL; INTIMATE protected automatically by the `ModelRouter` chokepoint (#642).

## Child Issues

- [x] #758 Per-user encrypted volume + user-held passphrase key (Argon2id) + recovery key at setup
- [x] #759 Network MCP transport + per-consumer auth + boundary `TierCeiling`
- [x] #760 Attested GPU-CC enclave provider in `ModelRouter` (`is_cloud=False`, attest-then-mount)
- [ ] #761 BYOK key plumbing for the `generation` stage (user-supplied provider key)
- [ ] #762 ADR: confidential per-user hosting + key custody + recovery

## Constraints

INTIMATE never reachable remotely and never routed to any cloud provider (enforced at the transport boundary AND by `ModelRouter`). The enclave's `is_cloud=False` is valid **only** when it verifies attestation before keys/data flow. Local-first stays the default. Keep in lock-step with `geoffe-ga/adepthood#927` and the handshake epic #748.

## Labels

`epic`, `spec-decomposition`, `epic:confidential-hosting`

Refs #755 · #748 · #642 · geoffe-ga/adepthood#927

[...]

PR #769, opened 2026-07-01
feat(confidential): user-held at-rest volume key — Argon2id + recovery key, no escrow (#758)

### LINKED ISSUE #758: feat: per-user encrypted volume + user-held passphrase key (Argon2id) + recovery key at setup
## Goal

The vault's durable volume is encrypted with a **user-held** key (passphrase-derived, Argon2id); the operator never stores it. At setup the user is shown a generated **recovery key** to save (option E — no operator escrow/reset). This is the data-layer the per-user VM mounts.

## Context

- Vault is a file tree (`~/Obsidian/Creek-Vault`); `creek init` scaffolds it. Encryption-at-rest wraps the volume the VM mounts.
- The unwrapped key is released into the enclave only after attestation (the enclave-provider issue handles attestation; this issue owns the at-rest key + recovery-key UX/flow).
- Decision: #757 / #755. Recovery model is deliberately minimal — Obsidian Sync's optional-E2EE parallel (unrecoverable by the operator).

## Acceptance Criteria

- [ ] Volume key is derived from a user passphrase via Argon2id (sane params); plaintext key never persisted operator-side.
- [ ] A one-time recovery key is generated and surfaced for the user to store; it can unwrap the volume independently of the passphrase.
- [ ] No operator escrow, no reset path; losing both passphrase and recovery key is (documented as) unrecoverable.
- [ ] Key handling never logs secrets; at-rest artifacts are ciphertext only.

## Constraints / DoD

stay-green TDD. No secret material in code/tests/fixtures or logs. `pre-commit` clean; coverage met; no bypasses. PR: `Refs #757`, `Closes #<this>`.

Refs #757
Refs #755

Geoffe-Ga:
## Summary

The confidential-hosting data layer: the vault's durable volume is encrypted with a **volume master key (VMK)** the operator never stores in plaintext. New `creek/confidential/keyvault.py` wraps the random VMK twice, giving **two independent unwrap paths**:

- **passphrase path** — KEK derived with **Argon2id** (memory-hard) over a random salt, AES-256-GCM-wraps the VMK.
- **recovery path** — a one-time, 256-bit **recovery key** is generated and surfaced for the user to store; its KEK is derived with HKDF-SHA256 and wraps the VMK a second time.

First epic-`confidential-hosting` deliverable. Refs the hosting/enclave decision (#757/#755).

## Security properties (each AC → test)

- **No operator escrow / no reset.** The persisted `KeyVault` is **ciphertext + public params only** (salt, Argon2id cost, AEAD nonces). The VMK, passphrase, and recovery key are never persisted. Losing *both* the passphrase and the recovery key is unrecoverable **by design** (documented; mirrors Obsidian Sync's optional E2EE). → `test_persisted_vault_holds_no_plaintext_key_material`, `test_persisted_file_is_ciphertext_only_and_owner_readable`.
- **Argon2id, memory-hard.** 64 MiB / t=3 / p=4, above the OWASP floor; params persisted per-vault so a future increase doesn't lock out old vaults. → `test_argon2_parameters_are_memory_hard`.
- **Independent recovery.** Either path recovers the identical VMK. → `test_passphrase_and_recovery_unwrap_the_same_key`.
- **Fail-closed.** AES-GCM authentication + **domain-separated AAD** per path: a wrong passphrase/recovery key, a malformed recovery key, or a tampered artifact raises `UnlockError` (a single fixed, non-revealing message — leaks nothing about *why*), never garbage. → `test_wrong_passphrase_fails_closed`, `test_wrong_recovery_key_fails_closed`, `test_malformed_recovery_key_fails_closed`, `test_tampered_ciphertext_fails_closed`.
- **No secret logging.** The module never logs; secret material lives only in locals + the returned `bytes` / one-time recovery string.
- On-disk artifact written `0600` (defence in depth, though it's ciphertext).

## Design notes

- One dependency, already declared: `cryptography>=48` (49.0.0) — `Argon2id`, `AESGCM`, `HKDF` from `hazmat`. No `argon2-cffi` needed.
- Recovery key is grouped, unpadded, uppercase base32 (human-storable; decoding is case/space-tolerant).
- `save_key_vault`/`load_key_vault` persist the ciphertext-only artifact.

## Test plan

- `tests/test_confidential_keyvault.py` (12): both-paths-same-VMK, wrong/malformed/tampered fail-closed, ciphertext-only vault + on-disk `0600`, memory-hard params, per-setup uniqueness (fresh randomness), JSON round-trip still unlocks, recovery-key format tolerance, empty-passphrase rejected.
- **No secret material hardcoded** (passphrases are test literals; every key is generated per test). `bandit -r creek/confidential/` clean. tryceratops/refurb/ruff/mypy --strict clean. `keyvault.py` 98% coverage.

Closes #758
Refs #757

### DIFF
--- a/creek-tools/creek/confidential/__init__.py
+++ b/creek-tools/creek/confidential/__init__.py
@@ -1,0 +1,29 @@
+"""Confidential-hosting data layer: at-rest volume key management (#758).
+
+Exposes the user-held encryption key model for the vault volume: a passphrase
+(Argon2id) and an independent one-time recovery key both unwrap the same volume
+master key, with NO operator escrow and no reset path. See
+:mod:`creek.confidential.keyvault`.
+"""
+
+from creek.confidential.keyvault import (
+    KeyVault,
+    SetupResult,
+    UnlockError,
+    create_key_vault,
+    load_key_vault,
+    save_key_vault,
+    unlock_with_passphrase,
+    unlock_with_recovery,
+)
+
+__all__ = [
+    "KeyVault",
+    "SetupResult",
+    "UnlockError",
+    "create_key_vault",
+    "load_key_vault",
+    "save_key_vault",
+    "unlock_with_passphrase",
+    "unlock_with_recovery",
+]
--- a/creek-tools/creek/confidential/keyvault.py
+++ b/creek-tools/creek/confidential/keyvault.py
@@ -31,17 +31,16 @@ from __future__ import annotations
 import base64
 import json
 import os
+import tempfile
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Final
+from pathlib import Path
+from typing import Any, Final
 
 from cryptography.exceptions import InvalidTag
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.ciphers.aead import AESGCM
 from cryptography.hazmat.primitives.kdf.argon2 import Argon2id
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
-
-if TYPE_CHECKING:
-    from pathlib import Path
 
 _VMK_BYTES: Final[int] = 32
 """Volume master key size — a 256-bit AES key."""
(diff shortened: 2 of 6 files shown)


Geoffe-Ga:
Thorough review — all three worth doing, done in the latest push:

- **Direct `cryptography` dependency**: added `cryptography>=48.0.0` to `[project] dependencies` and regenerated `uv.lock`. You're right the module imports `hazmat.*` unconditionally, so it can't rely on the `mcp` -> `pyjwt[crypto]` transitive edge. The lock diff adds only the direct dependency edge (no version bumps).
- **Atomic `save_key_vault`**: now writes to a `0600` tempfile (`mkstemp`, created owner-only — no widen-then-narrow), `fsync`s, and `os.replace`s into place; cleans up the temp on any error. This directly serves the threat model — the artifact is the only copy and "no reset" means a torn write would be accidental permanent lockout. Added an assertion that no temp file is left behind.
- **ADR-0005**: `docs/architecture/ADR/0005-confidential-volume-key-no-escrow.md` documents the no-escrow/no-reset decision — the escrow-vs-user-held trade-off, the irreversible-on-double-loss consequence, and explicit revisit triggers (so a future recovery path gets re-litigated, not added silently).

All gates clean locally (ruff/refurb/tryceratops/mypy/bandit); 12 keyvault tests pass; keyvault.py 98% coverage.

Geoffe-Ga:
Thanks — filed all three as the fast-follow issues you asked for (kept them out of this foundational PR per your call):

- **#770** — enforce a passphrase-strength floor at setup (the HIGH passphrase-floor item).
- **#771** — bounds-check untrusted KDF params in `from_dict` before Argon2id (the HIGH untrusted-on-disk-params item).
- **#772** — regression test for cross-path AAD domain separation (the test gap).

All three land under `epic:confidential-hosting`, `Refs #758`. This PR itself has all prior review items addressed (direct dep, atomic 0600 save, ADR-0005) and no blockers — I'll merge once CI is fully green.

PR #782, opened 2026-07-01
perf(classify): make llm.max_concurrent effective (engine loop was serial)

### LINKED ISSUE #764: perf(classify): llm.max_concurrent has no effect on classify throughput (per-fragment work is GIL-serialized)
## Problem

`creek classify --method llm` throughput is effectively independent of `llm.max_concurrent`. Raising the classification stage's `max_concurrent` from the default **5 → 12** produced **no change** in throughput (~15–20 fragments/min in both cases). For a real backfill (~21k unclassified fragments on `creek-demo-2026-06-07`) this means an **~18–24h** run where the cloud provider (Haiku) latency budget predicts closer to ~3h.

At a measured ~15–20/min with 12 workers, the *effective* per-call latency works out to ~40–48s/call — absurd for a Haiku network call (~1–2s). So the wall-clock is dominated by per-fragment work that does **not** parallelize, not by the LLM round-trip.

## Evidence

- Two live runs on the same vault (`/Users/.../creek-demo-2026-06-07`, ~21,464 `primary: unclassified` fragments), resumable so each only processes the remaining set:
  - `max_concurrent: 5`  → +29 `classification_method: llm` in 91s (~19/min)
  - `max_concurrent: 12` → +22 `classification_method: llm` in 91s (~15/min)
  - Identical within noise; the higher worker count did nothing.
- Provider was **not** the limiter: 0 rate-limit / 429 / backoff lines in the run log; Ollama had no model loaded (Intimate-tier redirects are rare), so the local path wasn't saturating either.

## Root cause (hypothesis)

`run_batch` itself is correct — `creek/classify/llm/batch.py:79-89` submits **all** tasks to a `ThreadPoolExecutor(max_workers=config.max_concurrent)` up front and drains via `as_completed`, so the pool is genuinely wide. The serialization is therefore **inside `classifier.classify(fragment)`**: the per-fragment pipeline (rules pre-pass, local embedding generation for few-shot/similarity, frontmatter read + rewrite) is **CPU-bound work under the GIL**, which pins effective parallelism near 1 regardless of worker count. Python threads only overlap on I/O; the dominant cost here is local CPU, so more threads ≠ more throughput.

(Contributing: the submission loop also sleeps `RATE_LIMIT_DELAY=0.1s` between every submit — `batch.py:87-88` — which caps *submission* at 10/s. Not the main bottleneck at 15/min, but worth noting.)

## Impact

- A one-time full-vault LLM backfill is a multi-hour-to-overnight job instead of a coffee-break job.
- `llm.max_concurrent` is a documented, advertised knob (config comment, `--help`) that silently has no effect on the cloud classify path — misleading.

## Suggested fix strategy

1. **Profile `classifier.classify`** on a cloud provider to confirm where wall-clock goes (rules pre-pass vs. embedding vs. file I/O vs. network). A 50-fragment `cProfile` run should settle it.
2. If local CPU dominates (likely the embedding/few-shot step): move the network-bound LLM calls onto an async path (`anthropic.AsyncAnthropic` + `asyncio.gather` with a semaphore), or run the CPU work in a `ProcessPoolExecutor` so it actually parallelizes; keep the thread pool only for the network leg.
3. Alternatively, wire the **Anthropic Message Batches API** (50% cheaper + async) for large backfills — addresses both cost and the serialization, since submission/poll is decoupled from per-fragment CPU work.
4. Either fix the knob or document that `max_concurrent` is ineffective for cloud classify and why.

## Repro

```bash
# resumable; safe to stop/restart
creek classify --method llm --vault <vault>
# bump llm.classification.max_concurrent 5 -> 12 in creek_config.yaml, restart, compare
# count: grep -rlE '^\s*classification_method:\s*llm' <vault>/01-Fragments | wc -l   (sample over 90s)
```

Discovered while backfilling `creek-demo-2026-06-07` with Haiku (`llm.classification = {provider: anthropic, model: claude-haiku-4-5}`).

Geoffe-Ga:
## Summary

`creek classify --method llm` throughput was independent of `llm.max_concurrent` (5→12 workers gave no change; a ~21k-fragment backfill ran ~18–24h instead of the ~3h the Haiku latency budget predicts).

**Root cause — a correction to the issue's hypothesis.** The issue guessed GIL-serialized per-fragment CPU (local embeddings). Reading the code, that's not it: the LLM classify path does **no** local embedding, and few-shot selection is `@lru_cache`d. The real cause is simpler — **`run_classify` iterated fragments in a plain serial `for md_file in rglob(...)` loop and never used a thread pool.** The `ThreadPoolExecutor` in `batch.py`'s `run_batch` is a *separate* code path the engine doesn't call. So `max_concurrent` had no effect because the loop was single-threaded, not because of the GIL. At ~3–4s/fragment (Haiku two-step reasoning prompt) run serially, that's exactly the observed 15–20/min.

## Fix
Drive the engine loop through a `ThreadPoolExecutor` sized by the classification stage's `max_concurrent`:
- the network-bound provider call (`_classify_one`) runs **outside** a lock, so calls from different worker threads genuinely overlap — that's what makes the knob effective;
- the cheap shared-state sections (`counts` mutation, progress/trace-log appends) run **under a `threading.Lock`**, so counts stay exact under concurrency (no lost updates);
- classifiers are pre-warmed by the existing pre-loop availability check, so concurrent first-access doesn't race on the lazy provider build;
- **re-atomization** (shared `VaultWriter`) and the **rules** path (local CPU, no I/O to overlap) stay serial;
- in-flight requests are bounded by `max_concurrent`, so there's no request burst (no new 429 risk).

Extracted `_dispatch_files` + `_assert_classifiers_available` so `run_classify` stays under the complexity gate.

## Test plan
`tests/test_classify_concurrency.py` (3 cases, LLM faked — no network):
- **Concurrency proven deterministically:** a `threading.Barrier(4)` inside the faked classify call — with `max_concurrent=4` the four calls rendezvous (barrier trips, `max_active==4`); a serial loop can *never* trip it (it would time out). This is the direct regression guard for the bug.
- **`max_concurrent=1` stays strictly serial** (`max_active==1`) — backward-compatible.
- **Correctness under concurrency:** a wide run over 12 fragments with 4 workers classifies and counts every fragment exactly once, and each file carries the `classification_method: llm` stamp.

Local `./scripts/check-all.sh`: ruff/format/mypy/pylint/refurb/tryceratops/complexity/security clean; existing classify-engine / tier-routing / resume / interrupt suites green (56 pass). (The 8 author-desk credential-dependent failures are the known local-only set green in CI.)

Closes #764

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -28,7 +28,10 @@ from __future__ import annotations
 
 import json
 import logging
+import threading
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
+from functools import partial
 from pathlib import Path  # noqa: TC003  # no issue: runtime dataclass field
 from typing import TYPE_CHECKING
 
@@ -77,7 +80,7 @@ _PROCESSING_LOG_SUBDIR = ("00-Creek-Meta", "Processing-Log")
 """Per-vault subdirectory carrying the LLM progress + trace logs."""
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import Callable, Iterator
 
     from creek.config import ClassificationConfig, CreekConfig, LLMConfig
 
@@ -308,19 +311,7 @@ def run_classify(
     # #647). Resolving here also fails the run loudly (IntimateRoutingError) when
     # classification is cloud and no local default exists to route Intimate to.
     tier_classifiers = build_tier_classifiers(config) if method == LLM_METHOD else None
-    if tier_classifiers is not None:
-        for classifier in tier_classifiers.distinct():
-            if not classifier.available:
-                # Refuse to iterate when a provider is unreachable so the vault
-                # is never stamped with ``classification_method: llm`` for
-                # fragments the LLM never actually saw. The orchestrator already
-                # logged the provider-specific reason (missing API key, missing
-                # consent env var, Ollama down) as a WARNING; we capture it here
-                # so the remediation hint is not buried in stderr.
-                raise LLMProviderUnavailableError(
-                    provider=classifier.config.provider,
-                    detail=_describe_llm_unavailability(classifier.config.provider),
-                )
+    _assert_classifiers_available(tier_classifiers)
     counts = _RunCounts()
     progress_path: Path | None = None
     trace_log_path: Path | None = None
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
Thanks — thorough review, no blockers. Addressed both notes now (pushed 9fbe72e):

- **Pre-warm race invariant (Problem 1):** added a thread-safety note to `LLMClassifier._provider()` spelling out that a classifier must be pre-warmed serially before concurrent use (the pre-loop `_assert_classifiers_available` does this), so a future concurrent caller that skips it can't silently reintroduce the double-build race.
- **Reatomize-serial guarantee (Testing gap):** added `test_reatomize_forces_serial_even_with_high_max_concurrent` — with `reatomize=True` and `max_concurrent=4` the loop stays serial (`max_active==1`), so the shared-`VaultWriter` guard is now a named, tested invariant.
- **Per-tier pool sizing (Problem 2):** filed as follow-up **#783** — the current pool is sized by the cloud `classification` stage's `max_concurrent`, which could oversubscribe a local provider on a mixed-tier run. Hardening, not a correctness bug.

@claude please re-review.

PR #773, opened 2026-07-01
feat(mcp): network transport + per-consumer auth + boundary TierCeiling

Geoffe-Ga:
## Summary

Exposes `creek-tools-mcp` over an **authenticated streamable-http network transport** so a remote Adepthood backend can reach a user's per-user-VM vault — without weakening the privacy guarantees. Stdio remains the default for local consumers (Claude Code, CrawDad), unchanged. Resolves the transport fork (#755) with the decided option (network MCP), per operator choice: **per-consumer bearer tokens from env, remote callers capped below INTIMATE**.

### What changed
- **`creek_mcp/remote_auth.py`** (new) — `CREEK_MCP_CONSUMER_TOKENS` holds `consumer=token` pairs in the **environment only** (never in code/config). `ConsumerTokenVerifier` verifies a bearer constant-time against every known token and yields an `AccessToken` whose `client_id` is the consumer. Empty/unset ⇒ no tokens ⇒ network mode denied (no anonymous access).
- **`creek_mcp/server.py`** —
  - `_BoundedFastMCP` overrides `call_tool` as the single boundary chokepoint: when the call is **remote** (`get_access_token()` is set), a requested `privacy_tier_ceiling` above `personal` (`intimate`/`all`, or any unrecognised value) is **refused before dispatch** — intimate content is never even read for a network consumer. Local stdio calls (no token) pass through unchanged.
  - `_effective_consumer` audits each remote call under the token's consumer identity (not the process-env default).
  - `--transport network` wiring (with `--host`/`--port`) that refuses to start without configured tokens; stdio path is untouched.
- **`docs/mcp.md`** — new "Network transport" section documenting the env-only tokens, per-consumer identity, and the INTIMATE-never-remote cap.

## Test plan
`tests/test_mcp_remote.py` (17 cases) asserts the security properties:
- **No anonymous access** — `main(["--transport","network"])` with no tokens exits; over-the-wire `TestClient` POST with no / wrong bearer → `401` before any tool runs.
- **INTIMATE never reachable remotely** — a remote caller requesting `intimate`/`all`/garbage is refused before dispatch (asserted: `status == "refused"`, nothing audited); `open`/`personal` dispatch.
- **Per-consumer identity** — a dispatched remote call is audited under the bearer's consumer, overriding the env default.
- **Stdio unchanged** — with no verifier wired, a local `intimate` call is *not* capped.
- Unit coverage for `load_consumer_tokens` parsing and `ConsumerTokenVerifier.verify_token`.

[...]

Issue #759, opened 2026-06-30
feat(mcp): network transport + per-consumer auth + boundary TierCeiling


Geoffe-Ga:
## Goal

Expose `creek-tools-mcp` over an authenticated **network transport** so a remote Adepthood backend can reach a user's per-user-VM vault — without weakening the privacy guarantees. Resolves the transport fork (#755) with the decided option (network MCP).

## Context

- Today: `creek-tools/creek_mcp/server.py` runs `run(transport="stdio")` with local consumers (CrawDad, Claude Code) via `CREEK_MCP_CONSUMER`.
- Add a network transport path (HTTP/SSE/streamable per the MCP SDK) alongside stdio; do not break existing stdio consumers.
- Boundary enforcement: per-consumer identity + auth, the existing hash-chained audit, and a `TierCeiling` **default-deny** at the boundary so INTIMATE is never reachable remotely.

## Acceptance Criteria

- [ ] The server can serve over an authenticated network transport; stdio remains for local consumers.
- [ ] Every remote call is authenticated and carries a consumer identity; unauthenticated/over-ceiling calls are refused.
- [ ] `TierCeiling` is enforced at the boundary with a safe default (remote consumers cannot read INTIMATE); audit chain extends to remote calls.
- [ ] Exposure is hardened (no anonymous access; least-privilege per consumer).

## Constraints / DoD

stay-green TDD. INTIMATE never reachable remotely (assert). `pre-commit` clean; coverage met; no bypasses. PR: `Refs #757`, `Closes #<this>`. Depends on the per-user volume/key (#758) for the hosting context but the transport layer is testable independently.

Refs #757
Refs #755 · #748

PR #788, opened 2026-07-01
test(mcp): over-the-wire valid-bearer happy-path through real middleware

### LINKED ISSUE #775: test(mcp): over-the-wire valid-bearer happy-path through real streamable_http_app middleware
Follow-up from #773 (COMMENTS review on #759) — https://github.com/Geoffe-Ga/Creek-Vault/pull/773

`test_streamable_http_rejects_unauthenticated_request` covers the 401 rejection paths through the real `TestClient`/HTTP layer, but the ceiling-cap + dispatch success path is only verified via direct `server.call_tool(...)` with a monkeypatched `_current_access_token` — bypassing the SDK's real auth middleware for the happy path.

## Actionable item
Add an over-the-wire test that presents a **valid** bearer token and asserts a full authenticated `tools/call` round-trip actually dispatches through `streamable_http_app()` (real `RequireAuthMiddleware`), and that the remote ceiling cap engages on that real path (a remote `intimate` request is refused end-to-end, an `open`/`personal` request succeeds). This closes the loop that the happy path is reached through the actual middleware, not just the failure path.

Note: a streamable-http `tools/call` requires the MCP session-init handshake; the test must drive the SDK client (or the initialize→call sequence) rather than a bare POST.

## Done-Done
- [ ] New test in tests/test_mcp_remote.py; `./scripts/check-all.sh` clean; PR `Refs #757`.

Geoffe-Ga:
## Summary

Follow-up from the #773 COMMENTS review on #759. Previously the authenticated **happy path** and the remote **ceiling cap** were only verified via direct `server.call_tool(...)` with a monkeypatched `_current_access_token` — bypassing the SDK's real auth middleware. This adds two tests that drive a full authenticated `tools/call` through `streamable_http_app()` via `TestClient`, so the **real `RequireAuthMiddleware`** runs in-process.

## Test plan
`tests/test_mcp_remote.py` (2 new cases; the LLM/vault are stubbed, no network):
- **`test_authenticated_tools_call_dispatches_over_the_wire`** — runs the MCP `initialize` → `notifications/initialized` → `tools/call` handshake with a **valid bearer** and an `open` ceiling; asserts `creek.wheel` dispatches end-to-end (`structuredContent.tool == "creek.wheel"`, not refused).
- **`test_remote_intimate_is_refused_over_the_wire`** — the same authenticated path with an `intimate` ceiling is **refused by `_BoundedFastMCP` end-to-end**. `get_access_token()` is populated by the real middleware from the verified token (no monkeypatch), so this proves the cap engages on the production path, not just via a patched context var.

Implementation note: DNS-rebinding host validation (a concern separate from auth) is disabled for the in-test `TestClient` host so requests reach the auth middleware; the 401 rejection tests already cover the unauthenticated paths.

Local `./scripts/check-all.sh`: ruff/format/mypy/pylint/refurb/tryceratops/complexity/security clean; `test_mcp_remote.py` passes (19). (The 8 author-desk credential-dependent failures are the known local-only set green in CI.)

Refs #757
Closes #775

### DIFF
--- a/creek-tools/tests/test_mcp_remote.py
+++ b/creek-tools/tests/test_mcp_remote.py
@@ -25,6 +25,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 from mcp.server.auth.provider import AccessToken
+from mcp.server.transport_security import TransportSecuritySettings
 from starlette.testclient import TestClient
 
 from creek_mcp import server as server_mod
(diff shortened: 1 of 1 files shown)


Geoffe-Ga:
Thanks — COMMENTS, no blockers. Took the Code Quality win: annotated `_network_server` as `-> FastMCP` (in the `TYPE_CHECKING` block, annotation-only), which drops all 4 `# type: ignore[attr-defined]` and matches this file's existing pattern — cleaner and consistent with §6.2. The DNS-rebinding-disable stays scoped to the test helper only (production `main()` is untouched, as you noted).

@claude please re-review.

Issue #666, opened 2026-06-26
feat(classify): per-tier classification routing — Haiku for non-Intimate, local for Intimate


Geoffe-Ga:
## Role
Senior Python engineer in `creek/classify/` (the classification engine + the LLM provider path) and `creek/classify/llm/router.py`, fluent in the #642 per-stage routing chokepoint and the `creek classify` run loop.

## Goal
Make `creek classify --method llm` route **per fragment by privacy tier**, so the model-routing assignment "**Haiku for non-Intimate, local for Intimate**" is configurable in one classify run. Concretely: an `Intimate`-tier fragment is classified by a **local** provider (via the `_enforce_local_for_intimate` chokepoint) even when the `classification` stage is configured for a cloud model, while non-`Intimate` fragments use the configured (possibly cloud) model.

## Context
- **The gap (surfaced by the #663 `/model-setup` skill):** every classification call site resolves the stage with **no tier** — `resolve("classification")`, tier `None` — so the chokepoint never fires per fragment and there is no mechanism to split classification by tier:
  - `creek/pipeline.py:170`, `creek/cli.py:1156`, `creek/cli.py:2512`, `creek/cli.py:2611`, `creek/classify/classify_engine.py:215`.
- **Architectural crux:** the classifier is built **once per run** — `run_classify` does `classify_llm = config.model_router.resolve("classification")` then `LLMClassifier(config=classify_llm)` (`classify_engine.py:~213`) and iterates fragments with a fixed provider. Per-tier routing therefore needs either (a) **partition fragments by tier** and run the local provider over the `Intimate` partition and the configured provider over the rest, or (b) make the per-fragment classify step resolve its provider from the fragment's `privacy_tier`. Pick per the trade-off; (a) is likely simpler and batch-friendly.
- **Reuse the chokepoint:** `ModelRouter.resolve("classification", tier)` already redirects `Intimate` → local `default` (or raises `IntimateRoutingError` if `default` is cloud) — `router.py:80/142`. Do NOT add a second tier check; pass the fragment's tier and trust the result. This is exactly the per-fragment-tier wiring **#660** (tier-filter desk retrieval) and **#661** (route the voice call by content tier) did for the author desk — mirror that precedent.
[...]

PR #799, opened 2026-07-07
perf(classify): size the classify pool per-tier via per-provider semaphores (#783)

### LINKED ISSUE #783: perf(classify): size the classify pool per-tier, not only by the cloud classification stage
Follow-up from #764 / PR #782 (COMMENTS review).

#764 made `creek classify --method llm` concurrent, sizing the thread pool by the **non-Intimate `classification` stage's** `max_concurrent` (`classify_engine.py` `run_classify`). But INTIMATE fragments route to a *different* (local) provider via `TierClassifiers`, so a mixed-tier vault applies the cloud-sized concurrency to local-model calls too.

## Goal
Consider sizing concurrency per-tier (or capping the local path separately) so a large cloud `max_concurrent` doesn't oversubscribe a local provider (e.g. Ollama) when a run mixes tiers. Likely fine in practice (Ollama queues internally), so this is a hardening/robustness follow-up, not a correctness bug.

## Options
- Bucket files by resolved provider and size each bucket's pool by that provider's `max_concurrent`.
- Or cap the local/enclave path at a smaller fixed width regardless of the cloud knob.

## Done-Done
- [ ] Local-provider calls aren't oversubscribed by the cloud `max_concurrent`; tested; `./scripts/check-all.sh` clean.

Geoffe-Ga:
## Summary
- `creek classify --method llm` sized its single thread pool by the non-Intimate `classification` stage's `max_concurrent` (#764), so a mixed-tier vault applied the cloud width to the local provider serving INTIMATE fragments — oversubscribing it.
- Now the outer pool is sized as the **sum** of the distinct classifiers' `max_concurrent`, and each fragment's leaf classify call is guarded by a `threading.BoundedSemaphore` owned by its resolved provider (keyed by classifier identity). Each provider is capped by its own knob regardless of pool width; both tiers saturate independently and the local path is never oversubscribed by the cloud knob.
- When `TierClassifiers.distinct()` returns a single classifier (all-cloud, all-local, or deferred-intimate) the pool width collapses to the exact pre-#783 value — regression-identical. Reatomize and rules paths stay strictly serial. The semaphore guards only the leaf call (outside the shared lock, released on exception), so it never nests with the lock and is never held during the vault write.

## Test plan
- Added four mixed-tier tests to `tests/test_classify_concurrency.py` (extending the #764 instrumented-classifier fixtures, now keyed per-provider): local-tier calls capped by the local knob not the cloud knob; cloud + local tiers saturating independently (a `Barrier` kills a naive `min()`-based fix); single-shared-classifier pool width unchanged; deferred-intimate route runs with no intimate fragments. Confirmed the two cap tests fail RED against current `main` for the right reason, then GREEN after the fix.
- `env -u ANTHROPIC_API_KEY -u CREEK_CLOUD_CONSENT -u CREEK_ANTHROPIC_CONSENT ./scripts/check-all.sh` — all 12 checks pass (mypy strict, ruff, pylint, refurb, tryceratops, interrogate, bandit, complexity, 5958 tests, coverage 93.86%, per-file gate).

## Note (out-of-scope follow-up)
Self-review flagged that `LLMConfig.max_concurrent` (`config.py`) has no `Field(ge=1)` lower bound: a `max_concurrent: 0` misconfig would now build `BoundedSemaphore(0)` and hang on `acquire()` (previously harmless serial). Left out of scope for this issue; worth a hardening follow-up to reject `max_concurrent < 1` at config load.

Closes #783

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -30,6 +30,7 @@ import json
 import logging
 import threading
 from concurrent.futures import ThreadPoolExecutor
+from contextlib import contextmanager
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path  # noqa: TC003  # no issue: runtime dataclass field
@@ -333,6 +334,20 @@ def run_classify(
         VaultWriter(vault_path=vault_path) if reatomize_enabled else None
     )
 
+    # Concurrency (#764/#783): the per-fragment LLM call is network-bound, so
+    # running files through a thread pool lets the calls overlap — previously
+    # this loop was serial and ``max_concurrent`` had no effect. Each distinct
+    # provider gets its OWN bounded semaphore sized by its own stage's
+    # ``max_concurrent``: INTIMATE fragments route to a local provider (#666),
+    # which must not inherit the cloud stage's width (#783). The outer pool is
+    # sized as the SUM of the distinct providers' widths so both tiers can
+    # saturate their own cap at the same time without oversubscribing either.
+    # Re-atomization shares one writer, so it stays serial (workers forced to 1);
+    # the rules path builds no throttle and also stays serial.
+    throttle, workers = _build_classify_throttle(tier_classifiers)
+    if vault_writer is not None:
+        workers = 1
+
     files = sorted(fragments_root.rglob("*.md"))
     process_one = partial(
         _process_file,
@@ -346,18 +361,8 @@ def run_classify(
         progress_path=progress_path,
         trace_log_path=trace_log_path,
         vault_writer=vault_writer,
+        throttle=throttle,
         lock=threading.Lock(),
-    )
-
-    # Concurrency (#764): the per-fragment LLM call is network-bound, so running
-    # files through a thread pool sized by the classification stage's
-    # ``max_concurrent`` lets the calls overlap — previously this loop was serial
-    # and the knob had no effect. Re-atomization keeps a single shared writer, and
-    # the rules path is local CPU with no I/O to overlap, so both stay serial.
-    workers = (
-        config.llm.for_stage("classification").max_concurrent
-        if method == LLM_METHOD and vault_writer is None
-        else 1
     )
     _dispatch_files(files, process_one, workers=workers)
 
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real regression from #764: sizing the classify thread pool by only the cloud `classification` stage's `max_concurrent` caused a mixed-tier vault to apply the cloud provider's width to the local Intimate-tier provider, oversubscribing it. The fix gives each distinct resolved classifier (`TierClassifiers.distinct()`) its own `threading.BoundedSemaphore` keyed by `id(classifier)`, sums the distinct widths for the outer pool, and wraps only the leaf `_classify_one` call in the corresponding permit.

## Strengths
- Clean separation of concerns: `_build_classify_throttle` (sizing) and `_classify_permit` (per-call guard) are small, single-purpose, and easy to reason about independently.
- The semaphore keying (`id(classifier)`) correctly lines up with `TierClassifiers.for_tier`'s instance resolution — traced through `build_tier_classifiers` → `distinct()` → `_process_file`'s `tier_classifiers.for_tier(tier_of(fragment))`, the same instances are used for both throttle construction and per-fragment lookup, so the permit always matches the actual provider making the call.
- The permit is acquired outside the shared `lock` (preserving the #764 fix that let calls overlap) and released on exception (the `with semaphore:` block wraps `_classify_one`), so a raised classify error can't leak a permit.
- Backward-compatibility is explicitly preserved and tested: when `distinct()` collapses to one classifier (shared instance), the pool width is identical to pre-#783 behavior.
- Docstrings are thorough (Args/Returns/Yields, cross-references to `#764`/`#666`/`#783`) and explain *why*, not just *what*.

## Security Concerns
None identified. 🟢 No new I/O, no new external input handling, no auth/secrets touched. Concurrency change is confined to the classify engine's internal thread pool.

## Problems
No blocking correctness issues found. One item worth flagging (see Code Quality) is a hang risk under misconfiguration, but it's pre-existing in scope and already called out by the author.

## Code Quality
- 🟡 `LLMConfig.max_concurrent` (`creek/config.py:166`) has no `Field(ge=1)` lower bound, unlike sibling knobs in the same file (`resonance_top_k`, `exchange_max_gap_minutes`, `reatomize_max_depth`, etc. all use `ge=1`). Before this PR, `max_concurrent: 0` was silently harmless (the old ternary `workers = ... if ... else 1` meant `_dispatch_files` just ran serially). After this PR, the same misconfiguration builds `threading.BoundedSemaphore(0)` — every call to `_classify_permit` for that provider will block forever on `with semaphore:` since a zero-value semaphore can never be acquired, hanging the entire run with no error surfaced. The PR description explicitly flags this as an out-of-scope follow-up, which is reasonable given the repo's pattern of small dedicated hardening PRs — but since this change is precisely what converts the gap from harmless to a silent hang, and the fix is a one-line `Field(ge=1)` consistent with the rest of the file, it's worth strong consideration for landing in this PR (or as an immediate fast-follow) rather than deferred indefinitely.
- Minor: when the two tiers use distinct providers, the outer pool is sized as the *sum* of both widths even in a run containing fragments for only one tier (e.g., an all-OPEN vault with a configured-but-unused local Intimate route still gets `cloud_width + local_width` threads). This is harmless (idle threads, no oversubscription) but is worth a one-line docstring note if it ever surprises someone profiling thread counts.

## CLAUDE.md Adherence
✅ Coverage: PR description reports `check-all.sh` green with 93.86% coverage (≥90% threshold met).
✅ Docstrings: All new/changed public and private functions carry full Google-style docstrings (Args/Returns/Raises/Yields) — consistent with the ≥95% interrogate gate.
✅ Complexity: New logic is factored into two small helpers (`_build_classify_throttle`, `_classify_permit`) rather than growing `_process_file`/`run_classify`'s complexity — in the spirit of the ≤10 cyclomatic complexity ceiling.
✅ Commit/branch conventions: `perf(classify): ...` follows Conventional Commits; branch name `issue/783-perf-classify-size-the-classify-pool-per` follows the issue-first workflow.
✅ No shortcuts: no `noqa`/`type: ignore` additions; the one known gap (`max_concurrent` bound) is disclosed rather than silently swept under the rug.

## Testing
Strong. TDD is evident from the PR description (RED confirmed against `main` before GREEN). The four new tests correctly target the actual regression:
- `test_intimate_calls_capped_by_local_max_concurrent_not_cloud` and `test_cloud_and_local_tiers_saturate_independently` prove the fix using a `Barrier` for the cloud side (deterministic proof of genuine overlap, not just "didn't crash") while asserting the local side's `max_active` bound — this is exactly the kind of test that would fail for the right reason against the pre-fix `min()`-style approach the docstring calls out.
- `test_single_shared_classifier_pool_width_unchanged` and `test_deferred_intimate_route_runs_with_no_intimate_fragments` pin the two backward-compatible shapes.
- The per-provider `probes: ClassVar[dict[str, _Probe]]` refactor (keyed by `self.config.provider`) is a sensible adaptation to let a single test assert on two providers independently.
No test exercises the `max_concurrent: 0` edge case noted above, but that's consistent with it being explicitly out of scope.

## Documentation
Docstrings on the new functions and the updated module-level test docstring are thorough and explain the causal chain (#764 → #666 → #783) clearly enough for a future reader unfamiliar with the history. No external/user-facing docs (README, ADRs) needed for this internal performance fix.



## Verdict: COMMENTS

No bugs block merge: the per-provider semaphore design is correct (verified id()-keyed throttle maps to the same classifier instances `for_tier` resolves, the single-classifier collapse path is regression-tested, and the reatomize-forces-serial path is preserved), and the four new tests are deterministic (Barrier-based) where it matters. The one substantive item — `LLMConfig.max_concurrent` has no `ge=1` floor, so this PR turns a previously-harmless `max_concurrent: 0` misconfig into an indefinite hang via `BoundedSemaphore(0)` — is real and trivial to fix (one line, consistent with `Field(ge=1)` used elsewhere in `config.py`), but the author already disclosed it transparently as a scoped follow-up, so it doesn't block this merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: COMMENTS
**Action**: pull comment 4907520970 to see in-depth feedback and continue iterating

Geoffe-Ga:
Follow-ups for the two review items: #800 (LLMConfig.max_concurrent ge=1 floor — pre-filed from the PR's own out-of-scope note, now agent-ready) and the docstring note on pool-width summing (issue above). Merging on green CI per the COMMENTS verdict.

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: COMMENTS
**Action**: pull comment 4907520970 to see in-depth feedback and continue iterating

PR #792, opened 2026-07-07
chore(ralph): true up subagent fleet, maintenance scans, and hopper from well-worn-tools

Geoffe-Ga:
## Summary

True-up with [`well-worn-tools`](https://github.com/Geoffe-Ga/well-worn-tools): the full Ralph fleet, its subagent taxonomy, and the complete maintenance-scan pipeline that feeds the hopper — every piece adapted for this monorepo (uv + `creek-tools/` working dir + `crawdad/`, no frontend).

### Subagent fleet (`.claude/agents/`)
- **chief-architect** (fable) plans; **ralph-worker** (opus) conducts one issue per worktree; 6 specialists (test/implementation/security/performance/documentation/dependency-review) build; **code-review-orchestrator** runs the Gate 2.5 pre-push self-review.
- `shared/house-rules.md` filled in with the real stack (Typer CLI + MCP server + CrawDad) and real thresholds: ≥90% branch coverage (per-file ≥80%, waiver floor 65%), interrogate ≥95%, complexity ≤10, mypy strict, ruff zero, pylint ≥9.0.
- Security/perf scopes reframed to this codebase: MCP bearer auth, vault path traversal, ingest input validation; O(n²) link passes and 35k-fragment scale.

### Parallel fleet mechanics (`scripts/ralph/`)
- `fleet.sh` (worktree pool, default 4 lanes), `pr-ready.sh` (deterministic merge-readiness token), `FLEET.md`, and three offline shell test suites — **62/62 passing locally**.
- `pick-next.sh` upgraded to the fleet-aware priority picker (P0–P3 + `priority-*` tiers, `solo`/`parallelizable`, same-epic guard — works with this repo's `epic:*` labels).
- `PROMPT.md` + `/ralph-tick` upgraded from the single-worker versions to the fleet-orchestrator contract; `state.json` gains the deslop-cadence + fleet keys (live counters preserved).

### Maintenance scans → the hopper
- Reusable `_claude-scan.yml` core (locked-uv install mirroring ci.yml) + 12 thin wrappers: bugs, complexity, coverage, dead-code, deps, docs, groom, mutation, perf, security, todo, types. **a11y dropped** — no frontend surface. mutmut installs ad hoc in the scan toolbox (not a repo dep).
- `hopper.yml`: hourly queue-depth guard that dispatches the most-stale producer scan when the `agent-ready` runway drops below 12.
[...]

PR #807, opened 2026-07-07
fix(ralph): honor die() exit code, case-fold epic guard, lock cmd_assign (#797)

### LINKED ISSUE #797: Ralph fleet script nits: die() exit code, epic-label case, assign flock, stale filename comments, FLEET.md link
## Role
Engineer sweeping small correctness/consistency nits in the Ralph fleet scripts.

## Goal
Fix five low-severity nits from the PR #792 review in one pass.

## Context / Requested changes
1. `scripts/ralph/fleet.sh` `die()` silently drops its documented second (exit-code) argument — always exits 1 even when called `die "..." 2`, contradicting its own contract comment.
2. `scripts/ralph/pick-next.sh`'s same-epic guard compares label text case-sensitively while `has_label()` lowercases — `Epic-42` vs `epic-42` can defeat the guard.
3. `fleet.sh cmd_assign`'s cap-check-then-`git worktree add` is not lock-guarded (TOCTOU). Sequential-orchestrator usage makes it unreachable today; add a cheap `flock` (or `mkdir` lock, macOS-safe) for future-proofing.
4. Stale comments in `iteration-trigger.yml` and `dependabot-to-ralph-issue.yml` still reference `claude-code-review.yml` by its old well-worn-tools filename instead of `code-review.yml`.
5. Link `scripts/ralph/FLEET.md` from `fleet.sh`'s header comment (the design doc resolves the concurrency-model questions a reader of the script alone will have).

## Test idea
Extend `test_fleet.sh` with a `die`-exit-code assertion and `test_pick_next.sh` with a mixed-case epic-label case; the rest are comment/doc changes verified by grep.

## Constraints
bash 3.2-safe; keep all 62 existing shell tests green.

Follow-up from #792 — https://github.com/Geoffe-Ga/Creek-Vault/pull/792#issuecomment-4906558689

Geoffe-Ga:
## Summary
- `fleet.sh die()` now honors its documented optional second (exit-code) argument and prints only `$1`, so the code no longer leaks into the message; `reconcile`'s "gh CLI required" path exits `2` as its contract promised.
- `pick-next.sh epic_labels()` lowercases epic labels before `sort -u`, making the same-epic guard's `comm -12` intersection case-insensitive so `Epic-42` vs `epic-42` can no longer defeat it.
- `fleet.sh cmd_assign()` wraps its cap-check → `git worktree add` critical section in an atomic mkdir-based lock (macOS/bash-3.2-safe; no `flock`) to close a TOCTOU, released on every exit path via an EXIT trap.
- Fixed stale `claude-code-review.yml` comment references → `code-review.yml` in `iteration-trigger.yml` and `dependabot-to-ralph-issue.yml` (comment-only), and linked `scripts/ralph/FLEET.md` from `fleet.sh`'s header (widening `usage()`'s sed range to keep help complete).

## Test plan
- Added RED-first tests: die exit-code + message-purity + lock-hygiene checks in `test_fleet.sh`; a mixed-case epic-guard scenario in `test_pick_next.sh`. Confirmed each failed for the right reason, then went green.
- `test_fleet.sh` (31), `test_pick_next.sh` (24), `test_pr_ready.sh` (15) all pass; `test_fleet.sh` also passes under system `/bin/bash` 3.2 (bash 3.2 compatibility).
- `cd creek-tools && ./scripts/check-all.sh` exits 0 (12/12) — no Python was touched.
- Verdict from pre-push self-review: CLEAN.

Closes #797
Refs #792

### DIFF
--- a/.github/workflows/dependabot-to-ralph-issue.yml
+++ b/.github/workflows/dependabot-to-ralph-issue.yml
@@ -65,7 +65,7 @@ jobs:
 
       - name: Bridge Dependabot PRs into Ralph issues
         env:
-          # Prefer a real PAT (same convention as claude-code-review.yml and
+          # Prefer a real PAT (same convention as code-review.yml and
           # deslop.yml) so the issue/PR edits are authored by an account
           # the mobile webhook recognizes; fall back to the run's GITHUB_TOKEN.
           GH_TOKEN: ${{ secrets.GEOFFE_GA_PAT || secrets.GITHUB_TOKEN }}
--- a/.github/workflows/iteration-trigger.yml
+++ b/.github/workflows/iteration-trigger.yml
@@ -67,7 +67,7 @@ jobs:
           fi
 
           # Find the latest Claude review comment via the deterministic
-          # "## Verdict: <verdict>" line that claude-code-review.yml appends.
+          # "## Verdict: <verdict>" line that code-review.yml appends.
           CLAUDE=$(jq -c '[.[] | select(.body | test("(^|\\n)## Verdict:"))] | last // empty' comments.json)
           if [ -z "$CLAUDE" ]; then
             echo "No Claude review yet - skipping"
--- a/scripts/ralph/fleet.sh
+++ b/scripts/ralph/fleet.sh
@@ -10,6 +10,8 @@
 # inspect, sync, and tear down those worktrees — never more than
 # `max_workers` (default 4) at a time.
 #
+# Design & concurrency model: scripts/ralph/FLEET.md
+#
 # Design contract ("optimistic parallelism, pessimistic merge"):
 #   * Parallel work is a speculation that the chosen issues are independent.
 #   * Correctness is guaranteed at MERGE time, not pick time: the orchestrator
@@ -50,9 +52,12 @@ readonly DEFAULT_MAX_WORKERS=4
 readonly WORKTREE_ROOT=".ralph/worktrees"
 readonly STATE_FILE="scripts/ralph/state.json"
 
+# Print an actionable error and exit. Callers pass the message as ONE arg and an
+# optional exit code as the second ($2) — so the code never leaks into the
+# message. Defaults to exit 1 for the common single-arg callers.
 die() {
-  echo "fleet: $*" >&2
-  exit 1
+  echo "fleet: $1" >&2
+  exit "${2:-1}"
 }
 
 # Resolve the repository root so the script works from any worktree/subdir.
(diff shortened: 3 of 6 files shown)


Geoffe-Ga:
## Summary
This PR fixes three low-severity nits flagged in PR #792's review: `fleet.sh die()` now correctly separates its message ($1) from an optional exit-code arg ($2) instead of leaking the code into the printed message; `pick-next.sh epic_labels()` lowercases labels before dedup so the same-epic guard is case-insensitive; and `fleet.sh cmd_assign()` closes a TOCTOU between the worker-cap check and `git worktree add` with an atomic `mkdir`-based lock plus an EXIT trap. Two stale `claude-code-review.yml` comment references were corrected to the real `code-review.yml` filename, and `fleet.sh`'s header now links `FLEET.md`.

## Strengths
- Root-caused rather than papered over: `die()`'s old `echo "fleet: $*"` really did concatenate the exit code into the user-visible message; the fix (`$1` for message, `${2:-1}` for exit) is minimal and correct, and every existing call site was verified to pass a single message argument, so no call site regresses.
- The `cmd_assign` lock is genuinely careful: `mkdir` is atomic (portable, no `flock` dependency, explicitly justified for bash 3.2/macOS), the EXIT trap fires on every exit path including `die`'s `exit` call (verified: `set -euo pipefail` + single-subcommand-per-invocation `main()` dispatch means there's no risk of a stale/duplicate trap across calls), and the happy path clears the trap after releasing the lock to avoid a redundant `rmdir` later.
- Re-entrancy is preserved: the existing-worktree check happens *before* acquiring the lock, so re-entrant calls never contend for it.
- `epic_labels()` fix is precise — lowercasing before `sort -u` fixes both the case-insensitive intersection and the sort collation, and the new test (`epic_guard_mixed_case`) exercises exactly the `Epic-Foo` vs `epic-foo` scenario the bug allowed through.
- `usage()`'s `sed -n '2,48p'` range was widened in lockstep with the new header comment line, and I confirmed by direct inspection that line 48 is still the "Exit codes:" line — the help text doesn't get truncated or overrun.
- Verified the `code-review.yml` comment fixes are factually correct (`.github/workflows/code-review.yml` is indeed the real file; `claude-code-review.yml` doesn't exist in the repo).

## Security Concerns
None. 🟢 LOW — this is internal tooling (no untrusted input, no network-facing surface); the lock path is a fixed, repo-relative directory name, not derived from user input.

## Problems
None found that block merge.

## Code Quality
- Minor: `acquire_assign_lock`'s polling loop uses `sleep 1` per retry (up to `FLEET_LOCK_TIMEOUT`, default 10s) — coarse-grained but adequate for this use case (low contention, human-scale operations); not worth the complexity of a finer-grained backoff.
- `repo_root()` (unchanged by this PR) resolves via `git rev-parse --show-toplevel`, which would return a worktree's own root rather than the main repo root if `cmd_assign` were ever invoked from inside a spawned worktree rather than the orchestrator's main checkout — the lock path is built from `$root`, so that scenario could bypass the new lock. This is pre-existing behavior outside this PR's diff and consistent with how `assign` is documented to be called (by the orchestrator, not from within a worker's own worktree), so it's not a regression here — just worth keeping in mind if that invocation contract ever changes.

## CLAUDE.md Adherence
✅ No Python files touched, so the coverage/mypy/ruff/docstring gates don't apply to this PR. ✅ Commit message follows Conventional Commits (`fix(ralph): ...`). ✅ Work done on a feature/issue branch, not directly on `main`. ✅ PR description documents a TDD (RED-first) test plan and references `./scripts/check-all.sh` passing, consistent with the repo's "Stay Green" workflow.

## Testing
New tests directly target each behavioral change: `test_fleet.sh` adds explicit-exit-code, default-exit-code, message-purity, and three lock-hygiene checks (clean release on success, release on a refused/`die`d assign, and fail-fast behavior against a pre-existing stale lock); `test_pick_next.sh` adds the mixed-case epic-guard scenario. These are well-targeted at the exact edge cases the PR fixes (not just happy-path re-verification), which is what a fix-only PR like this needs.

## Documentation
Comment updates are appropriately scoped: the new `die()` and `acquire_assign_lock` comments explain the non-obvious *why* (trap fires on `exit` not `return`; message/code separation contract for callers) rather than restating the code. The `FLEET.md` cross-link in the header is a nice discoverability touch matched by a corresponding widened `usage()` sed range so the printed help stays complete.



## Verdict: LGTM

The change is a small, well-tested shell-script fix (die() exit-code handling, case-insensitive epic guard, TOCTOU lock for cmd_assign) with RED-first tests added for every behavior change and no Python touched, so the CLAUDE.md coverage/lint/type gates don't apply. I verified die() call sites, the epic_labels lowering, and the lock/trap logic by reading the full functions and confirmed the two workflow comment fixes point at the actual file name (code-review.yml) — nothing here blocks merge or warrants a substantive follow-up ask.

Geoffe-Ga:
**CI**: 8/11 Green
**VERDICT**: LGTM
**Action**: pull comment 4909397129 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 14/14 Green
**VERDICT**: LGTM

PR #818, opened 2026-07-08
fix(ci): grant scan wrappers the token scopes the reusable core requests

Geoffe-Ga:
## Summary
All 11 core-based scan wrappers startup_failed on every trigger (cron + hopper dispatch) since the #792 port: a reusable workflow may not request more GITHUB_TOKEN scopes than its caller grants, the repo default is read-only, and the wrappers declared no permissions. This grants contents:read / issues:write / id-token:write at the caller job level, matching _claude-scan.yml's declared needs. scan-groom/deslop are standalone and unaffected.

Note: the same latent break exists in the well-worn-tools template.

## Test plan
- yaml parse + actionlint clean on all wrappers
- After merge: workflow_dispatch a scan and confirm it reaches the job (validated live below)

### DIFF
--- a/.github/workflows/scan-bugs.yml
+++ b/.github/workflows/scan-bugs.yml
@@ -19,6 +19,13 @@ on:
 
 jobs:
   run:
+    # A reusable workflow may not request more GITHUB_TOKEN scopes than its
+    # caller grants; the repo default is read-only, so grant the core's needs
+    # here (contents: read / issues: write / id-token: write) explicitly.
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
     uses: ./.github/workflows/_claude-scan.yml
     with:
       scan_name: bugs
--- a/.github/workflows/scan-complexity.yml
+++ b/.github/workflows/scan-complexity.yml
@@ -19,6 +19,13 @@ on:
 
 jobs:
   run:
+    # A reusable workflow may not request more GITHUB_TOKEN scopes than its
+    # caller grants; the repo default is read-only, so grant the core's needs
+    # here (contents: read / issues: write / id-token: write) explicitly.
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
     uses: ./.github/workflows/_claude-scan.yml
     with:
       scan_name: complexity
--- a/.github/workflows/scan-coverage.yml
+++ b/.github/workflows/scan-coverage.yml
@@ -19,6 +19,13 @@ on:
 
 jobs:
   run:
+    # A reusable workflow may not request more GITHUB_TOKEN scopes than its
+    # caller grants; the repo default is read-only, so grant the core's needs
+    # here (contents: read / issues: write / id-token: write) explicitly.
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
     uses: ./.github/workflows/_claude-scan.yml
     with:
       scan_name: coverage
--- a/.github/workflows/scan-dead-code.yml
+++ b/.github/workflows/scan-dead-code.yml
@@ -19,6 +19,13 @@ on:
 
 jobs:
   run:
+    # A reusable workflow may not request more GITHUB_TOKEN scopes than its
+    # caller grants; the repo default is read-only, so grant the core's needs
+    # here (contents: read / issues: write / id-token: write) explicitly.
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
     uses: ./.github/workflows/_claude-scan.yml
     with:
       scan_name: dead-code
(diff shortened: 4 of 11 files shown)


Geoffe-Ga:
## Summary
This PR fixes a startup failure affecting all 11 `_claude-scan.yml`-based scan wrapper workflows (`scan-bugs`, `scan-complexity`, `scan-coverage`, `scan-dead-code`, `scan-deps`, `scan-docs`, `scan-mutation`, `scan-perf`, `scan-security`, `scan-todo`, `scan-types`). Each wrapper now declares an explicit job-level `permissions:` block (`contents: read`, `issues: write`, `id-token: write`) matching what the reusable `_claude-scan.yml` core requests, since GitHub Actions forbids a reusable workflow from requesting more `GITHUB_TOKEN` scope than its caller grants, and the repo's default token permissions are read-only.

## Strengths
- Root-cause diagnosis is correct and precisely scoped: GitHub's reusable-workflow permission-escalation rule is exactly why an unscoped caller + a scoped callee (`_claude-scan.yml` declares `contents: read` / `issues: write` / `id-token: write`) fails at dispatch time.
- The fix is applied identically and completely across all 11 affected wrappers — no stragglers, no partial fixes.
- Correctly identifies and excludes `scan-groom.yml` and `deslop.yml`, which already declare their own top-level `permissions:` blocks and don't route through `_claude-scan.yml`.
- Each added block includes an inline comment explaining *why* the permissions are needed, which is genuinely useful context for future readers of these thin wrapper files.
- Grants the minimum needed scopes (no excess privileges) — matches the reusable core's actual `secrets`/output needs (issue creation via `issues: write`, OIDC via `id-token: write`).
- PR description is thorough: explains the mechanism, notes the parallel latent bug in the "well-worn-tools template," and documents a test plan (YAML parse + actionlint, plus live post-merge validation).

## Security Concerns
🟢 LOW — No security regressions. The granted scopes (`contents: read`, `issues: write`, `id-token: write`) are the minimum the reusable workflow needs and match the least-privilege scoping already used by the sibling `scan-groom.yml`/`deslop.yml` workflows. No new write-to-code or admin scopes are introduced.

## Problems
None found. The diff is mechanical, uniform, and consistent with the existing `_claude-scan.yml` contract.

## Code Quality
- Very minor duplication: the same 7-line permissions block (4 lines of comment + 3 lines of YAML) is now repeated verbatim in 11 files. This is consistent with the existing "thin wrapper, clone-and-tweak" pattern already used in this file family (per the wrappers' own header comments), so it's not a new pattern — but if a 12th scan wrapper is added later, consider whether the reusable core itself could declare `permissions` in a way that's inherited, or whether a short doc note in `_claude-scan.yml`'s header could reduce the copy-paste surface. Not a blocker.

## CLAUDE.md Adherence
✅ No Python code is touched, so the `creek-tools` coverage/mypy/ruff/complexity gates are not applicable here.
✅ Follows the "Always use scripts" spirit in intent (test plan references actionlint), though this is a workflow-only change outside `creek-tools/`.
✅ Uses a feature branch (`fix/scan-wrapper-permissions`) rather than committing to `main`, consistent with repo conventions.
N/A Commit conventions — PR title follows Conventional Commits (`fix(ci): ...`).

## Testing
The PR body states the change was validated via YAML parsing and `actionlint`, plus a live `workflow_dispatch` post-merge check. There is no automated CI gate in this repo that lints GitHub Actions workflow files (no `actionlint`/`yamllint` step found in `.github/workflows/ci.yml` or elsewhere), so this fix is inherently validated by manual/local checks and live dispatch rather than an automated test suite — appropriate given the nature of the change (declarative CI config, not application code) and consistent with how CI-configuration fixes are typically verified in this kind of repo.

## Documentation
Each modified file carries a clear inline comment explaining the "why" (GitHub's reusable-workflow scope-escalation rule), which is exactly the kind of non-obvious constraint worth documenting per this repo's own comment philosophy. The PR description itself is a model of clarity: problem, root cause, scope, and exclusions are all called out explicitly.



## Verdict: LGTM

The change is a minimal, mechanically-uniform fix (identical 7-line `permissions:` block added to 11 wrapper workflows) that correctly addresses a real, well-diagnosed bug — GitHub's rule that a reusable workflow cannot request more token scopes than its caller grants. The added permissions exactly match what `_claude-scan.yml` declares, the YAML is well-formed and consistently indented, and the PR body documents a live validation plan. Nothing here is blocking; there's no code (Python) surface for creek-tools' coverage/lint/mypy gates to apply to.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 4916569929 to see in-depth feedback and continue iterating

Issue #861, opened 2026-07-17
CI: pip-audit fails all Code Quality jobs — 15 known vulnerabilities in 5 packages


Geoffe-Ga:
## Problem

The `Code Quality & Testing (3.11/3.12/3.13)` jobs (and therefore `Quality Gate`) currently fail on **every run** at the security-audit step:

```
Found 15 known vulnerabilities in 5 packages
##[error]Process completed with exit code 1.
```

Observed on PR #860 (a CI/config-only PR touching no Python dependencies): https://github.com/Geoffe-Ga/Creek-Vault/actions/runs/29557787717 — pylint (9.70 ≥ 9.0), ruff (450 files clean), and the CrawDad suites all pass; only pip-audit is red. This is time-based advisory drift (new CVE/GHSA advisories published against pinned versions), so `main` fails identically — no branch can go green until the dependencies move.

## Fix (upgrade, don't suppress)

1. Run `pip-audit -r creek-tools/requirements*.txt` (or the CI invocation) locally to get the exact 5 package names + advisory IDs.
2. For each, check the **live** PyPI registry for the current fixed version (training-data/version intuitions are stale) and upgrade the pin.
3. Only if an upgrade path is genuinely unavailable (e.g., fix unreleased): a narrowly-scoped, commented ignore with the advisory ID and a removal condition — last resort, not the default.

## Acceptance criteria

- `pip-audit` step exits 0 on `main`.
- All version bumps are real upgrades (no blanket `--ignore-vuln` sweeps).
- Test suite still green at the new pins.

Geoffe-Ga:
Update from PR #866's CI run ([run 29849491275](https://github.com/Geoffe-Ga/Creek-Vault/actions/runs/29849491275), 2026-07-21): pip-audit now reports **28 known vulnerabilities in 6 packages** (up from 15/5 when this was filed). Current offenders per the uploaded `pip-audit-report.json`:

- `pillow 12.2.0` — 20 advisories (see #843)
- `mcp 1.27.1` — CVE-2026-52869/-52870, CVE-2026-59950 (see #862)
- `soupsieve 2.8.3` — PYSEC-2026-3071/-3072 (see #836)
- `httplib2 0.31.2` — PYSEC-2026-3444 (see #844)
- `setuptools 81.0.0` — PYSEC-2026-3447 (new since filing)
- `torch 2.12.1` — PYSEC-2025-194 (new since filing)

[...]

PR #884, opened 2026-07-22
fix(mcp): confine creek.ingest input_path to the vault root (#819)

### LINKED ISSUE #819: [scan:security] creek.ingest input_path is unconfined — remote MCP consumer can read arbitrary host files into the vault
## Role
suppressions).

## Goal
Confine `creek.ingest`'s client-supplied `input_path` to the vault root for
**remote (network-authenticated) callers**, so an authenticated remote consumer
can no longer read arbitrary host files into the vault. Verified by a failing
test that a remote-context `ingest_tool(..., input_path="/etc/passwd")` (or a
`../` escape) is refused before any bytes are read, while local (stdio) ingest of
external paths keeps working unchanged.

## Context
- File(s): `creek-tools/creek_mcp/tools/ingest.py:65-84` (the unconfined
  `source = Path(input_path)` → `.exists()` → `ingestor_cls().ingest(source)`),
  reached from `creek-tools/creek_mcp/server.py:434-447` (the `creek.ingest`
  tool). The remote transport is wired in `server.py:678-690` and the only
  remote gate is the tier-ceiling check in `_BoundedFastMCP.call_tool`
  (`server.py:109-138`) — there is **no path confinement**.
- Scanned at commit: `668d568b497c536757b27d2efedd73952b657bba` — re-verify
  against HEAD before starting.
- Evidence:
  - `ingest.py:65` — `source = Path(input_path)` uses the raw client string with
    no `resolve()` / `relative_to(vault_root)` confinement; `ingest.py:74`
    reads and persists its contents as a vault fragment.
  - Reachability over the authenticated network transport:
    1. `remote_auth.py` requires a bearer token → the caller is an
       authenticated **consumer**, but consumer ≠ trusted-to-read-host-disk.
    2. `server.py:73-75` `_REMOTE_ADMITTED_CEILINGS = {OPEN, PERSONAL}` — a
       remote caller may request `privacy_tier_ceiling="personal"`.
    3. `ingest.py:25` `DEFAULT_INGEST_TIER = PERSONAL`, and
       `tier_ceiling.write_tier_allowed(PERSONAL, PERSONAL)` → `1 <= 1` → `True`
       (`tier_ceiling.py:50-85`), so the tier gate **admits** the call.
    4. `input_path` is unconfined → `Path("/etc/passwd")` / `~/.ssh/id_rsa` /
       any host file `.exists()` → its content is ingested into
       `01-Fragments/` and becomes readable/retrievable vault content.
  - Corroboration (project's own stated intent): the sibling read-only tool
    `creek_mcp/tools/redact.py:107-125` `_resolve_within_vault()` deliberately
    confines its `input_path` "to prevent the MCP surface from scanning
    arbitrary disk content" — yet `ingest`, which *reads and persists* full file
    bodies (strictly more dangerous), performs no such confinement. The
    asymmetry is the tell that this is an oversight, not a design choice.
  - `journal.py` (`_safe_stem`), `compile` engine (`_resolve_target_path`,
    `engine.py:492-513`), and `attachments.py` (`sanitize_filename`) all confine
    their client-supplied path components; `ingest` is the lone gap.
- Related: sibling of the FEAT-010/011 MCP tool surface and the #759
  remote-auth boundary; mirror `redact.py`'s `_resolve_within_vault` helper.

## Output Format
"Closes #N".

## Examples
Before (`ingest.py`, remote caller reaches this with any host path):
```python
    source = Path(input_path)
    if not source.exists():
        return refusal_response(..., reason=f"input path not found: {input_path}")
    writer = VaultWriter(vault_path=vault_path)
    ingest_result = ingestor_cls().ingest(source)   # reads /etc/passwd
```
After (confine for remote callers before any read — mirror redact.py):
```python
    # Remote (network-authed) callers may only ingest paths inside the vault;
    # local stdio callers keep full-disk ingest of external exports.
    if _is_remote_call():                       # get_access_token() is not None
        source = _resolve_within_vault(vault_path, input_path)
        if source is None:
            return refusal_response(
                ..., reason=f"input_path {input_path!r} resolves outside the vault",
            )
    else:
        source = Path(input_path)
    if not source.exists():
        return refusal_response(..., reason=f"input path not found: {input_path}")
```
Factor the vault-confinement helper so `ingest` and `redact` share one
implementation rather than diverging.

## Constraints
- Preserve local (stdio) behaviour: ingesting external export paths from a
  local operator must keep working — the confinement applies to the remote,
  token-authenticated transport only.

Geoffe-Ga:
## Summary
- Close an arbitrary-host-file-read vector: `creek.ingest` built `Path(input_path)` from the raw client string and handed it to the ingestor, so a remote MCP consumer could read any host file into the vault (the tier-ceiling check was the only gate).
- Route `input_path` through a new shared `creek_mcp/path_confinement.py:resolve_within_vault` helper — resolves the path (collapsing `..`, following symlinks) and refuses anything outside the vault root. The check runs after the tier-ceiling and unknown-`source_type` gates and before existence; the *resolved* path (not the raw string) is what gets existence-checked and ingested, so there is no re-derivation/TOCTOU gap.
- Extract that confinement logic as the single source of truth and refactor `creek.redact.scan` to import it (deleting its private duplicate) so the two consumers can never drift.

## Test plan
- New `tests/test_path_confinement.py` — dedicated unit tests for the SSOT helper: in-vault absolute/relative/vault-root accepted; absolute-outside, `..`-traversal, and escaping-symlink rejected; in-vault symlink accepted.
- New confinement tests in `tests/test_mcp_write_tools.py` — `creek.ingest` refuses absolute-outside / `..`-traversal / escaping-symlink paths, does not invoke the ingestor on a confinement refusal, treats a relative `input_path` as vault-relative, and still checks the tier-ceiling gate first.
- `cd creek-tools && ./scripts/check-all.sh` green except the pre-existing `pip-audit` failure (transitive CVEs in pillow/pyasn1/setuptools/soupsieve/torch, red on all branches, tracked in #861 — unrelated; this change touches no dependencies).

## Security follow-ups (deliberately out of scope, to keep this fix minimal)
- Audit the confinement-refusal branch specifically (an out-of-vault `input_path` is an attempted arbitrary-file-read probe worth logging for forensics), while leaving not-found / unknown-source silent.
- Pre-existing (not introduced here): `redact_scan_tool`'s ok-path calls `resolved.relative_to(vault_path)` against the *unresolved* `vault_path`; if a caller ever passes an unresolved vault root this raises rather than returning a structured response (fails closed). Worth a tracking ticket.

Closes #819

### DIFF
--- a/creek-tools/creek_mcp/path_confinement.py
+++ b/creek-tools/creek_mcp/path_confinement.py
@@ -1,0 +1,56 @@
+"""Shared vault-root path confinement for the MCP tool surface.
+
+MCP tools accept an *input_path* string from an untrusted caller. To stop
+the surface from reading or ingesting arbitrary disk content, every tool
+that touches a caller-supplied path routes it through
+:func:`resolve_within_vault`, which confines the path to the vault root.
+
+The confinement contract:
+
+* An absolute path is accepted only if it resolves *inside* the vault.
+* A relative path is joined under the resolved vault root (never the
+  process cwd) before resolution.
+* Resolution collapses ``..`` segments and follows symlinks, so neither a
+  ``..`` traversal nor an in-vault symlink pointing outside can escape.
+* Paths that resolve outside the vault return ``None`` for the caller to
+  turn into a structured refusal.
+
+Existence is intentionally *not* checked here (``resolve(strict=False)``
+semantics) so callers can distinguish an outside-the-vault refusal from a
+not-found refusal with their own, clearer messages.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+
+def resolve_within_vault(vault_path: Path, input_path: str) -> Path | None:
+    """Return *input_path* resolved inside *vault_path*, or ``None`` if outside.
+
+    Accepts either an absolute path inside the vault or a vault-relative
+    path. Uses ``Path.resolve(strict=False)`` so non-existent staging
+    directories still validate; existence is checked separately by the
+    caller so it can emit a clean ``not found`` message instead of a
+    silent ``None`` collapse.
+
+    Args:
+        vault_path: Vault root the path must resolve inside of.
+        input_path: Caller-supplied path; absolute or relative to the
+            vault root.
+
+    Returns:
+        The resolved absolute path when it lies inside the vault, or
+        ``None`` when it escapes (after collapsing ``..`` and following
+        symlinks).
+    """
+    vault_resolved = vault_path.resolve()
+    candidate = Path(input_path)
+    if not candidate.is_absolute():
+        candidate = vault_resolved / candidate
+    resolved = candidate.resolve()
+    try:
+        resolved.relative_to(vault_resolved)
+    except ValueError:
+        return None
+    return resolved
--- a/creek-tools/creek_mcp/tools/ingest.py
+++ b/creek-tools/creek_mcp/tools/ingest.py
@@ -15,6 +15,7 @@ from creek.ingest import INGESTOR_REGISTRY, assemble_ingested_fragment
 from creek.models import PrivacyTier
 from creek.vault.writer import VaultWriter
 from creek_mcp.audit import MCPAuditLog
+from creek_mcp.path_confinement import resolve_within_vault
 from creek_mcp.tier_ceiling import (
     TierCeiling,
     refusal_response,
@@ -39,6 +40,22 @@ def ingest_tool(
     callers with ``ceiling=open`` cannot run an ingest because the
     fragments would exceed the ceiling. The wrapper refuses *before*
     touching the source file so no side effects occur on rejection.
+
+    Args:
+        vault_path: Vault root used for path validation and audit logging.
+        source_type: Registry key selecting the ingestor to run.
+        input_path: Source to ingest; may be absolute or relative to the
+            vault root. The path must resolve *inside* the vault to prevent
+            the MCP surface from ingesting arbitrary disk content. Refusals
+            are checked in order: tier ceiling, unknown ``source_type``,
+            path confinement, then existence.
+        privacy_tier_ceiling: Caller's tier ceiling; the personal-tier
+            default is refused when the ceiling is lower.
+        consumer: Identifier of the calling client (recorded in audit).
+
+    Returns:
+        A dict with ``status`` (``ok`` / ``refused``), the count and ids of
+        written fragments, and any per-fragment errors.
     """
     if not write_tier_allowed(DEFAULT_INGEST_TIER, privacy_tier_ceiling):
         MCPAuditLog(vault_path).append(
@@ -62,8 +79,17 @@ def ingest_tool(
             ceiling=privacy_tier_ceiling,
             reason=f"unknown source_type {source_type!r}",
         )
-    source = Path(input_path)
-    if not source.exists():
+    resolved = resolve_within_vault(vault_path, input_path)
+    if resolved is None:
+        return refusal_response(
+            tool=TOOL_NAME,
+            ceiling=privacy_tier_ceiling,
+            reason=(
+                f"input_path {input_path!r} resolves outside the vault root; "
+                "the ingest tool only operates on vault-relative paths."
+            ),
+        )
+    if not resolved.exists():
         return refusal_response(
             tool=TOOL_NAME,
             ceiling=privacy_tier_ceiling,
@@ -71,7 +97,7 @@ def ingest_tool(
         )
 
     writer = VaultWriter(vault_path=vault_path)
-    ingest_result = ingestor_cls().ingest(source)
+    ingest_result = ingestor_cls().ingest(resolved)
     written_ids: list[str] = []
     errors: list[str] = list(ingest_result.errors)
     for parsed in ingest_result.fragments:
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
PR #884 closes an arbitrary-host-file-read vulnerability in `creek.ingest`, where a raw client-supplied `input_path` was turned directly into a `Path` and handed to the ingestor with no confinement to the vault root. It extracts the vault-confinement logic (previously private and duplicated inside `creek_mcp/tools/redact.py`) into a shared `creek_mcp/path_confinement.py:resolve_within_vault` helper, wires it into `creek.ingest`, and refactors `creek.redact.scan` to consume the same helper instead of its own copy.

## Strengths
- **Correct fix, no TOCTOU gap**: `resolve_within_vault` resolves the candidate path once (collapsing `..`, following symlinks) and the *same* `resolved` `Path` object is reused for the existence check and passed to the ingestor — there's no re-derivation from the raw string after validation.
- **DRY**: genuinely duplicated logic (the old private `_resolve_within_vault` in `redact.py`) is deleted and replaced with a single shared source of truth, per CLAUDE.md §1.2.
- **Check ordering is deliberate and documented**: tier-ceiling → unknown `source_type` → path confinement → existence, matching both the docstring and the code, and covered by a dedicated test (`test_ingest_tier_ceiling_checked_before_confinement`).
- **No side effects on refusal**: `test_ingest_confinement_refusal_skips_ingestor_invocation` proves the ingestor is never touched when the path escapes the vault (via a raising stub ingestor).
- **Thorough test matrix**: absolute-outside, `..`-traversal, escaping symlink, and vault-root-itself are all covered at the helper level (`test_path_confinement.py`) and re-verified end-to-end at the tool level (`test_mcp_write_tools.py`), including that a relative `input_path` binds to the vault root and not the process cwd.
- **Honest PR description**: explicitly calls out two things left out of scope — logging confinement refusals for forensics, and a pre-existing (unrelated) `relative_to` bug in `redact.py`'s ok-path — rather than silently leaving them or overclaiming completeness.

## Security Concerns
No blocking vulnerabilities found in the new logic.
- 🟢 LOW: Confinement refusals in `creek.ingest` are not written to `MCPAuditLog` (only the tier-ceiling refusal is). An out-of-vault `input_path` is effectively an attempted file-read probe and would be useful to have in the audit trail for forensics. The author already flagged this explicitly as a deliberate, out-of-scope follow-up in the PR description, so this isn't a new gap being introduced — just noting it's worth actually filing as a tracked issue if it isn't already.
- 🟢 LOW (pre-existing, not introduced here): `redact_scan_tool`'s ok-path does `resolved.relative_to(vault_path)` against the *unresolved* `vault_path` (`creek_mcp/tools/redact.py:98`) while `resolved` is fully resolved — if a caller ever passes an unresolved vault root this could raise instead of returning a structured response. Untouched by this diff; already called out by the author as a follow-up.

## Problems
None found that block merge.

## Code Quality
- Non-blocking: with `input_path=""` (or any path that resolves to the vault root itself), `resolve_within_vault` returns the vault root, `resolved.exists()` is `True`, and `ingestor_cls().ingest(resolved)` — which is not wrapped in try/except — would be called on a directory. This isn't a security escape (still confined to the vault), and the same gap existed for any directory path before this PR, so it's not a regression; just an edge case worth a follow-up if it ever surfaces in practice.
- Moving the `Path` import in `redact.py` under `TYPE_CHECKING` (since `from __future__ import annotations` is in effect and `Path` is now only used in annotations) is a clean, correct cleanup enabled by extracting the helper.

## CLAUDE.md Adherence
- ✅ DRY / single source of truth (§1.2): duplicate confinement logic eliminated.
- ✅ Docstrings: new module and both modified tool functions have Google-style docstrings with Args/Returns, including the refusal-order contract.
- ✅ Type hints: fully typed, consistent with the rest of the codebase's `from __future__ import annotations` + strict-mypy style.
- ✅ Tests appear TDD-appropriate: dedicated unit tests for the new SSOT helper plus integration tests at the tool boundary, matching the "Testing Strategy" expectations in `creek-tools/CLAUDE.md`.
- ✅ Commit/PR title follows Conventional Commits (`fix(mcp): ...`) and links the closing issue (#819).
- Unable to execute `./scripts/check-all.sh` in this review sandbox (bash execution was blocked), so lint/type/coverage/complexity gates were verified by static reading only, not by running the scripts. The PR description states `check-all.sh` is green except a documented pre-existing `pip-audit` failure unrelated to this change (tracked in #861).

## Testing
Strong. The helper is tested directly (7 focused cases: absolute inside, relative-binds-to-vault, vault-root-itself, absolute outside, `..` traversal, symlink escaping, symlink staying inside) and the tool wrapper is tested end-to-end for the same escape vectors plus check-ordering and no-side-effects-on-refusal. Nothing insufficient here.

## Documentation
Good. The new module's docstring clearly states the confinement contract (absolute-must-resolve-inside, relative-joins-under-vault-root, `..`/symlink handling, existence deliberately not checked here). The `ingest_tool` docstring was updated to document the new `input_path` semantics and refusal ordering, matching the implementation exactly.



## Verdict: LGTM

This is a minimal, well-scoped security fix (arbitrary host-file-read via creek.ingest, issue #819) with a correct confinement helper, no TOCTOU/re-derivation gap, and thorough unit + integration tests covering absolute/relative/dot-dot/symlink escape cases and check-ordering. No bugs, security gaps, or CLAUDE.md violations found that would block merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5052456533 to see in-depth feedback and continue iterating

Issue #901, opened 2026-07-23
docs(redact): refresh patterns.py false-positive notes for span-merge overlap semantics


Geoffe-Ga:
Doc-accuracy follow-up from #832.

Several `false_positive_notes` in `creek-tools/creek/redact/patterns.py` describe pre-#832 overlap behavior, e.g. `stripe_key`: "The api_key pattern overlaps for the `sk_test_` prefix; both would fire, which is acceptable since the action is the same." Under the span-based single pass introduced by #832, truly overlapping matches now merge into ONE union labeled by the highest-severity contributor — the patterns no longer both fire independently on apply. Update the affected notes (`stripe_key`, `anthropic_key`, `openai_project_key`, `discord_bot_token`/`jwt`) to describe the merge semantics.

Documentation only; no behavior change.

Refs #832.

Issue #900, opened 2026-07-23
redact: scan/apply parity gap for custom line-anchored patterns (per-line scanner vs full-text redactor)


Geoffe-Ga:
Follow-up from the #832 security review (pre-existing asymmetry, not introduced by the fix).

`RedactionScanner.scan_file` scans `text.splitlines()` line-by-line, while `Redactor.redact_content` runs `finditer` over the whole multi-line content. These coincide for all built-in patterns (char classes exclude `\n`), but a **custom** pattern using `$`, `^`, inline `(?m)`, or DOTALL can be flagged per-line by `--scan` yet missed by the full-text `--apply` (e.g. `secret\d+$` matches at each line-end for the scanner but only at end-of-document for the redactor), leaving cleartext after apply — violating the scan/apply parity guarantee the module documents.

Fix options: have the redactor collect spans per-line (mirroring the scanner, mapping line offsets back to document offsets), or document the constraint on custom patterns and validate/reject line-anchored custom regexes at config load.

Refs #832.

Issue #899, opened 2026-07-23
redact: fold high-entropy detection into the span-based single pass


Geoffe-Ga:
Follow-up from #832 (span-based redaction rewrite) and its security review.

`Redactor._replace_high_entropy` still runs as a second pass on the post-splice (already-mutated) text — the same "anchor on already-redacted text" anti-pattern #832 fixed for the regex phase, narrowed to entropy-vs-regex overlaps:

1. **Entropy bisection**: if a >=20-char high-entropy blob is bisected by an interior regex redaction, the remaining halves can each fall under the 20-char candidate floor and escape the entropy pass, leaving sub-threshold cleartext fragments. Not exploitable with built-in patterns today, but a custom interior-matching pattern could trigger it.
2. **Marker mangling**: the marker name `email_password_combo` is exactly 20 chars with Shannon entropy ~3.68; at `min_confidence` below ~0.59 the entropy pass rewrites `[REDACTED:email_password_combo]` into `[REDACTED:[REDACTED:high_entropy_string]]`. No leak (content already gone), but it corrupts downstream marker parsing/telemetry. Default 0.6 is just safe (3.68 < 3.70).

Fix: collect entropy candidates against the ORIGINAL content and fold them into the same span set that gets merged and spliced (see `_collect_spans`/`_merge_spans` in `creek/redact/redactor.py`), or exclude text inside existing `[REDACTED:...]` markers from the entropy pass.

Refs #832.

PR #902, opened 2026-07-23
fix(redact): stop leaking password half of email:password combos on apply

### LINKED ISSUE #832: [scan:bugs] Redactor leaks password half of email:password combos on --apply (pattern-order bug)
## Role
suppressions).

## Goal
Ensure `Redactor.redact_content` never leaves the password half of an
`email:password` credential pair in cleartext, verified by a test that asserts
the secret is absent from the redacted output for input
`user@example.com:hunter2secret`.

## Context
- File(s): `creek-tools/creek/redact/redactor.py:101-108` (the sequential
  per-pattern replacement loop) and `creek-tools/creek/redact/patterns.py:75-106`
  (pattern ordering: `email` at line 75 precedes `email_password_combo` at line 99).
- Evidence: `redact_content` iterates patterns in `PATTERN_METADATA` insertion
  order and mutates `content` in place:
  ```python
  for name, pattern in patterns_to_use.items():
      marker = self.config.replacement_template.format(name=name)
      content = self._replace_pattern(content, pattern, marker, name)
  ```
  Because `email` (severity `medium`) runs before the critical
  `email_password_combo` pattern, the `email` regex matches `user@example.com`
  first (its `\b` boundary lands before the `:`), rewriting it to
  `[REDACTED:email]`. The broader `email_password_combo` pattern
  (`...@...:[^\s]+`) can then no longer match, so the password survives.
  Reproduced live at the scanned SHA:
  ```
  >>> from creek.config import RedactionConfig
  >>> from creek.redact.redactor import Redactor
  >>> Redactor(RedactionConfig(), salt=b'\x00'*16).redact_content(
  ...     'login user@example.com:hunter2secret here')
  'login [REDACTED:email]:hunter2secret here'   # 'hunter2secret' LEAKED
  ```
  The `RedactionScanner` scans each pattern independently against the original
  text, so `--scan` *does* flag the `email_password_combo` critical finding while
  `--apply` fails to remove it — violating the contract asserted in
  `_replace_high_entropy`'s own docstring ("a `--scan` finding cannot survive a
  `--apply` step"). This is a cleartext credential leak in the redaction
  pipeline.
- Related: sibling scan-run findings filed this run (purge wikilink RTBF gap,
  broken-links anchor false-positive, Google Drive version-suffix data loss).

## Output Format
"Closes #N".

## Examples
```python
# tests/test_redact.py
def test_apply_redacts_full_email_password_combo() -> None:
    r = Redactor(RedactionConfig(), salt=b"\x00" * 16)
    out = r.redact_content("login user@example.com:hunter2secret here")
    assert "hunter2secret" not in out          # fails today

# Fix sketch (redactor.py): collect match spans from every pattern against the
# ORIGINAL content in one pass (as RedactionScanner already does), resolve
# overlapping spans by keeping the widest / highest-severity one, then splice
# replacements right-to-left — OR order composite patterns
# (email_password_combo) ahead of their constituents (email) in PATTERN_METADATA.
```

## Constraints

Geoffe-Ga:
## Summary
- Fixes a P1 privacy bug: `Redactor.redact_content` applied patterns sequentially with in-place mutation, so the `email` pattern destroyed the anchor of the critical `email_password_combo` pattern and the password half of `user@example.com:hunter2secret` survived `--apply` in cleartext (while `--scan` correctly flagged it).
- Rewrites the replacement loop as a span-based single pass over the original content: collect spans from every in-scope pattern (allowlist + `post_validate` filtered), merge truly overlapping spans into unions, label each union with its highest-severity contributor (tie: widest span, then collection order), splice right-to-left. The high-entropy detector stays an unchanged second pass. Scan/apply parity is now a structural invariant — including for custom composite patterns, which reordering `PATTERN_METADATA` could never protect.
- Public API unchanged; `patterns.py`, `scanner.py`, and the CLI/file layers untouched; the removed `_replace_pattern` had no remaining references.

## Test plan
- New `TestPatternOrderLeak` class (7 tests) written RED-first: 5 failed at HEAD (password leak, combo marker, scan/apply parity, custom composite leak, partial-overlap union), 2 regression guards (standalone email marker, allowlisted-email-inside-combo). All green after the fix.
- All 260 redaction tests pass (`test_redact.py`, `test_properties_redaction.py`, `test_cli_redact.py`, `test_mcp_redact.py`, `test_redact_audit.py`).
- `cd creek-tools && ./scripts/check-all.sh` exits 0 (6028 tests; coverage, per-file, docstring, complexity, mypy strict, ruff all green).
- Security-specialist review of the rewritten pipeline: CLEAN; three minor pre-existing advisories filed as follow-ups #899, #900, #901.

Closes #832

### DIFF
--- a/creek-tools/creek/redact/redactor.py
+++ b/creek-tools/creek/redact/redactor.py
@@ -4,6 +4,12 @@ The :class:`Redactor` re-scans content to locate match positions (since
 :class:`RedactionMatch` intentionally does **not** store the matched text)
 and replaces each hit with a ``[REDACTED:type]`` marker.
 
+Replacement is a single pass over the *original* content: every in-scope
+pattern is matched against the untouched text, truly overlapping spans
+are unioned, and the merged spans are spliced out right-to-left. Because
+no pattern ever anchors on partially redacted text, a ``--scan`` finding
+cannot survive a ``--apply`` step (Issue #832).
+
 Redaction logs are written as JSON with a session salt in the header so
 that hashes can be correlated within a session but not reversed.
 """
@@ -11,10 +17,10 @@ that hashes can be correlated within a session but not reversed.
 import json
 import re
 from pathlib import Path
-from typing import Any
+from typing import Any, NamedTuple
 
 from creek.config import RedactionConfig
-from creek.redact.patterns import REDACTION_PATTERNS
+from creek.redact.patterns import PATTERN_METADATA, REDACTION_PATTERNS
 from creek.redact.scanner import (
     HIGH_ENTROPY_CANDIDATE,
     HIGH_ENTROPY_PATTERN_NAME,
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
Fixes a real P1 privacy bug: `Redactor.redact_content` used to apply patterns sequentially with in-place string mutation, so the `email` pattern could destroy the anchor text the higher-severity `email_password_combo` pattern needed, letting the password half of `user@example.com:hunter2secret` survive `--apply` in cleartext even though `--scan` correctly flagged it. The fix replaces the sequential loop with a single span-collection pass over the *original* content, unions truly-overlapping spans, labels each union by its most-severe contributor, and splices right-to-left. Public API, `patterns.py`, `scanner.py`, and CLI layers are untouched.

## Strengths
- Correct core algorithm: matching against the untouched original text for every in-scope pattern eliminates the root cause (anchor destruction from earlier substitutions) rather than patching around it.
- The interval-merge (`_merge_spans`) is a standard, correctly-implemented sweep: sorts by start, merges only on strict overlap (`start < prev.end`) so adjacent-but-distinct findings keep separate markers, and grows the merged end via `max(...)` so transitive overlaps (A↔C↔B) fold into one region — I traced this by hand against `test_partial_overlap_union_redacted` and it produces the single expected `[0, 30)` union.
- Severity-based marker selection (`_select_marker_name`) has a well-defined, deterministic tie-break order (severity → widest span → collection order), and correctly gives `email_password_combo` (critical) priority over `email` (medium) in the flagship regression test.
- Splicing right-to-left avoids the classic index-invalidation bug from replacing overlapping regions in a mutable string.
- Test suite (`TestPatternOrderLeak`, 7 cases) is well-targeted: it directly encodes the bug (password leak), the marker-precedence fix, scan/apply parity, an allowlist-interacting case, a custom/composite pattern case, and a genuine partial-overlap union — exactly the edge cases a fix like this needs to prove.
- Docstrings on every new function/class explain the *why* (Issue #832, parity guarantee), not just restate the code.
- Confirmed no dangling references to the removed `_replace_pattern` anywhere in the tree.

## Security Concerns
🟢 LOW — none found in this diff. This *is* a security/privacy fix (previously-leaking credentials now redacted), and I did not find any way the new merge/splice logic could under-redact: every collected span is unioned before splicing, so no fragment of any surviving-post-allowlist match can leak. The PR notes a prior security-specialist pass came back clean with three pre-existing (unrelated) advisories filed separately (#899–#901).

## Problems
None blocking.

## Code Quality
- 🟢 Non-blocking: `_severity_rank` gives unknown/custom patterns (no `PATTERN_METADATA` entry) a rank that always loses to *every* known severity, including `low`. So if a user's custom pattern overlaps a low-severity built-in (e.g. `phone_number`), the built-in's name wins the marker even though the custom pattern was arguably the more actionable finding for that user. Redaction itself is unaffected (the whole union is still replaced), so this is only a labeling/telemetry nit — worth a mention in a follow-up issue if custom-pattern marker attribution matters to users, not a blocker.
- 🟢 Non-blocking: `_select_marker_name`'s tie-break key uses `span.start - span.end` (a negative width) to prefer the widest span via `min()`. It's correct and documented, but a reader has to do the sign flip in their head; a one-line comment or naming it `-width` would remove that friction. Very minor.
- I could not execute `./scripts/check-all.sh` or pytest in this review sandbox (Bash execution was blocked), so I relied on careful manual tracing of the algorithm against the new tests rather than running them; the PR description states all 6028 tests and quality gates pass locally.

## CLAUDE.md Adherence
✅ TDD followed explicitly (tests written RED-first per the PR description, 5 failing at HEAD reproducing the bug plus 2 regression guards).
✅ No shortcuts: no `noqa`/`type: ignore`, no threshold changes, dead code (`_replace_pattern`) fully removed rather than left stubbed.
✅ Type hints throughout (strict-mode-compatible: `NamedTuple`s, `dict[str, re.Pattern[str]]`, etc.).
✅ Docstrings on all new public/private members (Google-style, Args/Returns sections), consistent with the ≥95% interrogate gate.
✅ Scope discipline: patterns.py/scanner.py/CLI untouched as claimed, keeping the blast radius to the one module that owned the bug.

## Testing
Strong. The new `TestPatternOrderLeak` class isolates exactly the failure mode (sequential mutation destroying a later pattern's anchor), adds a scan/apply parity test that scans-then-applies against a real file, covers allowlist interaction with a combo pattern, and — notably — covers a custom/composite pattern and a genuine partial (non-containment) overlap, which are the two hardest cases for a span-merge rewrite to get right. This is not just coverage-for-coverage's-sake; each test would have failed under the old implementation and now passes under the new one per the stated RED/GREEN history.

## Documentation
Good. The module docstring, `redact_content`'s docstring, and every new helper's docstring were updated to describe the new single-pass/merge/splice model and explicitly reference Issue #832, so the "why" (scan/apply parity) is discoverable from the code itself, not just the PR description.



## Verdict: LGTM

The rewrite correctly fixes the P1 leak (verified by hand-tracing the span-collect/merge/splice algorithm against all 7 new tests, including the partial-overlap and custom-pattern cases), keeps the public API and all other modules untouched, and is backed by thorough RED-first tests plus a clean security-specialist pass per the PR description. The one quirk I found (custom/unknown-severity patterns always lose the marker-naming tie-break to built-ins) is cosmetic — the merged span is still fully redacted either way — so it doesn't block merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5054522030 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #903, opened 2026-07-23
purge: case/Unicode-fold wikilink variants ([[secret title#h]]) survive purge and leak the title


Geoffe-Ga:
## Context
Follow-up from #833 (heading-suffixed wikilink RTBF leak, fixed). Security review of that fix found an adjacent, pre-existing leak vector in the same class.

## Problem
`_build_wikilink_pattern` in `creek-tools/creek/purge/engine.py` matches the purged title case-sensitively (no `re.IGNORECASE`, no casefold). Obsidian resolves wikilinks case-insensitively, so after `purge_fragment` of a fragment titled "Secret Title", a live link `[[secret title]]` or `[[secret title#h]]` elsewhere in the vault still resolves and displays the purged title — an RTBF gap for case/Unicode-fold variants.

## Suggested direction
Make the wikilink scrub case-insensitive (and consider Unicode casefolding) in `_build_wikilink_pattern`, with tests mirroring the #833 parametrized suite. Take care not to over-match distinct titles that differ only beyond case folding semantics Obsidian actually uses.

## Lower-priority hardening noted in the same review (not necessarily actionable)
- Markdown-link text `[Secret Title](path)` is never scrubbed (arbitrary prose scrubbing may be by-design out of scope).
- Bare title strings in frontmatter fields (e.g. `aliases:`) are not scrubbed.
- Scrubbing a transclusion `![[Title#h]]` leaves a stray `!` (cosmetic, pre-existing).

Refs #833

PR #904, opened 2026-07-23
fix(purge): scrub heading-suffixed wikilinks on fragment purge

### LINKED ISSUE #833: [scan:bugs] purge_fragment leaves [[Title#heading]] refs intact, leaking purged title (RTBF gap)
## Role
suppressions).

## Goal
Make `purge_fragment` scrub `[[Title#heading]]` and `[[Title#heading|alias]]`
wiki-link references (not just `[[Title]]` / `[[Title|alias]]`), verified by a
test asserting a purged fragment's title no longer appears in a
heading-suffixed link and that `PurgeResult.wikilinks_removed` counts it.

## Context
- File(s): `creek-tools/creek/purge/engine.py:1221-1231` (`_build_wikilink_pattern`).
- Evidence: the wiki-link scrub pattern only tolerates an optional alias, not a
  heading/block-reference suffix:
  ```python
  escaped = re.escape(title)
  return re.compile(rf"\[\[{escaped}(?:\|[^\]]*)?\]\]")
  ```
  For `[[Secret Title#Part]]` the char after the escaped title is `#`, which
  matches neither the optional alias group nor `\]\]`, so nothing matches and the
  reference survives the scrub. The provenance pass only rewrites the bare
  fragment ID, never the title. Reproduced live at the scanned SHA:
  ```
  >>> from creek.purge.engine import _build_wikilink_pattern
  >>> pat = _build_wikilink_pattern('Secret Title')
  >>> bool(pat.search('x [[Secret Title]] y'))        # True  (handled)
  >>> bool(pat.search('x [[Secret Title#Part]] y'))   # False (LEAKED)
  ```
  After a right-to-be-forgotten `purge_fragment`, any other vault file containing
  `[[Secret Title#Part]]` still displays the purged title verbatim, and
  `PurgeResult.wikilinks_removed` under-counts — a privacy / RTBF correctness gap.
- Related: sibling scan-run findings filed this run (redactor password leak,
  broken-links anchor false-positive, Google Drive version-suffix data loss).
  Note the analogous heading-aware fix already present in
  `creek/lint/checks/orphan_compiled.py` (`[^\]|#]+?`).

## Output Format
"Closes #N".

## Examples
```python
# Vault with 00-Creek-Meta/creek_config.yaml marker,
# 01-Fragments/secret.md (id: frag-x, title: "Secret Title"),
# 02-Threads/t.md body = "See [[Secret Title#Part]]".
def test_purge_scrubs_heading_suffixed_wikilink(vault: Path) -> None:
    result = PurgeEngine(vault_path=vault).purge_fragment("frag-x")
    assert "Secret Title" not in (vault / "02-Threads" / "t.md").read_text()
    assert result.wikilinks_removed == 1        # both fail today

# Fix sketch (engine.py): accept an optional heading/block suffix and alias:
#   re.compile(rf"\[\[{escaped}(?:#[^\]|]*)?(?:\|[^\]]*)?\]\]")
```

## Constraints

Geoffe-Ga:
## Summary
- `_build_wikilink_pattern` now accepts an optional heading/block-reference segment, so `[[Title#Heading]]`, `[[Title#^id]]`, `[[Title#Heading|alias]]`, and `[[Title#]]` wikilinks are scrubbed on purge instead of leaking the purged title (RTBF gap), and `PurgeResult.wikilinks_removed` counts them correctly.
- The fix lands at the single pattern chokepoint shared by both purge entry points (`purge_fragment` and `_purge_single`), so every purge path is covered; a regression guard pins that prefix titles (`[[AlphaBeta]]`, `[[AlphaBeta#H]]`) are still untouched when purging "Alpha".
- Security leak-vector review of the purge path came back clean in scope; the pre-existing case-fold variant leak was filed separately as #903 (out of scope here, as are #845 and #835).

## Test plan
- TDD: added `test_fragment_purge_removes_heading_wikilinks` (4 parametrized variants), `test_fragment_purge_removes_all_wikilink_variants` (mixed body, count == 4, `[[Beta]]` survives), and `test_fragment_purge_leaves_prefix_titled_wikilinks` — the leak tests failed RED against the old pattern, then went GREEN with the fix.
- `cd creek-tools && ./scripts/check-all.sh` → exit 0 (12/12 checks; 6034 tests passed, coverage 93.90%, per-file gate green).

Closes #833

### DIFF
--- a/creek-tools/creek/purge/engine.py
+++ b/creek-tools/creek/purge/engine.py
@@ -1219,7 +1219,12 @@ def _build_source_path_matcher(
 
 
 def _build_wikilink_pattern(title: str) -> re.Pattern[str]:
-    """Build a regex matching ``[[title]]`` and ``[[title|alias]]``.
+    """Build a regex matching every wikilink form targeting ``title``.
+
+    Matches ``[[title]]`` and ``[[title|alias]]``, plus heading
+    (``[[title#Heading]]``), block-reference (``[[title#^id]]``), and
+    heading+alias (``[[title#Heading|alias]]``) variants, so a purge
+    removes all links that leak the title.
 
     Args:
         title: The target title to match exactly.
@@ -1228,7 +1233,7 @@ def _build_wikilink_pattern(title: str) -> re.Pattern[str]:
         A compiled regex pattern.
     """
     escaped = re.escape(title)
-    return re.compile(rf"\[\[{escaped}(?:\|[^\]]*)?\]\]")
+    return re.compile(rf"\[\[{escaped}(?:#[^\]|]*)?(?:\|[^\]]*)?\]\]")
 
 
 def _build_provenance_pattern(fragment_id: str) -> re.Pattern[str]:
--- a/creek-tools/tests/test_purge.py
+++ b/creek-tools/tests/test_purge.py
@@ -289,6 +289,73 @@ def test_fragment_purge_removes_wikilinks(tmp_path: Path) -> None:
     assert "[[Beta]]" in post.content
 
 
+@pytest.mark.parametrize(
+    "link",
+    [
+        "[[Alpha#Heading]]",
+        "[[Alpha#Heading|alias]]",
+        "[[Alpha#^blk01]]",
+        "[[Alpha#]]",
+    ],
+)
+def test_fragment_purge_removes_heading_wikilinks(tmp_path: Path, link: str) -> None:
+    """Heading/block-suffixed wiki-links to the purged title are scrubbed (#833)."""
+    vault = _make_vault(tmp_path)
+    _write_fragment(vault, "frag-A", "Alpha")
+    linker = _write_fragment(vault, "frag-B", "Beta", body=f"See {link} today.")
+    engine = PurgeEngine(vault)
+
+    result = engine.purge_fragment("frag-A")
+
+    assert result.wikilinks_removed == 1
+    post = frontmatter.load(str(linker))
+    assert link not in post.content
+    assert "Alpha" not in post.content
+
+
+def test_fragment_purge_removes_all_wikilink_variants(tmp_path: Path) -> None:
+    """A mixed body loses every Alpha link variant while [[Beta]] survives (#833)."""
+    vault = _make_vault(tmp_path)
+    _write_fragment(vault, "frag-A", "Alpha")
+    linker = _write_fragment(
+        vault,
+        "frag-B",
+        "Beta",
+        body=(
+            "See [[Alpha]] and [[Alpha|a]] and [[Alpha#H]] "
+            "and [[Alpha#H|a]] and [[Beta]]."
+        ),
+    )
+    engine = PurgeEngine(vault)
+
+    result = engine.purge_fragment("frag-A")
+
+    assert result.wikilinks_removed == 4
+    post = frontmatter.load(str(linker))
+    assert "Alpha" not in post.content
+    assert "[[Beta]]" in post.content
+
+
+def test_fragment_purge_leaves_prefix_titled_wikilinks(tmp_path: Path) -> None:
+    """Links to a longer title sharing the purged prefix are untouched (#833)."""
+    vault = _make_vault(tmp_path)
+    _write_fragment(vault, "frag-A", "Alpha")
+    linker = _write_fragment(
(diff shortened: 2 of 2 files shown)


Geoffe-Ga:
## Summary
This PR fixes an RTBF (right-to-be-forgotten) leak in `creek-tools/creek/purge/engine.py`: `_build_wikilink_pattern` previously only matched `[[Title]]` and `[[Title|alias]]`, so heading-suffixed and block-reference wikilinks (`[[Title#Heading]]`, `[[Title#^id]]`, `[[Title#Heading|alias]]`, `[[Title#]]`) survived a fragment purge and continued to leak the purged title. The fix extends the shared regex with an optional `#...` segment before the existing optional `|alias` segment.

## Strengths
- The fix lands at the single shared chokepoint (`_build_wikilink_pattern`, consumed by `_scrub_references`), which both `purge_fragment` and `_purge_single` call — so every purge path benefits without duplicated logic.
- `re.escape(title)` is preserved, so titles containing regex metacharacters remain handled safely; no injection risk introduced.
- Counting continues to use `pattern.subn(...)`, so `wikilinks_removed` accurately reflects the number of variants actually removed (verified: 4 distinct variants in one body → count of 4).
- Good negative-case regression test (`test_fragment_purge_leaves_prefix_titled_wikilinks`) proves the fix doesn't over-match longer titles sharing a prefix (e.g., purging "Alpha" leaves `[[AlphaBeta]]` / `[[AlphaBeta#H]]` untouched) — this is exactly the kind of edge case a naive fix could break.
- PR description explicitly scopes out the pre-existing case-fold variant leak (filed as #903) rather than scope-creeping the fix — good discipline.
- Manually traced the regex against all PR-cited variants plus the prefix-title case; behavior matches expectations (heading segment correctly excludes `]` and `|` so it stops at the right boundary, alias segment still works after a heading).

## Security Concerns
None. This is itself a security/privacy fix (closing an RTBF leak vector), and it doesn't introduce new risk — the regex is a bounded, non-nested-quantifier pattern (no ReDoS risk), and title interpolation is escaped via `re.escape`.

## Problems
None found.

## Code Quality
- Minor: the docstring on `_scrub_references` (engine.py ~line 808-810) still says wikilinks are matched as "`[[title]]` / `[[title|alias]]`" only — it wasn't updated to mention the new heading/block-reference variants now handled by `_build_wikilink_pattern`. Non-blocking; purely a stale internal comment, not user-facing documentation.

## CLAUDE.md Adherence
- ✅ TDD followed: tests were added and described as failing RED against the old pattern before the fix went GREEN.
- ✅ `./scripts/check-all.sh` reported as exit 0 with 6034 tests passing and 93.90% coverage (per PR description), above the ≥90% aggregate threshold.
- ✅ Change is scoped and minimal — no unrelated refactoring bundled in.
- ✅ Docstring on the changed function (`_build_wikilink_pattern`) is updated and accurate for the new behavior.
- Could not independently verify live CI status or re-run `check-all.sh` in this review environment; relying on the PR's self-reported check-all output, which is consistent with the code inspected.

## Testing
Strong test coverage for this fix:
- 4 parametrized heading/block-ref/heading+alias/empty-heading variants, each individually asserted to be scrubbed with `wikilinks_removed == 1`.
- A mixed-body test combining all variants plus a same-body `[[Beta]]` control, asserting exact count (4) and that the unrelated link survives.
- A negative/regression test confirming prefix-sharing titles (`AlphaBeta`) are not falsely matched when purging `Alpha`.

This is exactly the kind of edge-case coverage (positive + negative) the fix needs; no gaps identified.

## Documentation
Function docstring for `_build_wikilink_pattern` was properly updated to describe all newly-supported variants. The only gap is the stale summary in `_scrub_references`'s docstring noted above (non-blocking).



## Verdict: LGTM

The fix is small, correctly scoped to the single shared regex chokepoint used by both purge entry points, verified by parametrized positive tests, a mixed-body count test, and a negative (prefix-title) regression guard; no security, correctness, or CLAUDE.md gate issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5054808924 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #913, opened 2026-07-24
fix(mcp): enforce 32-char entropy floor on CREEK_MCP_ELEVATED_TOKEN

### LINKED ISSUE #907: [scan:security] CREEK_MCP_ELEVATED_TOKEN gating vault destruction has no 32-char entropy floor (unlike consumer tokens #838)
## Role
suppressions).

## Goal
Reject a `CREEK_MCP_ELEVATED_TOKEN` shorter than 32 characters at server
startup — failing closed with the same rotation recipe used for consumer
tokens — so the secret that gates irreversible vault destruction is held to at
least the same entropy floor as the lower-value per-consumer bearer tokens.
Verified by a test asserting a too-short elevated token is refused before any
purge tool is reachable, and that a compliant token still authorizes.

## Context
- File(s): `creek-tools/creek_mcp/auth.py:25-40` (`is_elevated`, the elevated
  gate); the missing floor contrasts with
  `creek-tools/creek_mcp/remote_auth.py:90-113` (`_validated_token`,
  `_MIN_TOKEN_LEN = 32`) which enforces the consumer-token floor added by #838.
  The gated destructive surface is `creek-tools/creek_mcp/tools/purge.py`
  (`_gate` at line 156 → `is_elevated` at line 170; the five `creek.purge.*`
  tools, including `creek.purge.vault` which destroys all fragments).
- Scanned at commit: `47fe41dec29ea975db6778f33e81801ea403000b` — re-verify
  against HEAD before starting.
- Evidence: `is_elevated` accepts **any** non-empty token:
  ```python
  # creek_mcp/auth.py:34-40
  expected = os.environ.get(ELEVATED_TOKEN_ENV, "")
  if not expected or not provided_token:
      return False
  return hmac.compare_digest(
      expected.encode("utf-8"),
      provided_token.encode("utf-8"),
  )
  ```
  There is no length/entropy check. Meanwhile the lower-value consumer bearer
  token is explicitly refused below 32 chars:
  ```python
  # remote_auth.py:106-113 (_validated_token, #838)
  if len(token) < _MIN_TOKEN_LEN:  # _MIN_TOKEN_LEN = 32
      msg = (f"consumer {name!r} token is {len(token)} chars, below the "
             f"{_MIN_TOKEN_LEN}-char minimum; rotate it with ...")
      raise ValueError(msg)
  ```
  So the **higher-value** secret (guards `creek.purge.vault`, the single most
  destructive operation in the system) is held to a **weaker** standard than
  the lower-value one. `auth_token` is a caller-supplied tool argument
  (`purge.py` `_gate(..., auth_token=...)`); the compare is constant-time but
  imposes no cost ceiling, so a weak/short operator-chosen elevated token is
  online-guessable by a network-authenticated consumer with no throttle to stop
  it. Bandit + pip-audit are clean at this SHA; this is a code-review finding in
  the scan's "MCP bearer-token transport" priority hunting ground.
- Related: distinct from #895 (consumer-token *rotation*, not the elevated-token
  entropy floor); direct sibling of #838 (which added `_MIN_TOKEN_LEN` for
  consumer tokens). No other open `[scan:security]` issue covers this.

## Output Format
"Closes #N".

## Examples
Hoist a shared minimum-length check and apply it to the elevated token at the
point it is loaded/first used, mirroring `_validated_token`:

```python
# before — auth.py: any non-empty token is accepted
expected = os.environ.get(ELEVATED_TOKEN_ENV, "")
if not expected or not provided_token:
    return False
return hmac.compare_digest(expected.encode("utf-8"),
                           provided_token.encode("utf-8"))

# after — refuse a too-short configured elevated token, fail closed
# (reuse the _MIN_TOKEN_LEN = 32 floor and rotation recipe from remote_auth)
expected = os.environ.get(ELEVATED_TOKEN_ENV, "")
if not expected or not provided_token:
    return False
if len(expected) < _MIN_TOKEN_LEN:
    # configured elevated token is too weak to guard vault destruction;
    # surface at startup with the secrets.token_urlsafe(32) rotation recipe,
    # and deny here so a short token never authorizes a purge.
    return False
return hmac.compare_digest(expected.encode("utf-8"),
                           provided_token.encode("utf-8"))
```

Prefer validating at server startup (so a misconfigured operator gets a clear,
named error with the rotation recipe rather than a silent deny), while keeping
`is_elevated` fail-closed on a short token as defense in depth. Factor the
`_MIN_TOKEN_LEN` floor and the rotation-recipe message into one shared helper
rather than duplicating the constant across `auth.py` and `remote_auth.py`.

## Constraints
- Never echo the token value in any error message — name the length and the
  required minimum only, matching `_validated_token`'s existing message.
- Keep the gate fail-closed: an absent, empty, or too-short elevated token must
  deny every `creek.purge.*` call.
- Scope: this issue only (the elevated-token entropy floor) — file follow-up
  issues for adjacent ideas such as failed-attempt backoff/rate-limiting on the
  purge gate.
- If the finding no longer reproduces at HEAD (e.g. a floor was already added),
  close this issue with a comment explaining what changed instead of forcing a
  PR.

Geoffe-Ga:
## Summary

`CREEK_MCP_ELEVATED_TOKEN` gates the five `creek.purge.*` tools — including `creek.purge.vault`, the single most destructive operation in the system — but `is_elevated` accepted **any** non-empty value. Meanwhile the *lower-value* per-consumer bearer tokens have enforced a 32-char minimum since #838. The higher-value secret was held to the weaker standard, and `auth_token` is a caller-supplied tool argument with no throttle, so a short operator-chosen token was online-guessable by a network-authenticated consumer.

The floor now lives in exactly one place:

- **New `creek_mcp/token_policy.py`** — `MIN_TOKEN_LEN = 32`, the rotation recipe, and the rejection wording, exposed as `meets_min_length` (predicate) and `require_min_length` (raising). Stdlib-only and imports nothing from `creek_mcp`, so both auth surfaces can depend on it without a cycle.
- **`remote_auth._validated_token`** delegates to it; `_MIN_TOKEN_LEN` is deleted. The consumer-token `ValueError` message is **byte-identical** to before, so the unmodified `tests/test_mcp_remote.py` is a regression fence on that.
- **`server.main`** gains `_require_strong_elevated_token`, called *before* the transport branch so **both** stdio and network abort via `parser.error(...)` (exit 2), naming the env var, the observed length, the 32 floor, and the rotation recipe — never the token value. An absent or empty token is still **not** an error: that is the supported "purge disabled" posture.
- **`auth.is_elevated`** denies a sub-floor *configured* token **silently**. It keeps its `-> bool` signature and never raises, because `tools/purge.py::_gate` does not catch and an escaping exception would skip the audit entry every purge call is required to write. Silent also means a hostile caller gets no oracle distinguishing "wrong token" from "server misconfigured". Only the **server's** value is measured; the attacker-supplied one never is. This is the chokepoint covering embedders who call `build_server()` directly and bypass startup.

**`tools/purge.py` is deliberately untouched.** The caller-visible refusal reason stays generic for the same no-oracle reason — telling a possibly-hostile remote caller "the server secret is 11 chars" would narrow a brute-force search against the very secret this hardens. The operator's actionable signal is the startup error on stderr.

### Breaking change

A server running a sub-32-char `CREEK_MCP_ELEVATED_TOKEN` today will **refuse to start** on next launch, on both transports. Rotate with `python -c "import secrets; print(secrets.token_urlsafe(32))"`. There is no grace period, warning-only mode, or opt-out — either would reintroduce exactly the weakness being closed. Called out in `docs/mcp.md`.

## Test plan

Written RED first, then GREEN.

- **New `tests/test_mcp_token_policy.py`** — pins `MIN_TOKEN_LEN == 32` against silent weakening; boundary behavior at 32/31/empty; compliant token returned unchanged; and that the `ValueError` names the subject, the observed length, the floor, and the recipe while **never** echoing the token.
- **`tests/test_mcp_auth.py`** — `test_is_elevated_denies_sub_minimum_configured_token` is the crux: a 31-char server secret with an **exact match** supplied must still return `False` (and returning a bool proves it does not raise). `test_is_elevated_accepts_exact_boundary_token` fences the off-by-one at exactly 32. The AST walk banning `==`/`!=` inside `is_elevated` is unmodified and still passes.
- **`tests/test_mcp_purge.py`** — `test_purge_refused_when_configured_token_is_sub_minimum` calls `purge_vault_tool` with a correct `confirm_vault_path` and the exact (weak) token, then asserts `refused`, the fragment survives on disk, a refusal entry landed in the audit log, and the refusal reason contains neither `32`, `chars`, nor the token value.
- **`tests/test_mcp_server.py`** — startup rejection on stdio *and* on `--transport network` (the latter with valid consumer tokens set, proving the elevated check fires *first*); `build_server` is stubbed to raise if invoked, proving refusal happens before any purge tool is reachable. Three regression-fence cases assert compliant / unset / empty all still start normally.
- Repaired short test literals that would otherwise have passed for the wrong reason — notably `test_crawdad_consumer_cannot_purge`, which with an 11-char server secret would have been refused because the *server* was weak rather than because CrawDad has no token.

`cd creek-tools && ./scripts/check-all.sh` → exit 0: 12/12 checks, 6068 tests passed, aggregate branch coverage 93.90%, per-file gate green, `creek_mcp/token_policy.py` and `creek_mcp/auth.py` both at 100%.

Closes #907

### DIFF
--- a/creek-tools/creek_mcp/auth.py
+++ b/creek-tools/creek_mcp/auth.py
@@ -11,12 +11,24 @@ relative to the input length, so an attacker cannot probe the secret
 through repeated calls. The :mod:`creek_mcp.auth` test module asserts
 via an AST walk that ``is_elevated`` contains no ``==``/``!=`` on the
 hot path; the rule is mechanical, not stylistic.
+
+The configured token must also clear the shared 32-character floor in
+:data:`creek_mcp.token_policy.MIN_TOKEN_LEN` (#907) — a guessable secret
+must not guard irreversible destruction. :func:`creek_mcp.server.main`
+turns a weak token into a loud startup error, but the check is repeated
+here because it is the only chokepoint that covers embedders calling
+:func:`creek_mcp.server.build_server` directly, which bypasses startup
+entirely. Here the deny is **silent**: ``is_elevated`` returns ``False``
+rather than raising or explaining, so a possibly-hostile caller is handed
+no oracle about the server's configuration.
 """
 
 from __future__ import annotations
 
 import hmac
 import os
+
+from creek_mcp.token_policy import meets_min_length
 
 ELEVATED_TOKEN_ENV = "CREEK_MCP_ELEVATED_TOKEN"
 """Env var the server reads at startup to learn the expected token."""
@@ -25,14 +37,37 @@ ELEVATED_TOKEN_ENV = "CREEK_MCP_ELEVATED_TOKEN"
 def is_elevated(provided_token: str | None) -> bool:
     """Return ``True`` when *provided_token* matches the configured secret.
 
-    The gate fails closed when either side is missing: an absent or
-    empty ``CREEK_MCP_ELEVATED_TOKEN`` denies every call, and a missing
-    client token cannot accidentally match an empty server token. Only
-    a non-empty server token combined with a constant-time match
-    against the client value returns ``True``.
+    The gate fails closed three ways. An absent or empty
+    ``CREEK_MCP_ELEVATED_TOKEN`` denies every call, and a missing client
+    token cannot accidentally match an empty server token. A *configured*
+    token below :data:`creek_mcp.token_policy.MIN_TOKEN_LEN` characters
+    also denies every call (#907): a secret weak enough to guess is not
+    allowed to authorize vault destruction, even on an exact match. Only
+    a strong server token combined with a constant-time match against the
+    client value returns ``True``.
+
+    Only the *server's* token is measured. The client-supplied value is
+    never length-checked: it is attacker-controlled, so measuring it
+    proves nothing and only risks another signal leaking back out.
+
+    The weak-configuration deny is silent — this function returns a
+    ``bool`` and never raises, both to avoid disclosing server
+    configuration to the caller and because
+    :func:`creek_mcp.tools.purge._gate` does not catch: an exception here
+    would escape into the FastMCP tool surface and skip the audit entry
+    every purge call is required to write.
+
+    Args:
+        provided_token: The ``auth_token`` the caller presented, if any.
+
+    Returns:
+        ``True`` only when a floor-clearing server token is configured
+        and the caller's token matches it exactly.
     """
     expected = os.environ.get(ELEVATED_TOKEN_ENV, "")
     if not expected or not provided_token:
+        return False
+    if not meets_min_length(expected):
         return False
     return hmac.compare_digest(
         expected.encode("utf-8"),
--- a/creek-tools/creek_mcp/remote_auth.py
+++ b/creek-tools/creek_mcp/remote_auth.py
@@ -38,6 +38,8 @@ from typing import TYPE_CHECKING, Final
 from mcp.server.auth.provider import AccessToken, TokenVerifier
 from mcp.server.auth.settings import AuthSettings
 
+from creek_mcp.token_policy import require_min_length
+
 if TYPE_CHECKING:
     from collections.abc import Mapping
 
@@ -52,9 +54,6 @@ TOKEN_TTL_ENV: Final[str] = "CREEK_MCP_TOKEN_TTL_SECONDS"
 
 _REMOTE_TOKEN_TTL_SECONDS: Final[int] = 3600
 """Default lifetime of a verified bearer's ``AccessToken`` (one hour)."""
-
-_MIN_TOKEN_LEN: Final[int] = 32
-"""Minimum consumer-token length — the ``secrets.token_urlsafe(32)`` floor (#838)."""
 
 # Monkeypatchable clock alias: tests pin `_now` to a fixed instant so the
 # expires_at arithmetic is exact; production keeps wall-clock time.
@@ -88,10 +87,12 @@ def _token_ttl_seconds() -> int:
 
 
 def _validated_token(name: str, token: str) -> str:
-    """Return *token* if it clears :data:`_MIN_TOKEN_LEN`, else raise (#838).
+    """Return *token* if it clears the shared length floor, else raise (#838).
 
-    The error names the consumer and the observed/required lengths and gives
-    the rotation recipe — it never echoes the token value itself.
+    Delegates to :func:`creek_mcp.token_policy.require_min_length`, which the
+    elevated-token gate shares (#907), so both surfaces enforce one number
+    with one wording. The error names the consumer and the observed/required
+    lengths and gives the rotation recipe — it never echoes the token itself.
 
     Args:
         name: The consumer the token belongs to (already stripped).
@@ -101,16 +102,10 @@ def _validated_token(name: str, token: str) -> str:
         The token, unchanged, when it meets the minimum length.
 
     Raises:
-        ValueError: If the token is shorter than :data:`_MIN_TOKEN_LEN`.
+        ValueError: If the token is shorter than
+            :data:`creek_mcp.token_policy.MIN_TOKEN_LEN`.
     """
-    if len(token) < _MIN_TOKEN_LEN:
-        msg = (
-            f"consumer {name!r} token is {len(token)} chars, below the "
-            f"{_MIN_TOKEN_LEN}-char minimum; rotate it with "
-            'python -c "import secrets; print(secrets.token_urlsafe(32))"'
-        )
-        raise ValueError(msg)
-    return token
+    return require_min_length(f"consumer {name!r} token", token)
 
 
 def load_consumer_tokens(
@@ -121,7 +116,8 @@ def load_consumer_tokens(
     Format: ``adepthood=<token>;other=<token>``. Blank entries and entries
     without a token are skipped. Returns an empty dict when unset — the caller
     treats "no tokens configured" as "network mode not permitted" (no anonymous
-    access). A *present* token shorter than :data:`_MIN_TOKEN_LEN` characters
+    access). A *present* token shorter than
+    :data:`creek_mcp.token_policy.MIN_TOKEN_LEN` characters
     is refused outright (#838); generate compliant tokens with
     ``python -c "import secrets; print(secrets.token_urlsafe(32))"``.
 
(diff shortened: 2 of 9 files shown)


Geoffe-Ga:
## Summary
PR #913 closes issue #907: `CREEK_MCP_ELEVATED_TOKEN` — which gates the most destructive MCP tool, `creek.purge.vault` — previously accepted any non-empty value, while the lower-value per-consumer bearer tokens already enforced a 32-char minimum (#838). The PR extracts the shared floor into a new stdlib-only `creek_mcp/token_policy.py`, enforces it at startup for both transports (`server.main`), and enforces it defensively (and silently) inside `auth.is_elevated` for embedders that call `build_server()` directly and bypass `main`.

## Strengths
- **Single source of truth.** `token_policy.py` centralizes `MIN_TOKEN_LEN`, the rejection wording, and the rotation recipe; both `remote_auth._validated_token` and the new elevated-token checks delegate to it instead of duplicating a magic number.
- **Correct silent-vs-loud split.** Startup (`server.main`) fails loudly via `parser.error` (exit 2, message naming the env var/length/floor/recipe, never the token). The runtime gate (`auth.is_elevated`) fails **silently** (returns `False`, no exception) — correctly reasoned: `tools/purge.py::_gate` doesn't catch exceptions, so a raise there would skip the mandatory audit-log entry, and a bool return avoids handing a hostile caller a configuration oracle. I traced `_gate` and confirmed the refusal reason (`_REFUSAL_NO_TOKEN`) is generic regardless of *why* `is_elevated` returned `False`.
- **No new oracle.** Only the server-configured token's length is measured; the attacker-supplied token is never length-checked (correctly, since checking it would prove nothing and could leak a timing signal). The server's expected token isn't attacker-influenced per-call, so there's no exploitable timing channel from the early-return-on-weak-token branch.
- **Regression fencing.** `tests/test_mcp_remote.py` is explicitly left unmodified and its exact error-message assertions (`assert "secrets.token_urlsafe(32)" in message`, etc.) still pass against `require_min_length`, proving the consumer-token error text is byte-identical post-refactor.
- **Honest, well-flagged breaking change.** Both the PR body and `docs/mcp.md` clearly document that a sub-32-char token now refuses to boot, with the exact rotation command, and correctly note there is no grace period/opt-out (appropriately, since a grace period would reintroduce the exact weakness being closed).
- Docstrings are thorough (Google-style, Args/Returns/Raises), matching creek-tools' ≥95% interrogate requirement, and explain *why* (silent vs. loud, why only the server token is measured) rather than restating the code.

## Security Concerns
None found. 🟢 The change is itself a security hardening fix; I did not find a way to bypass either the startup check or the runtime gate, and the "which chokepoint covers embedders" reasoning in the docstrings checks out against the actual `build_server`/`main` call graph.

## Problems
None blocking.

## Code Quality
- 🟢 **Non-blocking:** `docs/mcp.md`'s consumer-token section (`CREEK_MCP_CONSUMER_TOKENS`, around the "Generate each token as high-entropy hex" bullet and the `export CREEK_MCP_CONSUMER_TOKENS=...` example) still recommends `secrets.token_hex(32)`, while this PR's own `token_policy._ROTATION_RECIPE` and the elevated-token doc examples now standardize on `secrets.token_urlsafe(32)`. Not a security issue (`token_hex(32)` produces 64 chars, well above the 32-char floor either way) and it predates this PR, but since the PR's stated goal is "one number, one wording" shared across both auth surfaces, it would have been a natural place to also unify the example command.

## CLAUDE.md Adherence
✅ TDD followed (PR description states tests written RED first, then GREEN implementation).
✅ Coverage: PR reports `token_policy.py` and `auth.py` both at 100%, aggregate 93.90% (above the 90% floor).
✅ Docstrings: Google-style with Args/Returns/Raises throughout the new/changed public functions.
✅ Type hints: `Final`, `str | None`, etc., consistent with strict mypy.
❓ CI status could not be independently confirmed in this sandbox (`gh pr checks` required interactive approval I couldn't obtain), but the PR body reports `./scripts/check-all.sh` exit 0 with 12/12 checks and 6068 tests passing.

## Testing
Strong. New/changed tests cover:
- The policy module itself (boundary at 32/31/empty, unchanged-token round-trip, message contents) in `test_mcp_token_policy.py`.
- `is_elevated`'s weak-token silent deny and exact-boundary accept in `test_mcp_auth.py`.
- End-to-end purge refusal on a weak *server* secret with an exact-match caller token, verifying the vault survives, the refusal is audited, and the reason leaks neither the floor number nor the token (`test_mcp_purge.py`).
- Startup rejection on both stdio and network transport, proving the elevated check runs *before* `build_server`/`_serve_network` are ever invoked, plus regression fences for compliant/unset/empty tokens (`test_mcp_server.py`).
- A repaired pre-existing test (`test_crawdad_consumer_cannot_purge`) that would previously have passed for the wrong reason (weak server config rather than CrawDad's missing token) — good catch, this is exactly the kind of masked-assertion bug mutation testing is meant to surface.

## Documentation
`docs/mcp.md` is updated with the new floor, the breaking-change callout, and the rotation recipe in the right places. Only gap is the leftover `token_hex(32)` example noted above under Code Quality.



## Verdict: LGTM

The fix is narrowly scoped, closes a real vulnerability (a guessable CREEK_MCP_ELEVATED_TOKEN could authorize vault destruction), and is backed by thorough boundary/regression tests plus a defense-in-depth design (loud startup failure on both transports, silent runtime deny for embedders bypassing main). I verified the logic directly (auth.py, server.py, tools/purge.py, token_policy.py) and found no bugs, oracles, or missed call sites; the only issue is a pre-existing, non-blocking doc inconsistency (token_hex vs token_urlsafe example in the consumer-token section) that this PR didn't need to fix.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5073800626 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #921, opened 2026-07-24
fix(mcp): enforce the read-side tier ceiling on creek.reflect entry_ref

### LINKED ISSUE #846: [scan:bugs] creek.reflect entry_ref returns verbatim INTIMATE content to a lower-ceiling caller
## Role
suppressions).

## Goal
Enforce the read-side tier ceiling on `creek.reflect`'s `entry_ref` path so a
caller whose `privacy_tier_ceiling` is below the referenced fragment's tier gets
a structured refusal instead of verbatim above-ceiling content — verified by a
test that an OPEN-ceiling call against an INTIMATE fragment refuses.

## Context
- File(s): `creek-tools/creek_mcp/tools/reflect.py:232-254` (`_resolve_entry`)
  and `creek-tools/creek_mcp/tools/reflect.py:302-352` (`reflect_tool`).
- Evidence: `reflect_tool` resolves `entry_ref` to a fragment body and its
  classified tier via `_resolve_entry` (reflect.py:302), then uses that tier
  **only** to pick the LLM routing tier — `_routing_tier(privacy_tier_ceiling,
  entry_tier)` (reflect.py:320) forces local processing for INTIMATE. It
  **never checks `entry_tier` against `privacy_tier_ceiling`**. The returned
  `notes[].quote` values are verified verbatim substrings of the entry
  (`_clean_notes`, reflect.py:339), and the optional `essay` is model prose over
  the entry — both are returned in the result (reflect.py:340-352). So a caller
  with `privacy_tier_ceiling=TierCeiling.OPEN` who passes `entry_ref` of an
  INTIMATE fragment receives literal spans of that intimate content. The tool's
  own docstring scopes the ceiling to "gates corpus admission and, via
  `_routing_tier`, the local-vs-cloud routing tier" (reflect.py:283-284) — i.e.
  the ceiling gates *grounding corpus* and *LLM routing* but not the primary
  entry itself. The sibling read tool `creek.author` instead threads the ceiling
  to specialists that **exclude** above-ceiling fragments; reflect has no such
  exclusion/refusal. Remote callers reach these tools under an admitted ceiling
  (`server.py` `_REMOTE_ADMITTED_CEILINGS`), so this is a reachable egress path,
  not local-only.
  body survives purge, reflect unguarded `frontmatter.load` crash, compile
  write-side tier-ceiling gap. Confirm intended read-side semantics against
  `creek_mcp/tier_ceiling.py` and `creek/author` before implementing.

## Output Format
"Closes #N".

## Examples

```python
def test_reflect_refuses_entry_ref_above_ceiling(tmp_path):
    vault = _init_vault(tmp_path)
    _write_fragment(vault, id="frag-x", tier="intimate",
                    body="the distinctive intimate body")

    def echo_factory(_tier):
        # model that surfaces a note quoting the entry verbatim
        return lambda _prompt: '{"notes": [{"quote": "the distinctive '\
                               'intimate body", "note": "n"}]}'

    result = reflect_tool(
        vault_path=vault,
        llm_factory=echo_factory,
        entry_ref="frag-x",
    )

    assert result["status"] == "refused"
```

Today `status == "ok"` and `result["notes"][0]["quote"]` is a verbatim span of
the intimate body. After the fix — refuse when the resolved `entry_tier` is not
`tier_allowed`/`write_tier_allowed` under `privacy_tier_ceiling`, mirroring
`creek.author`'s exclusion — the call returns a structured refusal.

Fix sketch (in `reflect_tool`, right after `_resolve_entry`):

```python
entry, entry_tier = _resolve_entry(content, entry_ref, vault_path)
if entry_ref and entry_tier is not None and not tier_allowed(
    entry_tier, privacy_tier_ceiling
):
    return refusal_response(
        tool=TOOL_NAME, ceiling=privacy_tier_ceiling,
        reason="entry_ref tier exceeds ceiling",
    )
```

## Constraints
- Only the `entry_ref` path needs gating — raw `content` carries no
  classification (`entry_tier is None`) and is the caller's own text.

Geoffe-Ga:
## Summary

`creek.reflect` resolved an `entry_ref` to a vault fragment's body **and** its classified `privacy_tier`, then used that tier for exactly one thing: picking the LLM routing tier. It never compared the tier to the caller's `privacy_tier_ceiling`. Since every returned `notes[].quote` is a *verified verbatim substring* of the entry, a caller at `ceiling=open` who passed the id of an INTIMATE fragment received literal spans of that intimate content — plus a free-prose `essay` written over it.

This was reachable over the network. `server.py` caps remote consumers at `_REMOTE_ADMITTED_CEILINGS = PERSONAL` precisely so intimate content is unreachable remotely (#759); reflect's `entry_ref` path walked around that cap. The contract doc has promised the opposite all along — *"Reads above the ceiling are refused… a rank comparison applied uniformly to reads and writes"* (`docs/decisions/2026-06-30-adepthood-creek-mcp-contract.md`) — so this is a spec-conformance bug, not a semantics change.

### The fix

A single read-side admission gate. New module-level `_above_ceiling` delegates the rank comparison to `creek_mcp.tier_ceiling.tier_allowed` — the canonical read-side predicate, so reflect cannot drift from the rest of the MCP surface — and `reflect_tool` returns the standard `refusal_response` when it fires. Raw inline `content` is never gated: it resolves to `entry_tier is None`, carries no classification, and is the caller's own text.

Three placement decisions, each load-bearing:

| Placement | Why |
|---|---|
| **After** the unconditional `MCPAuditLog` append | A refused probe stays audited. No second append on refusal, and the fragment id is deliberately not logged — so probing surfaces as a *rate* of `has_entry_ref` calls per consumer, not as named targets. Both choices are now recorded in a comment rather than being incidental. |
| **After** the `"entry_ref not found"` refusal | Keeps that distinct, pinned, debuggable reason intact. |
| **Before** the `care_guard` seam | An `escalate` response is a one-bit oracle telling a caller who may not read a fragment that it carries acute-distress markers. Care still runs for every raw-`content` call and every within-ceiling `entry_ref`, so no legitimate care pathway is lost — a client that *wrote* an entry necessarily holds a ceiling that admits it. |

The refusal reason does **not** echo the offending tier, diverging from `save.py` on purpose: there the tier came from the caller's own input; here it is derived from content they are not admitted to, so echoing it would be a tier-classification oracle. The residual coarse oracle in the not-found-vs-above-ceiling split is documented as accepted risk, with the note that unifying the two reasons would also require equalising the vault scan or the timing difference would preserve it as a side channel.

`_fragment_tier` and `_routing_tier` are unchanged. The routing fold-in is now unreachable-by-construction through the public API (every admitted `entry_tier` is at or below the ceiling-derived tier), so it is pinned by a **direct unit test** rather than deleted — it remains defense in depth if the gate is ever reordered or bypassed.

### Two existing tests changed — neither to make a metric pass

- `test_entry_ref_resolves_a_fragment_body` — its fixture had no `privacy_tier` key and would now fail closed, so it declares `privacy_tier: open`. Intent untouched.
- `test_entry_ref_intimate_fragment_routes_local_despite_open_ceiling` — asserted a scenario that is now refused *before* the LLM factory is ever built, making `factory.asked_tier is PrivacyTier.INTIMATE` unreachable-by-construction. Replaced by the same guarantee at an admitting ceiling. Refusal is strictly stronger than routing local.

## Test plan

`cd creek-tools && ./scripts/check-all.sh` → **exit 0** (12 passed / 0 failed; 6083 tests; 93.91% aggregate branch coverage; per-file gate, mypy strict, pylint, bandit, refurb, tryceratops, interrogate, complexity all clean).

Gate 1 was genuinely RED first — 4 failures against unfixed `reflect.py`, the headline one asserting the intimate span was absent from `json.dumps(result)` so the failure names the leak itself rather than a status mismatch.

New coverage:

- **The leak** — INTIMATE fragment at `ceiling=open` is refused; no span of the body appears anywhere in the serialized result; the LLM factory is never asked.
- **Rank, not a special case** — PERSONAL refused under OPEN; PERSONAL admitted under PERSONAL; INTIMATE admitted under INTIMATE (and still routes local) and under ALL.
- **Fail-closed** — a fragment with no `privacy_tier` key, and one with an unparseable value, are both refused; each asserted at `ceiling=personal` as well as `open`, so weakening `_fragment_tier`'s default from INTIMATE to PERSONAL would fail.
- **Non-bypass** — supplying `content` *and* `entry_ref` together: the fragment is never read, the span appears in neither the result nor the recorded prompt, and routing follows the ceiling.
- **Ordering** — `care_guard` is not consulted for an above-ceiling entry, yet still fires for an admitted one; and exactly one audit record is written for a refused call (a refusal-path append would make refused reads distinguishable by record count).
- **Defense in depth** — `_routing_tier` parametrized directly across 10 cases, including the ties and the `None` entry tier, so collapsing it to `return ceiling_tier` fails.

Gate 2.5 self-review (implementation + test + security dimensions, plus a dedicated prior security pass) found no residual egress path and returned one blocker — the missing direct `_routing_tier` test — which is fixed here along with every non-blocking nit.

## Out of scope, already tracked

- #847 — unguarded `frontmatter.load` in `_resolve_entry`.
- #848 — `creek.compile`'s unenforced write-side tier ceiling.

Two further findings from the security pass have no issue yet and are worth filing: `creek/vault/writer.py`'s `_clear_classification` keeps `privacy_tier` across a material body rewrite (a stale-tier admission path straight through this new gate), and `tier_allowed`'s bare `_TIER_RANK[tier]` subscript would `KeyError` rather than fail closed if a `PrivacyTier` member were ever added without updating the map.

Closes #846

### DIFF
--- a/creek-tools/creek_mcp/tools/reflect.py
+++ b/creek-tools/creek_mcp/tools/reflect.py
@@ -7,8 +7,14 @@ grounded in retrieval over their corpus. Distinct from the essay-shaped
 ``draft``/``author``/``mine`` surface; this is the reflection surface Adepthood's
 journal needs.
 
-Three guarantees this module enforces:
+Four guarantees this module enforces:
 
+- **The ceiling is enforced on read (#846).** An ``entry_ref`` resolves a vault
+  fragment the caller did not supply, so its classified tier is checked against
+  the caller's ``privacy_tier_ceiling`` before anything is done with its text: an
+  above-ceiling fragment is refused outright rather than reflected. Raw inline
+  ``content`` is the caller's own text and carries no classification, so it is
+  never gated here.
 - **INTIMATE never egresses.** The LLM callable is obtained from a *tier-keyed
   factory* (``llm_factory(tier)``). The production factory resolves through
   :class:`creek.classify.llm.router.ModelRouter`, whose
@@ -18,11 +24,12 @@ Three guarantees this module enforces:
   only derives the routing tier and hands it over. The routing tier is the
   **more sensitive** of the entry's own classified ``privacy_tier`` (when it came
   from a vault fragment via ``entry_ref``) and the ceiling-derived tier — so an
-  INTIMATE fragment routes local even under a low ceiling. Raw inline ``content``
-  carries no classification, so there the guarantee is bounded by the caller's
-  declared ceiling (a caller cannot mark intimate content ``open`` and also have
-  it treated as intimate without classifying it first). Both fail closed to
-  INTIMATE on anything unrecognised.
+  admitted INTIMATE fragment still routes local under a broader ceiling, which is
+  defense in depth behind the read-side gate above. Raw inline ``content`` carries
+  no classification, so there the guarantee is bounded by the caller's declared
+  ceiling (a caller cannot mark intimate content ``open`` and also have it treated
+  as intimate without classifying it first). Both fail closed to INTIMATE on
+  anything unrecognised.
 - **Quotes are verbatim.** Every returned ``quote`` is validated to be a
   substring of the entry (whitespace-normalised). Model-supplied spans that are
   not are dropped — the client re-anchors verbatim quotes to character offsets
@@ -44,7 +51,12 @@ from typing import TYPE_CHECKING, Any, Protocol
 from creek.care.guardrail import CARE_POLICY, CARE_SIGNAL
 from creek.models import PrivacyTier
 from creek_mcp.audit import MCPAuditLog
-from creek_mcp.tier_ceiling import TierCeiling, refusal_response, to_privacy_override
+from creek_mcp.tier_ceiling import (
+    TierCeiling,
+    refusal_response,
+    tier_allowed,
+    to_privacy_override,
+)
 
 if TYPE_CHECKING:
     from collections.abc import Callable
@@ -110,11 +122,37 @@ def _routing_tier(ceiling: TierCeiling, entry_tier: PrivacyTier | None) -> Priva
 
     Failing closed: an unrecognised ceiling, or an *entry_tier* outside the known
     ranks, routes as INTIMATE (local-only).
+
+    An above-ceiling *entry_tier* is now refused upstream by
+    :func:`_above_ceiling` before this function ever runs (#846), so in practice
+    this only reconciles an *admitted* ``entry_ref`` (or raw ``content``, which
+    has no *entry_tier*) against the ceiling. It stays in place as defense in
+    depth — the load-bearing INTIMATE-never-egresses guarantee — and must not be
+    removed.
     """
     ceiling_tier = _CEILING_ROUTING_TIER.get(ceiling, PrivacyTier.INTIMATE)
     if entry_tier is None:
         return ceiling_tier
     return max(ceiling_tier, entry_tier, key=_sensitivity)
+
+
+def _above_ceiling(entry_tier: PrivacyTier | None, ceiling: TierCeiling) -> bool:
+    """Return whether a resolved entry's classified tier exceeds *ceiling*.
+
+    Delegates the rank comparison to :func:`creek_mcp.tier_ceiling.tier_allowed`,
+    the canonical read-side admission predicate, so this tool cannot drift from
+    the rest of the MCP surface.
+
+    Args:
+        entry_tier: The entry's classified ``privacy_tier``, or ``None`` for raw
+            inline ``content`` — which carries no classification and is the
+            caller's own text, so it is never above the ceiling.
+        ceiling: The caller's declared ceiling.
+
+    Returns:
+        ``True`` when the entry must be refused rather than reflected.
+    """
+    return entry_tier is not None and not tier_allowed(entry_tier, ceiling)
 
 
 def _normalise(text: str) -> str:
@@ -237,9 +275,21 @@ def _resolve_entry(
     Raw *content* carries no classification, so its tier is ``None`` and the
     caller falls back to the ceiling. An *entry_ref* is resolved to a fragment
     markdown file under ``01-Fragments``; its body becomes the entry **and its
-    persisted ``privacy_tier`` becomes the routing tier** — this is what stops an
-    INTIMATE fragment being reflected through a cloud model under a low ceiling.
-    A blank result yields ``("", None)``, which the caller turns into a refusal.
+    persisted ``privacy_tier`` becomes both the admission tier and the routing
+    tier** — the caller checks the admission tier against the ceiling via
+    :func:`_above_ceiling` (#846) before doing anything else with the text, and
+    separately folds it into :func:`_routing_tier` so an admitted INTIMATE
+    fragment still routes local. A fragment with the ``privacy_tier`` key
+    missing entirely fails closed to INTIMATE (:func:`_fragment_tier`) — but
+    that only fires for hand-edited or legacy files. A normally
+    pipeline-written, not-yet-classified fragment carries an *explicit*
+    ``privacy_tier: unclassified`` (the ``Fragment`` model default,
+    serialised by every write), which ranks as open and is admitted at any
+    ceiling; that is a separate, deliberate policy owned by
+    ``creek_mcp.tier_ceiling._TIER_RANK`` and
+    :func:`creek.classify.privacy_filter.tier_of`, not by this fail-closed
+    path. A blank result yields ``("", None)``, which the caller turns into
+    a refusal.
     """
     if content and content.strip():
         return content, None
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
PR #921 fixes a real privacy leak in `creek.reflect`'s `entry_ref` path (#846): a caller could pass the id of an INTIMATE/PERSONAL vault fragment while declaring a lower `privacy_tier_ceiling`, and receive verbatim spans of that fragment (plus free-form model prose) despite the ceiling. The fix adds a read-side admission gate (`_above_ceiling`) that refuses the call outright before the fragment text touches the care guard or the LLM, delegating the rank comparison to the existing canonical `tier_ceiling.tier_allowed` predicate. Only `creek-tools/creek_mcp/tools/reflect.py` and its test file change.

## Strengths
- Uses the *existing* canonical predicate (`tier_allowed`) instead of re-deriving tier ranking locally — avoids the exact kind of drift that caused the original bug (a second, unsynchronized notion of "admissible").
- Gate ordering is deliberate and tested: above the care-guard seam (so an `escalate` response can't act as a one-bit oracle revealing that an unadmitted fragment contains acute-distress markers), and reuses the single unconditional audit-log append rather than adding a second one (avoids making refused vs. admitted calls distinguishable by record count).
- Explicit non-bypass test (`test_content_wins_over_entry_ref_and_is_not_a_ceiling_bypass`) proves supplying both `content` and `entry_ref` can't be used to smuggle an above-ceiling fragment into the prompt — asserts on the actual prompt text sent to the LLM, not just the response shape.
- Fail-closed behavior is pinned from three distinct angles: missing `privacy_tier` key, unparseable/unknown tier value, and explicit `unclassified` (which is deliberately *not* fail-closed, since it ranks as open by design) — these are easy to conflate and the tests keep them separate.
- `_routing_tier`'s defense-in-depth fold-in (INTIMATE-never-egresses) is kept and directly unit-tested even though the new gate makes it unreachable-by-construction from the public API — correctly reasoned as insurance against the gate being bypassed or reordered later.
- PR description transparently documents two related, out-of-scope findings from the security self-review (a `_clear_classification` stale-tier path, and `tier_allowed`'s bare dict subscript risk) rather than silently leaving them undiscovered.

## Security Concerns
🟢 LOW — Two residual risks are called out explicitly in code comments/PR body rather than left implicit:
- The distinct `"entry_ref not found"` vs `"entry_ref tier exceeds ceiling"` refusal reasons form a coarse existence/rank oracle across repeated probes. Mitigated by unguessable fragment ids (`frag-` + 12 hex) and the fact that a probing consumer can only learn a coarse rank of a fragment they already know the id of; accepted risk is documented in-code with a note on what would need to change (equalizing scan cost) if the two reasons are ever merged.
- `tier_allowed`'s bare `_TIER_RANK[tier]` subscript would raise `KeyError` rather than fail closed if a new `PrivacyTier` member were ever added without updating the map. Not introduced by this PR (pre-existing helper), and explicitly flagged as unresolved in the PR body rather than silently ignored. Worth a follow-up issue but not blocking this fix.

Neither rises to blocking: both are pre-existing/inherent tradeoffs of the design that the PR discloses rather than introduces or hides.

## Problems
None found. The gate placement, fail-closed defaults, and precedence of `content` over `entry_ref` all check out against the source in `reflect.py` and `tier_ceiling.py`.

## Code Quality
- `reflect.py` maintains its own `_TIER_SENSITIVITY`/`_sensitivity` ranking (used by `_routing_tier`) alongside `tier_ceiling.py`'s `_TIER_RANK`/`_CEILING_RANK`. This duplication pre-dates this PR and the new `_above_ceiling` correctly avoids adding to it by delegating to `tier_allowed` — but the two parallel ranking tables remain a latent drift risk or that some future change updates one and not the other. Not blocking; worth a follow-up to unify at some point.
- Comment density in `reflect_tool` (especially around the audit log and refusal) is high but each one earns its place by documenting a non-obvious invariant (append-count-as-oracle, timing side-channel) rather than restating the code.

## CLAUDE.md Adherence
✅ TDD — PR body states Gate 1 was genuinely red first (4 failing tests against unfixed code), consistent with the repo's Stay-Green workflow.
✅ Local checks — PR body reports `./scripts/check-all.sh` exit 0 (6083 tests, 93.91% aggregate branch coverage, mypy strict, pylint, bandit, refurb, tryceratops, interrogate, complexity all clean).
🟡 CI — at review time, the `Code Quality & Testing` jobs (3.11/3.12/3.13) and `claude-review` were still `IN_PROGRESS`; `CrawDad Quality` jobs had already passed. Should confirm all jobs land green before merge per the repo's Gate 3.
✅ Commit/PR hygiene — feature branch off `main`, descriptive PR body with rationale, out-of-scope items explicitly tracked (#847, #848), closes #846.

## Testing
Very strong. New coverage includes: the leak scenario itself (INTIMATE fragment under OPEN ceiling, asserting the exact literal span is absent from `json.dumps(result)`); rank-based refusal (PERSONAL under OPEN) proving it's a general comparison, not an INTIMATE-only special case; admission at equal/broader ceilings; fail-closed for missing and unparseable tier values at two different ceilings each; the `content`-wins non-bypass case verified against the actual LLM prompt; gate-vs-care-guard ordering (guard never consulted when refused, still consulted when admitted); and audit-log-once-per-call semantics. The `_routing_tier` fold-in is also pinned directly via parametrized tests since the public API can no longer reach its above-ceiling branch. No gaps identified.

## Documentation
Excellent — module docstring updated to state four guarantees (was three), each new/changed function (`_above_ceiling`, `_routing_tier`, `_resolve_entry`, `reflect_tool`) has a docstring explaining not just what it does but why, including the accepted-risk reasoning for the refusal-reason side channel. Test docstrings likewise explain the specific regression each test would catch, which is unusually thorough and will pay off for future maintainers.



## Verdict: LGTM

This closes a real, network-reachable privacy leak (intimate/personal fragment content egressing via `entry_ref` to a caller below the required ceiling) with a minimal, correctly-ordered gate that delegates to the canonical `tier_allowed` predicate rather than reimplementing rank comparison, and is backed by exhaustive tests covering the leak itself, ordering vs. the care seam, fail-closed defaults, and a documented non-bypass via combined `content`+`entry_ref`. Nothing here blocks merge; the PR body itself transparently scopes out two related follow-ups (#847, #848) rather than silently leaving gaps.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5074528092 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #926, opened 2026-07-24
[scan:bugs] unparseable fragments under 01-Fragments are silently skipped everywhere and never reported to the operator


Geoffe-Ga:
## Role
You are a senior Python engineer working in this project's codebase, following its existing conventions (TDD via stay-green, the `cd creek-tools && ./scripts/check-all.sh` gate, >=90% branch coverage aggregate and >=80% per file, >=95% docstring coverage, complexity <=10, mypy strict, zero lint/type suppressions).

## Goal
Give the operator a way to *find* vault files that the pipeline silently skips: a `creek lint` check that reports every file under `01-Fragments` whose front matter cannot be parsed or read, so a corrupt fragment gets repaired instead of quietly disappearing from every scan.

## Context
Surfaced by the Gate 2.5 review of #847. #847 fixed a real crash — `creek.reflect` no longer dies on a corrupt fragment — but the fix is a *silent* skip (`logger.debug` + `continue`, at DEBUG, which is off at the default WARNING level). The same silent-skip pattern already exists in `PurgeEngine._load_frontmatter` (`creek/purge/engine.py:882-886`) and `HygieneEngine._read_content` (`creek/clean/hygiene.py:785-789`).

Net effect today: a corrupt fragment is invisible. It is skipped by reflect, skipped by purge, skipped by hygiene — and nothing anywhere tells the user it exists. The failure mode is quiet data loss from the working set: `entry_ref` resolution returns the ordinary `"entry_ref not found"` refusal, indistinguishable from a fragment that was never there (deliberately so — a distinct reason would be an existence oracle, see #847). So the *tool* surface cannot report it; a lint pass is the right place.

Two parts, both worth doing:
1. A `creek lint` check (see `creek/lint/checks/` for the existing check shape) that walks `01-Fragments`, attempts the load, and reports each unparseable/unreadable path with the exception class name. Report the class name only, never `str(exc)` — `yaml.MarkedYAMLError.__str__` embeds the offending source snippet, which would put tier-unknown vault content into lint output.
2. Optionally, in the silent-skip sites, an aggregated `logger.warning` carrying a **count** of files skipped during a scan (never an id, never a path, never the probed `entry_ref`) so a total-outage scenario — whole `01-Fragments` unreadable, every lookup silently becoming "not found" — is operable rather than invisible.

## Output Format
[...]

Issue #925, opened 2026-07-24
[scan:security] creek_mcp/ is excluded from bandit and pylint — the MCP auth/tier/audit surface is unscanned


Geoffe-Ga:
## Role
You are a senior Python engineer working in this project's codebase, following its existing conventions (the `cd creek-tools && ./scripts/check-all.sh` gate, zero lint/type suppressions, no lowering of thresholds).

## Goal
Bring `creek-tools/creek_mcp/` under the same static-analysis gates as `creek-tools/creek/` — bandit and pylint — and burn down whatever findings that surfaces, so the MCP surface is not the one unscanned package in the repo.

## Context
Surfaced by the Gate 2.5 review of #847. Every static scanner in the gate targets `creek/` explicitly and therefore silently excludes `creek_mcp/`:
- `creek-tools/scripts/security.sh:73` — `bandit -r creek/ -ll`
- `.github/workflows/ci.yml:117-121` — `bandit -r creek/ -f json ...` and `bandit -r creek/ -ll`
- `creek-tools/scripts/pylint.sh:69,77` — `python -m pylint creek/ --fail-under=...`
- `.github/workflows/ci.yml:85-91` — same, via `./scripts/pylint.sh`
- `scripts/lint-tryceratops.sh`, `scripts/lint-interrogate.sh`, `scripts/complexity.sh`, `scripts/lint-refurb.sh`, `scripts/typecheck.sh` are also `creek/`-scoped (confirm each; some may be deliberate).

This is not a cosmetic gap. `creek_mcp/` is where the security-relevant MCP surface lives: bearer/elevated token verification (see #897, #913), the read-side privacy-tier ceiling (#846/#921), `tier_ceiling.py`'s admission predicate (#923), and the `MCPAuditLog` append chain. Ruff, mypy (via the pre-commit hook), pytest and coverage DO cover `creek_mcp/`, so the gap is specifically bandit + pylint + the `creek/`-scoped extended linters.

## Output Format
A single PR that: (1) widens the scanner targets, (2) fixes — not suppresses — whatever it surfaces, (3) passes `cd creek-tools && ./scripts/check-all.sh`, and (4) references this issue with "Closes #N".

## Constraints
- Expect this to surface real findings; budget for them. If the pylint score for `creek_mcp/` lands below the 9.0 floor, fix the code rather than carving out a separate lower threshold.
[...]

Issue #924, opened 2026-07-24
[scan:bugs] frontmatter.load guards across creek/ are too narrow: YAMLError uncaught at 9 sites, TypeError at 18


Geoffe-Ga:
## Role
You are a senior Python engineer working in this project's codebase, following its existing conventions (TDD via stay-green, the `cd creek-tools && ./scripts/check-all.sh` gate, >=90% branch coverage aggregate and >=80% per file, >=95% docstring coverage, complexity <=10, mypy strict, zero lint/type suppressions).

## Goal
Audit and fix every `frontmatter.load` / `frontmatter.loads` call site in `creek-tools/creek/` whose exception guard is too narrow to actually survive a corrupt vault file, so one hand-edited fragment cannot crash an unrelated pipeline stage.

## Context
Spun out of the Gate 2.5 review of #847 (which fixed the same class of bug in `creek_mcp/tools/reflect.py::_resolve_entry`). Two distinct holes, both verified empirically at HEAD against the installed `python-frontmatter` + PyYAML:

**Hole A — `yaml.YAMLError` is not caught at all (the more severe one).**
9 sites guard a frontmatter load with `except (OSError, ValueError):`. `yaml.YAMLError` derives directly from `Exception`, NOT from `ValueError` — confirmed: `yaml.YAMLError.__mro__ == (YAMLError, Exception, BaseException, object)` and `issubclass(yaml.parser.ParserError, ValueError) is False`. So plain malformed YAML front matter (e.g. `foo: [unterminated`) still raises straight through these guards. Examples:
- `creek/lint/checks/compost.py:32-34`
- `creek/lint/checks/synchronicity.py:32-34`

**Hole B — `TypeError` from a non-string front-matter key.**
18 sites guard with `except (OSError, ValueError, yaml.YAMLError):`. `python-frontmatter` splats parsed metadata as `Post(content, handler, **metadata)`, so a non-string key raises a bare `TypeError`, which none of those three cover. Verified:
```
>>> frontmatter.load(p)   # p contains: ---\n2024-05-01: note\nid: frag-x\n---\nbody\n
TypeError: keywords must be strings
```
[...]

Issue #923, opened 2026-07-24
[scan:bugs] tier_allowed's bare _TIER_RANK subscript would KeyError instead of failing closed on a new PrivacyTier


Geoffe-Ga:
Found during the #846 security review (PR #921), out of scope there.

## Evidence
`creek-tools/creek_mcp/tier_ceiling.py` `tier_allowed` (~L76-85) ends with:

```python
return _TIER_RANK[tier] <= _CEILING_RANK[ceiling]
```

`_TIER_RANK` (~L50-55) enumerates `OPEN`, `UNCLASSIFIED`, `PERSONAL`, `INTIMATE` explicitly. Every *other* unknown-tier path in the codebase fails closed via `.get(..., <most restrictive>)` — e.g. `creek_mcp/tools/reflect.py` `_sensitivity`, `_CEILING_ROUTING_TIER.get(...)`, and `_fragment_tier`'s `except ValueError` — but this one is a bare subscript.

Not reachable today: `PrivacyTier` is a closed enum and all callers normalise strings through `PrivacyTier(...)` first. But `tier_allowed` is now the single canonical read-side admission predicate (it is what the #846 gate delegates to), so a future `PrivacyTier` member added without updating `_TIER_RANK` would raise `KeyError` out of the admission check rather than refusing. Depending on the caller, that is a crash across the MCP boundary — and any `except`-and-continue caller would turn it into an admission.

## Goal
Make the canonical admission predicate fail closed on an unranked tier, consistent with the rest of the module's defaults: rank an unknown tier as the most restrictive (`_TIER_RANK.get(tier, _TIER_RANK[PrivacyTier.INTIMATE])`), and do the same for `_CEILING_RANK` in the same expression. Add a test that pins the behaviour for a tier not present in the map, and a docstring line stating the fail-closed contract so the invariant is explicit rather than incidental.

## Related
- #846 / PR #921 made `tier_allowed` the shared chokepoint for `creek.reflect`'s read-side gate.
- Re-verify against HEAD before starting; close with a comment if it no longer reproduces.

Geoffe-Ga:
**This may already be fixed on `main`** — worth verifying and closing rather than working.

Checked at `a580ce5`. The bare `_TIER_RANK` subscript this issue describes no longer exists:

```
creek_mcp/tier_ceiling.py:104   return _TIER_RANK.get(tier, _TIER_RANK[PrivacyTier.INTIMATE])
[...]

PR #927, opened 2026-07-24
fix(mcp): skip unreadable fragments in creek.reflect entry_ref resolution

### LINKED ISSUE #847: [scan:bugs] creek.reflect crashes on malformed frontmatter in 01-Fragments (unguarded frontmatter.load)
## Role
suppressions).

## Goal
Make `creek.reflect` tolerate malformed frontmatter in any `01-Fragments/*.md`
file — resolving `entry_ref` must skip unreadable files (debug-logged) and
return a structured dict, never raise `yaml.YAMLError` across the MCP boundary —
verified by a test with a corrupt sibling fragment present.

## Context
- File(s): `creek-tools/creek_mcp/tools/reflect.py:249-254` (`_resolve_entry`).
- Evidence: `_resolve_entry` iterates
  `(vault_path / "01-Fragments").rglob("*.md")` and calls
  `frontmatter.load(path)` (reflect.py:249-250) with **no** `try`/`except`.
  `python-frontmatter` raises `yaml.YAMLError` (ScannerError/ParserError) on
  malformed YAML frontmatter. The reflect module elsewhere goes out of its way
  to never raise across the MCP boundary — `_parse_notes` explicitly catches
  `yaml.YAMLError` for exactly this reason (reflect.py:150-155) — and
  `server.py`'s tool wrapper adds no boundary catch. Every other vault walk in
  the codebase guards the load (`PurgeEngine._load_frontmatter`
  engine.py:789-793 catches `Exception`; `hygiene._read_content`
  hygiene.py:770-774 catches `Exception`); this one does not. A single corrupt
  fragment anywhere under `01-Fragments` that sorts before the target turns any
  `entry_ref` reflection into an unhandled crash.
  body survives purge, reflect entry_ref intimate egress, compile write-side
  tier-ceiling gap.

## Output Format
"Closes #N".

## Examples

```python
def test_reflect_survives_malformed_sibling_frontmatter(tmp_path):
    vault = _init_vault(tmp_path)
    frags = vault / "01-Fragments"
    # sorts before the target and has broken YAML frontmatter
    (frags / "aaa-bad.md").write_text(
        "---\nfoo: [unterminated\n---\nbody\n", encoding="utf-8"
    )
    _write_fragment(vault, id="frag-target", tier="open", body="entry text")

    result = reflect_tool(
        vault_path=vault,
        llm_factory=lambda _t: (lambda _p: '{"notes": []}'),
        entry_ref="frag-target",
    )

    assert isinstance(result, dict)  # today: raises yaml.YAMLError
```

Fix sketch (mirror `PurgeEngine._load_frontmatter`):

```python
for path in (vault_path / "01-Fragments").rglob("*.md"):
    try:
        post = frontmatter.load(path)
    except Exception:  # noqa: BLE001 — skip unreadable fragment, keep scanning
        logger.debug("Unable to parse frontmatter in %s", path)
        continue
    if str(post.metadata.get("id", "")) == entry_ref:
        return post.content, _fragment_tier(post.metadata)
```

(Resolve the correct exception-hygiene form for `check-all.sh` — the repo's
tryceratops/ruff gate may prefer a narrower catch or a documented `# noqa`.)

## Constraints
- No lint/type suppressions (max-quality-no-shortcuts): fix root causes; if a
  broad-except carve-out is genuinely needed, follow the repo's documented
  `# noqa: BLE001` / `TRY` convention with a one-line justification.

Geoffe-Ga:
## Summary

`creek.reflect`'s `_resolve_entry` walked `(vault_path / "01-Fragments").rglob("*.md")` calling `frontmatter.load` with **no guard**, so a single corrupt fragment anywhere under `01-Fragments` turned *every* `entry_ref` reflection into an unhandled exception crossing the MCP boundary — not a structured refusal. The corrupt file did not have to be related to the target; whichever one the scan reached first decided whether the tool worked at all.

This was the one path in the module that broke its own never-raises contract. `_parse_notes` already catches `yaml.YAMLError` for exactly this reason, `_default_retrieve` catches `Exception`, and `server.py`'s tool wrapper adds no boundary catch — so the tool is the last line of defense. Every other vault walk in the codebase guards the load (`PurgeEngine._load_frontmatter`, `hygiene._read_content`); this one did not.

The fix wraps only the load: skip, debug-log, `continue`. Three properties are deliberate.

**Fail-closed.** An unparseable fragment can never be *returned* as an entry either, so it is unreadable-by-anyone rather than readable-without-a-tier. Its front matter never became usable metadata, so its `privacy_tier` is unknown and unknowable; salvaging the id or body out of the raw text would hand an `open`-ceiling caller content nobody can vouch for — walking *around* the #846 read-side gate rather than through it. `continue` is the only correct action. The id comparison and `_fragment_tier` stay outside the `try`, so tier derivation keeps its own INTIMATE fail-closed default.

**No new oracle.** A skip is indistinguishable from not-found: same status, same `"entry_ref not found"` reason, same single audit record. A distinct `"entry_ref unreadable"` reason would confirm to an unadmitted caller that a fragment with that id exists — in the one place where the tier is unknown, so nothing can be reasoned about it.

**The #846 gate is untouched.** The gate landed in #921 and deliberately sits after the unconditional `MCPAuditLog` append and before `care_guard`, because an `escalate` response would otherwise be a one-bit oracle to an unadmitted caller. Neither the gate, its position, nor the audit append is modified. Its accepted-residual-risk note is amended by one phrase only — the not-found path *attempts to* parse every fragment file — since corrupt files now abort mid-parse rather than parsing cleanly.

### Why `except Exception` and not the house tuple

`frontmatter.load` splats parsed metadata as `Post(content, handler, **metadata)`, so front matter with a **non-string key** — a bare YAML date such as `2024-05-01: note`, a realistic hand-edited-vault case — raises a bare `TypeError`. That is not an `OSError`, not a `ValueError`, and not a `yaml.YAMLError`, so the house `(OSError, ValueError, yaml.YAMLError)` tuple used at other call sites would miss it entirely. Verified at HEAD:

```
>>> frontmatter.load(p)   # p: ---\n2024-05-01: note\nid: frag-x\n---\nbody\n
TypeError: keywords must be strings
```

The debug log carries `type(exc).__name__` **only**, never `str(exc)` / `exc_info` / `logger.exception`: `yaml.MarkedYAMLError` stringifies with the offending source snippet, which would write tier-unknown vault content into the log. That constraint is recorded in a comment so it is not "improved" into a leak later.

## Test plan

Eight new tests in `tests/test_mcp_reflect.py`, all verified RED before the fix and mutation-tested after:

- **Revert the guard entirely → all 8 fail.**
- **Narrow the catch to `(OSError, ValueError, yaml.YAMLError)` → the non-string-key test fails, the other 7 pass.** The breadth is regression-tested, not just asserted in prose.

Coverage: `creek_mcp/tools/reflect.py` 93.75% → **94.58%**; the previously-partial `rglob` loop branch is now fully covered. No coverage waiver touched.

Failure modes pinned individually, each with an order-independent test whose vault contains only the corrupt file so the loop must exhaust it on every filesystem: `yaml.ParserError` (malformed YAML), `UnicodeDecodeError` (invalid UTF-8), `IsADirectoryError`/`OSError` (a directory named `*.md`, which `rglob` sweeps), and `TypeError` (non-string key).

**An ordering false-green was found and closed.** `rglob` yields `os.scandir` order, which `pathlib` never sorts and which is hash-ordered on APFS/XFS. A corrupt sibling "named to sort first" can be visited *after* the target, in which case `_resolve_entry` early-returns at the id match and the test passes against the live bug — silently and permanently. One test did exactly that on the first run. The fix is not more names: each sibling test now asserts, via a spy on `frontmatter.load` plus a second probe that cannot early-return, that the sweep actually reached every file. That assertion also kills a `break`/early-return mutant that naming alone could not distinguish.

Gate 2: `cd creek-tools && ./scripts/check-all.sh` → exit 0, 12/12 checks, 6098 tests pass.

## Follow-ups filed (out of scope here)

- #924 — the same guard is too narrow at 27 other `frontmatter.load` sites in `creek/`: `yaml.YAMLError` uncaught at 9, `TypeError` at 18.
- #925 — `creek_mcp/` is excluded from bandit and pylint; the MCP auth/tier-ceiling/audit surface is unscanned.
- #926 — silently skipped fragments are never reported to the operator; a corrupt fragment is invisible to reflect, purge and hygiene alike.

Already tracked and deliberately untouched: #922 (stale `privacy_tier` surviving a body rewrite), #923 (`tier_allowed`'s bare `_TIER_RANK[tier]` subscript).

Closes #847

### DIFF
--- a/creek-tools/creek_mcp/tools/reflect.py
+++ b/creek-tools/creek_mcp/tools/reflect.py
@@ -40,12 +40,19 @@ Four guarantees this module enforces:
   :func:`creek.care.guardrail.acute_distress_guard`, threaded in by
   ``build_server``; the seam stays injectable so tests can substitute their own.
 
+Underneath those guarantees, every read and parse failure answers with a
+structured dict rather than raising: a malformed or unreadable file under
+``01-Fragments`` is skipped during ``entry_ref`` resolution (#847), and a
+malformed model turn degrades to no notes. No parse error crosses the MCP
+boundary.
+
 The LLM and retrieval are injected so the tool is unit-testable with no live
 calls; ``build_server`` supplies the production factory + retrieval.
 """
 
 from __future__ import annotations
 
+import logging
 from typing import TYPE_CHECKING, Any, Protocol
 
 from creek.care.guardrail import CARE_POLICY, CARE_SIGNAL
@@ -61,6 +68,8 @@ from creek_mcp.tier_ceiling import (
 if TYPE_CHECKING:
     from collections.abc import Callable
     from pathlib import Path
+
+logger = logging.getLogger(__name__)
 
 TOOL_NAME = "creek.reflect"
 
@@ -290,6 +299,21 @@ def _resolve_entry(
     :func:`creek.classify.privacy_filter.tier_of`, not by this fail-closed
     path. A blank result yields ``("", None)``, which the caller turns into
     a refusal.
+
+    A fragment file that cannot be parsed or even read is skipped, debug-logged,
+    and the scan continues (#847) — one hand-edited or half-written file must not
+    take down ``entry_ref`` resolution for the whole vault. The skip is
+    fail-closed: such a fragment can never be *returned* as an entry either, so
+    it is unreadable-by-anyone rather than readable-without-a-tier. Its front
+    matter never became usable metadata — the file may not even have been read —
+    so its ``privacy_tier`` is unknown and unknowable, and salvaging the id or
+    the body out of the raw text would hand an ``open``-ceiling caller content
+    nobody can vouch for — walking around the #846 gate rather than through it.
+    The skip is also deliberately indistinguishable from not-found (both land on
+    the caller's ``"entry_ref not found"`` refusal): a distinct "unreadable"
+    reason would confirm to an unadmitted caller that a fragment with that id
+    exists, a new existence oracle over the corpus in the one place where the
+    tier is unknown.
     """
     if content and content.strip():
         return content, None
@@ -297,7 +321,28 @@ def _resolve_entry(
         import frontmatter
 
         for path in (vault_path / "01-Fragments").rglob("*.md"):
-            post = frontmatter.load(path)
+            try:
+                post = frontmatter.load(path)
+            except Exception as exc:
+                # Broader than the house ``(OSError, ValueError, yaml.YAMLError)``
+                # tuple on purpose. ``frontmatter.load`` splats the parsed
+                # metadata as ``Post(content, handler, **metadata)``, so front
+                # matter with a non-string key — a bare YAML date such as
+                # ``2024-05-01: note``, a realistic hand-edited-vault case —
+                # raises ``TypeError``, which is none of those three (see
+                # ``test_nonstring_frontmatter_key_is_skipped_not_crashed``). The
+                # corpus is user-managed and arbitrarily messy, so the failure
+                # surface is open-ended *beyond* decode/IO/YAML, and per the
+                # never-raises contract none of it may cross the MCP boundary.
+                #
+                # Log the exception's *type name only* — never ``str(exc)``,
+                # ``exc_info``, or ``logger.exception``. ``yaml.MarkedYAMLError``
+                # stringifies with the offending source snippet, so the message
+                # text would write tier-unknown vault content into the log.
+                logger.debug(
+                    "Skipping unreadable fragment %s (%s)", path, type(exc).__name__
+                )
+                continue
             if str(post.metadata.get("id", "")) == entry_ref:
                 body: str = post.content
                 return body, _fragment_tier(post.metadata)
@@ -341,7 +386,9 @@ def reflect_tool(
         ``{status, tool, tier_ceiling, ...}`` — ``ok`` with ``notes`` (+ optional
         ``essay``), ``escalate`` with a ``reason``, or a structured ``refused``
         response for one of: no *entry_ref*/*content* resolves to text
-        (``"entry_ref not found"`` / ``"no entry content supplied"``); the
+        (``"entry_ref not found"`` — which also covers a fragment whose front
+        matter is unparseable or unreadable, see :func:`_resolve_entry` — or
+        ``"no entry content supplied"``); the
         *entry_ref*'s classified tier exceeds *privacy_tier_ceiling*
         (``"entry_ref tier exceeds ceiling"``, #846 — see :func:`_above_ceiling`);
         or the LLM provider is unavailable/raises (``"reflection unavailable:
@@ -396,9 +443,9 @@ def reflect_tool(
             # probe costs a vault scan — while the distinct not-found reason is
             # what makes a legitimate client's bug debuggable. If the two reasons
             # are ever unified, the scan must be equalised too: the not-found
-            # path parses every fragment file whereas this path early-returns at
-            # the match, so the timing difference would preserve the oracle as a
-            # side channel.
+            # path attempts to parse every fragment file whereas this path
+            # early-returns at the match, so the timing difference would preserve
+            # the oracle as a side channel.
             reason="entry_ref tier exceeds ceiling",
         )
 
--- a/creek-tools/tests/test_mcp_reflect.py
+++ b/creek-tools/tests/test_mcp_reflect.py
@@ -20,7 +20,8 @@ contract this suite pins, in order of stakes:
 3. **Quotes are verbatim** — every returned ``quote`` is a substring of the
    input; model-supplied spans that are not are dropped, never trusted.
 4. Corpus-grounded retrieval, audit logging, read-only wrt the corpus, clean
-   degradation with no provider, and the care seam (#753).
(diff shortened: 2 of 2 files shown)


Geoffe-Ga:
## Summary
`_resolve_entry` in `creek_mcp/tools/reflect.py` walked every fragment under `01-Fragments` calling `frontmatter.load` with no guard, so a single corrupt/unreadable fragment anywhere in the vault turned an unrelated `entry_ref` lookup into an unhandled exception crossing the MCP boundary. This PR wraps just the `frontmatter.load(path)` call in `except Exception: log type name, continue`, keeping the tier-derivation logic outside the guarded block, and adds 8 new tests covering the distinct failure modes (`yaml.YAMLError`, `UnicodeDecodeError`, `OSError`/`IsADirectoryError`, and a `TypeError` from non-string frontmatter keys).

## Strengths
- **Root-caused, not papered over.** Confirmed via `grep` that the repo's usual guard is `except (OSError, ValueError, yaml.YAMLError)` at 17 other call sites, and that this really would miss the `TypeError` `frontmatter.loads(**metadata)` raises for a non-string YAML key (e.g. a bare `2024-05-01:` date) — the broader `except Exception` here is justified, documented, and scoped to exactly one line.
- **Security invariants preserved, not just claimed.** The skip is indistinguishable from "not found" (same reason string, verified by a whole-dict-equality test), so it introduces no new existence oracle; the debug log carries only `type(exc).__name__`, never `str(exc)`, so a `yaml.MarkedYAMLError`'s embedded source snippet (i.e., vault content) can't leak into logs; the #846 read-side ceiling gate, its position, and the single unconditional audit append are all untouched.
- **Genuinely rigorous test suite.** Beyond the four failure-mode arms, the author caught and fixed their own ordering false-green (`rglob`/`os.scandir` order isn't guaranteed, so a naive "corrupt sibling present" test could pass without the skip path ever running) by asserting via a `frontmatter.load` spy that every named file was actually swept — a real find documented plainly in the PR body and test docstrings.
- **Scope discipline.** Three related-but-broader problems (the same narrow-tuple gap at 27 other sites, `creek_mcp/` excluded from bandit/pylint, silent skips invisible to the operator) were filed as follow-ups (#924/#925/#926) rather than folded in here.
- Consistent with existing precedent elsewhere in the codebase (`PurgeEngine._load_frontmatter`, `hygiene._read_content` also already use a bare `except Exception` for the same "corpus is arbitrarily messy" reason).

## Security Concerns
None found. 🟢 The change is fail-closed (an unparseable fragment can never be resolved as an entry, by anyone), avoids a new information oracle, and avoids logging tier-unknown content. No new attack surface introduced.

## Problems
None blocking.

## Code Quality
- Minor, non-blocking: the debug log includes the fragment's file path (not its content). Path/filenames are unlikely to be sensitive in this system, and it's debug-level only, but worth keeping in mind if vault filenames are ever derived from content.
- The `_resolve_entry` docstring has grown quite long (three paragraphs of security rationale). Given this repo's documentation-heavy style and that the rationale is genuinely load-bearing (explains *why* not to "simplify" the catch later), this is appropriate rather than bloat.

## CLAUDE.md Adherence
✅ TDD workflow followed (PR states 8 tests written and verified RED before the fix). ✅ `./scripts/check-all.sh` reported 12/12 checks passing, 6098 tests, per the PR description. ✅ Coverage improved (93.75% → 94.58%) rather than waived. ✅ No `noqa`/`type: ignore`/threshold-lowering shortcuts present in the diff. ✅ Docstrings extensive and Google-style-consistent with the rest of the module. Note: CI was still `IN_PROGRESS` on the main "Code Quality & Testing" jobs at review time (CrawDad jobs already green); worth confirming those finish green before merge per Gate 3, though nothing in the diff suggests they won't.

## Testing
Strong. 8 new tests exercise each distinct exception arm, the fail-closed contract (corrupt fragment never resolvable), the no-new-oracle contract (skip == not-found byte-for-byte), non-bypass of the #846 ceiling gate, and — notably — an explicit defense against ordering-dependent false greens via a `frontmatter.load` spy. This exceeds what's typically needed for a fix this size.

## Documentation
Module docstring, function docstring, and inline comments were all updated to describe the new never-raises behavior and the reasoning behind the broad `except Exception`, matching the implementation exactly. The #846 gate's "accepted residual risk" note was updated by exactly the one phrase the PR claims ("attempts to parse every fragment file"), which I verified against the diff.



## Verdict: LGTM

The fix is narrowly scoped, correctly justified against the codebase's existing conventions (verified `except (OSError, ValueError, yaml.YAMLError)` is indeed used at 17 other sites and would miss the documented `TypeError` case), preserves the security invariants it discusses (fail-closed, no new oracle, no content leak into logs), and ships with an unusually thorough, verified test suite. No bugs, security issues, or CLAUDE.md violations found; only trivial non-blocking notes remain.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5075461801 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #933, opened 2026-07-25
reflect.py residual-risk comment overstates fragment-id unguessability (ids are deterministic and child ids enumerable)


Geoffe-Ga:
## Context
Found while fixing #848. #848's own equivalent comment was corrected in that PR; `reflect.py` now carries the stale claim and the two should agree.

## The inaccuracy
`creek_mcp/tools/reflect.py` (the ACCEPTED RESIDUAL RISK block in `reflect_tool`'s ceiling gate, from #846/PR #921) justifies its refusal oracle partly with:

> fragment ids are unguessable (`frag-` + 12 hex), are only learnable from content already admitted to the caller

The second half is not true. Ids are **deterministic**:

- `creek/ingest/base.py::generate_fragment_id` — `sha256(f"{source}:{timestamp}:{content}")[:12]`
- `creek/ingest/base.py::generate_child_fragment_id` — `sha256(f"{parent_id}:{level}:{index}")[:12]`

Child ids are derivable from a single known parent id by enumerating `index`, with **zero** knowledge of the content. So the id space around known content is sweepable; only a blind sweep of the full 48-bit space is infeasible. (`creek/models.py::synthetic_fragment_id` is random, but is documented as the exception — production ingestors must use the deterministic generator.)

The comment also omits a real mitigation worth stating: above-ceiling probes are audited, so a sweep is detectable by rate.

## Ask
Correct the wording to rest on the audit trail rather than on id entropy. `creek_mcp/tools/compile.py`'s `_ABOVE_CEILING_REASON` comment (from #848) has the corrected phrasing to mirror, including the probe-cost timing asymmetry note.

Comment-only change; no behaviour change. The gate itself is correct.

Refs #846, #848

Issue #932, opened 2026-07-25
Sweep: audit every MCP tool that reads vault fragments for a source-side tier-ceiling gate


Geoffe-Ga:
## Context
Two instances of the same bug class have now been fixed one at a time:

- #846 / PR #921 — `creek.reflect` resolved an `entry_ref` fragment without checking its classified tier against the caller's ceiling (read side).
- #848 — `creek.compile` rolled up source `fragment_ids` without checking their tiers (write side).

Both had docstrings asserting a ceiling rule that the code did not implement.

## Ask
Audit every MCP tool that reads vault fragments the caller did **not** supply inline, and confirm each checks the resolved fragment's classified tier against `privacy_tier_ceiling`. Candidates: `creek.link`, `creek.report`, `creek.draft`, `creek.author`, `creek.mine`, `creek.wheel`, `creek.state.read`, `creek.lint`.

For each, record either the gate's location or an explicit note that the tool cannot reach above-ceiling content and why.

## Suggested guardrails
- A shared helper so a new tool gets the gate by default rather than by remembering.
- A test that enumerates the registered tool surface and asserts each read-of-unsupplied-content tool has a gate, so a newly added tool fails the suite until it is triaged.

## Prior art to copy
- `creek_mcp/tools/reflect.py` — `_above_ceiling` plus the gate-ordering comment (an `escalate` response is a one-bit oracle to an unadmitted caller).
- `creek_mcp/tools/compile.py` — `_sources_above_ceiling`, the fail-closed `_tier_of`, the fixed generic refusal reason, and the audited-refusal ordering.

Both establish the same two rules: gate ordering is the substance, and refusals must not become a config or content oracle.

Refs #846, #848

Geoffe-Ga:
Umbrella note: this sweep already has two confirmed instances filed separately — #928 (creek.compile's LLM is not tier-routed, so admitted intimate source titles egress) and #931 (structural_path ancestry leaks above-ceiling parent titles into the compile prompt). Two more MCP read paths were hardened today and should be treated as *done*, not re-audited: creek.reflect entry_ref (#846 / PR #921) and creek.compile source selection (#848 / PR #934).

Whoever takes this sweep should enumerate the remaining unaudited tools rather than re-deriving the whole surface, and close it by filing one issue per real gap — not by fixing them inline.

Geoffe-Ga:
Sweep done — PR #973 (the full tool-by-tool table is in the PR body).

[...]

Issue #931, opened 2026-07-25
structural_path ancestry leaks above-ceiling parent titles into the compile prompt via an admitted child


Geoffe-Ga:
## Context
Found while fixing #848, which gates the tiers of the fragment ids a caller **names**.

## The gap
`creek/compile/engine.py::_build_prompt` emits `structural_path:` for every admitted source, and `creek/compile/hierarchy.py::structural_path_context` prefers the fragment's persisted `structural_path` — a list of **parent titles**.

So an `open`-tier child of an above-ceiling parent is admitted by the #848 gate (its own tier is fine), yet still carries its parent's title into the compile prompt — and potentially into the LLM-authored page body and its provenance.

#848's gate checks only the requested ids, so its guarantee is precisely "the fragments you named", not "everything the prompt will see". The module docstring in `creek_mcp/tools/compile.py` was tightened to say `named` for exactly this reason.

## Suggested fix
Either check ancestry tiers alongside the named ids, or strip/redact `structural_path` entries whose owning fragment is above the caller's ceiling. The former is more complete; the latter is cheaper and local to `_build_prompt`.

Worth auditing whether the same ancestry channel exists in other tools that render `structural_path`.

Refs #848

Geoffe-Ga:
**Cross-reference from #1036** (layer (g): a prompt-channel canary layer for the MCP read surface).

#1036 adds a `creek.compile` prompt probe that captures what crosses to the model and asserts no above-ceiling sentinel appears in it. That probe runs over a **flat** fixture — three sibling fragments, no ancestry — so it does *not* cover the channel this issue is about. `creek/compile/engine.py:451-459` emits

```
sections.append(f"  structural_path: {' > '.join(breadcrumb)}")
```

for every admitted source, so an admitted open child of an above-ceiling parent carries that parent's **title** into the compile prompt. A flat corpus has no breadcrumb, so #1036's probe passes and says nothing about this.

That limitation is stated explicitly in the probe's own docstring (the same way #1068 stated `creek.report`'s), precisely so a green compile probe cannot be misread as certifying this surface.

[...]

Issue #930, opened 2026-07-25
creek.compile walks 01-Fragments twice per call; closes a TOCTOU window and halves I/O to load once


Geoffe-Ga:
## Context
Found while fixing #848, which added a source-tier ceiling gate to `compile_tool`.

## The problem
`creek_mcp/tools/compile.py::_sources_above_ceiling` walks `01-Fragments` via `iter_vault_fragments`, then `creek/compile/engine.py::_load_fragments_for_compile` walks it **again** with the same loader. Every admitted compile therefore parses the whole fragment corpus twice. At 35k fragments that is a doubled `frontmatter.load` pass, and it hands a remote consumer 2x amplification.

Sharing the loader was deliberate (#848): it guarantees the gate and the engine can never disagree about which files exist. But the parity is by construction, not instantaneous.

## The TOCTOU window
Between the gate's walk and the engine's walk, a concurrent writer that *raises* a fragment's tier (`creek classify`, the `creek.classify` MCP tool, a text editor) would have that fragment compiled under its pre-write tier. Not caller-controlled and low likelihood, but real.

## Suggested fix
Load once in the wrapper and hand the `(Fragment, body)` pairs to `compile_to_vault`. This closes the window and halves the I/O. It requires an engine signature change, which is why #848 explicitly deferred it.

While in there: `_load_fragments_for_compile` tests membership with `if fragment.id in requested` against a **list** (O(n*k)); a set would make it O(n).

Refs #848

PR #934, opened 2026-07-25
fix(mcp): enforce creek.compile's documented write-side tier ceiling

### LINKED ISSUE #848: [scan:bugs] creek.compile does not enforce its documented write-side tier ceiling
## Role
suppressions).

## Goal
Enforce the write-side tier-ceiling rule the `creek.compile` docstring promises:
refuse a compile whose source `fragment_ids` include any fragment whose tier
exceeds the caller's `privacy_tier_ceiling`, instead of silently compiling it —
verified by a test that an OPEN-ceiling compile of an INTIMATE fragment refuses.

## Context
- File(s): `creek-tools/creek_mcp/tools/compile.py:1-108` (`compile_tool` and
  module docstring).
- Evidence: the module docstring states (compile.py:3-5): "The write-side
  tier-ceiling rule applies: a caller cannot create a compiled page whose source
  fragments include a tier they could not read." But `compile_tool`
  (compile.py:55-108) **never** loads the source fragments' tiers, **never**
  imports or calls `write_tier_allowed`, and **never** passes
  `privacy_tier_ceiling`/an override into `compile_to_vault` (the engine call at
  compile.py:94-102 receives only `fragment_ids`, `vault_path`, `target_kind`,
  `target_id`, `target_title`, `llm`). The ceiling is used only to echo
  `refusal_response` for unrelated validation errors and in the audit entry. By
  contrast `save_tool` and `journal_ingest_tool` both gate writes via
  `write_tier_allowed`. Result: with `ceiling=OPEN` a caller can compile
  INTIMATE/PERSONAL fragment IDs; the call succeeds (`status="ok"`/`"noop"` and
  a page is written) instead of refusing — and intimate titles/IDs flow into the
  compiled page provenance and the (possibly cloud) compile LLM.
  body survives purge, reflect entry_ref intimate egress, reflect unguarded
  `frontmatter.load` crash. Confirm the intended semantics against
  `creek_mcp/tier_ceiling.py` (`write_tier_allowed`) and the `save`/`journal`
  gates before implementing.

## Output Format
"Closes #N".

## Examples

```python
def test_compile_refuses_source_fragment_above_ceiling(tmp_path):
    vault = _init_vault(tmp_path)
    _write_fragment(vault, id="frag-int", tier="intimate", body="secret")

    result = compile_tool(
        vault_path=vault,
        fragment_ids=["frag-int"],
        target_kind="eddy",
        target_id="e1",
        target_title="T",
        llm_factory=lambda: (lambda _p: '{"body": "x"}'),
    )

    assert result["status"] == "refused"
```

Today the call returns `status="ok"`/`"noop"` and writes a compiled page. After
the fix — load each `fragment_ids` tier and refuse via `refusal_response` when
any exceeds the ceiling per `write_tier_allowed`, before calling
`compile_to_vault` — it refuses.

Fix sketch (guard before the engine call at compile.py:93):

```python
above = _fragments_above_ceiling(fragment_ids, vault_path, privacy_tier_ceiling)
if above:
    return refusal_response(
        tool=TOOL_NAME, ceiling=privacy_tier_ceiling,
        reason=f"source fragments exceed ceiling: {', '.join(above)}",
    )
```

## Constraints
- Reuse `creek_mcp.tier_ceiling.write_tier_allowed` — do not re-derive tier
  ordering. Mirror the existing `save_tool`/`journal_ingest_tool` gate.

Geoffe-Ga:
## Summary

`creek.compile`'s module docstring claimed *"a caller cannot create a compiled page whose source fragments include a tier they could not read"* — but `compile_tool` never loaded the source fragments' tiers, never called `write_tier_allowed`, and never passed the ceiling into `compile_to_vault`. The ceiling was used only to echo unrelated validation refusals and to stamp the audit entry.

With `ceiling=open`, a caller could compile INTIMATE fragment ids. The call returned `status="ok"` and wrote a page, sending intimate ids and titles into three places at once: the compiled page's provenance, the paradox log, and the compile LLM prompt — which is **not** tier-routed.

This adds the gate the docstring already promised.

### Design

**`_sources_above_ceiling`** walks the vault through `iter_vault_fragments` — deliberately the *same* loader `compile_to_vault` uses — so the set of files the gate inspects and the set the engine compiles are identical by construction. A bespoke `frontmatter.load` scan would diverge (the shared reader rejects `type != fragment` and schema failures that a raw scan reads happily), and that divergence is the exact bug class this gate exists to prevent. Admission goes through `write_tier_allowed`, mirroring `save_tool` / `journal_ingest_tool`.

**Gate ordering is the substance**, following the read-side precedent set by #846 / PR #921 in `reflect.py`:

- It sits *below* the `target_kind` and empty-`fragment_ids` checks. Neither order leaks — those read only caller-supplied input and touch no vault state — so the tiebreaker is cost asymmetry: a full `01-Fragments` walk should not be reachable from a syntactically invalid call.
- It sits *above* everything after it, each of which is a distinct egress: `llm_factory()` is evaluated as an *argument* to the engine call and the prompt emits `id:`/`title:` unconditionally (`_fragment_excerpt_for_prompt` redacts only the body); the engine writes per-claim provenance naming the source ids into `02-Threads`/`03-Eddies`/`06-Frequencies`, which any open-ceiling read tool then serves; `_append_paradox_log` records them too; and the hash marker must not leave idempotency state behind on a refused call.
- The ceiling refusal deliberately wins over the engine's missing-id error, so a caller cannot pair one above-ceiling id with a batch of guesses and read back which guesses exist.

**Refusals are not a config oracle.** `_ABOVE_CEILING_REASON` is a fixed constant naming no ids, titles, tiers, or even a count — echoing which ids were above-ceiling would turn a single call into a bulk tier-classification oracle over the corpus. The `ACCEPTED RESIDUAL RISK` comment records what remains (a bare boolean is still a group-testing oracle) and rests the argument on the audit trail rather than on id entropy, because ids are deterministic and child ids are enumerable from a known parent.

**Two deliberate departures from #846**, both documented in the code:

1. *The refusal audits at the refusal site, not unconditionally at the top of the call.* `reflect` appends unconditionally; doing that here would append on every compile including no-ops, breaking the FEAT-011 acceptance test that asserts exactly one `creek.compile` entry across two calls. Compile's audit contract is "one entry per materially-new page", not "one per call". This follows `save`/`journal` instead.
2. *No `TierCeiling.ALL` fast path.* One route through `write_tier_allowed` only; a second would silently desync if `ALL`'s semantics changed.

**`_tier_of` fails closed.** A fragment file with *no* `privacy_tier` key would otherwise be defaulted by the model to `unclassified` (rank 0, admitted at every ceiling) — failing open on exactly the hand-edited or legacy file whose tier nobody can vouch for. `creek.reflect` already refuses that file closed to INTIMATE, and two MCP tools disagreeing about the same file is the divergence the shared-loader design exists to prevent. A fragment carrying an *explicit* `privacy_tier: unclassified` is untouched and stays admitted — that ranking is deliberate policy owned by `_TIER_RANK` (#923), and a test pins it so any change is deliberate.

## Test plan

- `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 checks, 6118 tests passed, aggregate coverage 93.92%, `creek_mcp/tools/compile.py` per-file 90.67% (gate 80%), ruff lint+format clean, mypy strict clean. No suppressions added.
- Verified genuine RED before implementing: 9 of the new tests failed against the unfixed code, each on the bug itself.

11 new tests in `tests/test_mcp_write_tools.py`:

- Acceptance: intimate source at `ceiling=open` is refused.
- **The LLM factory is never invoked on refusal** — the load-bearing egress assertion. Note the trap this avoids: `llm_factory()` is evaluated *inside* the pre-existing `try/except (ValueError, RuntimeError)`, so a test using a *raising* factory would see `status="refused"` even with no gate at all. Every test uses a counting `_RecordingLLMFactory` instead, and the allowed cases assert `calls == 1` so an unrelated short-circuit cannot masquerade as the gate passing.
- Refusal leaves no page, no paradox log, no hash marker — with a stub whose payload *would* produce both a page and a paradox, so each absent artifact is positive evidence the engine never ran.
- Refusal is audited with no ids, titles, or tiers — asserted against the raw `mcp.jsonl` bytes, and the hash chain still verifies.
- Above-ceiling source at any list position (both orderings parametrized; assertions are position-independent, so nothing depends on filesystem traversal order).
- Full tier/ceiling matrix mirroring `write_tier_allowed`.
- Missing `privacy_tier` key fails closed across all four ceilings.
- Explicit `unclassified` stays admitted at `open` (pins existing policy, points at #923).
- Missing ids still report "not found"; a ceiling violation outranks a missing id; unknown `target_kind` still wins over the gate.

The four pre-existing compile tests are untouched and green.

## Follow-ups filed (deliberately out of scope)

- #928 — the compile LLM is not tier-routed, so an *admitted* intimate source under a local `intimate`/`all` ceiling can still egress its title.
- #929 — `_build_compile_llm` imports a nonexistent `_default_llm`; MCP compile raises `ImportError` rather than a structured refusal.
- #930 — `01-Fragments` is walked twice per compile; loading once closes a TOCTOU window and halves the I/O (needs an engine signature change).
- #931 — `structural_path` ancestry can carry an above-ceiling parent's title in via an admitted child.
- #932 — sweep every MCP tool that reads unsupplied vault fragments for the same gate.
- #933 — correct the same residual-risk wording in `reflect.py`.

Closes #848

### DIFF
--- a/creek-tools/creek_mcp/tools/compile.py
+++ b/creek-tools/creek_mcp/tools/compile.py
@@ -1,8 +1,34 @@
 """``creek.compile`` MCP tool — roll fragments into a compiled page.
 
-Wraps :func:`creek.compile.engine.compile_to_vault` (FEAT-003). The
-write-side tier-ceiling rule applies: a caller cannot create a compiled
-page whose source fragments include a tier they could not read.
+Wraps :func:`creek.compile.engine.compile_to_vault` (FEAT-003).
+
+Four guarantees this module enforces (#848 — before that fix the first
+of them was asserted in this docstring and implemented nowhere):
+
+- **The write-side ceiling is checked against the *source* tiers, before
+  any state-affecting step.** Compile's variant of the FEAT-011 write-side
+  rule gates the classified tiers of the fragments being rolled up
+  rather than the tier of the page being created: a caller cannot
+  compile a page out of fragments they *named* and could not read.
+  (Only the named ids are checked — an admitted fragment's own
+  ``structural_path`` ancestry is not, which is a tracked follow-up.)
+  The check runs
+  ahead of any LLM client being built, any page being written, and any
+  paradox being logged — the three places the source ids would
+  otherwise escape (enumerated in :func:`compile_tool`).
+- **The refusal names nothing.** It carries the fixed
+  :data:`_ABOVE_CEILING_REASON`: no ids, no titles, no tiers, not even a
+  count of how many sources were above the ceiling.
+- **The refusal is audited, while the argument-validation refusals are
+  not.** An attempted ceiling violation is an operator-relevant signal;
+  a typo'd ``target_kind`` is not. This mirrors ``save`` / ``journal``
+  and is a deliberate departure from ``reflect``, which appends
+  unconditionally at the top of the call — :func:`compile_tool`
+  explains why compile cannot do the same.
+- **The gate and the engine share one fragment loader.** Both read the
+  vault through :func:`creek.vault.reader.iter_vault_fragments`, so the
+  two can never disagree about which files exist or which of them are
+  fragments (see :func:`_sources_above_ceiling`).
 
 **Idempotency semantics (audit-dedup only).** The wrapper fingerprints
 the target file after each compile and stamps the hash under
@@ -20,18 +46,62 @@ Engine-side LLM-call dedup is tracked separately as a follow-up.
 from __future__ import annotations
 
 import hashlib
-from typing import TYPE_CHECKING, Any, Protocol, cast
+from typing import TYPE_CHECKING, Any, Final, Protocol, cast
 
 from creek.compile.engine import TARGET_KINDS, compile_to_vault
+from creek.models import PrivacyTier
+from creek.vault.reader import iter_vault_fragments
 from creek_mcp.audit import MCPAuditLog
-from creek_mcp.tier_ceiling import TierCeiling, refusal_response
+from creek_mcp.tier_ceiling import (
+    TierCeiling,
+    refusal_response,
+    write_tier_allowed,
+)
 
 if TYPE_CHECKING:
     from pathlib import Path
 
-    from creek.models import CompileTargetKind
+    from creek.models import CompileTargetKind, Fragment
 
 TOOL_NAME = "creek.compile"
+
+# The above-ceiling refusal reason is a fixed string naming no fragment id, no
+# title, no tier, and no count. Echoing the offending ids would turn a single
+# call into a bulk tier-classification oracle — submit 100 ids, learn 100 tiers
+# in one round trip. A count would be nearly as useful to an attacker: it makes
+# group testing *exact*, so a caller could binary-search a batch knowing
+# precisely how many above-ceiling ids each half holds.
+#
+# ACCEPTED RESIDUAL RISK: a bare boolean is still a group-testing oracle — a
+# caller who resubmits subsets of a batch identifies every above-ceiling id in
+# O(k log n) calls. Two things that argument must NOT lean on:
+#
+# - *Id unguessability is not a control here.* Ids are deterministic
+#   (``creek.ingest.base.generate_fragment_id`` hashes source+timestamp+content;
+#   ``generate_child_fragment_id`` hashes ``f"{parent_id}:{level}:{index}"``), so
+#   from a single admitted parent id a caller can enumerate every child id with
+#   no knowledge of the content. The id space around known content IS sweepable;
+#   only a blind sweep of the full 48-bit space is infeasible.
+# - *Probe cost does NOT leak where the offending fragment sits.* Worth stating
+#   because the symmetry is accidental and a refactor could destroy it:
+#   ``iter_vault_fragments`` materialises the whole directory into a list before
+#   returning, so the rglob + parse cost is paid in full before
+#   ``_sources_above_ceiling``'s loop begins. The early ``return True`` therefore
+#   short-circuits only in-memory iteration, not the I/O. Unlike
+#   ``reflect.py::_resolve_entry``, which walks a raw lazy ``rglob`` and returns
+#   at the match — a genuine timing channel that comment correctly records. If
+#   this gate is ever switched to a lazy loader (see the load-once follow-up),
+#   that channel appears here too and must be re-analysed.
+#
+# Accepted because the real control is the audit trail, not id entropy: every
+# True probe appends a ``creek.compile`` entry distinguishable from a success
+# (no ``created_path``, no ``affected_fragment_ids``), so a sweep is detectable
+# by rate. Noting the attacker's counter honestly — appending one known-missing
+# id makes the negative half fall through to the engine's not-found refusal,
+# which is NOT audited, so only positive probes are logged. Accepted also
+# because *some* distinguishable refusal is required for a legitimate client's
+# bug to be debuggable at all.
+_ABOVE_CEILING_REASON: Final[str] = "source fragments exceed tier ceiling"
 
 
 class CompileLLMFactory(Protocol):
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
## Summary
PR #934 closes issue #848: `creek_mcp/tools/compile.py`'s module docstring claimed a caller "cannot create a compiled page whose source fragments include a tier they could not read," but `compile_tool` never checked source-fragment tiers against the caller's ceiling. With `ceiling=open`, a caller could compile INTIMATE fragments into a page whose provenance, paradox log entry, and compile-LLM prompt all carried the intimate ids/titles — the LLM prompt path is not tier-routed. This PR adds the missing gate (`_sources_above_ceiling`), wires it in ahead of every egress point, and adds 11 new tests plus a docs update.

## Strengths
- **Shared-loader parity by construction**: the new gate walks the vault via the exact same `iter_vault_fragments` call `compile.engine._load_fragments_for_compile` uses, so gate and engine can never disagree about which files are fragments — verified directly in `creek/compile/engine.py:480` and `creek/vault/reader.py:77`.
- **Correct gate ordering**, and it's justified rather than asserted: cheap argument validation (`target_kind`, empty `fragment_ids`) runs before the vault-walking gate (cost asymmetry — no full `01-Fragments` parse on a malformed call), and the gate itself runs before `llm_factory()` is evaluated, before `compile_to_vault` writes anything, and before the paradox log / hash marker are touched. It also wins over the engine's missing-id error, closing an id-existence oracle.
- **Fail-closed tier resolution** (`_tier_of`): a fragment with no `privacy_tier` key is treated as `INTIMATE` rather than inheriting the `Fragment` model's `unclassified` default (rank 0, admitted everywhere) — this correctly mirrors `reflect.py::_fragment_tier` (confirmed at `creek_mcp/tools/reflect.py:262`), and a test (`test_compile_admits_unclassified_source_at_open_ceiling`) pins that an *explicit* `unclassified` tier stays admitted, so the two cases aren't conflated.
- **Refusal audited without leaking**: the audit entry omits `target_title`, `affected_fragment_ids`, `created_path`, `created_tier`, and collapses `fragment_ids` to `{"count": N}` via the existing `summarise_args` — verified against `creek_mcp/audit.py:91`. A test asserts the raw `mcp.jsonl` bytes never contain the fragment id, a synthesized title string, or the word "intimate", and that the hash chain still verifies after a refusal append.
- **Load-bearing egress test**: `test_compile_refusal_never_invokes_the_llm_factory` uses a call-counting factory rather than trusting `status == "refused"` alone — correctly avoiding the trap where a factory that merely raises would produce a refusal with no gate at all (the pre-fix code evaluates `llm_factory()` inside the existing `try/except (ValueError, RuntimeError)`).
- **Reuses `write_tier_allowed`** rather than re-implementing tier comparison, keeping all write-tools' gates on one predicate (verified in `creek_mcp/tier_ceiling.py:88`), and deliberately has no `ALL`-ceiling fast path to avoid a second route desyncing later.
- Follow-up work that's explicitly out of scope (LLM prompt not tier-routed post-admission, double vault walk, `structural_path` ancestry not checked) is filed as separate issues (#928–#933) rather than silently deferred.

## Security Concerns
🟢 LOW — This *is* the security fix; I found no new vulnerability introduced by it.
- The PR's own "ACCEPTED RESIDUAL RISK" comment (lines 76–104 of the diff) correctly documents that a caller who resubmits subsets of a batch can still binary-search which above-ceiling ids exist in O(k log n) calls, and correctly argues the mitigating control is the audit trail (every positive probe is logged) rather than id-unguessability. This residual risk is inherent to any boolean admission gate and is reasonably accepted here, not ignored.
- The documented TOCTOU window (gate walks the vault, returns, then the engine walks it again — a concurrent `creek classify` could raise a fragment's tier in between) is real but narrow, not caller-controlled, and tracked as a follow-up (#930) requiring an engine signature change to fix properly. Acceptable to defer.

## Problems
None found. I traced the modified `compile_tool` function end-to-end against `tier_ceiling.py`, `vault/reader.py`, `compile/engine.py`, and `audit.py` and every behavioral claim in the code and PR description holds up against the actual source.

## Code Quality
- The inline comment block explaining gate placement (~48 lines within `compile_tool`) and the `_ABOVE_CEILING_REASON` residual-risk comment (~35 lines) are unusually long. This matches the established documentation style already present in this codebase (e.g. `reflect.py`, `tier_ceiling.py` module docstrings), so it's a style observation rather than a deviation — not blocking.
- Minor: `created_tier=None` is passed explicitly on the success path (pre-existing pattern, unchanged by this PR) — fine as-is since compiled pages don't currently carry a single tier.

## CLAUDE.md Adherence
✅ TDD ordering matches the PR's own account (9 of 11 new tests asserted to fail RED against the unfixed code).
✅ Reuses existing write-gate machinery (`write_tier_allowed`, `MCPAuditLog`) instead of duplicating tier logic — consistent with `save.py`/`journal.py` precedent this PR explicitly follows.
✅ mypy-strict-compatible typing throughout the new code (`_tier_of`, `_sources_above_ceiling` fully annotated; `Fragment` import correctly gated under `TYPE_CHECKING` since `from __future__ import annotations` is in effect).
⚠️ I was unable to execute `./scripts/check-all.sh` / `./scripts/test.sh` in this review sandbox (all script and `pytest` invocations were blocked from running), so coverage/lint/mypy numbers claimed in the PR description are not independently re-verified by me — only static-traced. CI on the PR was still `IN_PROGRESS` (not failed) for the 3.11/3.12/3.13 Code Quality & Testing jobs at the time of this review; `CrawDad Quality` jobs are green.

## Testing
Strong. The 11 new tests in `tests/test_mcp_write_tools.py` cover: the acceptance case (intimate source refused at `ceiling=open`), the LLM-factory-never-invoked egress assertion, absence of all on-disk artifacts (page/paradox log/hash marker) on refusal using a stub LLM that *would* produce both, audit-log redaction plus hash-chain integrity, position-independence of the above-ceiling scan, the full tier×ceiling admission matrix, the fail-closed missing-`privacy_tier`-key case across all four ceilings, the explicit-`unclassified`-stays-admitted policy pin, missing-id-still-reported behavior, ceiling-wins-over-missing-id ordering, and unknown-`target_kind`-wins-over-the-gate ordering. All referenced test helpers (`_write_fragment`, `_noop_llm_factory`, `_read_audit`) exist in the file. The four pre-existing compile tests are untouched.

## Documentation
`docs/mcp.md` is updated to describe compile's source-side (not target-side) tier gate in both the tools table and the write-side tier-ceiling section. The module docstring in `compile.py` and the new functions' docstrings are thorough and accurately describe the implementation as verified above.



## Verdict: LGTM

The fix closes a real vulnerability (intimate-tier fragments could be compiled and leaked into open-tier pages, provenance, the paradox log, and an untier-routed LLM prompt) with a correct, minimal implementation that I traced end-to-end against the actual source (tier_ceiling.py, vault/reader.py, compile/engine.py, audit.py) — every claim in the PR body checks out. Gate ordering, audit-log field redaction, fail-closed handling of missing `privacy_tier`, and shared-loader parity with the engine are all verified correct, and the 11 new tests cover the acceptance case, the load-bearing LLM-non-invocation assertion, no-trace-on-refusal, audit redaction, ordering vs. missing-id, and the full tier×ceiling matrix. Nothing here blocks merge; the only remark (very long inline comments) matches this codebase's established documentation style and isn't a deviation worth blocking on.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5075856258 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #876, opened 2026-07-22
fix(classify): PrivacyClassifier is never invoked — every fragment stays privacy_tier: unclassified, which filters treat as OPEN


Geoffe-Ga:
## Role

You are a security-minded pipeline engineer working in `creek-tools/` (Python, mypy strict, TDD per `creek-tools/CLAUDE.md`).

## Goal

Wire `creek.classify.privacy.PrivacyClassifier` into the classification pipeline so every fragment receives a real `privacy_tier` (`open`/`personal`/`intimate`), and stop treating `unclassified` as effectively-open in downstream filters.

## Context

- `PrivacyClassifier` (creek/classify/privacy.py) is fully implemented and tested — but has **zero production callers**. Only `tests/test_privacy.py` instantiates it. Neither `creek classify`, `creek process`, nor any pipeline stage assigns a tier.
- The LLM classification parser deliberately blocks `privacy_tier` from LLM output (`_ALLOWED_TOP_LEVEL_KEYS` in creek/classify/llm/parsing.py, SEC-004) — that guard is correct and must stay. The tier must come from the deterministic classifier.
- Evidence (demo vault `creek-demo-2026-06-07`, 35,330 fragments, every creek command run): **100% of fragments carry `privacy_tier: unclassified`** after full `classify --method llm`.
- Consequence: `creek/classify/privacy_filter.py` ranks `UNCLASSIFIED` at 0 — the same rank as `OPEN`, documented as "pass through with full body". So the entire private corpus (30k+ Discord DMs, journals) is draft-eligible, mine-eligible, and voice-proxy-eligible at the *open* tier, and the Intimate-never-cloud model routing (#666, ADR-0003) has nothing to act on. The module docstring promises "when in doubt … PERSONAL"; the runtime reality is fail-open.
- Related open scan findings that assume tiers exist: #846, #848.

## Proposed Solution

1. Add a privacy-tier assignment step inside `run_classify` (creek/classify/classify_engine.py) that runs the deterministic `PrivacyClassifier` for every fragment whose `privacy_tier` is `unclassified` (and on `--force`). It is rules-based (source kind, authorship, recovery keywords, confessional voice) — no LLM, no consent needed.
2. Ordering: run it **after** frequency/voice classification within the same pass, so the confessional-register signal is available, but **before** the per-tier model router selects a provider for any subsequent LLM work on that fragment. Tier must never be *downgraded* automatically once set to `intimate` (fail-closed).
[...]

Issue #634, opened 2026-06-21
feat(voice): real audience-facing classifier (beyond grep), wired into voice scoping


Geoffe-Ga:
**Epic:** #632 (buffs the signal that #1 consumes)

## Problem
The only audience-vs-private signal today is grep/platform: PrivacyClassifier (classify/privacy.py:80-232 — RECOVERY_KEYWORDS substring scan, platform/channel-name token match) is a *privacy/leakage* gate (OPEN ≠ audience-facing); SourceKind.WRITING (models.py:299-312) is set only by the Substack ingester (ingest/substack.py:400) and read nowhere. Grep-only is too lightweight to tell an essay from a casual OPEN Discord post.

## Change
Add a real **audience-facing classifier** that tags each fragment's audience axis (audience-facing | private | mixed), persisted to frontmatter, running in the classify pipeline. Beyond grep: combine robust signals (platform/medium, representativeness, length, titling/sectioning/quotation presence, self-authored long-form) with a seam for an optional LLM-assisted pass (consistent with `classify --method llm`). Wire its output into the #1 fingerprint scoping, replacing the v0 existing-signal heuristic.

**Constraint:** knowledge/retrieval corpus unchanged; adds an attribution axis used for VOICE scoping only.

## Acceptance
- Fragments carry a persisted `audience` axis; classifier is not pure grep (documented heuristic signals + optional LLM seam).
- Runs in `creek classify`, idempotent like other classification.
- #1's fingerprint scoping consumes this axis.
- Tested: a casual OPEN chat is NOT audience-facing; a Substack essay IS. `./scripts/check-all.sh` green.

## Seams
models.py:299-312/623, classify/privacy.py (do not conflate with privacy), classify/classify_engine.py, ingest/*, generate/ai_style/fingerprint.py.

Refs #632

Issue #584, opened 2026-06-07
feat: give 05-Wavelength/Observations/ a producer (⚠ product decision first)


Geoffe-Ga:
> Part of epic #578 · Refs #578 · Depends on #582 (pattern) · SPEC: `git-issues/2026-06-07_SPEC_summary.md`

## Role
Senior Python engineer in `creek-tools`, plus a product decision from the
maintainer before implementation starts.

## Goal
Give `05-Wavelength/Observations/` a producer so it stops being a read-only
dead end. The folder is read by `DecisionContextGatherer` but nothing ever
writes to it.

## ⚠️ Open product decision (resolve first)
Ontology calls Observations "Raw wavelength-related reflections." That phrasing
is ambiguous between two implementations — **pick one before coding:**

1. **User-authored** — Observations are journaling content the user writes.
   Then the gap is a *save* surface: add `creek save --target observation`
   (extend `SaveTarget` in `creek/save/router.py`, mapping to
   `("05-Wavelength", "Observations")`), mirroring the existing `praxis`/
   `paradox` targets. No auto-generation.
2. **Auto-derived** — Observations are generated from wavelength signal (like
   Phase-Maps). Then add a generator + `creek report --type observations`.

Default recommendation: **option 1 (save target)** — "raw reflections" reads as
user input, it's the smaller change, and `DecisionContextGatherer` already
consumes whatever lands there. Confirm with the maintainer on the issue thread
before starting.

## Context
- Consumer that already reads the folder: `DecisionContextGatherer`
  (`creek/generate/decisions.py:783`, `_current_wavelength` /
  `05-Wavelength/Observations`).
- Save router + targets: `creek/save/router.py` (`SaveTarget`, the
  `_TARGET_DIRS` map ~line 47) — the extension point for option 1.
- Report dispatch: `creek/cli.py` `_REPORT_DISPATCH` + MCP `_VALID_TYPES` — the
  extension point for option 2.
- `Observation` model: `creek/models.py:915`.

## Output format
- Option 1: new `SaveTarget.OBSERVATION` + router mapping + CLI `save` wiring +
  tests (round-trip a saved observation that `DecisionContextGatherer` then
  reads). Update `docs/slash-commands.md` save targets.
- Option 2: `ObservationGenerator` + `_report_observations` handler + MCP type
  + integration test. Update `docs/slash-commands.md` report types.

## Examples
```text
# option 1
[...]

PR #936, opened 2026-07-25
fix(classify): give praxis_potential a producer so 04-Praxis and 08-Decisions are reachable

### LINKED ISSUE #877: fix(classify): praxis_potential has no producer — 04-Praxis and 08-Decisions are structurally unreachable
## Role

## Goal

Give `praxis_potential` a producer so the consumers that gate on it — decision detection, praxis-chain mining, and compost's praxis check — can ever fire. Today `04-Praxis/` and `08-Decisions/` are structurally unreachable.

## Context

- `Fragment.praxis_potential` defaults to `PraxisPotential.NONE` (creek/models.py:770) and **no pipeline stage ever changes it**:
  - The LLM classify parser only accepts `frequency`, `wavelength`, `voice`, `confidence_scores` (`_ALLOWED_TOP_LEVEL_KEYS`, creek/classify/llm/parsing.py) — `praxis_potential` is not in the schema or prompt.
  - The rules classifier doesn't set it either. `grep -rn "PraxisPotential\." creek/` shows only consumers.
- Consumers gating on `EXPLICIT`:
  - creek/generate/decisions.py:307 — a fragment is a decision candidate only if `praxis_potential == "explicit"`
  - creek/generate/mining.py:1114 — praxis-chain mining strategy
  - creek/generate/compost.py:119 — praxis check
- Evidence (demo vault `creek-demo-2026-06-07`, 35,330 fragments, full LLM classify done): a 3,000-fragment sample shows `praxis_potential: none` on **100%**. `creek report --type decisions` (run 2026-07-22) → "No decision notes generated: no new decision candidates found." `08-Decisions/` and `04-Praxis/` contain only `.gitkeep`.
- Related closed issues #44/#45 built decision detection on top of this field; the field was never fed. Same folder-without-producer pattern as #584.

## Proposed Solution

Extend the LLM classification response schema with a `praxis_potential` section (`explicit` / `latent` / `none`) and prompt guidance ("does this fragment state an actionable insight, practice, or decision-in-progress?"):

1. Add `praxis` (or extend an existing section) to `CLASSIFICATION_PROMPT` and `_ALLOWED_TOP_LEVEL_KEYS`, with enum-parsing that falls back to `none` (same `_parse_enum` pattern as mode/dosage).
2. Rules-classifier fallback: mark `explicit` on cheap signals (imperative "I should/I will/next time", checkbox lists, "decision:"-style markers) so offline vaults aren't dead.
3. Provide a resume-safe backfill path: `creek classify --method llm` without `--force` currently skips already-classified fragments, so add the field to the "material change / upgrade" detection used by `creek fill` (#736 pattern) — otherwise existing 35k-fragment vaults can never backfill without a full forced reclassify.

## Acceptance Criteria

- [ ] LLM classify sets `praxis_potential` on newly classified fragments; unknown values fall back to `none`
- [ ] Rules classify sets `explicit` on the cheap-signal heuristics, with tests
- [ ] `creek fill` (or classify upgrade offer) detects fragments classified before this field existed and offers a backfill
- [ ] On a fixture vault with an actionable-insight fragment, `creek report --type decisions` writes a draft note to `08-Decisions/Active/`

## Constraints

- Keep SEC-004: `privacy_tier` remains rejected from LLM output; only add the documented new key
- Backfill must be opt-in (it re-runs the LLM over the corpus — cost surfaced to the operator first)
- Follow stay-green TDD

Geoffe-Ga:
## Summary

`Fragment.praxis_potential` shipped with a default of `none` and **zero production writers**. All three consumers gate on `explicit` — `generate/decisions.py:307`, `generate/mining.py:1114`, `generate/compost.py:119` — so decision detection, praxis-chain mining and compost's praxis check were structurally unreachable. 100% of the 35,330-fragment demo vault reads `praxis_potential: none`; `creek report --type decisions` reports "no new decision candidates found"; `04-Praxis/` and `08-Decisions/` hold only `.gitkeep`. Same folder-without-producer shape as #584, and the same shape as #876 — whose `privacy_pass.py` this change deliberately mirrors.

**Policy lives in the new pure `creek/classify/praxis_pass.py`.** A table-driven, precompiled signal set (strong, weight 2: a markdown task checkbox, a line-initial `Decision:`/`Praxis:`/`Next step:`/`Action item:` label; moderate, weight 1: sentence- or line-anchored first-person commitments) plus a monotone `none < latent < explicit` merge.

The threshold is the substance. `_EXPLICIT_AT = 2` means one strong marker or two *distinct* moderate ones, so a bare "I will" never fires. Over-firing would not fix this bug — it would flood mining, compost and the decisions report with noise, which is strictly worse than the current silence. Scoring is presence-based over distinct patterns: a repeated idiom is a verbal tic, not a second piece of evidence. Anchoring is what rejects reported speech ("he said I should… and that I will…") and the second/third-person lookalikes that dominate a chat-log corpus.

**The heuristic never emits `latent`.** "There is a practice hiding in here that the author has not named" is a judgment no regular expression can make, so `latent` reaches a fragment only through the new `praxis:` section in the LLM response schema and prompt.

**`escalate` is the single chokepoint for every producer, and it never lowers.** The LLM's verdict merges against the verdict the fragment *already carries*, not merely against `none`. Refusing to write only on `none` is not enough: `latent` is also weaker than `explicit`, and nothing downstream could repair such a demotion, because the keyword pass re-derives from the body and only ever proposes `explicit` or `none` — so a verdict that came from judgment rather than keywords would keep the demotion to disk. (This was caught at self-review; the regression test asserts on frontmatter read back off disk.)

**Ordering is load-bearing.** The pass runs in `_process_file` *after* `_classify_one` (so the merge already sees any LLM verdict) and *before* the #876 `reassess` (so the privacy pass stays the last mutation before the write). `_prepare_fragment` is byte-identical to `9af26cc`: the tier is still assigned before `TierClassifiers.for_tier` reads it and above the OPS-001 short-circuit. A regression test now pins that, so a future change cannot hoist praxis into that window and slide the privacy pass below the router.

**`_write_tier_only` is deliberately not widened** — a departure from the #876 precedent, not an oversight. That writer's narrow promise ("changing ONLY the privacy-tier fields") is itself load-bearing when reviewing the #876 path, and the justification does not transfer: an untiered fragment is a live cloud-egress hole, whereas a missing praxis verdict is a missing feature, and silently rewriting the praxis axis of every already-curated fragment in a mature vault is not something to do unasked. So an `llm`/`manual`-stamped vault backfills only under an explicit, paid `--force`.

**`creek fill` surfaces the outstanding count and names that command's token cost.** The gap is detected as "the free heuristic proves `explicit` while the disk says `none`" — positive evidence a producer never ran, zero tokens to compute, and self-limiting. "Value == `none`" would have nagged forever (most fragments genuinely are `none`); "key absent" can never fire, because `model_dump` always writes the key. Both fill hints share one vault walk, so `creek fill` still parses the 35k-fragment vault once.

**Security.** SEC-004 is intact: the allow-list gains exactly `praxis`, a *section* name and not a `Fragment` field name, and `privacy_tier` stays rejected. `_YAML_HEAD_RE` is deliberately not widened. Every signal pattern is linear-time (no nested quantifiers) and compiled once at import, because it runs over unbounded, attacker-influenced bodies. The fill hint reports a **count only** — never a fragment id or body excerpt (config-oracle rule, #846/#848).

`ClassifySummary.praxis_marked` and the `creek.classify` MCP payload report what each run raised. A silent producer is how this bug survived.

### Known limitations (documented in `praxis_pass.py`)

- Fragments the resume short-circuit preserves get no praxis verdict without `--force`; `creek fill` reports the count.
- `creek/classify/weighted.py` carries its own prompt and allow-list, so on the weighted path praxis comes from the heuristic only and the model can never contribute `latent`.

### Follow-up observed, not fixed here

`creek/pipeline.py::_classify_fragments` never calls `AudienceClassifier` — the identical shape of hole for the #634 audience axis. `creek classify` stamps it; `creek process` does not. Out of scope for #877.

## Test plan

- `./scripts/check-all.sh` → **exit 0**, 12/12 checks. mypy strict clean, ruff clean, xenon clean, pylint 9.70, refurb + tryceratops clean, aggregate coverage **93.93%**, per-file gate 217 files >= 80%.
- Full suite including the `integration` and `e2e` markers: **6326 passed, 5 skipped** (the 5 are live-API / Ollama smoke tests with no credentials locally).
- Baseline on the base commit `9af26cc` was independently confirmed exit 0 with 0 failures before any edit, so every failure seen during this change was real.
- New `tests/test_praxis_pass.py` — each signal in isolation; **one moderate hit alone stays `none`** (the anti-noise contract); a repeated single idiom stays `none` (presence-based scoring); near-misses (`you should`, `it will`, `they will`) do not fire; `escalate` never lowers across the full matrix; `apply_praxis` returns the *same object* when nothing escalates; idempotence; empty body; the stored value is a plain `str`.
- New `tests/test_classify_praxis_e2e.py` (integration) — rules classifier only, no LLM, no network: a seeded fragment goes `creek classify --method rules` → `praxis_potential: explicit` on disk → `creek report --type decisions` → exactly one note under `08-Decisions/Active/` with `detection_method: pattern`. The fixture title deliberately contains **no** `DECISION_KEYWORDS`, so the keyword strategy cannot make the test pass without the new field.
- `tests/test_classify_engine.py` — a rules run stamps `explicit`; a neutral fragment stays `none`; an already-`explicit` fragment is never demoted; `summary.praxis_marked` counts only raises; and the **#876 ordering regression test** (tier assigned before the router; preserved fragments still tier-only-written, praxis untouched without `--force`).
- `tests/test_classify.py` — `validate_response` accepts `praxis` and **still rejects `privacy_tier`**; `_apply_praxis` escalate-only matrix; `latent` over `explicit` writes nothing, end to end through `classify()`; the prompt renders through `str.format` without a `KeyError`.
- `tests/test_pipeline.py` — `creek process` stamps praxis on both the no-LLM and the LLM path, and an LLM `latent` answer never demotes an `explicit` fragment on disk.
- `tests/test_cli_fill.py` — the backfill count fires only on provable gaps, is zero without `01-Fragments`, names the `--force` cost, is **silent when there is nothing to backfill**, and an unreadable fragment warns without aborting `fill`.

Closes #877

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -57,6 +57,42 @@ call it:
   non-``unclassified`` tier on disk also survives any non-``--force``
   run untouched: the operator's decision outranks the heuristic.
 
+Praxis pass (issue #877)
+------------------------
+
+Every fragment this run actually (re-)classifies also leaves the engine
+carrying a scored :class:`~creek.models.PraxisPotential`. Before #877 the
+field had a default and **no producer at all**, so the three consumers
+gated on ``explicit`` (:mod:`creek.generate.decisions`,
+:mod:`creek.generate.mining`, :mod:`creek.generate.compost`) were
+structurally unreachable. The policy lives in
+:mod:`creek.classify.praxis_pass`; the engine only decides *when*:
+
+* **After the classifier, before the privacy reassess.** Running after
+  :func:`_classify_one` means the escalate-only merge already sees any
+  LLM verdict, and running before :func:`~creek.classify.privacy_pass.reassess`
+  keeps the #876 privacy reassess as the last mutation before the write.
+  It must NOT be hoisted into :func:`_prepare_fragment`: that would put
+  it between the tier assignment and
+  :meth:`TierClassifiers.for_tier`, where the #876 ordering above is
+  load-bearing security.
+* **NOT on the preserved short-circuit** — a deliberate departure from
+  the tier precedent one section up. :func:`_write_tier_only` makes a
+  narrow, auditable promise ("changing ONLY the privacy-tier fields")
+  that is itself load-bearing when reviewing the #876 path, and the
+  justification for the tier exception does not transfer: an untiered
+  fragment is a live cloud-egress hole, whereas a missing praxis verdict
+  is a missing feature. So an ``llm``/``manual``-stamped vault backfills
+  this axis only under an explicit ``creek classify --method llm
+  --force``. ``creek fill`` surfaces the outstanding count and names that
+  command's token cost (:func:`creek.cli._hint_praxis_backfill`).
+* **Escalate-only.** A rules re-run can raise ``none`` to ``explicit``
+  but can never demote a verdict the LLM or the operator put on record.
+
+:attr:`ClassifySummary.praxis_marked` reports how many fragments the run
+actually raised — the observability that was missing when this field had
+an invisible (in fact absent) producer.
+
 Known residual (not fixable by ordering)
 ----------------------------------------
 
@@ -104,6 +140,7 @@ from creek.classify.constants import (
 )
 from creek.classify.llm import LLMClassifier
 from creek.classify.llm.router import IntimateRoutingError
+from creek.classify.praxis_pass import apply_praxis
 from creek.classify.privacy import PrivacyClassifier
 from creek.classify.privacy_filter import tier_of
 from creek.classify.privacy_pass import (
@@ -150,6 +187,9 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Iterator
 
     from creek.config import ClassificationConfig, CreekConfig, LLMConfig
+
+    # Annotation-only: the praxis verdict is compared, never constructed here.
+    from creek.models import PraxisPotential
 
 logger = logging.getLogger(__name__)
 
@@ -233,6 +273,16 @@ class ClassifySummary:
             This counts the decision, not the bytes: a fragment whose
             subsequent write fails is counted here *and* reported on
             :attr:`errors`.
+        praxis_marked: Fragments this run raised to a stronger
+            :class:`~creek.models.PraxisPotential` (issue #877) — almost
+            always ``none`` → ``explicit`` from the keyword pass. The
+            pass is escalate-only and idempotent, so a second run over an
+            unchanged vault reports ``0``. Fragments the resume
+            short-circuit preserved are **not** counted: they are not
+            given a praxis verdict at all without ``--force`` (see the
+            module docstring). Like :attr:`privacy_tiers_assigned` this
+            counts the decision, not the bytes — a fragment whose write
+            then fails is counted here *and* reported on :attr:`errors`.
     """
 
     total: int
@@ -241,10 +291,11 @@ class ClassifySummary:
     preserved_llm: int
     skipped_high_confidence: int
     errors: tuple[str, ...]
-    # Appended last, with a default, so the positional
+    # Appended last, with defaults, so the positional
     # ``ClassifySummary(0, 0, 0, 0, 0, ())`` construction below still
     # compiles unchanged.
     privacy_tiers_assigned: int = 0
+    praxis_marked: int = 0
 
 
 @dataclass
@@ -261,6 +312,7 @@ class _RunCounts:
     preserved_llm: int = 0
     skipped: int = 0
     privacy_tiers_assigned: int = 0
+    praxis_marked: int = 0
     errors: list[str] = field(default_factory=list)
 
 
@@ -459,6 +511,7 @@ def run_classify(
         # underlying list and reach into completed-run state.
         errors=tuple(counts.errors),
         privacy_tiers_assigned=counts.privacy_tiers_assigned,
+        praxis_marked=counts.praxis_marked,
     )
 
 
@@ -563,6 +616,29 @@ def _record_if_preserved(
         counts.preserved_llm += 1
         return True
     return False
+
+
+def _record_praxis(
+    before: PraxisPotential,
+    fragment: Fragment,
+    counts: _RunCounts,
+) -> None:
+    """Tally a fragment whose praxis verdict this run raised (issue #877).
+
+    Kept as its own function so the ``if`` lives here rather than in
+    :func:`_process_file`, whose cyclomatic complexity is already at its
+    working ceiling. Comparing before/after is safe precisely because
+    :func:`~creek.classify.praxis_pass.apply_praxis` is escalate-only:
+    the values can only ever differ upwards, so this counts real work and
+    never churn.
+
+    Args:
+        before: The verdict the fragment carried before the praxis pass.
+        fragment: The fragment after the pass.
+        counts: Mutable per-run counters; mutated in place.
+    """
+    if fragment.praxis_potential != before:
+        counts.praxis_marked += 1
 
 
 def _assert_classifiers_available(tier_classifiers: TierClassifiers | None) -> None:
@@ -825,6 +901,16 @@ def _process_file(
     # so it runs on both the rules and LLM paths and stays idempotent.
     new_fragment = audience.classify_and_enforce(new_fragment, body)
 
+    # #877: score the praxis axis from the free keyword pass. Unconditional
+    # by design — the purity of ``apply_praxis`` absorbs the decision (it
+    # returns the same object when nothing escalates), so this adds no
+    # branch to this function. It runs AFTER ``_classify_one`` so the
+    # escalate-only merge already sees any LLM verdict, and BEFORE the
+    # reassess below so the #876 privacy pass stays the last mutation
+    # before the write.
+    praxis_before = new_fragment.praxis_potential
+    new_fragment = apply_praxis(new_fragment, body)
+
     # #876: classification may have hardened the fragment's voice signals
     # (confessional + conviction is one of the INTIMATE triggers), which the
     # pre-pass could not see. Take the stricter of the two verdicts —
@@ -834,6 +920,8 @@ def _process_file(
         new_fragment = reassess(new_fragment, body, classifier=privacy)
 
     with lock:
+        _record_praxis(praxis_before, new_fragment, counts)
+
         # FEAT-023 wire-up (issue #318): when ``classification.reatomize`` is
         # True we run the orchestrator on the root fragment and persist any
         # newly-derived child fragments. The root's frontmatter still flows
--- a/creek-tools/creek/classify/llm/orchestrator.py
+++ b/creek-tools/creek/classify/llm/orchestrator.py
@@ -20,6 +20,7 @@ from creek.classify.llm.batch import run_batch
 from creek.classify.llm.calibration import _apply_wavelength
 from creek.classify.llm.parsing import (
     _apply_frequency,
+    _apply_praxis,
     _apply_voice,
     _split_reasoning_and_yaml,
     validate_response,
@@ -33,6 +34,7 @@ from creek.classify.llm.providers import (
     provider_display_name,
     provider_is_cloud,
 )
+from creek.models import PraxisPotential
 
 if TYPE_CHECKING:
     from creek.classify.llm.base import LLMProvider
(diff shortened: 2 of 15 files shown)


Geoffe-Ga:
## Summary
Fixes #877: `Fragment.praxis_potential` had a default of `none` and no producer anywhere in the pipeline, so the three consumers gated on `explicit` (`decisions.py`, `mining.py`, `compost.py`) were structurally unreachable. The PR adds a pure, side-effect-free `creek/classify/praxis_pass.py` (a precision-first, table-driven keyword heuristic plus an escalate-only merge), wires it into both `creek classify` (`classify_engine.py`) and `creek process` (`pipeline.py`), extends the LLM response schema with an optional `praxis:` section that can additionally emit `latent`, and adds a `creek fill` backfill hint.

## Strengths
- **Root-cause fix, not a patch.** The policy lives in one pure module (`praxis_pass.py`) reused by both the classify engine and the one-shot pipeline, avoiding the "fixed classify but not process" half-fix the PR explicitly calls out as a known follow-up for a *different* axis (#634 audience).
- **Escalate-only merge is the correct invariant and is enforced at a single chokepoint** (`escalate`), with both the rules heuristic and the LLM path routed through it. The PR description notes this exact demotion bug (LLM `latent` answer silently lowering a recorded `explicit`) was caught at self-review and is now pinned by `test_a_latent_answer_leaves_an_explicit_fragment_explicit` and `test_llm_answer_never_demotes_an_explicit_fragment_on_disk` — a genuinely good regression test targeting an end-to-end (not just unit) path.
- **Precision-over-recall design is well justified and tested.** `_EXPLICIT_AT = 2` (one strong marker or two *distinct* moderate ones), presence-based (not occurrence-based) scoring, and sentence/line anchoring to reject second/third-person lookalikes and reported speech are all backed by targeted tests (`test_one_idiom_repeated_is_still_a_single_signal`, `test_moderate_marker_must_be_anchored_to_a_sentence_start`, near-miss parametrization).
- **Security invariants preserved deliberately, not accidentally.** SEC-004's `_ALLOWED_TOP_LEVEL_KEYS` is widened by exactly one key (`praxis`), with an explicit test that `privacy_tier` still gets rejected even alongside a valid `praxis` section. `_YAML_HEAD_RE` is deliberately *not* widened, with a documented rationale (schema lists `praxis` last, so it can never lead a payload). All new regex patterns are linear-time and compiled once at import — relevant given they run over unbounded, attacker-influenced fragment bodies at 35k-fragment vault scale.
- **#876 ordering is protected with a dedicated regression test** (`test_praxis_pass_runs_after_the_router_reads_the_tier`) guarding against a future refactor hoisting the praxis pass into `_prepare_fragment` and sliding the privacy reassess below the tier router — a real, previously-load-bearing security ordering.
- **The `--force`-only backfill decision for preserved fragments is well-reasoned** and explicitly contrasted with the #876 tier precedent (untiered = live cloud-egress hole vs. missing praxis = missing feature), with tests proving the asymmetry (`test_preserved_fragment_gets_praxis_only_under_force`).
- **The `creek fill` hint reports a count only**, with tests explicitly asserting no fragment id or body excerpt leaks into stdout (config-oracle rule, #846/#848), and the shared single-walk optimization (`_scan_fill_gaps`) avoids parsing the vault twice.
- **The e2e test is designed to avoid a false-green.** `test_classify_then_report_produces_a_pattern_detected_decision` deliberately picks a fixture title containing none of `DECISION_KEYWORDS`, so the test can only pass if the *pattern* strategy (which needs `praxis_potential`) actually fired — not the pre-existing keyword strategy.

## Security Concerns
🟢 LOW — No new vulnerabilities identified. The one deliberately-not-widened surface (`_YAML_HEAD_RE`) and the one deliberately-narrow allow-list addition (`praxis` only, `privacy_tier` still rejected) are both correctly reasoned and tested. Regex patterns are ReDoS-safe (no nested quantifiers, all precompiled).

## Problems
None identified that would block merge.

## Code Quality
- The module and function docstrings are extremely long (in places verging on essay-length, e.g. `praxis_pass.py`'s module docstring and several individual function docstrings). This matches the project's established `interrogate` ≥95% docstring-coverage requirement and the precedent set by the #876 privacy-pass module, but a future reviewer may want to consider whether some of this rationale belongs in `docs/decisions.md` (which is also updated here) rather than duplicated across module docstring, function docstrings, and test-file docstrings. Non-blocking — consistent with existing repo conventions.
- `_record_praxis`'s parameter is typed as `PraxisPotential` but the runtime value is a plain `str` (due to `Fragment.model_config`'s `use_enum_values=True`); this mismatch is pre-existing pattern in the codebase (documented elsewhere via `.value` comments) and not introduced by this PR.

## CLAUDE.md Adherence
✅ TDD: tests were clearly written alongside (or before, per the PR's stated regression discovered "at self-review") the implementation, covering pure-function policy, engine integration, pipeline integration, MCP surface, and CLI hints.
✅ Coverage: PR reports 93.93% aggregate, per-file gate 217 files ≥80% (per PR description; CI still finishing at review time).
✅ Docstring coverage: consistent with ≥95% interrogate threshold given the docstring density throughout.
✅ Complexity: new logic is factored into small pure functions (`detect`, `escalate`, `apply_praxis`, `_apply_praxis`, `_record_praxis`) specifically to avoid raising `_process_file`'s cyclomatic complexity, per the `_record_praxis` docstring's stated rationale.
✅ MyPy strict / Ruff: no obvious violations spotted (proper `TYPE_CHECKING` guards, `from __future__ import annotations` present in touched modules).
✅ Commit/PR conventions: Conventional Commits title (`fix(classify): ...`), issue-linked, follows the git-workflow issue-first pattern.

## Testing
Testing is comprehensive and well-targeted, not just high-volume:
- Pure policy layer (`test_praxis_pass.py`): full signal-table coverage, anchoring edge cases, near-miss rejection, escalation matrix (all 9 combinations plus a symmetry property test), idempotence, and an explicit "never emits `latent`" invariant test parametrized across every fixture body.
- Engine integration (`test_classify_engine.py`): stamping, idempotence (`praxis_marked == 0` on second run), preserved-fragment `--force` asymmetry, and the #876 ordering regression.
- LLM parsing (`test_classify.py`): schema acceptance, SEC-004 regression, escalate-only merge at both the unit (`_apply_praxis`) and full-call (`classifier.classify`) levels — notably including the demotion regression test that started from a non-default (`explicit`) fragment, which the PR description says earlier unit tests structurally could not catch.
- Pipeline integration (`test_pipeline.py`): no-LLM heuristic-only path, escalate-only merge under `no_llm=True`, and the full LLM-dispatch demotion-prevention path with real vault writes.
- CLI hint (`test_cli_fill.py`): provable-gap-only counting (4-way fixture matrix), zero-signal silence (no permanent nag), disclosure-safety assertions, and best-effort failure isolation from the sibling #876 hint.
- True e2e (`test_classify_praxis_e2e.py`): proves `08-Decisions/` reachability through the shipped CLI via the *pattern* strategy specifically, guarding against a false-green through the pre-existing keyword strategy.

## Documentation
`docs/decisions.md` is updated to explain the pattern-strategy's now-satisfied `praxis_potential` input and the backfill story. Module and function docstrings extensively document the rationale, known residuals (preserved-fragment backfill needs `--force`; the weighted path is heuristic-only), and forward pointers to related issues (#876, #634, #846/#848). Prompt changes (`prompts.py`) are documented with an explicit warning about `str.format` brace-doubling for future editors — a nice defensive touch given a real class of prompt-template bugs.



## Verdict: LGTM

The change is correct, narrowly scoped to the stated bug, exhaustively tested (unit, integration, e2e, and a specific regression test for the exact demotion defect the author caught at self-review), and preserves the SEC-004 injection boundary and #876 ordering invariants with a dedicated regression test. Nothing here would block merge; CI (3.11/3.12/3.13 quality jobs) was still in progress at review time and should be confirmed green before merge, but the diff itself gives no reason to expect a failure.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5076915863 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #946, opened 2026-07-28
[hardening] redact --apply token-boundary snapping: prose over-redaction, O(S x R) cost, and allowlist docs


Geoffe-Ga:
Follow-ups from the security audit of the #909 fix. None is a leak; all three were rated MINOR with no blockers or majors. Recording them rather than expanding a P1 security PR.

Background: #909 added token-boundary snapping to `Redactor._snap_to_candidate_runs` — any match span whose start or end falls strictly inside a contiguous `HIGH_ENTROPY_CANDIDATE` run (`[A-Za-z0-9+/=_-]{20,}`) is widened to that run edge, so a regex match can never bisect a token and leave half a secret in cleartext. That is deliberately fail-closed and independent of `min_confidence`.

## 1. Snapping can swallow hyphen/slash-joined prose

The candidate class includes `-`, `_`, `/`, `+`, `=` — exactly the joiners used by slugs, kebab-case identifiers, and paths. So a self-delimiting PII match can drag a whole slug into the marker:

```python
redactor.redact_content("order-555-123-4567-confirmation")
#  before #909: "order-[REDACTED:phone_number]-confirmation"
#  after  #909: "[REDACTED:phone_number]"
```

Same shape for `ssn` inside a path (`path/123-45-6789/more` is one 21-char run). JSON/CSV are largely protected because `"`, `:`, `,` break runs; markdown prose is not. `--apply` writes in place, so this is only reversible via the queue.

Proposed remedy from the audit: restrict snapping to token-shaped credential detectors and exempt the patterns whose matches are inherently self-delimiting — `phone_number`, `ssn`, `ipv4`, `ipv6`, `email` — since those cannot leave "the rest of the secret" behind and so gain nothing from widening.

**This needs an explicit architectural ruling before implementation.** Exempting selected detectors from the span union is structurally the same move that left the high-entropy detector outside the #832/#902 union and thereby created #909. Any carve-out must come with a test proving the exempted pattern cannot bisect a token.

## 2. `_snap_to_candidate_runs` is O(spans x runs)

Both terms grow linearly in file size on token-dense input. `.json` is in `supported_extensions` and files are read whole, so a ChatGPT `conversations.json` full of 36-char UUIDs (each a run at ~3.9 bits/char) makes spans ~= runs:

| runs | file size | inner-loop iterations |
|---|---|---|
| 20k | ~2 MB | 4e8 (~30 s) |
[...]

Issue #945, opened 2026-07-28
[bug] redact --apply rewrites its own markers into nested markers at low min_confidence


Geoffe-Ga:
Found while fixing #909; deliberately out of scope there.

## Symptom

`Redactor.redact_content` is not idempotent at low `min_confidence`. Reproduced at the #909 branch HEAD:

```python
out(  "[REDACTED:email_password_combo]", min_confidence=0.0)
# -> "[REDACTED:[REDACTED:high_entropy_string]]"
```

Re-running `creek redact --apply` over an already-redacted tree with `min_confidence <= 0.59` corrupts previously-written markers into nested markers.

## Cause

The marker name `email_password_combo` is exactly 20 characters, which is the `HIGH_ENTROPY_CANDIDATE` floor (`[A-Za-z0-9+/=_-]{20,}`), so the marker body is itself a candidate run. Its Shannon entropy is **3.6842 bits/char**, which sits just **0.0158 bits** under the default 3.7 threshold — that margin is the only reason this is invisible at default settings. Lower `min_confidence` and the threshold drops to 2.5, the marker body clears it, and the entropy detector redacts the marker.

`email_password_combo` is currently the only name in `PATTERN_METADATA` that reaches the 20-char floor (`openai_project_key` is 18, `high_entropy_string` is 19), so today it is the single load-bearing case — but any future pattern name of 20+ chars inherits the bug.

## Not a leak

No secret is exposed; the damage is marker corruption and loss of idempotence. Filing as P2 rather than P1.

## Suggested fix

Make emitted markers structurally inert to the detectors — e.g. exclude regions already occupied by a spliced marker from entropy candidacy, or guarantee marker syntax outside the candidate character class. A regression test exists in spirit at `tests/test_redact.py::TestHighEntropyOverlapLeak::test_markers_are_inert_for_every_pattern_name`, which currently pins inertness only at the default threshold and notes this issue as the reason it is not parametrized over `min_confidence=0.0`.

Refs #909

Issue #435, opened 2026-06-01
bug(redact): property test found a rare case where redaction leaves part of an email-like string


Geoffe-Ga:
## Context
During FEAT-040 work, `tests/test_properties_redaction.py::test_redacted_marker_is_present_when_secret_matches` failed on a Hypothesis example (reproduced on clean `main` — **pre-existing, unrelated to FEAT-040**). It did not recur after clearing the local `.hypothesis` cache, so it is a rare example the generator occasionally hits (hence fresh CI has been green).

## Observed
`'user@example.com' not in '<redacted output>'` failed; output looked like `[REDACTED:high_entropy_string]...example.com0` — the redactor replaced part of an email-like high-entropy token but left a trailing fragment intact, so the original substring survived.

## Why needs-spec
The exact failing input was lost when the cache was cleared. A maintainer should reproduce (higher `max_examples` or a pinned `--hypothesis-seed`), pin the example, and decide the correct redaction boundary at email/high-entropy overlaps.

Surfaced from the autonomous FEAT-040 loop rather than silently cleared.

Geoffe-Ga:
## Spec / RCA (reproduced deterministically)

**The framing in the original report ("email/high-entropy overlap") is a red herring.** The truncated output `...example.com0` made it look like high-entropy redaction partially ate an email. It didn't — the `high_entropy_string` pattern is `[A-Za-z0-9+/=_\-]{20,}`, which contains neither `@` nor `.`, so it can never span an email, and every email sub-token (`user`, `example`, `com`) is < 20 chars. The marker and the surviving secret are **independent events**.

### Root cause
The property `test_redacted_marker_is_present_when_secret_matches` asserts:
```python
if "[REDACTED:" in redacted:
    assert secret not in redacted
```
This conflates *"some marker appeared"* with *"this secret was redacted."* Two unrelated facts collide:

1. **The email pattern's `\b` anchors fail when the secret is flanked by word-characters not in its charset.** Pattern: `\b[a-zA-Z0-9._%+\-]+@[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}\b`. A preceding Unicode letter (e.g. `é`) or a trailing alphanumeric (e.g. `0`) removes the `\b`, so `user@example.com` is **not** matched → not redacted.
[...]

PR #947, opened 2026-07-28
fix(redact): stop leaking a high-entropy secret remainder on apply

### LINKED ISSUE #909: [scan:bugs] redact --apply leaks high-entropy secret remainder when a regex pattern partially overlaps it
## Role
suppressions).

## Goal
Make `Redactor.redact_content` redact a high-entropy secret in full even when a
regular pattern covers only part of it, restoring the #832 scan/apply parity
invariant — verified by a test that a high-entropy blob partially overlapped by
a regex pattern leaves **no** cleartext remainder after `--apply`.

## Context
- File(s): `creek-tools/creek/redact/redactor.py:228-234` (`redact_content`) and
  `creek-tools/creek/redact/redactor.py:300-321` (`_replace_high_entropy`).
- Evidence: `redact_content` splices the regular-pattern markers first and only
  then runs the high-entropy detector over the already-mutated text:
  ```python
  spans = self._collect_spans(content, patterns_to_use)
  content = self._splice_markers(content, _merge_spans(spans))   # 229
  if self._should_apply_high_entropy(pattern_types):
      content = self._replace_high_entropy(content)               # 232 — runs on spliced text
  ```
  Unlike the #832 fix, the high-entropy detector is **not** collected as a span
  against the ORIGINAL text and merged; it is a separate post-pass. So when a
  regular pattern (e.g. an `api_key` match of exactly 20 chars) covers only the
  PREFIX of a longer high-entropy blob, the marker replaces the prefix and the
  trailing remainder is `< 20` chars, so the `HIGH_ENTROPY_CANDIDATE` (`{20,}`)
  regex can no longer match it. Meanwhile `RedactionScanner._scan_high_entropy`
  DID flag the full blob, so a `--scan` finding survives `--apply`.
  Live-reproduced: input `AKIAIOSFODNN7EXAMPLEZq7Wp3Rf9Tk2Ng` →
  `scan_file` reports `{'api_key', 'high_entropy_string'}`, but
  `redact_content(...)` returns `... [REDACTED:api_key]Zq7Wp3Rf9Tk2Ng` — the
  14-char remainder `Zq7Wp3Rf9Tk2Ng` is left in cleartext. This is the exact CLI
  `--apply` path (`_apply_redactions` calls `redact_content` with
  `pattern_types=None`).
- Related: regression of the #832 scan/apply parity invariant; sibling
  `[scan:bugs]` findings from this run (purge RTBF silent-skip, generic-ingestor
  duplication, attribution always-escalate). Deferred sibling: purge
  UnicodeDecodeError crash in `_scrub_one_file`.

## Output Format
"Closes #N".

## Examples
```python
def test_high_entropy_remainder_survives_apply():
    cfg = RedactionConfig()
    scanner = RedactionScanner(cfg)
    secret = "AKIAIOSFODNN7EXAMPLEZq7Wp3Rf9Tk2Ng"
    assert "high_entropy_string" in {m.match_type for m in scanner.scan_content(secret)}
    out = Redactor(cfg, scanner.salt).redact_content(secret)
    assert "Zq7Wp3Rf9Tk2Ng" not in out   # fails today: remainder leaks in cleartext
```
Fix sketch — collect high-entropy candidates as `_Span` entries against the
ORIGINAL content inside (or alongside) `_collect_spans`, subject to the same
allowlist + entropy threshold, and let `_merge_spans` union them with the
regular spans, instead of running `_replace_high_entropy` as a sequential
post-pass on the spliced text:
```python
spans = self._collect_spans(content, patterns_to_use)
if self._should_apply_high_entropy(pattern_types):
    spans.extend(self._collect_high_entropy_spans(content))   # against ORIGINAL
content = self._splice_markers(content, _merge_spans(spans))
```
Then a partial regular-pattern overlap can never leave an unredacted
high-entropy tail.

## Constraints
- Preserve existing behavior when a high-entropy blob does NOT overlap a regular
  pattern (the existing high-entropy tests must still pass), and keep the
  allowlist/entropy-threshold gating intact.

Geoffe-Ga:
## Summary

Fixes a P1 secret leak in `creek redact --apply`.

`Redactor.redact_content` spliced the regex-pattern markers first and only then ran the generic high-entropy detector as a *post-pass over the already-mutated text*. When a regex match covered only part of a longer high-entropy run, the marker replaced that part and the remainder was left shorter than the `HIGH_ENTROPY_CANDIDATE` `{20,}` floor, so the detector could no longer re-find it. `AKIAIOSFODNN7EXAMPLEZq7Wp3Rf9Tk2Ng` redacted to `[REDACTED:api_key]Zq7Wp3Rf9Tk2Ng` — a 14-char cleartext tail of a secret that `--scan` had correctly flagged as `high_entropy_string`. This is the exact CLI `--apply` path: `_apply_redactions` calls `redact_content` with no `pattern_types`.

Two changes, both in `creek/redact/redactor.py`:

1. **The entropy detector joins the span union.** `_replace_high_entropy` (a `re.sub` post-pass) is replaced by `_collect_high_entropy_spans`, which locates candidates against the **original** content and emits `_Span` entries that flow through the existing `_merge_spans` union and one right-to-left splice. Gating is unchanged and still mirrors `RedactionScanner._scan_high_entropy` exactly — same shared `HIGH_ENTROPY_CANDIDATE` object, same allowlist, same `entropy_threshold(min_confidence)`.

2. **Token-boundary snapping**, which is what makes the fix hold. Step 1 alone is *threshold-dependent*, and that is the same bug in a new dress: the union only fires when the whole combined run clears the entropy threshold. `AKIAIOSFODNN7EXAMPLE` + `"a"*14` measures 3.145 bits/char against the 3.7 default, so no entropy span is emitted at all and the output was still `[REDACTED:api_key]aaaaaaaaaaaaaa` — 14 chars of the same contiguous token a *critical* detector had matched. Worse, the reported fixture clears `min_confidence=1.0`'s 4.5 threshold by only 0.07 bits, so an operator raising that knob to cut false positives silently reopened the bug. `_snap_to_candidate_runs` now widens any span whose start or end falls **strictly inside** a candidate run out to that run's boundary, before merging. A match can no longer bisect a contiguous token, and the guarantee never consults the threshold. Runs whose exact text is in `false_positive_allowlist` are never widened — explicit user intent wins.

Because snapping widens, `--apply` may now redact slightly **more** than `--scan` reported. That is deliberate: a missed secret is unrecoverable once written, whereas over-redaction is visible and fixable.

**Relationship to PR #902 (issue #832):** this is #902's fix being *incomplete*, not a regression of it. #902 introduced the span-union pipeline for the regular patterns, and its own PR body records the carve-out verbatim — "The high-entropy detector stays an unchanged second pass." That exemption is exactly the hole #909 falls through. The `email_password_combo` leak #902 fixed is a different symptom (regex-vs-regex anchor destruction); this is detector-vs-regex partial overlap. Same root class, same structural remedy, now applied to the last detector left outside the union.

Public API unchanged. `patterns.py`, `scanner.py`, and the CLI/file layers are untouched; `redact --apply` still operates on fragment bodies only.

## Test plan

- New `TestHighEntropyOverlapLeak` (18 tests), written RED-first and verified against real implementations rather than asserted:
  - Against **HEAD**: 8 fail.
  - Against the **intermediate** (span union, no snapping): exactly **1** fails — `test_sub_threshold_entropy_tail_after_api_key_still_redacted`. So the snapping is load-bearing, not decorative.
  - Against the **final** implementation: all 18 pass.
- Every core assertion compares the **complete** returned string with `==`, never `secret not in out`. A containment assertion passes while this bug is live — that is exactly how it shipped — so the three pre-existing `TestRedactorHighEntropy` tests and the #435 property test were tightened from containment to full-output equality too. No assertion was removed: each replacement equality strictly subsumes the containment checks it replaced, and one docstring that claimed `min_confidence=0` was corrected to match its body's `1.0`.
- Threshold gating is bracketed from both sides by a 24-char probe at exactly 3.0 bits/char — inert at the 3.7 default, redacted at the 2.5 floor `min_confidence=0.0` selects — so a refactor that hard-coded or dropped the config read fails. Threshold-independence of the fix itself is pinned at `min_confidence=1.0`.
- `test_markers_are_inert_for_every_pattern_name` guards a real 0.016-bit cliff: `email_password_combo` is a 20-char run at 3.6842 bits/char against the 3.7 default.
- Scan/apply parity re-probed by hand across ten overlap geometries (leading tail, trailing tail, hyphen-joined, space-separated, low-entropy tail, `token=` prefix, card/ipv4 adjacency, `email:password`). In every case where `scan_file` reports `high_entropy_string`, the full run is gone from the applied output.
- All 278 redaction tests pass (`test_redact.py`, `test_properties_redaction.py`, `test_cli_redact.py`, `test_mcp_redact.py`, `test_redact_audit.py`).
- `cd creek-tools && ./scripts/check-all.sh` exits 0 — 12/12 checks, 6315 tests, 93.93% coverage, per-file gate green.
- Security-specialist audit of the snapping mechanism: **no blockers, no majors**. Confirmed no span can end strictly inside a non-allowlisted run; the allowlist carve-out cannot be turned into a leak amplifier (it keys on the whole maximal run, so a benign string cannot suppress an overlapping secret); snapping cannot widen across a newline (the candidate class excludes every character `str.splitlines()` treats as a boundary); and marker mislabeling can only ever go *upward* in severity, never downward.

Follow-ups filed rather than scope-creeping this P1: #945 (markers rewritten into nested markers at low `min_confidence` — pre-existing, distinct) and #946 (prose over-redaction from snapping, the O(spans × runs) cost, and allowlist documentation).

Closes #909

### DIFF
--- a/creek-tools/creek/redact/redactor.py
+++ b/creek-tools/creek/redact/redactor.py
@@ -5,10 +5,32 @@ The :class:`Redactor` re-scans content to locate match positions (since
 and replaces each hit with a ``[REDACTED:type]`` marker.
 
 Replacement is a single pass over the *original* content: every in-scope
-pattern is matched against the untouched text, truly overlapping spans
+detector — the regex patterns *and* the generic high-entropy detector
+alike — is matched against the untouched text, truly overlapping spans
 are unioned, and the merged spans are spliced out right-to-left. Because
-no pattern ever anchors on partially redacted text, a ``--scan`` finding
-cannot survive a ``--apply`` step (Issue #832).
+no detector ever anchors on partially redacted text, a ``--scan`` finding
+cannot survive a ``--apply`` step (Issue #832). Keeping the entropy
+detector inside that same union is what stops a regex match that covers
+only *part* of a high-entropy run from leaving the remainder — now too
+short for the detector to re-find — in cleartext (Issue #909).
+
+That union alone is not sufficient, because the entropy detector only
+fires when the *whole* run clears the confidence-derived threshold: a
+key with a low-entropy tail (or any operator who raises
+``min_confidence`` to cut false positives) would reopen the same leak.
+So spans are also **snapped to token boundaries** before merging — a
+span whose edge falls strictly inside a ``HIGH_ENTROPY_CANDIDATE`` run
+is widened to that run's edge, making the fix boundary-driven rather
+than threshold-driven. Runs the operator explicitly allowlisted are
+excluded from snapping: user intent wins, and a regex match inside such
+a run still redacts its own span unwidened.
+
+A consequence worth stating plainly: ``--apply`` may therefore redact
+slightly **more** than ``--scan`` reported, because snapping widens on
+token shape rather than on a reported finding. That asymmetry is
+deliberate, fail-closed behaviour — a missed secret is unrecoverable
+once written, whereas over-redaction is visible in the output and
+fixable via ``false_positive_allowlist``.
 
 Redaction logs are written as JSON with a session salt in the header so
 that hashes can be correlated within a session but not reversed.
@@ -70,6 +92,36 @@ class _MergedSpan(NamedTuple):
     start: int
     end: int
     contributors: list[_Span]
+
+
+_Run = tuple[int, int]
+"""Half-open ``(start, end)`` offsets of one high-entropy candidate run."""
+
+
+def _snap_one(span: _Span, runs: list[_Run]) -> _Span:
+    """Widen *span* to the boundaries of any candidate run it bisects.
+
+    A boundary is only moved when the span's edge falls *strictly*
+    inside a run, so a span already flush with a run boundary — or one
+    that fully contains the run — is returned unchanged. The pattern
+    name and collection order are preserved verbatim so the merged
+    span's marker selection is unaffected.
+
+    Args:
+        span: The match span to snap.
+        runs: Half-open offsets of the candidate runs to snap against.
+
+    Returns:
+        The span, widened to whole-token boundaries where it bisected a
+        run; otherwise an equal copy of the input.
+    """
+    start, end = span.start, span.end
+    for run_start, run_end in runs:
+        if run_start < span.start < run_end:
+            start = run_start
+        if run_start < span.end < run_end:
+            end = run_end
+    return _Span(start, end, span.pattern_name, span.order)
 
 
 def _severity_rank(pattern_name: str) -> int:
@@ -202,12 +254,37 @@ class Redactor:
 
         Re-scans *content* against the configured patterns (since matched
         text is never stored in :class:`RedactionMatch`) in a single pass
-        over the original text: every in-scope pattern is matched against
+        over the original text: every in-scope detector is matched against
         the untouched content, truly overlapping spans are unioned, and
         each merged region is replaced by the marker of its most severe
-        contributor. This guarantees scan/apply parity — a ``--scan``
-        finding cannot survive ``--apply``, because no pattern ever
-        anchors on already-redacted text (Issue #832).
+        contributor. This guarantees scan/apply parity — when
+        *pattern_types* is not narrowed, a ``--scan`` finding cannot
+        survive ``--apply``, because no detector ever anchors on
+        already-redacted text (Issue #832). Narrowing *pattern_types*
+        deliberately takes the excluded detectors out of scope, so their
+        findings are left untouched by design.
+
+        The generic high-entropy detector participates in that *same*
+        single-pass union rather than running as a post-pass over the
+        spliced output. Otherwise a regex match covering only part of a
+        high-entropy run would splice its marker over that part and leave
+        a remainder too short for ``HIGH_ENTROPY_CANDIDATE`` to re-match,
+        leaking the tail of a secret in cleartext (Issue #909).
+
+        Because that union still depends on the *combined* run clearing
+        the entropy threshold, spans are additionally snapped to token
+        boundaries by :meth:`_snap_to_candidate_runs`: any span edge
+        falling strictly inside a ``HIGH_ENTROPY_CANDIDATE`` run is
+        pushed out to that run's edge, so no contiguous token is ever
+        left half-redacted regardless of how ``min_confidence`` is
+        tuned. Allowlisted runs are exempt — an explicitly allowlisted
+        token is never widened onto.
+
+        Snapping is shape-driven, so ``--apply`` may redact slightly
+        **more** than ``--scan`` reported. That is deliberate,
+        fail-closed behaviour: a missed secret is unrecoverable once
+        written, while over-redaction is visible and fixable by
+        allowlisting the affected token.
 
         Args:
             content: The text to redact.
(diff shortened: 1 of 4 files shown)


Geoffe-Ga:
## Summary
PR #947 fixes a P1 secret-leak bug in `creek redact --apply`: the high-entropy detector ran as a post-pass over already-redacted text, so a regex match that only partially covered a high-entropy run (e.g. an AWS key with a high-entropy or even low-entropy tail) could leave a cleartext remainder. The fix folds the entropy detector into the same original-text span-union pipeline used for regex patterns (mirroring #832's structural remedy) and adds token-boundary snapping so the guarantee doesn't depend on the entropy threshold at all.

## Strengths
- Correct root-cause fix, not a patch over symptoms: moving `_replace_high_entropy` (a `re.sub` post-pass) to `_collect_high_entropy_spans` (spans against the *original* content) directly addresses the ordering bug described in the issue.
- The token-boundary snapping (`_snap_one` / `_snap_to_candidate_runs`) closes a second, subtler hole — the union alone is still threshold-dependent, and the PR's own fixture (`AKIAIOSFODNN7EXAMPLE` + 14 `"a"`s at 3.145 bits/char) demonstrates that the union-only fix still leaks. Good adversarial thinking against the intermediate fix.
- Test quality is excellent: every core assertion uses `result == "..."` full-string equality rather than `not in`/`in` containment. The PR body documents this was deliberate because a containment check is exactly how the original bug shipped undetected — this is the right level of rigor for a secret-leak fix.
- Snapping correctly excludes allowlisted runs (user intent wins) and doesn't cross newlines (the candidate character class `[A-Za-z0-9+/=_\-]{20,}` structurally can't match `\n`), verified by dedicated tests.
- Gating parity between `_collect_high_entropy_spans` and `RedactionScanner._scan_high_entropy` is exact (same regex object, same allowlist check, same `entropy_threshold` call), which is what preserves the scan/apply parity guarantee from #832.
- Good scope discipline: two related-but-distinct issues (nested-marker rewriting at low confidence, and the O(spans×runs) cost / prose-over-redaction risk of snapping) were filed as follow-ups (#945, #946, both verified to exist and open) instead of scope-creeping this P1.
- Docstrings and `docs/redaction.md` were updated to match the new behavior, including the intentional scan/apply asymmetry (apply may now redact slightly more than scan reported) — a real behavior change that's called out rather than buried.

## Security Concerns
🟢 LOW — None found. I manually traced `_snap_one`'s boundary logic against overlap cases (span inside a run, span spanning two runs with a gap, span flush with a run boundary, span fully containing a run) and it behaves correctly in each. The fail-closed design choice (over-redact rather than under-redact, with `false_positive_allowlist` as the escape hatch) is the right tradeoff for a secret-scanning tool. `RedactionMatch` still never stores matched text, so no new information is logged.

## Problems
None blocking.

## Code Quality
- `_snap_to_candidate_runs` and `_collect_high_entropy_spans` both call `HIGH_ENTROPY_CANDIDATE.finditer(content)` independently, and `_snap_one` is O(spans × runs) — already tracked in follow-up #946 rather than fixed here, which is reasonable for a P1 given the current input sizes (fragment bodies, not whole vaults).
- Everything else is clean: small, single-purpose functions, consistent NamedTuple usage, no dead code.

## CLAUDE.md Adherence
✅ Coverage — PR reports 93.93% branch coverage (≥90% threshold) and per-file gate green.
✅ Docstrings — every new function/method has a full Google-style docstring with Args/Returns; PR reports `check-all.sh` (which includes interrogate) exits 0.
✅ Types — new code is fully typed (`_Run = tuple[int, int]`, `list[_Span]`, etc.), consistent with MyPy strict mode.
✅ Ruff / patterns — code follows existing module conventions (splice right-to-left, NamedTuple spans, severity ranking) rather than introducing a new style.
✅ Commits/workflow — feature branch off an issue (`issue/909-...`), not a direct push to main; PR references and closes #909.
⚠️ I could not independently re-run `./scripts/check-all.sh` in this review sandbox (tooling restrictions), so I'm relying on the PR body's reported results plus my own static trace of the logic. At the time of review, GitHub's "Code Quality & Testing" jobs (3.11/3.12/3.13) were still IN_PROGRESS; "CrawDad Quality" jobs were already green. Worth confirming the quality jobs finish green before merge.

## Testing
Strong. 18 new tests in `TestHighEntropyOverlapLeak`, each targeting a distinct overlap geometry (trailing tail, leading tail, sub-threshold tail requiring snapping specifically, multiline boundary, allowlist carve-out, `pattern_types` narrowing, severity tie-break, idempotency, marker self-inertness at a documented 0.016-bit cliff). Three pre-existing tests were tightened from containment to full-string equality, and one incorrect docstring (`min_confidence=0` vs the actual `1.0`) was corrected. The PR body documents a RED→intermediate→GREEN verification process (8 fail at HEAD, exactly 1 fails against the union-only intermediate, all pass at final) which is exactly the kind of test-honesty this fix needed.

## Documentation
`docs/redaction.md` and the module/method docstrings in `redactor.py` were updated in lockstep with the behavior change, including the new scan/apply size asymmetry and the allowlist interaction. No gaps found.



## Verdict: LGTM

This is a well-scoped, thoroughly tested P1 security fix: the implementation correctly folds the high-entropy detector into the same original-text span union used for regex patterns, and the token-boundary snapping step closes the threshold-dependent gap the union alone would leave open. I traced the merge/splice/snap logic by hand against several boundary cases (partial overlap, sub-threshold tail, allowlisted runs, multiline content) and found no correctness or security gap; the 18 new tests use exact-string equality (not containment), which is the right rigor for a secret-leak fix. Nothing here blocks merge — CI's core quality jobs were still in progress at review time (CrawDad jobs already green), which the author should confirm goes green before merging.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5108269811 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #950, opened 2026-07-28
[security][purge] _purge_intimate_stub follows any in-vault pointer; NUL byte in pointer can raise; audit failure_reason interpolates exception text


Geoffe-Ga:
Surfaced by the Gate 2.5 security review on #910. **All three are pre-existing** in code that #910 does not touch (#910 changed only the frontmatter *loader* feeding these paths), so they were correctly deferred rather than folded into that PR.

All in `creek-tools/creek/purge/engine.py`.

## 1. `_purge_intimate_stub` — containment guard too wide (GAP-012)
It resolves `saved_from.intimate_body_pointer` and checks only `stub_path.is_relative_to(vault_root)`. Its sibling `_purge_journal_staged_entry` is stricter: it additionally scopes the resolved path to the journal staging directory.

Consequence: a hand-edited or malicious `intimate_body_pointer` aimed anywhere else *inside* the vault — `00-Creek-Meta/creek_config.yaml`, another fragment, an audit artifact — is followed and **deleted** during a purge. Reachable today via any ordinary valid-UTF-8 fragment.

**Fix:** mirror the `stub_root` / `INTIMATE_STUB_RELPATH` guard pattern already used by `_purge_journal_staged_entry`, so only paths under `10-Liminal/Compost/intimate-stubs/` are ever unlinked.

## 2. `Path.resolve()` can raise on an embedded NUL byte
Both `_purge_intimate_stub` and `_purge_journal_staged_entry` document a "missing/!bad pointer is a no-op, never an error" contract, but a double-quoted YAML scalar containing a NUL (`intimate_body_pointer: "a\0b"`) makes `Path.resolve()` raise `ValueError`, which is uncaught — breaking that contract and aborting the purge.

**Fix:** catch `ValueError` (alongside `OSError`) around the resolve and treat it as a no-op.

## 3. Audit `failure_reason` interpolates full exception text
`_run_audited` builds `failure_reason = f"{type(exc).__name__}: {exc}"` and writes it to the append-only, deliberately non-purgeable audit log. Exception text can carry vault-derived content (e.g. a title-derived filename, or a `yaml.MarkedYAMLError` source snippet). #910 established the house rule for the *logger* — type name only, never `str(exc)` — but the audit path predates it.

[...]

Issue #949, opened 2026-07-28
[purge] fragment_count drifts when a thread/eddy file is undecodable


Geoffe-Ga:
Split out of #910 (pre-existing; do not fix there).

## Context
`PurgeEngine._decrement_counts` (`creek-tools/creek/purge/engine.py`) loads each thread/eddy file through the **strict** `_load_frontmatter` and skips it on `None`.

#910 deliberately kept this call site on the strict loader: `_decrement_counts` rewrites the file with `frontmatter.dumps(post)` + `write_text`, so feeding it a lossy (`errors="replace"`) read would persist U+FFFD over the operator's original bytes. That trade-off is pinned by `test_decrement_counts_does_not_corrupt_undecodable_thread_file`.

## The gap
When a `02-Threads/` or `03-Eddies/` file is not valid UTF-8 (or is otherwise unparseable), purging one of its fragments leaves its `fragment_count` un-decremented. The count drifts high and stays wrong.

## Why it is low priority
This is pre-existing behaviour, not introduced by #910, and it is strictly preferable to the alternative (corrupting the file). `fragment_count` is display metadata, not a correctness or privacy surface — no RTBF implication.

## Suggested approach
Byte-level rewrite of just the `fragment_count` scalar, preserving all other bytes verbatim; or surface the drift so `creek lint` can flag it (coordinate with #926, which owns operator-facing reporting of skipped files).

## Acceptance
- Purging a fragment decrements the count on an undecodable thread/eddy file, **or** the drift is reported to the operator.
- The file's invalid bytes are preserved exactly; no U+FFFD is ever written.
- `cd creek-tools && ./scripts/check-all.sh` exits 0.

Issue #948, opened 2026-07-28
[purge] Wikilink/provenance references inside an undecodable file survive a purge unscrubbed


Geoffe-Ga:
Split out of #910 (do not fix there).

## Context
#910 fixed the RTBF correctness bug where a fragment with an undecodable (non-UTF-8) body silently escaped a scoped purge. As part of that fix, `PurgeEngine._scrub_one_file` (`creek-tools/creek/purge/engine.py`) widened `except OSError` to `except (OSError, UnicodeDecodeError)`, logging a WARNING and returning `(0, 0)`.

Before #910 that same file raised `UnicodeDecodeError` out of the purge, aborting it mid-run with `status="partial"` **before the target was ever unlinked**. Skipping is strictly better, but it leaves a residual gap.

## The gap
A wiki-link (`[[Title]]`) or a bare fragment-ID provenance mention living *inside* a file that is not valid UTF-8 is no longer scrubbed when the fragment it points at is purged. The dangling reference survives.

## Why it was not fixed in #910
The obvious fix — read with `errors="replace"` like the delete-decision path does — is **wrong here**, because `_scrub_one_file` writes `text` back via `write_text`. A lossy read followed by a write would persist U+FFFD over the operator's original bytes, i.e. recreate the exact data-loss bug #910 was careful to avoid on the rewrite paths. Skipping was the correct minimal choice.

## Suggested approach
Scrub at the byte level for this case (operate on `read_bytes()` and substitute encoded patterns) so the undecodable regions are preserved verbatim while references are still removed. Needs its own tests pinning byte-identity of the non-reference regions.

## Acceptance
- A purge scrubs references inside an undecodable `.md` file.
- The file's invalid bytes are preserved exactly; no U+FFFD is ever written.
- `cd creek-tools && ./scripts/check-all.sh` exits 0.

PR #951, opened 2026-07-28
fix(purge): stop a scoped purge silently skipping an undecodable fragment

### LINKED ISSUE #910: [scan:bugs] Scoped purge silently skips a matching fragment with an undecodable body, defeating RTBF
## Role
suppressions).

## Goal
Ensure a fragment that matches a scoped purge is never silently skipped merely
because its body is not valid UTF-8, so a right-to-be-forgotten purge cannot
report success while leaving the target file (and its secret body) on disk —
verified by a test that `purge_source` deletes a matching fragment whose body
contains invalid UTF-8 bytes.

## Context
- File(s): `creek-tools/creek/purge/engine.py:789-793` (`_load_frontmatter`),
  consumed by `_find_fragment_by_id`, `_fragments_from_source`,
  `_fragments_from_source_path`, `purge_daterange`, and `purge_classifications`.
- Evidence: `_load_frontmatter` swallows every exception and returns `None`:
  ```python
  try:
      return frontmatter.load(str(path))
  except Exception:
      logger.debug("Unable to parse frontmatter in %s", path)
      return None
  ```
  `frontmatter.load` raises `UnicodeDecodeError` on a file with invalid UTF-8
  bytes **even when the YAML frontmatter itself is well-formed**. Every scoped
  purge treats a `None` load as "skip this file". So a fragment that genuinely
  matches the purge criteria is dropped with only a debug log; the file survives
  and the operation records `status='complete'` with `fragments_deleted=0`.
  Live-reproduced: a vault whose only `claude` fragment has body bytes
  `0xff 0xfe` → `PurgeEngine(v).purge_source('claude')` returns
  `fragments_affected=0` and the file remains on disk with its secret body,
  while the audit records a successful purge. This defeats the RTBF guarantee.
- Related: same-root-cause sibling deferred this run —
  `_scrub_one_file` (engine.py:871-874) catches only `OSError`, so the same
  invalid-UTF-8 file crashes reference-scrubbing mid-run (partial purge). A fix
  here should coordinate with (but is independent of) that one. Sibling
  `[scan:bugs]` findings: redactor `--apply` leak, generic-ingestor duplication,
  attribution always-escalate.

## Output Format
"Closes #N".

## Examples
```python
def test_purge_source_deletes_fragment_with_undecodable_body(tmp_path):
    vault = _seed_vault(tmp_path)
    bad = vault / "01-Fragments" / "bad.md"
    bad.write_bytes(
        b"---\nid: frag-bad\nsource:\n  platform: claude\n---\nsecret \xff\xfe body\n"
    )
    res = PurgeEngine(vault).purge_source("claude")
    assert res.fragments_affected == 1     # today: 0
    assert not bad.exists()                # today: file survives
```
Fix sketch — distinguish "no/failed frontmatter" from "file exists but cannot be
decoded". Load with `errors="replace"` (matching the redaction scanner), or on a
decode failure still treat the file as a purge candidate for file-deleting
operations (and surface a WARNING), so an undecodable fragment cannot silently
defeat a right-to-be-forgotten purge. Narrow the bare `except Exception` to the
specific parse/decode errors so genuine bugs are not masked.

## Constraints
- No lint/type suppressions (max-quality-no-shortcuts): fix root causes — do not
  keep a bare `except Exception` that hides real defects.
- A destructive/RTBF operation must fail loud or complete, never silently
  under-delete while reporting success.

Geoffe-Ga:
## Summary

A fragment that matched a scoped purge could survive it, with its private body intact, while the audit recorded a successful purge. That defeats the right-to-be-forgotten guarantee, so this is a P1 correctness fix.

`PurgeEngine._load_frontmatter` swallowed every exception and returned `None`, and every scoped purge reads `None` as "skip this file". `frontmatter.load` raises `UnicodeDecodeError` on a file whose **body** carries non-UTF-8 bytes **even when its YAML frontmatter block is well-formed ASCII** — so a genuinely matching fragment was dropped with only a `logger.debug` line.

Live-reproduced on `main` before the fix, and again after:

```
# before                          # after
fragments_affected: 0             fragments_affected: 1
file still exists: True           file still exists: False
```

### The design: split the loader by write-safety

The helper was doing two different jobs for two classes of caller, and that collision is the bug.

| Caller class | Sites | What it does with the post |
|---|---|---|
| **Delete-decision** | `_find_fragment_by_id`, `_fragments_from_source`, `_fragments_from_source_path`, `purge_daterange` | reads `id`/`platform`/`original_file`/`created`, then `unlink()`s the file |
| **Rewrite** | `purge_classifications`, `_decrement_counts` | `frontmatter.dumps(post)` + `write_text` |

An `errors="replace"` read is exactly right for the first group and **catastrophic** for the second — it would persist U+FFFD over the operator's original bytes during a *non-destructive* metadata reset. So:

- **`_load_frontmatter_for_match`** (new, lossy) serves the four delete-decision sites. Strict read first; on `UnicodeDecodeError` it warns and re-reads through `path.open(encoding="utf-8", errors="replace")`. Its docstring opens by forbidding write-back, and no code path can write it.
- **`_load_frontmatter`** (body unchanged, strict) keeps the two rewrite sites. Two tests assert **byte-identity** on those paths and fail the instant anyone makes the shared loader lossy.

The lossy read is safe for that one job because a purge match is a function of frontmatter metadata only, never of the body. Replacing bad body bytes cannot change the outcome — it can only stop a matching fragment from escaping.

Chose two named helpers over a `lossy: bool` flag deliberately: a boolean trap on a destructive path is worse than two names, and two names make the call-site audit a single `grep`.

### Supporting changes

- Bare `except Exception` → shared `_FRONTMATTER_LOAD_ERRORS = (OSError, TypeError, ValueError, yaml.YAMLError)`. **`TypeError` is load-bearing** and must never be dropped: `frontmatter.load` splats metadata as `Post(content, handler, **metadata)`, so a bare YAML date key raises a plain `TypeError` the house `(OSError, ValueError, yaml.YAMLError)` tuple misses (precedent: #847 / PR #927). `UnicodeDecodeError` is a `ValueError` subclass, so **arm order is load-bearing** — it is caught ahead of the tuple.
- Every new warning logs `type(exc).__name__` **only** — never `str(exc)`, `exc_info`, or `logger.exception`. `yaml.MarkedYAMLError` stringifies with the offending source snippet, which would copy the very secret being purged into the log (house rule from PR #927).
- `_scrub_one_file` widens `except OSError` → `except (OSError, UnicodeDecodeError)`. This is a **minimal necessity, not scope creep**: before this change `purge_source` with only undecodable matches walked nothing, but now it matches them, and `_scrub_references` walks every vault `.md` — so a second undecodable file would raise out of `body()`, `_run_audited` would write `status="partial"` and re-raise, and the target would **never be unlinked** (unlink follows the scrub in `_purge_single`). That would convert a silent under-delete into a loud one, which does not satisfy the goal. It deliberately does **not** get an `errors="replace"` read: it writes `text` back, so a lossy read there would recreate the corruption bug in a second location.

### Where fail-closed stops

Per the operator constraint from PR #947 (#909), fail closed rather than open. The lossy re-read *dissolves* the ambiguity rather than resolving it toward deletion, so there is no residual "matches but undecidable" class.

A file that can be **neither decoded nor parsed** exposes no criteria surface at all and is left on disk. Deleting it under a *scoped* purge would be purging by absence of evidence — unbounded collateral loss in a vault that may hold many malformed notes. `purge_vault` remains the backstop: it wipes folder contents without ever parsing frontmatter, so the total-RTBF hammer still reaches those files (verified). And the residual skips are no longer silent — WARNING, not debug.

### Notes for the reviewer

- `frontmatter.load` is itself `open(fd, "r", encoding=...)` + `loads(text)`, so the lossy path goes through the **file object** rather than decoding bytes and calling `frontmatter.loads`. That keeps `frontmatter.load` the single parse entry point and preserves universal-newline handling — a CRLF vault file behaves identically on both paths (verified).
- Keeping the strict helper on `frontmatter.load` also keeps `test_load_frontmatter_handles_unreadable_file`, which monkeypatches that symbol, meaningful and **untouched**.
- The test diff is **purely additive**: 458 insertions, **zero deletions**. No pre-existing test was modified.

### Scope

#926 owns operator-facing *reporting* of silently-skipped files and is untouched — no new `PurgeResult` field, counter, summary object, or lint surface. This PR is the correctness half only.

Filed as follow-ups, not fixed here:
- **#948** — a reference inside an undecodable file is now skipped rather than crashing, so it survives unscrubbed; needs a byte-level scrub, since `errors="replace"` there would corrupt.
- **#949** — `fragment_count` drifts when a thread/eddy file is undecodable (pre-existing; preferable to corrupting it).
- **#950** — three pre-existing findings from the Gate 2.5 security review, in `_purge_intimate_stub` and `_run_audited`; none in code this PR touches.

## Test plan

15 new tests in `tests/test_purge.py`.

**10 fail before this change** (verified RED, then GREEN):
- Wrong counts / surviving files on all five scoped paths — `purge_source`, `purge_fragment`, `purge_source_path`, `purge_daterange`, and the audit record (`fragments_deleted == 1`, `affected_fragments == ["frag-A"]` while `status == "complete"` — the "reports success" half).
- A real `UnicodeDecodeError` out of `_scrub_one_file` on three (non-matching sibling, garbage file, undecodable thread file).
- A missing WARNING record on one (asserts level + path substring only, never wording).

**5 green-today regression guards**, each documented as such in its docstring:
- Two assert **byte-identity** on the rewrite paths (`purge_classifications`, `_decrement_counts`) — these fail the instant anyone makes the shared loader lossy.
- Two pin `TypeError` (bare YAML date key) and `yaml.YAMLError` (unterminated flow sequence) in the tolerated tuple.
- Two pin **no-over-deletion**: a non-matching undecodable file and a file that is both undecodable *and* unparseable both survive with identical bytes.

Test-quality guards: no assertion depends on file-visit order — `rglob` yields `os.scandir` order (hash-ordered on APFS), and an order-dependent test can false-green against this exact bug. The new byte-level `_vault_files_containing_bytes` helper is used instead of the existing `_vault_files_containing_secret`, which strict-decodes and would itself raise on these fixtures.

**Gate 2:** `cd creek-tools && ./scripts/check-all.sh` → exit 0, 12/12 checks, 6332 tests pass, aggregate coverage 93.95%, per-file gate clean. `creek/purge/engine.py` at 94.57% with every newly added line covered.

**Gate 2.5:** `code-review-orchestrator` (incl. an independent security pass over the RTBF deletion path) returned **CLEAN** — no blockers. The three security findings it surfaced are pre-existing in untouched code and are filed as #950.

Closes #910

### DIFF
--- a/creek-tools/creek/purge/engine.py
+++ b/creek-tools/creek/purge/engine.py
@@ -37,6 +37,7 @@ from datetime import UTC, date, datetime
 from typing import TYPE_CHECKING
 
 import frontmatter
+import yaml
 from pydantic import BaseModel, Field
 
 from creek.ingest.journal_staging import JOURNAL_STAGING_RELDIR
@@ -123,6 +124,29 @@ must tolerate nested-list entries (or skip purged drafts). The
 regression test ``test_purged_marker_reparses_as_nested_list_in_yaml``
 pins this trade-off so a future marker swap has to update the test and
 this docstring together.
+"""
+
+_FRONTMATTER_LOAD_ERRORS: tuple[type[Exception], ...] = (
+    OSError,
+    TypeError,
+    ValueError,
+    yaml.YAMLError,
+)
+"""Exceptions a hand-edited or corrupt vault file raises out of ``frontmatter.load``.
+
+Widens the house ``(OSError, ValueError, yaml.YAMLError)`` tuple with
+``TypeError``, which is load-bearing here and must never be dropped:
+``frontmatter.load`` splats the parsed metadata as
+``Post(content, handler, **metadata)``, so frontmatter carrying a
+non-string key — a bare YAML date like ``2024-05-01: note``, a realistic
+hand-edited-vault case — parses to a ``datetime.date`` key and raises a
+plain ``TypeError: keywords must be strings`` that the house tuple
+misses (precedent: PR #927 / issue #847).
+
+``UnicodeDecodeError`` is a ``ValueError`` subclass and is therefore
+covered deliberately: a file whose *body* carries non-UTF-8 bytes fails
+at read time, before its (possibly byte-clean ASCII) frontmatter block
+is ever parsed.
 """
 
 
@@ -453,7 +477,7 @@ class PurgeEngine:
         def body() -> None:
             """Run the date-range destructive ops, mutating ``result``."""
             for frag_file in self._list_fragment_files():
-                post = self._load_frontmatter(frag_file)
+                post = self._load_frontmatter_for_match(frag_file)
                 if post is None:
                     continue
                 created = _coerce_date(post.get("created"))
@@ -793,7 +817,7 @@ class PurgeEngine:
             Tuple of ``(path, post)`` or ``(None, None)`` if not found.
         """
         for frag_file in self._list_fragment_files():
-            post = self._load_frontmatter(frag_file)
+            post = self._load_frontmatter_for_match(frag_file)
             if post is not None and post.get("id") == fragment_id:
                 return frag_file, post
         return None, None
@@ -812,7 +836,7 @@ class PurgeEngine:
         """
         matches: list[tuple[Path, frontmatter.Post]] = []
         for frag_file in self._list_fragment_files():
-            post = self._load_frontmatter(frag_file)
+            post = self._load_frontmatter_for_match(frag_file)
             if post is None:
                 continue
             if _extract_source_platform(post) == source_type:
@@ -841,7 +865,7 @@ class PurgeEngine:
         predicate = _build_source_path_matcher(source_path, match=match)
         matches: list[tuple[Path, frontmatter.Post]] = []
         for frag_file in self._list_fragment_files():
-            post = self._load_frontmatter(frag_file)
+            post = self._load_frontmatter_for_match(frag_file)
             if post is None:
                 continue
             original_file = _extract_source_original_file(post)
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
Fixes a P1 right-to-be-forgotten bug: `PurgeEngine._load_frontmatter` swallowed every exception (including `UnicodeDecodeError` raised on a non-UTF-8 fragment *body*, even with well-formed ASCII frontmatter) and returned `None`, which every scoped-purge call site read as "skip this file." A matching fragment could therefore survive a purge with its private body intact while the audit log recorded success. The fix splits the loader into a strict, write-safe path (`_load_frontmatter`, used only by the two in-place rewrite sites) and a new lossy, delete-decision-only path (`_load_frontmatter_for_match` / `_load_frontmatter_lossily`, used by the four sites that only ever `unlink`), narrows the bare `except Exception` to a documented tuple (`OSError, TypeError, ValueError, yaml.YAMLError`), and makes `_scrub_one_file` tolerate an undecodable file instead of aborting the purge mid-flight.

## Strengths
- Correct, narrowly-scoped root-cause fix: the two-helper split (rather than a `lossy: bool` flag) makes the write-safety invariant grep-able at every call site, and is enforced by byte-identity regression tests on both rewrite paths.
- Security-conscious logging: every new log line logs `type(exc).__name__` only, never `str(exc)`, specifically to avoid leaking vault content (a YAML parse error can otherwise stringify the offending source snippet) or the very secret being purged into logs.
- I independently confirmed via the source (not just the PR description) that the two remaining strict-loader call sites are exactly the rewrite paths (`purge_classifications` at engine.py:429, `_decrement_counts` at engine.py:1137) — no delete-decision site was left on the old, buggy path.
- `TypeError` is deliberately retained in the tolerated tuple with a documented precedent (#847/PR #927: a bare YAML date key becomes a non-string kwarg to `Post(**metadata)`), preventing a plausible future regression from "cleaning up" the exception tuple.
- Explicit non-goals and follow-ups (#948, #949, #950) keep this PR's diff focused on the correctness half only, rather than scope-creeping into reporting/lint/security fixes it doesn't own.
- Test suite is unusually rigorous for this class of bug: it pins fail-closed behavior (no over-deletion of non-matching/garbage/malformed files), pins the two rewrite paths as lossless via byte-equality, and pins WARNING-level visibility without coupling to exact wording.

## Security Concerns
🟢 LOW — None introduced by this change. The lossy re-read is scoped strictly to delete-decision call sites that never write the post back, and the design correctly reasons that a purge-match decision depends only on frontmatter metadata, never body bytes, so lossy substitution of body bytes cannot change a match outcome. Log messages avoid echoing file content or exception string representations that could leak vault secrets.

## Problems
None found blocking merge.

## Code Quality
- No changes requested. The module-level `_FRONTMATTER_LOAD_ERRORS` constant with a trailing docstring, and the three-method split, are consistent with the file's existing conventions and read clearly.

## CLAUDE.md Adherence
✅ TDD: PR description documents a RED→GREEN cycle (10 new tests fail before the fix, pass after) plus 5 already-green regression guards explicitly labeled as such.
✅ Coverage/complexity/typing: per the PR description, `check-all.sh` passed 12/12 with 93.95% aggregate coverage and `engine.py` at 94.57% with every new line covered; mypy strict and ruff are part of that gate.
✅ Commit/workflow: issue-linked branch (`issue/910-...`), Gate 2.5 orchestrator review reported clean with out-of-scope findings correctly filed as follow-up issues rather than silently dropped or fixed out-of-scope.
I was not able to independently re-run `./scripts/check-all.sh` in this review environment (sandboxed, no shell execution permission granted), so the above is trust-but-verify against the PR's self-reported gate results plus my own static review of the diff, which found no correctness or scope gaps.

## Testing
Strong. 15 new tests in `tests/test_purge.py` cover: all four scoped-purge call sites (`purge_source`, `purge_fragment`, `purge_source_path`, `purge_daterange`) actually deleting an undecodable-bodied match; the audit record correctly counting the deletion; non-matching and syntactically-unparseable files surviving byte-for-byte; both rewrite paths (`purge_classifications`, `_decrement_counts`) refusing to corrupt an undecodable file; the two previously-fragile exception classes (`TypeError` from a bare YAML date key, `yaml.YAMLError` from malformed flow syntax) pinned into the tolerated tuple; two simultaneously-undecodable fragments both being purged without aborting the walk; and WARNING-level log visibility asserted on level + path substring only (not exact wording, avoiding brittleness). Notably the test diff is purely additive — no existing test was altered to make this pass.

## Documentation
Docstrings on all three loader methods explicitly state the write-safety contract, cross-reference each other (`See :meth:`), and explain *why* the lossy read is safe for match-only use — this is exactly the kind of non-obvious invariant that should live in a docstring, and it does.



## Verdict: LGTM

The fix closes the exact RTBF gap described (undecodable-body fragments surviving a scoped purge while the audit log claims success), correctly splits the frontmatter loader by write-safety so the lossy retry can never corrupt a rewritten file, and is backed by 15 new tests covering all four delete-decision call sites plus non-matching/garbage/malformed-YAML boundary cases and log-visibility. I verified in the actual source that the only two remaining strict `_load_frontmatter` call sites are exactly the two rewrite paths (`purge_classifications`, `_decrement_counts`), confirming no delete-decision site was missed. No blocking issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5108996516 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #952, opened 2026-07-28
fix(purge): stop an intimate-body pointer deleting any in-vault file

### LINKED ISSUE #950: [security][purge] _purge_intimate_stub follows any in-vault pointer; NUL byte in pointer can raise; audit failure_reason interpolates exception text

Geoffe-Ga:
## Summary

`PurgeEngine._purge_intimate_stub` resolved `saved_from.intimate_body_pointer` and guarded it with **only** `is_relative_to(vault_root)` before calling `unlink(missing_ok=True)`. Any hand-edited pointer aimed at another path *inside* the vault — `00-Creek-Meta/creek_config.yaml`, an unrelated fragment, the audit log — passed that sole guard and the victim file was deleted. That was an arbitrary in-vault file-deletion primitive reachable today through any ordinary valid-UTF-8 fragment.

The sibling `_purge_journal_staged_entry` was already correct: it scopes deletion to the journal staging dir with a second guard. **That sibling is the model this PR follows** — no departure from its shape.

Three findings from #950, all in `creek/purge/engine.py`:

1. **Containment (the P1).** A second guard requires the **resolved** stub path to be under `10-Liminal/Compost/intimate-stubs/` (`INTIMATE_STUB_RELPATH`, the existing single source of truth). It runs *before* the counter increment, so a dry run cannot preview a deletion the real run would refuse. Guarding the resolved path means an `intimate-stubs/../../../00-Creek-Meta/creek_config.yaml` walk is refused too; using `is_relative_to` means a legitimately nested stub is still swept.
2. **NUL byte.** An embedded NUL made `Path.resolve()` raise an uncaught `ValueError` that aborted the entire purge, violating the documented "a missing or bad pointer is skipped, never an error" contract. A new shared helper `_resolve_pointer_in_vault(pointer, *, field)` owns join + resolve + vault-root guard for **both** sweeps and catches exactly `(OSError, ValueError)` around the resolve alone.
3. **Audit leak.** `_run_audited` wrote `f"{type(exc).__name__}: {exc}"` into the deliberately **non-purgeable** audit log, so vault-derived exception text (a title-derived filename, a `yaml.MarkedYAMLError` source snippet) outlived the RTBF request that produced it. Now the type name only — matching the rule #910 set for the logger on this same module. The full message still propagates on the exception, and `creek purge` does not swallow it.

**Fail closed.** Every new branch either refuses or behaves exactly as before; nothing is ever deleted that was not deleted before. One deliberate, documented side effect: a hand-authored stub parked *outside* the canonical directory now survives a purge. That is the safe direction — the note itself is still deleted, so the RTBF request is honoured, and the operator can remove the stray file explicitly.

**No config oracle.** Refusal warnings name only the raw (attacker-controlled) pointer and the caller-supplied root — never the resolved victim path — so the log cannot be turned into a filesystem existence oracle. A test pins this.

Does not touch `_load_frontmatter` / `_load_frontmatter_for_match` / `_load_frontmatter_lossily`: PR #951's write-safety split and its byte-identity tests are left fully intact.

## Test plan

Gate 1 RED was proven by data loss, not by an incidental counter — on `main` these tests fail with `00-Creek-Meta/creek_config.yaml` genuinely deleted from the vault.

12 new tests, plus one existing test strengthened:

- Refuses a pointer at the vault config, at a sibling fragment, and at the audit log (each asserts the victim survives **byte-identical**; the audit-log case also asserts the intent+outcome pair survives and `verify()` holds).
- Refuses a `..` walk back out of the stub dir — the test that proves this is resolved-path containment, not a string-prefix check.
- Dry run refuses out-of-scope pointers (counter stays 0).
- Refusal warning does not echo the resolved victim path (asserts level + absence, never wording).
- Absolute-path pointer refused — pins the `Path("/vault") / "/etc/passwd"` footgun, where the join discards the left operand entirely.
- Pointer naming the stub root itself is a no-op.
- Nested stub (`intimate-stubs/2026/Alpha.md`) still swept — green before and after, so an exact-parent-equality guard would fail it.
- NUL byte in `intimate_body_pointer` and in `source.origin_key` are both no-ops that complete the purge with `status="complete"`. Both hand-write the markdown (PyYAML's emitter will not round-trip a NUL) and assert the loaded pointer really contains `\x00` *before* purging, so they cannot pass vacuously.
- `failure_reason == "OSError"` with the message absent from the audit log file on disk (the RTBF assertion). The existing partial-outcome test was strengthened, not weakened: `assert outcome.failure_reason` is kept and a `not in` assertion added.

No test depends on `rglob` visit order.

Gate 2: `./scripts/check-all.sh` exits **0** — 12/12 checks, 6345 passed, coverage 93.97%. Gate 2.5: `code-review-orchestrator` returned **CLEAN**; its three non-blocking coverage suggestions were folded in (absolute-path and stub-root-itself tests) rather than deferred.

Closes #950

### DIFF
--- a/creek-tools/creek/purge/audit.py
+++ b/creek-tools/creek/purge/audit.py
@@ -106,8 +106,12 @@ class PurgeAuditEntry(BaseModel):
         status: For outcome entries only — ``"complete"`` if the body
             ran without raising, ``"partial"`` if it aborted partway.
             ``None`` for intent entries and for pre-GAP-002 outcomes.
-        failure_reason: Short string capturing the exception type and
-            message when ``status="partial"``; ``None`` otherwise.
+        failure_reason: The exception **type name only** (e.g.
+            ``"OSError"``) when ``status="partial"``; ``None``
+            otherwise. The message is deliberately excluded: this log
+            is preserved by every purge, so vault-derived text quoted
+            in an exception would outlive the right-to-be-forgotten
+            request that produced it.
         target: Legacy field preserved for backward compatibility on
             read; populated only when reading pre-Batch-C entries.
         count: Legacy fragment count preserved for backward compatibility
--- a/creek-tools/creek/purge/engine.py
+++ b/creek-tools/creek/purge/engine.py
@@ -657,6 +657,54 @@ class PurgeEngine:
         if not self.dry_run:
             frag_file.unlink()
 
+    def _resolve_pointer_in_vault(self, pointer: str, *, field: str) -> Path | None:
+        """Resolve a frontmatter-supplied path pointer inside the vault.
+
+        Pointers such as ``saved_from.intimate_body_pointer`` and
+        ``source.origin_key`` are vault *content*: an operator (or
+        anyone who can write one note) controls their text, so they are
+        untrusted input to a destructive path. This helper turns such a
+        pointer into an absolute path only when it is safe to follow —
+        a refusal is a no-op, never an error, per the purge engine's
+        "a missing or bad pointer is skipped" contract.
+
+        A pointer is refused when it cannot be resolved at all (an
+        embedded NUL byte makes :meth:`~pathlib.Path.resolve` raise
+        ``ValueError``; a resolution loop or unreadable component
+        raises ``OSError``) or when it resolves outside the vault root.
+        Both refusals are logged at WARNING so an ignored pointer is
+        distinguishable from a sweep that never ran.
+
+        Args:
+            pointer: Vault-relative path string taken from frontmatter.
+            field: Frontmatter field name the pointer came from, used
+                verbatim in the refusal logs so operators can grep for
+                the field that carried the bad value.
+
+        Returns:
+            The resolved absolute path, or ``None`` meaning *refuse* —
+            the caller must not follow this pointer.
+        """
+        try:
+            resolved = (self.vault_path / pointer).resolve()
+        except (OSError, ValueError):
+            logger.warning(
+                "Ignoring %s %r that cannot be resolved to a path",
+                field,
+                pointer,
+            )
+            return None
+        vault_root = self.vault_path.resolve()
+        if not resolved.is_relative_to(vault_root):
+            logger.warning(
+                "Ignoring %s %r that resolves outside the vault root %s",
+                field,
+                pointer,
+                vault_root,
+            )
+            return None
+        return resolved
+
     def _purge_intimate_stub(
         self,
         post: frontmatter.Post,
@@ -673,10 +721,19 @@ class PurgeEngine:
         right-to-be-forgotten request.
 
         The method is defensive: a missing or empty pointer, an
-        already-deleted stub, and a pointer that resolves outside the
-        vault root are all treated as no-ops rather than errors. The
-        last case prevents a hand-edited or malicious pointer
-        (``../../secret``) from steering a delete outside the vault.
+        unresolvable pointer, an already-deleted stub, and a pointer
+        that resolves outside the vault root are all treated as no-ops
+        rather than errors. A second containment guard additionally
+        scopes deletion to the canonical stub directory itself, so a
+        hand-edited or malicious pointer (``../../secret``,
+        ``00-Creek-Meta/creek_config.yaml``, or a ``..`` walk back out
+        of the stub dir) can never steer a delete at an arbitrary file.
+
+        That guard is deliberately fail-closed, with one accepted side
+        effect: a hand-authored stub parked *outside* the canonical
+        directory now survives the purge. That is the safe direction —
+        the note itself is still deleted, so the RTBF request is
+        honoured, and the operator can purge the stray file explicitly.
 
         Args:
             post: Frontmatter of the note being purged.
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real security bug in `PurgeEngine._purge_intimate_stub` (creek-tools/creek/purge/engine.py): a note's `saved_from.intimate_body_pointer` was resolved and guarded only with `is_relative_to(vault_root)`, so a hand-edited pointer could delete *any* in-vault file (e.g. `00-Creek-Meta/creek_config.yaml`, a sibling fragment, or the purge audit log itself) during an ordinary scoped purge. The fix adds a shared `_resolve_pointer_in_vault` helper (join + resolve + vault-root guard, refusing on `OSError`/`ValueError` instead of crashing on an embedded NUL byte) and a second containment guard scoping the intimate-stub sweep to `10-Liminal/Compost/intimate-stubs/` and the journal-staging sweep to its staging dir. It also stops the audit log from persisting exception *messages* (only the exception type name is kept), since vault-derived text in a message would otherwise outlive an RTBF purge.

## Strengths
- Root-caused correctly: the vulnerability is "resolved path under vault root" being far too permissive a containment check for a destructive delete driven by attacker-controlled frontmatter content.
- Fix is fail-closed and applies identically to both pointer types (`intimate_body_pointer`, `source.origin_key`) via one shared helper — no asymmetric coverage.
- The dry-run containment check runs *before* the counter increment, so `--dry-run` can't misreport a deletion the real run would refuse (this is directly tested).
- Warning logs deliberately omit the *resolved* victim path (only the raw attacker-controlled pointer and the caller-supplied root are logged), preventing the log from being used as a filesystem-existence oracle — also directly tested.
- The audit-log message-leak fix (type name only, no `str(exc)`) is a good defense-in-depth catch: the audit log is intentionally non-purgeable, so it's exactly the wrong place for vault-derived exception text to persist past an RTBF request.
- Excellent test coverage: pointer-at-config, pointer-at-sibling-fragment (with proof the sibling remains independently purgeable), pointer-at-audit-log (with hash-chain verification), `..`-walk-back-out (proves resolved-path containment, not string-prefix), absolute-pointer footgun (`Path("/vault") / "/etc/passwd"` join behavior), NUL-byte handling for both pointer types, and a no-regression test for legitimately nested stubs.
- Local-import-with-explanatory-comment for `INTIMATE_STUB_RELPATH` is verified correct: importing `creek.save._constants` does trigger `creek/save/__init__.py`, which does pull in the heavier `writer`/`privacy_filter`/LLM-provider chain, so keeping this import function-scoped is justified, not cargo-culted.
- Docs (`cleaning-and-purge.md`, `save.md`) updated to describe the new containment scope and the accepted "stray stub outside canonical dir survives" side effect.

## Security Concerns
None remaining 🟢 — this PR *is* the security fix. The new guard is resolved-path containment (not a string-prefix check, confirmed by the `..`-walk test), refusal logs don't leak the resolved victim path, and the swept scope (`is_relative_to`, checked before increment) is proven correct across the added tests. No sibling module (e.g. redact) was found with the same unguarded pattern.

## Problems
None found. This is a tightly scoped, correctly targeted fix.

## Code Quality
- Non-blocking nit: `_purge_journal_staged_entry`'s own out-of-scope warning doesn't have a dedicated test asserting it also omits the resolved path (only the intimate-stub sweep has `test_refused_stub_pointer_warning_does_not_echo_resolved_target`). Given both paths now share `_resolve_pointer_in_vault` and the second-guard code is structurally identical, this is very low risk, but a parallel test would make the origin_key path's oracle-safety explicit rather than inferred by symmetry.

## CLAUDE.md Adherence
✅ Commit message follows Conventional Commits (`fix(purge): ...`).
✅ PR description states Gate 2 (`check-all.sh` exit 0, 6345 tests passed, 93.97% coverage) and Gate 2.5 (`code-review-orchestrator` returned CLEAN) were run, consistent with the repo's 4-gate "Stay Green" workflow.
✅ TDD or its equivalent evidenced by "12/12 checks... Gate 1 RED was proven by data loss" language and thorough test additions.
✅ Docstrings on all new/changed public and private methods, in the repo's Google-style format with `Args`/`Returns` sections — consistent with the ≥95% interrogate requirement.
✅ Typing uses modern `Path | None` syntax consistent with strict mypy mode elsewhere in the file.

## Testing
Strong. 12 new tests plus one strengthened existing test, each targeting a distinct attack surface: same-vault victim (config/sibling/audit-log), path-walk-back-out, absolute-pointer join footgun, stub-root-itself as a no-op, dry-run non-preview, oracle-safe warning content, and NUL-byte handling for both pointer fields — plus an audit-log message-leak regression test with a synthetic marker asserted absent from the raw file bytes on disk. No test appears to rely on filesystem iteration order. This meets the bar for a security-fix PR.

## Documentation
`cleaning-and-purge.md` and `save.md` were both updated to reflect the new containment scope and the one accepted behavioral side effect (a stub parked outside the canonical directory now survives a purge and must be removed explicitly) — documentation matches the implementation.



## Verdict: LGTM

The fix closes a real arbitrary in-vault file-deletion primitive (attacker-controlled frontmatter pointer + `is_relative_to(vault_root)` alone was too weak a guard), applies the same containment fix symmetrically to both pointer types, fails closed with a documented and safe side effect, avoids a filesystem oracle in its warning logs, and ships 12 new tests plus one strengthened test that directly prove the vulnerability existed on `main` and is now closed. No scope creep, no blocking issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5109831995 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #953, opened 2026-07-28
Ledger-wire the generic source so an edited .txt/.log updates in place instead of orphaning its fragment


Geoffe-Ga:
## Context

Split out of #911. That issue makes `GenericIngestor.parse` stamp a stable, mtime-derived timestamp so re-ingesting an *unchanged* file is a genuine no-op (the deterministic id dedups the write). That closes unbounded duplication, but it deliberately does **not** close two adjacent identity problems, because both need the ledger rather than a better timestamp:

1. **An edited `.txt`/`.log` mints a new id and orphans the old fragment.** Content changes -> hash input changes -> new `frag-…`. The prior fragment is left behind with no pointer to its successor. This is exactly the #672/#673 problem class that the markdown/journal source solved with `source_key` + `SourceLedger` + update-in-place.
2. **A bare `touch` (mtime bump, byte-identical content) mints a new id.** The timestamp is part of `generate_fragment_id`'s hash input, so a no-op mtime change looks like a new unit. The ledger's content-hash comparison is what distinguishes "touched" from "edited"; `--incremental` already relies on it for markdown (`unit_is_changed` in `creek/pipeline.py`).

## Why this is not a one-line change

`ledger_for_source` (`creek/ingest/pipeline.py:73`) returns `None` for everything but `markdown` on purpose — the ledger was scoped to mutable journal units living *inside* the vault. Wiring `generic` in has a real hazard:

- `derive_source_key` (`creek/ingest/pipeline.py:57-71`) falls back to `candidate.name` — a **bare filename** — when the source path is not relative to the vault root. Generic sources are almost always outside the vault (`~/Downloads/*.txt`, log files), so `~/a/notes.txt` and `~/b/notes.txt` would collide on a single `source_key` and update-in-place over each other. That is data loss, not duplication.
- A non-`None` ledger also switches on `tomb_missing_units` for the source type, changing directory-pass semantics for generic.

So this needs a **collision-safe `source_key` for out-of-vault sources** (e.g. hash the resolved absolute path, or namespace by resolved parent) before the generic source can safely be ledger-backed.

## Scope

- Design and implement a collision-safe `derive_source_key` for sources outside the vault root, without changing the existing in-vault key shape (markdown ledgers on disk must keep resolving).
- Wire `generic` into `ledger_for_source` and verify update-in-place, touch-without-edit skip, and tombing semantics for a directory pass.
[...]

Issue #672, opened 2026-06-26
feat(ingest): add source_key provenance + SourceLedger skeleton


Geoffe-Ga:
## Role

You are a Senior Python engineer working in this repo's `creek/ingest/` +
`creek/vault/` modules.

## Goal

Add a stable **`source_key`** provenance field to the fragment model /
frontmatter and a **`SourceLedger`** abstraction that round-trips
`00-Creek-Meta/State/ingest/<source>.jsonl` (`source_key → {fragment_id,
content_hash, last_seen}`), wired into the markdown ingest path as a
**behaviour-preserving no-op pass-through** — it records ledger entries and
populates `source_key`, but changes no write/skip behaviour.

## Context

- **Parent epic:** #668
- **Predecessor issue(s):** none — this is the skeleton issue.
- **SPEC section:**
  `creek-tools/plans/2026-06-26_SPEC_continuous-idempotent-ingest.md` — R1 (§3,
  lines 106–128), specifically the `source_key` (line 109–112) and the per-source
  ledger (lines 113–114). Phase 1 (§5, lines 216–218). Resolved decision #1 (§6,
  lines 235–237).
- **Files involved:**
  - `creek/ingest/base.py` — `generate_fragment_id` (line 329) and
    `assemble_ingested_fragment` (line 384, injects id into frontmatter); add
    `source_key` derivation + injection here.
  - `creek/ingest/markdown.py` — `MarkdownIngestor` (journal source); the wiring
    site for the pass-through ledger record.
  - `creek/models.py` (Fragment model / provenance) — add the optional
    `source.origin_key` provenance field.
  - `creek/config.py` — `SourcePaths` (line 1244) names the `00-Creek-Meta` state
    location; use it to resolve `00-Creek-Meta/State/ingest/<source>.jsonl`.
  - **New:** a `SourceLedger` module (e.g. `creek/ingest/ledger.py`) — read/write
    the per-source JSONL.
  - `tests/` — new smoke tests for the ledger round-trip and `source_key`
    population.
- **Prior decisions:**
  - UPDATE IN PLACE is the Phase-1 design (SPEC §6 #1) — but **no update/tomb
    logic in this issue**; this is the skeleton seam only.
  - `source_key` = vault-relative **source path** for file sources (Drive file-id
    and append-only per-event ids are out of scope here).
  - Fragment ids today are `frag-sha256(source:timestamp:content)[:12]`
    (`base.py:329`); the writer skips a known id and never updates
[...]

PR #954, opened 2026-07-28
fix(ingest): derive generic fragment id from file mtime, not wall clock

### LINKED ISSUE #911: [scan:bugs] GenericIngestor stamps wall-clock timestamp, duplicating fragments on every re-ingest
## Role
suppressions).

## Goal
Make `GenericIngestor.parse` stamp fragments with a stable, source-derived
timestamp so the deterministic fragment id is identical across re-ingests of an
unchanged file, eliminating unbounded duplicate fragments — verified by a test
asserting two `parse` calls on the same file yield an equal `timestamp` (and
hence an equal `generate_fragment_id`).

## Context
- File(s): `creek-tools/creek/ingest/generic.py:247-259` (`GenericIngestor.parse`),
  with the id derivation in `creek-tools/creek/ingest/base.py:329-345`
  (`generate_fragment_id`) and `base.py:~425` (`assemble_ingested_fragment`).
- Evidence: `parse` stamps the fragment with wall-clock `now`:
  ```python
  now = datetime.now(tz=LA_TZ)                 # 247
  authored_at = _safe_file_mtime(raw.path)     # 248 — stable, but only used for metadata
  return [ParsedFragment(..., timestamp=now)]  # 258
  ```
  The fragment id is derived deterministically from that timestamp:
  `assemble_ingested_fragment` → `generate_fragment_id(source, parsed.timestamp,
  content)`, and `generate_fragment_id` hashes
  `f"{source}:{timestamp.isoformat()}:{content}"`. Because `parsed.timestamp` is
  the wall clock at parse time rather than a stable source-derived value, every
  ingest run produces a DIFFERENT id for the same unchanged file. Every other
  ingestor uses a stable source timestamp (markdown: file creation time;
  code/spreadsheets: file mtime; discord/chatgpt/claude: message epoch), and the
  generic path has **no ledger** (`ledger_for_source` returns `None` for anything
  but `markdown`), so it relies solely on the deterministic id for dedup. The
  stable mtime is already computed in this same function (`authored_at`) but is
  only used for metadata. Result: re-ingesting the same unclaimed `.txt`/`.log`
  file writes a brand-new fragment on **every run** — unbounded duplication.
  `tests/test_ingest_generic.py::test_parse_sets_timestamp` only asserts
  `isinstance(datetime)`, so it does not catch this.
- Related: sibling `[scan:bugs]` idempotency finding deferred this run —
  `write_fragment_idempotent` unchanged-branch (pipeline.py) also duplicates a
  fragment after an in-place edit. Both are ingest-idempotency defects but have
  distinct root causes and fixes.

## Output Format
"Closes #N".

## Examples
```python
def test_parse_timestamp_is_stable_across_reingests(tmp_path):
    f = tmp_path / "note.txt"
    f.write_text("stable content")
    raw = _raw_document_for(f)               # build/discover a RawDocument
    first = GenericIngestor().parse(raw)[0]
    second = GenericIngestor().parse(raw)[0]
    assert first.timestamp == second.timestamp   # today: two datetime.now() differ
```
End-to-end variant: ingest a directory with one `.txt` through the generic path
twice and assert exactly one fragment file exists on disk.

Fix sketch — use the file's mtime for the id-bearing timestamp, reusing the
value already computed for `authored_at`:
```python
timestamp = _safe_file_mtime(raw.path) or datetime.now(tz=LA_TZ)
return [ParsedFragment(..., timestamp=timestamp)]
```
Keep `now()` only as the fallback for the synthetic/no-file-on-disk case.

## Constraints
- Preserve behavior for the synthetic case where no file mtime is available.

Geoffe-Ga:
## Summary

`GenericIngestor.parse` stamped `timestamp=datetime.now(tz=LA_TZ)`, and `generate_fragment_id` hashes `f"{source}:{timestamp.isoformat()}:{content}"`. The generic source has **no ledger** (`ledger_for_source` returns `None` for anything but `markdown`), so it relies solely on that deterministic id for dedup — and the id was not deterministic. Re-ingesting an unchanged `.txt`/`.log` minted a fresh id and wrote a brand-new fragment **on every run**: unbounded duplication. The stable mtime was already computed one line above, for `metadata['authored_at']`, and then discarded.

`parse()` now calls `_safe_file_mtime` **once** and reuses that single value for both `timestamp` and `authored_at`. The wall clock survives only as the fallback for a pathless/synthetic `RawDocument` where `stat()` fails, where `authored_at` stays honestly `None`.

### Approach: deterministic identity, not post-hoc de-duplication

Per the issue and the prior art (#672/#673 solved a similar id-derivation problem with a stable `source_key` + ledger rather than de-duplicating after the fact), the fix is at the **derivation**, not a cleanup pass. Two candidates were considered:

- **Chosen — mtime-derived timestamp.** Minimal, uses the value already computed in the same function, and matches every sibling file-based ingestor (code/spreadsheets key on mtime; markdown on file creation time; discord/chatgpt/claude on message epoch).
- **Rejected for now — wire `generic` into the `SourceLedger`.** Not a one-liner, and actively hazardous today: `derive_source_key` falls back to a **bare filename** when the source lives outside the vault, which generic sources almost always do (`~/Downloads/*.txt`, log files). Two different `notes.txt` in different folders would collide on one `source_key` and update-in-place over each other — data loss, not duplication. A non-`None` ledger also switches on `tomb_missing_units` for the source type. This needs a collision-safe `source_key` first; filed as **#953**.

### Timezone — load-bearing

The mtime is used **verbatim in UTC** and never converted to `LA_TZ`. The UTC offset string is part of the hashed `isoformat()`, so rendering in local time would bake `-07:00`/`-08:00` into fragment identity and make it a function of the host's tzdata and DST state. `file_modified_time`'s `datetime.fromtimestamp(st_mtime, tz=UTC)` is a pure function of the epoch float — identical on every machine and under every `TZ`. A test pins `utcoffset() == timedelta(0)` and the `+00:00` suffix so this can't regress.

### Blocking prerequisite included: `source.platform`

`generate_frontmatter` emitted `"platform": "unknown"`, which is **not a member of `SourcePlatform` at all** (the fallback member is `OTHER`). Every generic fragment therefore raised a pydantic `ValidationError` inside `assemble_ingested_fragment`, which `run_ingest` swallowed into `result.errors` — **generic fragments have never reached a vault.** Verified by running it against `main`.

This is a prerequisite rather than scope creep: the issue's mandated "ingest the same source twice, assert the fragment count is unchanged" test would otherwise pass **vacuously on `0 == 0`**, and would stay vacuous after a timestamp-only fix. Corrected to `SourcePlatform.OTHER`, whose routing target is the unchanged `01-Fragments/Unsorted/` (`_PLATFORM_SUBFOLDER[OTHER] == "Unsorted"`) — exactly what `generic.py`'s docstrings already promised.

### Why this matters past the duplicate count

This is an **idempotency** bug, not a disk-bloat one. Duplicated fragments are classified independently, so the same content can land at **different privacy tiers** across runs — and the merged tier gates (#846 read-side, #848 write-side, #876 the classifier) all trust that value. A source-scoped right-to-be-forgotten purge (#910) has to find *every* copy, and copies created under drifting identity are exactly what such a purge misses. Fixing the derivation is what makes those guarantees hold.

### Back-compat: no migration needed, nothing orphaned

Stated explicitly because the alternative would be silent orphaning: because `"unknown"` never validated, **there is no on-disk population of generic fragments carrying wall-clock-derived ids to migrate.** There is also nothing to migrate *from* — no ledger, no `origin_key`, so no persisted mapping of which wall-clock id came from which file. Rewriting ids in place would in any case be more destructive than the duplicates, since ids are referenced by resonance links, thread/eddy membership, the per-directory `.id-index.jsonl`, and the chained provenance log.

### Behaviour changes to note at review

- `source.platform` for generic fragments: invalid `"unknown"` → `"other"` (routing unchanged).
- `created` in generic frontmatter: parse-time wall clock → source mtime in UTC. Deliberate and more honest — `creek/time.py` documents that `created` is **not** in the authored-at precedence chain (all time-bucketing routes through `effective_authored_at`: `authored_at` → `ingested`), and other file-based ingestors already key on mtime. The only cross-fragment `.created` consumer, `atomize/aggregate.py`'s gap-minute sort, is moot here because `GenericIngestor.parse` returns exactly one fragment per file.
- Side benefit: `creek ingest --since` now actually works for the generic source. `unit_is_changed` compares `parsed.timestamp` to the cutoff ("mtime-style", per its own docstring); with a wall-clock stamp every generic unit always looked newer than any cutoff, so nothing was ever skipped.

### Known limitation (deferred, tracked)

A bare `touch` — or any edit — bumps the mtime and still mints a new id, because the generic source has no ledger. Recorded in the `parse` docstring and in `docs/ingestion.md`, and filed as **#953**.

## Test plan

Gate 1 was genuinely RED first: **7 failed / 45 passed** before the implementation, **52 passed** after. The two pre-existing tests that changed are assertion *corrections* to now-known-wrong expectations, both additive (`test_parse_sets_timestamp` retains its original `isinstance` assertion); I diffed the file against `git show HEAD:tests/test_ingest_generic.py` and byte-hashed both test files before and after the implementation run to confirm nothing was removed or weakened.

New `tests/test_ingest_generic_idempotent.py`:

- **`test_reingest_of_unchanged_file_writes_no_duplicate`** — the acceptance test. Ingests a source dir **twice** through the real `run_ingest` pipeline and asserts `errors == []` **and** exactly **1** `.md` on disk after *each* run, plus a byte-identical frontmatter `id` across runs. The `errors == []` + exact-count pair is what stops it passing vacuously on `0 == 0`. Confirmed to exercise the real production dedup path: with `ledger=None`, `write_fragment_idempotent` falls through to `writer.write_fragment`, and dedup is entirely `_write_model`'s id-index lookup.
- **`test_parse_timestamp_is_stable_across_calls`** — equal `timestamp` *and* equal assembled fragment id across two parses.
- **`test_parse_timestamp_equals_file_mtime_in_utc`** — tz-stability guard (`utcoffset() == timedelta(0)`, `isoformat()` ends `+00:00`).
- **`test_changed_mtime_yields_new_id`** — pins the accepted trade-off and guards against a "fix" that simply drops `timestamp` from the hash.
- **`test_synthetic_document_falls_back_to_wall_clock`** — preserves the pathless-`RawDocument` contract, and doubles as the guard that the injected clock genuinely **advances**.
- **`test_generic_frontmatter_platform_is_a_valid_source_platform`** — `SourcePlatform(...)` validates and the fragment still routes to `01-Fragments/Unsorted/` through the real `VaultWriter`.

Clock handling: no freezegun/time_machine in this repo, so `creek.ingest.generic.datetime` is monkeypatched with an **advancing** fake (advancing, not frozen — a frozen clock would make the *buggy* code produce equal ids and destroy the RED). Only that module attribute is patched; `Ingestor.ingest`'s provenance `datetime.now()` is left alone. No test depends on real elapsed time, and no assertion depends on filesystem visit order (`rglob` is unsorted; `_discover_directory` already wraps it in `sorted()`).

Gates:
- **Gate 2:** `cd creek-tools && ./scripts/check-all.sh` → exit 0, 12/12 checks, **6351 passed**, 93.99% coverage, per-file gate green. No suppressions, no threshold changes, no skipped or deleted tests.
- **Gate 2.5:** `code-review-orchestrator` over the full diff → **CLEAN**, zero blockers.

Closes #911
Refs #953

### DIFF
--- a/creek-tools/README.md
+++ b/creek-tools/README.md
@@ -242,7 +242,7 @@ The ingestion pipeline currently ships **11** registered `Ingestor`s plus a read
 
 `claude`, `chatgpt`, `discord`, `code`, `document` (.docx / .pdf), `markdown`, `spreadsheet` (.xlsx / .csv), `presentation` (.pptx), `image` (with OCR), `substack` (newsletter exports), and `generic` (fallback for unknown text).
 
-`gdrive` is a downloader, not an ingestor — it stages files locally and dispatches each one to the appropriate ingestor by extension. The `other` enum value on `SourcePlatform` is reserved for downstream consumers (e.g. fragments synthesised from praxes) and has no parser.
+`gdrive` is a downloader, not an ingestor — it stages files locally and dispatches each one to the appropriate ingestor by extension. The `other` enum value on `SourcePlatform` is what the `generic` fallback ingestor writes, and it also covers downstream consumers (e.g. fragments synthesised from praxes).
 
 Each ingestor follows the same four-stage contract — `discover` → `parse` → `convert_to_markdown` → `generate_frontmatter` — and writes one fragment per logical unit (per chat thread, per sheet, per slide deck, per file). See [`docs/ingestion.md`](docs/ingestion.md) for which is right for which export.
 
--- a/creek-tools/creek/ingest/generic.py
+++ b/creek-tools/creek/ingest/generic.py
@@ -4,7 +4,7 @@ Handles files that are not claimed by any specialized ingestor (Markdown,
 ChatGPT, Claude, Discord).  Attempts text reading with multiple encoding
 strategies (UTF-8, UTF-16, Latin-1, chardet fallback), skips binary files,
 and routes unclassified content to ``01-Fragments/Unsorted/`` with
-``source.platform: "unknown"`` in frontmatter.
+``source.platform: "other"`` in frontmatter.
 
 Exports:
     GenericIngestor: Concrete ``Ingestor`` subclass for unclaimed files.
@@ -30,6 +30,7 @@ from creek.ingest.base import (
     file_modified_time,
     normalize_encoding,
 )
+from creek.models import SourcePlatform
 
 logger = logging.getLogger(__name__)
 
@@ -139,7 +140,7 @@ class GenericIngestor(Ingestor):
     Discovers all files whose extensions are not in ``_CLAIMED_EXTENSIONS``,
     attempts multi-encoding text decoding, skips binary files, and routes
     successfully parsed content to ``01-Fragments/Unsorted/`` with
-    ``source.platform: "unknown"``.
+    ``source.platform: "other"``.
     """
 
     def discover(self, source_path: Path) -> list[RawDocument]:
@@ -226,6 +227,30 @@ class GenericIngestor(Ingestor):
         on disk: those produce ``authored_at = None`` rather than
         raising.
 
+        Issue #911 — re-ingest identity contract. The mtime is stamped on
+        ``timestamp`` as well, because
+        :func:`creek.ingest.base.generate_fragment_id` hashes
+        ``timestamp.isoformat()``: keying it on ``datetime.now()`` made every
+        re-ingest of an *unchanged* file mint a fresh id and write a duplicate
+        fragment. Derived from the mtime, an unchanged file re-ingests to the
+        same deterministic id and the write is a no-op.
+
+        The mtime is used verbatim in UTC — never converted to
+        :data:`~creek.ingest.base.LA_TZ` — because the UTC offset string is
+        part of the hashed input. ``datetime.fromtimestamp(st_mtime, tz=UTC)``
+        is a pure function of the epoch float, so identity does not vary with
+        the host's tzdata, ``TZ`` env var, or DST state; rendering in LA would
+        bake ``-07:00``/``-08:00`` into the hash.
+
+        Known limitation: the generic source has no ingest ledger, so a bare
+        ``touch`` (or any edit) bumps the mtime and therefore mints a new
+        fragment rather than updating the existing one. Ledger-backed
+        update-in-place is tracked in issue #953.
+
+        The wall clock survives only as the fallback for the pathless /
+        synthetic case where ``stat()`` fails: there is no mtime to key on, so
+        ``timestamp`` is ``now`` and ``authored_at`` stays honestly ``None``.
+
         Args:
             raw: The raw document to parse.
 
@@ -244,8 +269,10 @@ class GenericIngestor(Ingestor):
         if not text.strip():
             return []
 
-        now = datetime.now(tz=LA_TZ)
+        # One stat, two uses: the same mtime is both the honest authored_at and
+        # the identity-bearing timestamp (issue #911). Kept in UTC verbatim.
         authored_at = _safe_file_mtime(raw.path)
+        timestamp = authored_at if authored_at is not None else datetime.now(tz=LA_TZ)
         return [
             ParsedFragment(
                 content=text,
@@ -255,7 +282,7 @@ class GenericIngestor(Ingestor):
                     "authored_at": authored_at,
                 },
                 source_path=str(raw.path),
-                timestamp=now,
+                timestamp=timestamp,
             )
         ]
 
@@ -282,11 +309,25 @@ class GenericIngestor(Ingestor):
     def generate_frontmatter(self, fragment: ParsedFragment) -> dict[str, Any]:
         """Generate YAML frontmatter for an unclaimed file.
 
-        Sets ``source.platform`` to ``"unknown"`` and routes to
-        ``01-Fragments/Unsorted/``. FEAT-031 (issue #335): emits
-        ``authored_at`` as an ISO string when the parse stage
-        captured the file's mtime; omits the key when the metadata
-        value is missing or ``None`` (honest absence).
+        Sets ``source.platform`` to :data:`~creek.models.SourcePlatform.OTHER`
+        (``"other"``) and routes to ``01-Fragments/Unsorted/``. Issue #911:
+        this previously emitted ``"unknown"``, which is not a
+        :class:`~creek.models.SourcePlatform` member at all, so
+        ``assemble_ingested_fragment`` raised a pydantic ``ValidationError``
+        that ``run_ingest`` swallowed into ``result.errors`` — generic
+        fragments never reached the vault. ``OTHER`` is the enum's fallback
+        member and maps to the same ``Unsorted`` subfolder, so the routing
+        target is unchanged.
+
+        FEAT-031 (issue #335): emits ``authored_at`` as an ISO string
+        when the parse stage captured the file's mtime; omits the key
+        when the metadata value is missing or ``None`` (honest absence).
+
+        ``created`` is the parse-stage ``timestamp``, i.e. the file's mtime in
+        UTC since #911. That is deliberate: :mod:`creek.time` documents that
+        ``created`` is not part of the authored-at precedence chain (all
+        time-bucketing routes through ``effective_authored_at``), and the other
+        file-based ingestors already key on mtime.
 
         Args:
             fragment: The parsed fragment.
@@ -299,7 +340,7 @@ class GenericIngestor(Ingestor):
             "type": "fragment",
             "title": title,
             "source": {
-                "platform": "unknown",
+                "platform": SourcePlatform.OTHER,
                 "original_file": fragment.source_path,
             },
             "created": fragment.timestamp.isoformat(),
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
Fixes an idempotency bug in `GenericIngestor.parse`: fragment identity was keyed on wall-clock `datetime.now()` instead of the already-computed file mtime, so re-ingesting an unchanged `.txt`/`.log` file minted a fresh id and wrote a duplicate fragment on every run. Along the way it fixes a latent, unrelated defect where `source.platform: "unknown"` was not a valid `SourcePlatform` enum member, causing every generic fragment to fail pydantic validation and silently never reach the vault at all.

## Strengths
- Root-cause fix at the derivation point, not a post-hoc dedup pass — consistent with prior art (#672/#673) cited in the PR body.
- The `"unknown"` → `SourcePlatform.OTHER` fix is a genuine prerequisite, not scope creep: without it, the acceptance test would pass vacuously (`0 == 0` fragments), and the PR correctly identifies and fixes this rather than letting it slide.
- UTC-verbatim timestamp handling is deliberate and well-justified (avoids baking host TZ/DST into the hashed id) and is pinned by a dedicated test (`utcoffset() == timedelta(0)`).
- New idempotency test (`test_ingest_generic_idempotent.py`) drives the real `run_ingest` pipeline end-to-end (not mocks) and asserts both `errors == []` and exact fragment counts, which is what makes it non-vacuous.
- The advancing fake clock (vs. a frozen one) is a nice detail — a frozen clock would have made the pre-fix buggy code pass by accident, defeating the RED phase.
- Docs (`docs/ingestion.md`, `README.md`) updated in the same PR to match the corrected behavior, including an honest note on the known ledger-less limitation, tracked as a separate follow-up (#953) rather than silently deferred.
- Verified independently: `file_modified_time` (`creek/ingest/base.py`) uses `datetime.fromtimestamp(st_mtime, tz=UTC)` as claimed; `SourcePlatform.OTHER` maps to `"Unsorted"` in `_PLATFORM_SUBFOLDER`; `VaultWriter.write_fragment` serializes via `model.model_dump(mode="json")` before frontmatter dump, so the enum is written as a plain string — no yaml-serialization risk from passing the enum member directly in `generate_frontmatter`.

## Security Concerns
None identified. 🟢 No user input, subprocess, or path-traversal surface touched by this change.

## Problems
None blocking.

## Code Quality
- Non-blocking: the docstring additions in `generic.py` are fairly long (multi-paragraph) for inline module docs; could be trimmed to just the load-bearing invariant (UTC-verbatim requirement) with the rest left to the PR description/docs, but this is a style preference and matches the repo's existing dense-docstring convention elsewhere in this file.
- Non-blocking: `timestamp = authored_at if authored_at is not None else datetime.now(tz=LA_TZ)` is a touch implicit about why `authored_at is None` implies "no file" — a one-line inline comment (already partially present above it) covers this adequately.

## CLAUDE.md Adherence
✅ TDD workflow followed — PR body documents genuine RED (7 failed/45 passed) before GREEN (52 passed).
✅ `./scripts/check-all.sh` reported 12/12 checks green, 93.99% coverage, per-file gate green (per PR description; CI run was still in progress at review time — `Code Quality & Testing` jobs for 3.11/3.12/3.13 and `claude-review` were `IN_PROGRESS`, `CrawDad Quality` jobs already `SUCCESS`).
✅ Commit/PR follows conventional-commit style (`fix(ingest): ...`) and issue-first workflow (`Closes #911`, `Refs #953`).
✅ No suppressions, threshold changes, or deleted tests — confirmed via diff review.

## Testing
Strong. The new test file directly exercises the real `run_ingest` pipeline (not just the `parse()` unit in isolation) for the acceptance scenario, and separately covers: timestamp stability across repeated parses, UTC-ness of the derived timestamp, the accepted trade-off that a changed mtime yields a new id, the synthetic/pathless fallback to wall clock, and frontmatter platform validity end-to-end through `VaultWriter`. Existing tests were corrected (not weakened) to match the new, previously-incorrect expectations (`"unknown"` → `"other"`), and the PR explicitly notes a byte-hash diff was done to confirm nothing was silently dropped.

## Documentation
Thorough — `docs/ingestion.md` gained a new "Generic" section and the "Idempotency" section was rewritten to explain per-ingestor timestamp derivation and the ledger-less caveat for `generic`, with a clear pointer to the follow-up issue (#953) for the known limitation. `README.md`'s one-line description of the `other` platform value was also corrected to reflect that `generic` now actually emits it.



## Verdict: LGTM

The core fix is correct and verified against the actual source (mtime derivation via the already-UTC `file_modified_time`, `SourcePlatform.OTHER` is a real StrEnum member that serializes correctly through `model_dump(mode="json")` and routes to the existing `Unsorted` subfolder via `_PLATFORM_SUBFOLDER`), the accompanying `ValidationError`/swallowed-error claim checks out against `pipeline.py`'s `except (KeyError, ValueError)`, and the new test suite genuinely exercises the real pipeline with a non-vacuous RED/GREEN contrast. No blocking issues found; only minor stylistic nits worth a mention.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5110268762 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #956, opened 2026-07-28
fix(crawdad): dispatch intents bundled with compose=true instead of dropping them

### LINKED ISSUE #915: Fix agent loop dropping intents bundled with compose=true (workflow/register silently no-op)
**Labels:** `de-slop`, `crawdad-bot`, `bug`, `priority-high`, `agent-ready`
**Detected by:** de-slopify scan (crawdad-bot) 2026-07-24
**Severity:** High
**Remediation:** fix (correctness)

## Problem
The agent loop drops **all intents that arrive in the same router response as `compose=true`** — silently. `crawdad/crawdad/loop.py:176` breaks out of the round loop *before* dispatching whenever `response.compose` is truthy:

```python
if response.compose or not response.intents:
    break
```

The router's own tests document the **canonical, well-behaved Haiku reply** for a workflow/register phrase as the intent **bundled with `compose: True` in one response** (`tests/test_router.py:311-355`, `:358-396` — both assert the intent is present *and* `response.compose is True`). When the router emits that exact shape, the `break` at loop.py:176 fires, `aggregated` stays `[]`, the dispatcher is never called, and the composer produces a generic reply as if nothing was requested. This is Family 3/0 (dropped functionality / lying-flow) — the `/crawdad workflow` and register-switch features silently no-op on a plausible, test-sanctioned model output. (Paradox routing escapes this only because `_maybe_route_paradox` re-injects a dropped `creek.save`; `run_workflow`/`activate_register` have no such fallback.)

**Current state:** the loop tests (`test_loop.py:539-551`, `:640-655`) sidestep the bug by scripting these flows as **two separate rounds** (`{intents:[X]}` then `{intents:[], compose:True}`), so the bundled shape the router tests call canonical is never exercised end-to-end.

## Evidence (corroboration)
- **Signal 1 — reproducing test (ran, failed against current code):**
  ```
  # RouterResponse(intents=[run_workflow(...)], compose=True)  -- the canonical bundled shape
  outcome.kind = composed
  runner_calls = []                       # workflow NEVER ran
  composer tool_results = []              # composer saw nothing
  AssertionError: BUG REPRODUCED: workflow never ran; runner_calls=[]
  ```
  (Driven through the real `AgentLoop.run("draft phase-transitions")` with an injected workflow runner; the runner is never called.)
- **Signal 2 — in-repo contradiction:** `tests/test_router.py:333` and `:377` fix the canonical Haiku reply as `{"intents":[<intent>], "compose": True}` and assert `response.compose is True` with the intent present; `loop.py:176` discards exactly that. The router prompt (`router.py:117-138`) also tells Haiku to "emit … intent … BEFORE setting `compose: true`", which a model routinely satisfies inside one JSON object.

## Scope
Fixes the intent-dropping in `AgentLoop.run`. Does NOT change the router prompt or the router parser. Does NOT touch paradox routing (already has its own fallback).

## Tasks
1. In `crawdad/crawdad/loop.py`, change the round-loop control so a response carrying intents is **dispatched even when `compose=true`**, then composition proceeds. Only break-without-dispatch when `response.intents` is empty.
2. Ensure the round cap and `too_deep` reset semantics are preserved (a compose-bundled round should still count and terminate the loop after dispatch, not spin).
3. Add a loop test that scripts the **single-response bundled shape** `RouterResponse(intents=[run_workflow], compose=True)` and asserts the workflow runner is invoked and the composer receives the tool result. Add the equivalent for `activate_register`.

## Acceptance Criteria
- [ ] A single router response with `intents=[crawdad.run_workflow(...)]` **and** `compose=True` invokes the workflow runner and threads its result to the composer.
- [ ] The same holds for `crawdad.activate_register` bundled with `compose=True` (register actually switches).
- [ ] Existing two-round loop tests still pass; the round-cap `too_deep` behavior is unchanged.
- [ ] All pre-commit hooks pass.

## Files to Create/Modify
| File | Action |
|------|--------|
| `crawdad/crawdad/loop.py` | Modify |
| `crawdad/tests/test_loop.py` | Modify (add bundled-shape tests) |

Geoffe-Ga:
## Summary

`AgentLoop.run` dropped **every intent that arrived in the same router response as `compose=true`** — silently. The guard at `crawdad/crawdad/loop.py:176` was one conditional making two different decisions:

```python
if response.compose or not response.intents:
    break
```

It broke out of the round loop *before* dispatching, so the dispatcher was never called, `aggregated` stayed `[]`, and the composer produced a generic reply as if nothing had been requested. `/crawdad workflow` phrasing and natural-language register switches silently no-opped.

**The dropped shape is not hypothetical.** `router.py:117-138` instructs Haiku to emit the intent *"BEFORE setting `compose: true`"* — which a model routinely satisfies inside one JSON object — and `tests/test_router.py:311-355` / `:358-396` pin exactly that bundled single object as the canonical well-behaved reply for both `run_workflow` and `activate_register`, asserting the intent is present **and** `response.compose is True`. The loop discarded precisely what the router's own tests sanctioned. Paradox routing escaped the bug only because `_maybe_route_paradox` re-injects a dropped `creek.save`; `run_workflow` and `activate_register` have no such fallback.

The guard now splits the two decisions it was conflating: an empty intent list breaks without dispatching (nothing to do), while `compose` dispatches first and breaks after. The bundled round still calls `_record_tool_round`, so its history is byte-identical to the two-round path; the `break` immediately after guarantees no second router pass and therefore no duplicate `tool` turn.

Notes on the edges:

- **`for/else` / `too_deep` untouched.** The cap is now reached only when the router returns non-empty intents with `compose=false` for every round — exactly what its log message already claimed. `max_rounds=1` with a bundled response composes rather than falling through to `too_deep` (the `break` fires at the end of iteration 0); that case has its own test, since it is where a `continue`-shaped fix would regress.
- **Complexity is net zero, deliberately.** radon counts boolean operators, so the removed `if A or B` was +2 and the two replacement conditionals are also +2. `AgentLoop.run` stays at **B (10)** — measured before and after — which is exactly the `xenon --max-absolute B` ceiling. A `_run_rounds` extraction was prepared as a contingency and proved unnecessary.
- **One intentional behavior change:** a bundled round can now surface `unknown_intent` / `mcp_unavailable` where the old code silently composed. That is the correct direction — the silent drop was the bug.
- **Docs that sanctioned the bug were corrected.** `loop.py`'s module flow diagram and `RouterResponse`'s docstring both stated the buggy contract verbatim (`compose=True or intents=[] ─► composer.compose`). That false normative statement is why the bug survived review, so both were rewritten to match the implementation.

## Test plan

Three tests added to `crawdad/tests/test_loop.py` (pure addition — **189 insertions, 0 deletions**; no existing test modified, reordered, or deleted, verified by hash across the implementation run).

Because the failure mode is a *silent* no-op, every test asserts the **effect**, never "no exception raised". All three were confirmed RED against the unfixed code first:

| Test | RED failure before the fix |
|---|---|
| `test_loop_dispatches_run_workflow_bundled_with_compose` | `assert runner_calls == [...]` → `AssertionError: assert [] == [('phase-transitions', {'topic': 'spirals'})]` — the workflow runner was never called |
| `test_loop_activate_register_bundled_with_compose_switches_stack` | `assert registry.stack is not initial` → failed; the registry was never mutated, so the register never switched |
| `test_loop_bundled_compose_round_composes_at_max_rounds_one` | `assert runner_calls == [...]` → `[]`, at `max_rounds=1` |

The third test is the reason effect-assertions matter: `outcome.kind == "composed"` **passed against the live bug** on the line directly above the failure. Asserting the call survived would have proved nothing.

Supporting assertions: `len(router.calls) == 1` (the bundled round costs no extra router pass), the composer receives exactly one `ToolResult` with the right `intent_type` and `body`, a non-`user`/`assistant` history turn carries that body (history parity with the two-round path), and `composer.calls[0]["skills"] is registry.stack` (the composer got the switched-*to* stack).

Gate results, run the way the **CrawDad Quality** CI job does (`cd crawdad && ./scripts/check-all.sh` — note the `creek-tools` gate does *not* cover this subproject):

- **Baseline on `origin/main`:** exit 0, 507 passed, 94.57% branch coverage.
- **After the change:** exit 0, all 7 checks green, **510 passed** (507 + 3 new), **94.58%** branch coverage.
- Both arms of the new `if response.compose:` are covered — the `True` side by the three new tests, the `False` side by the existing multi-round aggregation and cap tests.

Gate 2.5 self-review returned **CLEAN** with no blockers.

Closes #915

### DIFF
--- a/crawdad/crawdad/intents.py
+++ b/crawdad/crawdad/intents.py
@@ -64,11 +64,16 @@ class Intent(BaseModel):
 class RouterResponse(BaseModel):
     """The exact JSON shape Haiku must emit — no prose, no extras.
 
-    FEAT-015 adds the ``compose`` flag: when ``True`` (or when
-    ``intents`` is empty) the loop terminates router iteration and
-    invokes the Sonnet composer. The pair ``{intents: [], compose:
-    true}`` is the canonical "I have enough context; compose now"
-    signal.
+    FEAT-015 adds the ``compose`` flag: it ends router iteration, so the
+    Sonnet composer produces the reply for this round. An empty
+    ``intents`` list ends iteration too. The pair ``{intents: [],
+    compose: true}`` is the canonical "I have enough context; compose
+    now" signal.
+
+    ``compose`` and ``intents`` are independent: a response may carry
+    both, and the loop then dispatches those intents *before* composing
+    rather than discarding them (#915). The router prompt invites this
+    shape by asking for the intent "BEFORE setting ``compose: true``".
     """
 
     model_config = ConfigDict(frozen=True)
--- a/crawdad/crawdad/loop.py
+++ b/crawdad/crawdad/loop.py
@@ -10,7 +10,7 @@ Sonnet composer. The loop's contract:
        ▼  ── ROUND N ──┐
     router.extract_intents
        │
-       ├─ compose=True or intents=[] ─► composer.compose ─► reply
+       ├─ intents=[] ─► composer.compose ─► reply
        │
        └─ intents non-empty
               │
@@ -19,13 +19,23 @@ Sonnet composer. The loop's contract:
               │
               ▼
           dispatcher.dispatch(intents)  →  aggregate ToolResults
+              │                            + record the ``tool`` turn
               │
-              ▼
+              ├─ compose=True ─► composer.compose ─► reply
+              │
+              └─ compose=False ─► ROUND N+1 ──┘
        (loop until ``max_rounds`` rounds; the next attempt is refused)
+
+``compose`` is read *after* the dispatch, not before it: the router is
+prompted to emit its intents and set ``compose: true`` in the same
+response once it has everything it needs, so a bundled round must run
+the tools it asked for and then compose within that same round (#915).
+Only an *empty* intent list short-circuits straight to the composer.
 
 The cap defaults to :data:`MAX_LOOP_ROUNDS` (5) and can be raised or
 lowered per deployment via the ``max_loop_rounds`` key in
-``crawdad.yaml`` (FEAT-036, bounded ``[1, 50]``).
+``crawdad.yaml`` (FEAT-036, bounded ``[1, 50]``). The cap is reached
+only when the router keeps returning intents with ``compose=false``.
 
 Side-effects: the loop appends ``user`` and ``assistant`` turns to the
 shared :class:`ConversationHistory` (the source the next router pass
@@ -173,7 +183,8 @@ class AgentLoop:
                 _LOGGER.warning("router parse error mid-loop: %s", exc)
                 return LoopOutcome(kind="router_parse_error", reply=_ROUTER_PARSE_REPLY)
 
-            if response.compose or not response.intents:
+            # Nothing to dispatch: go straight to the composer.
+            if not response.intents:
                 break
 
             try:
@@ -186,6 +197,14 @@ class AgentLoop:
                 return LoopOutcome(kind="mcp_unavailable", reply=_MCP_UNAVAILABLE_REPLY)
             aggregated.extend(results)
             self._record_tool_round(results)
+
+            # Intents bundled with compose=true (#915): the tools above
+            # have now run, so this round composes rather than costing an
+            # extra router pass. Recording the round above keeps history
+            # identical to the two-round path — the break below guarantees
+            # no second router pass, so there is no duplicate turn.
+            if response.compose:
+                break
         else:
             # Exhausted max_rounds without the router setting compose=true.
             _LOGGER.warning(
--- a/crawdad/tests/test_loop.py
+++ b/crawdad/tests/test_loop.py
@@ -675,6 +675,195 @@ async def test_loop_run_workflow_intent_invokes_workflow_runner(
     assert tool_results[0].body == "workflow phase-transitions done"
 
 
+async def test_loop_dispatches_run_workflow_bundled_with_compose(
+    state: SessionState, skills: VoiceSkillStack
+) -> None:
+    """#915: intents bundled with ``compose=true`` must still be dispatched.
+
+    The router prompt tells Haiku to emit the intent *before* setting
+    ``compose: true``, and ``test_router.py`` pins that bundled single
+    object as the canonical well-behaved reply. The loop must therefore
+    dispatch this round's intents and *then* compose, rather than
+    breaking out before dispatch and silently dropping the work.
+    """
+    from crawdad.intents import RUN_WORKFLOW_INTENT_TYPE
+
+    runner_calls: list[tuple[str, dict[str, str]]] = []
+
+    async def _workflow_runner(name: str, inputs: dict[str, str]) -> str:
+        runner_calls.append((name, inputs))
+        return f"workflow {name} done"
+
+    router = _ScriptedRouter(
+        [
+            RouterResponse(
+                intents=[
+                    Intent(
+                        type=RUN_WORKFLOW_INTENT_TYPE,
+                        args={
+                            "name": "phase-transitions",
+                            "inputs": {"topic": "spirals"},
+                        },
+                    )
+                ],
+                compose=True,
+            )
+        ]
+    )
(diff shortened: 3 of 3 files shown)


Geoffe-Ga:
## Summary
Fixes #915: `AgentLoop.run` previously broke out of the round loop on `response.compose or not response.intents`, which discarded any intents bundled with `compose=true` in the same router response before they were ever dispatched — exactly the shape the router prompt asks Haiku to produce and that `test_router.py` pins as canonical. The fix splits the single conditional into two: an empty intent list still short-circuits to the composer, but a non-empty list with `compose=true` now dispatches first and composes in the same round.

## Strengths
- Root-cause fix, not a workaround: the guard is restructured to match the two decisions it was conflating, rather than special-casing around the bug.
- `_record_tool_round` still runs for the bundled round, so history stays identical between the one-round and two-round paths — verified directly by a test assertion on `composer_history`.
- Tests assert effects (workflow runner actually called, registry stack actually swapped, composer received the switched-to stack), not just "no exception" — and the PR description confirms all three were RED against the unfixed code first.
- The `max_rounds=1` test specifically guards against a plausible alternative fix (`continue` instead of `break` after dispatch) that would regress into `too_deep` instead of composing — a good adversarial test to include.
- Module docstring flow diagram and `RouterResponse` docstring, which both stated the buggy contract verbatim, were corrected to match the new implementation — removes a latent source of future regressions.
- Documented, deliberate behavior change (bundled rounds can now surface `unknown_intent`/`mcp_unavailable` instead of silently composing) is called out explicitly rather than left implicit.

## Security Concerns
None identified. 🟢 No new external input handling, no changes to the trust boundary, subprocess invocation, or MCP dispatch surface beyond reordering existing calls.

## Problems
None found. The `for/else` cap logic, paradox auto-routing, and `_current_skills()`/registry read-after-dispatch ordering all remain correct under the new control flow (verified by reading `loop.py:170-249`).

## Code Quality
- Nothing blocking. The two-line comment additions above the new `if response.compose:` check are a touch verbose but genuinely clarify non-obvious history/round-count invariants, consistent with the file's existing comment density.

## CLAUDE.md Adherence
- ✅ TDD: tests added and confirmed RED before the fix per the PR description.
- ✅ Complexity: PR states `AgentLoop.run` measured at Xenon `B (10)` before and after, at the `--max-absolute B` ceiling — worth a quick independent confirmation in CI but not blocking given the explicit before/after claim.
- ✅ Coverage: both branches of `if response.compose:` are exercised (new tests for `True`, existing multi-round/cap tests for `False`).
- ✅ Docs: docstrings that sanctioned the old (buggy) contract were corrected to match implementation.
- Could not independently execute `./scripts/check-all.sh` or `gh pr checks` in this environment (sandboxed, no tool approval available) — relying on the PR description's reported gate results (510 passed, 94.58% branch coverage, 7/7 checks green) and static reading of the diff.

## Testing
Three new tests in `crawdad/tests/test_loop.py`, pure addition (189 insertions, 0 deletions per PR description). They cover: bundled `run_workflow` dispatch, bundled `activate_register` dispatch (including the subtler "old voice stack" failure mode), and the `max_rounds=1` edge case. Assertions target effects (runner calls, registry mutation, composer inputs, history contents) rather than absence-of-exception, which is the right shape for a silent-no-op bug. Not blocking.

## Documentation
Both docstrings touched by the change (`loop.py` module flow diagram, `RouterResponse` class docstring) were rewritten to reflect the fixed behavior. No further documentation gaps identified.



## Verdict: LGTM

The fix correctly separates the "nothing to dispatch" and "compose after dispatch" decisions, is exercised by three new effect-asserting tests (including the max_rounds=1 edge case that would catch a continue-shaped regression), and both touched docstrings were corrected to match the new implementation — no blocking issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5110575256 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #957, opened 2026-07-28
fix(crawdad): stop false MIME mismatch when a UTF-8 char straddles the 1024-byte sample

### LINKED ISSUE #916: Fix false MIME mismatch when a UTF-8 multibyte char straddles the 1024-byte text sample
**Labels:** `de-slop`, `crawdad-bot`, `bug`, `priority-high`, `agent-ready`
**Detected by:** de-slopify scan (crawdad-bot) 2026-07-24
**Severity:** High
**Remediation:** fix (correctness)

## Problem
`_verify_text_sample` (`crawdad/crawdad/attachments.py:262-276`) falsely reports a **valid UTF-8 text file as a MIME mismatch** when a multi-byte UTF-8 character straddles the 1024-byte sample boundary. It slices a fixed byte prefix and strict-decodes it:

```python
sample = data[:_TEXT_SAMPLE_BYTES]          # _TEXT_SAMPLE_BYTES = 1024
if b"\x00" in sample or not _decodes_as_utf8(sample):   # strict decode, no codepoint-boundary handling
```

`_decodes_as_utf8` calls `sample.decode("utf-8")` with no `errors=` and no incremental/partial-codepoint handling (attachments.py:279-285). A byte cut at offset 1024 can split a 2–4 byte UTF-8 sequence, so a perfectly valid `.md`/`.txt` file is classified `mismatch`. This is Family 0 (boundary/off-by-one): with default config the user sees a spurious "⚠ MIME mismatch" warning on a valid file; with `reject_on_mime_mismatch=True` the valid file is **hard-rejected** at the gate (attachments.py:590-601).

**Current state:** every text-match test uses short ASCII payloads far under 1024 bytes (`test_attachments.py` around `test_verify_mime_type_matched_text_sample_returns_match`), so the boundary case is never exercised.

## Evidence (corroboration)
- **Signal 1 — reproducing test (ran, failed against current code):**
  ```
  # payload = b"a"*1023 + "é".encode() + b" more perfectly valid text\n"   (valid UTF-8)
  _TEXT_SAMPLE_BYTES = 1024
  status = mismatch | detected = application/octet-stream | expected = text/markdown
  AssertionError: BUG REPRODUCED: valid UTF-8 .md flagged as mismatch
  ```
- **Signal 2 — reading:** the fixed slice at attachments.py:265 followed by strict `decode("utf-8")` at :282 has no codepoint-boundary handling. Python's `codecs.getincrementaldecoder("utf-8")` exists precisely to decode a partial byte stream without failing on a truncated trailing sequence.

## Scope
Fixes the false-negative in the text-sample UTF-8 check. Does NOT change the NUL-byte binary heuristic, the `filetype`-based path for non-text extensions, or the reject/warn policy wiring.

## Tasks
1. In `_verify_text_sample`, decode the sample tolerantly at the boundary — e.g. use an incremental UTF-8 decoder, or decode `data[:_TEXT_SAMPLE_BYTES]` while ignoring only a truncated trailing multi-byte sequence — so a validly-encoded file whose 1024th byte splits a codepoint is treated as `match`. Keep rejecting genuinely invalid UTF-8 and NUL-containing samples.
2. Add tests covering: (a) a 2-byte char straddling byte 1024 → `match`; (b) a genuinely invalid UTF-8 byte sequence within the sample → `mismatch`; (c) a NUL byte within the sample → `mismatch`.

## Acceptance Criteria
- [ ] `verify_mime_type("notes.md", b"a"*1023 + "é".encode("utf-8") + b" ...")` returns `status="match"`.
- [ ] Genuinely invalid UTF-8 and NUL-containing samples still return `status="mismatch"`.
- [ ] With `reject_on_mime_mismatch=True`, a valid boundary-straddling text file is no longer hard-rejected.
- [ ] All pre-commit hooks pass.

## Files to Create/Modify
| File | Action |
|------|--------|
| `crawdad/crawdad/attachments.py` | Modify (`_verify_text_sample` / `_decodes_as_utf8`) |
| `crawdad/tests/test_attachments.py` | Modify (add boundary tests) |

Geoffe-Ga:
## Summary

`_verify_text_sample` sliced `data[:_TEXT_SAMPLE_BYTES]` (1024) and strict-decoded it. A multibyte UTF-8 codepoint straddling the window edge raised `UnicodeDecodeError`, which the verifier misread as "this is a binary blob" — so a perfectly valid `.md`/`.txt` came back `status="mismatch"`, `detected_mime="application/octet-stream"`. By default the user gets a spurious "MIME mismatch" warning on a good file; under `AttachmentConfig(reject_on_mime_mismatch=True)` the file is **hard-rejected at the gate** and never lands.

**Confirmed broader than filed.** The issue cites one `é` case; all **six** straddle positions reproduce — 2-byte, 3-byte CJK at both split offsets, and 4-byte emoji at all three. Since the cut is a byte cut and not a character cut, a CJK-heavy note (3 bytes/char) has roughly a **2/3** chance of landing mid-character, so a Japanese/Chinese/Korean journal attachment is more likely than not to be falsely flagged.

### The fix

Decode the sample with `codecs.getincrementaldecoder("utf-8")` and pass `final=not truncated`, where `truncated = len(data) > _TEXT_SAMPLE_BYTES`.

The strict `>` is load-bearing: at exactly 1024 bytes the sample *is* the whole file, so nothing was cut and a dangling partial there is genuine corruption that must still mismatch. `final=not truncated` collapses both modes into a single call with zero added branches — both helpers stay at cyclomatic complexity A.

**Invariant:** the sample is `match` iff (a) it holds no NUL byte anywhere in the 1024-byte window, and (b) every byte decodes as valid UTF-8 except — when and only when the body actually runs past the window — a trailing run of at most 3 bytes forming a legal prefix of one unfinished sequence.

This is a bounded allowance, not a blanket "ignore the tail". The decoder raises on invalid leads, bare or malformed continuations, and overlong or out-of-range forms wherever the disqualifying byte falls inside the window. The NUL scan still runs **first**, over all 1024 raw bytes, independent of decoding — NUL bytes decode as valid UTF-8, so reordering that check would have been a silent security regression.

### Security review

This deliberately loosens a content-sniffing validator, so it got a dedicated adversarial pass: **CLEAN**, no blockers. The strongest polyglot enabled is a blob with 1021 bytes of valid NUL-free UTF-8 followed by a legal lead byte at offsets 1021-1023. That buys zero payload capacity — the region past byte 1024 was never inspected before this change either — so it sits inside the false-negative surface ADR-0002 already documents rather than opening a new one. The effective validated prefix is 1021 bytes worst-case; ADR-0002 now records this.

Empirically verified against CPython's decoder: `\xc0\xaf`, `\xc0\x80` (overlongs), `\xed\xa0\x80` (surrogate), `\xf5…` (out of range), bare continuations and `\xfe\xff` are all **rejected**. A trailing `\xed` or `\xed\xa0` is tolerated, because the byte that would disqualify it lies past the cut and was never inspected — documented precisely in the docstring rather than glossed over.

## Test plan

- **RED proven before implementing:** 9 failed / 53 passed, `AssertionError: assert 'mismatch' == 'match'`.
- Parametrized straddle test over all six positions (2/3/4-byte at every split), each carrying self-check assertions that the codepoint *actually* straddles byte 1024 and that the payload is valid UTF-8 end to end — so a subtly-wrong fixture fails loudly instead of passing for the wrong reason.
- The acceptance criterion spelled out verbatim so it stays greppable against the issue.
- Eight negative tests, each pinning one invariant and each killing a specific weaker implementation: `1024 -> mismatch` / `1025 -> match` as an adjacent pair (the `>` vs `>=` off-by-one); `\xff\xfe` in the last 2 bytes (kills "shave 3 trailing bytes and re-decode"); `\xe6\x28` at the boundary (legal lead, illegal continuation — kills "forgive any tail"); NUL at offset 1023 (kills "shrink the NUL window"); whole-file truncated tail; invalid bytes mid-sample.
- Integration test through `process_attachments` with `reject_on_mime_mismatch=True` proving a valid boundary-straddling file is accepted and written to staging (acceptance criterion 3).
- Test diff is **+236/-0** — purely additive, byte-verified against `HEAD` at every handoff. No existing test touched.

**Gates (both subprojects, since the diff spans `crawdad/` and a `creek-tools/` ADR):**

| gate | result |
|---|---|
| `crawdad/scripts/check-all.sh` | exit 0, 7/7 checks, **526 passed**, **94.59%** coverage (baseline 510 / 94.58%) |
| `creek-tools/scripts/check-all.sh` | exit 0, 12/12 checks, **6351 passed**, **93.99%** coverage |

No `# noqa`, no `# type: ignore`, no skips, no lowered thresholds, no deleted tests, no new dependency (`codecs` is stdlib).

Closes #916

### DIFF
--- a/crawdad/crawdad/attachments.py
+++ b/crawdad/crawdad/attachments.py
@@ -26,6 +26,7 @@ before any ``creek.ingest`` call.
 
 from __future__ import annotations
 
+import codecs
 import hashlib
 import logging
 import re
@@ -152,7 +153,9 @@ _EXTENSION_MIME_SPECS: dict[str, _ExtensionMimeSpec] = {
 # in this set are expected to be plain text, so a NUL byte or undecodable
 # bytes in the first ``_TEXT_SAMPLE_BYTES`` of the body are treated as
 # evidence that the file is actually a binary blob renamed to a text
-# extension (the polyglot case the issue calls out). The dict carries
+# extension (the polyglot case the issue calls out). One byte pattern is
+# *not* such evidence: a codepoint the sampling itself cut in half at the
+# window edge — see ``_TEXT_SAMPLE_BYTES`` and issue #916. The dict carries
 # the canonical text MIME per extension so a successful match surfaces
 # the right label (``text/markdown`` for ``.md``, ``application/json``
 # for ``.json``, etc.) instead of a generic ``text/plain`` blanket.
@@ -174,6 +177,13 @@ _TEXT_EXTENSIONS: frozenset[str] = frozenset(_TEXT_EXTENSION_TO_MIME)
 # Sample size for the text-content heuristic. One kilobyte is enough to
 # catch every common binary header (PE / ELF / OLE / PDF / ZIP) without
 # materially extending the download path's CPU cost.
+#
+# The window is a byte cut, not a character cut, so it can land in the
+# middle of a multibyte codepoint. Issue #916 pins the resulting edge
+# semantics: the sample is text iff (a) it holds no NUL byte anywhere in
+# the window, and (b) every byte decodes as UTF-8 except — when and only
+# when the body actually runs past the window — a trailing run of at most
+# three bytes forming a legal prefix of one unfinished sequence.
 _TEXT_SAMPLE_BYTES: int = 1024
 
 # Canonical MIME the text verifier uses when it has to label something
@@ -260,10 +270,24 @@ def _verify_binary_signature(suffix: str, data: bytes) -> MimeVerification:
 
 
 def _verify_text_sample(suffix: str, data: bytes) -> MimeVerification:
-    """Sample-check that *data* is plausibly UTF-8 text with no NUL bytes."""
+    """Sample-check that *data* is plausibly UTF-8 text with no NUL bytes.
+
+    The sample is a fixed-width byte window, so a multibyte codepoint can
+    straddle its edge. When — and only when — *data* actually extends past
+    the window, an unfinished-but-well-formed trailing sequence is
+    tolerated rather than read as binary (issue #916). A body that ends
+    at or before the window was never cut by us, so a dangling partial
+    there is genuine corruption and still reads as a mismatch.
+    """
     expected = _TEXT_EXTENSION_TO_MIME[suffix]
     sample = data[:_TEXT_SAMPLE_BYTES]
-    if b"\x00" in sample or not _decodes_as_utf8(sample):
+    # Strict ``>``: at exactly _TEXT_SAMPLE_BYTES the sample *is* the whole
+    # file, so nothing was cut off and no tail may be forgiven.
+    truncated = len(data) > _TEXT_SAMPLE_BYTES
+    # NUL check first, over the full raw window: NUL bytes decode as valid
+    # UTF-8, so this heuristic is independent of (and unweakened by) the
+    # truncation tolerance below.
+    if b"\x00" in sample or not _decodes_as_utf8(sample, truncated=truncated):
         return MimeVerification(
             status="mismatch",
             detected_mime=_BINARY_CANONICAL_MIME,
@@ -276,10 +300,44 @@ def _verify_text_sample(suffix: str, data: bytes) -> MimeVerification:
     )
 
 
-def _decodes_as_utf8(sample: bytes) -> bool:
-    """Return ``True`` when *sample* is fully decodable as UTF-8."""
+def _decodes_as_utf8(sample: bytes, *, truncated: bool) -> bool:
+    """Return ``True`` when *sample* decodes as UTF-8, tail cut allowance aside.
+
+    Args:
+        sample: The bytes to decode — the leading window of a larger body.
+        truncated: ``True`` when *sample* was cut out of a longer body, so
+            its final codepoint may legitimately be incomplete. ``False``
+            demands a complete decode, matching a plain
+            ``sample.decode("utf-8")``.
+
+    Returns:
+        ``True`` when every byte decodes as UTF-8, except that a
+        *truncated* sample may end in a trailing run of **at most three**
+        bytes forming a legal prefix of one unfinished sequence — issue
+        #916, where a codepoint straddling the window edge was misread as
+        a binary blob. This is a bounded allowance, not a blanket "ignore
+        the tail": the incremental decoder still raises on the first byte
+        that cannot begin or continue a valid sequence, so invalid leads,
+        bare or malformed continuations, and overlong or out-of-range
+        forms are rejected wherever the byte that disqualifies them falls
+        inside the window.
+
+        The allowance is therefore bounded by what the window can see. A
+        partial sequence whose disqualifying byte lies *past* the cut is
+        tolerated, because that byte was never inspected — a trailing
+        ``\\xed\\xa0`` passes even though it can only continue into a
+        surrogate, and a bare trailing ``\\xed`` passes too, since one
+        lead byte alone cannot yet be range-checked. The effective
+        validated prefix is
+        ``_TEXT_SAMPLE_BYTES - 3`` bytes in the worst case, which is why
+        this sits inside the false-negative surface ADR-0002 already
+        documents rather than opening a new one.
+    """
+    decoder = codecs.getincrementaldecoder("utf-8")()
     try:
-        sample.decode("utf-8")
+        # ``final=True`` flushes the decoder, so a dangling partial raises;
+        # ``final=False`` lets a legal unfinished tail stay buffered.
+        decoder.decode(sample, final=not truncated)
     except UnicodeDecodeError:
         return False
     return True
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
## Summary
Fixes issue #916: `_verify_text_sample` sliced the first 1024 bytes of an attachment and strict-decoded it as UTF-8; a multibyte codepoint straddling that byte boundary raised `UnicodeDecodeError`, which was misread as "this is binary" and produced a spurious (or, under `reject_on_mime_mismatch=True`, hard-blocking) MIME mismatch on perfectly valid `.md`/`.txt` files. The fix swaps the strict `bytes.decode("utf-8")` call for `codecs.getincrementaldecoder("utf-8")` with `final=not truncated`, so a legal-but-unfinished trailing sequence is tolerated only when the sample was actually cut out of a longer body.

## Strengths
- The `truncated = len(data) > _TEXT_SAMPLE_BYTES` / `final=not truncated` collapse is the right fix: one incremental-decoder call replaces the previous strict decode, with no added branching, and it correctly refuses to forgive a dangling partial when the body ends exactly at the window (still genuine corruption).
- NUL-byte scan is deliberately left running first, over the full raw sample, independent of the decode step — correctly preserves the existing security invariant (NUL bytes decode as valid UTF-8, so reordering would have silently weakened the polyglot check).
- Test coverage is exhaustive and adversarial: all six straddle positions (2/3/4-byte codepoints at every split), the `1024 -> mismatch` / `1025 -> match` off-by-one pair, malformed-continuation-at-boundary, invalid-bytes-far-from-boundary, and two NUL-placement invariants (including NUL at the very last sampled byte). Several tests are explicitly designed to kill specific weaker/incorrect implementations (e.g., "shave trailing bytes and re-decode"), which is good defensive test design.
- An end-to-end integration test through `process_attachments(reject_on_mime_mismatch=True)` proves the fix actually prevents the hard-rejection path, not just the low-level helper.
- ADR-0002 is updated in the same PR to document the new carve-out and its bounded false-negative surface (`_TEXT_SAMPLE_BYTES - 3` worst case), keeping design docs in sync with the code.
- Docstrings are unusually precise about the exact invariant being enforced, which will help future maintainers reason about edge cases without re-deriving the byte arithmetic.

## Security Concerns
🟢 LOW — This is a content-sniffing validator being deliberately loosened, so it deserves scrutiny, but the loosening is tightly bounded: only the last ≤3 bytes of the 1024-byte window can go unvalidated, and only when the body extends past the window (bytes past 1024 were never inspected before this change either, so no new surface is opened). The PR body documents a dedicated adversarial pass against overlongs, surrogates, out-of-range sequences, and bare continuations, all confirmed rejected. No issues found on review of the logic itself.

## Problems
None found. Logic was traced by hand against the incremental-decoder semantics (`final=False` buffers a legal partial trailing sequence without raising; both `final=True` and any mid-stream invalid byte still raise regardless of `final`), and it matches every test's expected outcome.

## Code Quality
- Non-blocking: the docstring on `_decodes_as_utf8` is quite long (mirrors the ADR content). Not a problem, but could eventually be trimmed to a pointer at the ADR/issue if it starts drifting from the code.
- Non-blocking: `truncated` is recomputed as `len(data) > _TEXT_SAMPLE_BYTES` in `_verify_text_sample` rather than being derivable from `sample` alone — worth a passing glance if this helper is ever reused with a pre-sliced buffer where the original `data` length isn't available, but that's a hypothetical, not a current issue.

## CLAUDE.md Adherence
✅ Uses project scripts per crawdad/CLAUDE.md convention (PR body reports `crawdad/scripts/check-all.sh` and `creek-tools/scripts/check-all.sh` both exit 0).
✅ No `# noqa`, `# type: ignore`, or skipped checks — matches the "no shortcuts" rule.
✅ New parameter (`truncated: bool`) is keyword-only and fully typed, consistent with MyPy-strict requirements.
✅ Coverage reported above the ≥90% branch-coverage gate (94.59% crawdad, 93.99% creek-tools) and docstring coverage requirement is met via the added docstrings.
✅ Conventional-commit-style PR title (`fix(crawdad): ...`) and commit message, matching repo convention.
✅ No new dependency — `codecs` is stdlib.

## Testing
Strong. TDD was followed (RED proven per the PR description: 9 failing / 53 passing before the fix), tests are purely additive (+236/-0, no existing test touched), and they cover the positive case (all six straddle splits), the off-by-one boundary pair, and multiple negative cases specifically designed to catch under- or over-corrected implementations. This is well above the bar needed to merge.

## Documentation
ADR-0002 was updated in the same PR to describe the new carve-out, its rationale, and its bounded false-negative surface, and both the `_TEXT_SAMPLE_BYTES` module comment and the two touched function docstrings were expanded to state the exact invariant. Documentation is complete and consistent with the code.



## Verdict: LGTM

The fix is narrowly scoped, the incremental-decoder logic is correct (verified by tracing the truncated/final semantics against every test case), the NUL-check-before-decode ordering preserves the existing security property, and the PR's own reported numbers (check-all.sh green in both subprojects, coverage above threshold, RED-then-GREEN TDD) line up with what the diff actually contains. Nothing here should block merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5110985242 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #960, opened 2026-07-29
fix(vault): re-derive privacy_tier on material fragment rewrite (#922)

### LINKED ISSUE #922: [scan:bugs] _clear_classification keeps a stale privacy_tier across a material fragment rewrite
Found during the #846 security review (PR #921), out of scope there.

## Evidence
`creek-tools/creek/vault/writer.py` `_clear_classification` (~L407-425) drops `classification_method`, `classified_at` and `classification_reasoning` when a fragment's body is materially rewritten on re-ingest — but it **keeps `privacy_tier`**.

So a fragment whose content was replaced retains its old, possibly lower, tier until something re-classifies it. Every read-side ceiling check keys on that persisted `privacy_tier`: `creek_mcp/tools/reflect.py` `_fragment_tier` -> `_above_ceiling` (the gate added in #846), `creek/classify/privacy_filter.tier_of`, and `creek_mcp/tier_ceiling.tier_allowed`. A body that is now intimate but still carries `privacy_tier: open` is admitted to an OPEN-ceiling caller — including a remote one — straight through the new gate.

The classification-provenance fields are cleared precisely because they no longer describe the new body. `privacy_tier` is derived from the same stale classification and should be treated the same way.

## Goal
Decide and implement the correct stale-tier semantic on material rewrite. Options, in rough order of safety:
1. Reset `privacy_tier` to the most restrictive tier until re-classification (fail closed; may over-refuse until the next classify pass).
2. Reset to `unclassified` (consistent with the other cleared provenance fields, but `unclassified` ranks as *open* system-wide, so this is a no-op for exposure — see the related policy note below).
3. Keep the tier but gate reads on `classified_at` being present and consistent with the current body.

Whichever is chosen, pin it with a test that rewrites a fragment body and asserts the read-side ceiling behaviour.

## Related
- The `unclassified`-ranks-as-open policy (`creek_mcp/tier_ceiling.py` `_TIER_RANK`, `creek/classify/privacy_filter.py`) is the reason option 2 does not help; that policy is deliberate and is a separate, larger architectural question.
- #846 / PR #921 added the read-side gate this stale tier defeats.
- Re-verify the finding against HEAD before starting; close with a comment if it no longer reproduces.

Geoffe-Ga:
## Summary

`VaultWriter.update_fragment` — the idempotent in-place re-ingest path — cleared a fragment's classification provenance keys on a material rewrite but left `privacy_tier` describing the body that had just been overwritten. Every read-side ceiling check keys on that persisted tier (`creek_mcp/tools/reflect.py::_fragment_tier` → `_above_ceiling`, `creek/classify/privacy_filter.py::tier_of`, `creek_mcp/tier_ceiling.py::tier_allowed`), so a fragment rewritten from an essay into intimate content but still carrying `privacy_tier: open` was admitted straight through the #846 gate to an OPEN-ceiling caller, including a remote one.

The fix re-derives the tier **from the new body** on the material branch, merges it **escalate-only** against the tier on disk, and refreshes the derived `voice_proxy_eligible` key.

### Which of the issue's three options — and why none of them, quite

The issue offered three options and steered toward option 1 (reset to the most restrictive tier). I verified both obvious options are traps by executing the relevant code in this worktree, not just reading it:

**Option 1 (blanket `privacy_tier: intimate`) is not "over-refuse until the next classify pass" — it is permanent and irreversible.**
- `privacy_pass.needs_tier` returns `False` for any explicit non-`unclassified` tier, so no ordinary classify pass ever revisits it (`needs_tier({"privacy_tier": "intimate"})` → `False`).
- Even `creek classify --force` merges through `privacy_pass.escalate`, which never lowers a tier: `escalate(INTIMATE, OPEN)` → `intimate`.
- Nothing in the package downgrades `privacy_tier` — `creek/purge/engine.py`'s `_CLASSIFICATION_RESET_FIELDS` explicitly excludes it.

So a blanket reset would bury **every materially edited fragment at `intimate` forever**, recoverable only by hand-editing vault YAML. On a continuous-ingest vault where journal entries get edited routinely, that progressively buries the corpus. That is a worse bug than the one being fixed, and it is *not* contingent on classifier availability — the ratchet, not the classifier, is what traps it.

**Option 2 (reset to `unclassified`) is a genuine no-op for exposure.** `creek_mcp/tier_ceiling.py::_TIER_RANK` maps `UNCLASSIFIED` to `0`, identical to `OPEN`; `tier_allowed(UNCLASSIFIED, TierCeiling.OPEN)` returns `True`. Same trap as #876.

**Option 3** (gate reads on `classified_at` consistency) spreads the fix across every read site instead of fixing it at the one write site that creates the staleness.

**What this PR does instead:** re-derive the tier from the new body using the existing `PrivacyClassifier.classify_tier`, then merge escalate-only against the on-disk tier. This is the only option that is both fail-closed *and* self-healing:
- the tier always describes the **current** body, so there is no staleness window at all;
- it costs **no LLM call** — `PrivacyClassifier` is pure keyword/platform heuristics, no network, no cloud consent, no Ollama;
- escalate-only preserves operator curation (a manual `intimate` survives a benign rewrite);
- crucially, it does **not** pessimistically bury benign edits, so there is no availability loss.

### On the availability cost the issue asked about

Because the tier is derived rather than blanket-stamped, the permanent-invisibility failure mode does not arise for benign rewrites — a rewritten essay stays `open` and stays visible (pinned by `test_rewrite_into_benign_content_stays_open`). Re-tiering needs no classifier infrastructure, so a host with no cloud consent and no reachable Ollama behaves identically.

One residual, inherited from repo-wide policy rather than introduced here: the escalate-only merge means a fragment that was legitimately `intimate` and is later rewritten into genuinely open content stays `intimate` until an operator edits the frontmatter. That is the same never-auto-downgrade rule `privacy_pass` already enforces everywhere, and lowering is the one direction that leaks.

### Deliberately out of scope

`creek_mcp/tier_ceiling.py::_TIER_RANK[UNCLASSIFIED] = 0` diverges from `creek/classify/privacy_filter.py::_TIER_RANK[UNCLASSIFIED] = 1`, which #876 raised to rank with `PERSONAL` on the reader side. The MCP ceiling is also a reader-side gate, so this looks like a #876 straggler rather than a deliberate policy — but the issue explicitly scopes it out, and it is pinned by deliberate tests (`tests/test_mcp_tier_ceiling.py:51`, `tests/test_mcp_reflect.py:698`, `tests/test_mcp_write_tools.py:1205`). Filed separately; not touched here.

## Changes

- `creek/vault/writer.py` — new `_tier_on_disk` (fail-closed frontmatter tier parse, mirroring `tier_of`/`_fragment_tier`) and `_retier_after_rewrite`, called on the material branch only. Deferred imports, matching the load-time cycle already documented on `_clear_classification` (`classify_engine` imports `VaultWriter`). The escalate base is the **on-disk** tier, never `tier_of(fragment)` — the ingest pipeline hands over a freshly built `Fragment` still carrying the model's `unclassified` default, which ranks *below* `open` in `_ESCALATION_RANK`, so escalating against it would silently discard an operator's `intimate`.
- `docs/idempotent-ingest.md` — documents the re-derivation, the LLM-free heuristic, and the escalate-only merge.

## Test plan

7 new tests in `tests/test_ingest_reclassify_threshold.py::TestMaterialRewriteRetiers`. Every re-tiering test asserts the **read-side ceiling behaviour** (`tier_allowed(_fragment_tier(post.metadata), TierCeiling.OPEN)`), not just the frontmatter string — that is the assertion that actually protects the invariant.

RED before the fix (3), verified by running them:
- `test_rewrite_into_recovery_content_is_refused_at_open_ceiling` — `AssertionError: assert 'open' == 'intimate'`
- `test_rewrite_refreshes_derived_voice_proxy_eligible`
- `test_untiered_fragment_gets_a_tier_assigned_outright`

Guard tests (pinning what must *not* change):
- `test_rewrite_into_benign_content_stays_open` — no blanket blackout
- `test_rewrite_never_downgrades_an_intimate_tier` — escalate-only; also pins the on-disk merge base
- `test_trivial_edit_leaves_the_tier_untouched` — a typo-scale edit does not re-tier even when the appended text carries a recovery keyword
- `test_unrecognised_disk_tier_fails_closed_to_intimate` — covers the fail-closed parse arm with a *benign* rewrite, so the only route to `intimate` is the fail-closed parse

Materiality is guaranteed by construction: difflib's ratio is bounded above by `2*min(la,lb)/(la+lb)`, and the fixture lengths put that bound near 0.5, well below the 0.9 threshold.

**Gate 2** (`creek-tools/scripts/check-all.sh`, the creek-tools gate — the diff does not touch `crawdad/`): **12/12 checks passed**, 6357 tests passed, **93.99%** branch coverage. Full unit suite separately: 6382 passed, 5 skipped, **0 failures — no fixture fallout**, which is itself a consequence of the design: a body-derived, escalate-only merge has far less blast radius than a blanket tier reset would have had.

**Gate 2.5**: `code-review-orchestrator` over the full diff — verdict **CLEAN**, no blockers. It independently confirmed the on-disk merge base, the short-circuit safety of the `post.metadata[PRIVACY_TIER_KEY]` subscript, the `voice_proxy_eligible` formula matching the model's computed field, and that no other write path (`creek/save/writer.py`, `creek/purge/engine.py`, `creek/ingest/refresh.py`) rewrites a body while leaving a stale tier.

Closes #922

### DIFF
--- a/creek-tools/creek/vault/writer.py
+++ b/creek-tools/creek/vault/writer.py
@@ -36,6 +36,7 @@ from creek.models import (
     Authorship,
     DecisionStatus,
     PraxisType,
+    PrivacyTier,
     SourcePlatform,
 )
 from creek.vault.authors import (
@@ -81,6 +82,12 @@ _PLATFORM_SUBFOLDER: dict[str, str] = {
 
 # Destination for soft-tombed fragments whose source unit has vanished (#674).
 _ORPHANED_RELPARTS: tuple[str, ...] = ("10-Liminal", "Orphaned")
+
+# Frontmatter key holding the serialised ``Fragment.voice_proxy_eligible``
+# computed field (BUG-009). It is derived from ``privacy_tier`` +
+# ``source.author``, so any code path that rewrites the tier on disk owes this
+# key a refresh or the snapshot goes stale (#922).
+_VOICE_PROXY_ELIGIBLE_KEY: str = "voice_proxy_eligible"
 
 # Map an AuthorManifest.author_kind -> the fragment's Authorship axis (#470).
 # Any unrecognised kind (the manifest already fails closed to ``human_source``)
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
Follow-up filed for the deliberately out-of-scope divergence noted in the PR body: #961 — `creek_mcp/tier_ceiling.py` ranks `UNCLASSIFIED` as `OPEN` while `creek/classify/privacy_filter.py` ranks it as `PERSONAL` (#876 straggler).

Geoffe-Ga:
## Summary
`VaultWriter.update_fragment`'s material-rewrite path cleared classification provenance keys but left `privacy_tier` describing the *old* body, so a fragment rewritten from benign into intimate content stayed admissible at an OPEN MCP ceiling. This PR adds `_retier_after_rewrite`, which re-derives the tier from the new body via the existing LLM-free `PrivacyClassifier`, merges it escalate-only against the on-disk tier, and refreshes the derived `voice_proxy_eligible` flag.

## Strengths
- Correct, minimal fix at the single write site that creates the staleness, rather than spreading defensive checks across every read site.
- The escalate-only merge is verified to use the **on-disk** tier as its base, not `tier_of(fragment)` — I confirmed the freshly-built `Fragment` the ingest pipeline hands to `update_fragment` still carries the `unclassified` pipeline default, which ranks *below* `open` in `privacy_pass._ESCALATION_RANK`; escalating against it would have silently discarded an operator's `intimate`. Getting this right is the crux of the fix and it's correct.
- `_tier_on_disk` correctly mirrors the fail-closed parsing pattern already used by `creek/classify/privacy_filter.py::tier_of` and `creek_mcp/tools/reflect.py::_fragment_tier` (unrecognised value → `INTIMATE`, never `open`).
- `voice_proxy_eligible` is recomputed with the exact formula from `Fragment.voice_proxy_eligible` (`tier != INTIMATE and author == SELF`) — I checked both definitions side by side and they match, so the frontmatter snapshot can't drift from the computed field.
- The PR body explicitly rules out both cheaper alternatives (blanket `intimate` stamp is a permanent ratchet since nothing ever downgrades `privacy_tier`; reset to `unclassified` is a no-op because `tier_ceiling._TIER_RANK` ranks it with `OPEN`) and I independently verified both claims against `privacy_pass.needs_tier`/`purge/engine.py` and `tier_ceiling._TIER_RANK`. This is unusually rigorous self-verification.
- Test suite asserts the actual protected invariant (`tier_allowed(_fragment_tier(post.metadata), TierCeiling.OPEN)`), not just the frontmatter string — this is the assertion that matters, and most of the new tests use it.
- Good coverage of the policy's edge cases: benign rewrite stays open, intimate never downgrades, trivial edits don't re-tier even with a keyword injected, unclassified-on-disk assigns outright, unrecognised on-disk value fails closed.

## Security Concerns
None outstanding — this PR *fixes* a real read-side exposure (stale permissive tier surviving a material rewrite into sensitive content, reachable by a remote OPEN-ceiling MCP caller). The fix is fail-closed throughout (unrecognised tier → INTIMATE, escalate-only merge, no auto-downgrade path). 🟢

## Problems
None found blocking merge.

## Code Quality
- Docstrings are extremely dense (the `_retier_after_rewrite` docstring runs ~50 lines re-explaining rejected alternatives). Useful for a security-sensitive change, but it's a lot to maintain if the policy shifts again — consider trimming to what a future reader needs to *not break the invariant*, with the design-rationale detail left in the PR description / linked doc. Non-blocking style preference.
- `docs/idempotent-ingest.md` addition is accurate and appropriately concise relative to the code docstring.
- The known `#876` divergence between `tier_ceiling._TIER_RANK[UNCLASSIFIED]=0` and `privacy_filter._TIER_RANK[UNCLASSIFIED]=1` is correctly identified as out of scope and pinned by existing tests rather than silently left inconsistent — good call to file separately rather than scope-creep this PR.

## CLAUDE.md Adherence
- ✅ TDD: RED commit (`9aec9e1`) precedes the fix commit, with tests demonstrated failing before the fix per the PR description.
- ✅ Conventional commits: `test(vault): ...` then `fix(vault): ...`, both referencing the issue.
- ✅ Docstrings: Google-style, present on all new public/private functions well above the 95% interrogate floor.
- ✅ Complexity: `_retier_after_rewrite` and `_tier_on_disk` are both low-branch-count, well under the ≤10 cyclomatic limit.
- ✅ Deferred imports follow the exact documented pattern for the `classify_engine` ↔ `VaultWriter` load cycle already established by `_clear_classification`.
- Per the PR description, `./scripts/check-all.sh` reports 12/12 checks passing locally (6357 tests, 93.99% branch coverage) and a prior `code-review-orchestrator` pass came back clean. I independently traced every factual claim in the PR body (escalation ranks, `needs_tier`, `PrivacyClassifier.classify_tier` order-of-checks, `voice_proxy_eligible` formula, the `#876` rank divergence) against the current source and all checked out.
- CI (`Code Quality & Testing` 3.11/3.12/3.13, `claude-review`) was still `IN_PROGRESS` at review time; `CrawDad Quality` jobs (irrelevant — no `crawdad/` changes) are green. Per Gate 3/4, merge should wait for those to go green and for the automated `claude-review` verdict, but this is a process gate rather than a defect in the change itself.

## Testing
Strong. 7 new tests in `TestMaterialRewriteRetiers` cover: refusal-at-ceiling after a rewrite into sensitive content, benign rewrite staying open, never-downgrade of an existing intimate tier, trivial edits leaving the tier untouched even with an injected keyword, the derived `voice_proxy_eligible` flag following the tier down, outright assignment when the on-disk value is `unclassified`, and fail-closed behavior on an unrecognised on-disk tier value (with the rewrite body deliberately kept benign so the fail-closed path is the *only* route to the asserted result — a nice touch that avoids a false-positive test). Materiality of the fixture bodies is guaranteed by construction via a documented `difflib` ratio bound rather than eyeballed, which is credible.

## Documentation
`docs/idempotent-ingest.md` is updated to describe the new re-derivation behavior, the LLM-free heuristic, and the escalate-only merge — accurate and matches the implementation.



## Verdict: LGTM

The fix correctly closes the security gap (a materially rewritten fragment kept a stale, permissive privacy_tier that governed read-side admission), the escalate-only/on-disk-merge-base design is verifiably correct against the existing privacy_pass policy module, tests pin both the fix and the read-side ceiling behavior it protects, and every factual claim in the PR description checks out against the actual code. Nothing here blocks merge; the only remarks are minor nits an author could take or leave.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5111675010 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #962, opened 2026-07-29
creek compile CLI resolves the generation stage with no tier — the Intimate-never-cloud gate is a no-op


Geoffe-Ga:
## Summary

`creek/cli.py:2477` builds the compile LLM as:

```python
llm=default_llm(config.model_router.resolve("generation")),
```

It goes *through* `ModelRouter`, which looks correct at a glance — but it passes **no tier**. `ModelRouter._enforce_local_for_intimate` (`creek/classify/llm/router.py:142`) only acts when `tier == PrivacyTier.INTIMATE`, so with `tier=None` the chokepoint is a **no-op** and the call resolves to whatever `generation` is configured as.

Empirically confirmed against `main` @ 756bf8d:

```
tier=None      -> provider=anthropic
tier=open      -> provider=anthropic
tier=personal  -> provider=anthropic
tier=intimate  -> provider=ollama      <-- only this one is gated
```

(with `LLMRoutingConfig(default=ollama, generation=anthropic)`)

So `creek compile` sends intimate source material to a cloud provider. What egresses is specifically **titles**: `creek/compile/engine.py::_build_prompt` (lines 334-370) emits `id:` and `title:` for every source unconditionally, and `_fragment_excerpt_for_prompt` (line 318) redacts only the *body* — replacing it with `[Intimate-tier summary: {title}]`, which puts the title straight back in. Persisted `structural_path` breadcrumbs carry ancestor titles too.

## Relationship to #928 / #929

This is the CLI-side twin of #928. #928/#929 were fixed for the **MCP** surface in PR #<TBD>, which added `creek_mcp.tier_ceiling.routing_tier(ceiling, content_tier)` and keys the compile factory with the more sensitive of the caller's ceiling and the max tier across admitted sources.

The CLI was deliberately left out of that PR's scope: neither issue named it, and the threat model differs (the CLI is the vault owner on their own machine; the MCP surface serves remote consumers). But the Intimate-never-cloud invariant (#647) is supposed to be absolute, so this should still be closed.

## Why it is not a one-liner

The CLI has no tier to pass at the call site. `compile_to_vault` loads the fragments *internally* (`_load_fragments_for_compile`), so only the engine knows the source tiers, and the LLM is already built by then.

[...]

Issue #958, opened 2026-07-29
creek.draft MCP tool bypasses ModelRouter — Intimate-never-cloud not enforced (fails closed only by accident)


Geoffe-Ga:
**Found while verifying #918** (forwarding `CREEK_CLOUD_CONSENT` to the MCP subprocess). Filed separately to keep that PR scoped; #918 does **not** introduce or widen this hole, but it removes one accidental layer of obscurity in front of it.

## Problem

The MCP `creek.draft` tool builds its LLM client **without going through `ModelRouter`**, so the Intimate-never-cloud chokepoint (epics #642/#643) is never consulted on that path.

`creek-tools/creek_mcp/server.py:156-174`:

```python
def _build_draft_llm() -> Callable[[str], str]:
    from creek.classify.llm import LLMClassifier

    config = load_config()
    classifier = LLMClassifier(config.llm)      # no router, no stage, no tier
    ...
    return classifier.invoke_prompt
```

No `model_router`, no `resolve()`, no `PrivacyTier`. Compare the sibling `_build_reflect_llm_factory` (`server.py:190-214`), which does it correctly:

```python
router = load_config().model_router          # server.py:202
cfg = router.resolve("generation", tier)     # server.py:205
```

`ModelRouter._enforce_local_for_intimate` (`creek/classify/llm/router.py:142-176`) only fires when a `tier` argument is actually supplied — `tier=None` passes straight through (router.py:165). Since `_build_draft_llm` never supplies one, there is no enforcement at all.

Intimate content genuinely reaches that prompt: `draft_tool` turns the caller's ceiling into a `PrivacyTierOverride` (`creek_mcp/tools/draft.py:54`) and hands it to `DraftGenerator` (draft.py:74-78). With `privacy_tier_ceiling` of `intimate`/`all`, `filter_fragments_by_tier` (`creek/generate/drafts.py:599`) admits intimate fragments with **full bodies**, which land in the prompt at `drafts.py:1349` (`self._llm(prompt)`). The ceiling is chosen by the MCP caller, so it is the mechanism that *enables* intimate inclusion, not a defense against it.

By contrast `creek.author` and `creek.classify` are correctly router-enforced (`creek/author/client.py:122`; `creek/classify/classify_engine.py:390-393`), and `creek.mine` makes no LLM call at all.

## Why it hasn't leaked yet — it fails closed *by accident*

[...]

PR #963, opened 2026-07-29
fix(mcp): tier-route the creek.compile LLM through ModelRouter (#928, #929)

### LINKED ISSUE #928: creek.compile MCP LLM is not tier-routed: admitted intimate source titles can egress to cloud
## Context
Found while fixing #848 (source-tier ceiling gate for `creek.compile`).

#848 stops **above-ceiling** source fragments from reaching the compile LLM. It does **not** stop an **admitted** intimate source under a local `intimate`/`all` ceiling from egressing.

## The gap
`creek_mcp/server.py::_build_compile_llm` resolves `load_config().llm` directly, with **no `ModelRouter` and no tier key** — unlike `_build_reflect_llm_factory`, which is tier-keyed so `ModelRouter._enforce_local_for_intimate` can force INTIMATE onto a local model or raise `IntimateRoutingError`.

Meanwhile `creek/compile/engine.py::_build_prompt` emits `id:` and `title:` for every source fragment **unconditionally**; `_fragment_excerpt_for_prompt` redacts only the *body*.

So a local caller with `privacy_tier_ceiling=intimate` compiling an intimate fragment ships that fragment's id and title to whatever provider is configured — violating the Intimate-never-cloud house rule.

## Suggested fix
Mirror `_build_reflect_llm_factory`: make the compile factory tier-keyed and route through `ModelRouter`, deriving the routing tier from the (now ceiling-checked) source fragments. Consider also redacting `title:` for intimate sources in `_build_prompt`.

## Notes
- Not remotely exploitable: `_BoundedFastMCP` caps remote consumers at `personal`, and #848's gate now refuses above-ceiling sources.
- Sibling bug in the same function filed separately (the `_default_llm` import).

Refs #848

### LINKED ISSUE #929: creek_mcp.server._build_compile_llm imports a nonexistent _default_llm; MCP compile raises ImportError, not a refusal
## Context
Found while fixing #848.

## The bug
`creek_mcp/server.py::_build_compile_llm` does:

```python
from creek.compile.engine import _default_llm as _engine_default_llm
return _engine_default_llm(load_config().llm)
```

`creek/compile/engine.py` exports **`default_llm`** (no leading underscore) at line 516. There is no `_default_llm`. So `_build_compile_llm()` raises `ImportError` on every production invocation of the `creek.compile` MCP tool.

Two problems:

1. **MCP compile has never worked in production.** Only tests, which inject their own factory, exercise the path.
2. **The error is not caught as a structured refusal.** `creek_mcp/tools/compile.py` catches `(ValueError, RuntimeError)` around the engine call; `ImportError` is neither, so it escapes the MCP boundary as an unhandled transport error instead of `status="refused"`.

There is likely a second latent bug behind it: `load_config().llm` is an `LLMRoutingConfig` (per the per-stage routing work) while `default_llm` expects an `LLMConfig`. Compare `creek/cli.py:2256`, which uses `config.model_router.resolve("generation")`.

## Suggested fix
Fix the import name, resolve the config through `model_router` the way the CLI does, and add a regression test that builds the real server factory. Consider widening the wrapper's `except` so a provider-construction failure becomes a refusal rather than a transport error.

## Note
Fixing this will make #928 (compile LLM not tier-routed) live rather than theoretical, so land them together or land #928 first.

Refs #848

Geoffe-Ga:
## Summary

`creek.compile` on the MCP surface had two entangled defects. Fixing either alone would have been wrong, so both land here.

- **#929** — `creek_mcp/server.py:185` imported `_default_llm`, but `creek/compile/engine.py:516` defines `default_llm` (no underscore). Every production invocation raised `ImportError`. Tests never caught it because `server.py:257` lets them inject `compile_llm_factory` — factory injection is exactly what hid the bug.
- **#928** — the same line passed `load_config().llm` (an `LLMRoutingConfig`, which has no `.provider`) straight into the provider factory, bypassing `ModelRouter`, the Intimate-never-cloud chokepoint.

**Why one PR:** the broken import was the *only* thing preventing the unrouted egress. Correcting the symbol name in isolation would have converted a dead path into a live, router-bypassing one — shipping the leak the breakage was accidentally suppressing. This path has never worked in any form; #929 was never a one-character fix.

### What egresses

`_build_prompt` emits `id:` and `title:` for **every** source unconditionally; `_fragment_excerpt_for_prompt` redacts only the *body*, replacing it with `[Intimate-tier summary: {title}]` — putting the title straight back. So intimate **titles** are live egress payload, exactly as the issue says.

### The fix

`_build_compile_llm(tier)` resolves the `generation` stage via `ModelRouter.resolve(stage, tier)`. Empirically, `tier=None` returns the cloud provider and only `tier=INTIMATE` triggers `_enforce_local_for_intimate` — so passing a tier is the whole ballgame.

`compile_tool` keys the factory with **the more sensitive of** the caller's ceiling and the maximum tier across admitted sources:

| ceiling | sources | routes to |
|---|---|---|
| OPEN | all open | cloud |
| PERSONAL | open+personal | cloud |
| INTIMATE / ALL | open only | **local** |
| INTIMATE / ALL | any intimate | **local** |
| any | no `privacy_tier` key | **local** (fails closed) |
| any | zero ids resolve | **local** (fails closed) |

Taking the max is what makes it uncheatable: a low declared ceiling loses to the sources' own tiers, and low-tier sources lose to a high ceiling. No declaration a remote consumer can make buys cloud routing for intimate content.

This mirrors `reflect._routing_tier`, whose docstring calls the ceiling term "defense in depth ... must not be removed". That reconciliation is promoted into public `creek_mcp.tier_ceiling.routing_tier` and now **shared** by reflect and compile, so the two MCP surfaces cannot drift on the privacy rule — and it collapses a duplicated tier ranking.

The tier is computed in the **existing** #848 gate walk (`_sources_above_ceiling` → `_survey_sources`, returning a `_SourceGate`), so there is **no extra vault I/O** — important at 35k fragments. The early `return True` is gone, which makes the probe-cost uniformity that the #848 comment called "accidental" true by construction. `max_tier` never enters any response, refusal, or audit payload; echoing it would create a per-call tier-classification oracle strictly worse than the documented bare-bool one, and there is a regression test pinning that.

## Test plan

Gate 2: `./scripts/check-all.sh` → **exit 0**, 6388 passed, coverage 94.00% (baseline on `main` was 6354 passed, also exit 0).

Because `creek_mcp/` is outside the static-analysis net (`scripts/typecheck.sh:61` runs `mypy creek/` only — per #925, and *that exclusion is why both defects survived*), I ran `python -m mypy creek_mcp/` by hand: **7 errors → 4**. The three eliminated are precisely this bug (`server.py:185 has no attribute "_default_llm"; maybe "default_llm"?`, the `no-any-return` beside it, and compile.py's `object` arg-type). The 4 remaining are pre-existing and out of scope. `typecheck.sh` is deliberately unchanged — that is #925's scope.

Key tests:
- `test_compile_routes_intimate_sources_to_the_local_provider` — the flagship. Mixed tiers (intimate among open) at a ceiling that **admits** intimate, so it is the live-egress case, not the refusal case. Nothing is stubbed at the seam under test: the real `_build_compile_llm` is the factory, a **real `ModelRouter`** resolves the stage, and only `build_provider` is spied. Asserts the resolved provider is local **and** that the intimate title is genuinely in the prompt that reached it (and the body is not). A single-tier compile could not pin this.
- `test_compile_keys_the_llm_factory_with_the_routing_tier` — 6-way matrix; asserts exactly one call with the exact tier, and never `None`.
- `test_build_server_wires_the_production_compile_llm_factory` — builds the server **omitting** `compile_llm_factory` entirely. This is the test that would have caught #929.
- Fail-closed pins for absent `privacy_tier` and zero-resolving ids; `IntimateRoutingError` → structured refusal; and an anti-oracle guard that the refusal gains no tier-derived field.

Gate 2.5 self-review returned **CLEAN** with no blockers. Its one nit — the source-gate docstring overstated its not-found guarantee — is fixed in the third commit, with a test pinning the argument-evaluation ordering.

## Notes

- Test integrity: exactly **6** deleted lines across all test files, every one verified signature-only (adding the `tier` parameter to stub factories) plus one additive `title=` kwarg. Assertion counts rose 162→191, 59→69, 4→13. Reflect's tests are **unmodified and green**, which is the proof the shared-helper extraction changed no behavior.
- Out of scope, filed separately: **#962** — `creek compile` on the **CLI** (`cli.py:2477`) calls `resolve("generation")` with no tier, so the same gate is a no-op there. It needs the engine-signature change already tracked as a follow-up, and the threat model differs (vault owner vs. remote consumer).
- **#931** (persisted `structural_path` can carry an unrequested intimate ancestor's title into the prompt) is a real residual this PR does **not** close.
- Added evidence to **#958**: `creek.draft` is broken the same way (`AttributeError: 'LLMRoutingConfig' object has no attribute 'provider'`) and latently suppresses the same unrouted egress — so it needs the same both-at-once treatment.

Closes #928
Closes #929

### DIFF
--- a/creek-tools/creek_mcp/server.py
+++ b/creek-tools/creek_mcp/server.py
@@ -65,7 +65,9 @@ if TYPE_CHECKING:
     from mcp.server.auth.provider import AccessToken, TokenVerifier
     from mcp.types import ContentBlock
 
+    from creek.compile.engine import CompileLLM
     from creek.models import PrivacyTier
+    from creek_mcp.tools.compile import CompileLLMFactory
     from creek_mcp.tools.reflect import _LLM, _LLMFactory
 
 SERVER_NAME = "creek-tools-mcp"
@@ -174,17 +176,45 @@ def _build_draft_llm() -> Callable[[str], str]:
     return classifier.invoke_prompt
 
 
-def _build_compile_llm() -> Callable[[str], str]:
-    """Return the compile-side LLM callable, mirroring ``creek compile``.
+def _build_compile_llm(tier: PrivacyTier) -> CompileLLM:
+    """Return the compile-side LLM callable, routed for *tier* (#928/#929).
 
-    Lazy import so the server still boots when the LLM provider is not
-    configured; only ``creek.compile`` then fails. Calls into the CLI's
-    private factory rather than re-deriving the configuration so the
-    behaviour stays in lock-step with ``creek compile``.
+    Mirrors :func:`_build_reflect_llm_factory`'s inner factory: the
+    ``generation`` stage is resolved through the config's
+    :class:`~creek.classify.llm.router.ModelRouter`, so an INTIMATE compile is
+    forced onto the local ``default`` model — or the router raises
+    ``IntimateRoutingError`` rather than egressing the source titles the
+    compile prompt carries. Nothing here re-checks the tier or picks a
+    provider: ``compile_tool`` derives the routing tier and this hands it to
+    the one chokepoint that owns the decision.
+
+    Lazy imports keep the server bootable with no provider configured — only a
+    ``creek.compile`` call then fails, and it fails as a structured refusal.
+
+    Args:
+        tier: The routing tier ``compile_tool`` computed from the caller's
+            ceiling and the source fragments' own classifications.
+
+    Returns:
+        A prompt → completion-text callable bound to the resolved provider.
+
+    Raises:
+        RuntimeError: When the resolved provider is unavailable, or — as
+            :class:`~creek.classify.llm.router.IntimateRoutingError`, a
+            subclass — when intimate content has no local backend to fall
+            back to. ``compile_tool`` turns either into a refusal.
     """
-    from creek.compile.engine import _default_llm as _engine_default_llm
+    from creek.classify.llm.providers import build_provider
 
-    return _engine_default_llm(load_config().llm)
+    cfg = load_config().model_router.resolve("generation", tier)
+    provider = build_provider(cfg)
+    if not provider.available:
+        msg = (
+            "LLM provider unavailable for compile. "
+            "Check Ollama or ANTHROPIC_API_KEY configuration."
+        )
+        raise RuntimeError(msg)
+    return lambda prompt: provider.complete(prompt).text
 
 
 def _build_reflect_llm_factory() -> _LLMFactory:
@@ -219,7 +249,7 @@ def build_server(
     *,
     vault_path: Path | None = None,
     draft_llm_factory: Callable[[], Callable[[str], str]] | None = None,
-    compile_llm_factory: Callable[[], Callable[[str], str]] | None = None,
+    compile_llm_factory: CompileLLMFactory | None = None,
     reflect_llm_factory: Callable[[], _LLMFactory] | None = None,
     token_verifier: TokenVerifier | None = None,
 ) -> FastMCP:
@@ -232,9 +262,10 @@ def build_server(
         draft_llm_factory: Optional factory for the draft LLM. The
             factory is invoked lazily so only ``creek.draft`` needs an
             LLM provider.
-        compile_llm_factory: Optional factory for the compile LLM.
-            Invoked lazily so only ``creek.compile`` needs an LLM
-            provider.
+        compile_llm_factory: Optional tier-keyed factory for the compile
+            LLM — ``factory(tier)``. Passed straight to ``creek.compile``,
+            which invokes it lazily (and only after its source-tier gate),
+            so only ``creek.compile`` needs an LLM provider.
         reflect_llm_factory: Optional thunk returning the tier-keyed LLM
             factory for ``creek.reflect``. Invoked lazily per call so only
             ``creek.reflect`` needs an LLM provider.
--- a/creek-tools/creek_mcp/tier_ceiling.py
+++ b/creek-tools/creek_mcp/tier_ceiling.py
@@ -5,6 +5,16 @@ content above the ceiling is omitted or returned as a title-only stub.
 The four ceiling values mirror
 :class:`creek.classify.privacy_filter.PrivacyTierOverride` so the MCP
 surface and ``--include-tier`` stay in lock-step.
+
+Two distinct questions are answered here, both off the same ranking:
+
+- *admission* — :func:`tier_allowed` / :func:`write_tier_allowed`: may this
+  content be read (or created) under the caller's ceiling at all?
+- *routing* — :func:`routing_tier`: given that it was admitted, which
+  :class:`~creek.models.PrivacyTier` must the LLM call be keyed with so
+  :class:`creek.classify.llm.router.ModelRouter` applies the
+  ``Intimate``-never-cloud gate (#928)? Every tool that hands content to a
+  model derives its tier here rather than deciding for itself.
 """
 
 from __future__ import annotations
@@ -47,6 +57,10 @@ _CEILING_TO_OVERRIDE = {
 }
 
 
+# The one sensitivity ranking in the MCP surface: open/unclassified <
+# personal < intimate. Both the admission predicates and the routing helpers
+# read it through :func:`tier_sensitivity` so a tier can never rank one way
+# for "may I read this?" and another for "where may I send it?".
 _TIER_RANK = {
     PrivacyTier.OPEN: 0,
     PrivacyTier.UNCLASSIFIED: 0,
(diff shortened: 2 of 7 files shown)


Geoffe-Ga:
## Summary
PR #963 fixes two entangled defects in `creek.compile`'s MCP surface: a broken `_default_llm` import (#929, engine.py actually defines `default_llm`) that made every production compile call raise `ImportError`, and — the defect that import failure was accidentally masking — a complete bypass of `ModelRouter`'s Intimate-never-cloud gate (#928), which would have let intimate fragment titles egress to a cloud provider. The fix threads a computed routing tier (the more sensitive of the caller's ceiling and the max tier among admitted sources) through `_build_compile_llm(tier)` into `ModelRouter.resolve`, and promotes the ceiling/content reconciliation into a shared `creek_mcp.tier_ceiling.routing_tier` used by both `compile` and `reflect`.

## Strengths
- Root-caused correctly and bundled deliberately: fixing the import alone first would have converted a dead path into a live egress path — explained clearly in the PR body and commit messages.
- The tier is derived from the *existing* `#848` gate walk (`_survey_sources`), adding zero extra vault I/O at 35k-fragment scale, and turns an "accidental" probe-cost uniformity into one guaranteed by construction (no more early-`return True`).
- Fail-closed everywhere it matters: unrecognized tiers/ceilings, fragments with no `privacy_tier` key, and zero-resolving ids all route to `INTIMATE`/local rather than defaulting open.
- `max_tier` is verified to never leak into any response, refusal, or audit payload — pinned by an explicit anti-oracle test asserting the exact key sets of both the refusal dict and the audit entry.
- The flagship test (`test_compile_routes_intimate_sources_to_the_local_provider`) exercises a *real* `ModelRouter`, not a stub's opinion of what it should do, and asserts the intimate title (not just the tier) actually reaches the local provider and never the cloud one.
- The `reflect.py` extraction into shared `tier_ceiling.routing_tier` is a clean, behavior-preserving refactor (reflect's own tests are unmodified and green, proving no regression), and it closes a duplicated ranking table that could have drifted between the two MCP surfaces.
- Verified independently by reading `creek/classify/llm/router.py` (`ModelRouter.resolve`/`_enforce_local_for_intimate`) and `creek/compile/engine.py` (`default_llm`, `_build_prompt`, `_fragment_excerpt_for_prompt`) — every claim in the PR description about the pre-fix behavior and the router's semantics checks out against the actual source.

## Security Concerns
None outstanding — this PR *closes* a real cloud-egress-of-intimate-content vulnerability (#928) and a totally broken production path (#929). 🟢 No new concerns introduced; the anti-oracle guard against a tier-classification side channel is a nice piece of defense-in-depth reasoning.

## Problems
None found blocking merge.

## Code Quality
- `_survey_sources`'s docstring documents a subtle, deployment-dependent ordering seam (the router's `IntimateRoutingError` can preempt the engine's "not found" message when `llm_factory(tier)` is evaluated as a call argument before the not-found check runs). This is honestly disclosed, judged safe (fails closed, names no id), and pinned with a dedicated regression test rather than silently left as a footgun — good judgment call, not a smell.
- The in-code documentation is very dense/verbose (near-essay-length docstrings and comments throughout `compile.py`/`tier_ceiling.py`). This matches the pre-existing style already present in `reflect.py`, so it's consistent with repo convention rather than a new deviation, but a future pass to trim some of the repetition between module docstring, function docstring, and inline comments (all explaining the same routing invariant) could improve scanability. Non-blocking.

## CLAUDE.md Adherence
✅ TDD 3-commit structure (RED → GREEN → docs-refinement-after-self-review) matches the Stay Green workflow.
✅ Reported `check-all.sh` exit 0, 6388 passed, coverage 94.00% (verified plausible against the actual test additions inspected).
✅ Shared single source of truth for the tier ranking (`tier_ceiling.tier_sensitivity`/`routing_tier`) — satisfies the DRY principle explicitly.
✅ No shortcuts/suppressions introduced; `# noqa: E731` in test lambdas is pre-existing style, not new.
✅ mypy strict manually verified for `creek_mcp/` (7→4 errors, remaining are pre-existing/out-of-scope per #925) since that directory sits outside the `typecheck.sh` net — appropriately scoped rather than silently expanding scope.

## Testing
Extensive and well-targeted: real-`ModelRouter` flagship test proving actual prompt content is tier-routed, a 6-row routing-tier matrix, fail-closed pins for absent-tier and zero-resolving-id cases, an `IntimateRoutingError`-surfaces-as-refusal test, the ordering-seam regression test, and an anti-oracle guard on the refusal/audit payload shape. `test_build_server_wires_the_production_compile_llm_factory` specifically targets the exact gap (no `compile_llm_factory` injected) that let #929 ship undetected. Test-only diff lines are minimal and verified signature-only per the PR's own audit. No gaps identified.

## Documentation
Docstrings are exhaustive to the point of redundancy in places (see Code Quality) but fully accurate against the implementation — every behavioral claim (router semantics, egress path, fail-closed defaults) was cross-checked against source and holds up.



## Verdict: LGTM

The fix is verified correct by direct inspection of `ModelRouter.resolve`/`_enforce_local_for_intimate`, `engine.py`'s `default_llm` (confirming the `_default_llm` import really didn't exist) and `_build_prompt`/`_fragment_excerpt_for_prompt` (confirming titles egress unconditionally). The tier-threading through `_build_compile_llm(tier)` → `CompileLLMFactory` → `compile_tool`'s `routing_tier(ceiling, gate.max_tier)` is sound, fails closed on every edge case (unresolved ids, missing `privacy_tier`, unknown tier/ceiling), never leaks the computed tier into refusal/audit payloads (test-pinned), and the reflect.py extraction is a pure behavior-preserving refactor. Tests are thorough and exercise the real `ModelRouter`, not a stub, at the flagship seam. No blocking issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5112230820 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #972, opened 2026-07-29
creek.redact.scan returns intimate fragment filenames (slugified titles) to an open-ceiling caller, plus absolute paths in report_markdown


Geoffe-Ga:
## Context

Found by the #932 sweep, during Gate-2.5 self-review. This is the **only
read-side leak** the sweep found — every other gap it turned up (#968, #970,
#971) is write-side. It also falsifies the sweep's own working conclusion that
"no unaudited tool leaks above-ceiling content into its MCP *response*". That
conclusion was drawn from probing the tools that walk the corpus on their own;
`creek.redact.scan` was set aside as "the caller named the path", which is true
and turns out not to be sufficient.

## The gap

`creek_mcp/tools/redact.py` confines `input_path` to *inside the vault* via
`resolve_within_vault`, and never consults `privacy_tier` at all. FEAT-027's own
module docstring frames the intended use as scanning Discord attachments staged
under `00-Creek-Meta/Inbound/` before ingest — but nothing restricts it to that
subtree. A caller at `privacy_tier_ceiling=open` can point it at
`01-Fragments`.

The docstring's safety argument is:

> the scan only reports counts + line numbers + salted hashes — no matched text
> and no file body content leaves the tool.

All true. The problem is what else leaves the tool: **the file path**. Creek
fragment filenames are slugified titles.

## Reproduced

Vault with one fragment, `privacy_tier: intimate`, titled "My Affair With Dana",
stored at `01-Fragments/Notes/my-affair-with-dana.md`, body containing a phone
number and an email.

```python
redact_scan_tool(vault_path=v, input_path="01-Fragments",
                 privacy_tier_ceiling=TierCeiling.OPEN)
```

Response (`status: ok`):

```json
"findings": [
  {"file_path": "01-Fragments/Notes/my-affair-with-dana.md",
   "line_number": 17, "match_type": "email",
   "severity": "medium", "salted_hash": "e15b2dcc…"},
  {"file_path": "01-Fragments/Notes/my-affair-with-dana.md",
   "line_number": 17, "match_type": "phone_number",
   "severity": "medium", "salted_hash": "aa31203e…"}
]
```

An `open`-ceiling caller learns: an intimate fragment exists, **what it is
called**, that it contains an email address and a phone number, and on which
line. The title is the sensitive part — "no matched text" does not cover it.

## A second, separate bug in the same response

`_finding_to_dict`'s docstring says:

[...]

Issue #971, opened 2026-07-29
creek.skills.refresh never threads privacy_tier_ceiling: personal bodies pass unsummarised at ceiling=open


Geoffe-Ga:
## Context

Found by the #932 sweep ("audit every MCP tool that reads vault fragments for a
source-side tier-ceiling gate").

`creek_mcp/tools/skills.py` says:

> The generator already excludes ``intimate`` exemplars so the operation is safe
> at any ceiling.

The first half is **true** — I verified `creek/generate/skills.py::_is_snapshot_fragment`
returns `allow_intimate` (default `False`) for `PrivacyTier.INTIMATE`, and
`skills_refresh_tool` never passes `allow_intimate=True`. The intimate exclusion
is real.

The second half — "safe at any ceiling" — is overbroad, because the ceiling is
not threaded at all:

- `skills_refresh_tool` never calls `to_privacy_override`, and
  `SkillTreeGenerator` never touches `creek.classify.privacy_filter`.
- So `PERSONAL` fragments contribute their **full bodies** to the generated
  skill tree, at every ceiling including `open`.

Every other generation tool that reaches unsupplied vault fragments
(`creek.mine`, `creek.draft`, `creek.author`) routes through
`filter_fragments_by_tier` / `tier_within_override`, where a personal fragment at
`ceiling=open` contributes a **title-only summary**
(`_summarize_personal`). `creek.skills.refresh` is the odd one out: it has a
hand-rolled, intimate-only filter instead of the shared one, so it is stricter
than nothing but looser than every sibling.

The output lands in `<vault>/creek-skills/`, an untiered vault artifact.

## Why this is its own issue

It is the same *family* as the `creek.report` gap (privacy-blind or
partially-blind generators materialising above-ceiling content into untiered
artifacts) but a different fix: `report`'s generators have no tier awareness at
all, whereas `SkillTreeGenerator` has its own and simply needs to defer to the
shared one.

## Suggested shape

1. Give `SkillTreeGenerator.generate_all_skills` an
   `override: PrivacyTierOverride | None` parameter and route eligibility through
   `creek.classify.privacy_filter` rather than the hand-rolled
   `_is_snapshot_fragment` tier branch — keeping `allow_intimate` working as the
   `INTIMATE`/`ALL` override values.
2. Thread `to_privacy_override(privacy_tier_ceiling)` from `skills_refresh_tool`,
   and `--include-tier` from the CLI.
[...]

Issue #970, opened 2026-07-29
creek.journal update-in-place lets an unadmitted caller overwrite an above-ceiling fragment's body


Geoffe-Ga:
## Context

Found by the #932 sweep ("audit every MCP tool that reads vault fragments for a
source-side tier-ceiling gate"). This is the write-side mirror of #848: a tool
that acts on a vault fragment the caller did not supply, without checking that
fragment's classified tier against the caller's ceiling.

`creek.journal` is idempotent by `external_id` (#754): the same id
**updates the existing fragment in place**. `journal_ingest_tool` gates the
*incoming* tier with `write_tier_allowed(entry_tier, privacy_tier_ceiling)`
(`creek_mcp/tools/journal.py:158`) — but it never loads or consults the tier of
the fragment the update is about to overwrite.

## Reproduced

```python
journal_ingest_tool(vault_path=v, content="SECRET intimate confession",
                    external_id="e1", tier="intimate",
                    privacy_tier_ceiling=TierCeiling.ALL)
# -> status=ok, fragment_id=frag-02ae40155b73, action=created, tier=intimate

journal_ingest_tool(vault_path=v, content="benign public text",
                    external_id="e1", tier="open",
                    privacy_tier_ceiling=TierCeiling.OPEN)
# -> status=ok, fragment_id=frag-02ae40155b73, action=updated, tier=open
```

Resulting fragment on disk:

- `privacy_tier: intimate` — **preserved** (the escalate-only privacy merge holds;
  the tier is *not* downgraded, which is the good news)
- the intimate body — **gone**, replaced by the `open`-ceiling caller's text

So a caller at `privacy_tier_ceiling=open` destroyed the contents of an
`intimate` fragment it was never admitted to read. `creek.purge.*` requires
`CREEK_MCP_ELEVATED_TOKEN` for exactly this kind of destruction
(`docs/mcp.md` §"Elevated-authorization model"); `creek.journal` achieves a
targeted version of it with no token and at the most restrictive ceiling.

## Bounding (this is narrow, and that is why it is a separate issue, not a fix
inside #932)

`external_id` is chosen by the calling client, and Adepthood is the only journal
producer today, so in practice a caller can only clobber its own entries. The
exposure is real in the multi-consumer / shared-vault case the remote transport
(`docs/mcp.md` §remote) already contemplates, where a `personal`-capped remote
consumer and a local `all` consumer share one vault.

[...]

Issue #969, opened 2026-07-29
creek.state.read / creek.state.render have no tier-ceiling gate; state.read's docstring asserts one that does not exist


Geoffe-Ga:
## Context

Found by the #932 sweep ("audit every MCP tool that reads vault fragments for a
source-side tier-ceiling gate"). `creek.state.read` is the clearest remaining
instance of the exact #846/#848 defect signature: **a docstring asserting a
ceiling rule the code does not implement.**

`creek_mcp/tools/state_read.py:27-35` says:

> the tier-ceiling enforcement here is the gate on whether the report itself can
> be returned at all, not a body-level filter

There is no such enforcement anywhere in the module. `state_read_tool` audits
the call, then returns `content = (vault / "00-Creek-Meta/State/latest.md")
.read_text()` verbatim. `privacy_tier_ceiling` is echoed and never consulted.
`creek.state.render` (`creek_mcp/tools/state.py`) is the same: it re-renders and
returns the full `content`, with no ceiling threaded into
`StateReportGenerator`.

The #932 sweep is filing this rather than fixing it inline because the correct
fix is a design decision, not a comparator — see below.

## What is actually in the state report

I verified `creek/generate/state.py` contains **no** privacy filtering at all.
Its sections draw on:

- **Eddy / Thread titles** (`section_active_eddies`, `section_active_threads`),
  liminal entry names, synchronicity rows — unfiltered.
- **Mined seed prompts** (`section_suggested_questions` → `phase_filtered_seeds`)
  — called with **no** override, which `filter_fragments_by_tier` treats as
  equivalent to `OPEN` (intimate dropped, personal bodies summarised). So this
  path is **safe by default**; it is not the gap.

I probed a temp vault holding one `intimate` and one `personal` fragment with
canaries in title, tags and body: neither canary appeared in the `state.render`
or `state.read` response. So there is **no demonstrated body-level leak today** —
the exposure is structural, at the title level, and blocked on the sub-issue
below.

### 1. The cross-ceiling cache

`latest.md` carries no record of the ceiling it was rendered under. It may have
been produced by an MCP caller at `ceiling=all`, or by the CLI with
`--include-tier all`, and is then served verbatim to a caller at
`ceiling=open`. A per-call gate on `state.read` has nothing sound to compare
against, because the artifact does not say what is in it.

The fix is to **stamp the render-time ceiling (and the maximum content tier
actually admitted) into the state artifact**, then gate read-back on that stamp.
[...]

PR #973, opened 2026-07-29
sweep(mcp): machine-checked read-posture manifest for the tier-ceiling surface

### LINKED ISSUE #932: Sweep: audit every MCP tool that reads vault fragments for a source-side tier-ceiling gate

Geoffe-Ga:
## Summary

Issue #932 asked for a sweep: audit every MCP tool that reads vault fragments the caller did **not** supply inline, confirm each checks the resolved tier against `privacy_tier_ceiling`, and — the part that actually pays — build a guardrail so the next such tool fails the suite until it is triaged.

**The audit record is now code.** `creek_mcp/read_gate.py` carries `TOOL_POSTURES`: one entry per registered MCP tool, each naming its gate or justifying why the ceiling cannot be reached. `tests/test_mcp_read_gate.py` verifies it in six layers, so the manifest cannot lie.

Five gaps were found and **filed, not fixed inline** (per the issue author's own instruction): #968, #969, #970, #971, #972.

### The tool-by-tool table

The surface was enumerated from `build_server()` at runtime via `await server.list_tools()` — 23 tools — not from the issue's candidate list, which was incomplete (it omitted `classify`, `state.render`, `skills.refresh`, `redact.scan`, `handshake`, and the five `purge.*`).

| Tool | Reads unsupplied vault content? | Gate / why not reachable | Verdict |
|---|---|---|---|
| `creek.reflect` | yes (`entry_ref`) | `_above_ceiling`, ordered above the care seam | done (#846) |
| `creek.compile` | yes (`fragment_ids`) | `_survey_sources` + `write_tier_allowed` | done (#848) |
| `creek.wheel` | yes | `to_privacy_override` → `tier_within_override` | gated |
| `creek.mine` | yes | override → `IdeaMiner` → `filter_fragments_by_tier` | gated |
| `creek.draft` | yes | same override → `IdeaMiner` **and** `DraftGenerator` | gated |
| `creek.author` | yes | override → `run_author`/`plan_author` (#660) | gated |
| `creek.report` | **yes** | **none** — ceiling audited and echoed, never converted | **gap → #968** |
| `creek.state.read` | **yes** (cached artifact) | **none**, and the docstring asserted one | **gap → #969** |
| `creek.state.render` | **yes** | **none**; generator takes no override | **gap → #969** |
| `creek.skills.refresh` | yes | partial: intimate excluded by generator hardcode; ceiling never threaded | **gap → #971** |
| `creek.redact.scan` | **yes** (caller-named, unconfined) | path confined to the *whole vault*, tier never consulted | **gap → #972** |
| `creek.journal` | read: no | write: update-in-place ignores the existing fragment's tier | **gap → #970** |
| `creek.lint` | yes — **frontmatter only** | response returns only name + count-shaped summary + `len(findings)`; titles live in `findings`, never returned | metadata-only |
| `creek.link`, `creek.classify` | yes | counts only; no new tiered content | metadata-only |
| `creek.handshake` | no | tool names + versions | n/a |
| `creek.save`, `creek.ingest` | no | caller's content; `write_tier_allowed` gates the created tier | write gate |
| `creek.purge.*` (5) | no | no ceiling param by design; `CREEK_MCP_ELEVATED_TOKEN` | auth-token |

### What the sweep actually found

Most of the remaining exposure is **write-side** — a tool *acts on* content above the caller's ceiling rather than handing it back. That shape is invisible to `docs/mcp.md`'s ceiling language, which is written entirely in read-back terms ("omitted", "returned as a title-only stub"). That is why these survived.

Reproduced, not inferred:
- `creek.report(report_type="tags")` at `ceiling=open` wrote an **intimate** fragment's tag verbatim into `00-Creek-Meta/Tag-Garden.md` and `Processing-Log/tag-history.json`. Four of six report generators contain zero tier filtering.
- `creek.journal` at `ceiling=open` **destroyed the body** of a fragment persisted at `privacy_tier: intimate`. The tier itself was not downgraded (the escalate-only merge held) — the content was.
- `creek.redact.scan` at `ceiling=open`, aimed at `01-Fragments`, returned `01-Fragments/Notes/my-affair-with-dana.md`. Creek filenames are slugified titles.

### Two corrections to the brief I was given

1. **#932's "prior art" section names `_sources_above_ceiling`. No such symbol exists** — `compile.py` defines `_survey_sources`, which returns a `_SourceGate` the caller branches on. Its docstring explains it is *deliberately* not named as a predicate, so renaming it would reintroduce the bug that comment prevents. Pinned to the real name.
2. **This PR overturns its own conclusion.** It initially concluded no unaudited tool leaks above-ceiling content into an MCP *response*, having probed the tools that walk the corpus themselves. `creek.redact.scan` was set aside as "the caller named the path" — true, and not sufficient. Gate-2.5 review caught it; `docs/mcp.md` no longer makes the claim.

### Proof the guardrail is not vacuous

Six mutations injected, each run, each reverted. Working tree verified clean after every one.

| # | Mutation | Result |
|---|---|---|
| 1 | Added `@server.tool(name="creek.dummy")` reading a fragment by id and returning its body, no tier check | **2 fail** (a) |
| 2 | Pointed reflect's `gate_symbol` at a nonexistent name | **3 fail** (c) |
| 3 | Kept the symbol, removed the `if _above_ceiling(...)` call site | **1 fail** (c) |
| 4 | Removed `gap_issue=968` | **2 fail** (d) |
| 5 | **wheel keeps its gate call and adds a parallel ungated read** | **1 fail** (f) |
| 6 | **call-and-discard**: bare `_above_ceiling(t, c)`, fall through | **3 fail** (c)+(f) |

Mutations 5 and 6 are the ones that matter. Before this PR's final commit, **both passed all 101 tests — and mutation 5 also passed `tests/test_mcp_wheel.py`.** Structural checks prove a gate is declared, exists, and is invoked; they cannot prove it is the *only* path to the corpus, and they cannot tell a decision from a gesture. Layer (f) (runtime canary probe, manifest-driven coverage) and the `_calls_symbol` load-bearing-value check close those. `creek.draft` and `creek.author` are probe-exempt with written reasons, which is why the (c) fix matters independently — mutation 6 was verified against `creek.author` specifically.

### Deliberately not done

- No behaviour change to any tool. Every non-test edit outside `read_gate.py` is inside a docstring or markdown.
- `reflect` and `compile` are **not** retrofitted onto the new primitives. Their refusal reasons and ordering are tool-specific and load-bearing — reflect's ACCEPTED RESIDUAL RISK block documents why its reason must stay distinct from `entry_ref not found`, including a timing-equalisation caveat that would have to be handled if they were ever unified. Recorded in the module docstring as a decision.
- Not re-filed: #958, #962, #931, #964, #961, #928, #923.
- Not filed: `creek.link` writing cross-tier wikilinks (core-pipeline, adjacent to #931); `Eddy`/`Thread` having no `privacy_tier` (folded into #969, since it is the same fix).

### Docstrings corrected against their tracking issues, not softened

Four asserted ceiling behaviour the code does not implement — the literal #846/#848 signature:
- `state_read.py`: *"the tier-ceiling enforcement here is the gate"* — there is none (#969).
- `skills.py`: *"safe at any ceiling"* — intimate is excluded by a generator hardcode; personal bodies pass unsummarised (#971).
- `lint.py`: *"reads only compiled-layer metadata"* — `paradox`/`unnamed`/`orphan-compiled` read `01-Fragments`. **No issue filed**: the conclusion holds for a better reason, independently verified against all 13 `CheckResult` constructions.
- `report.py` and `journal.py` said nothing at all, which is how their gaps survived.

Also filed a note on #969 that `tests/test_mcp_tools.py:116::test_state_read_does_not_embed_fragment_bodies` is **vacuous** — it hand-writes `latest.md` as a fixed 3-line string, so its assertion is true by construction with or without a gate. A test asserting the ceiling property, beside a docstring asserting it, over code implementing neither.

## Test plan

- `./scripts/check-all.sh` → **12/12 passed**, 6521 tests, 93.99% branch coverage, `read_gate.py` 90.38% per-file. Run with credentials stripped.
- `tests/test_mcp_read_gate.py` → 116 passed (Gate 1 RED confirmed first: `ModuleNotFoundError: No module named 'creek_mcp.read_gate'`).
- Test integrity: `git diff --numstat origin/main -- creek-tools/tests/` → `1476 0` — one new file, **zero deletions, no existing test touched**. One post-RED rewrite by the test specialist was diffed and accepted: it removed no assertion and *added* one (`assert as_intimate is None`, making an equality comparison non-vacuous).

Closes #932

### DIFF
--- a/creek-tools/creek_mcp/read_gate.py
+++ b/creek-tools/creek_mcp/read_gate.py
@@ -296,12 +296,18 @@ TOOL_POSTURES: dict[str, ToolPosture] = {
         rationale=_COUNTS_ONLY_RATIONALE,
     ),
     "creek.redact.scan": ToolPosture(
-        posture=ReadPosture.CALLER_NAMED_PATHS,
+        posture=ReadPosture.UNGATED_KNOWN_GAP,
         rationale=(
-            "The caller names the path, which resolve_within_vault confines "
-            "to the vault; findings carry counts, line numbers and salted "
-            "hashes, never the matched text."
+            "The caller names the path and resolve_within_vault confines it to "
+            "the vault, and no matched text is returned — but the confinement "
+            "is to the whole vault, not to the FEAT-027 staging subtree, and "
+            "the tier is never consulted. An open-ceiling caller pointed at "
+            "01-Fragments gets back the filenames of intimate fragments, which "
+            "are slugified titles, plus which PII types each contains. This "
+            "posture was CALLER_NAMED_PATHS until the #932 review; the name "
+            "was true and the conclusion drawn from it was not (#972)."
         ),
+        gap_issue=972,
     ),
     "creek.save": ToolPosture(
         posture=ReadPosture.NO_UNSUPPLIED_READ,
--- a/creek-tools/creek_mcp/read_gate.py
+++ b/creek-tools/creek_mcp/read_gate.py
@@ -1,0 +1,426 @@
+"""Read-surface posture manifest + the two canonical ceiling gates (#932).
+
+:data:`TOOL_POSTURES` is the audit record this issue asked for: one entry per
+registered MCP tool, stating how that tool relates to the caller's
+``privacy_tier_ceiling``. A tool either enforces the ceiling through a named
+gate (``GATED``), returns nothing the caller did not supply
+(``NO_UNSUPPLIED_READ``), reads only paths the caller named
+(``CALLER_NAMED_PATHS``), returns count-shaped metadata rather than content
+(``METADATA_ONLY``), is gated by an elevated auth token instead
+(``AUTH_TOKEN``), or does not honour the ceiling at all and says so against a
+tracking issue (``UNGATED_KNOWN_GAP``). The record is worth keeping because a
+tool nobody triaged is otherwise indistinguishable, to a reader of the
+surface, from a tool somebody decided needs no gate.
+
+**Gate ordering.** The discipline is worked out at length in
+:mod:`creek_mcp.tools.reflect`, around its ceiling gate and the comment blocks
+either side of it. It is restated — not copied — here, because every adopter
+of the primitives below inherits it:
+
+1. Audit-log the attempt first and unconditionally, recording only ``has_*``
+   booleans. Never the probed target id and never the outcome, so the trail
+   cannot answer "did consumer X read fragment F?" in either direction; a
+   probing consumer shows up as a rate, never as a named target.
+2. Resolve the target and answer *not found* before anything else touches it.
+3. Run the ceiling gate **above** every derived-signal seam — care guards,
+   grounding retrieval, the model. Any signal downstream of the gate is a
+   one-bit oracle about content the caller is not admitted to: an
+   ``escalate`` response tells an unadmitted caller that the fragment carries
+   acute-distress markers just as surely as returning the body would.
+4. Keep the refusal reason generic (:data:`GENERIC_ABOVE_CEILING_REASON`) and
+   never echo the resolved tier. That tier was derived from content the
+   caller cannot read, so echoing it — in the reason or in an extra key added
+   "for debugging" — turns every refusal into a tier-classification oracle
+   over the corpus.
+5. Only *after* admission derive :func:`creek_mcp.tier_ceiling.routing_tier`
+   to key a model call.
+
+**Why these primitives have no production caller yet.** That is a decision,
+not an oversight. ``creek.reflect`` and ``creek.compile`` are deliberately
+*not* retrofitted onto :func:`refuse_above_ceiling`: their refusal reasons and
+ordering comments are tool-specific and load-bearing. Reflect's "ACCEPTED
+RESIDUAL RISK" block documents exactly why its reason must stay distinct from
+``entry_ref not found`` — including the timing-equalisation caveat that would
+have to be handled if the two were ever unified — and folding it into a shared
+reason would delete that reasoning along with the behaviour. The primitives
+exist so the four filed gaps (#968/#969/#970/#971) can be closed by *adoption*
+rather than by each tool re-deriving the policy, and so that a *new* tool
+finds an already-gated corpus walk on the path of least resistance.
+
+:data:`TOOL_POSTURES` is verified by ``tests/test_mcp_read_gate.py``, which
+fails if an entry lies: the manifest must match ``server.list_tools()``
+exactly, a ``GATED`` claim is checked against a real call site in the named
+module, a gap must name a positive issue that the tool's own source mentions,
+and a tool recorded as a gap must not already call a primitive below.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import TYPE_CHECKING
+
+from creek.classify.privacy_filter import filter_fragments_by_tier
+from creek.vault.reader import iter_vault_fragments
+from creek_mcp.tier_ceiling import refusal_response, tier_allowed, to_privacy_override
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from pathlib import Path
+
+    from creek.models import Fragment, PrivacyTier
+    from creek_mcp.tier_ceiling import TierCeiling
+
+_FRAGMENTS_SUBDIR = "01-Fragments"
+
+GENERIC_ABOVE_CEILING_REASON = "resolved content exceeds the declared tier ceiling"
+"""The one refusal reason for an above-ceiling read.
+
+Deliberately names no tier and no fragment. It has to be identical across
+every above-ceiling tier at a given ceiling, or the refusal itself ranks the
+content the caller was not admitted to.
+"""
+
+
+class ReadPosture(StrEnum):
+    """How a tool relates to the caller's ``privacy_tier_ceiling``.
+
+    Attributes:
+        NO_UNSUPPLIED_READ: Returns nothing the caller did not supply, so
+            there is no unadmitted content for a ceiling to gate.
+        CALLER_NAMED_PATHS: Reads only vault paths the caller named, and
+            returns no matched content from them.
+        GATED: Enforces the ceiling through the ``gate_module`` /
+            ``gate_symbol`` pair recorded alongside it.
+        METADATA_ONLY: Returns counts, check names and other count-shaped
+            summaries rather than fragment bodies or titles.
+        AUTH_TOKEN: Takes no ceiling at all; gated fail-closed by an elevated
+            token instead, because it destroys content rather than reading it.
+        UNGATED_KNOWN_GAP: Reads vault content without honouring the ceiling.
+            A tracked, reproduced gap — ``gap_issue`` names the issue.
+    """
+
+    NO_UNSUPPLIED_READ = "no-unsupplied-read"
+    CALLER_NAMED_PATHS = "caller-named-paths"
+    GATED = "gated"
+    METADATA_ONLY = "metadata-only"
+    AUTH_TOKEN = "auth-token"
+    UNGATED_KNOWN_GAP = "ungated-known-gap"
+
+
+@dataclass(frozen=True)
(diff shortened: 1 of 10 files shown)


Geoffe-Ga:
## Summary
This PR closes out issue #932 by adding `creek_mcp/read_gate.py`, a machine-checked manifest (`TOOL_POSTURES`) recording, for each of the 23 registered MCP tools, how it relates to the caller's `privacy_tier_ceiling` — plus two adoptable primitives (`refuse_above_ceiling`, `iter_admitted_fragments`) that gaps can migrate onto later. No tool behavior changes; every non-test edit outside the new module is a docstring/doc addition. Five real, reproduced ceiling gaps were found and filed as separate tracked issues (#968–#972) rather than fixed inline, per the issue's own instruction.

## Strengths
- The manifest is genuinely hard to falsify: layers (a)-(e) verify surface completeness, gate symbol existence *and* invocation (via AST, including a "call-and-discard" trap), and gap-issue integrity against the tool's own source — not just the manifest.
- Layer (f) (runtime canary probes) is the standout: it calls each `GATED` tool for real against a vault with sentinel-tagged intimate content and asserts the sentinel never appears in the serialized response. The PR documents having run this against a live mutation (a parallel ungated read added next to an untouched, still-passing gate) and shows layers (a)-(e) stay green while only (f) catches it — a concrete demonstration that the structural checks alone are insufficient and that this layer earns its keep.
- Every citable claim in the PR body checks out on inspection: all 23 tool names match `server.list_tools()`, all six `GATED` `(module, symbol)` pairs are real and invoked (verified directly against `reflect.py`, `compile.py`, `wheel.py`, `mine.py`, `draft.py`, `author.py`), and issues #968–#972 exist on GitHub with descriptions matching the manifest's stated rationale.
- Honest, well-scoped writeup: the PR explicitly documents overturning its own initial conclusion about `creek.redact.scan` after Gate-2.5 review, and corrects a wrong symbol name (`_sources_above_ceiling`) inherited from the issue's own "prior art" section rather than silently working around it.
- `refuse_above_ceiling` delegates to `tier_allowed` rather than re-deriving admission logic, avoiding a second private ranking that could drift from the rest of the MCP surface — and the refusal payload is tested to never echo the offending tier (avoiding a tier-classification oracle).

## Security Concerns
🟢 LOW — The PR itself introduces no new leak; it is a pure audit/manifest layer. The five real gaps it documents (`creek.report` write-side leak, `creek.state.read`/`render` cross-ceiling cache, `creek.skills.refresh` unsummarized personal bodies, `creek.journal` update-in-place tier bypass, `creek.redact.scan` filename disclosure) are genuine and currently unmitigated in production, but that is explicitly the scope issue #932 asked for ("audit, then file — don't fix inline"), and all five are filed with reproduction details. Worth confirming the team tracks #968–#972 to closure with the same rigor applied here.

## Problems
None found blocking merge.

## Code Quality
- `iter_admitted_fragments` re-filters fragments one pair at a time (`filter_fragments_by_tier([(fragment, body)], override=...)`) rather than passing the whole sequence — the docstring explains this is intentional (to keep path-to-body attribution correct across the filter's fragment-dropping behavior), which is a reasonable and clearly justified tradeoff, not a defect.
- The PR title `sweep(mcp): ...` uses a non-standard conventional-commit type (`sweep` isn't in the default `feat/fix/chore/docs/refactor/test/...` set enforced by `conventional-pre-commit`). The three actual commits (`test(mcp):`, `feat(mcp):`, `fix(mcp):`) are all valid conventional types, and no CI job appears to lint the PR title itself, so this only matters if the squash-merge commit message is taken verbatim from the PR title — worth a quick title edit before merge to keep history consistent.
- `read_gate.py` sits at 90.38% per-file coverage per the PR's own numbers, just above the repo's 90% floor — not a blocker, but the thinnest-covered branches (likely rare AST edge cases) may be worth a follow-up glance if coverage tightens later.

## CLAUDE.md Adherence
✅ Coverage — 93.99% branch overall, `read_gate.py` at 90.38% (above the 90% floor, per PR-reported `check-all.sh` run).
✅ Docstrings — every new/changed function and module carries the repo's characteristic dense, rationale-bearing docstrings; interrogate gate reported passing.
✅ Types — `from __future__ import annotations`, full type hints including `TYPE_CHECKING`-guarded imports, consistent with strict-mypy conventions elsewhere in the codebase.
✅ Patterns — new primitives delegate to existing `tier_ceiling` helpers rather than re-implementing policy, matching the codebase's stated "delegate, don't re-derive" discipline.
⚠️ Commits — individual commits are conventional; PR title itself uses a non-standard `sweep(...)` type (minor, see Code Quality).

## Testing
Exceptionally strong for this kind of change. `tests/test_mcp_read_gate.py` (1476 new lines, zero deletions/modifications to existing tests) implements six independent verification layers against the manifest, including AST-based anti-gaming checks and live runtime probes with three-location sentinel fragments (title/body/tags) to catch leaks in any response shape. The PR documents a concrete mutation-testing exercise (6 injected mutations, each reverted) proving the guardrail actually fails when the property it claims to check is violated — including one mutation that passed all 101 pre-existing tests and would have shipped a real leak undetected before this PR's final layer was added. This is well above the bar of "insufficient testing is blocking" — it's a positive example of the pattern.

## Documentation
`docs/mcp.md` is updated accurately: the `creek.report`, `creek.skills.refresh`, and `creek.journal` table rows now correctly describe the known gaps instead of the previously-inaccurate "any ceiling permitted" / "already excluded" claims, and a new "Read-side posture (#932)" section explains the write-side vs. read-side distinction that let these gaps hide. Per-tool module docstrings were corrected in place to stop asserting ceiling enforcement the code doesn't implement — a real documentation-accuracy fix, not just narrative padding.



## Verdict: LGTM

Every checkable claim in the PR body verifies against the actual code: all 23 registered MCP tools are covered by TOOL_POSTURES (matches server.list_tools() exactly), every GATED entry's (module, symbol) pair is a real, invoked call site, and all five gap issues (#968-#972) exist on GitHub with descriptions matching the manifest's rationale. The change is provably behavior-preserving — every non-test, non-read_gate.py diff hunk is a pure docstring addition — while the new test suite is unusually rigorous (structural + AST anti-gaming + runtime canary probes) with documented mutation-testing evidence. Only minor nits (PR-title type, threshold headroom) remain; nothing here blocks merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5113614062 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #974, opened 2026-07-29
fix(pipeline): creek process never runs the post-classification privacy reassess — confessional fragments stay PERSONAL and voice-proxy eligible


Geoffe-Ga:
## Role

You are closing a **fourth** instance of the recurring "wired into one entry point but not the other" defect class in the Creek classification pipeline. This one is security-adjacent: it leaves intimate content voice-proxy eligible.

## The defect

`creek/classify/privacy_pass.py::reassess` exists so that classification hardening the fragment's voice signals can *raise* its privacy tier before the write. `PrivacyClassifier.classify_tier`'s third INTIMATE trigger is "self-authored + confessional register + conviction confidence" (`creek/classify/privacy.py:140`), and that signal only exists **after** the classifier has run.

`creek/classify/classify_engine.py` calls it — `_process_file` L919-920, `if prepared.owns_tier: new_fragment = reassess(...)`, the last mutation before the write.

`creek/pipeline.py` **never calls `reassess`**. `_run_classification` stamps the tier with `apply_tier` at L560 — deliberately *before* the LLM dispatch at L571, because the per-tier router must see a real tier (#876, Intimate-never-cloud) — and then never looks again.

## Reproduction (verified at f72dfa7)

Drive `Pipeline._run_classification` with a self-authored `markdown` fragment whose rules pass leaves it unclassified and whose LLM classifier returns `voice_register: confessional`, `confidence: conviction`:

```
pipeline tier          : personal
pipeline voice         : voice_register='confessional' confidence='conviction'
pipeline voice_proxy_eligible: True

what the engine's reassess() would give:
                       : intimate
       voice_proxy_eligible: False
```

## Consequence

The same fragment, same body, same model verdict:

- via `creek classify --method llm` → `privacy_tier: intimate`, `voice_proxy_eligible: false`
- via `creek process` → `privacy_tier: personal`, `voice_proxy_eligible: true`

So a confessional fragment ingested through the one-shot command stays admissible to a PERSONAL-ceiling MCP caller and keeps feeding voice-proxy generation. Which command the user happened to run decides whether the tier is right.

## Approach

[...]

PR #975, opened 2026-07-29
fix(classify): run AudienceClassifier in creek process (#937)

### LINKED ISSUE #937: fix(classify): AudienceClassifier never runs in creek process — audience axis populated by classify only
## Role

You are closing the third instance of a recurring defect class in the Creek classification pipeline: a classifier that exists, works, and is wired into one entry point but not the other.

## Goal

Wire `AudienceClassifier` into `creek/pipeline.py::_classify_fragments` so `creek process` populates the audience axis (#634), matching what `creek classify` already does.

## Context

`creek/classify/classify_engine.py` imports and instantiates `AudienceClassifier` (lines 93, 387, and it is threaded through `_process_file` at line 734). `creek/pipeline.py` contains **zero** references to it. The consequence: `creek classify` stamps the audience axis, `creek process` silently does not — so which command a user runs determines whether the field is populated at all.

This is the same shape as two bugs fixed today:

- **#876 / PR #935** — `PrivacyClassifier` had no producer, so every fragment stayed `privacy_tier: unclassified` and downstream filters treated that as OPEN. Fixed with `creek/classify/privacy_pass.py` wired into `cli.py`, `pipeline.py`, and `creek_mcp/tools/classify.py`.
- **#877 / PR #936** — `praxis_potential` had no producer, so `04-Praxis/` and `08-Decisions/` were structurally unreachable. Fixed with `creek/classify/praxis_pass.py` following the same wiring.

Two specialists on the #877 lane independently flagged this gap; it was observed but deliberately left unfiled to keep that PR scoped.

## Approach

Read `privacy_pass.py` and `praxis_pass.py` and their `pipeline._classify_fragments` call sites first — the wiring pattern is established twice over, and a third bespoke solution would be waste. The likely fix is smaller than either predecessor because `AudienceClassifier` already exists and is proven in `classify_engine`; this is a missing call, not a missing producer.

Confirm before building whether `creek process` and `creek classify` should produce byte-identical frontmatter for the same input. If yes, say so in the PR body and assert it in a test — that invariant is what makes this class of bug detectable in future.

## Format

One PR closing this issue, following the four gates in `.claude/agents/shared/house-rules.md`.

## Constraints

- TDD: a failing test first, demonstrating that `creek process` leaves the audience axis unpopulated where `creek classify` sets it.
- Do **not** regress the #876 tier→router ordering or the #877 `escalate` chokepoint. PR #936 added a regression test pinning the tier ordering so it cannot be hoisted — keep it passing.
- Meet the repo thresholds: >=90% branch coverage, per-file >=80%, >=95% docstring, complexity <=10, mypy strict, ruff clean.
- No bypasses. No `# noqa` / `# type: ignore` / `@pytest.mark.skip` without an `Issue #N` justification.
- If fixtures in unrelated suites break because a previously-empty field is now populated, fix the **fixtures** to declare what they need — do not lower a threshold or weaken an assertion. (This happened on #876 across author-desk, voice, state, and synthesis suites.)

Refs #634, #876, #877

Geoffe-Ga:
## Summary

`AudienceClassifier` (the #634 audience axis) was wired into `creek/classify/classify_engine.py` but had **zero** references in `creek/pipeline.py`. So `creek classify` stamped the axis and `creek process` silently left every fragment at the `Fragment` model default `audience: mixed` — which on disk is indistinguishable from a fragment the classifier genuinely weighed and found ambiguous. Which command the user happened to run decided whether the field was populated at all.

That default is not inert. `_default_audience_authority` (`creek/config.py:1432`) weights `audience-facing` at **2.0** and `private` at **0.1**; `mixed` is **1.0**, neither. So a process-built vault weighs a published essay and a 3 a.m. journal entry at identical authority, and the voice fingerprint learns the average instead of the voice — silently, with nothing raised and nothing logged.

Third instance of one defect class, after #876 (`PrivacyClassifier`) and #877 (`praxis_potential`). Unlike both predecessors it needed no new policy module: `AudienceClassifier.classify_and_enforce` already existed and was proven in the engine, so this is a **missing call, not a missing producer**.

`creek/pipeline.py`, +26/−0:
- `AudienceClassifier()` built once in `Pipeline.__init__`, bare — exactly as `classify_engine.py:439` builds it. No `llm_assist`, so the rules path needs no provider and `process --no-llm` stays egress-free.
- One unconditional `frag = self.audience_classifier.classify_and_enforce(frag, item.body)` in `_run_classification`, after `apply_tier` and the maybe-LLM dispatch, before `apply_praxis` — mirroring the engine's `_classify_one` → audience → `apply_praxis` order.

Ordering is load-bearing and is pinned by a test: `AudienceClassifier.score` reads `fragment.privacy_tier` (`creek/classify/audience.py:160`), so hoisting the call above `apply_tier` would score every fresh fragment against `unclassified` and return different verdicts than `creek classify` does.

## The invariant question — settled, with evidence

The issue asked whether `creek process` and `creek classify` should produce **byte-identical frontmatter**. **As literally worded, no** — and I would rather pin the true contract than assert one that isn't true:

- `run_classify` adds `classification_method` and `classified_at`, which are not `Fragment` model fields at all (`_write_fragment`, `classify_engine.py:1564-1583`).
- YAML key ordering and timestamp round-tripping make a bytes comparison meaningless regardless.

The true, stronger, testable contract — and the one this PR asserts:

> For a vault written by `creek process --no-llm` from freshly-ingested, untiered fragments, a subsequent `creek classify --method rules --force` changes **no `Fragment` model field**. The only frontmatter delta is the *addition* of `classification_method` and `classified_at`.

I measured this both ways with a guard-free probe over three fragments (essay / journal / markdown):

- On `origin/main` @ f72dfa7 the complete delta was `audience` on all three (`mixed` → `audience-facing` / `private` / `private`), plus the two provenance keys. **Nothing else moved** — `privacy_tier`, `praxis_potential`, `frequency`, `wavelength`, `voice`, `voice_proxy_eligible` were already identical. The invariant already held for #876 and #877; `audience` was the last hole.
- With this fix the delta is **exactly** the two provenance keys.

`tests/test_process_classify_parity.py` pins it. The comparison is over the **whole parsed frontmatter dict minus a documented two-key exclusion set** — it names no axis — so a *fourth* classifier wired into one entry point and not the other fails this test automatically, without anyone remembering to extend a list. That is the point of the issue, and it is why this PR is larger than the two-line fix it could have been.

Preconditions are established deliberately in the test and documented in its module docstring: `no_llm=True` + `--method rules` (with `--force`, a rules re-run would legitimately overwrite LLM-derived `frequency`/`wavelength`/`voice` — an LLM-vs-rules disagreement is not this defect class); fragments enter untiered; parsed metadata, not bytes.

## The third call site — checked, and it needs nothing

#876 touched `cli.py`, `pipeline.py` and `creek_mcp/tools/classify.py`, so I checked the MCP surface rather than silently fixing two of three:

- `creek_mcp/tools/classify.py` **wires no classifier**. It calls `run_classify`, which already stamps audience, so it is correct by delegation. Its #876/#877 changes only surfaced *counts* (`privacy_tiers_assigned`, `praxis_marked`). I deliberately added no `audience_marked` count: unlike those two, audience is stamped unconditionally on every (re)classified fragment rather than escalate-only, so the count would always equal `classified` and carry no information.
- No MCP tool wraps `Pipeline` — there is exactly one construction site, `creek/cli.py:1152` (the `process` command).
- `creek/vault/writer.py::_retier_after_rewrite` (#922) re-derives only `privacy_tier` on a material edit; a stale `audience` there awaits the next classify pass exactly as a stale `frequency` does. Pre-existing uniform posture, not a #937 hole.

## Follow-up filed: #974

While verifying the invariant I found a genuine **fourth** instance of this defect class, and reproduced it:

`creek/pipeline.py` never calls `privacy_pass.reassess`, which the engine runs at `classify_engine.py:919-920`. A self-authored fragment the model reads as confessional + conviction ends:

- via `creek classify --method llm` → `privacy_tier: intimate`, `voice_proxy_eligible: false`
- via `creek process` → `privacy_tier: personal`, `voice_proxy_eligible: **true**`

So intimate content ingested through the one-shot command stays voice-proxy eligible and admissible to a PERSONAL-ceiling MCP caller. That is security-adjacent and deliberately **not** fixed here for scope discipline — filed as **#974** with the reproduction.

## Test plan

- `./scripts/check-all.sh` → **exit 0**, 12/12 checks, **6524 passed**, **93.99%** branch coverage, per-file gate green (218 files ≥ 80%).
- `./scripts/test.sh --all` (integration + e2e) → **6549 passed, 5 skipped** (all env-gated: no Ollama, no API keys). **Zero fixture fallout** — no existing test needed changing, unlike #876.
- Gate 1 RED proven before the fix, with the exact failures:
  - `test_no_llm_run_stamps_the_audience_axis` — `{'…pipe01': 'mixed', '…pipe02': 'mixed'}` != `{'…pipe01': 'audience-facing', '…pipe02': 'private'}`
  - `test_audience_is_scored_after_the_privacy_tier_is_stamped` — `assert 'mixed' == 'private'`
  - `test_rules_classify_after_process_changes_no_fragment_field` — `('intimate', 'explicit', 'mixed')` != `('intimate', 'explicit', 'private')`
- Test integrity: `git diff --numstat origin/main -- tests/` = **165/0 and 255/0 — zero deletions**. No existing test was modified, reformatted or removed; `tests/test_pipeline.py` is a pure append.
- Regressions explicitly re-run green: `TestPipelineTierRouting::test_untiered_journal_fragment_routes_local_and_gets_a_tier` (#876 tier→router ordering), `TestPipelinePraxisPass` (#877 escalate-only chokepoint), all of `tests/test_classify_engine.py` and `tests/test_audience.py`.
- Complexity unchanged: `_run_classification` stays **B (6)**; the new statement is unconditional.
- No bypasses: no `# noqa`, no `# type: ignore`, no skip marker, no threshold change, no deleted test or code.

New tests:
- `tests/test_pipeline.py::TestPipelineAudiencePass` — a two-outcome stamping test (essay → `audience-facing`, short journal → `private`, so a stamp-everything implementation cannot pass) and an ordering pin whose fixture returns `private` only when the tier is stamped first, and `mixed` if the call is hoisted.
- `tests/test_process_classify_parity.py` — the cross-entry-point invariant above.

Closes #937
Refs #634, #876, #877, #974

### DIFF
--- a/creek-tools/creek/pipeline.py
+++ b/creek-tools/creek/pipeline.py
@@ -37,6 +37,7 @@ from creek.audit.yield_summary import (
     format_yield_line,
     write_yield_summary,
 )
+from creek.classify.audience import AudienceClassifier
 from creek.classify.classify_engine import build_tier_classifiers
 from creek.classify.praxis_pass import apply_praxis
 from creek.classify.privacy import PrivacyClassifier
@@ -223,6 +224,9 @@ class Pipeline:
         rule_classifier: Keyword-based classifier.
         privacy_classifier: Assigns each fragment's privacy tier before the
             per-tier router reads it (#876).
+        audience_classifier: Stamps the #634 audience axis on every fragment
+            this pipeline classifies (#937), because the classify engine is
+            not the only writer of that axis.
         tier_classifiers: Per-tier LLM classifiers (#666/#706) — the fragment's
             privacy tier selects the provider so Intimate stays local.
         review_generator: Review queue generator for uncertain fragments.
@@ -260,6 +264,12 @@ class Pipeline:
         # so without this the per-tier router below saw "not INTIMATE" for
         # every fragment and shipped journal entries to the cloud.
         self.privacy_classifier = PrivacyClassifier()
+        # Issue #937: the classify engine is not the only writer of the #634
+        # audience axis, and ``creek process`` never calls it — so this entry
+        # point needs its own instance or every fragment it ingests keeps the
+        # model default. Built bare, exactly as the engine builds it: the rules
+        # path needs no provider, which keeps ``process --no-llm`` egress-free.
+        self.audience_classifier = AudienceClassifier()
         # Per-tier routing (#666/#706): each fragment is classified by the
         # provider its privacy tier resolves to — Intimate stays local even when
         # ``classification`` is cloud. Building here is safe with any config:
@@ -525,6 +535,15 @@ class Pipeline:
         router reads that tier to keep Intimate content off cloud
         providers and a freshly-ingested fragment carries none.
 
+        Every fragment is also stamped with the #634 audience axis once the
+        tier is known (#937). ``creek process`` never calls the classify
+        engine, so without this every one-shot-processed fragment keeps
+        ``audience: mixed`` — which the voice fingerprint reads as neutral
+        (``_audience_factor``), so the user's audience-facing writing never
+        earns its weight and their private writing is never discounted. The
+        pass runs after the tier because the score reads it, and it is
+        deterministic and idempotent, so re-processing never churns the axis.
+
         Every fragment is likewise scored for praxis potential once the
         classifiers are done (#877). ``creek process`` never calls the
         classify engine, so without this the one-shot command would leave
@@ -571,6 +590,13 @@ class Pipeline:
                     frag = classifier.classify(frag, content=item.body)
             else:
                 result.deterministic_classified += 1
+            # Issue #937: stamp the #634 audience axis here, after ``apply_tier``
+            # above — the score reads ``privacy_tier``, and a fragment still
+            # ``unclassified`` contributes nothing, so an earlier call would
+            # return a different verdict than ``creek classify`` does. This
+            # placement mirrors the engine's (classify -> audience -> praxis)
+            # so the two entry points agree fragment for fragment.
+            frag = self.audience_classifier.classify_and_enforce(frag, item.body)
             # Issue #877: score the praxis axis last, so the escalate-only
             # merge sees whatever the optional LLM dispatch above produced.
             frag = apply_praxis(frag, item.body)
--- a/creek-tools/tests/test_pipeline.py
+++ b/creek-tools/tests/test_pipeline.py
@@ -1375,3 +1375,168 @@ class TestPipelinePraxisPass:
         # The rest of the same response *did* land, so the assertion above
         # cannot be passing because the LLM was never consulted.
         assert on_disk["frag-praxispipe04"]["frequency"]["primary"] == "F3"
+
+
+class TestPipelineAudiencePass:
+    """``creek process`` stamps the ``audience`` axis too (issue #937).
+
+    The #634 audience classifier reached production wired into
+    :mod:`creek.classify.classify_engine` alone. ``creek process`` never
+    calls that engine — it ingests and classifies in one shot — so every
+    fragment that entered a vault through the one-shot command kept the
+    :class:`~creek.models.Fragment` model default ``audience: mixed``,
+    which on disk is indistinguishable from a fragment the classifier
+    genuinely weighed and found ambiguous.
+
+    That default is not inert. The voice fingerprint reads the persisted
+    axis in
+    :func:`creek.generate.ai_style.fingerprint._audience_factor`, which
+    multiplies audience-facing material by ``2.0`` and private material
+    by ``0.1``. ``mixed`` is neither, so a process-built vault weighs a
+    published essay and a 3 a.m. journal entry at exactly the same
+    authority: the fingerprint learns the average instead of the voice.
+    Nothing raises and nothing is logged, which is how this same
+    wired-into-one-caller defect got through twice before — #876 (privacy
+    tiers) and #877 (praxis potential).
+    """
+
+    @staticmethod
+    def _ingested(
+        frag_id: str,
+        title: str,
+        body: str,
+        platform: str,
+        kind: str = "unclassified",
+    ):
+        """An IngestedFragment whose metadata drives the audience score.
+
+        Args:
+            frag_id: Fragment id (the assertion key).
+            title: Fragment title.
+            body: Markdown body the audience heuristic scans for
+                headings, block quotations and word count.
+            platform: ``SourcePlatform`` value — the heaviest audience
+                signal (essay/Substack positive, journal negative).
+            kind: ``SourceKind`` value; ``writing`` is a further
+                audience-facing vote. Defaults to ``unclassified``.
+
+        Returns:
+            The assembled :class:`~creek.ingest.base.IngestedFragment`.
+        """
+        from creek.ingest.base import IngestedFragment
+        from creek.models import (
+            Authorship,
+            Fragment,
+            FragmentSource,
+            SourceKind,
+            SourcePlatform,
+        )
+
+        fragment = Fragment(
+            id=frag_id,
+            title=title,
+            source=FragmentSource(
+                platform=SourcePlatform(platform),
+                author=Authorship.SELF,
+                kind=SourceKind(kind),
+            ),
+        )
+        return IngestedFragment(fragment=fragment, body=body)
+
+    def test_no_llm_run_stamps_the_audience_axis(self, vault_path: Path) -> None:
+        """A published essay and a private blip land on opposite axes.
+
+        ``no_llm=True`` guarantees Pass 3 never runs, so the verdict
+        provably comes from the free, local, deterministic heuristic in
+        :class:`creek.classify.audience.AudienceClassifier` with zero
+        egress — the point of #634 being a rules axis with an *optional*
+        LLM seam rather than one more mandatory model call.
+
+        Two fragments with two *distinct* expected outcomes, so an
+        implementation that stamps a single value over everything cannot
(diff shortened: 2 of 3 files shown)


Geoffe-Ga:
## Summary
Fixes #937: `creek process` (`Pipeline._run_classification`) never stamped the #634 `audience` axis, unlike `creek classify`, leaving every process-built fragment at the model default `mixed` — which the voice fingerprint weights identically to a genuinely ambiguous fragment. The PR adds one `AudienceClassifier` instance and one unconditional `classify_and_enforce` call, placed to match `classify_engine.py`'s existing order (tier → optional LLM → audience → praxis), plus a new cross-entry-point parity test designed to catch a future unwired axis automatically.

## Strengths
- **Placement verified correct against the sibling implementation.** I read `classify_engine.py:889-912` and confirmed the engine's actual order is `_classify_one` (rules/LLM) → `audience.classify_and_enforce` → `apply_praxis`, and that `_prepare_fragment` stamps the tier before any of that. `pipeline.py:590-601` now reproduces this exactly. The claim that hoisting the call above `apply_tier` would change the verdict is correct: `AudienceClassifier.score` reads `fragment.privacy_tier` (`creek/classify/audience.py:160`), and an untiered fragment contributes `0` from that signal.
- **Regression-class test, not just a regression test.** `tests/test_process_classify_parity.py` diffs whole parsed frontmatter (minus a documented 2-key exclusion set) rather than enumerating axes by name, so a *future* fifth unwired axis fails this test without anyone extending a list. This is exactly the right response to "third instance of the same defect class."
- **Honest scope boundary.** The PR identifiably found and documented a fourth, more severe instance of the same defect class (`privacy_pass.reassess` never called in `pipeline.py`, filed as #974) and explicitly declined to fix it here to keep this PR scoped — with a reproduction in the description rather than silently leaving it out.
- **Deterministic, provider-free by construction.** `AudienceClassifier()` is built bare (no `llm_assist`), identical to `classify_engine.py:439`, so `process --no-llm` stays egress-free — verified by grep, not just by the PR's claim.
- Docstrings on the new `audience_classifier` attribute and inline comments on both the constructor and the call site correctly explain *why* the placement matters, not just what the line does.

## Security Concerns
🟢 LOW — None found. No new I/O, no new external calls, no secrets/paths touched. The classifier is pure-rules by default (no provider egress).

## Problems
None found blocking merge.

## Code Quality
- Non-blocking: the module docstring / test docstrings are unusually long and repeat context already in the PR description (e.g., the `_audience_factor` weighting rationale appears in three places — `pipeline.py`, `test_pipeline.py`, `test_process_classify_parity.py`). Not wrong, just a lot of prose to maintain if the weights ever change. Not a blocker.
- The exclusion-set docstring in `test_process_classify_parity.py` references exact line numbers (`classify_engine.py:1564-1583`) which will drift as the file is edited — a minor future-maintenance nit, not a functional issue.

## CLAUDE.md Adherence
✅ TDD followed — PR description documents Gate 1 RED failures with exact assertion diffs before the fix.
✅ No shortcuts — grep confirms zero `noqa`, `type: ignore`, or `skip()` additions in the diff.
✅ Complexity — PR states `_run_classification` stays at Xenon B(6); the added statement is a single unconditional call with no new branch, consistent with that claim.
✅ Docstrings — Google-style, present on all new/modified public surfaces (module, fixture, test methods, attribute).
✅ Commit message — Conventional Commits format, descriptive body citing the issue.
❌/N/A — I was not able to execute `./scripts/check-all.sh` or `./scripts/test.sh` in this review environment (sandbox blocked command execution), so the PR's self-reported "12/12 checks, 6524 passed, 93.99% coverage" is taken on trust from the description rather than independently re-verified here. Static review of the diff found nothing that would plausibly fail those gates.

## Testing
Strong. Two new test classes:
- `TestPipelineAudiencePass` in `test_pipeline.py`: a two-outcome stamping test (essay → `audience-facing`, journal → `private`) that a stamp-everything implementation cannot pass, plus an explicit ordering pin (`test_audience_is_scored_after_the_privacy_tier_is_stamped`) whose fixture is engineered to disagree between the two possible call placements.
- `test_process_classify_parity.py`: the cross-entry-point invariant, comparing whole parsed frontmatter dicts minus a documented 2-key provenance exclusion, with an explicit off-default anti-vacuity guard so the test can't silently pass by comparing two vaults that both landed on defaults.

Not independently re-run in this environment, but the assertions are logically sound against the source I read, and the PR documents the RED→GREEN transition with concrete failure output.

## Documentation
Docstrings are thorough to the point of being verbose (see Code Quality), but complete and accurate — I did not find a documentation claim (constant values, line references, call ordering) that didn't check out against the actual source.



## Verdict: LGTM

The fix is minimal, correctly placed (verified against classify_engine.py's ordering: tier → LLM dispatch → audience → praxis, which pipeline.py now mirrors exactly), and is backed by a strong regression-class test (test_process_classify_parity.py) that will catch a future fourth instance of this bug without naming any axis. No bypasses, no scope creep, and the deliberately-excluded follow-up (#974) is honestly disclosed rather than silently left out.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5114043279 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #976, opened 2026-07-29
Fragment accepts naive ingested/authored_at from vault frontmatter, breaking the "never naive" contract in creek/time.py


Geoffe-Ga:
## Context

Split out of #938 (which fixes only the `CompostTracker` blast radius). Filed per that issue's "file follow-up issues for adjacent problems" constraint.

## The defect

`creek/time.py::effective_authored_at` documents this contract for its fallback:

> 2. `Fragment.ingested` — the wall-clock moment the vault wrote the fragment. **Always present, never naive.**

That contract is **not enforced**. `Fragment` has no timezone validator (`creek/models.py` has `field_validator`s for `voice_weight`, `representativeness`, `audience`, … but none for `created` / `ingested` / `authored_at`). Verified by execution at `36a4e4f`:

```python
# tz-AWARE — default_factory=now_la
Fragment(id="d1", title="D", source=..., frequency=...).ingested.tzinfo
# -> America/Los_Angeles

# NAIVE — a frontmatter timestamp serialised without a UTC offset
Fragment.model_validate({..., "ingested": datetime(2024, 1, 1, 12, 0, 0)}).ingested.tzinfo
# -> None
```

So `effective_authored_at` can return either naive or aware depending on how a fragment's frontmatter was written, and `tests/test_time.py::test_result_is_always_tz_aware` is **vacuous** — its docstring asserts "Fragment validates that `ingested` and `authored_at` (when present) are tz-aware via Pydantic's datetime parsing for both defaults and round-tripped YAML values, so this helper inherits the invariant", which is false; the test only ever feeds it tz-aware inputs, so it cannot catch the regression it claims to pin.

## Blast radius

Every `effective_authored_at` / `effective_authored_date` consumer that sorts, `min`/`max`es, subtracts, or compares can raise `TypeError: can't compare offset-naive and offset-aware datetimes` on a vault that mixes serialisation styles:

- `creek/link/threads.py:414,451,453,483,547,548`
- `creek/link/temporal.py:175,180`
- `creek/link/eddies.py:325,327,541`
- `creek/generate/synchronicity.py:257,306`
- `creek/generate/wavelength.py:691,692`
- `creek/generate/unnamed.py:262,650`
- `creek/generate/compost.py:417` — **fixed in #938** at the call site only

#938 coerces at the compost call site because widening it there was out of that issue's scope. This issue is the root-cause fix.

## Proposed fix

[...]

Issue #955, opened 2026-07-28
fix(ingest): code.py renders mtime in LA_TZ while base.file_modified_time uses UTC — identity-bearing timestamps disagree


Geoffe-Ga:
## Role

You are unifying a timezone convention that two file-based ingestors disagree on, where the value feeds fragment identity.

## Goal

Make `creek/ingest/code.py` and `creek/ingest/base.py` agree on how a file's mtime becomes a timestamp, so the same file ingested by either path yields the same instant in the same representation.

## Context

Verified against `main`:

| location | conversion |
|---|---|
| `creek/ingest/base.py:443` (`file_modified_time`) | `datetime.fromtimestamp(path.stat().st_mtime, tz=UTC)` |
| `creek/ingest/code.py:430` (local `_file_modified_time`) | `datetime.fromtimestamp(mtime, tz=LA_TZ)` — imports `LA_TZ` from `base` at line 427 |

Surfaced by the Gate 2.5 review on #911 and deliberately left untouched there as pre-existing. The reviewer's assessment — which I agree with — is that this is **not a correctness defect on its own**: a fixed epoch rendered under an explicit `ZoneInfo` is still deterministic, so `code.py` is not reintroducing the wall-clock non-determinism that #911 fixed.

What makes it worth fixing is that **PR #954 (issue #911) just made mtime the basis of generic fragment identity**. Two file-based ingestors now derive identity-bearing timestamps through different timezone lenses. The same underlying instant serialises differently depending on which ingestor saw the file, which is a latent source of confusion for anything that compares, sorts, or dedupes across ingestors.

## Approach

Decide the convention explicitly rather than defaulting to whichever is easier to change:

- **UTC everywhere** matches `base.file_modified_time`, matches what #954 established for generic, and is the conventional choice for stored timestamps.
- **`LA_TZ`** is presumably deliberate somewhere — find out *why* `code.py` diverged before erasing it. Check git history and any docs on the code ingestor. If the local-time rendering is load-bearing for a user-facing display, keep the display and normalise the stored value.

The likely fix is deleting `code.py`'s local `_file_modified_time` and calling the shared `base.file_modified_time`, but confirm the above first and say in the PR body which convention you chose and why.

## Format

One PR closing this issue, following the four gates in `.claude/agents/shared/house-rules.md`.

## Constraints

- TDD: a failing test first, asserting both ingestors produce the same instant for the same file.
[...]

Issue #882, opened 2026-07-22
feat(compost): add a production 'creek compost scan' — 10-Liminal/Compost is unreachable (only calibrate exists)


Geoffe-Ga:
## Role

You are a pipeline engineer working in `creek-tools/` (Python, mypy strict, TDD per `creek-tools/CLAUDE.md`).

## Goal

Add a production `creek compost scan` command so the FEAT-018 compost detector can actually populate `10-Liminal/Compost/` — today the folder is unreachable by any command.

## Context

- `creek compost --help` exposes exactly one subcommand: `calibrate` (scores the detector against a labelled fixture, #266). There is no command that runs detection over a real vault.
- The detection machinery exists and is production-quality: embedding-similarity gate + LLM verifier (FEAT-018/#240), provider-neutral verifier (#667), review-queue routing for ambiguous verdicts (creek/generate/compost.py:347, `review_queue_relpath: 10-Liminal/Compost/Review` in creek/config.py:921).
- `creek fill --with-compost` only writes the *overview report* (`_Compost-Report.md`, a Dataview shell over whatever compost notes exist) — with no scan command, it reports over a permanently empty folder.
- Evidence (demo vault `creek-demo-2026-06-07`, 35,330 fragments, every command run): `10-Liminal/Compost/` contains only `.gitkeep`.

## Proposed Solution

Add `creek compost scan --vault <path>`:

1. Run the embedding gate over the vault (free, local) to produce candidates.
2. Route candidates through the LLM verifier **behind the standard consent gate** (`CREEK_ANTHROPIC_CONSENT` / configured provider via the model router); support `--no-llm` to stop at the gate and emit candidates to the review queue instead.
3. Write canonical compost notes for confirmed verdicts, review-queue entries for ambiguous ones (reusing the existing `for_review` routing), and append provenance.
4. Print a pre-flight estimate (candidate count → number of LLM calls) before verifying, so the operator can bail on cost.

## Acceptance Criteria

- [ ] `creek compost scan` on a fixture vault writes compost notes to `10-Liminal/Compost/` for confirmed candidates and review entries for ambiguous ones
- [ ] `--no-llm` runs the gate only and never egresses content
- [ ] LLM verification refuses to run without explicit consent env, matching classify/draft behaviour
- [ ] `creek fill --with-compost`'s report reflects the scanned notes
- [ ] `./scripts/check-all.sh` passes

## Constraints

[...]

PR #977, opened 2026-07-29
fix(compost): make CompostTracker timezone-safe on both comparison sides

### LINKED ISSUE #938: [scan:bugs] CompostTracker default naive clock crashes project-silence detection against tz-aware fragment timestamps
## Role
suppressions).

## Goal
Make `CompostTracker` use a timezone-aware default clock so project-silence
detection stops raising `TypeError: can't compare offset-naive and offset-aware
datetimes` against the tz-aware fragment timestamps the pipeline produces,
verified by a new test that fails today and passes after the fix.

## Context
- File(s): `creek-tools/creek/generate/compost.py:210` (naive default clock),
  `:417-420` (the crashing comparison); reachable in production via
  `creek fill --with-compost` → `_fill_compost_report` →
  `CompostTracker().generate_compost_report(...)` (`creek/cli.py:1990,1994-1998`).
- Evidence:
  ```python
  # compost.py:210 — DEFAULT clock is timezone-NAIVE
  self._now = now or datetime.now(tz=UTC).replace(tzinfo=None)
  # compost.py:417-420 — compared against tz-AWARE fragment timestamps
  last_seen = max(effective_authored_at(frag) for frag in frags)
  if last_seen >= cutoff:        # cutoff derives from the naive self._now
      return None
  days = (self._now - last_seen).days
  ```
  `effective_authored_at` (`creek/time.py`) returns `Fragment.authored_at` or
  falls back to `Fragment.ingested`, which defaults to `now_la()` — a tz-AWARE
  America/Los_Angeles datetime that the module contract says is "always present,
  never naive". `CompostTracker()` is constructed with NO args in production
  (`creek/cli.py:1998`), so the naive path is the production path. As soon as one
  tag reaches `project_min_fragments` (default 3) with tz-aware timestamps, the
  `last_seen >= cutoff` (naive vs aware) comparison raises `TypeError` and the
  entire compost scan aborts. The suite passes only because `tests/test_compost.py`
  deliberately sidesteps it (its `reference_now` is a naive literal and `_fragment`
  builds naive `ingested`/`created`; the fixture docstring even warns that
  "silence calculations against a naive self._now would raise").
- Related: #882 (adds a dedicated compost scan command) is a distinct feature,
  not this crash; sibling scan findings this run touch generate/ and classify/.

## Output Format
"Closes #N".

## Examples
Before (naive default → TypeError against tz-aware fragments):
```python
self._now = now or datetime.now(tz=UTC).replace(tzinfo=None)
```
After (tz-aware default consistent with the pipeline's LA-normalised clock):
```python
self._now = now or now_la()   # from creek.time
```
Reproducing test sketch (`tests/test_compost.py`):
```python
def test_project_silence_with_default_clock_and_tzaware_fragments():
    frags = [_journal_fragment(tag="ship-it") for _ in range(3)]  # tz-aware ingested
    tracker = CompostTracker(project_gap_days=30, project_min_fragments=3)
    candidates = tracker.detect_compost_candidates([], frags)  # raises TypeError today
    assert any(c.tag == "ship-it" for c in candidates)
```

## Constraints
- Keep any explicitly-passed naive `now` working, or migrate callers/tests to
  tz-aware — do not merely paper over the comparison with a bare `except`

Geoffe-Ga:
## Summary

`CompostTracker`'s default clock was `datetime.now(tz=UTC).replace(tzinfo=None)` — the **only** place in `creek/` that *stripped* tzinfo, against a house convention of attaching it everywhere else (`pipeline.py:105`, `cli.py:337`, `clean/validator.py:329`, `ingest/base.py:326`, `ingest/gdrive.py:420,711`, `clean/hygiene.py:394`, `link/embeddings.py:653`). Two distinct defects followed.

**1. The `TypeError`.** Project-silence detection raised `TypeError: can't compare offset-naive and offset-aware datetimes` — at **two independent sites**, only one of which #938 identified:

- `compost.py:418` — `last_seen >= cutoff`
- `compost.py:417` — inside `max()` itself, while ranking a tag group whose fragments mix naive and aware timestamps, **before `self._now` is ever consulted**

**2. A UTC/LA calendar bug reachable in production today.** `self._now.date()` was the *UTC* date, feeding the compost-note filename (`:471`), the report's `generated:` frontmatter (`:570`), and the thread-dormancy `today` (`:267`). For the ~7 hours a day that UTC leads LA, `creek fill --with-compost` stamped an LA-normalised vault with **tomorrow's date**. Measured live: `datetime.now(tz=UTC).replace(tzinfo=None).date()` = `2026-07-29` while `now_la().date()` = `2026-07-28`.

### Why the issue's proposed one-liner is not the fix

#938 proposed `self._now = now or now_la()`. I ran it: that **relocates** the crash rather than fixing it. `Fragment` has no timezone validator, so `Fragment.model_validate` on frontmatter serialised without an offset yields a **naive** `ingested` — verified by execution. Naive and aware fragments are therefore both real production inputs, and an aware clock against naive fragments raises the identical `TypeError`. The `max()` site is unreachable by any clock-only change.

### The fix

New `creek.time.ensure_aware()`, applied to **both sides** of every comparison:

- **aware in → returned untouched** — instant, `tzinfo` object and microseconds all survive, so a Sydney-sourced `authored_at` keeps its Sydney rendering (the contract `effective_authored_at` already promises)
- **naive in → `LA_TZ` *attached***, wall clock intact. Attaching rather than converting is precisely what keeps values off day boundaries. LA because `creek/time.py` mandates LA normalisation (ontology §8.3), so a naive value here is an LA wall clock that lost its offset in transit.

It tests awareness via `value.utcoffset() is not None` rather than `tzinfo is not None` — CPython's own definition, which also covers a `tzinfo` whose `utcoffset()` returns `None`.

**Backward compatibility** (an explicit #938 constraint): an explicitly-passed **naive** `now=` still works and is read as an LA wall clock. ~10 existing tests rely on this. Day-count arithmetic is provably unchanged — CPython short-circuits comparison and subtraction when both operands share a `tzinfo` object, and `ZoneInfo` is interned. Verified: a 100-day span straddling a DST boundary yields exactly 100 days both naive and LA-aware.

### Which of the three explanations for `.replace(tzinfo=None)` was true

**Explanation 2, decisively** — `effective_authored_at` does *not* always return aware, so the strip was load-bearing against real inputs and deleting it alone moves the crash. Explanation 1 was true but *not* load-bearing: the tests do inject a naive `now`, and `tests/test_compost.py`'s `_fragment` docstring openly admits it mirrors naive `created` into `ingested` because otherwise "silence calculations against a naive `self._now` would raise" — but no *production* caller injects `now` at all, so the strip existed to keep the fixtures self-consistent, never for a production reason.

### Scope notes

- `detect_compost_candidates` has **no production caller today** (only `generate_compost_report` is wired, via `creek fill --with-compost`), so the `TypeError` is latent until #882 wires the detector — but it is public exported API, and defect 2 above *is* live on the wired path.
- The model-level root cause — `Fragment` accepting naive timestamps, which leaves the same latent crash in the temporal linker, eddies, threads, synchronicity and wavelength — is deliberately **out of scope** and filed as **#976**. `tests/test_time.py::test_result_is_always_tz_aware` currently asserts an invariant that does not hold; #976 covers making it non-vacuous.
- Evidence bearing on **#955** posted there: the two sites produce the *same instant* from the same `st_mtime` and differ only in rendering, so #955 is cosmetic **until** something calls `.date()` on the value — which is exactly defect 2's bug class. Recommendation recorded: prefer `LA_TZ`. Not fixed here.
- `creek/pipeline.py::_as_aware` (UTC semantics, different domain) deliberately left alone.

## Test plan

TDD: all tests written and proven RED before implementation.

**RED signatures observed** (against `36a4e4f`):
- `TypeError: can't compare offset-naive and offset-aware datetimes` at `compost.py:418` × 3 — default clock + aware fragments; aware clock + naive fragments; naive clock + aware fragments
- the same `TypeError` at `compost.py:417` × 1 — mixed naive/aware under one tag, inside `max()`
- `AssertionError: ['generated: 2026-07-29'] == ['generated: 2026-07-28']` — end-to-end through `creek fill --with-compost`
- `ImportError: cannot import name 'ensure_aware' from 'creek.time'`

**11 new tests**, all timezone-independent:
- `tests/test_time.py::TestEnsureAware` (5) — naive→LA attach, wall-clock fields preserved, Sydney and UTC passthrough unchanged, comparability with `now_la()` in both operand orders
- `tests/test_compost.py::TestTimezoneAwareClock` (5) — the four crash paths, each asserting the **full expected candidate** rather than "no exception" (so a fix that swallowed the `TypeError` would not pass); the mixed-tag test spans two real zones (LA + Sydney), not merely naive-vs-aware
- `tests/test_cli_fill.py::test_fill_with_compost_dates_the_report_on_the_la_calendar` (1) — runs the **real** `CompostTracker` through `creek fill --with-compost` with both `creek.generate.compost.datetime` and `creek.time.datetime` frozen to one instant whose UTC and LA dates disagree (`2026-07-29T06:00Z` = `2026-07-28T23:00-07:00`)

**Timezone independence:** every datetime literal carries an explicit `tzinfo`; fixed past instants (2020) are used where the verdict must not depend on run date; the one live-clock test brackets `today_la()` either side of the call so it cannot flake across LA midnight. No test reads the `TZ` env var. No new dependency (`freezegun`/`time-machine` deliberately not added).

**Mutation-checked:** I reverted the clock to its old buggy form and confirmed the frozen e2e test fails with the exact date mismatch — it is a genuine detector, not a tautology.

**Gate 2:** `cd creek-tools && ./scripts/check-all.sh` → **12 passed / 0 failed**, **6535 tests passed**, **93.99% branch coverage**, per-file gate clean (218 files ≥80%), mypy strict clean, ruff clean, complexity clean. No suppressions of any kind.

**Gate 2.5:** `code-review-orchestrator` verdict **CLEAN**, no blockers. Its one MINOR nit (a theoretical LA-midnight flake) is fixed in this branch.

**Test integrity:** `git diff --numstat origin/main` shows **0 deleted lines** across all three test files (`+75`, `+224`, `+87`). No existing test, assertion, fixture, or explicit `tzinfo`/`TZ` setting was modified or removed.

Closes #938
Refs #976, #955

### DIFF
--- a/creek-tools/creek/generate/compost.py
+++ b/creek-tools/creek/generate/compost.py
@@ -19,7 +19,7 @@ import logging
 import re
 from collections import defaultdict
 from dataclasses import dataclass, field
-from datetime import UTC, date, datetime, timedelta
+from datetime import date, datetime, timedelta
 from typing import TYPE_CHECKING
 
 import frontmatter
@@ -33,7 +33,7 @@ from creek.models import (
     Thread,
     ThreadStatus,
 )
-from creek.time import effective_authored_at
+from creek.time import effective_authored_at, ensure_aware, now_la
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Mapping
@@ -178,8 +178,14 @@ class CompostTracker:
             project_min_fragments: Minimum fragment count for a project
                 (tag) to be tracked. Defaults to 2.
             now: Reference "now" for age calculations. Defaults to
-                :func:`datetime.now` (timezone-naive). Useful for
-                deterministic tests.
+                :func:`creek.time.now_la` — tz-aware, America/Los_Angeles
+                — so the dates stamped on compost notes and the overview
+                report follow the same calendar as the rest of the vault
+                (issue #938; a UTC-derived clock ran a day ahead of LA
+                for the last several hours of every LA day). A naive
+                value passed here is still accepted and read as an LA
+                wall-clock reading, keeping existing callers working.
+                Useful for deterministic tests.
             similarity_fn: Closure mapping a text string to its maximum
                 cosine similarity against the compost exemplar set
                 (see :mod:`creek.generate.compost_embedding`). When
@@ -207,7 +213,7 @@ class CompostTracker:
         self.dormancy_days = dormancy_days
         self.project_gap_days = project_gap_days
         self.project_min_fragments = project_min_fragments
-        self._now = now or datetime.now(tz=UTC).replace(tzinfo=None)
+        self._now = ensure_aware(now) if now is not None else now_la()
         self._similarity_fn = similarity_fn
         self._verifier = verifier
         self._embedding_threshold = embedding_threshold
@@ -408,13 +414,27 @@ class CompostTracker:
         frags: list[Fragment],
         cutoff: datetime,
     ) -> CompostCandidate | None:
-        """Return a project candidate for *tag*, or ``None`` if still active."""
+        """Return a project candidate for *tag*, or ``None`` if still active.
+
+        Both sides of the silence comparison are timezone-aware before
+        they meet: ``cutoff`` derives from the aware clock, and each
+        fragment timestamp is passed through
+        :func:`creek.time.ensure_aware`. A vault mixing naive and aware
+        fragment timestamps — the ordinary result of round-tripping
+        frontmatter without an offset — therefore ranks and compares
+        instead of raising ``TypeError`` (issue #938; the underlying
+        model-level normalisation is issue #976).
+        """
         if len(frags) < self.project_min_fragments:
             return None
         # FEAT-031: measure silence against the authored-date precedence
         # so a project whose fragments were merely *ingested* recently
         # but authored long ago is still recognised as silent.
-        last_seen = max(effective_authored_at(frag) for frag in frags)
+        # ``max`` compares its own candidates, so the normalisation has to
+        # happen inside the generator rather than around it: without it a
+        # single naive timestamp in the group raises before ``cutoff`` is
+        # ever consulted.
+        last_seen = max(ensure_aware(effective_authored_at(frag)) for frag in frags)
         if last_seen >= cutoff:
             return None
         days = (self._now - last_seen).days
--- a/creek-tools/creek/time.py
+++ b/creek-tools/creek/time.py
@@ -5,7 +5,9 @@ timestamp the pipeline produces is normalised to America/Los_Angeles.
 Bare :func:`datetime.now` calls leak the host timezone (or, worse,
 produce naive datetimes that fail to compare against tz-aware ones with
 ``TypeError``), so production code routes through :func:`now_la` /
-:func:`today_la` instead.
+:func:`today_la` instead. :func:`ensure_aware` is the repair valve for
+values that arrive from outside that discipline — persisted frontmatter,
+legacy callers — making them safe to compare without moving the clock.
 
 The constant :data:`LA_TZ` is re-exported from
 :mod:`creek.ingest.base` to preserve the historical import path while
@@ -33,6 +35,47 @@ def now_la() -> datetime:
 def today_la() -> date:
     """Return today's date as observed in America/Los_Angeles."""
     return now_la().date()
+
+
+def ensure_aware(value: datetime) -> datetime:
+    """Return *value* guaranteed comparable against other aware datetimes.
+
+    Comparing a naive datetime with a tz-aware one raises ``TypeError:
+    can't compare offset-naive and offset-aware datetimes``, which is the
+    failure mode this module exists to prevent. Callers that mix a
+    pipeline-generated clock with timestamps read back off disk route
+    through this helper first.
+
+    The contract is deliberately asymmetric:
+
+    * **Aware in, untouched out.** The instant, the ``tzinfo`` object and
+      the microseconds all survive — no normalisation to LA or UTC.
+      A fragment authored in Sydney keeps its Sydney rendering, matching
+      the promise :func:`effective_authored_at` already makes to callers
+      that display source-local time.
+    * **Naive in, LA attached — never converted.** Every wall-clock field
+      is preserved; only the missing offset is filled in. LA is the
+      right assumption because the ontology (§8.3) normalises every
+      timestamp this pipeline writes to LA, so a naive value is an LA
+      wall-clock reading that lost its offset in transit — typically
+      through YAML frontmatter serialised without one. Converting rather
+      than attaching would shift such a value by the LA offset and could
+      move it across a day boundary.
+
+    Args:
+        value: A datetime that may be naive or timezone-aware.
+
+    Returns:
+        *value* unchanged when it is already aware, otherwise the same
+        wall clock anchored to :data:`LA_TZ`.
+    """
+    # ``datetime.utcoffset()`` returns ``None`` for exactly the values
+    # CPython treats as naive when comparing: ``tzinfo is None``, plus the
+    # rarer ``tzinfo`` whose own ``utcoffset`` yields ``None``. Both need
+    # the anchor, so test the offset rather than ``tzinfo`` itself.
+    if value.utcoffset() is not None:
+        return value
+    return value.replace(tzinfo=LA_TZ)
 
 
 def effective_authored_at(fragment: Fragment) -> datetime:
--- a/creek-tools/tests/test_cli_fill.py
+++ b/creek-tools/tests/test_cli_fill.py
@@ -9,17 +9,20 @@ compost step, and the summary reports real per-folder counts.
 from __future__ import annotations
 
 import logging
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 
 from typer.testing import CliRunner
 
 import creek.cli as cli_mod
+import creek.time as creek_time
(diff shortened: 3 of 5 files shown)


Geoffe-Ga:
## Summary
This PR fixes `CompostTracker`'s timezone handling: the default clock previously stripped tzinfo (`datetime.now(tz=UTC).replace(tzinfo=None)`), which both raised `TypeError` when compared against tz-aware fragment timestamps and silently stamped compost notes/reports with the UTC calendar date instead of the LA one (a real day-boundary bug for ~7 hours a day). The fix introduces `creek.time.ensure_aware()` and applies it to both sides of every naive/aware comparison in the project-silence detector.

## Strengths
- Correctly identifies that the issue's proposed one-liner (`self._now = now or now_la()`) only relocates the crash rather than fixing it, and demonstrates why via the `max()` call site at `compost.py:437` which runs *before* `self._now` is ever consulted.
- `ensure_aware()`'s asymmetric contract (attach LA to naive, pass aware through untouched) is the right call — converting instead of attaching would itself introduce day-boundary shifts, which is exactly the bug class being fixed.
- Uses `value.utcoffset() is not None` rather than `value.tzinfo is not None` to test awareness — the correct, CPython-consistent check.
- Backward compatibility for existing callers that pass a naive `now=` is preserved and explicitly tested.
- Verified the fix is a genuine detector (not a tautology) by reverting to the old buggy clock and confirming the new e2e test fails with the expected date mismatch.
- Good scoping discipline: explicitly calls out and defers the model-level root cause (`Fragment` accepting naive timestamps) to a separate issue (#976) rather than scope-creeping the fix.

## Security Concerns
None identified. 🟢 No user input, secrets, or external I/O touched — pure datetime normalization logic.

## Problems
None found blocking merge.

## Code Quality
- The docstrings on `ensure_aware`, `_project_candidate`, and the constructor's `now` parameter are unusually long, but each sentence is carrying real, non-obvious information (why LA vs UTC, why attach vs convert, why both comparison sides need normalizing) rather than restating the code — consistent with the CLAUDE.md docstring-coverage requirement and not a concern.
- Minor style nit (non-blocking): `_FrozenDatetime` in `test_cli_fill.py` duplicates a similar frozen-clock stub pattern already present in `test_time.py` per the PR description ("unlike the narrower stub in tests/test_time.py"). A shared test fixture could reduce duplication, but the PR's rationale for keeping them separate (different zone-agnostic behavior needed) is reasonable and not worth blocking on.

## CLAUDE.md Adherence
- ✅ TDD: PR body documents RED signatures observed before the fix (TypeError at two call sites, ImportError for `ensure_aware`, date-mismatch assertion failure).
- ✅ Coverage: self-reported 93.99% branch coverage, per-file gate clean.
- ✅ Linting/typing: self-reported ruff and mypy strict clean.
- ✅ Commits: on a feature branch (`issue/938-scan-bugs-composttracker-default-naive`), not direct-to-main.
- ✅ No suppressions (`noqa`, `type: ignore`, etc.) introduced, matching the `max-quality-no-shortcuts` philosophy.
- Note: CI was still in progress (`IN_PROGRESS`) for the 3.11/3.12/3.13 Code Quality & Testing jobs at review time; the self-reported local gate results are consistent with what's expected to pass, but the reviewer should confirm the CI run finishes green before merge.

## Testing
Thorough and well-targeted:
- `tests/test_time.py::TestEnsureAware` (5 tests) covers naive→LA attachment, wall-clock field preservation, aware passthrough for both Sydney and UTC zones, and bidirectional comparability with `now_la()`.
- `tests/test_compost.py::TestTimezoneAwareClock` (5 tests) exercises all three previously-crashing combinations (aware clock/naive fragment, naive clock/aware fragment, default clock/aware fragment, and the `max()` site with mixed naive+aware+Sydney fragments under one tag) — each asserting the full expected candidate rather than merely "no exception," which is what makes them resistant to a fix that swallows the error instead of resolving it.
- `tests/test_cli_fill.py::test_fill_with_compost_dates_the_report_on_the_la_calendar` freezes both `creek.generate.compost.datetime` and `creek.time.datetime` to a single instant chosen so UTC and LA dates disagree, and asserts the report is dated on the LA calendar end-to-end through the real CLI path — a strong regression guard for the day-boundary bug specifically.
- Timezone independence of the suite (explicit tzinfo on every literal, no reliance on host `TZ`) is well reasoned and verified in the code.

## Documentation
Docstrings for `ensure_aware`, the constructor's `now` parameter, and `_project_candidate` are updated to explain the asymmetric attach-vs-convert contract and why both comparison operands need normalizing — accurate and consistent with the implementation as reviewed.



## Verdict: LGTM

The fix correctly resolves both the reported TypeError and the previously-unreported UTC/LA date-boundary bug, applies ensure_aware() at every site where a naive/aware mismatch could occur (constructor, and both operands of the max()/>= comparison in _project_candidate), preserves backward compatibility for naive now= callers, and ships with thorough, non-tautological tests (verified by reading the diff and the surrounding compost.py/time.py code) plus a clean self-reported quality gate.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5114512665 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #982, opened 2026-07-29
chore(crawdad): consolidate the three near-duplicate helper pairs in test_dependency_pins.py


Geoffe-Ga:
## Context

`crawdad/tests/test_dependency_pins.py` now guards three security pins — `mcp` (#862), `pyasn1` (#867), and `aiohttp` (#978) — and each one carries its own near-identical helper pair:

- `_mcp_specifier()` / `_locked_mcp_version()`
- `_pyasn1_constraint_specifier()` / `_locked_pyasn1_version()`
- `_aiohttp_constraint_specifier()` / `_locked_aiohttp_version()`

The lock-reading helpers are byte-identical apart from the package name; the specifier helpers differ only in which table they read (`[project].dependencies` for the direct `mcp` pin vs `[tool.uv].constraint-dependencies` for the two transitive ones).

## Why it was left alone in #978

Deliberate call by the architect and confirmed at Gate 2.5 review on PR #981: consolidating would have required editing every existing test body during a **P0 security** fix, and this repo has a documented failure mode of specialists silently mangling live assertions during "refactors". Keeping #978 purely additive (`105 insertions, 0 deletions`) made the no-assertion-lost property trivially verifiable in the diff. The triplication was accepted as the lower-risk trade.

Three copies is where the pattern stops being a coincidence, so it is worth doing now — outside a security PR, where a careful diff review is affordable.

## Proposal

Extract two generic helpers:

- `_locked_version(name: str) -> Version`
- `_constraint_specifier(name: str) -> SpecifierSet` and `_declared_specifier(name: str) -> SpecifierSet` (or one helper with a table selector)

Keep every existing test function, assertion, and assertion message intact — this is a plumbing change only. The assertion messages are deliberately specific (they name advisory IDs and explain the floor) and must not be genericised away.

`creek-tools/tests/test_dependency_pins.py` has the same shape and should get the same treatment in the same PR if it has drifted the same way.

## Acceptance
- No test function removed, renamed, or weakened; `git diff --numstat` deletions accounted for line by line.
- `cd crawdad && ./scripts/check-all.sh` exits 0.
- Deliberately break each pin locally (floor and lock, per package) and confirm the corresponding test still fails with a useful message.

Refs #978, PR #981

PR #967, opened 2026-07-29
fix(quality): bring creek_mcp under mypy, bandit and pylint (#925)

Geoffe-Ga:
## Summary

`creek_mcp/` — the network-facing package (MCP server, bearer auth, remote auth, token policy, path confinement, audit log, 20 tool modules) — was excluded from **mypy, bandit and pylint**. Not by any config exclusion: every invocation simply passed `creek/` as its only CLI target. This PR widens all three to `creek/ creek_mcp/` at every surface and pins that with a gate-contract test.

This is not hygiene. PR #963 ran a standalone `mypy creek_mcp/` and went 7 errors → 4; one of the three it eliminated was `server.py:185: has no attribute "_default_llm"; maybe "default_llm"?` — mypy would have caught #929, a production path that had **never worked**, at authoring time for free.

### How each scanner was scoped, before → after

| Scanner | Before | After |
|---|---|---|
| mypy | `scripts/typecheck.sh` → `python -m mypy creek/` | `creek/ creek_mcp/` |
| bandit (local) | `scripts/security.sh` → `bandit -r creek/ -ll` | `creek/ creek_mcp/`, `-ll` kept |
| bandit (CI) | `ci.yml` **inlines** `bandit -r creek/ …` — does *not* call the script | both lines widened |
| pylint (gate) | `scripts/pylint.sh` → `python -m pylint creek/` ×2 (gating + JSON) | both widened |
| pylint (extended) | `scripts/lint-extended.sh` → `pylint creek/` | widened |
| pylint (pre-commit) | `pylint-fast` hook `files: ^creek-tools/creek/` | `^creek-tools/(creek\|creek_mcp)/` |

There was **no** `[tool.bandit] exclude_dirs` entry, mypy `exclude`, or per-module override for `creek_mcp` — the target argument was the whole mechanism.

**CI parity was the trap.** mypy and pylint reach CI through `./scripts/typecheck.sh` and `./scripts/pylint.sh`, so the script edits propagate. The bandit step **inlines** its own `bandit -r` and never calls `scripts/security.sh`, so widening the script alone would have been local-only. Both `ci.yml` bandit lines were edited, and both files now carry a comment saying the two target lists are hand-synced. No other workflow gates on these three scanners.

The pre-commit `mypy` and `bandit` hooks are staged-file-typed with no `files:` filter, so they already covered `creek_mcp` — only `pylint-fast` needed widening.

### Findings the widening surfaced, and how each was routed

[...]

PR #985, opened 2026-07-29
fix(security): run pip-audit in crawdad's gate on both dependency surfaces

### LINKED ISSUE #979: [scan:security] crawdad security gate runs bandit but never pip-audit — 19 advisories accumulated undetected
## Role
its existing conventions (TDD via stay-green, the `cd crawdad &&
./scripts/check-all.sh` gate, ≥90% branch coverage, ≥95% docstring coverage,
complexity ≤10, mypy strict, zero lint/type suppressions). You are closing a
gap in the security gate itself, so treat `creek-tools/scripts/security.sh` as
the reference implementation to bring CrawDad up to parity with.

## Goal
Add a `pip-audit` dependency-vulnerability step to `crawdad/scripts/security.sh`
so `cd crawdad && ./scripts/check-all.sh` fails on any advisory in
`crawdad/uv.lock`, verified by the script exiting non-zero on the currently
locked (vulnerable) tree and exiting zero once #978 lands.

## Context
- File(s): `crawdad/scripts/security.sh:1-7` (entire script),
  `crawdad/scripts/check-all.sh:32`, `.github/workflows/ci.yml:216-255`
  (the `crawdad` / "CrawDad Quality" job)
- Scanned at commit: `36a4e4f219372dd91caad1f9b281ec6f9e9092fa` — re-verify against HEAD before starting
- Evidence: `crawdad/scripts/security.sh` runs Bandit and **nothing else** —
  there is no dependency audit anywhere in the CrawDad gate:

  ```bash
  #!/usr/bin/env bash
  # scripts/security.sh — Bandit on the source tree.
  set -euo pipefail
  SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
  cd "$(dirname "$SCRIPT_DIR")"
  bandit -r crawdad -ll
  ```

  `crawdad/scripts/check-all.sh:32` invokes it as
  `run "Security" "$SCRIPT_DIR/security.sh"`, and `.github/workflows/ci.yml`
  runs only `./scripts/check-all.sh` for the crawdad job ("Single source of
  truth: identical script developers run locally"). So no dependency audit
  runs for CrawDad locally *or* in CI.
- **Impact, corroborated by a second independent signal:** running the audit
  by hand at this SHA surfaces **19 known vulnerabilities in 5 packages** that
  the gate has never reported — `aiohttp` (11, see #978), `cryptography`,
  `pydantic-settings`, `python-multipart`, and `starlette`. Every one has a
  published fix. `creek-tools/uv.lock` already carries the fixed versions of
  four of them, so CrawDad has silently drifted behind its sibling package
  precisely because it has no gate to hold the line.
- Secondary drift to fix in the same PR: the CI job installs with
  `pip install -e ".[dev]"` (unpinned) rather than from the committed
  `uv.lock`, so CI does not even exercise the locked tree the audit is meant
  to protect. `creek-tools` installs from its lock; CrawDad should too.
- The reference implementation is `creek-tools/scripts/security.sh`, which
  runs `bandit -r creek/ creek_mcp/ -ll` **and** `pip-audit`, with each
  ignored advisory carrying an inline justification, an advisory link, and a
  re-audit date.
- Related: #978 (the aiohttp advisories this gate would have caught).

## Output Format
`cd crawdad && ./scripts/check-all.sh`, and (4) references this issue with
"Closes #N".

## Examples
```bash
# BEFORE — crawdad/scripts/security.sh
set -euo pipefail
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
cd "$(dirname "$SCRIPT_DIR")"
bandit -r crawdad -ll

# AFTER
set -euo pipefail
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
cd "$(dirname "$SCRIPT_DIR")"

echo "=== Security Checks (Bandit) ==="
bandit -r crawdad -ll || { echo "✗ Bandit found issues" >&2; exit 1; }

echo "=== Dependency Vulnerability Check (pip-audit) ==="
# Audit the committed lock, not whatever happens to be installed, so local
# and CI agree on the same dependency set. No --ignore-vuln entries today;
# any future entry needs a justification, an advisory link, and a re-audit
# date, matching creek-tools/scripts/security.sh.
uv export --frozen --all-extras --no-hashes -o "$(mktemp -t crawdad-reqs-XXXXXX.txt)" ...
pip-audit -r "$REQS" || { echo "✗ pip-audit found vulnerable dependencies" >&2; exit 1; }
```

Note the deliberate choice to audit the **exported lock** rather than the
ambient environment: `pip-audit` with no `-r` audits site-packages, which in
CI is an unpinned `pip install -e ".[dev]"` tree and therefore not the thing
the lockfile promises. Auditing the export is what makes the gate meaningful.

For the test, follow the existing `crawdad/tests/` conventions for
script-shape assertions — assert that `scripts/security.sh` actually invokes
`pip-audit` and that `check-all.sh` still wires `security.sh` in, so the gate
cannot be silently removed again.

## Constraints
- Do not land this with `|| true` or any escape hatch on the `pip-audit`
  step — a non-blocking security gate is the bug being fixed here
- Do not add `--ignore-vuln` entries to make the gate pass on today's
  vulnerable lock. The correct sequence is to land #978 (which clears all 19
  advisories) and let this gate go green honestly; if the two PRs race, this
  one may land second
- Keep `shellcheck` clean — it runs in pre-commit

Geoffe-Ga:
## Summary

`crawdad/scripts/security.sh` ran exactly one line — `bandit -r crawdad -ll`. There was no `pip-audit` anywhere in crawdad's gate, while the sibling `creek-tools/scripts/security.sh` has run both all along. Eight known vulnerabilities accumulated in `crawdad/uv.lock` behind a green CI.

This follows the #925/#967 template: turn the scanner on **and** land the remediation in one change, so the gate ends green with every finding **fixed** rather than suppressed. **Zero suppressions** — no `--ignore-vuln`, no `--strict`, no `|| true`, no skip branches.

### Scope correction — the advisories were lockfile-only

Verified, not assumed. The CrawDad Quality job provisions with `pip install -e ".[dev]"`, and pip honours neither `uv.lock` nor `[tool.uv]`. An audit of the actual pip-provisioned venv returned `No known vulnerabilities found` **both before and after** this change. All eight lived in the lockfile.

They reach `uv sync` users, and the lock is the repo's reproducibility contract and its audit surface — so this is real and worth fixing — but it is **not** vulnerable code on the live bot's network path. Worth re-checking the `P0` label against that.

### What pip-audit is pointed at, and why: **both** surfaces

| Surface | Invocation | Question it answers |
|---|---|---|
| Installed environment | `pip-audit` | "Is what CI actually imports vulnerable?" |
| Exported lock | `uv export --quiet --locked ... -o <mktemp>` → `pip-audit --requirement <tmp> --disable-pip` | "Is the reproducibility contract `uv sync` users install vulnerable?" |

The gap between those two answers is exactly what let this drift, so auditing only one would have been the same mistake in a new place:

- **Environment-only** would have found *nothing* — it reports clean today while the lock carried eight.
- **Lock-only** leaves the CI install path permanently unguarded, because `[tool.uv].constraint-dependencies` **cannot protect a pip install**. The four floors this PR adds protect `uv sync` users only; the environment audit is the sole guard on the pip path. That is the same drift class #983 independently found in the dev toolchain (CI resolves ruff 0.16.0 / mypy 2.3.0 vs lock 0.15.14 / 2.1.0).

`creek-tools` gets away with a single bare `pip-audit` only because its CI installs **from** the lock (`uv pip install -r locked-requirements.txt`), which makes environment and lock the same artifact. crawdad's CI does not, so crawdad needs both.

### Findings table — 8 → 0, all fixed, none suppressed

Every version checked against the **live PyPI/OSV registry**, not recalled. All four are transitive-only with open upstream floors, so floors live in `[tool.uv].constraint-dependencies` per the DEP-003 precedent; `mcp>=1.28.1,<2.0.0` is untouched.

| Package | Locked | Advisory | Floor added | Live-registry check | Resolution |
|---|---|---|---|---|---|
| `cryptography` | 48.0.0 | GHSA-537c-gmf6-5ccf | `>=48.0.1` | latest 49.0.0; 48.0.1 first patched | **fixed** → 49.0.0 |
| `pydantic-settings` | 2.14.1 | GHSA-4xgf-cpjx-pc3j | `>=2.14.2` | latest 2.14.2 = the fix | **fixed** → 2.14.2 |
| `python-multipart` | 0.0.29 | PYSEC-2026-3036, -3037 (fixed 0.0.30); PYSEC-2026-3040 (fixed 0.0.31) | `>=0.0.31` | latest 0.0.32; **0.0.30 is a partial fix** | **fixed** → 0.0.32 |
| `starlette` | 1.1.0 | PYSEC-2026-248 (fixed 1.3.0); PYSEC-2026-249 (fixed 1.3.1) | `>=1.3.1` | latest 1.3.1; **1.3.0 is a partial fix** | **fixed** → 1.3.1 |

`python-multipart` and `starlette` follow the first-fully-patched convention #978 set for `aiohttp` (3.14.0 was partial): flooring at 0.0.30 / 1.3.0 would still ship a known advisory.

No ceiling blocks any raise — `mcp` 1.28.1 declares `pydantic-settings>=2.5.2`, `python-multipart>=0.0.9`, `starlette>=0.27`; `sse-starlette` 3.4.4 declares `starlette>=0.49.1`; `google-auth`/`pyjwt` declare `cryptography>=38.0.3`/`>=3.4.0`. The relock changed only those four versions — nothing removed, nothing downgraded.

### pip-audit output, verbatim

Before (exported lock):
```
Found 8 known vulnerabilities in 4 packages
Name              Version ID                  Fix Versions
----------------- ------- ------------------- ------------
cryptography      48.0.0  GHSA-537c-gmf6-5ccf 48.0.1
pydantic-settings 2.14.1  GHSA-4xgf-cpjx-pc3j 2.14.2
python-multipart  0.0.29  PYSEC-2026-3037     0.0.30
python-multipart  0.0.29  PYSEC-2026-3036     0.0.30
python-multipart  0.0.29  PYSEC-2026-3040     0.0.31
starlette         1.1.0   PYSEC-2026-248      1.3.0
starlette         1.1.0   PYSEC-2026-249      1.3.1
starlette         1.1.0   PYSEC-2026-248      1.3.0
```

After (`./scripts/security.sh`, both surfaces):
```
=== bandit: source tree ===
	No issues identified.
=== pip-audit: installed environment ===
No known vulnerabilities found
=== pip-audit: exported uv.lock ===
No known vulnerabilities found
```

### How I verified the change actually reaches CI

The brief warned that a script edit may not propagate — in this repo the creek-tools **bandit** step is inlined in `ci.yml` rather than calling `scripts/security.sh`. **That trap does not apply to crawdad**, and I verified rather than assumed it. The full chain:

1. `.github/workflows/ci.yml` "CrawDad Quality" job → `pip install -e ".[dev]"` then `run: ./scripts/check-all.sh` (it has no inlined scanner steps).
2. `crawdad/scripts/check-all.sh:32` → `run "Security" "$SCRIPT_DIR/security.sh"`.
3. `crawdad/scripts/security.sh:38,58,59` → the two `pip-audit` invocations.
4. Both tools come from the dev extra — so I added `pip-audit>=2.10.0` and `uv>=0.11.16` there. A gate tool the extras omit is a gate that cannot run.

Confirmed in a **from-scratch venv**: `pip install -e ".[dev]"` provisions pip-audit 2.10.1 and uv 0.12.0, and the bare audit runs clean. **No `ci.yml` change is needed.**

### Implementation notes

- **Hashed export** (no `--no-hashes`): lets `--disable-pip` work without `--no-deps` and avoids a per-run warning.
- **`--quiet`** suppresses only the redundant ~145 KB copy of the export that `-o` *also* dumps to stdout, which would bury the audit results in the CI log. Verified it does **not** silence failures: `--locked` still exits 2 with `The lockfile at uv.lock needs to be updated` on stderr.
- **`--locked` doubles as a lock-freshness gate** — editing `pyproject.toml` without `uv lock` now fails the Security step loudly. Intentional; the fix is `uv lock`, never dropping the flag.
- **No `--strict`**, deliberately: the unpublished local `crawdad` package is a permanent benign pip-audit SKIP, and `--strict` would make it a permanent false failure. Commented in-script so nobody "hardens" it later.
- Operational note: the environment arm means a new advisory in `pip`/`setuptools` can redden CI before a fix ships. The job already runs `pip install --upgrade pip setuptools wheel` first, and creek-tools has carried this same posture all along.

## Test plan

- 15 new tests in `crawdad/tests/test_dependency_pins.py`, all confirmed **RED before / GREEN after**:
  - 12 pin guards (pyproject floor + locked version per package), including explicit partial-fix guards rejecting `0.0.30` and `1.3.0`.
  - **3 gate-wiring guards** pinning the exact regression class that caused #979 — a scanner the gate declares but never runs: `security.sh` must audit the environment, must audit the exported lock, and the dev extra must provision both tools.
- Positive control that the environment arm is not a no-op: injected `pyasn1==0.6.3` into a scratch venv → bare `pip-audit` flagged 3 advisories and exited 1.
- **Gate 2 (crawdad) green:** `cd crawdad && ./scripts/check-all.sh` → exit 0, 7/7 checks, **554 tests** (was 539), 94.59% branch coverage. `shellcheck` clean on `security.sh`. The creek-tools gate was not run because the diff touches no `creek-tools/` file.

### Test integrity

One test file touched. `git diff --numstat origin/main` → `571 insertions / 8 deletions`. All 8 deletions are **docstring prose**, not code: the module and two test docstrings claimed "`uv.lock` is what CI installs and what pip-audit inspects" — false for crawdad, and precisely the misconception that let the lock drift unaudited. Corrected as a **deliberate accuracy fix**, not a cleanup.

An AST comparison against `origin/main` (docstrings stripped) confirms: **zero pre-existing definitions removed, zero pre-existing code changed**; 28 new definitions added (13 helpers + 15 tests). SHA-256 `bd3907b7…` → `818ae14d…`.

## Known gaps / follow-ups

- The 3 gate-wiring guards parse `security.sh` as text, so they would not catch a future refactor to `python -m pip_audit` or a `$VAR`-indirected invocation. Real but theoretical for a script this size; flagged rather than over-engineered.
- This PR enlarges **#982**'s helper-duplication surface by four helper pairs (deliberate — consolidating is that issue's job).
- Dev-tool version ceilings stay out of scope (**#983**). The two tools added here are unbounded in the same style as crawdad's existing dev extras and should be swept in with that issue.
- `.github/dependabot.yml` covers `/creek-tools` but **not** `/crawdad`, so crawdad's deps get no automated bump PRs. Out of scope here; worth its own issue.

Closes #979
Refs #925, #967, #978, #982, #983

### DIFF
--- a/crawdad/CLAUDE.md
+++ b/crawdad/CLAUDE.md
@@ -22,6 +22,9 @@ These mirror `creek-tools/CLAUDE.md` at a smaller scale:
    - MyPy strict mode, zero violations
    - Ruff lint + format, zero violations
    - Bandit, zero medium-or-above findings
+   - pip-audit, zero known vulnerabilities across both the installed
+     environment and the exported `uv.lock` — no suppression without
+     a tracked issue
 
 ## 2. Project overview
 
@@ -145,3 +148,22 @@ The bot does **not** exit on MCP subprocess failure. The pattern is:
 | Type checking | strict, zero | `mypy --strict` |
 | Lint + format | zero | `ruff check` + `ruff format` |
 | Security | zero medium+ | `bandit -r crawdad/ -ll` |
+| Dependency vulnerabilities | zero known | `pip-audit` (installed env + exported `uv.lock` — `./scripts/security.sh`) |
+
+`scripts/security.sh` runs `pip-audit` twice because crawdad has two
+distinct dependency surfaces and each answers a different question.
+CI provisions with `pip install -e ".[dev]"`, and pip honours neither
+`uv.lock` nor `[tool.uv].constraint-dependencies` — both are invisible
+to it — so a bare `pip-audit` audits only what CI actually imports.
+The second run exports `uv.lock` (`uv export --quiet --locked
+--all-extras --no-emit-project`) and audits that instead, since
+`uv.lock` is the reproducibility contract `uv sync` users install and
+is where all eight advisories of #979 lived; an environment-only audit
+would have reported clean while the lock carried eight. `--locked`
+doubles as a lock-freshness gate — if `pyproject.toml` and `uv.lock`
+have drifted, the export fails with "The lockfile at `uv.lock` needs
+to be updated" and the fix is `uv lock`, never dropping `--locked`.
+Do not add `--strict` to either `pip-audit` call: the local `crawdad`
+package isn't published to PyPI, so pip-audit always reports it as a
+benign SKIP, and `--strict` would turn that permanent skip into a
+permanent false failure.
--- a/crawdad/pyproject.toml
+++ b/crawdad/pyproject.toml
@@ -48,6 +48,14 @@ dev = [
     "radon>=6.0.0",
     "xenon>=0.9.0",
     "types-PyYAML>=6.0.0",
+    # scripts/security.sh — the security gate — shells out to both of
+    # these: pip-audit scans the installed environment and the exported
+    # lock, and uv performs the `uv export --locked` that produces that
+    # export. CI installs only this extra before running check-all.sh,
+    # so the gate's own toolchain has to be declared here or the gate
+    # cannot run (#979).
+    "pip-audit>=2.10.0",
+    "uv>=0.11.16",
 ]
 
 [build-system]
@@ -72,6 +80,34 @@ constraint-dependencies = [
     # transitive via google-genai → google-auth → pyasn1-modules.
     # Both fixed in 0.6.4 (#867).
     "pyasn1>=0.6.4",
+    # Security: pin cryptography above GHSA-537c-gmf6-5ccf, carried by
+    # the previously-locked 48.0.0 — transitive via google-genai →
+    # google-auth → cryptography and via mcp → pyjwt → cryptography.
+    # Fixed in 48.0.1. Both consumers declare open floors
+    # (google-auth: cryptography>=38.0.3, pyjwt: cryptography>=3.4.0),
+    # so nothing upstream blocks the raise (#979).
+    "cryptography>=48.0.1",
+    # Security: pin pydantic-settings above GHSA-4xgf-cpjx-pc3j,
+    # carried by the previously-locked 2.14.1 — transitive via mcp →
+    # pydantic-settings. Fixed in 2.14.2; mcp 1.28.1 declares
+    # pydantic-settings>=2.5.2, so the pinned mcp is unaffected (#979).
+    "pydantic-settings>=2.14.2",
+    # Security: pin python-multipart above PYSEC-2026-3036,
+    # PYSEC-2026-3037, and PYSEC-2026-3040, all carried by the
+    # previously-locked 0.0.29 — transitive via mcp → python-multipart.
+    # 0.0.30 fixes only the first two, so the floor is 0.0.31, the first
+    # fully patched release (same first-fully-patched convention as
+    # aiohttp>=3.14.1 above). mcp 1.28.1 declares
+    # python-multipart>=0.0.9 (#979).
+    "python-multipart>=0.0.31",
+    # Security: pin starlette above PYSEC-2026-248 and PYSEC-2026-249,
+    # both carried by the previously-locked 1.1.0 — transitive via
+    # mcp → starlette and mcp → sse-starlette → starlette. 1.3.0 fixes
+    # only PYSEC-2026-248, so the floor is 1.3.1, the first fully
+    # patched release. mcp 1.28.1 declares starlette>=0.27 (>=0.48.0 on
+    # Python ≥3.14) and sse-starlette 3.4.4 declares starlette>=0.49.1,
+    # so both consumers accept the raise (#979).
+    "starlette>=1.3.1",
 ]
 
 [tool.setuptools.packages.find]
--- a/crawdad/scripts/security.sh
+++ b/crawdad/scripts/security.sh
@@ -1,8 +1,59 @@
 #!/usr/bin/env bash
-# scripts/security.sh — Bandit on the source tree.
+# scripts/security.sh — Bandit on the source tree, then pip-audit on
+# both dependency surfaces (installed environment and exported uv.lock).
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$(dirname "$SCRIPT_DIR")"
 
+echo "=== bandit: source tree ==="
 bandit -r crawdad -ll
+
+# pip-audit runs twice because crawdad has two distinct dependency
+# surfaces, and each one answers a different question.
+#
+# 1. The installed environment — "is what CI actually imports
+#    vulnerable?" CI provisions with `pip install -e ".[dev]"`, and pip
+#    honours neither uv.lock nor [tool.uv].constraint-dependencies:
+#    both are invisible to it. Nothing else in the gate guards this
+#    surface, so a floor can be correct in pyproject and in the lock
+#    while CI still resolves a vulnerable transitive release.
+# 2. The exported lock — "is the reproducibility contract `uv sync`
+#    users install vulnerable?" All eight advisories of #979 lived
+#    here: an environment-only audit reported "No known vulnerabilities
+#    found" while uv.lock carried eight, so auditing the environment
+#    alone would have missed this entire issue.
+#
+# The gap between those two answers is what let the drift go unnoticed.
+# The sibling creek-tools gate gets away with a single bare pip-audit
+# only because its CI installs FROM the lock (`uv pip install -r
+# locked-requirements.txt`), which makes environment and lock the same
+# artifact. crawdad's CI does not, so crawdad needs both.
+
+echo "=== pip-audit: installed environment ==="
+# No --strict here: the local `crawdad` package is not published to
+# PyPI, so pip-audit always reports it as a benign SKIP. --strict would
+# promote that permanent skip into a permanent false failure — do not
+# "harden" this by adding it.
+pip-audit
+
+echo "=== pip-audit: exported uv.lock ==="
+LOCK_REQUIREMENTS="$(mktemp)"
+trap 'rm -f "$LOCK_REQUIREMENTS"' EXIT
+# --locked doubles as a lock-freshness gate: the export fails instead
+# of silently relocking when pyproject.toml and uv.lock have drifted.
+# That failure is intentional — the fix is to run `uv lock`, never to
+# drop --locked.
+#
+# The export deliberately keeps hashes (no --no-hashes): the hashed
+# form lets `pip-audit --disable-pip` resolve the requirements without
+# --no-deps and without emitting the "users are encouraged to fully
+# hash" warning on every run.
+#
+# --quiet suppresses only the redundant copy of the export that `-o`
+# otherwise also dumps to stdout (~145 KB of hashes per run, which
+# would bury the audit results in the CI log). It does NOT silence
+# failures: `--locked` still exits non-zero with "The lockfile at
+# `uv.lock` needs to be updated" on stderr when the lock has drifted.
+uv export --quiet --locked --all-extras --no-emit-project -o "$LOCK_REQUIREMENTS"
+pip-audit --requirement "$LOCK_REQUIREMENTS" --disable-pip
--- a/crawdad/tests/test_dependency_pins.py
+++ b/crawdad/tests/test_dependency_pins.py
@@ -38,13 +38,45 @@ discord.py 2.7.1 declares ``aiohttp<4,>=3.7.4``, so nothing about
 discord.py changes, and the floor lives in
 ``[tool.uv].constraint-dependencies`` alongside pyasn1.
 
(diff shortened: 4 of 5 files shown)


Geoffe-Ga:
## Summary
Adds `pip-audit` to crawdad's security gate on both the installed environment and the exported `uv.lock`, and fixes the 8 advisories (across `cryptography`, `pydantic-settings`, `python-multipart`, `starlette`) that had accumulated in the lock unnoticed. Backed by 15 new tests, dev-extra additions (`pip-audit`, `uv`), and matching `CLAUDE.md`/docstring documentation.

## Strengths
- Root-caused *why* the gap existed (pip installs neither honor `uv.lock` nor `[tool.uv].constraint-dependencies`) instead of just bolting on a scanner, and audits both surfaces the gap implies.
- Correctly avoids the "partial fix" trap for `python-multipart` (0.0.30) and `starlette` (1.3.0), flooring at the first fully-patched release for each (verified: `uv.lock` resolves `python-multipart` to 0.0.32 and `starlette` to 1.3.1, both ≥ the stated floors).
- Zero suppressions (`grep`-checked `security.sh`: no `--ignore-vuln`, no `--strict`, no `|| true`).
- Traced the actual CI path (`ci.yml` → `check-all.sh` → `security.sh`) end-to-end rather than assuming the script edit propagates, and confirmed with a from-scratch venv that `.[dev]` provisions the new tools.
- Test coverage of the *wiring* itself (`test_security_script_audits_installed_environment`, `..._exported_lock`, `test_dev_extras_provision_audit_toolchain`) pins the exact regression class (#979: a scanner declared but never run), not just the pins.
- I hand-traced `_pip_audit_invocations`/`_is_pip_audit_command`/`_targets_requirements_file` against the real `security.sh` content and confirmed they classify the bare `pip-audit` vs. the `--requirement ... --disable-pip` invocation correctly, including not miscounting the `echo "=== pip-audit: ..."` lines.

## Security Concerns
🟢 None found. `mktemp`/`trap` cleanup is correct and race-free for this use (non-sensitive dependency-version data); `--locked` on `uv export` correctly turns lock drift into a hard CI failure rather than a silent relock. I could not independently re-verify the underlying CVE/GHSA/PYSEC advisory data against the live registries (no web access in this session) — the PR body documents that verification was done against live PyPI/OSV, and the internal consistency (constraint floor ↔ resolved lock version ↔ test assertions) all checks out, so I'm not blocking on this, just flagging it as unverified by me directly.

## Problems
None blocking.

## Code Quality
- The four new `_<pkg>_constraint_specifier()` / `_locked_<pkg>_version()` helper pairs are near-verbatim duplicates of each other (and of the pre-existing `aiohttp`/`pyasn1` ones). The author already flags this in the PR description as deliberately deferred to #982 (helper consolidation), so it's a known, tracked, non-blocking item rather than a fresh finding.
- Comment/docstring prose is repeated near-verbatim across `security.sh`, `CLAUDE.md`, and the test module docstring. Matches this repo's established documentation-heavy convention for security PRs (per #925/#967/#978 precedent) and satisfies the ≥95% interrogate gate, so not treating as a defect — just denser than a typical PR.

## CLAUDE.md Adherence
✅ Uses project scripts only (`scripts/security.sh`), no direct tool invocation changes.
✅ No shortcut suppressions added.
✅ `crawdad/CLAUDE.md` quality-bar table updated to reflect the new pip-audit gate.
✅ Conventional Commits format (`fix(security): ...`).
✅ TDD noted as followed (tests confirmed RED before / GREEN after per PR body).

## Testing
15 new tests, each package gets a floor-rejects / floor-accepts / locked-version guard triple plus 3 gate-wiring guards. Positive control (injecting a known-vulnerable `pyasn1==0.6.3` into a scratch venv to confirm the environment arm isn't a no-op) is a nice touch beyond the minimum bar. Coverage claims (94.59% branch, 554 tests) are asserted in the PR body but not independently re-run in this review (no `uv`/`pip-audit` available in this review sandbox); nothing in the diff suggests they wouldn't hold.

## Documentation
Docs updated in all the right places: `crawdad/CLAUDE.md` quality table + gate-rationale paragraph, `security.sh` inline comments, and the test module docstring — all consistent with each other and with the actual script behavior I read.



## Verdict: LGTM

The dependency floors, constraint block, and resolved uv.lock versions are internally consistent (verified cryptography→49.0.0, pydantic-settings→2.14.2, python-multipart→0.0.32, starlette→1.3.1 all resolve above the stated patched versions), the security.sh changes are correctly wired and shell-safe, and the new tests genuinely exercise both the pins and the gate wiring (traced the tokenizer logic against the actual script — it correctly classifies the bare vs. `-r` pip-audit invocations). No suppressions, no bypasses, nothing blocking; the one real nit (four near-duplicate helper-function pairs) is already self-acknowledged and tracked under #982, so it isn't a fresh ask.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5115452596 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #509, opened 2026-06-02
harden(author): _TIER_RANK fail-closed on future PrivacyTier extension


Geoffe-Ga:
Non-blocking note from the #473 review (PR #505, LGTM round 7). `creek/author/checks.py::_TIER_RANK` maps each `PrivacyTier` to a restrictiveness rank; the privacy check indexes it as `_TIER_RANK[tier]`. If a new `PrivacyTier` member is added without updating `_TIER_RANK`, the lookup raises `KeyError` at review time.

One-liner hardening: use `_TIER_RANK.get(tier, <max rank>)` so an unknown tier fails closed to the most-restrictive rank (consistent with the existing UNCLASSIFIED=most-restrictive convention), and/or add a test that every `PrivacyTier` member is present in `_TIER_RANK`.

Not auto-picked (no `spec-decomposition` label).

PR #988, opened 2026-07-29
fix(author): stop the HARD privacy gate missing punctuation-edged verbatim leaks (#939)

### LINKED ISSUE #939: [scan:bugs] _is_verbatim_leak misses HARD privacy-gate leaks when the protected fragment body ends or starts with punctuation
## Role
suppressions).

## Goal
Fix `_is_verbatim_leak` in `creek/author/checks.py` so the deterministic HARD
privacy gate flags a verbatim reproduction of an over-tier fragment even when
that fragment's body begins or ends with punctuation (i.e. virtually all real
prose), verified by a new test that fails today and passes after the fix.

## Context
- File(s): `creek-tools/creek/author/checks.py:239-244` (bug on line 243)
- Evidence:
  ```python
  # checks.py:239-244
  if len(protected.split()) < _MIN_PROTECTED_LEAK_WORDS:
      return False
  normalized_protected = _WHITESPACE_RE.sub(" ", protected).strip()
  normalized_body = _WHITESPACE_RE.sub(" ", body)
  pattern = rf"\b{re.escape(normalized_protected)}\b"          # line 243
  return re.search(pattern, normalized_body) is not None
  ```
  `protected` is the whole over-tier fragment body (stripped). When it ends in a
  non-word character — `.`, `?`, `!`, `"`, `)`, essentially every real sentence —
  `re.escape` turns the trailing `.` into `\.` and the closing `\b` must sit
  between that `.` (a non-word char) and the following space/EOS (also non-word).
  There is no word boundary there, so `re.search` returns `None`,
  `_is_verbatim_leak` returns `False`, `check_privacy_compliance` raises no
  finding, and an exact verbatim reproduction of an INTIMATE/PERSONAL fragment is
  published straight past the deterministic HARD gate. The symmetric case (a
  snippet *starting* with punctuation, leading `\b`) fails identically. Every
  existing test in `tests/test_reflection.py` uses word-char-terminated secrets
  (e.g. `'...session notes'`, `'four word protected secret'`), so this edge is
  never exercised.
- Related: privacy-tier siblings #922, #923, #909 from the same scan family; this
  is a distinct site (author-side reflection HARD gate, not classify/redact).

## Output Format
"Closes #N".

## Examples
Before (edge-char-dependent word boundaries — miss on punctuation-terminated body):
```python
pattern = rf"\b{re.escape(normalized_protected)}\b"
```
After (non-word lookarounds — still reject embedding inside a larger word, still
match a punctuation-terminated snippet):
```python
pattern = rf"(?<!\w){re.escape(normalized_protected)}(?!\w)"
```
Reproducing test sketch (`tests/test_reflection.py`):
```python
def test_privacy_leak_flags_when_protected_body_ends_with_period(...):
    # INTIMATE fragment body ends in '.', contract default tier OPEN,
    # draft body reproduces it verbatim.
    findings = check_privacy_compliance(
        body="Here is the leak: the intimate confession nobody should publish.",
        evidence=...,  # bundle carrying the INTIMATE fragment
        vault=...,
        contract=...,  # default_privacy_tier=OPEN
    )
    assert any(f.kind == "privacy_compliance" for f in findings)  # empty today
```

## Constraints

Geoffe-Ga:
## Summary

`_is_verbatim_leak` (`creek/author/checks.py`) wrapped the escaped protected body in unconditional `\b` anchors:

```python
pattern = rf"\b{re.escape(normalized_protected)}\b"
```

The escaped needle is a **whole vault fragment body**. A `\b` can only assert a word/non-word *transition*, so when the body's own first or last character is punctuation — a `.`, `?`, `"`, `)`, i.e. virtually every real sentence — and the draft's adjacent character is also punctuation or a string end, no boundary exists. `re.search` returns `None`, `check_privacy_compliance` raises no finding, and an **exact verbatim reproduction of an INTIMATE or PERSONAL fragment publishes past the deterministic HARD privacy gate in silence**. Reproduced at HEAD:

| protected body | old result |
| --- | --- |
| `I cheated on my partner last spring.` | `False` — **leak missed** |
| `I cheated on my partner last spring` | `True` — caught |

The fix asserts a boundary **only on an edge whose own character is a word character**:

```python
prefix = r"(?<!\w)" if _WORD_CHAR_RE.match(normalized_protected[0]) else ""
suffix = r"(?!\w)" if _WORD_CHAR_RE.match(normalized_protected[-1]) else ""
pattern = f"{prefix}{re.escape(normalized_protected)}{suffix}"
```

### Why not the unconditional `(?<!\w)…(?!\w)` the issue suggests

The issue (and the architect's first plan) proposed unconditional lookarounds on both edges. I built both patterns and ran them over a matrix of punctuation-edge cases: **the unconditional form is a partial fix that trades one fail-open for another.** It regresses three leak paths `\b` currently catches — precisely the cases where the snippet's punctuation edge abuts a *word* character in the draft, which is where `\b` happened to work:

| protected | draft | `\b` (old) | uncond. lookaround | conditional (this PR) |
| --- | --- | --- | --- | --- |
| `…last spring.` | `…last spring. That is all.` | ❌ miss | ✅ | ✅ |
| `—the affair nobody knows about` | `He said—the affair nobody knows about,` | ✅ | ❌ **regression** | ✅ |
| `I told a lie about the money.` | `…the money.Then it stops.` | ✅ | ❌ **regression** | ✅ |
| `private session notes record` | `Xprivate session notes recordY` | ✅ no flag | ✅ no flag | ✅ no flag |

Un-spaced em dashes are ordinary prose (and LLM prose loves them), so the em-dash row is a real disclosure path, not a synthetic one. The conditional form is a **strict superset** of the old behaviour: identical for word-char-edged snippets, additionally matching punctuation-edged ones, never rejecting anything `\b` accepted. Two tests pin the regression cases so the naive rewrite cannot land later.

### Direction-of-risk / false-positive analysis

A `privacy_compliance` finding is **not** a hard block — `ReflectionNode.review` maps any finding to a `REVISE` verdict (`reflection.py`); the conductor owns retries and the exhaustion→`ESCALATE` policy. So a false positive costs a revise round, while the bug being fixed costs an unredacted intimate disclosure. Even so the negatives are pinned as firmly as the positives: a mid-word substring, a paraphrase, a sub-threshold snippet, an at-ceiling OPEN fragment, and both the leading- and trailing-edge glue cases all assert **no** finding.

`_MIN_PROTECTED_LEAK_WORDS` is **4** and is left unchanged — it is adequate. It short-circuits on whitespace token count, which is orthogonal to edge punctuation; the loosening only affects snippets whose first/last char is non-word, and for those the punctuation must *also* appear literally in the draft, making a ≥4-word match *less* coincidence-prone, not more. Behaviour for every word-char-edged snippet is bit-identical to before.

### `\b` + `re.escape` sweep (`creek/` and `creek_mcp/`)

Nine other sites use this construction. I enumerated the actual runtime values: `classify/rules.py:457`, `author/checks.py:332`, `generate/voice.py:739`, `generate/ai_style/features.py:115` all escape **fixed module-level vocabulary constants** with zero punctuation-edged entries (verified programmatically) — not affected. `generate/cohesion.py:155,158`, `generate/jargon.py:89` and `generate/lexicon.py:287` escape caller-supplied ontology/lexicon vocabulary, and `author/checks.py:478` escapes a pipeline-derived `author_slug`; none is a privacy fail-open, but a punctuation-edged configured term would silently never match. Filed as a follow-up rather than widening this PR. **This site was the only one escaping arbitrary user prose, and the only privacy gate.**

## Test plan

- `cd creek-tools && ./scripts/check-all.sh` → **12/12 checks pass, 0 failed; 6546 passed, 30 deselected; 94.00% total branch coverage; per-file gate 218 files ≥80%.**
- 11 new tests in `tests/test_reflection.py`, all through the public `ReflectionNode().review` path (no private imports, matching the file's convention). RED before the fix:
  ```
  FAILED test_privacy_leak_flags_when_protected_body_ends_with_period - assert False
  FAILED test_privacy_leak_flags_when_protected_body_starts_with_a_quote - assert False
  FAILED test_privacy_leak_flags_when_protected_body_is_punctuated_on_both_edges - assert False
  3 failed, 34 passed
  ```
  after the fix: `39 passed`.
- Mutation-checked by hand — every mutant dies:
  | mutant | killed by |
  | --- | --- |
  | `prefix = ""` | `test_privacy_punctuated_secret_glued_inside_a_word_does_not_flag` |
  | `suffix = ""` | `test_privacy_punctuation_led_secret_glued_at_its_word_tail_does_not_flag` |
  | unconditional `(?<!\w)…(?!\w)` | the two `…abuts_a_word_character` tests |

  The suffix mutant survived the first round — Gate 2.5 review caught it (every trailing-edge case was already blocked by the *leading* assertion), and the third commit adds the isolating test plus a pin on the finding's `HIGH` severity and message content.
- Test-integrity: `git diff --numstat origin/main HEAD` → `creek-tools/tests/test_reflection.py 246 / 0` — **zero deleted lines**; no existing test, fixture, helper or literal was modified. No `# noqa`, `# type: ignore`, `# pylint: disable`, skip marker, threshold change or `--no-verify` anywhere in the diff.

Closes #939

### DIFF
--- a/creek-tools/creek/author/checks.py
+++ b/creek-tools/creek/author/checks.py
@@ -215,6 +215,15 @@ paraphrase and shorter snippets are left to the semantic judge (#474).
 
 _WHITESPACE_RE = re.compile(r"\s+")
 
+_WORD_CHAR_RE = re.compile(r"\w")
+r"""Matches a single word character, testing whether an edge can carry a boundary.
+
+A ``\b`` anchor asserts a word/non-word *transition*, so it can never hold on an
+edge whose own character is already punctuation — as a whole fragment body's
+first/last character usually is (#939). :func:`_is_verbatim_leak` uses this to
+apply a boundary assertion to exactly the edges where one is meaningful.
+"""
+
 
 def _is_verbatim_leak(protected: str, body: str) -> bool:
     """Return whether *protected* appears in *body* as a substantive phrase.
@@ -224,7 +233,10 @@ def _is_verbatim_leak(protected: str, body: str) -> bool:
     * the protected snippet must be at least :data:`_MIN_PROTECTED_LEAK_WORDS`
       words long (a single common word/phrase is too coincidence-prone for a
       HARD gate), and
-    * it must match on word boundaries, so it cannot match inside a larger word.
+    * it must not be glued into a larger word. The boundary is asserted only on
+      an edge whose own character is a word character, so a snippet that starts
+      or ends with punctuation is still detected rather than silently missed
+      (#939).
 
     Whitespace is normalised on both sides first, so a reflowed line break in
     the draft cannot hide an otherwise-verbatim leak.
@@ -240,7 +252,14 @@ def _is_verbatim_leak(protected: str, body: str) -> bool:
         return False
     normalized_protected = _WHITESPACE_RE.sub(" ", protected).strip()
     normalized_body = _WHITESPACE_RE.sub(" ", body)
-    pattern = rf"\b{re.escape(normalized_protected)}\b"
+    # The word-count guard above makes the snippet non-empty here, so its edge
+    # characters are safe to index. Assert a boundary only where one can hold:
+    # a `\b` on a punctuation edge never matches when the draft's adjacent
+    # character is also punctuation or a string end, which let an exact
+    # reproduction of a protected body slip through the HARD gate (#939).
+    prefix = r"(?<!\w)" if _WORD_CHAR_RE.match(normalized_protected[0]) else ""
+    suffix = r"(?!\w)" if _WORD_CHAR_RE.match(normalized_protected[-1]) else ""
+    pattern = f"{prefix}{re.escape(normalized_protected)}{suffix}"
     return re.search(pattern, normalized_body) is not None
 
 
--- a/creek-tools/tests/test_reflection.py
+++ b/creek-tools/tests/test_reflection.py
@@ -752,3 +752,51 @@ def test_privacy_open_tier_punctuated_body_does_not_flag(tmp_path: Path) -> None
     result = ReflectionNode().review(body, evidence, contract=contract, vault=tmp_path)
 
     assert not any(f.dimension == "privacy_compliance" for f in result.findings)
+
+
+def test_privacy_punctuation_led_secret_glued_at_its_word_tail_does_not_flag(
+    tmp_path: Path,
+) -> None:
+    """A punctuation-led secret glued onto a word at its tail does not flag.
+
+    Isolates the trailing boundary assertion (#939): the snippet opens on a
+    quote, so the leading assertion is dropped and cannot mask the result. Its
+    last character is a word character, so that edge must keep its assertion
+    and refuse a match that runs straight into the next word.
+    """
+    secret = '"I never told anyone what really happened'
+    _seed_fragment(tmp_path, "frag-a", secret, tier=PrivacyTier.INTIMATE)
+    evidence = EvidenceBundle(
+        claims=[EvidenceClaim(claim="a claim", source_fragments=["frag-a"])]
+    )
+    contract = MediumContract(medium="research", default_privacy_tier=PrivacyTier.OPEN)
+    body = 'She wrote: "I never told anyone what really happenedX'
+
+    result = ReflectionNode().review(body, evidence, contract=contract, vault=tmp_path)
+
+    assert not any(f.dimension == "privacy_compliance" for f in result.findings)
+
+
+def test_privacy_punctuated_leak_finding_is_high_and_names_the_fragment(
+    tmp_path: Path,
+) -> None:
+    """The punctuation-edged leak finding keeps HIGH severity and full detail.
+
+    Pins the payload of the finding this fix restores (#939) so a downgrade of
+    the HARD privacy gate's severity, or a message that stops naming the
+    offending fragment and its tier, cannot pass silently.
+    """
+    secret = "I cheated on my partner last spring."
+    _seed_fragment(tmp_path, "frag-a", secret, tier=PrivacyTier.INTIMATE)
+    evidence = EvidenceBundle(
+        claims=[EvidenceClaim(claim="a claim", source_fragments=["frag-a"])]
+    )
+    contract = MediumContract(medium="research", default_privacy_tier=PrivacyTier.OPEN)
+    body = "Here is the leak: I cheated on my partner last spring. That is all."
+
+    result = ReflectionNode().review(body, evidence, contract=contract, vault=tmp_path)
+
+    finding = next(f for f in result.findings if f.dimension == "privacy_compliance")
+    assert finding.severity == "HIGH"
+    assert "frag-a" in finding.message
+    assert "intimate" in finding.message
--- a/creek-tools/tests/test_reflection.py
+++ b/creek-tools/tests/test_reflection.py
@@ -554,3 +554,201 @@ def test_conductor_escalates_on_exhaustion(tmp_path: Path) -> None:
 
     assert draft.rounds == 2
     assert draft.verdict == "ESCALATE"
+
+
+def test_privacy_leak_flags_when_protected_body_ends_with_period(
+    tmp_path: Path,
+) -> None:
+    """A leak whose protected body ends in a period must still flag (#939).
+
+    The verbatim matcher wraps the escaped snippet in unconditional word-boundary
+    anchors. A trailing anchor can never assert after a period, so an exact,
+    word-for-word disclosure of an INTIMATE fragment passes the HARD privacy
+    gate in silence.
+    """
+    secret = "I cheated on my partner last spring."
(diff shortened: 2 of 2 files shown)


Geoffe-Ga:
## Summary
PR #988 fixes `_is_verbatim_leak` in `creek/author/checks.py`, which used unconditional `\b...\b` word-boundary anchors around an escaped, arbitrary fragment body. Since `\b` can only fire at a word/non-word transition, it silently failed to match whenever the protected snippet's first/last character was punctuation and the adjacent draft character was also punctuation or a string edge — letting an exact verbatim reproduction of an INTIMATE/PERSONAL fragment slip past the HARD privacy gate undetected. The fix replaces the unconditional anchors with boundary assertions applied only on edges whose own character is a word character.

## Strengths
- Root-caused correctly: the PR body includes a clear before/after matrix showing the old `\b` behavior, a naive unconditional-lookaround "fix" that would have regressed real cases (em-dash-glued and period-glued leaks), and the actual conditional-lookaround fix that is a strict superset of the old matching behavior.
- Excellent TDD trail across 3 commits: RED tests reproducing the bug → GREEN fix → a Gate-2.5-caught surviving mutant (dropped suffix assertion) closed with an isolating test, visible directly in `git log`.
- 11 new tests are well-targeted: the 3 original bug repros, 2 explicit regression guards against the "naive" unconditional-lookaround fix, and negative pins (glued-inside-word, paraphrase, sub-threshold length, at-ceiling OPEN tier, suffix-isolation) so future edits can't silently reopen any of these paths.
- Proactively audited the 9 other `\b`+`re.escape` sites in the codebase, confirmed none is a privacy fail-open, and filed the one legitimate follow-up as issue #989 rather than scope-creeping this PR.
- Zero deleted lines in the test file per the PR's own `git diff --numstat` check — no existing test/fixture was weakened to make this pass.

## Security Concerns
🟢 LOW — this PR *fixes* a real privacy-disclosure bug (an INTIMATE/PERSONAL fragment could publish verbatim past a HARD gate with no finding). The fix itself introduces no new vulnerability: it still uses `re.escape` on the untrusted fragment body, so there's no regex-injection or ReDoS surface (fixed-length literal, no nested quantifiers).

## Problems
None found blocking merge.

## Code Quality
- Minor stylistic note: the new `_WORD_CHAR_RE` module constant's trailing raw-string "docstring" (`r"""Matches a single word character..."""`) follows the exact existing convention in this file (see `_MOST_RESTRICTIVE_RANK`/`_LEAST_RESTRICTIVE_RANK` above it), so it's consistent rather than novel — not a real concern, just noting it's a Sphinx-style attribute-docstring pattern rather than a real Python docstring.
- The double reference to `#939` in both the new comment and the function docstring is slightly redundant but harmless and matches the file's existing citation style (e.g. `#508`, `#509` elsewhere).

## CLAUDE.md Adherence
✅ Conventional Commits followed (`fix(author): ...`, `test(author): ...`) across all 3 commits.
✅ TDD (Gate 1: RED then GREEN, per commit history) followed exactly as required by `creek-tools/CLAUDE.md` §4.1.
✅ No forbidden shortcuts observed: no `# noqa`, `# type: ignore`, skip markers, or threshold changes anywhere in the diff.
✅ PR body reports `./scripts/check-all.sh` green (12/12 checks, 6546 passed, 94.00% branch coverage, per-file gate satisfied) — I could not independently re-run this in the review sandbox (tool execution was blocked), so this is taken on the PR's self-report, which is corroborated by the visible, consistent RED→GREEN commit trail.
✅ Follow-up work correctly filed as a separate issue (#989) instead of being bundled in.

## Testing
Strong. The 11 new tests exercise the fix exclusively through the public `ReflectionNode().review()` path (no private-function imports, matching the file's existing test convention), cover both the positive bug-repro cases and the negative/regression cases that a shallower fix would miss, and one test pins the finding's severity (`HIGH`) and message content so a quiet downgrade of the HARD gate can't land later. This meets the bar for a privacy-critical HARD-gate fix.

## Documentation
Adequate. The function docstring for `_is_verbatim_leak` was updated to describe the new conditional-boundary behavior, and the new `_WORD_CHAR_RE` constant carries an explanatory docstring following the file's existing convention. The PR description itself is exceptionally thorough (root cause, alternative-fix rejection rationale, false-positive/false-negative risk analysis, and an audit of other similar call sites).



## Verdict: LGTM

The fix is a small, correctly-reasoned regex change with 11 new tests (RED before / GREEN after, per the PR's own commit history) that directly pin the bug, the two documented regression cases, and multiple negative/false-positive cases; I traced the `\w`-conditional lookaround logic by hand against every case in the PR's matrix and it holds. No blocking issues found — I'd merge this as-is.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5116112366 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #990, opened 2026-07-29
fix(wavelength): normalise the unclassified sentinel so mode profiles serialise (#940)

### LINKED ISSUE #940: [scan:bugs] Mode profile writes StrEnum members instead of .value, crashing frontmatter.dumps on all-unclassified buckets
## Role
suppressions).

## Goal
Fix `_write_mode_profile` in `creek/generate/wavelength.py` so it passes the
`.value` string (not the raw `StrEnum` member) as the unclassified sentinel,
preventing `yaml.representer.RepresenterError` when a mode bucket's fragments all
have unclassified frequency or phase, verified by a new test that fails today and
passes after the fix.

## Context
- File(s): `creek-tools/creek/generate/wavelength.py:1381-1388` (the bug),
  `:1406-1412` (where the enum member reaches `frontmatter.dumps`); reachable in
  production via the DEFAULT `creek fill` step `report/mode-profiles`
  (`creek/cli.py:1985`) → `ModeProfileGenerator().generate_mode_profiles(...)`.
- Evidence:
  ```python
  # wavelength.py:1381-1388 — passes the ENUM MEMBER as the sentinel
  dominant_frequency = _most_common_classified(
      [str(f.frequency.primary) for f in fragments], Frequency.UNCLASSIFIED)
  dominant_phase = _most_common_classified(
      [str(f.wavelength.phase) for f in fragments], Phase.UNCLASSIFIED)
  # _most_common_classified (wavelength.py:316-320) returns the sentinel unchanged:
  #   if not classified: return unclassified
  # wavelength.py:1406-1412 — the raw member is dumped to YAML frontmatter
  post = frontmatter.Post(..., dominant_frequency=dominant_frequency,
                          dominant_phase=dominant_phase, ...)
  target.write_text(frontmatter.dumps(post), encoding="utf-8")
  ```
  `Frequency` and `Phase` are `StrEnum` subclasses (`creek/models.py:115,136`).
  When every fragment in a mode bucket has an unclassified phase (or frequency) —
  a fragment can have a classified *mode* but unclassified *phase* — the raw
  `Phase.UNCLASSIFIED` / `Frequency.UNCLASSIFIED` member flows into
  `frontmatter.dumps`, which raises
  `yaml.representer.RepresenterError: cannot represent an object` (PyYAML has no
  representer for the enum subclass). Every OTHER call site in this module
  correctly passes `.value` (e.g. `wavelength.py:418`
  `_most_common_classified(phases, Phase.UNCLASSIFIED.value)`), confirming this is
  a slip. Mypy misses it because a `StrEnum` member satisfies the `unclassified:
  str` parameter type. Existing tests (wavelength report tests ~lines 630-658)
  only use fully-classified fragments, so the bucket-unclassified path is never hit.
- Related: sibling generate/ scan finding on compost naive clock this run.

## Output Format
"Closes #N".

## Examples
Before (enum member → RepresenterError on all-unclassified bucket):
```python
_most_common_classified([...], Frequency.UNCLASSIFIED)
_most_common_classified([...], Phase.UNCLASSIFIED)
```
After (plain-string sentinel, matching wavelength.py:418):
```python
_most_common_classified([...], Frequency.UNCLASSIFIED.value)
_most_common_classified([...], Phase.UNCLASSIFIED.value)
```
Reproducing test sketch (`tests/test_wavelength_reports.py`):
```python
def test_mode_profile_with_unclassified_phase_and_frequency(tmp_path):
    frag = _fragment(mode=Mode.EXPRESS, phase=Phase.UNCLASSIFIED,
                     frequency=Frequency.UNCLASSIFIED)
    ModeProfileGenerator().generate_mode_profiles(tmp_path)  # RepresenterError today
    post = frontmatter.load(str(tmp_path / _mode_profile_path("express")))
    assert post["dominant_phase"] == "unclassified"
```

## Constraints

Geoffe-Ga:
## Summary

`ModeProfileGenerator._write_mode_profile` passed the **`StrEnum` members** `Frequency.UNCLASSIFIED` / `Phase.UNCLASSIFIED` as the sentinel to `_most_common_classified`, whose no-classified-values fallback returned that sentinel unchanged. The live member then reached `frontmatter.dumps`, and PyYAML — which has no representer for the enum subclass — raised:

```
yaml.representer.RepresenterError: ('cannot represent an object', <Frequency.UNCLASSIFIED: 'unclassified'>)
```

A fragment can carry a **classified engagement mode** while its frequency and phase stay unclassified, so any such mode bucket crashed the **default** `creek fill` step `report/mode-profiles` (`creek/cli.py:1985`). Reproduced at HEAD `317ea3d` before touching anything.

Neither mypy nor the happy path could see this: `StrEnum` subclasses `str`, so a member satisfies the `unclassified: str` annotation, and it compares equal to its own value, so the filter above the fallback behaves correctly. **Only the fallback return leaked a live member.**

### Fix shape chosen, and why

The issue proposed passing `.value` at the two broken call sites (option A). I chose **option B — normalise inside the helper** — after auditing every call site:

* `_most_common_classified` is module-private with **exactly four** call sites, all in `creek/generate/wavelength.py` (`:418`, `:419`, `:1381`, `:1385`). The blast radius is fully enumerable, which is the precondition for touching a shared helper.
* Option A fixes two lines and leaves the identical landmine armed for call site #5. This exact slip has already been hit and patched **three separate times** in this codebase — `creek/generate/decisions.py:207` carries a `# noqa: FURB123 # yaml.safe_dump cannot serialise StrEnum directly`, `creek/classify/classify_engine.py:1117-1119` carries a `.value` plus an explanatory comment, and `creek/generate/wavelength.py:1351-1355` carries a third. "Remember to write `.value`" demonstrably does not hold at repo scale.
* The sentinel annotation is widened to `StrEnum | str`. This states the contract honestly (callers may legitimately pass either form) and, as a side effect, keeps `str(unclassified)` a genuine conversion rather than a redundant cast — with a bare `str` annotation, `lint-refurb.sh` would flag it `FURB123` and force a suppression, which the house rules forbid. Verified empirically: refurb is clean with the union, and `from __future__ import annotations` lets the `StrEnum` import stay under `TYPE_CHECKING`.
* **The two call sites are deliberately left passing the enum members.** That keeps the normalisation path exercised end-to-end by the new integration test rather than only by a direct unit call.

No tighter type would have caught this. `str` is already technically correct for a `StrEnum` member; the type system has nothing more to say here, which is exactly why the runtime normalisation is the right lever.

### Sweep for siblings

I swept every `frontmatter.dumps` / `frontmatter.Post(` / `yaml.dump` site in `creek/` and `creek_mcp/` (27 `Post` constructions, 20 files) against all 31 `StrEnum` classes, tracing what feeds each kwarg and each `post[...] = ` / `post.metadata.update(...)` assignment.

**Result: no other reachable instance.** The vault models all set `use_enum_values=True` and the writers serialise via `.model_dump(mode="json")`, so attribute access yields plain strings; `creek/purge/engine.py`'s `_CLASSIFICATION_DEFAULTS` are plain-string literals; `flagged_frequencies` and `paradox.frequencies` are `list[str]`.

One **latent, currently-unreachable** trap found and filed separately rather than widened into this PR: `DecisionDetector.update_decision_phase` (`creek/generate/decisions.py:229-264`) takes `new_phase: str`, validates it against `_VALID_PHASES: frozenset[str]` which is built from `DecisionStatus` **members** (`:67-77`), then writes it straight to `post["status"]`. A `DecisionStatus` member would pass the membership check and crash the dump — identical shape to this bug. It has no production caller today (only tests, all passing plain strings), so it is genuinely out of scope here.

### On the class-level guard — one negative result reported honestly

I added `TestUnclassifiedSentinelIsPlainStr`, which asserts `type(result) is str` (not `isinstance` — a `StrEnum` member passes `isinstance`) across every sentinel form. **This is the real family guard:** all four call sites converge on this one return, so any future caller passing a member is covered by construction. It was RED on 5 of its 6 cases.

I also built the broader guard the brief asked for — a class sweeping *every* wavelength note writer against an all-unclassified corpus — and then **deleted it, because I proved it vacuous**: I reintroduced the bug in the helper *and* stripped `.value` from the `:418`/`:419` call sites, and all three of its tests still passed. The reason is structural: `generate_report` / `generate_weekly_report` / `generate_monthly_report` never put a `_most_common_classified` result into frontmatter (their metadata is `period` / `week` / `year` / `tags`), so a sentinel slip there renders as body text and never reaches PyYAML. It also added **zero** marginal branch coverage (`creek/generate/wavelength.py` at 93.44% with and without it). Shipping it would have been three tests that look like a safety net and catch nothing.

## Test plan

* **Gate 1 RED**, final tests against unmodified `origin/main` production code: **7 failed, 29 passed**.
  * `test_bucket_with_no_classified_frequency_or_phase_round_trips` → `yaml.representer.RepresenterError: ('cannot represent an object', <Frequency.UNCLASSIFIED: 'unclassified'>)`
  * `test_fallback_returns_exact_str[frequency|mode|phase|dosage-member]` → `AssertionError: assert <enum 'Frequency'> is str`
  * `test_fallback_result_is_yaml_representable` → same `RepresenterError`
* **Gate 1 GREEN**: 36 passed in `tests/test_wavelength_reports.py`.
* **Gate 2**: `cd creek-tools && ./scripts/check-all.sh` → **12/12 passed, 0 failed**. Full suite **6555 passed, 30 deselected**; aggregate branch coverage **93.99%** (floor 90%); per-file gate 218 files ≥ 80%. One autofixable trailing-newline formatting nit was resolved with `./scripts/fix-all.sh`. No suppressions added anywhere.

### Test-integrity audit

`git diff --numstat origin/main` → `creek-tools/tests/test_wavelength_reports.py: 115 additions, 0 deletions`. No existing test, assertion, or fixture was modified — in particular `_make_fragment`'s classification defaults are byte-identical to `origin/main`, so the all-unclassified bucket this PR exists to fix is still genuinely all-unclassified. Post-commit hash: `b3bcdfc6cd89a9ec48a8bec20b9e1aaf5f3e0179a351b360130f2223e3e2b987`.

Closes #940

### DIFF
--- a/creek-tools/creek/generate/wavelength.py
+++ b/creek-tools/creek/generate/wavelength.py
@@ -28,6 +28,7 @@ from creek.models import Dosage, Fragment, Frequency, Mode, Phase
 from creek.time import effective_authored_date
 
 if TYPE_CHECKING:
+    from enum import StrEnum
     from pathlib import Path
 
 
@@ -313,11 +314,30 @@ def _fragment_in_window(fragment: Fragment, start: date, end: date) -> bool:
     return start <= frag_date <= end
 
 
-def _most_common_classified(values: list[str], unclassified: str) -> str:
-    """Return the most common *values* entry, ignoring *unclassified* markers."""
+def _most_common_classified(values: list[str], unclassified: StrEnum | str) -> str:
+    """Return the most common *values* entry, ignoring *unclassified* markers.
+
+    The sentinel is normalised to a plain :class:`str` before it is
+    returned. Callers may legitimately hand over a ``StrEnum`` member
+    (``Phase.UNCLASSIFIED``) or its ``.value``: the member satisfies a
+    ``str`` annotation, compares equal to its value, and so passes mypy
+    and every comparison here unnoticed — but PyYAML has no representer
+    for the member, so returning it unchanged made the note it landed in
+    fail to serialise with ``RepresenterError`` (issue #940). Normalising
+    at the single point of return keeps that trap disarmed for every
+    caller rather than relying on each one remembering ``.value``.
+
+    Args:
+        values: Candidate markers, typically one per fragment.
+        unclassified: The marker treated as "not classified", returned
+            (as a plain ``str``) when nothing else qualifies.
+
+    Returns:
+        The most frequent classified marker, or the normalised sentinel.
+    """
     classified = [v for v in values if v and v != unclassified]
     if not classified:
-        return unclassified
+        return str(unclassified)
     counter = Counter(classified)
     return counter.most_common(1)[0][0]
 
--- a/creek-tools/tests/test_wavelength_reports.py
+++ b/creek-tools/tests/test_wavelength_reports.py
@@ -17,6 +17,7 @@ from creek.generate.wavelength import (
     PHASE_DOMAIN_MAPPINGS,
     ModeProfileGenerator,
     WavelengthTracker,
+    _most_common_classified,
 )
 from creek.models import (
     Confidence,
@@ -668,3 +669,117 @@ class TestModeProfileGenerator:
 
         assert written == []
         assert not (vault_path / "05-Wavelength" / "Mode-Profiles").exists()
+
+    def test_bucket_with_no_classified_frequency_or_phase_round_trips(
+        self,
+        vault_path: Path,
+    ) -> None:
+        """A classified *mode* with unclassified frequency/phase still serialises.
+
+        A fragment may carry a classified engagement mode while its
+        frequency and phase remain unclassified, so a mode bucket can
+        contain nothing classified along either of those axes. The
+        dominant-value fallback must yield a plain ``str``; a raw
+        ``StrEnum`` member reaches PyYAML, which has no representer for
+        it and raises ``yaml.representer.RepresenterError`` (issue #940).
+        """
+        _seed_fragments(
+            vault_path,
+            [
+                _make_fragment(
+                    frag_id="mu1",
+                    created=datetime(2026, 4, 20, 12, 0, tzinfo=UTC),
+                    mode=Mode.EXPRESS,
+                    frequency=Frequency.UNCLASSIFIED,
+                    phase=Phase.UNCLASSIFIED,
+                ),
+            ],
+        )
+
+        written = ModeProfileGenerator().generate_mode_profiles(vault_path)
+
+        assert [p.name for p in written] == ["express.md"]
+        post = frontmatter.load(str(written[0]))
+        assert post["dominant_frequency"] == Frequency.UNCLASSIFIED.value
+        assert post["dominant_phase"] == Phase.UNCLASSIFIED.value
+
+    def test_bucket_with_classified_frequency_but_unclassified_phase(
+        self,
+        vault_path: Path,
+    ) -> None:
+        """The phase axis falls back independently of the frequency axis.
+
+        Guards the half-classified bucket: were only the frequency
+        sentinel repaired, this note would still fail to serialise.
+        """
+        _seed_fragments(
+            vault_path,
+            [
+                _make_fragment(
+                    frag_id="mp1",
+                    created=datetime(2026, 4, 20, 12, 0, tzinfo=UTC),
+                    mode=Mode.INHABIT,
+                    frequency=Frequency.F6,
+                    phase=Phase.UNCLASSIFIED,
+                ),
+            ],
+        )
+
+        written = ModeProfileGenerator().generate_mode_profiles(vault_path)
+
+        post = frontmatter.load(str(written[0]))
+        assert post["dominant_frequency"] == Frequency.F6.value
+        assert post["dominant_phase"] == Phase.UNCLASSIFIED.value
+
+
+class TestUnclassifiedSentinelIsPlainStr:
+    """Every dominant-value fallback must be YAML-representable (#940).
+
+    ``Frequency``/``Mode``/``Phase`` are ``StrEnum`` subclasses, so a
+    member satisfies a ``str`` annotation, compares equal to its value,
+    and passes mypy — yet PyYAML's ``SafeDumper`` has no representer for
+    it and raises ``RepresenterError`` on dump. These tests pin the
+    class of bug rather than the single instance of it.
+    """
+
+    @pytest.mark.parametrize(
+        "sentinel",
+        [
+            Frequency.UNCLASSIFIED,
+            Mode.UNCLASSIFIED,
+            Phase.UNCLASSIFIED,
+            Dosage.UNCLASSIFIED,
+            Frequency.UNCLASSIFIED.value,
+            "unclassified",
+        ],
+        ids=[
+            "frequency-member",
+            "mode-member",
+            "phase-member",
+            "dosage-member",
+            "frequency-value",
+            "bare-string",
+        ],
+    )
+    def test_fallback_returns_exact_str(self, sentinel: str) -> None:
+        """The no-classified-values fallback returns an exact ``str``.
+
+        ``type(...) is str`` rather than ``isinstance`` on purpose: a
+        ``StrEnum`` member passes ``isinstance(x, str)``, so only an
+        exact type check catches the leak.
+        """
+        result = _most_common_classified([], sentinel)
+
+        assert type(result) is str
+        assert result == "unclassified"
+
+    def test_fallback_result_is_yaml_representable(self) -> None:
+        """The fallback value survives a frontmatter round trip."""
+        result = _most_common_classified(
+            [Phase.UNCLASSIFIED.value],
+            Phase.UNCLASSIFIED,
+        )
+
+        dumped = frontmatter.dumps(frontmatter.Post(content="", phase=result))
+
+        assert frontmatter.loads(dumped)["phase"] == Phase.UNCLASSIFIED.value


Geoffe-Ga:
Sweep follow-up filed as #991 (the latent `update_decision_phase` trap described under "Sweep for siblings" above). It has no production caller today, so it stays out of this PR.

Geoffe-Ga:
## Summary
Fixes issue #940: `_most_common_classified`'s no-classified-values fallback returned the raw `StrEnum` sentinel (`Frequency.UNCLASSIFIED` / `Phase.UNCLASSIFIED`) unchanged, which PyYAML cannot serialize, crashing the default `creek fill` step whenever a fragment's mode was classified but its frequency/phase were not. The fix normalizes the sentinel to `str(unclassified)` at the single point of return inside the helper, rather than patching each call site individually.

## Strengths
- **Root-cause fix at the right altitude**: normalizing inside the shared helper (4 call sites, all in this module) closes the trap for every current and future caller, instead of patching just the two broken lines. The PR body documents an explicit audit of why this scope was chosen over the narrower alternative.
- **Type contract is honest**: widening the annotation to `StrEnum | str` (guarded by `TYPE_CHECKING` + `from __future__ import annotations`, confirmed present in the file) states what callers actually pass, and keeps `str(unclassified)` a real conversion rather than a redundant cast that refurb would flag.
- **Precise regression guard**: `TestUnclassifiedSentinelIsPlainStr` asserts `type(result) is str` rather than `isinstance`, which is the only way to actually catch a `StrEnum` member leaking through (`isinstance(x, str)` is true for `StrEnum` members, so a weaker assertion would pass even with the bug present).
- **Honest negative result reported**: the author built a broader "every wavelength writer" guard, proved it vacuous (still passed with the bug reintroduced, zero marginal coverage), and deleted it rather than shipping tests that look like safety nets but catch nothing. This is exactly the kind of test-integrity discipline the repo's quality bar calls for.
- **Consistency with prior art**: the PR body cross-references three earlier instances of this same StrEnum/YAML footgun elsewhere in the codebase (`decisions.py`, `classify_engine.py`), and I confirmed `classify_engine.py:1119` follows the same `.value` normalization idea — this fix's approach (normalize once, centrally) is a strict improvement over the ad hoc per-site pattern used previously.
- **Deliberate test design choice**: the two call sites that triggered the original bug are left passing raw enum members (not `.value`) so the new tests exercise the actual normalization path end-to-end rather than only unit-testing the helper directly.

## Security Concerns
None. This is a pure serialization/type-normalization bug fix with no security surface.

## Problems
None found. No bugs, no CLAUDE.md violations blocking merge.

## Code Quality
- The docstring on `_most_common_classified` is long (Args/Returns plus a paragraph explaining the historical bug) — reasonable given the non-obvious WHY (mypy/isinstance can't catch this), and it directly documents the invariant callers rely on, so this is not a nit worth blocking on.
- Minor stylistic note: the two call sites at `wavelength.py:1401-1408` still pass the bare enum member (`Frequency.UNCLASSIFIED`) while the other two at `:438-439` pass `.value`. This inconsistency is intentional per the PR body (to keep the normalization path covered by integration tests) and is explained, so not a real concern — just worth knowing if a future reader wonders why the call sites differ.

## CLAUDE.md Adherence
✅ TDD followed (RED reported: 7 failed/29 passed against unmodified production code; GREEN: 36 passed).
✅ `./scripts/check-all.sh` reported green (12/12), full suite 6555 passed, aggregate branch coverage 93.99% (floor 90%), per-file gate passing.
✅ No suppressions (`# noqa`, `type: ignore`) added.
✅ Docstrings added in Google style with Args/Returns sections, consistent with the ≥95% interrogate requirement.
✅ Root-cause fix, not a shortcut — matches the "No Shortcuts" principle (normalizes at the source rather than patching call sites and hoping future callers remember `.value`).
✅ Commit/PR body follows the issue-first workflow (`Closes #940`) with a detailed, honest test-integrity audit.

## Testing
Strong. Includes: (1) the original crash reproduction (all-unclassified bucket round-trips through frontmatter), (2) a half-classified variant (frequency classified, phase not) guarding independent-axis fallback, (3) a parametrized class-level guard across all four `StrEnum` types plus `.value` and bare-string forms verifying `type(result) is str`, and (4) a YAML round-trip test. The author also reports deleting a broader guard class after proving it added zero marginal coverage — a good instinct that keeps the test suite meaningful rather than padded.

## Documentation
The function docstring explains the non-obvious WHY (mypy/isinstance can't catch a `StrEnum` member satisfying a `str` contract) which is exactly the kind of comment worth keeping. PR description is unusually thorough — includes root cause, fix rationale, a sibling-site sweep, and a documented negative result. Nothing further needed.



## Verdict: LGTM

The fix correctly normalizes the sentinel to a plain str at the single return point in `_most_common_classified`, is exercised end-to-end by both call sites that previously crashed, includes a dedicated class-level regression guard (`type(result) is str`, not `isinstance`), and the PR body documents a rigorous TDD cycle (RED→GREEN), a full sweep for sibling instances, and an honest negative result (a broader guard class that was built then deleted after being proven vacuous). No blocking issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5116644224 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #1003, opened 2026-07-30
fix(compile): reject non-object claim/paradox elements at the parse boundary

### LINKED ISSUE #941: [scan:bugs] compile _parse_llm_payload skips element-type check: string claims crash with AttributeError instead of ValueError
## Role
suppressions).

## Goal
Make `_parse_llm_payload` in `creek/compile/engine.py` validate that each
`claims`/`paradoxes` element is a dict, so a well-formed-JSON-but-wrong-shape LLM
payload surfaces the intended unified `ValueError` at parse time instead of an
`AttributeError` crashing out of `compile_to_vault`, verified by a new test that
fails today and passes after the fix.

## Context
- File(s): `creek-tools/creek/compile/engine.py:447-458` (the incomplete
  validation), `:205` and `:238` (the downstream `.get` that raises), `:144-145`
  (the unguarded call site).
- Evidence:
  ```python
  # engine.py:453-458 — validates the containers are lists, NOT their elements
  claims = decoded.get("claims") or []
  paradoxes = decoded.get("paradoxes") or []
  if not isinstance(claims, list) or not isinstance(paradoxes, list):
      msg = "Compile LLM payload claims/paradoxes must be arrays."
      raise ValueError(msg)  # noqa: TRY004  # ValueError unifies all LLM-payload schema errors ...
  return {"claims": list(claims), "paradoxes": list(paradoxes)}
  # engine.py:205 — downstream assumes each element is a dict
  [c for c in claims if str(c.get("id", "")).strip()]
  ```
  The `# noqa: TRY004` comments (lines 452, 457) document the deliberate intent:
  "ValueError unifies all LLM-payload schema errors." But element types are never
  checked. A well-formed-JSON payload of the wrong shape — e.g.
  `{"claims": ["insight one", "insight two"], "paradoxes": []}`, which any LLM can
  emit — passes `_parse_llm_payload` and reaches `_filter_valid_claims`
  (`engine.py:145` → `:205`), where `c.get(...)` raises
  `AttributeError: 'str' object has no attribute 'get'` (same for string paradox
  elements at `:238`). There is NO try/except around `engine.py:144-145`, so the
  `AttributeError` propagates out of `compile_to_vault` as a raw traceback instead
  of the clean `ValueError` the tolerant-JSON machinery exists to produce. The
  declared return type `list[dict[str, object]]` is unenforced at runtime.
- Related: other open compile issues (#928/#929/#930/#931) concern tier routing,
  a bad import, and double-walk — this is a distinct payload-validation defect.

## Output Format
"Closes #N".

## Examples
Before (list check only — non-dict elements slip through to a downstream crash):
```python
if not isinstance(claims, list) or not isinstance(paradoxes, list):
    raise ValueError("Compile LLM payload claims/paradoxes must be arrays.")
return {"claims": list(claims), "paradoxes": list(paradoxes)}
```
After (reject non-dict elements at parse time, keeping the unified ValueError):
```python
if not isinstance(claims, list) or not isinstance(paradoxes, list):
    raise ValueError("Compile LLM payload claims/paradoxes must be arrays.")
if not all(isinstance(c, dict) for c in claims) or not all(
    isinstance(p, dict) for p in paradoxes
):
    raise ValueError("Compile LLM payload claims/paradoxes items must be objects.")
return {"claims": list(claims), "paradoxes": list(paradoxes)}
```
Reproducing test sketch (`tests/test_compile.py`):
```python
def test_parse_llm_payload_rejects_string_claim_elements():
    with pytest.raises(ValueError):
        _parse_llm_payload('{"claims": ["a", "b"], "paradoxes": []}')
    # today: returns successfully, then downstream raises AttributeError
```

## Constraints

Geoffe-Ga:
## Summary

`_parse_llm_payload` (`creek/compile/engine.py`) validated that `claims` and `paradoxes` were *lists* but never validated their *element* types. A well-formed-JSON payload of the wrong shape — e.g. `{"claims": ["insight one"], "paradoxes": []}`, which any model can emit — parsed cleanly and then crashed downstream in `_filter_valid_claims` with `AttributeError: 'str' object has no attribute 'get'`. String *paradox* elements crashed identically in `_payload_to_paradox_entries`. Nothing wraps those calls, so the `AttributeError` escaped `compile_fragments` as a raw traceback instead of the `ValueError` the tolerant-JSON machinery exists to produce — and `creek_mcp/tools/compile.py` catches `(ValueError, RuntimeError)` to build a structured refusal but **not** `AttributeError`, so the bug left an MCP tool unhandled.

The fix adds one guard at the parse boundary:

```python
if any(not isinstance(item, dict) for item in (*claims, *paradoxes)):
    msg = "Compile LLM payload claims/paradoxes must contain JSON objects."
    raise ValueError(msg)
```

### The design decision

Two shapes were considered — validate at the parse boundary (**reject the whole payload**), or make `_filter_valid_claims` skip non-dict elements (**drop the bad ones and proceed**). These differ in product behaviour, so the choice is stated explicitly:

**Chosen: reject the whole payload.**

- **Category.** `_parse_llm_payload` owns payload *schema* validation and already rejects a wrong-typed container with `ValueError`. `_filter_valid_claims` is a *render-safety* filter — its docstring says it exists so `_render_body` does not emit a broken `[^]` footnote — for claims that are structurally valid but incomplete. A bare-string element means the model ignored the schema outright, which belongs in the reject tier, not the drop tier. **`_filter_valid_claims` is unchanged**, and the two existing tests that pin its silent-drop semantics (`test_compile_fragments_skips_claim_with_missing_id`, `test_compile_fragments_skips_empty_claim_text`) still pass untouched — that is the proof this stayed consistent rather than inventing a third behaviour.
- **Blast radius of the alternative.** A model that emits one string claim emitted a *list* of strings, so dropping bad elements realistically leaves zero claims. `compile_to_vault` calls `_write_compiled_page` unconditionally, `page.body` is rendered fresh from the surviving claims, and `merge_provenance` with an empty new list *preserves* existing entries. So a re-compile would rewrite the page with the body hollowed down to `# Title` while stale provenance survived in the frontmatter — an internally inconsistent note, reported as success. `creek/compile/__init__.py` names exactly that hazard: per-claim provenance is "non-negotiable … lossy compression here [is] a load-bearing risk for voice fidelity."
- **Conservatism.** Today that payload already aborts before any write (the `AttributeError` fires first), so this changes only the exception *type*, never what gets written.

**User-visible consequence:** a compile whose LLM emits off-schema elements now fails with `ValueError: Compile LLM payload claims/paradoxes must contain JSON objects.` The MCP `creek.compile` tool surfaces that as a structured refusal instead of an unhandled exception. `creek/cli.py` does not catch `ValueError` either, so `creek compile` still shows a traceback — just a semantically correct one (pre-existing gap, filed as a follow-up). No page is written and no existing page is modified; the operator retries.

**`paradoxes` are covered too.** The issue focused on claims, but line 455 validates both arrays together and `_payload_to_paradox_entries` has the identical `item.get` crash with no filter of its own, so the single guard covers both and a test pins each independently.

## Test plan

Gate 1 RED first. All four new tests go through the **public** `compile_fragments` / `compile_to_vault`, not `_parse_llm_payload` directly — at HEAD a direct unit test on the parser would *pass* (it returns successfully), so it would not reproduce the bug at all. The defect only manifests across the parse to filter boundary.

Confirmed RED before implementation — `E AttributeError: 'str' object has no attribute 'get'` at `creek/compile/engine.py:205`, on all four:

- `test_compile_fragments_rejects_string_claim_elements` — the issue's literal reproduction.
- `test_compile_fragments_rejects_string_paradox_elements` — well-formed claims must not rescue malformed paradoxes.
- `test_compile_fragments_rejects_mixed_claim_elements` — one good claim plus one bare string; asserts the **whole** payload is rejected rather than the good claim surviving. This is where the design decision is encoded.
- `test_compile_to_vault_leaves_existing_page_untouched_on_string_claims` — compiles a page, captures its bytes, re-compiles with a string-claims payload, asserts the raise **and** byte-identity on disk. Pins the abort-before-write contract and guards against any future slide toward drop-and-continue.

Test file diff is **+134 / -0** — purely additive, no existing test or assertion modified.

Runtime edge-case probe of the guard: string / `null` / nested-list / numeric elements all rejected; `{}`, missing keys, and `"claims": null` still tolerated as `[]` (pre-existing `or []` behaviour, unchanged).

Gate 2: `cd creek-tools && ./scripts/check-all.sh` exits **0** — 6555 unit tests pass, coverage **93.99%**. Tryceratops passes with **no `# noqa: TRY004`** on the new raise: TRY004's detector only matches a bare `isinstance` as the `if` test and does not walk into generator expressions. Verified in both directions — the new raise without a noqa is clean, while stripping the two *existing* noqas on lines 452/457 does produce TRY004 errors. Adding a dead suppression there would itself violate the noqa hygiene rule. `radon` puts `_parse_llm_payload` at B(6) to B(8); `xenon --max-absolute B` passes with headroom.

Gate 2.5: independent `code-review-orchestrator` pass over the diff — verdict **CLEAN**, zero blockers.

## Non-blocking notes from review (not fixed here)

- `(*claims, *paradoxes)` materialises the concatenated tuple before `any()` short-circuits. Harmless at LLM-payload sizes; `itertools.chain` would restore true laziness if this line is ever touched again.
- `decoded.get("claims") or []` still coerces any *falsy* wrong-typed value (`0`, `""`, `False`) to `[]`, slipping past both the container check and this new guard. Pre-existing and out of scope; filed as a follow-up.

Closes #941

### DIFF
--- a/creek-tools/creek/compile/engine.py
+++ b/creek-tools/creek/compile/engine.py
@@ -127,7 +127,8 @@ def compile_fragments(
 
     Raises:
         ValueError: If *target_kind* is not one of the supported
-            compiled-page surfaces.
+            compiled-page surfaces, or if the LLM payload's ``claims``
+            or ``paradoxes`` are not arrays of JSON objects.
     """
     _validate_target_kind(target_kind)
     pairs = list(fragments_with_bodies)
@@ -455,6 +456,9 @@ def _parse_llm_payload(raw: str) -> dict[str, list[dict[str, object]]]:
     if not isinstance(claims, list) or not isinstance(paradoxes, list):
         msg = "Compile LLM payload claims/paradoxes must be arrays."
         raise ValueError(msg)  # noqa: TRY004  # ValueError unifies all LLM-payload schema errors with the JSONDecodeError branch above.
+    if any(not isinstance(item, dict) for item in (*claims, *paradoxes)):
+        msg = "Compile LLM payload claims/paradoxes must contain JSON objects."
+        raise ValueError(msg)
     return {"claims": list(claims), "paradoxes": list(paradoxes)}
 
 
--- a/creek-tools/tests/test_compile.py
+++ b/creek-tools/tests/test_compile.py
@@ -726,6 +726,140 @@ def test_compile_fragments_rejects_non_array_claims() -> None:
         )
 
 
+def test_compile_fragments_rejects_string_claim_elements() -> None:
+    """Bare-string ``claims`` elements are rejected at the parse boundary.
+
+    An LLM returning ``{"claims": ["insight one"]}`` — prose strings
+    where claim objects belong — cleared the list-shape check and then
+    crashed downstream in ``_filter_valid_claims`` with
+    ``AttributeError: 'str' object has no attribute 'get'``. Element
+    types are part of the payload schema, so the rejection belongs at
+    the parse boundary with the other schema errors.
+    """
+    frag = _make_fragment(frag_id="frag-aaa")
+    bad = json.dumps({"claims": ["insight one", "insight two"], "paradoxes": []})
+    llm, _ = _make_llm([bad])
+    with pytest.raises(ValueError, match="must contain JSON objects"):
+        compile_fragments(
+            [(frag, "body")],
+            llm=llm,
+            target_kind="thread",
+            target_id="t",
+            target_title="T",
+        )
+
+
+def test_compile_fragments_rejects_string_paradox_elements() -> None:
+    """Bare-string ``paradoxes`` elements fail under the same guard.
+
+    ``_payload_to_paradox_entries`` calls ``item.get`` exactly the way
+    the claim filter does, so a string paradox crashes identically.
+    Well-formed claims must not rescue a payload whose paradoxes are
+    malformed — both arrays are validated, not just the first one.
+    """
+    frag = _make_fragment(frag_id="frag-aaa")
+    bad = json.dumps(
+        {
+            "claims": [{"id": "c1", "text": "X", "fragment_ids": ["frag-aaa"]}],
+            "paradoxes": ["they contradict"],
+        },
+    )
+    llm, _ = _make_llm([bad])
+    with pytest.raises(ValueError, match="must contain JSON objects"):
+        compile_fragments(
+            [(frag, "body")],
+            llm=llm,
+            target_kind="thread",
+            target_id="t",
+            target_title="T",
+        )
+
+
+def test_compile_fragments_rejects_mixed_claim_elements() -> None:
+    """One bad element rejects the whole payload; the good claim is not kept.
+
+    Whole-payload rejection is the deliberate design decision here, and
+    this test is where it is encoded. The alternative — silently
+    dropping the bad element and compiling the survivors — would let a
+    truncated or prose-contaminated response write a compiled page whose
+    body is wiped down to just the title while stale provenance survives
+    in frontmatter. A quietly corrupted note is worse than a loud
+    failure the operator can retry, so the parser refuses the payload
+    whole rather than salvaging part of it.
+
+    This is distinct from ``_filter_valid_claims``, which keeps its
+    silent-drop semantics for structurally valid claim objects that are
+    merely incomplete (see
+    ``test_compile_fragments_skips_claim_with_missing_id``).
+    """
+    frag = _make_fragment(frag_id="frag-aaa")
+    payload: dict[str, object] = {
+        "claims": [
+            {"id": "c1", "text": "X", "fragment_ids": ["frag-aaa"]},
+            "a bare string",
+        ],
+        "paradoxes": [],
+    }
+    llm, _ = _make_llm([json.dumps(payload)])
+    with pytest.raises(ValueError, match="must contain JSON objects"):
+        compile_fragments(
+            [(frag, "body")],
+            llm=llm,
+            target_kind="thread",
+            target_id="t",
+            target_title="T",
+        )
+
+
+def test_compile_to_vault_leaves_existing_page_untouched_on_string_claims(
+    vault: Path,
+) -> None:
+    """A malformed re-compile aborts before it touches the page on disk.
+
+    ``compile_to_vault`` parses the LLM payload before ``_write_compiled_page``
+    runs, so a schema violation on a re-run must leave the previously
+    compiled page byte-identical. This pins the conservative
+    abort-before-write contract and guards against any future slide
+    toward drop-the-bad-element-and-continue, which would rewrite the
+    page with a hollowed-out body.
+    """
+    frag = _make_fragment(frag_id="frag-aaa")
+    _write_fragment_to_vault(vault, frag, "body")
+    good = _llm_response(
+        claims=[
+            {
+                "id": "claim-001",
+                "text": "Stable claim.",
+                "fragment_ids": ["frag-aaa"],
+            },
+        ],
+    )
+    bad = json.dumps({"claims": ["insight one"], "paradoxes": []})
+    llm, _ = _make_llm([good, bad])
+
+    target = compile_to_vault(
+        fragment_ids=["frag-aaa"],
+        vault_path=vault,
+        target_kind="thread",
+        target_id="thread-untouched",
+        target_title="Untouched",
+        llm=llm,
+    )
+    before = target.read_text(encoding="utf-8")
+    assert "Stable claim." in before
+
+    with pytest.raises(ValueError, match="must contain JSON objects"):
+        compile_to_vault(
+            fragment_ids=["frag-aaa"],
+            vault_path=vault,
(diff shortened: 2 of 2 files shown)


Geoffe-Ga:
## Summary
This PR closes a crash bug in `creek/compile/engine.py`: `_parse_llm_payload` validated that `claims`/`paradoxes` were lists but not that their *elements* were JSON objects, so a well-formed-but-wrong-shaped LLM payload (e.g. `{"claims": ["insight one"]}`) parsed successfully and then crashed downstream with an unhandled `AttributeError` — which also escaped the MCP tool's `(ValueError, RuntimeError)` catch clause. The fix adds a single `isinstance` guard at the parse boundary that rejects the whole payload with a `ValueError`, and is backed by four new tests.

## Strengths
- Root-cause fix at the correct architectural boundary (`_parse_llm_payload` already owns payload-schema validation; the new guard is one line, symmetric with the existing list-shape check).
- Excellent PR description: explicitly states the two behavioral options considered (reject-whole-payload vs. drop-bad-elements) and justifies the choice using the codebase's own stated non-negotiable (per-claim provenance integrity in `creek/compile/__init__.py`), including a concrete "hollowed body, stale provenance" failure scenario for the rejected alternative.
- Test coverage is thorough and non-redundant: string claim elements, string paradox elements (proving both arrays are validated, not just the first), mixed valid/invalid elements (pins the whole-payload-reject decision), and a byte-identity-on-disk regression test proving `compile_to_vault` aborts before any write on re-compile.
- Confirms `_filter_valid_claims`'s pre-existing silent-drop semantics for structurally-valid-but-incomplete claims are untouched and still pinned by their own tests — the PR doesn't conflate the two different validation tiers (schema-reject vs. render-safety-drop).
- Honest about two acknowledged, explicitly out-of-scope follow-ups (`(*claims, *paradoxes)` tuple materialization; falsy-value coercion via `or []`) rather than silently leaving them or over-scoping the fix.
- Confirmed the new `raise` needed no `# noqa: TRY004` (documented why, and verified in both directions) — no linter suppression added.

## Security Concerns
None. 🟢 LOW — this is a defensive input-validation fix for untrusted/model-generated payloads, and it tightens validation rather than loosening it. No injection, path traversal, or secret-handling surface touched.

## Problems
None found. No bugs, no regressions, no CLAUDE.md violations.

## Code Quality
- `(*claims, *paradoxes)` builds a concatenated tuple before `any()` can short-circuit — harmless at realistic LLM-payload sizes, and the author already flagged it as a non-blocking follow-up if the line is touched again (`itertools.chain` would restore laziness). Not worth blocking on.
- `decoded.get("claims") or []` still coerces falsy wrong-typed values (`0`, `""`, `False`) to `[]`, bypassing both the container check and the new guard. Acknowledged as pre-existing and out of scope by the author; worth an issue if not already filed.

## CLAUDE.md Adherence
✅ TDD followed (RED confirmed via `AttributeError` before the fix, then GREEN). ✅ `./scripts/check-all.sh` reported passing (6555 tests, 93.99% coverage, above the 90% branch-coverage gate). ✅ No new lint suppressions added; existing noqa hygiene preserved. ✅ Complexity within bounds (`_parse_llm_payload` at B(6)–B(8), `xenon --max-absolute B` passes). ✅ Docstrings updated for the new `Raises` behavior on `compile_fragments`. Diff is scoped to a feature branch off `main`, consistent with the repo's branch workflow.

## Testing
Strong. Four new tests exercise the bug through the public `compile_fragments`/`compile_to_vault` API (correctly avoiding a unit test directly on `_parse_llm_payload`, which the author notes would not reproduce the defect since the parser itself returns successfully — the crash is at the parse→filter boundary). Coverage includes: string claim elements, string paradox elements, mixed valid/invalid elements (encodes the whole-payload-reject decision), and a disk byte-identity check proving no partial write occurs on abort. Existing tests pinning `_filter_valid_claims`'s drop semantics were left untouched, confirming no behavioral drift elsewhere.

## Documentation
`compile_fragments`'s docstring `Raises` section was updated to describe the new rejection condition. The PR description itself is exceptionally thorough — states the bug, the design decision with alternatives considered and rejected, and explicit follow-up notes for two acknowledged non-blocking issues.



## Verdict: LGTM

The fix is a minimal, correct guard clause that closes a real unhandled-AttributeError crash path, is backed by four new tests (including a byte-identity regression test) that reuse existing fixtures, keeps `_filter_valid_claims`'s drop-semantics contract intact and pinned by pre-existing tests, and the PR description documents a deliberate, well-reasoned design choice (reject-whole-payload vs. drop-bad-elements) with two explicitly scoped follow-up notes. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5133792219 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #1010, opened 2026-07-30
fix(redact): detect a high-entropy secret masked by a low-entropy neighbour

### LINKED ISSUE #942: [redact] low-entropy neighbour masks an adjacent high-entropy secret from the detector
## Summary

The high-entropy detector scores an entire contiguous `[A-Za-z0-9+/=_\-]` run as a
single candidate. When a genuinely random secret is glued with no delimiter to a
low-entropy neighbour, the *average* entropy of the combined run can fall under
`entropy_threshold(min_confidence)` even though a >=20-char high-entropy sub-run
would clear it on its own. The low-entropy neighbour effectively **masks** the
secret from the detector.

This affects `RedactionScanner._scan_high_entropy` and
`Redactor._collect_high_entropy_spans` identically — they share the gating by
design — so both `--scan` and `--apply` miss it. Scan/apply parity is *not*
violated; the detector simply never sees the secret.

## Reproducer (verified at HEAD of the #909 branch)

```python
from creek.config import RedactionConfig
from creek.redact.scanner import RedactionScanner, shannon_entropy, entropy_threshold
from creek.redact.redactor import Redactor

cfg = RedactionConfig()
s = RedactionScanner(config=cfg)
r = Redactor(config=cfg, salt=s.salt)

content = "AKIA" + "A" * 16 + "Zq7Wp3Rf9Tk2NgXy5Bc8"

shannon_entropy(content)                       # 3.445  <- under threshold
entropy_threshold(cfg.min_confidence)          # 3.700
shannon_entropy("Zq7Wp3Rf9Tk2NgXy5Bc8")        # 4.322  <- would clear it alone

r.redact_content(content)
# '[REDACTED:api_key]Zq7Wp3Rf9Tk2NgXy5Bc8'   <- 20-char high-entropy tail in cleartext
```

A `--scan` of the same content reports only `{'api_key'}` — the tail is invisible
to the detector too.

## Why this is not #909

#909 was the *ordering* bug: the entropy detector ran as a post-pass over
already-spliced text, so a partially-overlapping regex match destroyed the run it
needed to match. That is fixed (PR for #909) — entropy candidates are now
collected against the original content and unioned.

This issue is a *detector granularity* bug that predates #909 and lives equally in
the scanner. It is reachable only when the masking neighbour is near-constant
(H1 <~ 1.7 bits/char); a realistic AKIA-shaped key measures ~3.7-4.2 bits/char, so
real secrets do not trigger it. It was found during the #909 Gate 2.5 review and
deliberately scoped out rather than widening that PR.

## Suggested fix

After merging spans, re-score the sub-runs of each entropy candidate that are
*not* covered by a regular-pattern span, and flag any that independently clear the
threshold. This must land in **both** `RedactionScanner._scan_high_entropy` and
`Redactor._collect_high_entropy_spans` in lockstep, or scan/apply parity breaks.
Add a shared helper rather than a third copy of the gate.

## Acceptance

- A test pinning the reproducer above: `--scan` reports `high_entropy_string`, and
  `redact_content` leaves no cleartext fragment of `Zq7Wp3Rf9Tk2NgXy5Bc8`.
- Existing entropy tests (threshold, allowlist, `min_confidence` tuning) stay green.
- `cd creek-tools && ./scripts/check-all.sh` exits 0.

Geoffe-Ga:
## Summary

A low-entropy neighbour glued to a genuine secret hid it from the generic
high-entropy detector, in both `--scan` and `--apply`. Verified live at HEAD:

```
"aB3xK9mQ7zR2vN5tL8wY"        -> 4.3219 bits/char, clears the 3.70 default
"a" * 20 + that same secret    -> 3.0160 bits/char, fails it
redact_content(...)            -> all 40 characters returned in cleartext
```

Both call sites scored the *whole* `HIGH_ENTROPY_CANDIDATE` run's average, and
concatenation averages. Scan/apply parity was never violated — the detector
simply never fired.

**Note on the issue text:** #942's own reproducer (`"AKIA" + "A"*16 + tail`) is
already fixed at HEAD — #909's token snapping widens the `api_key` match to the
whole run and it yields `[REDACTED:api_key]`. Snapping only helps when a regex
span already exists inside the run; the general case has no anchor. The issue's
suggested fix (re-score sub-runs *not covered by a pattern span*) does not reach
it for the same reason. This PR fixes the general case.

### The fix

One shared `creek.redact.scanner.has_high_entropy_region(text, threshold)` that
both `RedactionScanner._scan_high_entropy` and `Redactor._collect_high_entropy_spans`
now call, in the position their two hand-synchronised inline comparisons held —
immediately after the allowlist check. **Scan/apply parity now holds by
construction, not by convention.** A run qualifies when the whole run clears the
threshold **or** when any contiguous `HIGH_ENTROPY_MIN_RUN`-character window does.

Invariant restored: *a run the detector would flag standing alone stays flagged
when concatenated to a neighbour.* The emitted span is still the whole candidate
run, never the hot sub-window.

### Why the window is exactly 20

It is `HIGH_ENTROPY_CANDIDATE`'s own `{20,}` floor — the shortest run the detector
considers a candidate at all — lifted to `patterns.HIGH_ENTROPY_MIN_RUN` with the
regex now built from it, so the number has one source of truth (pinned by
`test_candidate_regex_built_from_min_run_constant`). A narrower window flags
fragments too short to be candidates, and below 14 it cannot fire at all
(`log2(13)` = 3.700 is under the 3.70 default — a width-12 window measures **+0**
new detections repo-wide, i.e. silently disabled). A wider window leaves the
shortest candidates unmeasurable.

### Rejected alternatives, and how they are pinned out

| Variant | Why rejected | Test that fails under it |
|---|---|---|
| Scan only runs >= 40 chars | **0% detection** for a 20-char secret behind a 12–19 char masker — a constructible, total, silent leak | `test_suffix_masked_secret_in_32_to_39_band_redacted` |
| Edge-anchored windows (prefix/suffix only) | misses a secret placed mid-run | `test_wide_mid_masked_secret_redacted` |
| Lower `entropy_threshold` | forbidden threshold change; would over-redact everything | — |

Both rejections were confirmed by running the variants against the suite.

### False positives — measured, not assumed

Sweep of this repository restricted to `RedactionConfig.supported_extensions`
(~9.9 MB; 29,533 candidate runs; 9,825 flagged before):

| ext | runs | flagged before | newly flagged |
|---|---|---|---|
| `.py` | 25,608 | 8,116 | +876 |
| `.md` | 2,455 | 1,028 | +31 (+3.0%) |
| `.txt` | 416 | 131 | +10 |
| `.yaml`/`.toml`/`.json` | 86 | 25 | +1 |
| **total** | **29,533** | **9,825** | **+928 (+9.4%)** |

By `min_confidence`: `0.0` → +0% (all already flagged at the 2.5 floor),
`0.4` → +1.1%, `0.6` (default) → +9.4%, `0.8` → +1.2%, `1.0` → **+0%, provably
inert** (threshold 4.5 > `log2(20)` = 4.3219, short-circuited by the
distinct-character prefilter).

Every newly flagged run contains a 20-char substring the detector **already
flagged when that substring stood alone** — so this is the existing
false-positive rate applied consistently, not a new class of false positive.
The gate's decision was non-monotone under concatenation; this deletes the
inconsistency.

An 18-member `_NEGATIVE_CORPUS` of realistic non-secrets — hyphenated slugs, a
filesystem path, snake_case and CamelCase identifiers, the UUID
`550e8400-e29b-41d4-a716-446655440000`, an ISO timestamp, digit runs, `----`,
`____`, repeated tokens, a two-segment low-entropy run — is asserted to come
back verbatim, and is green both before and after. Each literal carries its
measured (whole-run, best-window) entropy pair in a comment; the test docstring
states they are load-bearing.

Accepting the over-redaction is the same fail-closed trade `redactor.py` already
documents: a missed secret is unrecoverable once written; over-redaction is
visible and fixable via `false_positive_allowlist` or `min_confidence`.

### Cost

Incremental sliding window: a module-level `c*log2(c)` lookup table plus a
running accumulator, with window entropy read as `log2(L) - acc/L`, so each step
is O(1). Two exact prefilters skip only work that could not have returned `True`
— the whole-run fast path, and a `log2(min(distinct, L))` ceiling whose **strict**
comparison leaves an exact tie to be scanned (a uniform window *attains* the
bound).

| measurement | before | after |
|---|---|---|
| entropy gate over the 9.9 MB body | 44.7 ms | 71.9 ms |
| `redact_content` on 1 MB | 1104 ms | 1168 ms (+5.7%) |
| 160 KB single run needing a full slide | — | 3.7 ms (naive re-scan: 181.7 ms, **49x**) |

Complexity class is unchanged: **O(total candidate-run characters)** in file size.
Accumulator drift over 1,000,000 steps measures 4.2e-14 against the direct
formula — eight orders of magnitude below the tightest real threshold margin
(`log2(13)` − 3.7 = 4.4e-4).

### Related issues

- **#946** — `_snap_to_candidate_runs` is deliberately **untouched**; its
  O(spans × runs) term is that issue's PR. This change grows the span count by up
  to ~10%, which accounts for the +5.7% on `redact_content` above. Disclosed, not
  fixed here.
- **#945** — unchanged. Every `PATTERN_METADATA` key is <= 20 chars, and at run
  length exactly `L` the new gate reduces *identically* to the old one, so
  `[REDACTED:email_password_combo]`'s 0.016-bit inertness margin is untouched. A
  new tripwire, `test_pattern_metadata_names_fit_within_min_run`, fails the moment
  a future pattern name exceeds 20 characters and pushes markers into the sub-run
  regime.

### Also in this PR

- `docs/redaction.md`: the `--apply` section claimed snapping was needed because
  "the entropy detector only fires when the whole run clears the threshold". That
  is now false. Rewritten as a layered description — whole-run gate, sub-run
  window gate (#942), token-boundary snapping (#909) as the threshold-independent
  backstop for runs that clear neither — plus the measured over-redaction numbers.
  The same stale claim is corrected in `redactor.py`'s module docstring and in
  `_snap_to_candidate_runs`'s docstring.
- `config.py`: `min_confidence`'s docstring said the `0.6` default "corresponds to
  roughly 4.2 bits/char". It is exactly **3.7** (`2.5 + 0.6 × (4.5 − 2.5)`).
  Docstring only — the default value is unchanged.

## Test plan

- **Gate 1 RED verified at HEAD before implementing:** 14 failures, e.g.
  `AssertionError: assert 'high_entropy_string' in set()` for
  `test_scan_apply_parity_for_masked_run`, and `redact_content` returning
  `'aaaaaaaaaaaaaaaaaaaaaB3xK9mQ7zR2vN5tL8wY'` verbatim.
- **19 new tests** (redaction suites 278 → 297), additions only —
  `git diff --numstat origin/main -- creek-tools/tests/` is `584  0`, **zero
  deletions**; no existing test, fixture literal, or docstring was touched.
  - `TestHighEntropyMaskedRunLeak` (11) — the reproducer on both `--scan` and
    `--apply`; the 32–39 band; mid-run masking; surrounding text; forward parity
    round-tripped through a file on disk; **reverse parity** (a 40-char run that
    clears no window is reported by neither and rewritten by neither);
    `min_confidence=1.0` inertness; allowlist-before-gate ordering; the negative
    corpus.
  - `TestHasHighEntropyRegion` (6) — whole-run fast path, run of length exactly
    `L`, run shorter than `L`, the distinct-character prefilter, the empty-string
    guard, and a **500-string × 5-threshold seeded oracle** comparing the
    incremental accumulator against a naive re-scan. (I separately validated the
    algorithm on 20,000 strings × 5 thresholds with zero mismatches.)
  - `TestHighEntropyRegexSourceOfTruth` (+2) — the regex/constant weld and the
    #945 pattern-name tripwire.
- **Gate 2:** `cd creek-tools && ./scripts/check-all.sh` → **12 passed, 0 failed**;
  6578 tests; **94.00%** aggregate branch coverage; per-file gate green
  (`scanner.py` 98.45%, `redactor.py` 98.69%, `patterns.py` 100%). No `# noqa`,
  no `# type: ignore`, no skips, no threshold changes.
- **Gate 2.5:** independent `code-review-orchestrator` pass over the diff returned
  one blocking finding — the stale single-gate claim in `_snap_to_candidate_runs`'s
  docstring — which is fixed in this commit; `check-all.sh` re-run green after it.

Closes #942


Geoffe-Ga:
## Summary
PR #1010 fixes a real detection gap in the redaction tool: `RedactionScanner._scan_high_entropy` and `Redactor._collect_high_entropy_spans` both gated a `HIGH_ENTROPY_CANDIDATE` run on its *whole-run average* Shannon entropy, so gluing predictable filler next to a genuine secret (e.g. `"a"*20 + secret`) dragged the average below threshold and let the secret pass through both `--scan` and `--apply` in cleartext. The fix introduces one shared `has_high_entropy_region()` helper (a whole-run check plus an O(1)-per-step sliding 20-char window scan) that both call sites now delegate to, restoring the invariant "a run that would be flagged standing alone stays flagged when concatenated to a neighbour."

## Strengths
- Single source of truth: `HIGH_ENTROPY_MIN_RUN` drives both the candidate regex's `{20,}` floor and the window width, pinned together by `test_candidate_regex_built_from_min_run_constant`.
- Scan/apply parity is now structural (one shared decision function) rather than convention-based (previously two hand-synchronised inline comparisons).
- The O(1)-per-step incremental accumulator (`log2(L) - acc/L` with a `c*log2(c)` lookup table) is correct — I independently re-derived the Shannon-entropy identity and verified the sliding-window index arithmetic and table bounds (max window count is exactly `HIGH_ENTROPY_MIN_RUN`, table has `HIGH_ENTROPY_MIN_RUN + 1` entries, so no index errors).
- Two valid short-circuits: a whole-run fast path, and a distinct-character ceiling (`log2(min(distinct, L)) < threshold`) that is a legitimate upper bound since any window's distinct-symbol count cannot exceed the whole run's — with the comparison correctly kept strict so an exact tie still gets scanned.
- Exceptional test rigor: a seeded 500-sample × 5-threshold brute-force oracle comparing the incremental accumulator against a from-scratch re-scan, an 11-case masked-secret suite (including the exact reproducer, a measured 0%-detection band, mid-run masking, forward *and* reverse scan/apply parity, allowlist-before-gate ordering), and an 18-member documented negative corpus guarding against over-redaction regressions.
- Honest, quantified trade-off disclosure: measured false-positive cost (+9.4% flagged repo-wide at default confidence) and performance cost (+5.7% on `redact_content`), with rejected alternatives (>=40-char-only scan, edge-anchored windows) shown to fail specific constructible cases.
- Docstring fix in `config.py` corrects a previously wrong entropy figure (3.7 bits/char, not "roughly 4.2") — verified against `entropy_threshold()`'s actual linear formula.
- Test-only diff to `test_redact.py` is purely additive (584/0 per PR stats), so no existing coverage was touched to make this pass.

## Security Concerns
None — this is a fix that strengthens a defensive secret-detection tool, closing a false-negative window in the redaction pipeline (secrets could previously survive both `--scan` and `--apply` cleartext when abutted by low-entropy filler). No new attack surface is introduced; the change is a pure detection-logic refinement with no new I/O, subprocess, or trust-boundary changes. 🟢 LOW (informational only — no action needed)

## Problems
None found blocking merge.

## Code Quality
- Non-blocking observation: the window scan's fixed-width sliding approach is a "any 20-char window" heuristic, not a proof that *every* possible secret-length/masker-length combination is caught — this is explicitly disclosed in the PR body and is an inherent, reasonable trade-off given the fixed candidate floor, not a gap in this PR's stated scope.
- The `counts` dict in `_window_entropy_clears` never removes zero-count entries as characters leave the window; harmless given the bounded alphabet size relevant here (function-local, short-lived), not worth optimizing.

## CLAUDE.md Adherence
✅ TDD: RED verified before implementation (14 failing tests documented), then GREEN.
✅ `./scripts/check-all.sh` reported green per PR description (12 passed, 0 failed, 6578 tests, 94.00% aggregate coverage, per-file gate green for touched files) — not independently re-run in this review environment (no `uv`/deps available), but the diff shows no shortcuts (no `# noqa`, no `# type: ignore`, no skipped tests, no threshold changes).
✅ No duplicated logic — the fix explicitly replaces two hand-synchronised inline comparisons with one shared function (DRY principle, section 1.2).
✅ Conventional commit-style PR title (`fix(redact): ...`).
✅ Complexity: new functions (`has_high_entropy_region`, `_window_entropy_clears`, `_entropy_from_counts`) are small with straightforward control flow, well under the 10-branch complexity ceiling.
✅ Docstrings present and Google-style throughout new code, consistent with the interrogate ≥95% requirement.

## Testing
Very strong. New tests: `TestHasHighEntropyRegion` (whole-run fast path, exact-length boundary, sub-minimum-length runs, low-alphabet ceiling, empty-string guard, and the seeded oracle differential test), `TestHighEntropyMaskedRunLeak` (the exact reproducer on both scan and apply, the measured 32–39-char 0%-detection band, mid-run and wide-mid-run masking, surrounding-text positional correctness, forward and reverse scan/apply parity, `min_confidence=1.0` mathematical inertness proof, allowlist-before-gate ordering, and an 18-item false-positive negative corpus), plus regex/constant-source-of-truth tests. Every core assertion uses `==` on the full returned string rather than `not in`, explicitly avoiding the weaker-assertion mistake that let a prior related issue (#909) ship. No blocking test gaps identified.

## Documentation
`docs/redaction.md`, `redactor.py`'s module docstring, and `_snap_to_candidate_runs`'s docstring are all updated consistently to describe the new layered gating (whole-run → sub-run window → token-boundary snapping) and include the measured over-redaction numbers. `config.py`'s `min_confidence` docstring is corrected to the exact 3.7 bits/char figure. Documentation accurately reflects the implementation.



## Verdict: LGTM

The core algorithm is mathematically correct (I independently re-derived the O(1) sliding-window entropy update, the ceiling prefilter, and the loop index bounds against the `_COUNT_LOG_TABLE`), scan/apply parity is now enforced by construction via a single shared `has_high_entropy_region` helper, and the change is exhaustively tested (unit tests for the helper, a seeded 500×5 brute-force oracle, and an 11-test masked-secret suite covering the exact leak plus non-regression and reverse-parity cases). No bugs, security issues, or CLAUDE.md violations found; this is ready to merge as-is.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5134549991 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #964, opened 2026-07-29
creek.reflect production grounding has never run: _default_retrieve reads EvidenceClaim.text, which does not exist


Geoffe-Ga:
**Found while widening mypy/bandit/pylint to cover `creek_mcp/` (#925).** Filed separately and suppressed inline there, because fixing it is a privacy-surface *activation*, not a type fix.

## The defect

`creek-tools/creek_mcp/tools/reflect.py:477-490`:

```python
def _default_retrieve(query: str, vault_path: Path, override: object) -> list[str]:
    try:
        from creek.author.agents import RetrievalSpecialist

        bundle = RetrievalSpecialist().gather(query, vault_path, override=override)
        return [claim.text for claim in bundle.claims]
    except Exception:
        return []
```

`EvidenceClaim` (`creek-tools/creek/author/models.py:82-99`) has fields `claim`, `source_fragments`, `author_slug`. There is **no `text` attribute**. So the list comprehension raises `AttributeError` on the first claim, the bare `except Exception` swallows it, and the function returns `[]` — *always*.

`build_server` wires `reflect_tool(...)` **without** a `retrieve=` argument (`creek_mcp/server.py:312-320`), so `_default_retrieve` is the production grounding path. Every production `creek.reflect` call has therefore been ungrounded since the tool shipped, while the module docstring advertises reflections "grounded in retrieval over their corpus".

Only visible under a widened mypy:

```
creek_mcp/tools/reflect.py:488: error: "EvidenceClaim" has no attribute "text"  [attr-defined]
```

Same family as #929 (`_default_llm` vs `default_llm`) — a never-working code path that the static-analysis exclusion let survive.

## Why the one-word fix is not the fix

`claim.text` → `claim.claim` looks obvious. It is not safe as a drive-by, for two reasons.

**1. It activates a title-egress channel.** `_fragment_claim` (`creek/author/agents.py:143-149`) sets `claim=fragment.title`. So the "grounding" this activates is a list of **corpus fragment titles**, which `_build_prompt` then sends to the LLM. That is precisely the channel #962 and #931 document as a hazard class in its own right (`compile` egresses `title:` for every source even when the body is redacted).

[...]

PR #1012, opened 2026-07-30
fix(mcp): route creek.draft through ModelRouter with a source-derived tier

### LINKED ISSUE #958: creek.draft MCP tool bypasses ModelRouter — Intimate-never-cloud not enforced (fails closed only by accident)

Geoffe-Ga:
## Summary

`creek_mcp/server.py::_build_draft_llm` built its LLM with `LLMClassifier(config.llm)`, **bypassing `ModelRouter`** — the single chokepoint that enforces the Intimate-never-cloud invariant (#647). The path was dead only *by accident*: `config.llm` is an `LLMRoutingConfig`, so every production `creek.draft` call raised `AttributeError: 'LLMRoutingConfig' object has no attribute 'provider'`. Reproduced at `origin/main`:

```
type(config.llm) = LLMRoutingConfig   has .provider? False
creek_mcp.server._build_draft_llm() -> AttributeError: 'LLMRoutingConfig' object has no attribute 'provider'
```

PR #967 deliberately left `# type: ignore[arg-type]  # Issue #958` there rather than "correcting" the type, because doing so inline would have stood up a second provider path with no tier gate on it. **This PR therefore does not silence the error — it rewires the tool through the router**, mirroring what #963 did for `creek.compile`. `LLMClassifier` and the suppression disappear as a *consequence*, not as the goal.

### The tier passed to `resolve()`, and why

```python
if not idea.source_fragments:                          # no vault content in this prompt
    return routing_tier(ceiling, None)                 # ceiling-derived tier alone
tiers = source_tiers(vault_path, idea.source_fragments)
return routing_tier(ceiling, max_source_tier(tiers))   # fails closed to INTIMATE when empty
```

`routing_tier` (the helper #963 introduced, reused here rather than re-derived) takes the **more sensitive** of the ceiling-derived tier and the sources' own classifications, so no declaration a caller can make buys cloud routing for sensitive content.

**The ceiling alone is provably insufficient.** `creek/generate/drafts.py::_render_fragment_section` emits `### {fid}: {fragment.title}` *unconditionally*, and at an `open` ceiling `filter_fragments_by_tier` replaces a personal body with `[Personal-tier summary: {title}]` rather than dropping the fragment. Verified against the real payload, not the intent — an `open`-ceiling draft over a `personal` source produces a prompt that contains the personal fragment's **title** and the redaction stub, but not its body. (`intimate` fragments *are* dropped outright below an intimate ceiling, so personal is the case that makes the survey necessary.)

The two empty cases route differently, and the distinction is made **before** the survey runs because `source_tiers` returns `[]` for both:
- `source_fragments == ()` → ceiling alone. Only `UNEXPLORED_ONTOLOGY` produces this, and it is also the only strategy with empty `threads` *and* `eddies` — its title/description are built purely from ontology enum labels. Failing closed here would push every unexplored-ontology draft onto the local model for no privacy gain.
- named but none resolved → `INTIMATE`. With no evidence about what the call carries, the safe assumption is the worst one (mirrors `compile`).

### Shared survey

The fail-closed tier read is extracted to `creek_mcp/source_tiers.py` (`fragment_tier`, `source_tiers`, `max_source_tier`) and shared with `creek.compile`, so two tools can never read the same file's tier two different ways. `_survey_sources` remains a symbol in `creek_mcp.tools.compile` because `read_gate.py` pins it as that tool's `gate_symbol`. `compile`'s uniform-probe-cost analysis moved with the walk it describes. **`creek.compile`'s tests are unmodified and green — that is the parity evidence the extraction is behaviour-preserving.**

## Test plan

`cd creek-tools && ./scripts/check-all.sh` → **12/12 checks pass, 6586 passed, 93.98% branch coverage.**

Gate 1 RED first; the exact pre-fix failures were `TypeError: _build_draft_llm() takes 0 positional arguments but 1 was given` and, through the registered FastMCP handler, `ToolError: Error executing tool creek.draft: 'LLMRoutingConfig' object has no attribute 'provider'`.

Anti-vacuity was the priority — a test asserting "draft refuses intimate content" passes trivially while the path is still dead. Verified independently of the test suite, with a **real `ModelRouter`** and only the provider builder spied:

| case | routing tier | resolved provider |
|---|---|---|
| intimate source, `ceiling=open` | `intimate` | **ollama** (local) |
| *same case pre-fix* (`routing_tier(open, None)`) | `open` | **anthropic** (cloud egress) |
| `privacy_tier` key absent | `intimate` | ollama |
| named id does not resolve | `intimate` | ollama |
| sourceless seed (vault holds an intimate fragment) | `open` | anthropic |
| personal source, `ceiling=open` | `personal` | anthropic |
| intimate + cloud-only config | — | `status="refused"` |

Row 1 vs row 2 is the point: the survey changes behaviour in the leak direction, so it is load-bearing rather than decorative.

Also asserted: `ok` responses echo no derived tier; the audit payload records only the caller's own declared ceiling; the refusal names no fragment id; and neither the empty-seeds path nor the out-of-range-index path ever invokes the factory.

Added `tests/test_mcp_server.py::test_build_server_wires_the_production_draft_llm_factory` — no factory injected, driven through the registered handler. Every prior draft test injected one, which is exactly what let a production factory that raises on its first line go unnoticed.

**Two tests were removed and replaced, called out explicitly so it is not mistaken for test tampering:** `test_build_draft_llm_raises_when_classifier_unavailable` and `test_build_draft_llm_returns_invoke_prompt_when_available` monkeypatched `LLMClassifier` with a stub whose `__init__(self, _config)` accepted anything — that stub *is* what hid #958, and keeping them would have forced the fix to retain `LLMClassifier(config.llm)`. Both properties they asserted are re-asserted against the routed factory by `test_build_draft_llm_routes_by_tier_then_degrades`, which additionally pins tier-keying, the Intimate redirect, and `generation`-stage precedence. All other test changes are a mechanical `llm=` → `llm_factory=lambda tier:` signature migration with **zero assertion changes**.

## Reachability of related issues (please read)

- **#931 — materially more reachable, by design of this fix.** `creek.draft` over MCP was *totally* dead before this PR. Making it live activates latent holes behind it: the `## Threads` / `## Eddies` prompt sections render compiled-page bodies and frontmatter descriptions with **no tier filtering at all**, and they do **not** raise the routing tier. That is stated plainly in `draft.py`'s module docstring rather than left for the next reader, and it is the strongest argument for the conservative fail-closed branches above. Out of scope here; belongs with #931.
- **#962 — not more reachable.** The CLI's untiered `resolve("generation")` is untouched. It does, however, become the *last* unrouted `generation` resolve once this lands.
- **#964 — not more reachable** by this change, but if reflect grounding is ever activated it would feed titles into prompts through the same channel; worth re-triaging together with #931.
- Accepted residual, unchanged from `compile`: an *explicit* `privacy_tier: unclassified` routes cloud-eligible via `_TIER_RANK` (#923) while `filter_fragments_by_tier` treats it as personal for redaction (#876). Not repaired here — a fix belongs in the one ranking both tools read.

Closes #958

### DIFF
--- a/creek-tools/creek_mcp/server.py
+++ b/creek-tools/creek_mcp/server.py
@@ -68,6 +68,7 @@ if TYPE_CHECKING:
     from creek.compile.engine import CompileLLM
     from creek.models import PrivacyTier
     from creek_mcp.tools.compile import CompileLLMFactory
+    from creek_mcp.tools.draft import DraftLLMFactory
     from creek_mcp.tools.reflect import _LLM, _LLMFactory
 
 SERVER_NAME = "creek-tools-mcp"
@@ -155,31 +156,47 @@ def _consumer_from_env() -> str:
     return os.environ.get("CREEK_MCP_CONSUMER", "unknown")
 
 
-def _build_draft_llm() -> Callable[[str], str]:
-    """Return the production LLM callable, mirroring ``creek draft``.
+def _build_draft_llm(tier: PrivacyTier) -> Callable[[str], str]:
+    """Return the draft-side LLM callable, routed for *tier* (#958).
 
-    Imported lazily so an absent LLM provider only fails the ``draft``
-    invocation, not the whole server startup. ``state``/``lint``/``mine``
-    must remain callable on hosts without an Anthropic key or running
-    Ollama.
+    Mirrors :func:`_build_compile_llm`: the ``generation`` stage is resolved
+    through the config's :class:`~creek.classify.llm.router.ModelRouter`, so
+    an INTIMATE draft is forced onto the local ``default`` model — or the
+    router raises ``IntimateRoutingError`` rather than egressing the source
+    ids and titles the draft prompt carries. Nothing here re-checks the tier
+    or picks a provider: ``draft_tool`` derives the routing tier from the
+    seed's source fragments and this hands it to the one chokepoint that owns
+    the decision.
+
+    Lazy imports keep the server bootable with no provider configured — only
+    a ``creek.draft`` call then fails, and it fails as a structured refusal.
+    ``state``/``lint``/``mine`` stay callable on hosts with neither an
+    Anthropic key nor a running Ollama.
+
+    Args:
+        tier: The routing tier ``draft_tool`` computed from the caller's
+            ceiling and the seed's source fragments' own classifications.
+
+    Returns:
+        A prompt → completion-text callable bound to the resolved provider.
+
+    Raises:
+        RuntimeError: When the resolved provider is unavailable, or — as
+            :class:`~creek.classify.llm.router.IntimateRoutingError`, a
+            subclass — when intimate content has no local backend to fall
+            back to. ``draft_tool`` turns either into a refusal.
     """
-    from creek.classify.llm import LLMClassifier
+    from creek.classify.llm.providers import build_provider
 
-    config = load_config()
-    # ``config.llm`` is an ``LLMRoutingConfig``; ``LLMClassifier`` declares a
-    # plain ``LLMConfig``. Correcting the type here means choosing a concrete
-    # per-stage model inline — which is precisely the decision
-    # :class:`~creek.classify.llm.router.ModelRouter` owns, and doing it here
-    # would stand up a second provider path with no Intimate-never-cloud gate
-    # on it. Left suppressed until #958 rewires this tool through the router.
-    classifier = LLMClassifier(config.llm)  # type: ignore[arg-type]  # Issue #958
-    if not classifier.available:
+    cfg = load_config().model_router.resolve("generation", tier)
+    provider = build_provider(cfg)
+    if not provider.available:
         msg = (
-            "LLM provider unavailable; cannot generate draft. "
+            "LLM provider unavailable for draft. "
             "Check Ollama or ANTHROPIC_API_KEY configuration."
         )
         raise RuntimeError(msg)
-    return classifier.invoke_prompt
+    return lambda prompt: provider.complete(prompt).text
 
 
 def _build_compile_llm(tier: PrivacyTier) -> CompileLLM:
@@ -254,7 +271,7 @@ def _build_reflect_llm_factory() -> _LLMFactory:
 def build_server(
     *,
     vault_path: Path | None = None,
-    draft_llm_factory: Callable[[], Callable[[str], str]] | None = None,
+    draft_llm_factory: DraftLLMFactory | None = None,
     compile_llm_factory: CompileLLMFactory | None = None,
     reflect_llm_factory: Callable[[], _LLMFactory] | None = None,
     token_verifier: TokenVerifier | None = None,
@@ -265,9 +282,10 @@ def build_server(
         vault_path: Override vault root. Defaults to
             ``load_config().vault_path`` so the MCP surface honours the
             same configuration as the CLI.
-        draft_llm_factory: Optional factory for the draft LLM. The
-            factory is invoked lazily so only ``creek.draft`` needs an
-            LLM provider.
+        draft_llm_factory: Optional tier-keyed factory for the draft LLM —
+            ``factory(tier)``. Passed straight to ``creek.draft``, which
+            invokes it lazily (and only once it knows which seed it is
+            drafting), so only ``creek.draft`` needs an LLM provider.
         compile_llm_factory: Optional tier-keyed factory for the compile
             LLM — ``factory(tier)``. Passed straight to ``creek.compile``,
             which invokes it lazily (and only after its source-tier gate),
@@ -416,7 +434,7 @@ def build_server(
         """Generate an essay draft from a mined idea."""
         return draft_tool(
             vault_path=vault,
-            llm=factory(),
+            llm_factory=factory,
             privacy_tier_ceiling=privacy_tier_ceiling,
             phase=phase,
             index=index,
--- a/creek-tools/creek_mcp/source_tiers.py
+++ b/creek-tools/creek_mcp/source_tiers.py
@@ -1,0 +1,163 @@
+"""Shared, fail-closed source-tier survey for the MCP tools (#958).
+
+This is the **one place** an MCP tool reads a fragment's classified
+privacy tier in order to route an LLM call, so two tools can never
+disagree about the same file. ``creek.compile`` (#848/#928) and
+``creek.draft`` (#958) both name a set of source fragments, both put
+those fragments' ids and titles into a prompt, and both must therefore
+answer "how sensitive is this call?" the same way. A divergence would
+mean one tool egressing a fragment the other keeps local — which is not
+a cosmetic inconsistency but a leak.
+
+The walk deliberately goes through
+:func:`creek.vault.reader.iter_vault_fragments` — the same loader
+``creek.compile.engine._load_fragments_for_compile`` uses — so the set of
+files this survey inspects and the set the compile engine would actually
+roll up are identical *by construction*. A bespoke ``frontmatter.load``
+scan would diverge: :func:`creek.vault.reader.try_load_fragment` rejects
+files whose ``type`` is not ``fragment`` and files that fail
+:class:`~creek.models.Fragment` schema validation, both of which a raw
+scan happily reads. That divergence would create a class of file one side
+sees and the other does not — precisely the bug class the compile gate
+exists to prevent. A file the shared loader skips (unreadable,
+non-fragment, schema-invalid) is invisible to every caller here, and so
+fails closed to whatever the caller does with an unresolved id.
+
+Nothing in this module decides *admission*, and nothing here refuses: it
+only reports tiers and reduces them. Refusing on them, and choosing what
+an *empty* id list means, stay with the callers
+(:mod:`creek_mcp.tier_ceiling`, :mod:`creek_mcp.tools.compile`,
+:mod:`creek_mcp.tools.draft`).
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from creek.models import PrivacyTier
+from creek.vault.reader import iter_vault_fragments
+from creek_mcp.tier_ceiling import tier_sensitivity
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from pathlib import Path
+
+    from creek.models import Fragment
+
+
+def fragment_tier(fragment: Fragment, raw: dict[str, object]) -> PrivacyTier:
+    """Return *fragment*'s tier, failing closed when the key is absent.
+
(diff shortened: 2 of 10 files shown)


Geoffe-Ga:
## Summary
`creek_mcp/server.py::_build_draft_llm` previously built its LLM client via `LLMClassifier(config.llm)`, bypassing `ModelRouter` entirely — `creek.draft` was dead in production (`AttributeError: 'LLMRoutingConfig' object has no attribute 'provider'` on every real call). This PR rewires `creek.draft` through `ModelRouter`, mirroring the already-shipped `creek.compile` fix (#963), and extracts the shared fail-closed source-tier survey into `creek_mcp/source_tiers.py` so both tools can never disagree about a fragment's tier.

## Strengths
- Root-caused correctly: doesn't just fix the type error, it identifies *why* the suppression existed (`# type: ignore[arg-type] # Issue #958`) and removes the underlying second, ungated provider path instead of patching around it.
- Correctly identifies that the declared `privacy_tier_ceiling` alone is insufficient, because `_render_fragment_section` unconditionally emits `### {fid}: {title}` even for redacted personal fragments — so routing must consider the *sources'* own tiers, not just the caller's ceiling. This is genuinely non-obvious and well-evidenced (`test_draft_prompt_carries_the_personal_title_at_an_open_ceiling` asserts on the real rendered prompt bytes, not just the tier decision).
- Fail-closed defaults are handled correctly and are separately tested: unresolved named source ids → `INTIMATE`; missing `privacy_tier` frontmatter key → `INTIMATE`; the one deliberate exception (`UNEXPLORED_ONTOLOGY`'s genuinely sourceless seed) is justified with a concrete code reference and tested against a vault that holds an unrelated intimate fragment (to prove the seed's own sources are consulted, not the whole vault).
- `IntimateRoutingError` (a `RuntimeError` subclass) is caught and turned into a structured `refused` envelope rather than leaking as a transport error, and the refusal reason is asserted to never name the offending fragment id (`test_draft_refuses_when_the_router_has_no_local_backend`).
- Laziness is preserved and tested: `llm_factory` is invoked only after the two non-drafting exits (empty seeds, out-of-range index), verified by `test_draft_does_not_build_a_provider_when_it_cannot_draft` — avoids spending a cloud credential/handshake on a call that never drafts.
- The extraction to `creek_mcp/source_tiers.py` is real deduplication, not just code movement — `creek.compile`'s existing tests are left unmodified and still pass, which is solid parity evidence for a behavior-preserving refactor.
- The PR body's "Reachability of related issues" section proactively flags what this change does *not* fix (the untiered `## Threads`/`## Eddies` prompt sections, feeding #931) rather than leaving it for a future reader to discover the hard way.

## Security Concerns
🟢 LOW — Two residual gaps are called out explicitly by the author rather than hidden:
- `## Threads` / `## Eddies` prompt sections render compiled bodies/descriptions with no tier filtering and don't raise the routing tier. Documented in `draft.py`'s module docstring as out-of-scope, tied to #931. This is a legitimate scope boundary, not an oversight — the module docstring states it plainly for the next reader.
- The `unclassified`-tier ranking asymmetry between `tier_sensitivity` (ranks cloud-eligible) and `filter_fragments_by_tier` (treats as personal for redaction) is inherited from `compile` and explicitly deferred to a shared fix, not silently duplicated.

Neither is a regression introduced by this PR; both are pre-existing residuals now documented at the new call site instead of only at the old one.

## Problems
None found that block merge.

## Code Quality
- `creek_mcp/source_tiers.py`'s docstrings are quite dense (the "walk never short-circuits" timing-channel argument, for instance) — but the density is justified given this is the shared privacy chokepoint for two tools; skimming it would risk missing why the non-short-circuiting property is load-bearing.
- Minor: `_source_routing_tier`'s docstring repeats some reasoning already stated in the module docstring (empty-sources vs. no-ids-resolved distinction). Not a real issue, since both audiences (module skimmer vs. function reader) benefit from having it inline, but a future edit to one should remember to update the other.

## CLAUDE.md Adherence
✅ TDD workflow evident from PR description (RED-first, exact pre-fix failures quoted) and the two deliberately-removed classifier-stub tests are called out explicitly as tampering-adjacent risk mitigated by a replacement pinning test — matches the "no shortcuts" / "root cause not workaround" principle.
✅ PR body states `./scripts/check-all.sh` → 12/12 checks, 6586 passed, 93.98% branch coverage (I was not able to independently re-run CI in this environment, but the reported numbers meet the ≥90% branch coverage bar).
✅ Complexity kept in check by extracting `_source_routing_tier` out of `draft_tool` specifically to stay within the complexity budget, per the docstring.
✅ Commit message follows Conventional Commits (`fix(mcp): ...`) with a properly attributed co-author trailer.

## Testing
Strong. Coverage includes: tier-routing for personal/intimate/unresolved/sourceless sources, the real rendered-prompt payload check (not just the tier decision), the production-factory wiring test that "would have caught #958" (no `draft_llm_factory` injected, drives the real `ModelRouter`), laziness of the factory invocation, and the `IntimateRoutingError`→refusal path. The anti-vacuity framing in the PR body (a test that trivially passes while the path is dead is worse than no test) is applied consistently — e.g., the miner is stubbed specifically so a bare fixture vault can't make assertions pass vacuously via an early `empty` return.

## Documentation
Docstrings are extensive and specifically justify each non-obvious decision (why empty `source_fragments` routes differently from "named but unresolved," why the derived tier must never be echoed to the caller, why the walk must not short-circuit). This matches the repo's Google-style docstring + ADR-adjacent reasoning convention and exceeds the 95% interrogate bar in spirit.



## Verdict: LGTM

The fix closes a real cloud-egress hole (creek.draft was dead in production and would have leaked source-fragment titles/ids to Anthropic once fixed naively), routes correctly through the existing ModelRouter chokepoint mirroring the proven creek.compile pattern, and ships with exhaustive tier-boundary tests (personal-outranks-open-ceiling, fail-closed-on-unresolved-id, sourceless-seed-by-ceiling-alone, IntimateRoutingError→refusal). No blocking issues found; only a couple of very minor documentation nits that don't warrant holding the merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5135405864 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #1015, opened 2026-07-30
fix(mcp): rank UNCLASSIFIED with PERSONAL on the MCP tier ceiling (#961)

### LINKED ISSUE #961: [scan:bugs] tier_ceiling ranks UNCLASSIFIED as OPEN while privacy_filter ranks it as PERSONAL (#876 straggler)
Found while fixing #922 (PR #960), deliberately out of scope there.

## Evidence

Two reader-side rank tables disagree about `unclassified`:

- `creek-tools/creek/classify/privacy_filter.py:116-126` — `_TIER_RANK[PrivacyTier.UNCLASSIFIED] = 1`, ranking it with `PERSONAL`. The inline comment is explicit that this is **the reader's caution ordering**, raised by #876 because "content nobody has vouched for" ranking alongside `open` exposed whole private corpora at the default ceiling.
- `creek-tools/creek_mcp/tier_ceiling.py:50-55` — `_TIER_RANK[PrivacyTier.UNCLASSIFIED] = 0`, identical to `OPEN`.

Verified at HEAD:

```
tier_allowed(PrivacyTier.UNCLASSIFIED, TierCeiling.OPEN)  -> True
```

`creek_mcp/tier_ceiling.py` is also a reader-side gate — it is the MCP read-tool ceiling, including for **remote** callers. So #876 raised the caution ordering in one of the two reader tables and not the other.

## Impact

A pipeline-written, not-yet-classified fragment carries an explicit `privacy_tier: unclassified` (the `Fragment` model default, serialised on every write). Such a fragment is:

- **excluded** from CLI generation flows at the default ceiling (`privacy_filter.tier_within_override` / `_effective_tier` normalise it to `PERSONAL`), but
- **admitted** to an `open`-ceiling MCP caller via `tier_ceiling.tier_allowed`.

So the same fragment is private to `creek draft`/`mine` and public to a remote MCP reader. On a freshly ingested vault — where every fragment is `unclassified` until `creek classify` runs — that is the entire corpus.

## Why it was not fixed in #922

#922 explicitly scoped this out ("that policy is deliberate and is a separate, larger architectural question"), and the current ranking is pinned by deliberate tests that assert it as policy:

- `creek-tools/tests/test_mcp_tier_ceiling.py:51` — `(PrivacyTier.UNCLASSIFIED, TierCeiling.OPEN, True)`
- `creek-tools/tests/test_mcp_reflect.py:698` — `test_explicitly_unclassified_entry_ref_is_allowed_under_open_ceiling`
- `creek-tools/tests/test_mcp_write_tools.py:1205` — `test_compile_admits_unclassified_source_at_open_ceiling`

Changing the rank means inverting those three tests, which is a policy decision rather than a bug fix — hence this issue rather than a drive-by change.

## Ask

Decide whether the divergence is intentional. If it is not (the `privacy_filter.py:118-125` comment suggests it is a #876 straggler), raise `_TIER_RANK[UNCLASSIFIED]` to `1` in `creek_mcp/tier_ceiling.py` and update the three tests above to assert the new policy. If it *is* intentional, document why the MCP reader is deliberately more permissive than the CLI reader, directly in `tier_ceiling.py`.

## Related
- #922 / PR #960 — stale `privacy_tier` on material rewrite; chose to re-derive the tier rather than reset to `unclassified` precisely because a reset is a no-op for exposure under the current ranking.
- #876 — raised `UNCLASSIFIED` to rank with `PERSONAL` in `privacy_filter.py`.
- #846 — added the MCP read-side ceiling gate this ranking governs.

Geoffe-Ga:
## Summary

`creek_mcp/tier_ceiling.py::_TIER_RANK` ranked `UNCLASSIFIED` at **0** (with `OPEN`); `creek/classify/privacy_filter.py::_TIER_RANK` has ranked it **1** (with `PERSONAL`) since #876. Both are reader-side gates. This raises the MCP table to 1 so the two agree.

`Fragment.privacy_tier` defaults to `unclassified` and `VaultWriter` serialises it on every write, so a pipeline-written, not-yet-classified fragment carries an *explicit* `privacy_tier: unclassified`. At HEAD it was private to `creek draft`/`mine` and readable by an `open`-ceiling MCP consumer — on a freshly-ingested vault, the entire corpus.

**The decisive evidence is that the MCP surface already disagreed with itself.** `creek_mcp/read_gate.py` exports exactly two canonical gate primitives so no tool re-derives the policy — and they gave opposite answers about the same file:

| primitive | delegates to | verdict on explicit `unclassified` at `ceiling=open` |
|---|---|---|
| `refuse_above_ceiling` | `tier_ceiling.tier_allowed` | **admitted** |
| `iter_admitted_fragments` | `privacy_filter.filter_fragments_by_tier` | **summarised as personal** |

`creek.wheel`, `creek.mine`, `creek.draft`'s corpus walk and `creek.author` already excluded it. Only `creek.reflect` and `creek.compile` were out of step — this brings two tools into line, it does not tighten the surface as a whole. `creek_mcp/tools/draft.py` carried the asymmetry as an `ACCEPTED RESIDUAL RISK (#876/#923)` bullet stating a fix *"belongs in the one ranking both tools read rather than in this wrapper"*; that bullet is **deleted**, not reworded.

### `routing_tier` needed a second edit — it is not optional

`routing_tier` reconciles ceiling and content with `max(..., key=tier_sensitivity)`. At rank 0, `unclassified` lost or tied every comparison, so the function **could never return it**. At rank 1 it wins under `ceiling=open`, handing `ModelRouter` a tier its gate has no rule for:

```python
if tier != PrivacyTier.INTIMATE or not provider_is_cloud(cfg.provider):
    return cfg
```

An out-of-vocabulary tier falls through as the **least** restrictive — the same fail-open class this PR closes. A private `_routable_tier` normalises `UNCLASSIFIED` → `PERSONAL` (same rank, expressed in the routing vocabulary), mirroring `privacy_filter._effective_tier`. Provider-neutral today (only `INTIMATE` is gated); pinned exhaustively over every ceiling × tier pair.

### Blast radius — stated, not discovered

**What stops working:** at `ceiling=open`, `creek.reflect` refuses an `entry_ref` resolving to an explicit-`unclassified` fragment; `creek.compile` refuses the whole call if any source is one; `read_gate.refuse_above_ceiling` refuses (no production caller yet); an explicit `tier="unclassified"` write via `creek.save`/`creek.journal` is refused. All are admitted at `ceiling=personal` and above. `creek.ingest` is unaffected (`DEFAULT_INGEST_TIER` is already `personal`).

**How large is the population?** Binary, not marginal:

- **A classified vault: zero.** `creek classify` and `creek process` both call `apply_tier`, and `PrivacyClassifier.classify_tier` never returns `unclassified` (falls back to `PERSONAL`). Pinned by `test_run_classify_never_persists_unclassified_tier` (the #934 invariant).
- **An ingested-but-not-yet-classified vault: 100%.** `creek ingest` run on its own assigns no tier — `creek/ingest/pipeline.py::run_ingest` has zero `privacy_tier` references.

So: **on a vault ingested but not classified, an `open`-ceiling MCP consumer reads nothing until `creek classify` runs.** That is a real availability regression and it is the correct behaviour — failing open on unvouched content is the bug. It is documented in `docs/mcp.md`, in the published Adepthood contract, and here.

`CONTRACT_VERSION` is bumped `0.1.0` → `0.2.0` per `contract.py`'s own instruction ("bump when the tool surface or its semantics change"), with a dated amendment note and change-log row in `docs/decisions/2026-06-30-adepthood-creek-mcp-contract.md`. Nothing pins the literal — `test_handshake.py` and `test_mcp_server.py` both compare against the constant by reference. **Flagging for the operator: this is a deliberate signal to Adepthood to renegotiate. Say so if you would rather it stayed 0.1.0.**

### Pinning the intended divergence

Only the two *reader* tables converge. Two others must not, and now say so out loud:

| table | `UNCLASSIFIED` | question it answers |
|---|---|---|
| `privacy_filter._TIER_RANK` | 1 (with `PERSONAL`) | reader admission — "handle carefully" (#876) |
| `tier_ceiling._TIER_RANK` | 1 (with `PERSONAL`) | **same question at the MCP boundary — changed here** |
| `privacy_pass._ESCALATION_RANK` | **-1** (below `OPEN`) | escalate-only merge — "no claim made", must lose every merge so a real tier can be assigned |
| `author/checks._TIER_RANK` | **3** (above `INTIMATE`) | Writing Desk leak gate — a cited fragment nobody vouched for must always count as over-tier |

New `test_unclassified_ranks_differ_by_context_on_purpose` imports all four and pins each as a **literal and a relation**, with the reason for each in the docstring, so a future "let's have one ranking" refactor fails with the argument attached. `author/checks.py`'s comment claiming its table matched `privacy_filter` was **false and always had been** (3 vs 1) — corrected.

### Every changed assertion, individually

This lane has a narrow exception to the "never change an existing assertion" rule, because these assertions *encode the policy under review*. Each one, with before → after and why:

| # | Location | Before | After | Why the new value is correct |
|---|---|---|---|---|
| 1 | `test_mcp_tier_ceiling.py` `test_tier_allowed_matrix` | `(UNCLASSIFIED, OPEN, True)` | `False` | The row **is** the policy. Untiered content needs a `personal` ceiling. |
| 2 | same, **added** | — | `(UNCLASSIFIED, PERSONAL, True)`, `(UNCLASSIFIED, INTIMATE, True)` | Pins rank 1 from *both* sides. Rank 2 or 3 would also make row 1 false; without these, "refused at open" could pass with the `personal` ceiling silently broken. |
| 3 | `test_tier_sensitivity_ranks` | `(UNCLASSIFIED, 0)` | `(UNCLASSIFIED, 1)` | The rank change itself. |
| 4 | new `test_tier_sensitivity_ranks_unclassified_with_personal` | — | `== tier_sensitivity(PERSONAL)` and `> tier_sensitivity(OPEN)` | Pins the *equivalence*, not the literal — a renumber that preserves ordering stays green, one that re-splits `unclassified` from `personal` does not. |
| 5 | `test_routing_tier_picks_the_more_sensitive_signal` | `(OPEN, UNCLASSIFIED, OPEN)` | `(OPEN, UNCLASSIFIED, PERSONAL)` | `unclassified` now outranks `open`, so it *raises* the routing tier; `_routable_tier` expresses it in the routing vocabulary. |
| 6 | same, **added** | — | `(PERSONAL, UNCLASSIFIED, PERSONAL)` | Pins the equal-rank tie so it cannot resolve toward the content tier and answer `unclassified`. |
| 7 | new `test_routing_tier_answers_only_in_routing_vocabulary` | — | every ceiling × every tier (+`None`) ⇒ result ∈ `{OPEN, PERSONAL, INTIMATE}` | The invariant that fails if the rank is raised *without* `_routable_tier`. |
| 8 | new `test_unclassified_ranks_differ_by_context_on_purpose` | — | all four tables, literals + relations | See above. |
| 9 | `test_mcp_reflect.py` `test_routing_tier_takes_the_more_sensitive_of_entry_and_ceiling` | `(OPEN, UNCLASSIFIED, OPEN)` | `PERSONAL` | Same normalisation through reflect's wrapper. The inline comment said "on a rank tie" — no longer a tie; rewritten, noting the row is unreachable through `reflect_tool` because the #846 gate refuses first. |
| 10 | `test_explicitly_unclassified_entry_ref_is_allowed_under_open_ceiling` | one test, `status == "ok"` at `OPEN` | renamed `..._is_refused_under_open_ceiling`, parametrised: `OPEN` ⇒ refused + reason + **body absent from `json.dumps(result)`** + `factory.asked_tier is None`; `PERSONAL` ⇒ `ok` + body returned | Pure policy pin, inverted. The `PERSONAL` row is load-bearing: it is what still separates an *explicit* `unclassified` (rank 1, admitted at personal) from a *missing* `privacy_tier` key (fail-closed INTIMATE, refused at personal) — the pairing the neighbouring fail-closed test depends on. Net **more** assertions than before. |
| 11 | `test_mcp_write_tools.py` `test_compile_admits_unclassified_source_at_open_ceiling` | one test, `ok` + `factory.calls == 1` at `OPEN` | renamed `..._refuses_...`, parametrised `(OPEN, False)` / `(PERSONAL, True)`; refused ⇒ `_ABOVE_CEILING_REASON` + `factory.calls == 0` | Write-side follows the same rank. Its docstring said *"Issue #923 owns any change to the ranking; if it lands, update `_TIER_RANK` and this test together"* — done, except #923 was a **mis-citation** (it is the separate bare-subscript bug). Re-cited to #961. |
| 12 | `test_mcp_read_gate.py` `test_refuse_above_ceiling_admits_content_within_the_ceiling` | `(UNCLASSIFIED, OPEN)` | `(UNCLASSIFIED, PERSONAL)` | `personal` is now the lowest ceiling that admits it. Docstring said the split with `privacy_filter` was "a deliberate, documented split" — it is now **closed**. |
| 13 | new `test_both_gate_primitives_treat_unclassified_exactly_as_personal` | — | `refuse_above_ceiling` refuses **and** equals the `personal` refusal; `iter_admitted_fragments` yields a body **equal to** a same-titled personal fragment's, raw body absent, `open` body verbatim | The highest-value guard: the two canonical gates must agree. Asserted by comparison against a personal sibling, not a hardcoded `"[Personal-tier summary: …"`. |

**Docstring-only, assertions unchanged** (each stated the old policy as fact): `test_untiered_entry_ref_is_refused_fail_closed`, `test_compile_fails_closed_when_privacy_tier_key_is_absent`, `test_draft_fails_closed_when_a_source_fragment_has_no_privacy_tier`, `test_run_classify_never_persists_unclassified_tier`.

**Test-integrity ledger.** `git diff --numstat origin/main -- creek-tools/tests/` deletions: `test_classify_engine.py` 7, `test_mcp_read_gate.py` 6, `test_mcp_reflect.py` 19, `test_mcp_tier_ceiling.py` 5, `test_mcp_tools.py` 6, `test_mcp_write_tools.py` 19. Filtering those deletions to actual assertions/params/test-defs yields **exactly 11 lines**, all in the enumerated set above; every other deleted line is prose stating the old policy. No unrelated invariant was touched, and rows 2, 4, 6, 7, 8, 10, 13 *add* coverage.

Every `#923` citation removed was citing `_TIER_RANK` as the owner of the old open-ranking, not the actual #923 bug (a bare `_TIER_RANK[tier]` subscript in `privacy_filter.tier_within_override`, still present and untouched).

### Follow-ups observed, not fixed here

- `creek_mcp/tools/ingest.py` writes fragments at the model default `unclassified` while its audit entry and response both report `created_tier: "personal"` — `DEFAULT_INGEST_TIER` gates the ceiling but is never stamped on the fragment.
- `routing_tier` returns an *unrecognised* tier value verbatim (`max` returns the object, not `INTIMATE`); `tier_allowed` refuses it first at `open`/`personal`, so it is latent rather than live.
- `creek classify` walks only `01-Fragments`, so `09-Reference` and `11-Other-Authors` — both in the Writing Desk corpus — are never tiered.

## Test plan

- `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 checks, **6614 passed**, 30 deselected, total coverage **93.98%**, `creek_mcp/tier_ceiling.py` at **100%**, per-file gate 219 files ≥ 80%.
- RED first: 11 failures before the source change, headlined by
  `AssertionError: refuse_above_ceiling admitted an explicitly-unclassified fragment at ceiling=open while iter_admitted_fragments summarises the same fragment as personal; the two canonical gates disagree (#961)`
- Gate 2.5: independent `code-review-orchestrator` pass over the full diff — **CLEAN**, no blockers.
- Full suite re-run after the docs pass: 6638 passed, 5 skipped (skips are Ollama / live-API only).

Closes #961


Geoffe-Ga:
## Summary
PR #1015 closes a fail-open privacy bug: `creek_mcp/tier_ceiling.py`'s `_TIER_RANK` ranked `UNCLASSIFIED` at 0 (with `OPEN`), while `creek/classify/privacy_filter.py` has ranked it at 1 (with `PERSONAL`) since #876. Because `Fragment.privacy_tier` defaults to `unclassified` and every pipeline write serialises it explicitly, this meant an `open`-ceiling MCP caller could read an entire freshly-ingested, not-yet-classified vault. The fix raises the MCP rank to match, adds a `_routable_tier` helper so `routing_tier` never leaks `UNCLASSIFIED` into the router's vocabulary, and updates the contract version, docs, and a large, carefully targeted test suite.

## Strengths
- **Root-cause fix, not a patch**: unifies the two *reader-admission* rank tables (`tier_ceiling`, `privacy_filter`) while explicitly and correctly leaving the two *other* rank tables (`privacy_pass._ESCALATION_RANK`, `author/checks._TIER_RANK`) diverged on purpose — each answers a genuinely different question, and a new test (`test_unclassified_ranks_differ_by_context_on_purpose`) pins all four so a future "let's unify these" refactor fails loudly with the reasoning attached. I checked all four tables against source and the PR's characterization of each is accurate.
- **`routing_tier` correctly re-derived, not just `_TIER_RANK`**: raising the rank could have let `UNCLASSIFIED` win a `max()` comparison and leak an out-of-vocabulary tier into `ModelRouter`, whose gate fails open on unrecognised tiers. The new `_routable_tier` normalization closes this, and `test_routing_tier_answers_only_in_routing_vocabulary` pins the vocabulary invariant exhaustively (every ceiling × every tier + `None`).
- **Exceptional traceability**: the PR body itemizes every changed assertion individually with before/after and justification, and states a "test-integrity ledger" (line-count audit of test deletions) to show no unrelated invariant was weakened. I spot-verified several of these claims directly against the diff and source and found them accurate.
- Blast radius is explicitly measured and stated (zero impact on classified vaults per `test_run_classify_never_persists_unclassified_tier`; full impact on ingested-but-unclassified vaults), rather than hand-waved.
- Contract version bump (`0.1.0` → `0.2.0`) with a dated amendment note and changelog row, per `contract.py`'s own versioning instruction.
- Docstring corrections for prior mis-citations (`#923` → `#961`) and a previously-false comment in `author/checks.py` claiming its table matched `privacy_filter`'s (it didn't — 3 vs 1) are fixed as part of this change.

## Security Concerns
🟢 None — this PR *is* the security fix. It closes a fail-open read-authorization gap at the MCP boundary. I verified the corrected ranking, the `_routable_tier` routing fix, and cross-module consistency directly against source rather than taking the PR description at face value.

## Problems
None found blocking merge.

## Code Quality
- Non-blocking: the inline comment block above `_TIER_RANK` in `tier_ceiling.py` is quite long (lines 65–100) — arguably better suited to a docs page linked from a shorter comment, given how much of it duplicates the docstring content elsewhere in the same file. Not a blocker; the redundancy is intentional given how easy this class of policy is to silently regress.
- The `ACCEPTED RESIDUAL RISK (#876/#923)` bullet deleted from `draft.py` is genuinely resolved by this change (the asymmetry it described no longer exists), so its removal is correct rather than a swept-under-rug.

## CLAUDE.md Adherence
✅ TDD stated in test plan (RED-first failures listed with the exact assertion message).
✅ `./scripts/check-all.sh` reported exit 0, 6614 passed, 93.98% coverage, `tier_ceiling.py` at 100%.
✅ Gate 2.5 (`code-review-orchestrator`) reported clean.
✅ Docstrings updated alongside code (Google-style, Args/Returns sections present in new/changed functions I reviewed).
✅ No use of `# noqa`, `# type: ignore`, or other bypass patterns observed in the diff.
✅ Root-cause fix aligned with "No Shortcuts" principle — the fix lives in the one ranking table, not a wrapper-level patch (explicitly the opposite of what the deleted `ACCEPTED RESIDUAL RISK` bullet had punted on).

## Testing
Testing is extensive and precisely targeted: existing assertions that encoded the *old* (buggy) policy are inverted with clear before/after documentation, new tests pin the cross-module divergence/convergence properties, and a dedicated test (`test_both_gate_primitives_treat_unclassified_exactly_as_personal`) asserts the two canonical MCP gate primitives now agree with each other by comparing actual yielded bodies rather than hardcoded stub strings — avoiding brittleness. I verified the referenced helper functions (`_admitted_by_id`, `_write_fragment`) exist and match usage. This comfortably clears the bar; nothing here is insufficient.

## Documentation
`docs/mcp.md`, `docs/generation.md`, and the Adepthood contract decision doc are all updated consistently with the code change, including an explicit operator-facing amendment note about the availability trade-off (an `open`-ceiling consumer reads nothing from an ingested-but-unclassified vault until `creek classify` runs). Genuinely thorough — the PR even flags the contract version bump as a discussion point for the operator rather than assuming it unilaterally.



## Verdict: LGTM

This is a rigorously reasoned, thoroughly tested fix for a real fail-open privacy leak (MCP tier ceiling admitted explicit `unclassified` fragments — i.e. an entire freshly-ingested vault — at an `open` ceiling). I independently verified the core rank change, the new `_routable_tier` normalization, and every cross-module claim (`privacy_filter._TIER_RANK`, `privacy_pass._ESCALATION_RANK`, `author/checks._TIER_RANK`) against the actual source, and they all match the PR description exactly. Tests are extensive, docs/contract version are updated in lockstep, and nothing blocks merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5136280442 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1031, opened 2026-07-30
creek draft CLI resolves the classification stage with no tier — intimate bodies can egress under --include-tier


Geoffe-Ga:
## Summary

`creek/cli.py:3851` (`_build_draft_llm`) builds the draft LLM as:

```python
classifier = LLMClassifier(config.model_router.resolve("classification"))
```

**No tier.** `ModelRouter.resolve` defaults `tier=None`, and `_enforce_local_for_intimate` documents its own behaviour as *"Non-`Intimate` tiers (and `None`) pass through unchanged"* — so the Intimate-never-cloud chokepoint (#647) is a no-op on this path, exactly as it was on `creek compile` (#962).

Verified against the router directly, on `LLMRoutingConfig(default=ollama, classification=anthropic)`:

```
None         -> anthropic
open         -> anthropic
personal     -> anthropic
unclassified -> anthropic
intimate     -> ollama      <-- the only gated one
```

## Why this one actually egresses bodies, not just titles

Unlike `creek compile` — where `_fragment_excerpt_for_prompt` redacts the body and only the *title* leaks — `creek draft` can put **full intimate bodies** in the prompt:

- The default policy in `creek.classify.privacy_filter.filter_fragments_by_tier` drops `intimate` fragments outright, so the default `creek draft` invocation is safe today.
- But `--include-tier intimate` / `--include-tier all` makes `_allows_intimate(override)` true, and the fragment is yielded **with its full body**.
- `creek.generate.drafts._render_fragment_section` then emits `### {fid}: {fragment.title}` plus that body into the prompt, which goes to the tier-less `classification`-stage provider.

So on a config routing `classification` to a cloud provider, `creek draft --include-tier all` sends intimate fragment bodies to that provider. The `--include-tier` override *is* audited to `00-Creek-Meta/audit/privacy.jsonl`, so the operator's intent is recorded — but the audit records an admission decision, not a routing one, and #647 is meant to be absolute regardless of admission.

Note also that this call site resolves the **`classification`** stage for what is a *generation* workload. That is a separate (pre-existing, non-privacy) oddity worth a look while here, but it is not the bug.

## Relationship to #962 / #958

This is the CLI twin of the MCP `creek.draft` fix (#958), and the sibling of the CLI `creek compile` fix (#962).

[...]

PR #1032, opened 2026-07-30
fix(compile): key the compile LLM with its sources' privacy tier

### LINKED ISSUE #962: creek compile CLI resolves the generation stage with no tier — the Intimate-never-cloud gate is a no-op

Geoffe-Ga:
## Summary

`creek/cli.py:2477` built the compile client as `default_llm(config.model_router.resolve("generation"))` — **with no tier**. `ModelRouter.resolve` defaults `tier=None`, and `_enforce_local_for_intimate` documents its own behaviour as *"Non-`Intimate` tiers (and `None`) pass through unchanged."* So on that path the Intimate-never-cloud rule (#647) had nothing to act on: routing through `ModelRouter` there was structurally decorative.

Reproduced on `LLMRoutingConfig(default=ollama, generation=anthropic)` at `627530b`:

```
None         -> anthropic
open         -> anthropic
personal     -> anthropic
unclassified -> anthropic
intimate     -> ollama      <-- the only gated one
```

**Scope honesty:** the provider outcome is config-dependent. On a vault whose `generation` stage already resolves local, both `tier=None` and `tier=INTIMATE` return the local provider and nothing leaks. The exposure needs a config routing `generation` to a cloud provider — supported and documented, but not every vault. The durable claim is the structural one: a tier-less `resolve()` cannot enforce the tier gate under *any* config.

What egresses is **titles**, and the test asserts against the real payload rather than the provider name: `_build_prompt` emits `id:`/`title:` for every source unconditionally, and `_fragment_excerpt_for_prompt` redacts an intimate *body* only to substitute `[Intimate-tier summary: {title}]` — putting the title straight back in the prompt.

## Approach

Make the invariant **structural rather than documented**. `compile_to_vault` now takes a tier-keyed `llm_factory: Callable[[PrivacyTier], CompileLLM]` instead of a pre-built `llm`, and derives the tier itself from the fragments it already loads. The engine is the only layer that has seen them, so it is the only layer that can say how sensitive the call is; callers can no longer supply a client without a tier.

Two orderings inside `compile_to_vault` are load-bearing and documented as such:

- **The reduction spans every loaded fragment, before the level policy.** `_apply_level_policy` drops non-leaf pairs under the default `"leaves"`, but `_build_prompt` passes a `by_id` spanning *all* pairs to `structural_path_context`, which emits the dropped parents' **titles** as the surviving child's `structural_path:` breadcrumb. Reducing over the policy-selected subset would under-route — an intimate parent's title egressing on a call keyed by its open child. `test_compile_to_vault_routes_on_a_parent_the_level_policy_dropped` pins exactly that case.
- **The factory is invoked after the load and before `_resolve_target_path`.** After the load, so an unresolvable id gets the engine's not-found error rather than whatever the factory would have raised; before the `mkdir`, so a privacy refusal leaves zero state on disk.

`fragment_tier`, `max_source_tier` and `source_tiers` move from `creek_mcp/source_tiers.py` — which they empty, so it is deleted — into `creek/classify/privacy_filter.py`, alongside a new public `tier_sensitivity` reading that module's **existing** `_TIER_RANK` (no new rank table). `creek` may never import `creek_mcp` (pinned at `creek/ingest/journal_staging.py:10` and `creek/care/__init__.py:5`), and the survey had to move whole: it was extracted in #958 precisely so `creek.compile` and `creek.draft` could not come to disagree about the same file, and splitting it would have re-opened that seam from the other direction.

The CLI passes the **content tier alone, with no ceiling term**, unlike `creek_mcp.tools.compile` which reconciles both via `routing_tier`. The CLI operator *is* the vault owner, so there is no caller declaration to defend against — and modelling the CLI as `ceiling=all` would not be conservative but wrong: `CEILING_ROUTING_TIER[ALL] == INTIMATE` would pin every compile of every open fragment to the local model.

## The issue named one site; there were five

I swept every `resolve()` / `resolve_role()` / `provider_for()` call site in `creek/` and `creek_mcp/`, and made a per-site judgement rather than a reflex:

| Site | Decision |
|---|---|
| `cli.py:2477` `creek compile` | **Fixed** — the issue. |
| `cli.py:5518` `_build_compost_verifier` | **Not the same class**, contrary to the issue's framing. Its only caller is `creek compost calibrate`, and the verifier only ever sees `CompostCalibrationEntry` fields from a YAML fixture — it never reads the vault, so no fragment tier exists. Annotated, with a durable caveat: `CompostDetector` accepts a `verifier`, so a future vault-scanning path **must** resolve with a tier. |
| `cli.py:1796` `_run_calibration_cli` | **No tier exists** — a labelled classification fixture, not vault content. Annotated. |
| `cli.py:3950` `_build_ontology_detector` | **No tier exists** — only `OutlineSection.seed_text` from `--seed-outline` / `--seed-outline-text`. Annotated. |
| `cli.py:3851` `_build_draft_llm` | **A real, unfixed leak** — it sends vault fragment *bodies*, and `--include-tier intimate\|all` lets intimate ones through. Different fix shape (sources resolve inside `DraftGenerator`), so filed as **#1031** and pointed at from the site so nobody mistakes it for a considered decision. |

`classify_engine.py:391` and `fidelity.py:117` already pair their tier-less resolve with an explicit `INTIMATE` one (#666); `author/client.py:128` is a diagnostic comparison whose client-building resolve is tier-keyed; `cli.py:2065` resolves only to build a display label.

## A behaviour change worth flagging

Because the factory now runs *inside* the engine after the load, a zero-resolving MCP compile gets the engine's not-found `ValueError` — which names the caller's own requested id — where an all-cloud config previously got the router's id-free refusal.

Judged acceptable: the ordinary local-default config **already** behaved exactly this way, so the change makes the refusal *config-independent* rather than introducing a new class of disclosure; the id echoed is one the caller supplied; and the above-ceiling gate still wins over not-found, which is the ordering that actually prevents existence-probing. `test_compile_not_found_refusal_no_longer_depends_on_model_config` asserts the config-independence directly rather than describing it.

## Test plan

`./scripts/check-all.sh` → **12/12 passed, exit 0**. 6624 tests (baseline 6614), 93.99% branch coverage, per-file gate clean with no new waiver.

**Non-vacuity.** A test asserting "intimate content routes local" passes for free on a config whose `generation` is already local. Every routing test here uses `generation=anthropic` + `default=ollama` and asserts the *unrouted* resolution would have been cloud **before** asserting the routed one is local, with the reason in its docstring.

New coverage:
- **Headline CLI test** — intimate fragment, cloud `generation`: asserts `default_llm` receives the **local** config, and that the prompt kept local really carries both `"Therapy session with Dana"` and `"[Intimate-tier summary: Therapy session with Dana]"`.
- **Engine tier derivation** — intimate-only → `INTIMATE`; open-only → `OPEN`; open-then-intimate → `INTIMATE` (max, not first-seen).
- **Fail-closed** — empty `fragment_ids` → `INTIMATE`; a fragment with **no** `privacy_tier` key → `INTIMATE` (asserting first that the model alone would report `unclassified`, so "read the model" cannot pass).
- **Level-policy trap** — intimate parent dropped by `"leaves"`: routes `INTIMATE`, and the parent's title is shown present in the prompt as a breadcrumb, proving the reduction's breadth is necessary rather than paranoid.
- **All-cloud refusal** — `IntimateRoutingError` propagates; at CLI level exit 1 with the router's wording, no page, **no target directory created**, and a provider spy showing nothing was built.
- **Rank-table agreement** — `privacy_filter.tier_sensitivity` and `tier_ceiling.tier_sensitivity` agree for every `PrivacyTier` plus an unknown one. Their agreement was incidental before and is load-bearing now that MCP *routing* reduces through creek's table while *admission* uses the MCP's. The test explicitly does **not** license merging them; `test_unclassified_ranks_differ_by_context_on_purpose` is untouched.

**Test integrity.** No test was deleted; counts went 44→51, 58→58 (one rename), 14→15. Exactly five assertions were removed repo-wide, each an inversion or a strict strengthening:
- `factory.calls == 1` → `== 0` and `factory.tiers == [INTIMATE]` → `== []`: a zero-resolving request now never reaches a model at all, which is stronger than routing it local. The fail-closed `INTIMATE` default is still live and still tested via the empty-`fragment_ids` path.
- Three inverted in `test_compile_routing_refusal_can_preempt_the_not_found_refusal` → `test_compile_not_found_refusal_no_longer_depends_on_model_config`, whose premise this change closes.

Three `_stub_compile` doubles now invoke `llm_factory` themselves, because `compile_tool` no longer does. Without that, `factory.calls` would have quietly become a dead signal — zero for admitted *and* refused rows — making six `assert factory.calls == 1` vacuous rather than failing loudly. Their six assertions are byte-identical.

## Residuals, unchanged and still documented

- The MCP **admission**-side TOCTOU: this closes the *routing* half only (the engine derives its tier from bytes it holds); the gate still walks the vault before the engine does.
- `structural_path` **ancestors outside the compile set** contribute titles from their persisted field without their tiers being surveyed — the pre-existing #928-family residual.
- `creek draft` CLI — **#1031**.

Closes #962
Refs #647, #928, #958, #1031


Geoffe-Ga:
## Summary
Fixes #962: `creek compile`'s CLI path built its LLM client via `config.model_router.resolve("generation")` with no tier argument, which makes `ModelRouter._enforce_local_for_intimate` a no-op — so the Intimate-never-cloud invariant (#647) was structurally unenforced on that path even though it was enforced on the MCP `compile` tool. The fix inverts control: `compile_to_vault` now takes a tier-keyed `llm_factory: Callable[[PrivacyTier], CompileLLM]` and derives the routing tier itself from the fragments it loads, so no caller can hand it a tier-less client. `fragment_tier`/`max_source_tier`/`source_tiers` move from `creek_mcp.source_tiers` (now deleted) into `creek.classify.privacy_filter` so the engine can compute this without importing `creek_mcp` (a documented layering rule).

## Strengths
- The core claim was verified directly against `creek/classify/llm/router.py`: `resolve(stage, tier=None)` really does pass an unmodified `LLMConfig` through when `tier` is `None`, confirming the described bug and its fix.
- `_routing_tier_for` correctly reduces over **every loaded fragment before the level policy** — verified this matters because `_build_prompt` passes a full `by_id` lookup to `structural_path_context`, which surfaces a dropped parent's title as the child's breadcrumb even under the default `"leaves"` policy. `test_compile_to_vault_routes_on_a_parent_the_level_policy_dropped` pins exactly this.
- Fail-closed defaults are correct and tested: empty `fragment_ids` → `INTIMATE`, a fragment missing the `privacy_tier` key entirely → `INTIMATE` (distinguished from explicit `unclassified` via raw frontmatter, not the validated model).
- Ordering is right and tested: `llm_factory` is invoked after the fragment load (so a bad id gets "not found," not a routing error) and before `_resolve_target_path`'s `mkdir` (so a privacy refusal leaves no directory behind).
- The deleted `creek_mcp/source_tiers.py` module has zero remaining live references — checked with a repo-wide grep; only historical docstring mentions remain.
- `creek_mcp/tools/compile.py`'s admission gate (`_survey_sources`/`gate.above_ceiling`) and its ordering relative to `compile_to_vault` are untouched; only the now-unused `max_tier` field was removed from `_SourceGate`, and nothing else references it.
- Tests include real anti-vacuity discipline: routing tests assert the *unrouted* resolution would have been cloud before asserting the routed one is local, so they can't pass "for free" on an already-local config. `test_privacy_filter_and_mcp_tier_sensitivity_agree_on_every_tier` pins that the two independent rank tables agree (and pins the absolute ranks, not just their equality).
- The five removed/inverted assertions are all justified in the PR body as strengthenings, and this matches what's in the diff (e.g., `factory.calls == 1 → == 0` for the not-found path, since the engine now never builds a client for an unresolvable id).
- The four CLI call sites in `cli.py` that deliberately do *not* thread a tier (calibration, ontology detector, compost verifier) get comments justifying why, distinguishing them from the real residual leak at `_build_draft_llm` (tracked separately as #1031) — this is good scope discipline rather than silently leaving look-alike sites unaddressed.

## Security Concerns
🟢 LOW — This PR *closes* a real privacy-routing gap rather than introducing one. Two residuals are explicitly named as unchanged/out-of-scope in the PR body and are reasonable to defer: the MCP admission-side TOCTOU (walk-then-load race) and `creek draft`'s CLI (#1031, a different fix shape since the tier is per-fragment there). Both are pre-existing and documented, not introduced by this change.

## Problems
None found blocking merge.

## Code Quality
- Non-blocking: the docstrings in this PR are extremely long and dense with cross-references and rationale (e.g., `tier_sensitivity`, `source_tiers`, `_routing_tier_for` each run 20–40 lines). This matches the repo's ≥95% docstring-coverage requirement and its stated "MAXIMUM QUALITY" documentation culture, but it does raise the maintenance cost of keeping every comment in sync with future edits — a future change to `_apply_level_policy` or `structural_path_context`, for instance, would need to update several docstrings that currently cite exact behavior. Worth being mindful of going forward, not a reason to hold this PR.
- The lambda-wrapping pattern in `creek_mcp/tools/compile.py` (`llm_factory=lambda content_tier: llm_factory(routing_tier(...))`) shadows the outer `llm_factory` parameter name with the closure variable of the same name — legal and correctly scoped here since it's invoked synchronously inside the same call, but a name that's easy to misread on a skim.

## CLAUDE.md Adherence
✅ Single conventional-commit message (`fix(compile): ...`). ✅ No `noqa`/`type: ignore`/threshold changes anywhere in the diff. ✅ `compile_to_vault`'s `llm` → `llm_factory` signature break has no compatibility shim — all in-repo callers were updated in the same commit, consistent with the repo's anti-shim stance. ✅ PR description states `./scripts/check-all.sh` passed 12/12 locally with 93.99% branch coverage (≥90% threshold) and no new per-file waiver; I did not re-run it myself in this environment, but nothing in the diff (unused imports, obviously broken types, dangling references) contradicts that claim under manual inspection.

## Testing
Strong. New coverage spans: the flagship CLI regression (intimate fragment + cloud `generation` config → local provider, with the actual prompt payload asserted rather than just the provider name), tier derivation (single-source, mixed-order max, empty), the level-policy breadcrumb trap, the all-cloud refusal path at both the engine and CLI layers (asserting zero directory/state creation and zero provider construction via a spy), and cross-table rank agreement. The three MCP-side stub doubles were correctly updated to call `llm_factory` themselves so `factory.calls` doesn't silently become a dead signal — verified this reasoning is sound given the engine, not the wrapper, now owns the call.

## Documentation
Docstrings are exhaustive to the point of being a style outlier for this codebase, but that's consistent with (and exceeds) the ≥95% interrogate requirement. The PR body itself is unusually thorough — it names the exact repro, the two load-bearing orderings, the four now-unified call sites, and the residuals left out of scope — which made this review straightforward to verify against the actual diff.



## Verdict: LGTM

The fix is structurally correct (verified directly against ModelRouter.resolve/_enforce_local_for_intimate and the title-egress path in _build_prompt/_fragment_excerpt_for_prompt), every caller of the changed compile_to_vault signature and the deleted creek_mcp/source_tiers module was updated with no orphaned references, and the new tests include genuine anti-vacuity assertions (unrouted-resolution-would-be-cloud checks, the level-policy-drop trap, ordering/no-disk-state-on-refusal). Nothing here blocks merge; the only remark is that the docstrings are unusually long, which is a style preference, not a defect.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5137404793 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1036, opened 2026-07-31
read_gate layer (f) cannot see a prompt-side leak: the canary probe asserts on the response, and refuses before the second corpus read


Geoffe-Ga:
**Found during the Gate 2.5 review of #964.** Not introduced there; #964 is what makes the gap reachable for `creek.reflect`.

## The gap

Layer (f) of `tests/test_mcp_read_gate.py` calls every `GATED` tool for real and asserts a canary appears nowhere in the **JSON-serialised response**. Its own preamble states why it exists:

> None of that can prove the gate is the *only* path to the corpus: a second, ungated read added alongside the still-present, still-called gate satisfies every structural check. Only calling the tool and looking at what comes back can see that, which is what (f) does.

For a tool that hands corpus content to an LLM, that is only half the surface. Two blind spots, and `creek.reflect` has both:

**1. The probe refuses before the second read runs.** `_probe_reflect` calls `reflect_tool` with `entry_ref=_RUNTIME_INTIMATE_ID` at `ceiling=open`. The #846 gate refuses it before the care seam, the grounder and the model — which is exactly what that probe is *for*. But it means the probe never exercises the grounding corpus walk at all. A raw-`content=` call is never ceiling-gated as an entry, proceeds all the way to the model, and is the only shape that reaches the second read.

**2. A prompt leak is invisible to a response assertion.** Grounding egresses to the **provider**, not to the caller. A tool that folded above-ceiling fragment text into its prompt while returning clean notes would leak to a third party and pass layer (f) unchanged. `tests/test_mcp_reflect.py`'s `_RecordingFactory` already exists precisely because of this, and says so:

> ``prompt`` captures what actually crossed to the model, which the response alone cannot show: a regression that folded above-ceiling fragment text into the prompt while leaving the returned notes clean would egress that text to the provider and still look correct from the outside.

So the strongest assertion available for LLM-backed tools lives in one tool's own suite, and the manifest layer that is supposed to be the forcing function does not require it.

## Why this is not urgent

#964 added targeted per-tier coverage of exactly this path for reflect — `test_grounding_admits_exactly_the_within_ceiling_corpus`, `test_above_ceiling_fragment_contributes_nothing_not_even_a_title` and `test_routing_tier_dominates_every_retrieved_fragment_tier` all assert against `_RecordingFactory.prompt`. Both mutants tried during that lane were killed:

- disabling `tier_within_override`'s ceiling cutoff -> 5 failures;
[...]

Issue #1034, opened 2026-07-31
creek.reflect grounding walks and re-embeds the whole corpus on every call; a cold embeddings cache is unbounded latency


Geoffe-Ga:
**Found while activating `creek.reflect`'s production grounding (#964).** The latency is inherent to `RetrievalSpecialist`, not to reflect; #964 is what makes an *interactive* MCP tool pay it.

## The cost

`_default_retrieve` (`creek_mcp/tools/reflect.py`) constructs a fresh `RetrievalSpecialist()` per call. `gather` then:

1. `_load_corpus` — parses **every** markdown file under `01-Fragments`, `09-Reference`, `11-Other-Authors`;
2. `_get_cache` — loads the persisted embeddings parquet (`embeddings_cache_path(vault)`) from scratch, because the instance is new every call;
3. `_rank_fragments` — embeds the query, then for **every admitted fragment** either reuses a cached vector whose `content_hash` still matches or **live-embeds it** (`_fragment_vector`);
4. keeps `config.author.retrieval_top_k` (default 5) and throws the rest away.

On a vault that has never run `creek link --method embeddings`, step 3 live-embeds the entire admitted corpus on **every** `creek.reflect` call. At the known 35k-fragment scale that is minutes of latency inside an interactive tool. Even with a warm cache, steps 1 and 2 are a full corpus parse plus a parquet load per call.

`_default_retrieve`'s `except Exception: return []` does **not** bound this — slowness is not an exception, so the tool has no degradation path for it. It will simply hang.

## Why the obvious fix is not obvious

A module-level `RetrievalSpecialist` singleton in `reflect.py` would amortise the model load and the parquet read. But `RetrievalSpecialist`'s own docstring states the linker is deliberately instance-level and "never a module global", so a fresh specialist per test starts cold and the conftest sentence-transformer mock cannot leak across tests. Any caching has to respect that, or move the seam.

## Scope

1. Measure a `creek.reflect` call against a realistically-sized vault, cold cache and warm.
2. Decide the bound: cache the specialist on the server (`build_server` already owns per-server state and could inject `retrieve=`, which is an existing seam), cap the corpus, or require a warm embeddings cache and refuse-with-reason otherwise.
3. Whatever is chosen, `_default_retrieve`'s docstring must state the cache dependency explicitly — #964 adds a caveat, not a bound.

Refs #964, #660

Geoffe-Ga:
[...]

Issue #1033, opened 2026-07-31
_load_corpus fails open on a missing privacy_tier key, disagreeing with the two fail-closed readers of the same file


Geoffe-Ga:
**Found while activating `creek.reflect`'s production grounding (#964).** Pre-existing since #660; not introduced there. Filed rather than fixed because the fix changes four other tools.

## The asymmetry

Two functions read the *same* fragment file and disagree about how sensitive it is when the `privacy_tier` key is **absent entirely**:

| reader | missing key becomes | admitted at |
|---|---|---|
| `creek_mcp/tools/reflect.py::_fragment_tier` | `INTIMATE` (fail closed) | `ceiling=intimate` and above |
| `creek/classify/privacy_filter.py::fragment_tier` | `INTIMATE` (fail closed) | — |
| `creek/author/agents.py::_load_corpus` | `UNCLASSIFIED` (rank 1) | `ceiling=personal` and above |

`_load_corpus` filters on `fragment.privacy_tier` read off the **validated `Fragment` model**, whose default for a missing key is `unclassified`. `tier_within_override` ranks `unclassified` with `personal` (#876), so the fragment is admitted from `ceiling=personal` up. The raw front matter that would distinguish "no key" from an explicit `privacy_tier: unclassified` is available in the same loop — `iter_vault_fragments` yields it as `_meta` — and is discarded (`creek/author/agents.py:114`).

## Reproduction

A fragment with **no** `privacy_tier` key at all:

```
---
type: fragment
id: frag-nokey
title: "T-NOKEY"
source: {platform: journal, author: self}
---
B-NOKEY
```

```python
>>> _fragment_tier({"id": "frag-nokey"})            # creek_mcp.tools.reflect
<PrivacyTier.INTIMATE: 'intimate'>

>>> for c in TierCeiling:
...     RetrievalSpecialist().gather("q", vault, override=to_privacy_override(c))
open      -> []
personal  -> ['T-NOKEY']      # <-- admitted
intimate  -> ['T-NOKEY']
all       -> ['T-NOKEY']
```

## Why it matters now

Before #964, `creek.reflect`'s `_default_retrieve` always returned `[]`, so the corpus walk never ran there. With grounding activated, a single tool now contains **both** readings at once: the same untiered, hand-edited fragment is

- **refused as an `entry_ref`** at `ceiling=personal` (`_fragment_tier` -> INTIMATE -> `_above_ceiling`), yet
[...]

PR #1037, opened 2026-07-31
fix(mcp): activate creek.reflect's grounding on ceiling-bounded titles (#964)

### LINKED ISSUE #964: creek.reflect production grounding has never run: _default_retrieve reads EvidenceClaim.text, which does not exist

Geoffe-Ga:
## Summary

`creek_mcp/tools/reflect.py::_default_retrieve` read `EvidenceClaim.text`. The model carries `claim`, not `text`, so the comprehension raised `AttributeError` into a bare `except Exception: return []`. `build_server` wires `reflect_tool` with no `retrieve=`, so **every production `creek.reflect` call has been ungrounded since the tool shipped**, while the module docstring advertised reflections "grounded in retrieval over their corpus". Third never-worked path found in `creek_mcp/` after #929 (compile) and #958 (draft), all three surfaced by widening mypy in #925.

Verified before touching anything — `_default_retrieve` returns `[]` at every ceiling:

```
open -> []   personal -> []   intimate -> []   all -> []
```

### The design decision: titles, not bodies or excerpts

The one-character fix makes it *run*; it does not settle what "grounding" should mean here. It should be **titles**, and that is a decision rather than a default-to-whatever-the-code-does:

1. **Grounding is structurally unquotable.** `_clean_notes` validates every returned `notes[].quote` as a verbatim span of the **entry**, never of a source. Grounding's only job is to tell the model which themes recur in this person's corpus. Titles do that; bodies would add egress without adding a quotable surface.
2. **Bodies would mean changing `_fragment_claim`**, whose "one assertion, one short sentence" contract is shared by `GraphSpecialist` and the Writing Desk's citation and leak-gate machinery. A body excerpt is not an assertion. That is a Writing-Desk-wide semantic change wearing a reflect-shaped costume.
3. **`retrieve=` is an injectable seam.** The *default* is where the conservative choice belongs; a deployment wanting richer grounding supplies its own callable (which replaces the default outright — the two are never merged).

### This is not the #931 / #1032 title-egress hazard

That was the sharpest question in the brief, and the answer is structural rather than lucky. #931/#1032 concern a fragment whose body policy already **dropped or summarised** still contributing its title — the residue of `filter_fragments_by_tier`'s summarise-the-body path (`[Intimate-tier summary: {title}]`, `structural_path`). Grounding retrieval never goes near that. `_load_corpus` applies `tier_within_override`, a **hard rank cutoff** that drops the record before ranking, so an above-ceiling fragment contributes **nothing**: no title, no body, no id. There is no summarise step for a title to survive.

`test_above_ceiling_fragment_contributes_nothing_not_even_a_title` pins that by construction, so a future "reuse the compile filter here for consistency" refactor — which would look like a tidy-up and would import the residue wholesale — fails loudly.

### The activated path is tier-routed, and here is the proof

The brief asked whether activation puts content through a resolve that is not tier-keyed. It does not, and the invariant is now pinned rather than argued. Measured on this code, one fragment per tier:

| ceiling | titles reaching the prompt | routed tier |
|---|---|---|
| `open` | open | `OPEN` |
| `personal` | open, personal, unclassified | `PERSONAL` |
| `intimate` | open, personal, unclassified, intimate | `INTIMATE` |
| `all` | open, personal, unclassified, intimate | `INTIMATE` |

Retrieval is bounded by `to_privacy_override(ceiling)`; `routing_tier`'s ceiling-derived **floor** is the most sensitive tier that same ceiling admits. So the routing tier always dominates every retrieved fragment, and **intimate titles are only reachable under a ceiling that already forces INTIMATE local-only routing**. `test_routing_tier_dominates_every_retrieved_fragment_tier` asserts both the table (so a `CEILING_ROUTING_TIER` retune fails here, where the grounding consequence lives) and the inequality *derived from the tiers whose titles actually reached the prompt* — so it survives a new tier being added, and it guards against the two halves drifting apart.

### The silent swallow

`except Exception: return []` is what hid this for the feature's whole life. It now logs the exception's **type name only** — never `str(exc)`, `exc_info`, or `logger.exception`, per the `_resolve_entry` precedent at `reflect.py:326-331`: `yaml.MarkedYAMLError` stringifies *with the offending source snippet*, which would write tier-unknown vault content into a log that carries no tier and is not covered by the ceiling. The catch stays broad on purpose — `gather` reaches `load_config`, a full corpus parse and the sentence-transformer load.

The `# type: ignore[attr-defined]  # Issue #964` suppression is removed as a **consequence** of the fix, not to satisfy `warn_unused_ignores`.

### Read-posture manifest (#973)

The guardrail did **not** fire, and the brief's expectation that it might is worth correcting: `_probe_reflect` calls reflect with an above-ceiling `entry_ref`, which refuses before grounding runs — and a grounding leak egresses to the **provider**, not into the response layer (f) inspects. So layer (f) is structurally unable to see this path. I filed that as **#1036** rather than papering over it.

What did change is the manifest **rationale**, deliberately: `creek.reflect` now has *two* unsupplied reads under *two* gates. `posture`, `gate_module` and `gate_symbol` are byte-identical and still true.

## Test plan

- `./scripts/check-all.sh` -> exit 0. **6637 passed** (baseline 6624), **94.01%** branch coverage (baseline 93.99%), 12/12 checks. Credentials stripped from the environment.
- **Gate 1 RED verified before the fix**: 11 failing invocations, every one on an empty `(none)` grounding block or the missing degradation log — `assert '(none)' != '(none)'` — never on a fixture or import error.
- **13 new tests**, none of which pass `retrieve=` (except the one whose subject *is* the seam). All assert on `_RecordingFactory.prompt`, not the response: a grounding leak egresses to the provider and leaves the returned notes clean.
- **Mutation-tested, both mutants killed:**
  - disable `tier_within_override`'s ceiling cutoff -> **5 failures**, including the derived routing inequality;
  - leak `fragment.id` into `_fragment_claim`'s rendering -> **1 failure**, caught by the exact-equality assertion on the rendered grounding line.
- **Test-integrity**: `git diff --numstat origin/main HEAD -- creek-tools/tests/` is `566 0` — **zero deletions**. `tests/test_mcp_reflect.py` sha256 `9c06cba46b9982ad9ee37616a31388f00e481ea3893622968957469429e90404`, unchanged from the RED commit through Gate 2.5.
- No `# noqa`, no `# type: ignore`, no skips, no threshold changes anywhere in the diff.

### The trap that would have made this whole suite vacuous

`tests/test_mcp_reflect.py`'s existing `_write_fragment` writes only `id` + `privacy_tier` and **no `type: fragment`**, so `try_load_fragment` returns `None` and the file is **invisible** to `_load_corpus`. Reusing it for corpus fixtures would have run every new assertion against an empty corpus and passed vacuously — the exact false green that let #964 survive. `_write_corpus_fragment` is a separate, model-valid writer whose docstring says why it must never be merged with its lookalike.

Every "nothing leaked" assertion carries a positive control (the `open` fixture must be present), so none of them can pass by the corpus being empty.

## Filed, not fixed

Three findings, deliberately out of scope because each has a blast radius beyond this issue:

- **#1033** — `_load_corpus` fails **open** on a *missing* `privacy_tier` key (the `Fragment` model default is `unclassified`), while `reflect._fragment_tier` and `privacy_filter.fragment_tier` both fail **closed** to INTIMATE on the same file. Reproduced. Activation puts both readings inside one tool: the same untiered fragment is refused as an `entry_ref` at `ceiling=personal` yet contributes its title as grounding there. Pre-existing since #660; the fix changes `creek.author`, `creek.mine`, `creek.draft` and `creek.wheel` too.
- **#1034** — grounding constructs a fresh `RetrievalSpecialist` per call, so it re-parses the corpus and re-loads the embeddings parquet every time, and live-embeds anything whose cached `content_hash` is stale. On a cold cache that is the whole admitted corpus. The `except Exception` does not bound it — slowness is not an exception. Documented honestly in the docstring.
- **#1036** — read_gate layer (f) asserts on the *response*, so it cannot see a prompt-side leak, and `_probe_reflect` refuses before the second corpus read. `creek.compile` is in the same position.

Also noted by review and *not* filed, because it is unreachable by construction: `tier_within_override` uses a bare `_TIER_RANK[tier]` lookup rather than the fail-closed `.get(..., INTIMATE)` its sibling `tier_sensitivity` uses. `PrivacyTier` has exactly the four members the table covers, and if it ever did raise, `_default_retrieve`'s broad catch degrades grounding to `[]` — it fails safe. Worth mentioning only because this PR is what makes that lookup execute in production for the first time.

Closes #964
Refs #925, #929, #931, #958, #962, #973, #1032

### DIFF
--- a/creek-tools/creek_mcp/read_gate.py
+++ b/creek-tools/creek_mcp/read_gate.py
@@ -167,9 +167,13 @@ TOOL_POSTURES: dict[str, ToolPosture] = {
     "creek.reflect": ToolPosture(
         posture=ReadPosture.GATED,
         rationale=(
-            "An entry_ref resolves a fragment the caller did not supply; "
-            "_above_ceiling refuses it before the care seam, the grounding "
-            "retrieval, or the model (#846)."
+            "Two unsupplied reads, under two gates. (1) An entry_ref resolves "
+            "a fragment the caller did not supply; _above_ceiling refuses it "
+            "before the care seam, the grounding retrieval, or the model "
+            "(#846). (2) The grounding corpus walk, live since #964, is gated "
+            "by to_privacy_override -> tier_within_override's hard rank cutoff "
+            "inside RetrievalSpecialist: only within-ceiling fragments are "
+            "admitted, and they contribute their titles only."
         ),
         gate_module="creek_mcp.tools.reflect",
         gate_symbol="_above_ceiling",
--- a/creek-tools/creek_mcp/tools/reflect.py
+++ b/creek-tools/creek_mcp/tools/reflect.py
@@ -3,9 +3,12 @@
 Takes one journal entry (raw ``content`` or an ``entry_ref`` fragment id) plus a
 privacy-tier ceiling and returns ``{notes: [{quote, kind, note}], essay?}`` —
 warm, second-person, anti-guru reflections that mirror the user's own wisdom,
-grounded in retrieval over their corpus. Distinct from the essay-shaped
-``draft``/``author``/``mine`` surface; this is the reflection surface Adepthood's
-journal needs.
+grounded in the *titles* of the ceiling-admitted corpus fragments nearest the
+entry (titles only, never body text). That grounding has been live only since
+#964: before it, the default grounder raised into a silent ``except`` and every
+reflection was ungrounded. Distinct from the essay-shaped
+``draft``/``author``/``mine`` surface; this is the reflection surface
+Adepthood's journal needs.
 
 Four guarantees this module enforces:
 
@@ -46,8 +49,10 @@ structured dict rather than raising: a malformed or unreadable file under
 malformed model turn degrades to no notes. No parse error crosses the MCP
 boundary.
 
-The LLM and retrieval are injected so the tool is unit-testable with no live
-calls; ``build_server`` supplies the production factory + retrieval.
+The LLM and retrieval are both injectable seams so the tool is unit-testable
+with no live calls. ``build_server`` supplies the production LLM factory and
+leaves ``retrieve`` unset, so production grounding is :func:`_default_retrieve`
+— the ceiling-bounded title retrieval described above.
 """
 
 from __future__ import annotations
@@ -358,8 +363,12 @@ def reflect_tool(
             *content* wins when both are given.
         entry_ref: A fragment id whose body is the entry, when *content* is
             absent.
-        retrieve: ``(query, vault, override) -> snippets`` grounding source;
-            defaults to the corpus retrieval specialist.
+        retrieve: ``(query, vault, override) -> grounding lines`` source. The
+            default (:func:`_default_retrieve`) returns ceiling-bounded corpus
+            fragment **titles**, never body text. An injected callable
+            *replaces* the default outright — the two are never merged — so a
+            deployment wanting richer grounding owns its own admission
+            filtering.
         care_guard: ``(entry) -> reason | None``; a non-``None`` reason escalates
             to a human and skips the model entirely (#753 seam).
         privacy_tier_ceiling: The ceiling; gates admission of the *entry itself*
@@ -485,25 +494,91 @@ def reflect_tool(
 def _default_retrieve(
     query: str, vault_path: Path, override: PrivacyTierOverride
 ) -> list[str]:
-    """Production grounding: top corpus fragments related to *query*.
+    """Production grounding: the *titles* of the corpus fragments nearest *query*.
 
-    Lazily imports the author retrieval specialist so the server still boots
-    when its (heavier) deps are unavailable; any retrieval failure degrades to
-    no grounding rather than failing the reflection.
+    Returns the titles of the top ``author.retrieval_top_k`` (default 5) corpus
+    fragments semantically nearest *query*, drawn from ``01-Fragments``,
+    ``09-Reference`` and ``11-Other-Authors``. The author retrieval specialist
+    is imported lazily so the server still boots when its (heavier) deps are
+    unavailable.
+
+    **Titles, not bodies — and that is the decision, not an accident.** Three
+    reasons, recorded because "grounding" reads like it ought to mean body text:
+
+    - :func:`_build_prompt` validates every returned ``notes[].quote`` as a
+      verbatim span of the *entry*, never of a source, so grounding supplies
+      thematic context and is structurally unquotable — titles serve that role.
+    - Bodies would mean changing ``creek.author.agents._fragment_claim``, whose
+      "one assertion, one short sentence" contract is shared by
+      ``GraphSpecialist`` and the Writing Desk's citation and leak-gate
+      machinery.
+    - :func:`reflect_tool`'s ``retrieve=`` is an injectable seam, so the
+      *default* is deliberately the conservative one; a deployment wanting
+      richer grounding supplies its own callable.
+
+    **The ceiling invariant.** ``creek.author.agents._load_corpus`` applies
+    ``tier_within_override`` — a **hard rank cutoff** — so an above-ceiling
+    fragment contributes *nothing*: no title, no body, no id. This is
+    explicitly NOT ``filter_fragments_by_tier``'s summarise-the-body path,
+    which is the shape #931/#1032 found leaking titles for fragments whose
+    bodies policy had already dropped. Verified admitted sets: ``open`` admits
+    ``open`` only; ``personal`` admits ``open`` + ``personal`` +
+    ``unclassified``; ``intimate`` and ``all`` admit all four.
+
+    **The routing consequence.** The most sensitive tier that can reach the
+    prompt is therefore bounded by the ceiling, and :func:`reflect_tool` keys
+    the model with :func:`creek_mcp.tier_ceiling.routing_tier`, whose
+    ceiling-derived floor is the most sensitive tier that ceiling admits. So
+    the routing tier always dominates every retrieved fragment's tier, and
+    intimate titles can only be retrieved under a ceiling that already forces
+    INTIMATE (local-only) routing. Pinned by
+    ``test_routing_tier_dominates_every_retrieved_fragment_tier``.
+
+    **The cost, honestly.** Each call constructs a fresh
+    ``RetrievalSpecialist``, so it parses the corpus and loads the embeddings
+    parquet per call, and live-embeds any fragment whose cached
+    ``content_hash`` is stale or missing — on a cold cache, that is the whole
+    admitted corpus. Tracked by #1034; the ``except Exception`` below does not
+    bound it, because slowness is not an exception.
+
+    A known asymmetry: a fragment with **no** ``privacy_tier`` key at all is
+    admitted here from ``ceiling=personal`` (the ``Fragment`` model default is
+    ``unclassified``), while :func:`_fragment_tier` fails that same file closed
+    to INTIMATE as an *entry*. Tracked by #1033.
+
+    Args:
+        query: The entry text to retrieve against.
+        vault_path: Vault root whose corpus subtrees are searched.
+        override: The ceiling-derived admission override, applied as a hard
+            cutoff inside the corpus walk.
+
+    Returns:
+        Fragment titles, most relevant first — ``[]`` when retrieval fails.
     """
     try:
         from creek.author.agents import RetrievalSpecialist
 
         bundle = RetrievalSpecialist().gather(query, vault_path, override=override)
-        # ``EvidenceClaim`` carries ``claim``, not ``text``, so this access has
-        # always raised and been swallowed below: production grounding has
-        # never actually run. Renaming the attribute switches it on, and what
-        # it would then feed the prompt is corpus fragment *titles*
-        # (``_fragment_claim`` sets ``claim=fragment.title``) — grounding on
-        # titles rather than bodies is an undecided design question that needs
-        # its own privacy tests. Suppressed rather than "fixed" so bringing
-        # this package under mypy does not quietly turn that egress on; #964
-        # owns the decision.
-        return [claim.text for claim in bundle.claims]  # type: ignore[attr-defined]  # Issue #964
-    except Exception:
+        return [claim.claim for claim in bundle.claims]
+    except Exception as exc:
+        # This log line is the #964 remedy. The attribute this comprehension
+        # reads was wrong for the entire life of the feature, so grounding
+        # raised on every production call and every reflection was ungrounded —
+        # undetected precisely because this swallow was silent. The log is what
+        # makes the next such failure findable.
+        #
+        # The catch stays broad on purpose: ``RetrievalSpecialist.gather``
+        # reaches ``load_config``, a full corpus parse and the
+        # sentence-transformer model load, so its failure surface is open-ended
+        # (ImportError, OSError, yaml errors, pydantic ``ValidationError``,
+        # model-load failures) and none of it may cross the MCP boundary.
+        #
+        # Log the exception's *type name only* — never ``str(exc)``,
+        # ``exc_info``, or ``logger.exception``, for the reason given at
+        # ``_resolve_entry`` above: these are the same user-managed corpus
+        # files, and ``yaml.MarkedYAMLError`` stringifies with the offending
+        # source snippet, so the message text would write tier-unknown vault
+        # content into a log that carries no tier and is not covered by the
+        # ceiling.
+        logger.debug("Grounding degraded to none (%s)", type(exc).__name__)
         return []
--- a/creek-tools/tests/test_mcp_reflect.py
+++ b/creek-tools/tests/test_mcp_reflect.py
@@ -22,6 +22,15 @@ contract this suite pins, in order of stakes:
 4. Corpus-grounded retrieval, audit logging, read-only wrt the corpus, clean
    degradation with no provider, the care seam (#753), and tolerance of
    malformed fragment frontmatter under ``01-Fragments`` (#847).
+5. **The default (production) grounder actually grounds (#964)** — every other
+   test in this file injects a ``retrieve=`` stub, so the retrieval
+   ``build_server`` really wires had zero coverage and silently returned no
+   grounding at all. The final section drives that path with no ``retrieve=``:
+   it feeds the prompt corpus fragment **titles** (never bodies), admits
+   exactly the within-ceiling corpus and leaves no trace of anything above it,
+   routes at a tier dominating every fragment it retrieved, and on a retrieval
+   failure degrades to ``(none)`` while logging the exception's *type name*
+   only.
 
 The LLM and retrieval are injected — no live calls.
 """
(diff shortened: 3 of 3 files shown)


Geoffe-Ga:
## Summary
PR #1037 fixes a real production bug in `creek_mcp/tools/reflect.py`: `_default_retrieve` read `EvidenceClaim.text`, an attribute that has never existed (the model carries `claim`), so grounding retrieval raised `AttributeError` into a bare `except Exception: return []` on every call — every `creek.reflect` reflection has been ungrounded in production since the tool shipped, despite the docstring's claim otherwise. The fix renames the attribute access, adds redacted debug logging for the (still intentionally broad) exception swallow, and adds 13 new tests that exercise the previously-never-executed default (unstubbed) grounding path.

## Strengths
- **Root cause, not a patch**: removes the `# type: ignore[attr-defined]` suppression that was hiding the bug, rather than papering over it — squarely in line with CLAUDE.md's "no shortcuts / fix root causes" principle.
- **TDD gate order verified**: commit history is `test(mcp): pin ... unstubbed production grounding` followed by `fix(mcp): activate ...`, i.e. RED before GREEN, matching CLAUDE.md's Gate 1.
- **Verified by independent trace, not just trust**: confirmed `EvidenceClaim.claim` (not `.text`) is the real attribute, `_fragment_claim` sets `claim=fragment.title` (so grounding is titles, not bodies, as documented), `_load_corpus`'s `tier_within_override` is a hard rank cutoff (no residue path, distinct from the #931/#1032 title-leak shape), and `routing_tier` takes the max of ceiling-floor and content tier — so the "routing tier always dominates every retrieved fragment" claim in the new tests is actually sound, not just asserted.
- **Security-conscious logging**: `logger.debug("Grounding degraded to none (%s)", type(exc).__name__)` logs only the exception type, matching the established precedent at `_resolve_entry` (reflect.py:326-336) for the same reason (user-managed corpus files; `yaml.MarkedYAMLError` embeds source text in `str(exc)`).
- **Test design actively defends against false-green**: `_write_corpus_fragment` is deliberately kept separate from the file's existing `_write_fragment` helper (which is invisible to `_load_corpus` since it omits `type: fragment`) — the PR description notes this is literally the trap that would have made the whole new suite pass vacuously.
- **Scope discipline**: three related-but-out-of-scope findings (#1033 fail-open vs fail-closed tier default asymmetry, #1034 per-call `RetrievalSpecialist` reconstruction cost, #1036 read-gate layer (f) blind spot) were filed as separate issues rather than folded in or ignored.
- `read_gate.py`'s `ToolPosture` rationale for `creek.reflect` was updated to describe both now-live unsupplied-read gates, keeping the manifest (which is itself test-verified per its module docstring) accurate.

## Security Concerns
🟢 LOW — none identified. The grounding retrieval path is correctly bounded by the same hard tier cutoff as other corpus reads, doesn't introduce a new read primitive, and the exception-logging change reduces exposure (type name only) rather than increasing it. The PR text also explicitly rules out the #931/#1032 title-leak shape with a targeted regression test (`test_above_ceiling_fragment_contributes_nothing_not_even_a_title`).

## Problems
None found blocking merge.

## Code Quality
- Non-blocking: `_default_retrieve`'s docstring is very long (multiple bolded sub-sections). It's justified here — this is exactly the kind of non-obvious design rationale (why titles not bodies, why the catch stays broad, a known asymmetry with #1033) CLAUDE.md's "why over what" comment guidance calls for — but future edits to this function should keep the docstring in sync rather than let it drift from the code.
- The performance cost noted in the new docstring (fresh `RetrievalSpecialist` per call, full corpus re-parse + re-embed on a cold cache) is honestly disclosed and correctly deferred to #1034 rather than silently accepted or fixed out-of-scope.

## CLAUDE.md Adherence
- ✅ TDD (test-first commit, then fix)
- ✅ No shortcuts — the pre-existing `# type: ignore` suppression is removed as a *consequence* of fixing the root cause, not a new one added
- ✅ Docstrings — Google-style Args/Returns sections added, matching module convention
- ⚠️ Gate 3 (CI green) — CI was still `IN_PROGRESS` at the time of this review (no failures observed); confirm it completes green before merge, per the "never request review with failing checks" / "only merge when CI is fully green" rule
- I could not execute `./scripts/check-all.sh`, `pytest`, or `ruff` directly in this review environment (command execution was restricted), so tool-reported results (coverage, mutation testing) are taken from the PR description and cross-checked via manual code trace rather than re-run locally.

## Testing
Strong. 13 new tests target exactly the path that had zero coverage before this PR (the unstubbed default grounder that `build_server` actually wires in production): title-not-body rendering, per-ceiling admission set-equality (not just leak-absence, so a filter that's too tight also fails), routing-tier dominance (both a hard-coded table check and a derived-inequality check), above-ceiling non-residue, retrieval-failure graceful degradation, and log redaction (type name present, message sentinel absent). The `retrieve=` injection-bypasses-default test guards against the two grounding sources silently merging now that the default is live. Per the PR description, none of the 566 test-file lines were deletions, and the file's sha256 is claimed unchanged from RED through Gate 2.5 — a reasonable integrity signal, though obviously unverifiable directly by me in this review pass.

## Documentation
Module docstring, `reflect_tool`'s `retrieve` parameter docs, and `_default_retrieve`'s docstring are all updated to accurately describe the newly-activated behavior (titles-only grounding, the ceiling invariant, the routing consequence, the known cost and the known tier-default asymmetry). The `read_gate.py` `ToolPosture` rationale for `creek.reflect` was updated in step with the code change so the manifest doesn't drift from reality.



## Verdict: LGTM

No blocking defects surfaced under independent verification: the root-cause attribute fix (EvidenceClaim.claim, not .text) is correct, the ceiling/routing-dominance invariants it relies on hold in creek/author/agents.py and creek_mcp/tier_ceiling.py, the new logging follows the exact type-name-only precedent already established at reflect.py:326-336, and the 13 new tests cover the previously-uncovered production grounding path (admission set-equality per ceiling, above-ceiling non-leakage, routing dominance, failure degradation, log redaction, and the retrieve= injection seam) without weakening any threshold or adding a suppression. CI was still in progress at review time (no failures observed), so confirm it finishes green before merge per Gate 3.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5137843471 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Fixes #964: `_default_retrieve` in `creek_mcp/tools/reflect.py` read `EvidenceClaim.text`, an attribute that doesn't exist (the model carries `claim`), so every production call raised `AttributeError` into a bare `except Exception: return []`. Since `build_server` never wires `retrieve=`, **every production `creek.reflect` call has been ungrounded since the feature shipped**. The fix renames the attribute access, adds a type-name-only debug log so future degradations are findable, and settles the previously-undecided design question of grounding on fragment *titles* (never bodies), enforced by a hard tier-rank cutoff (`tier_within_override`) rather than the summarise-and-leak-a-title path that caused #931/#1032.

## Strengths
- Root cause is precisely identified and the fix is a one-line attribute correction plus a log line — no scope creep into unrelated refactors.
- Extremely disciplined test suite: 13 new tests, each with a stated purpose, including an explicit "trap" avoidance (`_write_corpus_fragment` vs. the lookalike `_write_fragment`, which would silently make every new assertion pass vacuously against an empty corpus).
- Verified all load-bearing technical claims directly against source: `EvidenceClaim.claim` (`creek/author/models.py:89`), `_fragment_claim` setting `claim=fragment.title` (`creek/author/agents.py:143-149`), `tier_within_override`'s hard rank cutoff used inside `_load_corpus` (`creek/author/agents.py:115`), and that `build_server` truly leaves `retrieve` unset in production (`creek_mcp/server.py:332-346`) — all match the PR's description exactly.
- Good design reasoning for titles-vs-bodies grounding, tied to `_build_prompt`'s verbatim-quote validation and the shared `_fragment_claim` contract with `GraphSpecialist`/Writing Desk.
- The routing-tier-dominates-retrieval invariant is pinned both by a hardcoded table (catches a policy retune) and a derived inequality (catches drift/new tiers) — good belt-and-suspenders test design.
- Correctly distinguishes this activation from the #931/#1032 title-egress hazard by construction (`tier_within_override` drops the record before ranking; no summarise-the-body residue exists to leak), and pins that distinction with `test_above_ceiling_fragment_contributes_nothing_not_even_a_title`.
- Honest scope discipline: three follow-on findings (#1033 untiered-fragment inconsistency, #1034 per-call corpus reload cost, #1036 read-gate manifest blind spot) are filed rather than folded into this PR.
- `read_gate.py` posture rationale updated to reflect the second now-active unsupplied-read path, keeping the manifest accurate.

## Security Concerns
🟢 LOW — The exception handler intentionally logs only `type(exc).__name__`, never `str(exc)` or `exc_info`, to avoid writing tier-unknown vault content (e.g. a YAML error's offending source snippet) into an unclassified log stream. This mirrors the existing `_resolve_entry` precedent in the same file and is a sound, security-conscious choice. No new egress path is introduced; grounding remains title-only and is bounded by the same ceiling used elsewhere.

## Problems
None found blocking merge.

## Code Quality
- The degradation log uses `logger.debug(...)`. Since the whole point of this PR is that a production-breaking failure went undetected for the feature's entire life, consider whether `logger.warning` (still type-name-only) would make a *future* regression more discoverable at typical production log levels, where DEBUG is usually off. Not blocking — it matches the existing `_resolve_entry` convention in this file, and the failure now degrades gracefully rather than breaking the tool, so DEBUG is a defensible choice.
- `_default_retrieve`'s docstring is unusually long (relative to the function body) but the extra detail is genuinely load-bearing given how subtle the tier/routing invariants are — reasonable trade-off.

## CLAUDE.md Adherence
✅ TDD workflow followed (RED verified before fix per PR description). ✅ No `# noqa`, `# type: ignore`, skips, or threshold changes anywhere in the diff — the pre-existing `# type: ignore[attr-defined]` suppression is removed as a consequence of the real fix, not bypassed. ✅ Conventional commit-style PR title (`fix(mcp): ...`). ✅ No dependency/lockfile changes. Coverage/lint/mypy/complexity gates reported green in the PR's test plan (94.01% branch coverage, 12/12 checks) — consistent with the repo's ≥90% coverage and zero-Ruff-violation thresholds.

## Testing
Strong. Tests drive the actual production path (no `retrieve=` injected, except the one test whose subject is the seam itself), assert on the exact prompt content via a sliced `SOURCE FRAGMENTS` section (avoiding false positives/negatives from the `ENTRY:` block), use set-based (not order-based) assertions to avoid flakiness from the hash-randomized embedding mock, and include both a positive control (the `open` fixture) and mutation-style precision checks (exact equality on the rendered grounding line, not just "title is somewhere in the prompt"). The PR further reports 2 mutation kills. This satisfies the "insufficient testing is BLOCKING" bar comfortably — testing here is a non-issue.

## Documentation
Docstrings for `_default_retrieve`, `reflect_tool`'s `retrieve=` param, and the module-level docstring were all updated to reflect the new behavior, the design rationale, and the known follow-on issues (#1033, #1034). `read_gate.py`'s manifest rationale was updated in lockstep. Documentation is thorough to the point of being verbose in places, but every claim I spot-checked against source was accurate.



## Verdict: LGTM

All technical claims in the PR (EvidenceClaim.claim vs .text, _fragment_claim setting claim=fragment.title, tier_within_override's hard rank cutoff in _load_corpus, routing_tier/tier_sensitivity dominance, retrieve= left unwired in build_server) were independently verified against the actual source and are accurate. The fix is minimal and correct, thoroughly tested (13 new tests covering the exact regression, ceiling admission, routing dominance, leak-absence, and failure degradation), and the broad except/debug-log choices are deliberately justified and consistent with existing precedent in the file. No blocking issues found.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5138014257 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #1078, opened 2026-07-31
fix(mcp): gate creek.state.read/render on a tier-stamped state artifact (#969)

### LINKED ISSUE #969: creek.state.read / creek.state.render have no tier-ceiling gate; state.read's docstring asserts one that does not exist

Geoffe-Ga:
## Summary

`creek.state.read` and `creek.state.render` accepted a `privacy_tier_ceiling`, wrote it to the audit log, echoed it in the response — and never consulted it. `read` returned `00-Creek-Meta/State/latest.md` verbatim at any ceiling; `render` walked the whole vault with no override threaded into `StateReportGenerator`.

### Three leaks, reproduced at HEAD before writing a line

Running `StateReportGenerator(vault_path=v).render()` over a temp vault, no ceiling anywhere:

| Section | What escaped |
|---|---|
| `## Drift warnings` | `` - `/Users/…/vault/01-Fragments/Notes/private-<CANARY>-moment.md` `` — an `intimate` fragment's **slugified title**, inside an **absolute** path |
| `## Liminal Watch` | `- unnamed-<CANARY>-2` — the **file stem** of a `10-Liminal/Unnamed` note |
| `## Active eddies` | `- Eddy <CANARY> One — 0 fragment(s)` — an eddy whose only member fragment is above the ceiling |

This **corrects the issue on two points.** It says "there is no demonstrated body-level leak today — the exposure is structural, at the title level, and blocked on the sub-issue below." The exposure is at title level, but it is demonstrated, and it is **not** blocked on giving `Eddy`/`Thread` a `privacy_tier`: leaks 1 and 2 are tierable today, and leak 3 is tierable by **deriving** the eddy/thread tier from its constituent fragments at render time — the issue's own stated derivation, without a model field, a pipeline change, or a vault migration. The author's probe missed leak 1 only because their fixture had no fragment older than the 90-day `_STALE_FRAGMENT_AGE_DAYS` threshold.

The absolute path is a second defect in the same rows: `docs/generation.md` already claimed the artefact "does not leak an operator's home directory". Drift rows now render vault-relative, which makes the standing claim true.

## The cross-ceiling cache — the design question, and the answer

Gating the read alone does not fix it. If `latest.md` was rendered at `ceiling=all`, refusing an `open` read is right, but an `open` read is *also* wrong to serve from a cache built at a broader ceiling.

**(a) The artifact stamps itself**, as YAML frontmatter: `type: state-report`, `privacy_tier: <highest tier admitted>`, `tier_ceiling: <override it ran under>`. The key is `privacy_tier` **deliberately**, so the read side reuses `raw_privacy_tier` verbatim — `privacy_filter`'s docstring warns that "a tier reader somewhere else is a tier opinion the others can drift from", and this change adds none. Scalars only, `yaml.safe_dump(sort_keys=False)`: CrawDad keeps `raw_markdown` including frontmatter and feeds it to prompts, and its bullet regex is `^\s*-\s+`. Verified no consumer breaks — `state-budget` measures size, CrawDad's `_split_sections` discards everything before the first `## `, and no `type:`-dispatching loader is rooted at `00-Creek-Meta`.

**(b) `render` filters at render time.** Without it a `ceiling=open` render is either a lie (stamped `open`, full of intimate-derived titles) or useless (stamped `intimate`, unreadable by the caller who asked for it).

**(c) `read` compares the stamped `privacy_tier`, not the render ceiling.** Strictly better: a report rendered at `ceiling=all` over an all-`open` vault stamps `open` and stays readable at `open`. It is also the honest question — the ceiling is a statement of intent, the stamp is a fact about the bytes.

**(d) An unstamped legacy `latest.md` fails closed — accurate, not cautious.** Every `latest.md` written before this change was rendered completely unfiltered, i.e. the equivalent of `--include-tier all`. Failing closed states a fact about those bytes.

**Upgrade path for an operator with an existing `latest.md`.** CLI: nothing changes; `creek state` re-renders and stamps on the next run. MCP/CrawDad: the first `creek.state.read` at the default `ceiling=open` (so `/creek`, `/creek phase`, `/creek wavelength`, CrawDad's check-in) returns `status: "refused"` with `GENERIC_ABOVE_CEILING_REASON` — byte-identical to a legitimately-intimate refusal, because a distinguishable "legacy" reason would itself be an oracle for whether the vault holds above-ceiling content. Recovery is one command: `creek.state.render`, or `creek state --include-tier open`. Nothing is lost: the ISO-week archive is untouched and `ceiling=all` admits every stamp, including the unstamped one, so **no report is ever permanently unreachable.** Documented in `docs/mcp.md` and both tool docstrings, not in the refusal payload, which keeps exactly its four canonical keys and no `content` key at all.

**Refuse vs exclude, split down the middle, and the split is load-bearing.** `render` **excludes** — it names no target, it is the most corpus-walking tool on the surface. `read` **refuses** — it addresses one atomic cached artifact, and you cannot admit half a rendered markdown document without re-rendering it, which is what `render` is. The asymmetry is what makes recoverability provable: `render` is the recovery path for a refused `read`, so if `render` could refuse, a vault with one intimate fragment would have a permanently unreachable report.

## Per-section disposition (all eleven, triaged out loud)

Gated: wavelength snapshot and vault summary (**counts narrow with the ceiling** — `creek.wheel` is a frequency tally and is `GATED`, so a per-tier count is content here; an unfiltered count discloses the existence of what the rest of the report omitted, and it would make the stamp under-report); liminal watch (`within_ceiling` on raw frontmatter); active eddies/threads (derived tier = max over **all** constituent fragments — the unfiltered pass is load-bearing, since filtering first would resolve a mixed-tier eddy to `open` and leak its title); synchronicities (**already fail-closed via `_is_leaf` once the fragment load narrows — verified, zero code change**, and now pinned by a test); hyperedges (every `derived_from` id must resolve, and the span set is intersected with admitted eddies); drift warnings (admitted source paths only, rendered vault-relative); suggested questions (snapshot narrowed, and dropped whole below `ceiling=personal`); lint summary (`ceiling=intimate` or broader — an untierable verbatim copy).

Ungated, deliberately: **pre-LLM yield** alone. It reports one pipeline run's totals and names no fragment, so there is nothing in it to filter *by*.

A `_SECTION_DISPOSITIONS` table asserts this covers `SECTION_ORDER` exactly, in both directions. A new section must be triaged out loud or the stamp silently under-reports and the read gate fails open.

## Layer (f) is not the evidence, and I did not treat it as one

PR #1068 proved the shared read-gate envelope probe is structurally incapable of catching a write-side leak. `creek.state.render` is write-side too. So **every exclusion assertion in the new suite is on written artifact bytes** — the ISO-week file and `latest.md`, symlink-resolved — never only the response. On top of that: a **vault-wide byte-snapshot sweep** whose file list is derived from the filesystem diff, so an unanticipated artifact or a parallel ungated write is caught; and **inverting-predicate mutation rows** per gated section, so a gate that is called and discarded fails. Every exclusion has a paired positive control, so nothing passes by writing an empty report.

Manifest posture moved `UNGATED_KNOWN_GAP` → `GATED` for both tools, re-pointed at the gates that closed them, with two new runtime probes. `_probe_state_read` renders at `ceiling=all` **first**, so the artifact genuinely holds the canary — otherwise it would be vacuous on a vault with no report.

## The known-vacuous test, replaced

`tests/test_mcp_tools.py::test_state_read_does_not_embed_fragment_bodies`.

**Before:** wrote an intimate fragment, then **hand-wrote** `latest.md` as the fixed string `"# Audit report\n\n- Fragments: 1\n"`, then asserted `"secret journal body" not in result["content"]`. True by construction — nothing in `state_read_tool` ever read the fragment, so `_write_fragment` was decorative. It could not fail whether or not a gate existed, and it passed for the entire life of the gap.

**After** (`test_state_read_does_not_serve_above_ceiling_content_from_latest_md`): the artifact is produced through the production path, `state_render_tool(..., ceiling=ALL)`; the canary reaching `latest.md` is **asserted first** so the setup cannot rot back into vacuity; then `state_read_tool(..., ceiling=OPEN)` must refuse. **Executed against unfixed HEAD it fails** with `AssertionError: assert 'ok' == 'refused'` — and the pre-assertion passes there, which is itself proof the render leaked.

## Gate 2.5 found two more leaks — of the same class — and I reproduced both

An independent `code-review-orchestrator` pass returned `FIX_REQUIRED`. It was right twice, on exactly the bug this issue is about: an above-ceiling identifier riding in through a section nobody threaded the ceiling into.

1. **`## Hyperedges`** — `_fragment_to_eddies` mapped admitted fragments to *every* eddy they named without intersecting the admitted eddies, so an eddy correctly dropped from `## Active eddies` still printed in a `spans:` list at `ceiling=open`. Reproduced: `- Bridging praxis — spans: Mixed ZZEDDYCANARY, Plain`. Worse, its tier never reached the stamp, so the report was stamped `open` and `read` served it. Fixed by intersecting before the `len(spanning) >= 2` cut. **Its docstring claimed the opposite** — a fresh instance of this issue's own defect, now made true.
2. **`## Suggested questions`** — `mining.py`'s `_load_mining_snapshot` gates fragments but loads `02-Threads`/`03-Eddies` ungated, so a thread whose members are all `intimate` titled a mined seed at `ceiling=personal`. Reproduced: `- Secret ZZTHREADCANARY — Thread 't1' has accumulated 12 fragments…`. Fixed **inside `creek state`** — `phase_filtered_seeds` gained an additive `snapshot=` passthrough, and the generator narrows the snapshot with the lists it has already admitted. Deliberately **not** fixed by gating `mining.py`'s loaders: `Thread`/`Eddy` carry no `privacy_tier`, so a raw gate there would fail closed and silently delete two mining strategies for every `creek mine` caller — a separate decision on a separate tool.

Fixing 2 surfaced two further fail-open axes: `snapshot.fragments` (the miner reads the tier off the validated model, whose missing key defaults to `unclassified`, while the report's raw reader fails closed to `intimate`) and `snapshot.liminal_fragments`. Both closed.

All three new canaries were **executed against the pre-fix files and confirmed RED**. A re-review returned **`CLEAN`**.

## Consequences stated on the tin

- **`latest.md` is a single slot shared across ceilings**, and the MCP `render` default is the strictest `ceiling=open` — so a bare MCP render *narrows* the cached report for everyone, including subsequent CLI reads. Kept deliberately: the MCP boundary defaulting strict is the security posture, and per-ceiling filenames would multiply vault artifacts and break `latest.md` as the documented session-start context.
- `10-Liminal/Paradoxes` notes are `type: paradox` with no `privacy_tier`, so they fail closed and vanish from the Liminal Watch below `ceiling=intimate`.
- The **lint summary** is intimate-only. This closes the live caveat `creek_mcp/read_gate.py` already recorded under `creek.lint` — "unreachable through MCP today … read it against #969's state-artifact gap". `creek state` **is** the surface that served it back.
- **Behaviour widening, called out because it is a loosening inside a privacy PR:** the seeds section is now mined at the render's override. At `ceiling=open` this is unchanged (the section is dropped there anyway); the widening is at `ceiling=all`.
- The mining corpus's liminal axis is **accounted for on the stamp rather than narrowed** (state has no admitted list for `10-Liminal/Compost`). Only an opaque generated id ever renders, and an untiered note pushes the stamp to `unclassified` — ranked with `personal`, so still refused at `open`. The residual posture split on the same file is documented and filed as a follow-up.
- `creek_mcp/read_gate.py`'s "Why these primitives have no production caller yet" paragraph is **rewritten**, because this change makes it false. Leaving it would have created this issue's defect in the file whose whole subject is the manifest not lying.

## Test plan

`cd creek-tools && ./scripts/check-all.sh` → **exit 0, 12/12 checks, 6794 passed, 94.02→94.04% branch coverage**, per-file gate clean (219 files ≥ 80%).

- **New:** `tests/test_state_tier_ceiling.py`, 3240 lines — per-section exclusion + admission across all four ceilings on artifact bytes; the stamp (shape, body round-trip, `privacy_tier <= tier_ceiling` over all four ceilings, all-open-at-`ALL` → `open`, fresh vault → `open`, `latest.md` across the symlink path and **both** copy-fallback triggers); refusal (canonical four keys, byte-identical across stamped tiers, no `PrivacyTier` string beyond the caller's own echo, legacy unstamped, ten malformed-stamp shapes, missing report still `empty`); **recoverability executed end to end** over all four render ceilings; mutation resistance; the filesystem-diff sweep; six CLI tests.
- **Test integrity.** Deletions vs `origin/main`: `test_mcp_read_gate.py` **3** (two `_PINNED_GAPS` rows *re-pointed* into `_PINNED_GATE_ROWS`, which is the move the manifest's own failure message demands; one docstring line whose word "four" became factually wrong when two of four entries left, replaced by a longer one). `test_mcp_server.py` **1** (a docstring first line, preserved verbatim inside its expansion). `test_mcp_tools.py` **10** (the vacuous test's body and docstring, replaced above; and one fixture line gaining the stamp — all four of that test's assertions kept byte-identical, with a new test added beside it pinning the case the fixture change vacated). `test_state_tier_ceiling.py` **0**. **No test and no assertion was deleted or weakened.** SHA-256 of every touched test file was recorded at RED and re-checked after each later phase; the implementation phases changed none of them.

Closes #969
Refs #932, #846, #848, #968, #1068, #961, #1033

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -3251,6 +3251,11 @@ def report(
 @app.command()
 def state(
     vault: Path | None = typer.Option(None, help="Obsidian vault path"),
+    include_tier: str | None = typer.Option(
+        None,
+        "--include-tier",
+        help=_REPORT_INCLUDE_TIER_HELP,
+    ),
 ) -> None:
     """Render ``00-Creek-Meta/State/<iso-week>.md`` audit report.
 
@@ -3258,15 +3263,40 @@ def state(
     re-runs classification, linking, or compile. It reads existing
     fragments, threads, eddies, praxis, synchronicities, and the most
     recent ``run-summary.jsonl`` line, and writes a single markdown
-    document organised in seven sections (vault summary, pre-LLM yield,
-    active eddies, active threads, surprising connections, hyperedges,
-    drift warnings). ``latest.md`` next to the ISO-week file always
-    points at the most recent report.
+    document organised in eleven sections (wavelength snapshot, vault
+    summary, pre-LLM yield, liminal watch, active eddies, active threads,
+    surprising connections, hyperedges, drift warnings, suggested
+    questions, lint summary). ``latest.md`` next to the ISO-week file
+    always points at the most recent report.
+
+    ``--include-tier`` runs the same way round as it does on ``report``
+    (#968): omitting it leaves the render **unfiltered**, because the CLI
+    operator is the vault owner, and supplying it NARROWS what the artifact may
+    contain. It is also half of the documented upgrade path for a pre-#969
+    ``latest.md``: ``creek state --include-tier open`` re-renders and
+    re-stamps so an ``open``-ceiling MCP reader stops being refused.
+
+    Privacy override audit: the entry is written only when the flag was
+    **explicitly supplied**. ``override_elevates`` answers ``True`` for
+    ``PrivacyTierOverride.ALL``, so auditing the resolved ceiling would write a
+    line on every bare ``creek state`` — the loudest possible record of the one
+    case where the operator asked for nothing.
     """
     from creek.generate.state import StateReportGenerator
 
+    override = _parse_include_tier(include_tier)
     vault_path = _resolve_vault(vault)
-    written = StateReportGenerator(vault_path=vault_path).write()
+    if include_tier is not None:
+        _audit_privacy_override_if_needed(
+            vault_path=vault_path,
+            command="state",
+            override=override,
+            fragment_ids=[],
+        )
+    written = StateReportGenerator(
+        vault_path=vault_path,
+        override=override or PrivacyTierOverride.ALL,
+    ).write()
     console.print(f"[bold green]State report written: {written}[/bold green]")
 
 
--- a/creek-tools/creek/generate/mining.py
+++ b/creek-tools/creek/generate/mining.py
@@ -1635,6 +1635,7 @@ def phase_filtered_seeds(
     *,
     n: int = DEFAULT_PHASE_FILTERED_SEEDS,
     miner: IdeaMiner | None = None,
+    snapshot: MiningSnapshot | None = None,
 ) -> list[IdeaSeed]:
     """Return up to *n* :class:`IdeaSeed` objects filtered for *phase*.
 
@@ -1653,6 +1654,22 @@ def phase_filtered_seeds(
             zero.
         miner: Optional pre-configured :class:`IdeaMiner`. Defaults to
             a fresh instance with library defaults.
+        snapshot: Optional pre-loaded :class:`MiningSnapshot`, forwarded
+            verbatim to :meth:`IdeaMiner.mine_all`, which already accepts
+            one. ``None`` (the default) leaves the miner to load its own, so
+            no existing caller's behaviour changes.
+
+            ``creek state`` supplies a snapshot it has narrowed to the
+            threads and eddies its tier ceiling admitted (#969). That
+            narrowing cannot be made here, or in
+            :func:`_load_mining_snapshot`: :class:`~creek.models.Thread` and
+            :class:`~creek.models.Eddy` carry no ``privacy_tier`` field, so a
+            raw-frontmatter gate on ``02-Threads`` / ``03-Eddies`` would have
+            no evidence to read and would fail closed on *every* thread and
+            eddy — silently deleting the thread-terminus and liminal-cross-
+            eddy strategies for every ``creek mine`` caller below
+            ``ceiling=intimate``. A caller that can derive those tiers
+            passes the result in; this module keeps its own behaviour.
 
     Returns:
         Up to *n* seeds ordered by descending score, then strategy, then
@@ -1666,7 +1683,11 @@ def phase_filtered_seeds(
     except ValueError:
         current_phase = Phase.UNCLASSIFIED
     active_miner = miner if miner is not None else IdeaMiner()
-    all_seeds = active_miner.mine_all(vault_path, current_phase=current_phase)
+    all_seeds = active_miner.mine_all(
+        vault_path,
+        current_phase=current_phase,
+        snapshot=snapshot,
+    )
     allowed = _PHASE_AWARE_STRATEGIES.get(phase_value)
     filtered = (
         all_seeds
--- a/creek-tools/creek/generate/state.py
+++ b/creek-tools/creek/generate/state.py
@@ -20,6 +20,28 @@ reads what is already on disk.
 FEAT-007 inserts a wavelength snapshot and a suggested-questions
 section between sections 1 and 2; the section-ordering contract
 pinned by tests here makes that insertion straightforward.
+
+**Tier ceiling (#969).** The generator takes an
+:class:`~creek.classify.privacy_filter.PrivacyTierOverride` and admits content
+per section against it; :meth:`StateReportGenerator.write` then stamps the
+artifact with the highest tier it actually admitted, which is what
+``creek.state.read`` compares a caller's ceiling against. Two properties of
+that design are easy to break and are stated here rather than only at the call
+sites:
+
+* The corpus is loaded **unfiltered first** and only then admitted. An eddy's
+  tier is the maximum over *all* its member fragments, so filtering before
+  deriving would resolve an eddy with one ``open`` and one ``intimate`` member
+  to ``open`` and leak its title.
+* The default is ``ALL``. ``creek state``'s operator is the vault owner, and a
+  narrower default would silently delete most of an existing operator's report
+  — a behaviour regression wearing a privacy fix's costume.
+
+The type this module takes is ``PrivacyTierOverride``, never
+``creek_mcp.tier_ceiling.TierCeiling``: ``creek`` may never import
+``creek_mcp`` (the layering rule pinned at ``creek/ingest/journal_staging.py``
+and ``creek/care/__init__.py``), so the MCP wrapper converts at its own
+boundary.
 """
 
 from __future__ import annotations
@@ -29,24 +51,47 @@ import logging
 import os
 import re
 from collections import Counter
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from datetime import UTC, date, datetime
-from pathlib import Path  # noqa: TC003  # plain stdlib import; no lazy benefit
+from pathlib import Path  # plain stdlib import; no lazy benefit
+from typing import TYPE_CHECKING, TypeVar
 
 import frontmatter
 import yaml
 from pydantic import ValidationError
 
+from creek.classify.privacy_filter import (
+    PrivacyTierOverride,
+    max_source_tier,
+    raw_privacy_tier,
+    tier_of,
+    tier_within_override,
+    within_ceiling,
+)
 from creek.clean.hygiene import BrokenLinkScanner, OrphanScanner
 from creek.generate.indexes import FREQUENCY_NAMES
-from creek.generate.mining import IdeaSeed, phase_filtered_seeds
+from creek.generate.mining import (
+    IdeaMiner,
+    IdeaSeed,
+    MiningSnapshot,
+    _load_mining_snapshot,
+    phase_filtered_seeds,
+)
+from creek.generate.state_tiers import (
+    derived_link_tiers,
+    max_admitted_tier,
+    stamp_report,
+)
 from creek.generate.wavelength import (
     DEFAULT_CURRENT_PHASE_WINDOW_DAYS,
     CurrentPhaseSummary,
     current_phase_summary,
 )
 from creek.hierarchy import select_by_policy
-from creek.models import Eddy, Fragment, Frequency, Praxis, Thread
+from creek.models import Eddy, Fragment, Frequency, Praxis, PrivacyTier, Thread
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Sequence
 
 logger = logging.getLogger(__name__)
 
@@ -90,6 +135,9 @@ _LIMINAL_WATCH_TOP_N: int = 5
 """Cap on the number of fresh liminal items rendered **per subfolder** — see
 :meth:`StateReportGenerator.section_liminal_watch`."""
 
+_LIMINAL_ROOT: str = "10-Liminal"
+"""Vault subdirectory holding the liminal surfaces."""
+
 _LIMINAL_SUBDIRS: tuple[str, ...] = ("Unnamed", "Paradoxes")
 """Subfolders under ``10-Liminal`` that the liminal watch surfaces.
 
@@ -111,7 +159,7 @@ _FRAGMENTS_SUBDIR: str = "01-Fragments"
 _THREADS_SUBDIR: str = "02-Threads"
 _EDDIES_SUBDIR: str = "03-Eddies"
 _PRAXIS_SUBDIR: str = "04-Praxis"
-_SYNCHRONICITIES_SUBDIR: str = "10-Liminal/Synchronicities"
+_SYNCHRONICITIES_SUBDIR: str = f"{_LIMINAL_ROOT}/Synchronicities"
 _STATE_SUBPATH: tuple[str, str] = ("00-Creek-Meta", "State")
 _YIELD_RELATIVE: tuple[str, str, str] = (
     "00-Creek-Meta",
@@ -153,19 +201,70 @@ class _SynchronicityRow:
     time_gap_days: int
 
 
+@dataclass(frozen=True)
+class _TierIndex:
+    """Everything the ceiling decided, resolved once at load time (#969).
+
+    Lives on :class:`_VaultState` rather than on
+    :class:`StateReportGenerator` for two reasons. The generator already
+    carries exactly seven instance attributes, which is pylint's ``R0902``
+    ceiling, and a snapshot recording the ceiling it was taken under is the
+    semantically apt place for it anyway — the same claim the artifact stamp
+    makes about the document.
+
+    Attributes:
+        override: The ceiling the snapshot was taken under.
+        content_tiers: One tier per admitted item across every gated section
+            resolved at load time — fragments, eddies, threads, praxis and
+            liminal-watch notes. It is the base of the artifact stamp, not the
+            whole of it: :meth:`StateReportGenerator._content_tier` adds the
+            two contributions that are only knowable at render time, and
+            states there why each one is needed. A section that admits content
+            accounted for in neither place makes the stamp under-report and
+            the read gate fail open.
+        admitted_paths: ``str(path)`` of every admitted fragment file, in the
+            exact form the hygiene scanners return, so the drift section can
+            compare without re-normalising.
+        liminal: ``{subfolder: (admitted file stem, ...)}`` in render order.
+    """
+
+    override: PrivacyTierOverride = PrivacyTierOverride.ALL
+    content_tiers: tuple[PrivacyTier, ...] = ()
+    admitted_paths: frozenset[str] = frozenset()
+    liminal: dict[str, tuple[str, ...]] = field(default_factory=dict)
+
+
 @dataclass
 class _VaultState:
     """Snapshot of every collection :class:`StateReportGenerator` reads.
 
+    Every model collection below is the **admitted** slice: what survived the
+    ceiling, not what is on disk. The unfiltered corpus is deliberately not
+    retained past :func:`_load_vault_state`, so no section can reach around
+    the gate by accident.
+
     Attributes:
-        fragments: Fragment models from ``01-Fragments``.
-        threads: Thread models from ``02-Threads``.
-        eddies: Eddy models from ``03-Eddies``.
-        praxis: Praxis models from ``04-Praxis``.
+        fragments: Admitted Fragment models from ``01-Fragments``.
+        threads: Admitted Thread models from ``02-Threads``.
+        eddies: Admitted Eddy models from ``03-Eddies``.
+        praxis: Admitted Praxis models from ``04-Praxis``.
         synchronicities: Synchronicity rows from ``10-Liminal/Synchronicities``.
+            Not filtered here: a row names two fragment *ids*, and
+            :meth:`StateReportGenerator._is_leaf` already answers ``False`` for
+            an id absent from the admitted load, which fails the row closed.
         latest_yield: Most recent line of
             ``00-Creek-Meta/Processing-Log/run-summary.jsonl``, or
             ``None`` when the log is missing or empty.
+        tiers: What the ceiling decided — see :class:`_TierIndex`.
+        mining_corpus: Memoised narrowed snapshot backing
+            ``## Suggested questions``, or ``None`` before it is first built.
+            Loaded lazily rather than in :func:`_load_vault_state` because it
+            is needed only at ``ceiling=personal`` or broader and costs a
+            second vault walk. It lives here — not on
+            :class:`StateReportGenerator`, which is at pylint's seven-attribute
+            ``R0902`` limit — so both the section that renders it and the stamp
+            that must account for it read one memoised value rather than two
+            independently-built ones.
     """
 
     fragments: list[Fragment] = field(default_factory=list)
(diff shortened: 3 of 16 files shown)


Geoffe-Ga:
## Summary
PR #1078 closes a real privacy leak: `creek.state.read` and `creek.state.render` (MCP tools) previously ignored the caller's `privacy_tier_ceiling` entirely, serving intimate/private vault content (fragment titles via orphan paths, liminal file stems, eddy/thread titles) regardless of the requested ceiling. This change adds tier-aware admission to all but one of the state report's eleven sections, stamps the rendered artifact with the highest tier it actually admitted, and gates `creek.state.read` on that stamp (refuse) while `creek.state.render` filters at render time (exclude). It also fixes two additional leaks (`## Hyperedges` and `## Suggested questions`) found during the author's own Gate-2.5 self-review.

## Strengths
- Exceptional self-documentation: every non-obvious design decision (why `read` refuses but `render` excludes, why the stamp reduction differs from `max_source_tier`'s empty-case default, why fragments/eddies/threads/praxis each need a different admission rule) is stated in a docstring or module comment, with the "why", not just the "what".
- The eddy/thread derived-tier ordering — computing `derived_link_tiers` over the **unfiltered** fragment corpus before narrowing — is correctly implemented in `_load_fragments_admitted` (`creek/generate/state.py`), matching the documented rationale that reversing the order would leak a mixed-tier eddy's title.
- The hyperedges intersection fix (`_fragment_to_eddies` intersecting wikilink targets with `self._state.eddies` titles) happens before the `spans:` list and before the `len(spanning) >= 2` cut, closing both the direct leak and the indirect "pads the rank cut" leak.
- `stamped_content_tier` collapses three distinct failure modes (unparsable frontmatter, missing stamp, unrecognized value) into one `INTIMATE` fail-closed answer — correctly avoiding an oracle that would let a caller distinguish "legacy report" from "genuinely above ceiling".
- The claimed replacement of a vacuous test (`test_state_read_does_not_embed_fragment_bodies` in `test_mcp_tools.py`) is verified: the old test hand-wrote `latest.md` as a fixed string never touched by the fragment it created, so it passed regardless of whether a gate existed. The new test (`test_state_read_does_not_serve_above_ceiling_content_from_latest_md`) renders through the real production path (`state_render_tool(..., ceiling=ALL)`), asserts the canary reached the artifact first, then asserts the read refuses it — this actually exercises the gate.
- Docs (`docs/generation.md`, `docs/mcp.md`) are updated in lockstep with the code and accurately reflect per-section behavior, including surprising edge cases (e.g., `10-Liminal/Paradoxes` notes failing closed because the model has no `privacy_tier` field).
- Layering discipline preserved: `creek/generate/state.py` takes a `PrivacyTierOverride`, never `creek_mcp.tier_ceiling.TierCeiling`, keeping the `creek` → `creek_mcp` import boundary intact.
- Absolute-path leak fixed as a bycatch (`_vault_relative`, `broken_links.py`), making a previously-false claim in `docs/generation.md` ("no absolute path in the artifact") true.

## Security Concerns
🟢 LOW — None found that rise to blocking. The core gating logic was read section-by-section against the PR's own leak inventory (11 sections) and each admission path traces correctly to the ceiling. The two secondary leaks found during the author's own Gate-2.5 pass (hyperedges intersection, suggested-questions thread/eddy narrowing) are fixed and each has a dedicated regression test using a genuine production-path canary.

One residual, explicitly-documented and non-blocking gap: the mining corpus's `10-Liminal/Compost` axis is accounted for on the stamp rather than narrowed at the section level (only the fragment's opaque id ever renders, not a title), and the author explicitly flags a "posture split on the same file" as a filed follow-up rather than silently leaving it.

## Problems
None blocking. Nit only:
- `creek_mcp/tools/state.py`'s docstring for `render` still documents `## Pre-LLM yield` as the sole ungated section — worth double-checking this stays accurate if a future edit adds a twelfth section, since the PR itself notes "a new section must be triaged out loud or the stamp silently under-reports."

## Code Quality
- The `_TierIndex` / `_FragmentLoad` dataclasses and helper decomposition (`_admit_by_derived_tier`, `_admit_praxis`, `_admitted_liminal_notes`) keep `state.py`'s per-function complexity in check despite the large surface area — consistent with the repo's `<=10` cyclomatic complexity threshold.
- `derived_link_tiers`'s empty-case (`INTIMATE`, fail-closed "no evidence") is deliberately asymmetric with `max_admitted_tier`'s empty-case (`OPEN`, "admitted nothing") — this asymmetry is explained in both docstrings and is correct, but it's subtle enough that a future contributor could plausibly "fix" one to match the other. Worth a cross-reference comment linking the two functions explicitly (currently the reasoning is duplicated in prose rather than pointing at each other).

## CLAUDE.md Adherence
✅ Conventional Commits — single commit, `fix(mcp): gate creek.state.read/render on a tier-stamped state artifact (#969)`.
✅ Test coverage — PR body reports 12/12 local checks, 94.02→94.04% branch coverage, per-file gate clean; CI (`Code Quality & Testing` 3.11/3.12/3.13) was still `IN_PROGRESS` at review time — recommend confirming green before merge per the "Stay Green" 4-gate workflow.
✅ Docstring/documentation conventions — extensive, consistent with `creek-tools/CLAUDE.md`'s standards.
✅ Layering rules — `creek/` does not import `creek_mcp` (verified in state.py's module docstring and imports).
✅ Scripts-only workflow — PR body cites `./scripts/check-all.sh`, not raw tool invocations.

## Testing
Insufficient testing would be blocking, but this PR clears that bar comfortably. Verified directly:
- `test_mcp_tools.py`'s replaced test goes through the real production path and would fail on unfixed HEAD (confirmed by reading the diff, matching the PR's stated RED-then-GREEN claim).
- `test_mcp_server.py` adds an end-to-end MCP-transport-level refusal test alongside the existing stamped-`open` happy path, so the two are pinned together rather than one silently making the other redundant.
- `test_mcp_read_gate.py`'s manifest re-points the two `creek.state.*` entries from `UNGATED_KNOWN_GAP` to `GATED` with runtime probes (`_probe_state_render`, `_probe_state_read`) that seed genuine canaries via a carrier the report actually renders (`10-Liminal/Unnamed` file stem) rather than asserting on an empty report — avoiding the classic "gate that only proves nothing was written" failure mode.
- The new `test_state_tier_ceiling.py` (3240 lines) was not exhaustively read line-by-line in this review, but its sampled companion diffs and the PR's stated methodology (per-section exclusion/admission across all four ceilings on artifact bytes, mutation-resistance rows, a filesystem-diff sweep) are consistent with the rigor observed everywhere else in the diff.

## Documentation
Complete and accurate. `docs/generation.md` gained a full per-section ceiling behavior table and an "artefact stamps its own tier" explainer; `docs/mcp.md` documents the render/read asymmetry, the legacy-unstamped-report upgrade path, and cache-thrash consequences (`latest.md` as a single cross-ceiling slot). Both were spot-checked against the implementation and match.



## Verdict: LGTM

Independent line-by-line review of the full core diff (state.py, state_tiers.py, read_gate.py, mcp tools, mining.py, wavelength.py, lint checks, docs) found no bugs, no residual leak paths, and verified the PR's most load-bearing claim (that a previously-vacuous test in test_mcp_tools.py was replaced with a production-path test that actually fails on unfixed code) by reading the diff directly. Documentation, commit conventions, and layering rules (creek/ never imports creek_mcp) all check out. No blocking issues found; only a couple of non-blocking observations noted below.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5139879506 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Summary
Closes #969: `creek.state.read` and `creek.state.render` previously ignored the `privacy_tier_ceiling` entirely, serving `00-Creek-Meta/State/latest.md` verbatim regardless of caller ceiling. This PR threads a `PrivacyTierOverride` through `StateReportGenerator`'s eleven sections, derives tiers for models that carry none of their own (`Eddy`/`Thread`, via the fragments that name them), stamps the written artifact with the highest tier it actually admitted, and gates `read` (refuse) vs. `render` (exclude) accordingly.

## Strengths
- Correct core design: corpus is loaded **unfiltered** to derive per-title max tiers, then narrowed — this is the only way to avoid an eddy with one `open` + one `intimate` member leaking its title at `ceiling=open`. Verified this ordering directly in `_load_fragments_admitted`/`derived_link_tiers`.
- The read-vs-render asymmetry (refuse on a cached atomic artifact vs. exclude on a corpus walk) is well-reasoned and consistent with the existing `#1068` compile/reflect precedent in `read_gate.py`.
- Genuinely adversarial process: an independent Gate 2.5 pass found two more leaks of the same class (hyperedge `spans:` list not intersecting admitted eddies; `mining.py`'s thread/eddy loaders being ungated), and both are fixed with dedicated, executable-against-pre-fix-HEAD regression tests (`test_hyperedges_do_not_span_an_above_ceiling_eddy_title`, etc.) rather than just narrative claims.
- Replaced a previously vacuous test (`test_state_read_does_not_embed_fragment_bodies`, which hand-wrote `latest.md` so nothing could ever fail it) with one that exercises the real production path and is asserted to fail on unfixed HEAD.
- Fixed two related, previously-undocumented absolute-path leaks (`broken_links.py` lint check, drift-warnings section) as part of making `docs/generation.md`'s existing "never an absolute path" claim actually true.
- Fail-closed posture is applied consistently and correctly at every boundary I checked: unstamped/malformed artifact → `INTIMATE`; note missing `privacy_tier` → `INTIMATE` (not the model's `unclassified` default); empty derivation → `INTIMATE` (via `max_source_tier`) except the artifact-stamp reduction, which correctly diverges to `OPEN` for "admitted nothing" (a different, deliberately distinguished case, and reasoned about explicitly in `state_tiers.py`).
- `read_gate.py`'s `TOOL_POSTURES` manifest is updated in the same PR (both `state.read`/`state.render` moved from `UNGATED_KNOWN_GAP` → `GATED`), so the audit manifest can't drift from the actual gate.

## Security Concerns
🟢 LOW — None found in the reviewed logic. The fix closes a real, reproduced information-disclosure gap (fragment titles/paths leaking through eddy/thread/liminal/hyperedge sections regardless of declared privacy ceiling). I independently verified `tier_sensitivity`, `tier_within_override`, `raw_privacy_tier`, and `max_source_tier` in `privacy_filter.py` match the fail-closed behavior the PR claims, and traced `state_read_tool`'s refusal path through `refuse_above_ceiling` to confirm the four-key, tier-non-disclosing refusal payload.

## Problems
None blocking. No bugs found in the sampled logic paths (fragment/eddy/thread/praxis/liminal admission, hyperedge intersection, stamp computation, MCP tool wiring).

## Code Quality
- `_VaultState`/`_TierIndex`/`_FragmentLoad` add real structure but also real complexity (many new small dataclasses and helper functions in `state.py`); this is justified given the multi-axis admission logic and the pylint `R0902` attribute-count constraint called out in the docstrings, but it does raise the bar for future contributors to this module — the docstrings are doing a lot of load-bearing explanatory work to compensate.
- The `## Suggested questions` gating (`_mining_corpus`) is the most intricate piece — three separate narrowing mechanisms plus a stamp-side "addition one" for the one axis (`10-Liminal/Compost`) that can't be narrowed at all. This is called out honestly in the docstrings and covered by the residual-posture note in the PR body, but it's the part of this change most likely to regress silently if `mining.py` is touched later without re-reading `state.py`'s comments.
- Minor: the CLI `state` command's `override or PrivacyTierOverride.ALL` at `cli.py:58` reads a little indirectly (only meaningful because `_parse_include_tier(None)` returns falsy) — a one-line comment there could save a future reader the trip to `_parse_include_tier`.

## CLAUDE.md Adherence
✅ TDD workflow followed — PR body documents RED/GREEN status per phase and SHA-256-pins test files across implementation phases to prove no test/assertion was weakened.
✅ `./scripts/check-all.sh` reported clean per the PR description (12/12 checks, 6794 passed, 94.02→94.04% branch coverage, per-file gate clean).
✅ No shortcuts observed — no `noqa`/`type: ignore` additions visible in the reviewed diff, and the layering rule (`creek` never imports `creek_mcp`) is explicitly preserved (`PrivacyTierOverride` is the type `state.py` takes, not the MCP `TierCeiling`).
✅ Conventional commit / PR title (`fix(mcp): ...`) matches repo convention.
Note: CI (`Code Quality & Testing`, `claude-review`) was still `IN_PROGRESS` at review time — worth confirming green before merge per Gate 3, though `CrawDad Quality` had already passed on all three Python versions.

## Testing
Extensive and well-targeted: a new 3240-line `tests/test_state_tier_ceiling.py` covers per-section exclusion/admission across all four ceilings on artifact bytes, stamp shape/round-trip, refusal payload shape (including 10 malformed-stamp variants and the legacy-unstamped case), end-to-end recoverability, mutation-resistance rows, and a filesystem-diff sweep to catch unanticipated writes. The two Gate-2.5-discovered leaks (hyperedge span list, mining corpus thread/eddy titles) each have a canary reproduced and confirmed RED against pre-fix code, then GREEN after the fix — this is real regression-test discipline, not just coverage-number padding.

## Documentation
`docs/generation.md` and `docs/mcp.md` are updated with a detailed per-section ceiling-behavior table and the read/render posture rationale; docstrings throughout `state.py`/`state_tiers.py`/`read_gate.py` are unusually thorough about *why* each admission decision was made, which will materially help future maintainers avoid reintroducing the same class of leak.



## Verdict: LGTM

The core fix is correct and thoroughly verified: I traced the admission logic end-to-end (privacy_filter.py's tier_within_override/max_source_tier/within_ceiling → state.py's per-section admission → state_tiers.py's stamping → the MCP read/render tools) and it matches the PR's own description of the design. Two additional leaks found in the author's own Gate 2.5 pass (hyperedge `spans:` list, mining-corpus thread/eddy titles) are each pinned by a dedicated regression test that fails on pre-fix code. No blocking issues found; the only remarks are minor and non-blocking.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5139906299 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1083, opened 2026-07-31
VaultWriter.update_fragment rewrites the id-index's target without verifying the file's own id


Geoffe-Ga:
## Context

Found by the security review of #970 (defense-in-depth finding, pre-existing —
not introduced by that PR).

`VaultWriter.update_fragment` locates the file to rewrite through the
per-directory `.id-index.jsonl` cache: `_find_existing_locked`
(`creek/vault/writer.py`, ~1086-1119) returns `target_dir / index[fragment_id]`
and the caller then rewrites that path with `_atomic_write_text` **without
ever verifying that the located file's frontmatter `id` actually matches the
fragment being written**.

The index is a cache. If it is stale — a filename reused by a different
fragment, a hand-moved file, a partially-restored backup, an interrupted write
— `update_fragment` will silently overwrite whatever file the stale entry
names. The id in the index is trusted; the id in the file is never read back.

## Why it is worth closing

This is the one permissive divergence between the #970 overwrite gate's
locator and the writer's locator. #970's gate ranks the tier of the fragment
it finds by walking `01-Fragments` and matching frontmatter `id`; the writer
then rewrites whatever the index points at. In every case I could construct
these agree, or the gate is *more* conservative (an unresolvable id fails
closed to `intimate`). The exception needs a stale index entry whose filename
has since been reused **plus** a second live file still carrying the original
id — reachable only through out-of-band vault manipulation, which is why #970
shipped without it.

It is still worth closing as a class, because it protects every caller of
`update_fragment`, not just the journal path, and because "the index says so"
is a weaker guarantee than "the file says so" for an operation that destroys
the previous contents.

## Suggested shape

Before `_atomic_write_text` in the update path, load the located file's
frontmatter and confirm `post.metadata.get("id") == fragment.id`. On mismatch,
treat it as an index miss: fall back to the directory scan (or repair the
index and re-resolve) rather than rewriting. The verification is one
`frontmatter.load` of a file that is about to be read/written anyway, so the
cost is small relative to the write.

Add a regression test that plants a stale `.id-index.jsonl` entry pointing at
a file carrying a different `id` and asserts `update_fragment` does not
clobber it.

Refs #970

Issue #1082, opened 2026-07-31
Remote consumers lose idempotent re-sync on journal entries escalated to intimate (post-#970)


Geoffe-Ga:
## Context

Raised by the security review of #970 (the `creek.journal` write-side ceiling
gate). #970 ships as designed — this is a follow-on operational question the
gate creates, not a defect in it.

The #970 gate refuses an update-in-place when the caller's ceiling would not
have admitted a *read* of the fragment being overwritten. Correct. But it sits
**above** the `record.content_hash != new_hash` comparison in
`write_fragment_idempotent`, so it refuses *unchanged idempotent re-sends*
exactly as it refuses genuine rewrites.

Combine that with two other facts already in the system:

- `creek classify`'s privacy pass is **escalate-only**, so a journal fragment
  can move up to `intimate` and never back down.
- Remote consumers are capped at `personal` by `_BoundedFastMCP.call_tool`
  (`docs/mcp.md` §"INTIMATE is never reachable remotely"), and cannot request
  `ceiling=intimate`/`all` at all.

Adepthood is a remote consumer and the primary journal producer. So once a
journal entry is classified `intimate`, Adepthood can never again send that
`external_id` — not to edit it, and not as a no-op re-sync in a continuous
ingest loop. Every such call returns the generic refusal.

The recovery wording shipped with #970 — "recovery is one re-send at a broader
ceiling" — is true only for a **local stdio** caller. #970 documents that
caveat; this issue is about whether to narrow the refusal.

## Possible shape (needs an architecture decision, not a quick patch)

Admit the call when the incoming content hashes equal to the ledger's recorded
`content_hash` for that `source_key`. The argument that this leaks nothing:
the comparison is against **the caller's own previously-submitted text**, so a
caller learns only whether the bytes it just sent match the bytes it sent
before — not anything about the fragment's current body or tier. And the
`unchanged` branch is a dedup no-op that rewrites nothing, so there is no
destruction to gate.

Things that must be settled before implementing:

1. **Is `unchanged` genuinely write-free?** Verify `run_ingest`'s unchanged
   path touches neither the fragment nor the staged file. Note the staged file
   is rewritten by `_stage_entry` *before* `run_ingest` sees it, so admitting
   the call still overwrites the staged copy — with byte-identical content in
   the matching case, but the frontmatter `date` field is regenerated when the
   caller omits `timestamp`, which would make it non-identical. That has to be
[...]

Issue #1081, opened 2026-07-31
creek.journal's #970 overwrite gate costs a full 01-Fragments walk per update (~3.4s at 35k)


Geoffe-Ga:
## Context

The #970 write-side ceiling gate for `creek.journal` has to know the current
vault tier of the fragment an `external_id` already maps to, before it lets
the caller overwrite it. It gets that through
`creek.classify.privacy_filter.source_tiers(vault_path, [fragment_id])`
reduced by `max_source_tier`.

`source_tiers` is the right *correctness* choice and should not be swapped out
casually. It walks `01-Fragments` through the shared
`creek.vault.reader.iter_vault_fragments` loader, so the gate inspects exactly
the files the rest of the pipeline treats as fragments (a bespoke scan would
see a different set — the bug class the shared survey exists to prevent), and
its walk deliberately never short-circuits, which is what stops the refusal
leaking *where* the protected fragment sits through timing.

The cost is that it is a **full `01-Fragments` walk on every `creek.journal`
call that resolves a ledger record** — i.e. every update and every idempotent
no-op re-send. Creations are unaffected: with no ledger record the gate passes
`content_tier=None` and never calls `source_tiers`.

## Measured

Synthetic vault, fragments under `01-Fragments/Journal`, timing
`max_source_tier(source_tiers(vault, [one_id]))`:

| fragments | wall time | per fragment |
|-----------|-----------|--------------|
| 500       | 0.040 s   | 79 us |
| 2 000     | 0.175 s   | 88 us |
| 8 000     | 0.773 s   | 97 us |

Linear, extrapolating to roughly **3.4 s per journal update at 35k
fragments** — a per-call cost that did not exist before #970. The 35k-fragment
vault is a real configuration (see the scale notes on the `creek fill` work),
and Adepthood is a continuous journal producer, so this is on the hot path
rather than a batch job.

## Suggested shape

`VaultWriter` already maintains a per-directory `.id-index.jsonl` mapping
`id -> filename` precisely so duplicate detection is O(1) rather than a
directory scan (`creek/vault/writer.py`, `INDEX_FILENAME`), and
`update_fragment` locates the target through it. The gate is doing a full walk
to answer a question the index can answer directly.

A public `find_fragment_path(vault_path, fragment_id) -> Path | None` backed by
that index would let the gate load exactly one file. Two properties must be
preserved when doing it:

1. **Fail-closed on divergence.** The index locator and the shared-loader
[...]

Issue #1080, opened 2026-07-31
PurgeEngine leaves a dangling SourceLedger record pointing at the purged fragment


Geoffe-Ga:
## Context

Found while implementing the #970 write-side ceiling gate for `creek.journal`.

`PurgeEngine` deletes a purged fragment's markdown file and — for journal
fragments — follows `source.origin_key` back to the staged plaintext under
`00-Creek-Meta/adepthood/journal/` and deletes that too (#845). What it never
does is prune the **source ledger** record that maps that staged path to the
now-deleted fragment id. `creek/purge/engine.py` does not import
`SourceLedger` at all; `origin_key` is used only to locate the staged file.

## Reproduced

```python
res = journal_ingest_tool(vault_path=v, content="an entry to be purged",
                          external_id="adep-purge", tier="intimate",
                          privacy_tier_ceiling=TierCeiling.ALL)
PurgeEngine(v).purge_fragment(res["fragment_id"])

SourceLedger.load(v, source="markdown")._records
# {'00-Creek-Meta/adepthood/journal/adep-purge-b88e772cf4cc.md':
#     LedgerRecord(fragment_id='frag-ee5c425f5355', tombed=False)}
list((v / "01-Fragments").rglob("*.md"))
# []
```

The record survives the purge, still points at `frag-ee5c425f5355`, and is
**not** marked `tombed`. `SourceLedger.active_keys()` therefore still reports
the source key as live.

## Why it matters now

On its own this was latent — `run_ingest` re-creates the fragment on the next
send, so nothing visibly broke. The #970 gate makes it observable: that gate
resolves `external_id` -> ledger record -> fragment tier before a write, and
an id that resolves to a record but *no* fragment fails closed to `intimate`
(the correct safe direction — a fragment the locator cannot see must not be
assumed low-tier). Consequence: after a purge, re-sending the same
`external_id` is refused for any caller below `ceiling=intimate`.

That is safe and recoverable (`ceiling=intimate`/`all` still works, which is
what `test_resend_after_purge_restages_cleanly` exercises), so #970 ships as
is. But the underlying dangling record is still wrong, and a remote consumer
— capped at `personal` by the transport boundary — cannot re-create any
journal entry the operator has purged.

## Suggested shape

`PurgeEngine` should prune (or `tombed`-mark) the `SourceLedger` record for
every purged fragment whose `source.origin_key` names a ledger-backed source,
in the same pass that deletes the staged plaintext. Decide deliberately
[...]

PR #1084, opened 2026-07-31
fix(mcp): gate creek.journal's update-in-place on the overwritten fragment's tier (#970)

### LINKED ISSUE #970: creek.journal update-in-place lets an unadmitted caller overwrite an above-ceiling fragment's body

Geoffe-Ga:
## Summary

- **Gates `creek.journal`'s idempotent update-in-place on the tier of the fragment it would destroy** (#970): you may only overwrite what you could have read. `write_tier_allowed` ranked only the *incoming* entry, so a caller at `ceiling=open` could replace the body of a fragment persisted at `privacy_tier: intimate` — content it was never admitted to read. `creek.purge.*` demands `CREEK_MCP_ELEVATED_TOKEN` for exactly this kind of destruction; this path required only the caller's own, lower ceiling.
- **Moves `_stage_entry` below the gate.** This is the half of the fix that is easy to miss: staging used to run *first*, and the staged copy under `00-Creek-Meta/adepthood/journal/` has no escalate-only ratchet — so a gate bolted on after it returns a correct refusal over an already-destroyed staged entry whose `privacy_tier` has been rewritten *downward*. Reproduced before the fix; pinned by a byte-level test after it.
- **Flips `creek.journal`'s read-gate posture from `UNGATED`-with-`gap_issue=970` to `GATED`**, making it the second production adopter of `read_gate.refuse_above_ceiling` and the first that is a *write* gate — the question it asks ("could this caller have read what it is about to destroy?") is a read question, hence `tier_allowed` via the primitive rather than `write_tier_allowed`, which still gates the incoming tier above it.

### Design notes worth a reviewer's attention

**The gate keys on the existing fragment's CURRENT vault tier** — via the shared `source_tiers` + `max_source_tier` survey — never on the caller's declared `tier` and never on the staged copy's stamp. Both of those are stale the moment `creek classify` escalates a fragment, and a caller-supplied tier is not a gate at all. `test_gate_reads_the_vault_tier_not_the_staged_stamp` is the discriminator: it escalates the fragment on disk only, leaving the staged stamp and the caller's declaration both reading `open`.

**Two fail-closed rules, deliberately distinct — collapsing them breaks one.** No ledger record at all → `content_tier=None` → **admit**, because that is a *creation* and creation must keep working at every ceiling. A ledger record whose fragment does not resolve (purged, deleted out of band, schema-invalid, or tombed into `10-Liminal/Orphaned` where the `01-Fragments` walk cannot see it) → `max_source_tier([])` → `intimate` → **refuse**. Every divergence between this gate's locator and the writer's own id index therefore fails in the safe direction.

**Nothing becomes permanently un-updatable.** The `privacy_tier` ratchet is escalate-only in this codebase, so a fix that could bury an entry would be worse than the bug. No refusal is persisted; an `intimate` entry stays updatable at `ceiling=intimate`/`all`, proved by a parametrised test rather than reasoned about from the comparator.

**The refusal names nothing.** It carries `GENERIC_ABOVE_CEILING_REASON` and the four-key payload — no fragment id, no tier — because a refusal that ranks the content it withheld is a tier-classification oracle. The refusal audit carries the caller's own arguments only: no `created_tier`, no `affected_fragment_ids`. The accepted residual (an existence-*and-rank* oracle, deliberately blurred against the unresolvable cases) is written up in `journal.py`'s docstring in the style of `reflect.py`'s block.

**Recoverability of an already-clobbered vault: there is none, and the docs now say so.** The overwritten body survives nowhere — provenance records hashes and paths, not content. `docs/mcp.md` documents the operator path (re-send from the Adepthood-held original via a local stdio caller, vault backup, audit-log forensics) and states plainly that no pre-image backup is taken, deliberately: it would mint a second plaintext copy of intimate content.

### Verification is on the bytes, not the response

The read-gate guardrail's layer (f) asserts on the JSON response and is **structurally incapable** of catching this class — on #968 it passed 5/5 while all six report generators leaked to disk. `creek.journal`'s refusal envelope is canary-free whether or not the ceiling is enforced, so its layer-(f) entry passes *vacuously*. It is still added (as a tripwire against a future response shape that carries content), but the real evidence is a companion test asserting the fragment's bytes are unchanged, and its docstring says exactly that.

## Test plan

- `./scripts/check-all.sh` → **12/12 checks passed**, 6815 tests, **94.05%** branch coverage, per-file gate green (219 files ≥80%).
- `tests/test_mcp_journal.py` + `tests/test_mcp_read_gate.py` → **156 passed**.
- pylint **9.69** (≥9.0), xenon `--max-absolute B` clean, interrogate **97.7%** (≥95%).
- Gate 1 RED was verified failing first: 8 failed / 14 passed, with the two byte-level tests failing on **disk-byte inequality** (the staged-file diff landing at the `privacy_tier: intimate` → `open` frontmatter downgrade), not on response shape.

New tests (9 for the gate, plus a malformed-input guard and the read-gate probe pair):

| test | what it pins |
|---|---|
| `test_update_of_an_above_ceiling_fragment_is_refused` | the issue's verbatim reproduction; generic reason, no tier echoed |
| `test_refused_update_leaves_the_fragment_bytes_untouched` | `read_bytes()` equality + the caller's text in **no** `.md` anywhere |
| `test_refused_update_leaves_the_staged_entry_untouched` | fails if the gate is added without moving `_stage_entry` |
| `test_refused_update_is_audited_without_naming_the_fragment` | audit carries caller args only, by **exact equality** |
| `test_creating_a_new_entry_at_a_low_ceiling_still_works` | legitimate creation does not regress |
| `test_within_ceiling_update_still_works` | ordinary edits do not regress |
| `test_intimate_entry_stays_updatable_at_an_admitted_ceiling` | recoverability at `intimate`/`all`, and the tier stamp survives |
| `test_gate_reads_the_vault_tier_not_the_staged_stamp` | rejects any impl reading the caller's tier or the stale stamp |
| `test_a_ledger_record_whose_fragment_is_gone_fails_closed` | fail-closed **and** its recovery at `ceiling=all` |
| `test_overwrite_admission_tracks_the_ranking_at_every_tier` | the middle of the ranking, incl. `unclassified` (#961) |
| `test_journal_probe_refuses_and_leaves_the_fragment_bytes_untouched` | layer (f)'s vacuity for this tool, with bytes as the real evidence |

`test_resend_after_purge_restages_cleanly` is unmodified and still passes.

**Mutation-tested.** Neutering the gate (forcing it to resolve no existing
fragment) turns **11** of these tests red, including all three refusal legs of
the ranking table — so the suite fails for the right reason rather than
passing by construction.

### Manual verification

Reproduced the original bug and confirmed the overwritten intimate body survives nowhere in the vault; confirmed the staged copy was clobbered *and* tier-downgraded; measured the gate's cost (linear, ~3.4s at 35k fragments — filed, not fixed); and confirmed empirically that an unchanged idempotent re-send at the remote cap (`ceiling=personal`) is refused once a fragment is escalated to `intimate`, while local `ceiling=all` still succeeds.

## Process notes

Planned by an independent `chief-architect` pass; built by `test-specialist` → `implementation-specialist`; reviewed by `security-specialist` (no BLOCKER, no MAJOR) and `code-review-orchestrator` (**CLEAN**). Three of those specialists ran without a shell in their session and could not execute their own verification, so **every result quoted above was run and observed by the conductor**, not taken on report.

The one substantive review nit — no coverage of the *middle* of the tier ranking — was closed by adding `test_overwrite_admission_tracks_the_ranking_at_every_tier`, written by the conductor rather than a further specialist pass.

Two prose claims this PR originally introduced were found overstated by the security review and corrected before push: "recovery is one broader call" (true only for a *local stdio* caller — remote consumers are capped at `personal`) and "strictly weaker" of the residual oracle (it is no stronger on the *existence* question, but the rank bit is new).

## Follow-ups filed (not fixed here)

- #1080 — `PurgeEngine` leaves a dangling `SourceLedger` record pointing at the purged fragment. Latent before; this gate makes it observable, and it fails closed.
- #1081 — this gate costs a full `01-Fragments` walk per update (~3.4s at 35k, measured). The obvious speedup must not reintroduce the positional timing channel the non-short-circuiting walk exists to prevent.
- #1082 — remote consumers lose idempotent re-sync on entries escalated to `intimate`. Correct security behaviour; whether to carve out the unchanged-content case is a design decision.
- #1083 — `VaultWriter.update_fragment` rewrites the id-index's target without verifying the located file's own `id` (pre-existing; defense-in-depth).

Closes #970
Refs #932, #848, #754, #969, #968

### DIFF
--- a/creek-tools/creek_mcp/read_gate.py
+++ b/creek-tools/creek_mcp/read_gate.py
@@ -36,12 +36,20 @@ of the primitives below inherits it:
    to key a model call.
 
 **Who calls these primitives, and why the rest still do not.**
-:func:`refuse_above_ceiling` has exactly one production caller:
-``creek_mcp.tools.state_read``, which adopted it in #969. That adoption is the
+:func:`refuse_above_ceiling` has two production callers.
+``creek_mcp.tools.state_read`` adopted it in #969, and that adoption is the
 shape the primitives were written for — the tool addresses a *single* cached
 artifact, so there is nothing to partially admit, and its refusal has no
 tool-specific story to tell. It gets the generic reason, the four-key payload
 and the no-tier-echo rule for free rather than re-deriving any of them.
+``creek_mcp.tools.journal`` adopted it in #970 on the same property one step
+further out: it is a *write* gate asking a read question. Its idempotent
+update-in-place would overwrite the one fragment an ``external_id`` resolves
+to, and the rule is that you may only overwrite what you could have read — so
+admission is decided by ``tier_allowed`` through this primitive, not by
+``write_tier_allowed``, which ranks the incoming entry and knows nothing about
+the fragment on disk. Caller-addressed and singular again: you cannot
+overwrite half a body, so there is nothing to partially admit.
 
 The other refusing tools are deliberately *not* retrofitted.
 ``creek.reflect`` and ``creek.compile`` keep their own refusal reasons and
@@ -159,8 +167,9 @@ class ToolPosture:
         gate_symbol: Name of the gate callable within *gate_module*. Set on
             ``GATED`` entries only.
         gap_issue: Issue tracking an unenforced ceiling — required on every
-            ``UNGATED_KNOWN_GAP`` entry, and also carried by ``creek.journal``
-            whose gap is on its update-in-place path rather than a read.
+            ``UNGATED_KNOWN_GAP`` entry, and set on nothing else. It was also
+            carried by ``creek.journal`` until #970 gated its update-in-place
+            path; that entry is now ``GATED`` and names its gate instead.
     """
 
     posture: ReadPosture
@@ -347,14 +356,44 @@ TOOL_POSTURES: dict[str, ToolPosture] = {
         gap_issue=971,
     ),
     "creek.journal": ToolPosture(
-        posture=ReadPosture.NO_UNSUPPLIED_READ,
+        posture=ReadPosture.GATED,
         rationale=(
-            "The entry body is the caller's own, and write_tier_allowed gates "
-            "the tier it creates; but the idempotent update-in-place "
-            "overwrites the fragment an external_id already maps to without "
-            "consulting that fragment's tier (#970)."
+            "The entry body is the caller's own and write_tier_allowed gates "
+            "the tier it creates, but the idempotent update-in-place destroys "
+            "the fragment an external_id already maps to — so it refuses on "
+            "THAT fragment's current vault tier: you may only overwrite what "
+            "you could have read (#970). A write gate asking a read question, "
+            "hence refuse_above_ceiling (tier_allowed) rather than "
+            "write_tier_allowed; the target is caller-addressed and singular, "
+            "and you cannot overwrite half a body. Ordering is load-bearing: "
+            "the gate sits ABOVE _stage_entry, because the staged copy under "
+            "00-Creek-Meta/adepthood/journal/ has no escalate-only ratchet, so "
+            "a gate placed below staging would return a correct refusal over "
+            "an already-destroyed staged entry whose privacy_tier had been "
+            "rewritten downward. Two fail-closed rules, not one: no ledger "
+            "record at all passes content_tier=None and CREATES (creation must "
+            "keep working at every ceiling), while a ledger record whose "
+            "fragment does not resolve reduces max_source_tier([]) to intimate "
+            "and is refused — so any divergence from the writer's own id index "
+            "fails safe. Consequence to know: PurgeEngine leaves a dangling "
+            "ledger record (#1080), so a purged id is refused below "
+            "ceiling=intimate until it is re-sent by an admitted caller — a "
+            "LOCAL stdio caller only, since a remote consumer token is capped "
+            "at ceiling=personal and can never send ceiling=intimate/all "
+            "(#1082 tracks a possible content-hash carve-out for the "
+            "unchanged-resend case). Accepted residual: the refusal is an "
+            "existence-AND-rank oracle, matching reflect's own honesty "
+            "standard for its analogous oracle. No stronger than the "
+            "pre-existing action: created|updated bit on the existence "
+            "question those two share; the rank bit — 'above your ceiling' — "
+            "is new, and is the price of refusing at all. It is deliberately "
+            "blurred, not a clean tier read: every fail-closed unresolvable "
+            "case (purged, orphaned, schema-invalid, deleted out of band) "
+            "collapses into the identical refusal as a genuine above-ceiling "
+            "fragment, so refused means 'above your ceiling OR unresolvable'."
         ),
-        gap_issue=970,
+        gate_module="creek_mcp.tools.journal",
+        gate_symbol="refuse_above_ceiling",
     ),
     "creek.lint": ToolPosture(
         posture=ReadPosture.METADATA_ONLY,
--- a/creek-tools/creek_mcp/tools/journal.py
+++ b/creek-tools/creek_mcp/tools/journal.py
@@ -18,15 +18,99 @@ frontmatter (which the markdown ingestor carries onto the fragment), and the
 write-side ceiling refuses an entry whose tier exceeds the caller's ceiling — so
 an INTIMATE entry is stored intimate and never admitted under a lower ceiling.
 
-The idempotent update-in-place has a gap this guarantee does not cover
-(#970): overwriting the fragment an existing ``external_id`` already maps to
-replaces whatever body is there **without consulting that fragment's own
-tier**. Reproduced: a caller at ``ceiling=open`` replaced the body of a
-fragment persisted at ``privacy_tier: intimate``. The persisted tier itself
-was not downgraded — the escalate-only privacy merge held — but the intimate
-body was destroyed. Contrast ``creek.purge.*``, which requires
-``CREEK_MCP_ELEVATED_TOKEN`` for exactly this kind of destruction; this path
-requires only the caller's own (lower) ceiling.
+**The update-in-place path is gated too, on the tier of what it would destroy
+(#970): you may only overwrite what you could have read.** The write-side
+``write_tier_allowed`` check ranks the *incoming* entry and says nothing about
+the fragment an existing ``external_id`` already maps to, so a caller at
+``ceiling=open`` used to replace the body of a fragment persisted at
+``privacy_tier: intimate``. The persisted tier survived (the escalate-only
+privacy merge held) but the intimate body did not. Contrast ``creek.purge.*``,
+which requires ``CREEK_MCP_ELEVATED_TOKEN`` for exactly this kind of
+destruction; this path requires only the caller's own (lower) ceiling.
+:func:`_refuse_unadmitted_overwrite` closes that, and the question it asks —
+"could this caller have *read* the thing it is about to destroy?" — is a read
+question, which is why it goes through
+:func:`creek_mcp.read_gate.refuse_above_ceiling` (and so ``tier_allowed``)
+rather than through ``write_tier_allowed``.
+
+Gate ordering, all of it load-bearing:
+
+1. malformed calls (blank ``content``/``external_id``, unknown ``tier``) refuse
+   without an audit entry — there is no meaningful tier to record yet;
+2. ``write_tier_allowed`` on the *incoming* tier — audited, then refused;
+3. the overwrite gate on the tier of the *resolved existing* fragment —
+   audited, then refused;
+4. **only then** :func:`_stage_entry`, which before #970 ran first. Moving it
+   below the gate is the half of this fix that is easy to miss: the staged
+   copy under ``00-Creek-Meta/adepthood/journal/`` has no escalate-only
+   ratchet, so a gate added *after* the staging write returns a correct
+   refusal over an already-destroyed staged entry — body replaced and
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
Closes #970: `creek.journal`'s idempotent update-in-place previously overwrote whatever fragment an `external_id` already mapped to without checking that fragment's own privacy tier, letting a low-ceiling caller destroy `intimate` content. This PR adds a second gate — reusing the `read_gate.refuse_above_ceiling`/`tier_allowed` primitive rather than the write-side `write_tier_allowed` — keyed on the *existing* fragment's current vault tier, and moves `_stage_entry` below both gates so a refusal no longer clobbers the staged copy either.

## Strengths
- Correctly distinguishes the two fail-closed rules: no ledger record → `content_tier=None` → admitted (this is creation, must keep working at every ceiling); a ledger record whose fragment doesn't resolve → `max_source_tier([])` → `INTIMATE` → refused. Verified in `_existing_tier`/`_refuse_unadmitted_overwrite` (`creek-tools/creek_mcp/tools/journal.py:208-270`) that this matches the described contract.
- The reordering of `_stage_entry` below the new gate is real and load-bearing — confirmed the old code staged first; the new code only calls `_stage_entry` after both `write_tier_allowed` and `_refuse_unadmitted_overwrite` pass (`journal.py:385-413`). `_staged_path` was correctly split out as a pure path-compute helper so the gate can derive the ledger key before writing anything.
- Reuses the shared `source_tiers` walk, which is documented (and here relied upon) to never short-circuit — so the refusal cannot leak *which* fragment is protected via timing. This property is inherited correctly rather than re-implemented.
- Test suite verifies byte-level disk state, not just the JSON response shape — directly addressing the #968-style blind spot where a response-level guardrail can pass while a write-side leak still occurs. `test_refused_update_leaves_the_fragment_bytes_untouched` and `test_refused_update_leaves_the_staged_entry_untouched` are exactly the right shape of regression test for this bug class.
- `test_gate_reads_the_vault_tier_not_the_staged_stamp` is a strong discriminator test — it rules out an implementation that reads the caller-declared tier or the stale staged stamp instead of the live vault tier.
- `test_overwrite_admission_tracks_the_ranking_at_every_tier` walks the full tier×ceiling table (including the `unclassified` ranks-with-`personal` case from #961), not just the endpoints.
- Manifest (`read_gate.py` `TOOL_POSTURES`) and its pinning tests (`test_mcp_read_gate.py`) are updated in lockstep with the code change, keeping the posture claim and the actual gate from drifting apart.
- Docs (`docs/mcp.md`) transparently document the operational consequence (a remote consumer capped at `personal` can never re-send an `intimate`-escalated `external_id`, not even unchanged) and the recovery path, including that no pre-image backup is taken (correctly reasoned: a backup would itself mint a second plaintext copy of intimate content).
- Honest disclosure of out-of-scope follow-ups (#1080–#1083) rather than silently absorbing or ignoring adjacent issues surfaced during the fix.

## Security Concerns
🟢 LOW — none found. The core security property (only admit an overwrite the caller could have read) is correctly enforced, reads the live vault tier rather than a caller-supplied or staged-stale tier, fails closed on an unresolvable ledger target, and the refusal payload was verified (by test and by reading `refuse_above_ceiling`) to carry only the generic four-key response with no tier or fragment id echoed.

## Problems
None blocking.

## Code Quality
- The module docstring and several function docstrings in `journal.py` are extremely long (100+ lines for the module docstring). This matches the established house style elsewhere in this codebase (e.g. `reflect.py`), so it's a style note rather than an objection, but a future reader skimming for the gate's actual behavior has to wade through a lot of prose-as-documentation to find the four-line `_refuse_unadmitted_overwrite` body.
- `_validated_entry_tier`'s `PrivacyTier | dict[str, Any]` return union works with the `isinstance` narrowing at the call site, but is a slightly unusual signature shape; a small `Refused | PrivacyTier` sentinel type could read more explicitly. Not worth blocking on.

## CLAUDE.md Adherence
✅ TDD followed (PR description documents Gate 1 RED verified failing first, on disk-byte inequality specifically). ✅ `./scripts/check-all.sh` reported green with coverage/complexity/docstring thresholds met per the PR description. ✅ Follows the repo's established gate-primitive reuse pattern (`refuse_above_ceiling`) rather than inventing a parallel mechanism. ✅ Docs updated (`docs/mcp.md`) to match implementation. ✅ Follow-up issues filed for out-of-scope findings rather than silently deferred.

## Testing
Strong. 11 new tests cover: the issue's verbatim repro, byte-level fragment and staged-entry integrity across a refusal, audit-log content (present but silent on the protected fragment), non-regression of plain creation and same-tier edit-in-place, recoverability at an admitted ceiling (guarding against a "permanently un-updatable" regression), the vault-tier-vs-stale-stamp discriminator, the full tier-ranking table, the dangling-ledger-record fail-closed-then-recoverable case, and a runtime canary probe pinning that the shared response-level guardrail is vacuous for this tool while the byte-level check is not. Mutation testing (11/11 targeted tests turn red when the gate is neutered) is a meaningful extra signal that the tests fail for the right reason.

## Documentation
Thorough and accurate against the implementation — verified the `docs/mcp.md` table update and new prose sections correctly describe gate ordering, the two fail-closed branches, and the remote-vs-local recovery asymmetry, all of which match what the code actually does.



## Verdict: LGTM

The fix is correct (verified `_refuse_unadmitted_overwrite`/`_existing_tier` reads the vault-current tier via the shared `source_tiers`/`max_source_tier` primitives, and the gate now sits above `_stage_entry`), the fail-closed/fail-open split (no-ledger-record → creation vs. unresolvable-record → refuse) is sound, and the test suite includes byte-level evidence (not just response-shape assertions) plus a parametrised ranking-table test and a runtime canary probe. No blocking issues found; nothing here needs to be fixed before merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5140721460 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1089, opened 2026-07-31
resolve_within_vault raises an unhandled ValueError out of the MCP boundary on a NUL byte in a caller-supplied path


Geoffe-Ga:
## Role
You are a senior Python engineer working in this project's codebase, following
its existing conventions (TDD via stay-green, the `cd creek-tools &&
./scripts/check-all.sh` gate, >=90% branch coverage aggregate and >=80% per file,
>=95% docstring coverage, complexity <=10, mypy strict, zero lint/type
suppressions).

## Goal
Guard the `candidate.resolve()` call in
`creek_mcp/path_confinement.py::resolve_within_vault` so a caller-supplied path
that cannot be resolved at all (a NUL byte, and any other `ValueError`/`OSError`
shape) returns `None` — becoming the ordinary outside-the-vault refusal callers
already handle — instead of raising out of the MCP boundary. Verified by an
executed test asserting `creek.redact.scan` and `creek.ingest` return
`status="refused"` on a NUL-byte `input_path` and raise nothing.

## Context

Found by the Gate-2.5 security review of #972 and left out of that PR's scope,
because it is pre-existing and **shared** rather than specific to
`creek.redact.scan`.

- File(s): `creek-tools/creek_mcp/path_confinement.py:28-56` — specifically the
  unguarded `resolved = candidate.resolve()` at **line 51**.
- Scanned at commit: `9584959` (re-verified by the grooming pass 2026-08-10) —
  re-verify against HEAD before starting.
- Evidence: the existing `try`/`except ValueError` at lines 52-55 wraps only
  `resolved.relative_to(vault_resolved)`. The `resolve()` on line 51 sits
  **outside** it, so a NUL byte raises straight through:

  ```
  ValueError: lstat: embedded null character in path
  ```

  `input_path` arrives as a pydantic `str`, and JSON string escapes permit a NUL
  codepoint, so this is reachable from the wire. The exception escapes **after**
  the audit entry is written and **before** any refusal, so the caller gets a
  transport-level error where every other bad path gets `status="refused"`.
  `creek.ingest` takes caller input through the same helper and has the same
  behaviour.

  **Not a disclosure.** The message is `"embedded null character in path"` — it
  echoes nothing about the vault. This is an error-shape / robustness finding.
- Related: #972 (fixed the same class one layer up — `_vault_relative`'s
  unhandled `ValueError` on a vault root with a symlinked component, which is how
[...]

Issue #1088, opened 2026-07-31
CrawDad's configurable attachments.staging_subpath can fall outside creek.redact.scan's canonical scope, silently disabling the safety pass


Geoffe-Ga:
## Context

Found while fixing #972, which confines `creek.redact.scan` to the canonical
FEAT-027 staging subtree `00-Creek-Meta/Inbound/`: that subtree is admitted at
**every** ceiling, and any other vault path is ranked as if it held `intimate`
content (the scan reads no per-file tiers) and so needs `ceiling=intimate`/`all`.

## The gap

CrawDad's staging root is configurable — `AttachmentConfig.staging_subpath`
(`crawdad/crawdad/config.py:246`), default `00-Creek-Meta/Inbound`. Its
validator refuses only absolute paths and `..` segments, so **any** other
vault-relative path is accepted, and `crawdad/crawdad/bot.py::_run_safety_scan`
sends exactly that path to `creek.redact.scan` with the channel's tier as the
ceiling.

Two consequences after #972:

1. **Availability.** A deployment that overrides `staging_subpath` to anything
   outside `00-Creek-Meta/Inbound/` gets `status="refused"` on every safety scan
   from an `open`/`personal` channel. The bot degrades gracefully (the user still
   gets the download summary), but the safety pass silently stops running.
2. **Confinement.** `staging_subpath: 01-Fragments` is accepted today. Before
   #972 that made CrawDad itself the leak vector; after #972 it is refused at
   `open`/`personal` — which is the right outcome, but it is enforced on the
   creek-tools side by accident of scope rather than by CrawDad refusing an
   obviously wrong configuration.

## Suggested shape

Validate `staging_subpath` in `AttachmentConfig` against the canonical staging
root: accept `00-Creek-Meta/Inbound` and its descendants, refuse anything else
at config-parse time with a message naming #972's scope rule. Prefer refusing
early over a runtime warning — a misconfigured staging root means the safety
scan is not running, and that must not be discoverable only by reading logs.

Note the constant would then exist on both sides. `bot.py:71` already duplicates
`creek_mcp.tools.redact.TOOL_NAME` with a comment saying why; follow that
precedent rather than adding a cross-repo import.

## Acceptance criteria

- Executed test: `staging_subpath` outside `00-Creek-Meta/Inbound` raises at
  config-parse time with a message naming the canonical root.
- The default config is unchanged and every existing attachment test stays green.
[...]

Issue #1087, opened 2026-07-31
RedactionScanner.scan_batch follows symlinked files out of the scan root (SEC-003 guards --apply/--review, not the scan walk)


Geoffe-Ga:
## Context

Found while fixing #972 (`creek.redact.scan` tier scoping), and deliberately
left out of that PR's scope.

## The gap

`RedactionScanner.scan_batch` walks its root with `dir_path.rglob("*")` and
keeps every `child.is_file()`. `is_file()` follows symlinks, so a **symlinked
file** inside the scan root is opened and read — even when its target lies
outside the scan root, or outside the vault entirely.

Reproduced (executed) against `creek_mcp.tools.redact.redact_scan_tool`:

```
00-Creek-Meta/Inbound/ch1/sneaky.md -> 01-Fragments/Notes/my-affair-with-dana.md
```

A scan of `00-Creek-Meta/Inbound` at `privacy_tier_ceiling=open` reads the
intimate fragment's body and reports which PII types it contains.

`creek redact --apply` and `--review` both carry an explicit symlink-escape
guard (SEC-003, `creek/redact/cli_commands.py`). The **scan walk has none**, so
the guard protects the write path and not the read path that feeds it.

## What #972 does and does not fix

#972 makes the MCP response render every finding at its **as-scanned** path
(`00-Creek-Meta/Inbound/ch1/sneaky.md`), never its resolved target, so the
symlink no longer discloses the intimate fragment's slugified title. What it
does **not** fix is that the file was read at all: the caller still learns
*which PII types and line numbers* the out-of-scope target contains. That
residual is recorded in `creek_mcp/read_gate.py`'s posture rationale against
this issue.

## Suggested shape

Skip (or refuse) symlinked children in `scan_batch` whose resolved target
escapes the scan root, mirroring the existing SEC-003 guard rather than
inventing a second policy. Decide explicitly whether an in-root symlink is
skipped silently or surfaced as a skipped-file statistic — a silent skip in a
*safety* scanner is its own hazard.

## Acceptance criteria

- Executed test: a symlink under the scan root pointing outside it produces no
  finding, at any ceiling and through both the MCP tool and `creek redact --scan`.
- The CLI `--apply`/`--review` SEC-003 behaviour is unchanged.
- `docs/security/threat-model.md`'s SEC-003 entry states which walks the guard
  covers.

Refs #972, SEC-003

Geoffe-Ga:
[...]

Issue #1079, opened 2026-07-31
Two tier readers disagree about the same 10-Liminal note (state report vs mining corpus)


Geoffe-Ga:
Found during the Gate-2.5 review of #969 (PR #1078). Not a leak today — filed because the divergence is the bug class this codebase deliberately pins against elsewhere.

## The split

The same physical file — a `10-Liminal/Unnamed/` note with **no `privacy_tier` key at all** — is admitted at two different ceilings inside one rendered document:

- **`## Liminal Watch`** reads it through `_admitted_liminal_notes`, which uses `within_ceiling` on the note's **raw frontmatter**. A missing key fails closed to `INTIMATE`, so the note needs `ceiling=intimate` to appear.
- **`## Suggested questions`** reads it through the mining corpus, whose `_load_liminal_fragments` builds a validated `Fragment`. A missing key takes the model's `unclassified` default, which ranks with `personal` (#876/#961), so the note's opaque id reaches a mined prompt at `ceiling=personal`.

So one note needs `intimate` for its stem and only `personal` for its id, in the same report.

## Why it is not a leak today

The only liminal field a prompt renders is the fragment's generated 12-hex id (`_seed_from_liminal`), never its title or body. And #1078 folds those notes' tiers into the artifact stamp, so an untiered one pushes the stamp to `unclassified` — refused at `ceiling=open`. The stamp does not under-report. Verified end to end in the #1078 re-review.

## Why it should still be closed

`creek/classify/privacy_filter.py`'s module docstring is explicit that a second tier reader is "a tier opinion the others can drift from", and `tests/test_mcp_report_tier_ceiling.py::test_raw_and_model_tier_readers_agree_on_every_fragment` exists precisely to stop `raw_privacy_tier` and `fragment_tier` diverging. This is that divergence, reached through a different loader. It is exactly the #1033 fail-open shape: reading a tier off the validated model rather than the raw frontmatter.

## Suggested fix

Give `creek.generate.mining._load_liminal_fragments` the raw-frontmatter reader (`raw_privacy_tier`/`within_ceiling`) that the report side already uses, so both halves of one document agree about one file. Note this changes `creek.mine`'s corpus too, which is why #1078 did not do it inline — it needs its own before/after on that tool's suite.

Two smaller adjacencies worth folding in:

[...]

PR #1090, opened 2026-07-31
fix(mcp): scope creek.redact.scan to the staging subtree and render paths as scanned (#972)

### LINKED ISSUE #972: creek.redact.scan returns intimate fragment filenames (slugified titles) to an open-ceiling caller, plus absolute paths in report_markdown

Geoffe-Ga:
## Summary

- **Scopes `creek.redact.scan` instead of tier-filtering it.** `00-Creek-Meta/Inbound/` — the FEAT-027 staging subtree — is admitted at **every** ceiling; every other in-vault path is ranked as if it held `intimate` content, because the scan reads no per-file privacy tier at all, and so needs `ceiling=intimate`/`all`. The gate sits **above** the existence check, so `refused` vs `not found` is no longer an existence oracle over slugified fragment titles.
- **Gives one fail-closed helper ownership of every path in the response** (`input_path`, `findings[].file_path`, `report_markdown`), rendering each **as scanned** and vault-relative. It deliberately never calls `.resolve()` on a match path.
- **Moves the read-gate posture from `UNGATED_KNOWN_GAP`(#972) to `GATED`**, adopting neither canonical primitive, and adds the runtime canary probe that forcing function requires.

## The leak, and why the obvious fix was the wrong one

At `privacy_tier_ceiling=open` a caller could aim the scan at `01-Fragments` and get back intimate fragments' **filenames**. Creek fragment filenames are slugified titles, so the filename *is* the content — the tool's own defence ("no matched text leaves the tool") was true and insufficient. Reproduced by execution before the fix:

```
"file_path": "01-Fragments/Notes/my-affair-with-dana.md", "match_type": "email"
### `/private/var/…/vault/01-Fragments/Notes/my-affair-with-dana.md`
refused | input_path not found: 01-Fragments/Notes/my-secret-other-thing.md
```

Three leaks in that output: the slugified title, an absolute path leaking the operator's home directory (CrawDad posts `report_markdown` verbatim into a Discord channel), and an existence oracle — a caller could probe *any* guessed fragment title and read the answer off `refused` vs `ok`.

**A per-file tier filter was rejected** in favour of the issue's preferred scoping, for four reasons:

1. Files under `Inbound/` are pre-ingest raw attachments with **no `privacy_tier` frontmatter at all**, so any fail-closed per-file reader ranks them `intimate` and refuses the entire FEAT-027 use case. You would carve `Inbound/` out anyway — scoping by a longer road, plus a fourth tier reader while #1079 already tracks two existing ones disagreeing.
2. Exclusion leaves the `files_scanned` differential oracle the issue names.
3. A silently filtered scan under-reports, so a *redaction* tool would be lying about comprehensiveness.
4. The refusal now shipped is derived **only** from the caller's own `input_path` and their own declared ceiling — never from vault content — so it echoes no tier and is neither an existence nor a rank oracle.

## Scoping alone would not have closed it

`_finding_to_dict` resolved each match path before relativising, and `scan_batch`'s `rglob` yields symlinked children **unresolved**. So a staged symlink inside the admitted subtree rendered its target's slugified title. Reproduced at `ceiling=open` before the fix, and this is why the second half of the change is not path hygiene:

```
00-Creek-Meta/Inbound/ch1/sneaky.md -> 01-Fragments/Notes/my-affair-with-dana.md
  before: "file_path": "01-Fragments/Notes/my-affair-with-dana.md"
  after:  "file_path": "00-Creek-Meta/Inbound/ch1/sneaky.md"
```

## Why not `refuse_above_ceiling`

`GENERIC_ABOVE_CEILING_REASON` — "resolved content exceeds the declared tier ceiling" — asserts that content was ranked. This gate ranks no content; it ranks the caller's own requested path. Adopting the primitive would have shipped a false statement and told a CrawDad operator nothing actionable. Same call #968 made for `creek.report`: what the manifest checks is that the named gate exists and decides something, not which of two shapes it takes.

One consequence for reviewers: **the issue's acceptance criterion that "layer (e) will fail until the posture moves" does not hold as written.** Layer (e) only fires when a *gap-labelled* tool calls a canonical primitive, and this fix calls neither. What actually failed until the posture moved is `test_pinned_gaps_keep_their_posture_and_issue`; after the move, layers (c) and (f) check the claim harder than (e) would have.

## Accepted residuals (both recorded in the manifest rationale)

- Existence probing **within** `Inbound/` still works — that is the tool's job.
- `RedactionScanner.scan_batch` still follows symlinked **files**, so an out-of-scope target's PII types, line numbers and its `files_scanned` bit are still disclosed even though its path is not. Bounded — `rglob` does not descend into symlinked *directories*. Tracked by #1087.

## Out of scope, filed rather than fixed

- **#1087** — `scan_batch` follows symlinked files out of the scan root; the SEC-003 guard covers `--apply`/`--review`, not the scan walk.
- **#1088** — CrawDad's configurable `attachments.staging_subpath` can fall outside this tool's canonical scope, silently disabling the safety pass.
- **#1089** — a NUL byte in a caller-supplied path raises an unhandled `ValueError` out of the shared `resolve_within_vault` (pre-existing; `creek.ingest` shares it).

Also noted, not filed: `generate_markdown_summary` renders `` ### `{path}` ``, so a filename containing a backtick or newline could break the Discord code span. Not reachable through the one production path — CrawDad sanitises staged names to `[A-Za-z0-9._-]` — so no issue was opened.

## Test plan

Gate 1 was genuine RED→GREEN, re-demonstrated after an interruption by stashing only the production files: the two test modules then fail at collection, and the pre-fix behaviour reproduces (slugified title in `findings`, absolute path in `report_markdown`, `not found` existence oracle). Restored → 167 pass.

- `cd creek-tools && ./scripts/check-all.sh` → **12 passed / 0 failed**. `creek_mcp/tools/redact.py` at **100%** coverage.
  (Locally, 5 `consent-missing` tests fail unless `CREEK_CLOUD_CONSENT` is stripped from the environment — confirmed unrelated: they pass with the operator's real credential vars unset, both before and after this change.)
- `pytest tests/test_mcp_redact.py tests/test_mcp_read_gate.py` → **167 passed**.
- **Executed by hand, not read off a comparator:**
  - `01-Fragments` @ `open` → the canonical four-key refusal, no `findings`, no `statistics`.
  - Out-of-scope paths that exist, do not exist, and are symlinks resolving off the vault → **byte-identical** responses.
  - `00-Creek-Meta/Inbound` → `ok` with findings at **all four** ceilings (the issue's recoverability criterion).
  - `01-Fragments` @ `all` → `ok`, renders the intimate filename, so the `open` refusal withholds something real.
  - No absolute path in any response field, at any ceiling.
- New tests worth naming: `test_out_of_scope_refusals_are_indistinguishable_from_an_outward_symlink`, `test_scan_refuses_out_of_scope_paths_identically_whether_or_not_they_exist`, `test_a_symlinked_child_renders_as_its_staged_path_not_its_target`, `test_findings_and_report_markdown_render_the_same_paths` (the "one code path owns it" invariant — the #846/#848 signature, machine-checked), `test_scan_admits_the_staging_subtree_at_every_ceiling`, and layer (f)'s `test_redact_scan_probe_refuses_and_is_not_vacuous`.
- The layer (f) probe seeds its canary in a **filename**: `canary_vault` takes its file stems from `frag_id` and its fragments carry no PII, so the shared fixture is structurally blind to this leak class. The probe's anti-vacuity control is the same scan at `ceiling=all` returning the canary.

## Process notes

- Plan by `chief-architect`; specialists dispatched per its list (test → implementation → security → documentation). No spawn failed, so every pass named here really happened.
- A `security-specialist` pass at Gate 2.5 found one real oracle — a distinguishable "resolves outside the vault root" message that told an un-admitted caller *"there is an outward symlink at exactly this path"* — plus three factual errors in the new rationale prose (most notably "CrawDad calls it at `ceiling=open`"; it calls at the channel's configured ceiling, `personal` by default). All fixed, each via a failing test first where behaviour changed.
- `creek_mcp/server.py` is touched for one reason only: its `list_tools()` description advertised "over a vault-relative directory", which this fix narrows. The ceiling was **already** wired through to the tool at HEAD — this was the same root cause as #1037/#1068/#1078/#1084 (a ceiling received and ignored), not a wiring gap.

Closes #972
Refs #932, #846, #848

### DIFF
--- a/creek-tools/creek_mcp/read_gate.py
+++ b/creek-tools/creek_mcp/read_gate.py
@@ -92,6 +92,25 @@ refuses, on ``refuse_above_ceiling``. Read that as #1068's compile/reflect
 reasoning one level up: read's target is not caller-*named*, but it is
 caller-*addressed* and singular, which is the property that decides.
 
+#972 closed ``creek.redact.scan`` one step further out again: by narrowing the
+tool's *domain* rather than filtering its output. There was nothing to filter.
+The scan is a regex pass over bytes that opens no front matter, so it holds no
+tier for any file it names, and the leak was the filename itself — Creek
+fragment filenames are slugified titles. Its gate therefore decides *where*:
+the FEAT-027 staging subtree at every ceiling, because that is the one call
+CrawDad makes and it makes it at whatever ceiling the channel is configured
+for — ``personal`` unless an operator mapped that channel, ``open`` where one
+did — so the subtree has to be admitted at the lowest of them; everything else
+in the vault ranked as ``intimate``, because for all this tool knows it is.
+``refuse_above_ceiling`` was deliberately not adopted, for a sharper reason
+than ``report``'s: :data:`GENERIC_ABOVE_CEILING_REASON` asserts that *resolved
+content* exceeds the ceiling, which would be false here — nothing was resolved
+and nothing was ranked — and useless to a CrawDad operator, whose only
+actionable fact is where they pointed the scan. The gap also had a second half
+no gate could have closed, and it belongs to the same lesson: the response
+rendered a symlinked child under its target's name, so the scope fix is paired
+with a single as-scanned path renderer every field goes through.
+
 :data:`TOOL_POSTURES` is verified by ``tests/test_mcp_read_gate.py``, which
 fails if an entry lies: the manifest must match ``server.list_tools()``
 exactly, a ``GATED`` claim is checked against a real call site in the named
@@ -429,18 +448,71 @@ TOOL_POSTURES: dict[str, ToolPosture] = {
         rationale=_COUNTS_ONLY_RATIONALE,
     ),
     "creek.redact.scan": ToolPosture(
-        posture=ReadPosture.UNGATED_KNOWN_GAP,
+        posture=ReadPosture.GATED,
         rationale=(
-            "The caller names the path and resolve_within_vault confines it to "
-            "the vault, and no matched text is returned — but the confinement "
-            "is to the whole vault, not to the FEAT-027 staging subtree, and "
-            "the tier is never consulted. An open-ceiling caller pointed at "
-            "01-Fragments gets back the filenames of intimate fragments, which "
-            "are slugified titles, plus which PII types each contains. This "
-            "posture was CALLER_NAMED_PATHS until the #932 review; the name "
-            "was true and the conclusion drawn from it was not (#972)."
+            "Closed by narrowing the tool's domain, not by filtering its "
+            "output: the scan is a regex pass over bytes that reads no "
+            "per-file privacy_tier at all, so it has nothing to rank a "
+            "fragment with. _refuse_outside_scan_scope admits the FEAT-027 "
+            "staging subtree 00-Creek-Meta/Inbound/ at every ceiling — "
+            "CrawDad's safety pass runs there at the channel's configured "
+            "ceiling, personal by default and open only where an operator "
+            "mapped that channel (crawdad/crawdad/bot.py::_channel_tier), so "
+            "it has to keep working at the lowest of them — and ranks every "
+            "other in-vault target as if it held intimate content, so only "
+            "ceiling=intimate or all admits it. Consequence to know: that is "
+            "why a personal-ceiling caller cannot scan 09-Reference/ either — "
+            "the tool cannot tell it from 01-Fragments. Ordering is "
+            "load-bearing: the gate sits ABOVE the existence check, because "
+            "'input_path not found' versus a successful scan is a one-bit "
+            "existence oracle over every slugified fragment title in the "
+            "vault. Below ceiling=intimate all three out-of-scope answers are "
+            "now the same fixed string, derived from the caller's own "
+            "input_path and their own declared ceiling and echoing no tier, "
+            "no counts and no path: the file that is there, the name that is "
+            "nothing, and the symlink that resolves off the vault — which "
+            "resolve_within_vault necessarily catches ABOVE the gate, having "
+            "no resolved path to hand it, and which therefore collapses onto "
+            "the same reason in _outside_vault_reason rather than being "
+            "reordered under it. So for a caller the ceiling does not admit "
+            "vault-wide, the refusal is an oracle for neither existence nor "
+            "rank. The precise 'resolves outside the vault root' message "
+            "survives at ceiling=intimate and all, where it discloses nothing "
+            "the caller could not already read and a dangling link — a stale "
+            "sync folder, a moved drive — is the operator's one actionable "
+            "diagnostic. "
+            "refuse_above_ceiling was deliberately not adopted because its "
+            "generic reason asserts that resolved content was ranked, which "
+            "nothing here ever was. Deliberate divergence from rule 1 above: "
+            "the audit entry DOES record the probed target. It records the "
+            "caller's own input_path string and not the resolved path — which "
+            "for a staged symlink would be an intimate fragment's path — "
+            "because 'where did consumer X aim the scanner' is the actionable "
+            "fact about a security tool, and the string was already the "
+            "caller's. The second, independent half: every path in the "
+            "response — findings, the report_markdown headings CrawDad posts "
+            "verbatim into a channel, and the input_path echo — is "
+            "vault-relative through the single helper _vault_relative, and "
+            "rendered as scanned rather than as resolved, because a renderer "
+            "that resolved first let a symlink staged under Inbound/ disclose "
+            "an intimate fragment's slugified title from inside the admitted "
+            "subtree (#972). The echo is the one exception — it renders the "
+            "already-resolved target — and it is safe only because the gate "
+            "refuses an out-of-scope symlink before there is anything to "
+            "echo. Residuals to know: RedactionScanner.scan_batch still "
+            "follows symlinked children, so such a target's PII types and "
+            "line numbers are still disclosed even though its path no longer "
+            "is, and it counts toward statistics.files_scanned whether or not "
+            "it yields a finding, which is an existence-and-readability bit "
+            "about the target on top of the finding detail (#1087) — bounded "
+            "to symlinked FILES, since rglob does not descend into symlinked "
+            "directories and scan_batch keeps only is_file() children, so a "
+            "link to 01-Fragments/ staged as a directory yields nothing; and "
+            "existence probing *within* Inbound/ still works, which is the "
+            "tool's job rather than a leak."
         ),
-        gap_issue=972,
+        gate_module="creek_mcp.tools.redact",
+        gate_symbol="_refuse_outside_scan_scope",
     ),
     "creek.save": ToolPosture(
         posture=ReadPosture.NO_UNSUPPLIED_READ,
--- a/creek-tools/creek_mcp/server.py
+++ b/creek-tools/creek_mcp/server.py
@@ -512,7 +512,12 @@ def build_server(
         input_path: str,
         privacy_tier_ceiling: TierCeiling = TierCeiling.OPEN,
     ) -> dict[str, Any]:
-        """Read-only PII / secret scan over a vault-relative directory (FEAT-027)."""
+        """Read-only PII / secret scan of the FEAT-027 staging subtree.
+
+        Scoped to ``00-Creek-Meta/Inbound/``, which every ceiling admits. The
+        scan reads no per-file privacy tier, so any other vault path is ranked
+        as intimate content and needs privacy_tier_ceiling=intimate or all.
+        """
         return redact_scan_tool(
             vault_path=vault,
             input_path=input_path,
--- a/creek-tools/creek_mcp/tools/redact.py
+++ b/creek-tools/creek_mcp/tools/redact.py
@@ -7,43 +7,135 @@ a markdown summary suitable for embedding in a Discord reply.
 
 The tool is **read-only**: it never writes to the scanned files and
 never invokes the redactor. ``--apply`` lives behind the CLI for a
-deliberate human-driven step. The default tier is :data:`TierCeiling.OPEN`
-because the scan only reports counts + line numbers + salted hashes —
-no matched text and no file body content leaves the tool.
+deliberate human-driven step, and no matched text ever leaves the tool —
+findings carry counts, line numbers and salted hashes only.
 
 FEAT-027 introduces this tool so CrawDad can run the safety pass on
 Discord attachments staged under ``00-Creek-Meta/Inbound/`` before any
 ``creek.ingest`` call is dispatched.
 
-Read-side posture (#972): "no matched text leaves the tool" is true and is
-not sufficient. ``resolve_within_vault`` confines *input_path* to the whole
-vault rather than to the staging subtree above, and the tier is never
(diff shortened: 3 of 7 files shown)


Geoffe-Ga:
## Summary
PR #1090 closes #972, a read-side leak in `creek.redact.scan`: an `open`-ceiling caller aimed at `01-Fragments` could read back intimate fragments' slugified filenames (plus PII types/line numbers), and `report_markdown` separately leaked absolute filesystem paths. The fix scopes the tool to the FEAT-027 staging subtree (`00-Creek-Meta/Inbound/`) admitted at every ceiling, ranks everything else as `intimate`-tier, reorders the scope gate above the existence check to kill an existence oracle, and routes every path in the response through one fail-closed, non-resolving renderer to kill a paired symlink-disclosure leak.

## Strengths
- **Correct gate ordering.** The scope gate sits above `resolved.exists()`, which is exactly what's needed to prevent a one-bit existence oracle over slugified fragment titles — verified in the code and pinned by `test_scan_refuses_out_of_scope_paths_identically_whether_or_not_they_exist`.
- **The two-part fix is real, not cosmetic.** Scoping alone would not have caught a symlink staged *inside* the admitted subtree pointing at an intimate fragment (`rglob` yields symlinks unresolved); the `_vault_relative`/`_relativised_summary` renderer independently fixes this by never calling `.resolve()` on a match path. I confirmed `generate_markdown_summary` uses `str(match.file_path)` verbatim with no further resolution, so the relativized summary flows through cleanly to `report_markdown`, closing the field CrawDad posts to Discord.
- **Manifest posture correctly wired.** `creek_mcp.read_gate.TOOL_POSTURES["creek.redact.scan"]` moved `UNGATED_KNOWN_GAP` → `GATED` with `gate_module`/`gate_symbol` pointing at the real, actually-invoked `_refuse_outside_scan_scope` (verified via the AST-based layer (c) checks in `test_mcp_read_gate.py`), and the tool is wired into layer (f)'s probe sweep with a canary seeded in a **filename** — the one place this specific leak class is reachable, correctly identified since the shared `canary_vault` fixture is structurally blind to it.
- **Traced the supporting primitives and found them consistent**: `tier_allowed(INTIMATE, ceiling)` in `tier_ceiling.py` admits exactly `intimate`/`all` as claimed; `resolve_within_vault` fully resolves symlinks and `..` before the vault-containment check, so a staged symlink can't scope-escape by name; CrawDad's `_channel_tier` defaults to `personal` and the staging path passed matches the hardcoded `00-Creek-Meta/Inbound` scope (the configurable-staging-path mismatch is honestly flagged and filed as #1088 rather than silently left).
- **Excellent test coverage of edge cases**: sibling-prefix directory (`Inbound-other`) not mistaken for a child via `is_relative_to` vs string-prefix, vault-root-as-scan-target, symlinked vault-root component (the `/tmp`-on-macOS case), unrenderable-path fallback placeholder, and a cross-field invariant test (`findings` vs `report_markdown` headings name the same set of files) that would catch exactly the kind of divergence #972 itself was.
- **Honest scope-setting.** Residuals (symlinked-file PII/existence still disclosed, configurable staging path, NUL-byte `ValueError`) are explicitly acknowledged and filed as follow-ups (#1087, #1088, #1089) rather than silently left or over-claimed as fixed.

## Security Concerns
None found. 🟢 The fix appears to fully close the reported leak (filename disclosure + absolute-path disclosure) within its stated scope, and the accepted residuals are bounded and separately tracked. One thing worth double-checking by a maintainer with a live symlink-capable filesystem (not verifiable by static read alone): `_refuse_outside_scan_scope` computes `vault_path.resolve() / _STAGING_SUBDIR` without a further `.resolve()` on the joined result — if `00-Creek-Meta` or `Inbound` itself were ever a symlink, this could diverge from the actual on-disk staging directory. This fails *closed* (denies scanning) rather than open, so it's not a disclosure risk, just worth a mental note. 🟢 LOW, not blocking.

## Problems
None found that block merge.

## Code Quality
- No issues. Functions are small and single-purpose (`_admitted_vault_wide`, `_outside_vault_reason`, `_refuse_outside_scan_scope`, `_vault_relative`, `_relativised_summary`, `_finding_to_dict`), which should keep cyclomatic complexity well under the repo's ≤10 cap.
- The docstrings are extremely long-form/essay-style, but this matches the established convention already present throughout `read_gate.py` and neighboring tool modules (e.g. `creek.report`, `creek.journal`) — not a new pattern introduced by this PR, so not flagged as a deviation.

## CLAUDE.md Adherence
- ✅ TDD: test file docstrings and structure clearly reflect the described RED→GREEN cycle re-demonstrated by the author.
- ✅ Quality thresholds: PR claims 100% coverage on `creek_mcp/tools/redact.py` and `./scripts/check-all.sh` → 12/12 passed; consistent with the density of tests added.
- ⚠️ Gate 3 (CI green before review): at time of this review, `Code Quality & Testing (3.11/3.12/3.13)` and `claude-review` checks were still `IN_PROGRESS`; `CrawDad Quality` jobs had already passed. Confirm all jobs finish green before merging, per this repo's Stay-Green policy.
- ✅ Commit conventions: conventional-commit style message, references issues correctly (`Closes #972`, `Refs #932, #846, #848`).

## Testing
Strong, and specifically targeted at the failure modes named in the issue: existence-oracle collapse above/below the gate, the outward-symlink oracle (a real Gate-2.5-review-caught issue, per the PR body), scope-containment edge cases (sibling-prefix, vault-root), the filename-canary leak reproduction, and a cross-field consistency invariant between `findings` and `report_markdown`. Nothing insufficient here.

## Documentation
`docs/mcp.md` and `docs/redaction.md` updates accurately describe the new scoping rule, the local-operator escape hatch (`ceiling=intimate`/`all`), and the two accepted residuals with their tracking issues — consistent with what the code actually does.



## Verdict: LGTM

The scope gate is correctly ordered (above the existence check, above the per-file admission check), the path-rendering fix genuinely closes the symlink-disclosure sub-leak, and the test suite (existence-oracle collapse, outward-symlink oracle, sibling-prefix near-miss, vault-root-as-target, canary-based leak proof, cross-field path-equality invariant) verifiably covers the exact failure modes described in the PR body. Tracing the gate logic, tier_ceiling.tier_allowed, resolve_within_vault, and the scanner/CrawDad call sites found no bug or gap the tests miss. Nothing here should block merge; note only that CI ("Code Quality & Testing") was still IN_PROGRESS at review time and must go green per Gate 3 before this actually merges.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5145879215 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1094, opened 2026-07-31
docs: reconcile the two Adepthood<->Creek contract documents; the adepthood copy asserts intimate guarantees Creek does not implement


Geoffe-Ga:
## Context

Raised while writing the #1072 ADR, which had to describe the shipped surface
accurately and found the two published contract documents no longer do.

`docs/decisions/2026-06-30-adepthood-creek-mcp-contract.md` (Creek side) and
`docs/creek-vault-mcp-contract.md` in `Geoffe-Ga/adepthood` (their side) were
written as mirrored drafts. Both are now stale in ways that matter, and the
adepthood copy in particular asserts guarantees Creek does not implement.

## The divergences

**Adepthood's copy asserts things that are false against Creek at HEAD:**

1. **"Adepthood's Aspects, Creek's Frequencies, and the Wavelength phases are
   one vocabulary under three names"**, and the wheel returns "the same shape
   as adepthood's local `WheelBalanceResponse`". Both are wrong. `wheel_tool`
   returns APTITUDE frequency counts/shares over the classified corpus;
   `WheelBalanceResponse` is per-stage aspect *fullness* from Adepthood's own
   curriculum. Creek's `share` is a fraction of the classified corpus summing
   to 1.0 across ten frequencies; Adepthood's `fullness` is per-stage
   completion in `[0, 1]` with no such constraint. Different quantities that
   happen to share a cardinality. The #1072 ADR ratifies the ownership split
   and supersedes this claim.
2. **Decision (b): "`creek.journal` writes of intimate content ... are
   permitted only after the vault's enclave has successfully completed remote
   attestation."** No attestation path exists. `_BoundedFastMCP.call_tool`
   caps every remote consumer at `ceiling=personal` unconditionally
   (`_REMOTE_ADMITTED_CEILINGS`), so an intimate remote write is refused
   before dispatch with no attestation branch to reach. #757 owns the
   confidential path; until it lands, Decision (b) describes a capability that
   is not merely unimplemented but structurally unreachable.
3. **Decision (c): a reflection grounded in intimate content "may return"**
   over an end-to-end-encrypted channel. Same problem: there is no such
   channel, and a remote caller cannot reach intimate content to ground on.
4. **`CONSUMER_ID = "CREEK_MCP_CONSUMER"`** (`backend/src/domain/creek_vault.py:46`)
   is the *environment variable name*, sent as the consumer identity value.
[...]

Issue #1093, opened 2026-07-31
chore(api): tighten the /v1 contract models so the published schema enforces what the ADR states in prose


Geoffe-Ga:
## Context

Found during the Gate-2.5 review of #1072 (the `/v1` contract models). Not a
leak and not a bug today — a set of tightenings that would make the published
contract enforce mechanically what it currently states in prose.

`creek-tools/creek_mcp/api/models.py` is the canonical wire vocabulary for
`/v1`, and `docs/contracts/adepthood-v1/` is the bundle Adepthood vendors. The
whole point of the bundle is that a consumer validating against it gets the
same answer the server would give. Each item below is a place where the
*schema* is looser than the *prose*.

## Items

1. **`TierModel.ceilings` is unbounded.** Typed `list[WireTierCeiling]`, while
   the docstring promises exactly the two members "in ascending breadth". Pin
   with `min_length=2, max_length=2` plus an order validator, so the published
   schema says it.
2. **`CONTRACT_MODELS` is hand-maintained.** Nothing asserts it equals the set
   of `_WireModel` subclasses, so a new model that is never registered is
   silently absent from the bundle. Either derive it from
   `_WireModel.__subclasses__()` or add a test that compares the two.
3. **`write_bundle` overlays rather than prunes.** Its docstring claims "a
   model removed here disappears from the bundle", but a removed or renamed
   model leaves a stale file on disk until someone `git rm`s it. Prune the
   destination first, or correct the docstring.
4. **`OK_STATUS` does not constrain the fields it names.** Three separate
   sources of the string `"ok"` (`OK_STATUS`, and two `Literal["ok"]`
   annotations).
5. **`JournalUpsertRequest.timestamp` has no documented format.** It is
   ISO-8601 in prose only; the schema says `string`. Add a `format`/`pattern`
   or a documented example.
6. **No published model for the `X-Creek-Tier-Ceiling` header.** Nothing
   mechanically forces a future implementer to parse it via `WireTierCeiling`
   (two members) rather than `TierCeiling` (four, including `intimate`/`all`).
   A header model in `CONTRACT_MODELS` would close that.
7. **`FORBIDDEN_API_IMPORTS` covers only four frameworks** (`fastapi`,
   `starlette`, `uvicorn`, `httpx`). Widen to `flask`, `django`, `aiohttp`,
   `sanic`, etc., so the framework-free guarantee is not a whitelist of the
   frameworks we happened to think of.
[...]

Issue #1075, opened 2026-07-31
feat(api): PUT /v1/journal-entries/{external_id} over the shared journal service, with HTTP↔MCP parity tests


Geoffe-Ga:
## Role

Backend engineer replacing the first tracer stub with real behavior, and
proving the two adapters cannot diverge.

## Problem

`PUT /v1/journal-entries/{external_id}` returns `not_implemented` after #1074.
Journal ingest is the one Adepthood capability whose MCP parameter names already
match (`creek_mcp/tools/journal.py:119-129` ↔ `creek_vault_client.py:282-297`),
so it is the lowest-risk vertical to make real — and the one that proves the
"one behavioral implementation" claim is true rather than aspirational.

It is also the **write** path, which is where a bypass would do durable damage:
`journal_ingest_tool` refuses an entry whose tier exceeds the caller's ceiling
*before* staging (`tools/journal.py:168-179`), and audits both the refusal and
the success (`:169-174`, `:212-220`). An HTTP route that called `run_ingest`
directly, or that passed a ceiling the adapter had not admitted, would lose
both guarantees silently.

## Goal

The HTTP route is a parse-and-delegate shim over the **existing**
`journal_ingest_tool`, with idempotency, tier refusal, fragment reference and
audit behavior identical to MCP's — asserted by tests that run the same
scenarios through both adapters.

## Architectural context

HTTP is the Adepthood application adapter; MCP stays the agent adapter; both
call one implementation. If this route contains any privacy, idempotency or
audit logic of its own, the design is wrong — that logic belongs in
`tools/journal.py` where MCP already gets it.

## Proposed tracer

One OPEN-tier entry, sent twice with the same `external_id`, yields
`action: "created"` then `action: "unchanged"`, one fragment, and two audit
records — over HTTP, matching MCP exactly.

## Tasks

- Route handler: validate the request against #1072's model, resolve the
  admitted ceiling from #1073's policy, call `journal_ingest_tool`, project its
  return onto the canonical response model.
- Map `journal_ingest_tool`'s structured refusals onto the canonical error
  taxonomy: blank content/id → `invalid_request`; unknown tier →
  `invalid_request`; tier-exceeds-ceiling → `privacy_refused`; `"vault
  unavailable"` → `temporarily_unavailable`; `"ingest failed: …"` →
  `internal_error` **with the underlying message not echoed** (it can contain
  staged file paths).
- `external_id` arrives as a path segment: reject empty/oversized ids and any
[...]

Issue #1074, opened 2026-07-31
feat(api): authenticated /v1 tracer skeleton — real capabilities endpoint, honest not_implemented routes, generated OpenAPI


Geoffe-Ga:
## Role

Backend engineer wiring a tracer: the thinnest possible end-to-end path that is
honest about what it does and does not yet do.

## Problem

There is no HTTP entry point into Creek. `creek-tools-mcp --transport network`
(`creek-tools/creek_mcp/server.py:804-831`) serves streamable-HTTP MCP and
nothing else. Adepthood cannot begin its client work (Geoffe-Ga/adepthood#2045)
against anything real, and the two repos' only integration test surface today is
each other's imagination — which is how the four-way parameter mismatch
documented in #1071 survived shipping.

## Goal

`creek-tools-api` (or an equivalent documented entry point) starts, requires a
bearer token, serves a **real** `GET /v1/capabilities` built from the canonical
models, exposes journal/reflection/wheel routes that return an unmistakable
`not_implemented` error envelope, and generates OpenAPI. MCP is untouched.

## Architectural context

HTTP is the Adepthood application adapter; MCP stays the agent adapter; both
call one implementation. This issue establishes the adapter and its security
posture — it deliberately implements **no** capability behavior beyond
capabilities itself, so the tracer proves the path before the payloads.

## Proposed tracer

```
Adepthood test call
  → bearer-authenticated GET /v1/capabilities
  → shared admission/identity policy (#1073)
  → canonical CapabilitiesResponse model (#1072)
  → schema-valid JSON
```

## Tasks

**Adapter:**
- App factory + documented entry point (console script or
  `python -m creek_mcp.api`), with `--host`/`--port`/TLS flags mirroring
  `creek_mcp.server`'s. Default port must **not** be 8000 — that collides with
  Adepthood's own backend; pick and document another.
- Bearer auth verifying against the existing `CREEK_MCP_CONSUMER_TOKENS`
  registry via `creek_mcp.remote_auth` / `creek_mcp.token_policy`, honoring the
  ≥32-char floor. Constant-time compare. Missing or bad credentials → 401 with
  an error envelope that discloses no capability information.
- Build a **remote** `CallerIdentity` from the verified token and route every
  request through `creek_mcp.policy.admitted_ceiling` from #1073, *before* any
  handler runs — so `intimate`/`all` are refused at the adapter edge exactly as
  they are for remote MCP.
[...]

PR #1095, opened 2026-07-31
docs(adr): ratify the canonical Adepthood /v1 application contract — models, error taxonomy, versioning, and a published fixture bundle

### LINKED ISSUE #1072: docs(adr): ratify the canonical Adepthood application contract — models, error taxonomy, versioning, and a published fixture bundle
## Role

Principal architect writing a decision record and the Pydantic models that make
it executable. This issue settles what the two repos disagree about; every other
sub-issue of #1071 implements against it.

## Problem

The two repositories inferred the Adepthood↔Creek contract independently and
each implemented a different one. Verified at creek-vault `de33c3e` /
adepthood `fbc529d`, the disagreements are not cosmetic:

1. **`creek.reflect` parameter names share nothing.** Adepthood sends
   `{consumer, body, tier_ceiling}` (`backend/src/services/creek_vault_client.py:277-279`);
   `server.py:332-346` registers `(content, entry_ref, privacy_tier_ceiling)`.
2. **`creek.wheel` takes an unexpected `consumer`.** `server.py:349-358`
   accepts `privacy_tier_ceiling` only.
3. **Reflection result shape.** `reflect_tool` returns
   `{status, tool, tier_ceiling, routed_tier, notes[], essay_grounded, essay?}`
   (`creek-tools/creek_mcp/tools/reflect.py:479-490`); Adepthood reads a scalar
   `reflection` string.
4. **Wheel semantics are genuinely different concepts, not different spellings.**
   `wheel_tool` (`tools/wheel.py:95-110`) returns APTITUDE **frequency**
   counts/shares over the classified corpus; Adepthood's `WheelBalanceResponse`
   is per-**stage aspect fullness** from its own curriculum. Creek does not
   have Adepthood's stage/aspect vocabulary and must not invent it.
5. **The version gate cannot detect drift.** Creek publishes
   `CONTRACT_VERSION = "0.2.0"` (`creek_mcp/contract.py:18`); Adepthood
   advertises `"0.1.0-draft"` and compares the **major component only**
   (`creek_vault_client.py:79`, `:242`). Both are major `0`, so the check
   passes — on a 0.x pair where a minor bump is exactly what a breaking change
   looks like.
6. **The audited consumer and the remote ceiling cap are MCP-middleware-bound**
   (`server.py:94-101`, `:103-144`), so they have no definition at all for a
   non-MCP adapter.

Nothing here is fixable by one repo alone, and the current failure mode —
everything collapsing to "vault unavailable" on the client
(`creek_vault_client.py:381-391`) — means neither repo gets a signal when it
guesses wrong.

## Goal

An accepted ADR under `docs/decisions/` plus a canonical Pydantic model module
that together define, unambiguously, what `/v1` means — and a versioned fixture
bundle Adepthood can vendor so the two repos verify against the same bytes.

## Architectural context

HTTP/JSON `/v1` is the Adepthood application adapter; MCP stays the agent
adapter; both call one implementation. Protocol choice is separate from
confidential-compute guarantees — the ADR must say so in as many words.

## Scope

**Decide and record:**

- **Versioning.** `/v1` in the path. `contract_version` (the `/v1` surface) is
  distinct from `ontology_version` (`aptitude-wavelength/2026-05-23`, the shared
  frequency/phase vocabulary). Compatibility rule for a `0.x` contract: minor
  bumps may break, so **exact-minor** match is required until 1.0, not
  major-only. Deprecation policy for removing a field.
- **Capabilities.** `GET /v1/capabilities` shape: vault presence, both
  versions, the tier model, and the list of `/v1` capabilities actually served.
  It must distinguish *four* states a client can act on differently: vault
  present and usable; server reachable but vault uninitialized; server
  reachable but contract-incompatible; server unreachable.
- **Journal upsert.** `PUT /v1/journal-entries/{external_id}` with body
  `{content, timestamp, tier}` → `{status, fragment_id, action, tier}` where
  `action ∈ {created, updated, unchanged}`. Reconcile with
  `journal_ingest_tool`'s existing return (`tools/journal.py:221-229`), which
  already has this shape — prefer adopting it over inventing a new one.
- **Reflection.** `POST /v1/reflections`. The canonical result is
  **anchored notes plus an optional ungrounded essay** — Creek's existing shape.
  It must NOT be collapsed to a scalar string. Record: `content` vs `entry_ref`
  rules; that `essay_grounded` is load-bearing and a client may not treat the
  essay as grounded; that every `notes[].quote` is a verified verbatim span of
  the entry; which `routed_tier` provenance is safe to expose.
- **Wheel.** `GET /v1/wheel` returns **Creek's frequency counts/shares**,
  verbatim — the corpus fact Creek actually owns. The stage/aspect projection is
  **Adepthood's** to own and test (their #1937). Record this ownership split
  explicitly, and that the F1–F10 ↔ 10-stage numeric coincidence is not a
  semantic identity. All ten frequencies always present; deterministic ordering;
  empty corpus yields all-zero counts and `0.0` shares, not an error.
- **Error and refusal taxonomy.** Stable machine-readable `code` values —
  `unavailable`, `unsupported_capability`, `incompatible_version`,
  `invalid_request`, `privacy_refused`, `care_escalation`, `not_found`,
  `temporarily_unavailable`, `internal_error` — each mapped to an HTTP status.
  Clients must never parse prose to recover semantics. Error bodies may never
  echo journal content, credentials, or tier-unknown vault text.
- **Auth and consumer identity.** Bearer token, reusing the
  `CREEK_MCP_CONSUMER_TOKENS` registry and the ≥32-char floor from
  `creek_mcp/token_policy.py`. How the authenticated identity becomes the audit
  `consumer`. Loopback-plaintext vs remote-TLS posture, matching
  `_require_transport_confidentiality` (`server.py:770-802`). Missing or invalid
  credentials fail closed with no capability disclosure.
- **INTIMATE.** `/v1` refuses `intimate` and `all` ceilings exactly as
  `_BoundedFastMCP.call_tool` does for remote MCP today. The ADR states plainly
  that TLS does not satisfy the intimate-transit contract and that #757 owns the
  encrypted/attested path.
- **HTTP framework choice.** Verified: `pydantic>=2`, `httpx`, `uvicorn` are
  direct deps and `starlette` is present transitively via the MCP SDK
  (`creek-tools/uv.lock:3462`); **FastAPI is not a dependency today**. Record
  the choice — FastAPI (one new direct dep, OpenAPI generation for free) vs
  Starlette + hand-written OpenAPI — with the reasoning, so sub-issue 3 does not
  have to relitigate it.

**Build:** a `creek_mcp/api/models.py` (or equivalent) of canonical Pydantic
request/response/error models, and a `docs/contracts/adepthood-v1/` fixture
bundle — the generated JSON Schema plus example payloads for **success, empty,
refusal, care escalation, malformed input, incompatible version, and unavailable
service** for each of the four capabilities. The bundle carries the contract
version in its path or manifest so Adepthood can vendor a pinned copy.

**Not in scope:** any route, any change to MCP behavior, any change to
`tools/*.py`. This issue is contract + models + fixtures only.

## Acceptance criteria

- [ ] `docs/decisions/<date>-adepthood-http-application-api.md` exists, is
      linked from `docs/decisions.md`, and answers every bullet above with a
      decision — not a list of options.
- [ ] The ADR contains a table mapping each of the six verified divergences
      above to the resolution that supersedes it, naming which repo changes.
- [ ] Canonical Pydantic models exist for capabilities, journal upsert,
      reflection, wheel, and the error envelope; every error `code` in the
      taxonomy is a member of a closed enum, not a free string.
- [ ] The wheel model rejects a payload missing any of F1–F10 and preserves
      deterministic ordering.
- [ ] `docs/contracts/adepthood-v1/` contains JSON Schema plus the seven named
      example payloads per capability, and a test asserts every example
      validates against its model.
- [ ] A test asserts the ADR's declared `contract_version` equals
      `creek_mcp.contract.CONTRACT_VERSION`, so the doc cannot rot.
- [ ] The ADR states explicitly that HTTP delivers no encryption-at-rest,
      attestation or key-custody guarantee, and links #757.
- [ ] No MCP behavior changes; `./scripts/test.sh --all` green.

## Verification

```bash
```

Coverage ≥90% branch, interrogate ≥95%, mypy strict, ruff clean, complexity ≤10.
No suppressions.

- No example fixture contains a real journal body, a credential, or an
  `intimate`-tier success response.
- The INTIMATE examples are **refusals**.
- The error taxonomy forbids echoing request content in any error body.

## Files likely affected

- create `docs/decisions/<date>-adepthood-http-application-api.md`
- create `creek-tools/creek_mcp/api/models.py`
- create `docs/contracts/adepthood-v1/**`
- create `creek-tools/tests/test_adepthood_contract_models.py`
- modify `docs/decisions.md`, `creek-tools/pyproject.toml` (only if FastAPI is chosen)

## Dependencies

Blocks every other sub-issue of #1071. Counterpart: Geoffe-Ga/adepthood#2044
mirrors this ADR client-side. Supersedes the framing of #875 (which documents
the MCP surface) — keep #875, but it should land first or reference this.

## Budget escape hatch

If the wheel or reflection debate stalls, ship the ADR with capabilities +
journal decided and the other two marked `PROPOSED` with their open question
stated. Sub-issues 3 and 4 can then proceed. Never ship an ADR that leaves a
privacy, auth, or INTIMATE question open.

Geoffe-Ga:
## Summary

- **Ratifies the Adepthood↔Creek `/v1` contract** in a new accepted ADR, `docs/decisions/2026-07-31-adepthood-http-application-api.md`, with a table mapping each of the six verified divergences to the resolution that supersedes it and naming which repo changes. Indexed from a new `docs/decisions.md` (no ADR index existed).
- **Makes it executable**: `creek_mcp/api/models.py` is a framework-free Pydantic wire vocabulary (16 models, closed enums, total error tables); `creek_mcp/api/bundle.py` generates the published fixture bundle.
- **Publishes** `docs/contracts/adepthood-v1/` — 16 JSON Schemas + 28 example payloads (4 capabilities × 7 states) + a sha256 `manifest.json` + `retry-policy.json`, all generated, plus a hand-written README.
- **No route, no MCP behavior change, no new dependency.** `creek_mcp/server.py`, `creek_mcp/tools/*.py`, `pyproject.toml` and `uv.lock` are untouched.

## Decisions worth reviewing closely

**The error taxonomy is deliberately narrower than the issue sketched.** `not_found` is re-scoped to a **routing** code — emitted only for an unknown path — and every vault-object non-answer collapses to `privacy_refused` carrying the imported `read_gate.GENERIC_ABOVE_CEILING_REASON`. Letting `404` and `403` stay distinguishable over a vault object would hand back exactly the existence-and-rank oracle #846/#970/#972/#1090 spent five issues collapsing. The ADR says per distinction which are safe to expose (all derived from the caller's own input or static server config) and which must collapse, and fences `unavailable`/`temporarily_unavailable`/`internal_error` as **global-only**, never per-object.

**`care_escalation` is not an error code.** It was in the issue's proposed enum; it is removed and promoted to a 200-shaped `CareEscalationResponse`. A client that maps care escalation to a failure breaks the care contract.

**#1082 is answered by the retry table, not by a wire field.** `privacy_refused` → `terminal` for that `(endpoint, identifier)` pair. Retryability is a pure function of `code`, published as `retry-policy.json`, so it carries provably zero bits beyond the code — whereas a per-response `retryable` boolean would invite future refinement that varies *within* one code, which is how an oracle returns. The honest residual is stated: because the refusal is blurred, a purged fragment with a dangling ledger record (#1080) is also treated as terminal.

**`routed_tier` is exposed, with a proof.** `routing_tier(ceiling, tier) == CEILING_ROUTING_TIER[ceiling]` for every ceiling `/v1` permits and every tier that ceiling admits — verified empirically, then pinned by `test_routed_tier_is_a_constant_function_of_the_declared_ceiling` with a non-vacuity guard. If that test ever fails, the field must be removed, not the test relaxed.

**Versioning is a hard break, made legible.** Exact-minor match below 1.0 via `X-Creek-Contract-Version`, with `/v1/capabilities` exempt so the negotiation endpoint can never itself fail to negotiate. A client pinned to `0.1.0-draft` gets `409` plus a readable `/v1/capabilities`, instead of collapsing to "vault unavailable" — the collapse this epic exists to kill. `CONTRACT_VERSION` is deliberately **not** bumped: this publishes a new surface rather than changing one.

**Starlette, not FastAPI**, with four recorded reasons (MCP SDK pin coupling; a published OpenAPI doc must not be a function of a framework version; FastAPI's default 422 echoes the offending input, violating no-content-echo; five endpoints). #1074 should not relitigate it, and should promote `starlette` to a *declared* direct dep.

**INTIMATE is unreachable structurally**, not by a runtime branch: `WireTierCeiling` has two members mirroring `server._REMOTE_ADMITTED_CEILINGS`, so no wire position can name `intimate`. The ADR states plainly that TLS is transport confidentiality only, does not satisfy the intimate-transit contract, and that #757 owns the encrypted/attested path.

**A binding rule for #1075** came out of review: the `/v1` journal handler MUST enter through `journal_ingest_tool` and MUST NOT call `write_fragment_idempotent`/`run_ingest` directly. The overwrite gate is `_refuse_unadmitted_overwrite` in `creek_mcp/tools/journal.py`, *not* inside the pipeline function — wiring a route to the pipeline would bypass it and reopen the #970 bug class.

## Anti-rot

Two tests exist so this cannot decay: `test_committed_bundle_equals_a_fresh_build` asserts the committed bundle is byte-identical to `build_bundle()`, and the ADR's published `contract_version`/`ontology_version` are pinned to `creek_mcp.contract`. Change a model and the bundle goes red until regenerated.

## Test plan

- `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 checks. 7293 tests pass, 94.12% branch coverage, per-file gate passed (222 files ≥80%). ruff clean, mypy strict clean, xenon clean, interrogate 100% on `creek_mcp/api/`, pylint 10.00/10.
- New suite `tests/test_adepthood_contract_models.py`: **445 tests** in twelve groups — model validation, closed-enum membership mirrored against the runtime constants (`reflect._ALLOWED_KINDS`, `server._REMOTE_ADMITTED_CEILINGS`), wheel F1–F10 completeness + canonical ordering, error-envelope no-echo, the `routed_tier` non-oracle invariant, `essay_grounded`, INTIMATE unconstructibility, care-signal fidelity against `CARE_SIGNAL`, the bundle round-trip, the ADR↔code version pin, and AST layering checks.
- Coverage on the new package: `models.py` 100%, `bundle.py` 98.65%.
- Verified the baseline was green at the branch point (`04d355f`) before starting, so nothing here is masking a pre-existing failure.
- Divergences 1–6 were re-verified independently against `Geoffe-Ga/adepthood` at HEAD rather than taken from the issue body.

## Notes for the reviewer

- The issue's acceptance criterion says the ADR is "linked from `docs/decisions.md`" — that file did not exist, so it is created here as the ADR index. `creek-tools/docs/decisions.md` is an unrelated document (the decision-support *feature* manual); the new index carries a disambiguation note.
- Three of the 28 fixture cells (care escalation for capabilities / journal-upsert / wheel) are structurally unreachable, since the acute-distress guard runs only in `reflect_tool`. They validate against a real `NotApplicableExample` model and state why, so all 28 validate against their named model rather than three being fabricated or missing.
- Security invariants hold: no fixture contains a real journal body, a credential, or an `intimate`-tier success; every INTIMATE example is a refusal. Pinned by a value-walking test.
- Gate 2.5 ran twice. The first pass returned NEEDS WORK with eight MAJOR findings — including a factual misattribution of the journal overwrite gate, and two published-schema-looser-than-server bugs (`ReflectionResponse.status` admitting `escalate`; `essay_grounded` as a bare `bool` with a runtime-only validator). All were fixed and re-verified.
- Follow-ups filed: **#1093** (P3 — schema tightenings that would make the published schema enforce what the ADR states in prose) and **#1094** (P2 — the adepthood-side contract doc asserts attestation-gated intimate writes that Creek does not implement).

Closes #1072
Refs #1071


Geoffe-Ga:
## Summary
This PR ratifies the Adepthood↔Creek `/v1` contract as an accepted ADR (`docs/decisions/2026-07-31-adepthood-http-application-api.md`) and makes it executable: a new framework-free Pydantic vocabulary (`creek_mcp/api/models.py`), a deterministic fixture-bundle generator (`creek_mcp/api/bundle.py`), a 445-test suite, and the published bundle under `docs/contracts/adepthood-v1/`. No routes, no MCP behavior change, no new runtime dependency — this is a pure contract/vocabulary layer that #1074 will later mount behind Starlette.

## Strengths
- Every specific factual claim in the docstrings was independently verified against the *actual* referenced runtime code (`server._REMOTE_ADMITTED_CEILINGS`, `reflect._ALLOWED_KINDS`/`_DEFAULT_MAX_NOTES`, `journal._refuse_unadmitted_overwrite` ordering relative to the content-hash "unchanged" shortcut in `pipeline.py`) — all held up exactly as documented, including the subtle claim that the overwrite gate runs *above* the idempotency shortcut (which is what justifies `PRIVACY_REFUSED` being `TERMINAL` even for a re-sync).
- `INTIMATE` is made structurally unreachable (a 2-member `WireTierCeiling` enum) rather than gated by a runtime check that could be forgotten — verified no wire position can construct it, and the fixture bundle contains no `intimate` value anywhere.
- The `not_found`/`privacy_refused` taxonomy correctly collapses the existence-and-rank oracle the ADR says it's closing (#846/#970/#972/#1090) — `PRIVACY_REFUSED` is 403, never 404, and carries a constant, non-interpolated message.
- `RETRY_POLICY`, `ERROR_STATUS`, and `ERROR_MESSAGES` are genuinely total over all 9 `ErrorCode` members (directly enumerated, not just asserted by the accompanying test).
- The generated-bundle-must-equal-committed-bundle test (`test_committed_bundle_equals_a_fresh_build`) and the ADR-version-pin tests make this un-rottable by construction; a manifest hash was independently spot-checked via `sha256sum` against committed bytes and matched.
- The ADR is long (582 lines) but every paragraph carries a distinct decision or residual-risk note, consistent with this repo's established verbose-ADR style — nothing reads as padding on close inspection.
- `docs/decisions.md`'s disambiguation note about the unrelated `creek-tools/docs/decisions.md` is accurate — that's a genuinely different document (a decision-support feature manual).

## Security Concerns
🟢 LOW — None found. Fixture security invariants hold: no real journal body, credential, or `intimate`-tier success anywhere in the published bundle; the refusal examples correctly leak nothing beyond the generic above-ceiling reason.

## Problems
None found that block merge.

## Code Quality
- **(non-blocking)** No reflection-based test asserts that every `_WireModel` subclass in `models.py` is registered in `CONTRACT_MODELS` — currently true (16/16, verified by grep) but a future model added and forgotten in that dict wouldn't be caught, since the test fixtures (`HAPPY_PAYLOADS`/`MODEL_NAMES`) are keyed off `CONTRACT_MODELS` itself rather than off the module's live class list.
- **(non-blocking, forward-looking)** "`NOT_FOUND` never emitted for a vault object" is enforced only at the model/docstring level today. That's fine since #1074's actual HTTP handlers don't exist yet, but worth a tracked follow-up (or at least a comment/issue reference) so the invariant gets a repo-wide grep/AST guard once routes land — otherwise this security-relevant contract could quietly regress with no test noticing.
- Extra-field rejection for `ErrorEnvelope` is covered only via the generic all-model parametrized sweep, not a dedicated named test — a minor readability nit, not a coverage gap.

## CLAUDE.md Adherence
✅ New package/tests follow flat-layout conventions (`creek_mcp/api/`, `tests/`). ✅ Docstrings are extensive and specific (interrogate should be satisfied given the PR's claimed 100% on `creek_mcp/api/`). ✅ No new dependency, no framework import in the contract package (AST-checked by its own test). CI (`Code Quality & Testing` 3.11/3.12/3.13, `claude-review`) was still in progress at review time; `CrawDad Quality` jobs had already passed. The PR body reports `./scripts/check-all.sh` exit 0 locally (12/12, 94.12% branch coverage) — recommend confirming the in-progress CI jobs land green before merge, per the repo's 4-gate Stay-Green workflow.

## Testing
Strong. 445 tests (62 functions + 383 parametrized cases, verified by manual expansion) across the twelve claimed groups, all load-bearing invariant tests (`test_routed_tier_is_a_constant_function_of_the_declared_ceiling` with a genuine non-vacuity guard `test_admitted_ceiling_tier_sweep_is_not_vacuous`, five INTIMATE-unconstructibility tests, no-echo tests, AST layering tests, bundle round-trip, ADR version pin) exist and use exact-value assertions (sha256 hashes, byte-equality, frozenset membership against runtime constants) rather than weak `is not None` checks. No skips, broad excepts, sleeps, or non-determinism found. Coverage numbers (100%/98.65%) claimed in the PR body were not independently re-run in this review — flagged as unverified but plausible given the file's density.

## Documentation
Excellent. The ADR justifies all six divergence-resolution decisions in dedicated sections with citations (including a verified precise line-reference to `uv.lock` for the Starlette dependency claim), the new `docs/decisions.md` index is accurate, and the fixture bundle's `README.md` matches the actual generated schemas/examples (spot-checked field-ordering claims, `essay_grounded: const false`, and the three `NotApplicableExample` cells).



## Verdict: COMMENTS

No bugs, refuted claims, or CLAUDE.md violations surfaced across three independent deep-dive reviews (models/bundle correctness cross-checked against server.py/reflect.py/journal.py, the 445-test suite, and the ADR/fixture bundle) — everything the PR claims (invariants, test names, hashes, cross-references) checked out exactly. It stops short of LGTM only because there are two genuine, specific test-coverage suggestions worth the author weighing before merge: a reflection-based completeness check that no future model is added to `models.py` without being registered in `CONTRACT_MODELS`, and a forward-looking guard (even just a tracked follow-up issue) for the "NOT_FOUND is never emitted for a vault object" invariant once #1074 actually wires up HTTP handlers.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5146903822 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5146903822 to see in-depth feedback and continue iterating

Geoffe-Ga:
Thanks — all three Code Quality items addressed. Two changed code, one was already covered.

### 1. Reflection-based completeness check on `CONTRACT_MODELS` — fixed in this PR

You were right that this was a self-referential blind spot, not just a hypothetical: `HAPPY_PAYLOADS` and `MODEL_NAMES` are both keyed off `CONTRACT_MODELS`, so an unregistered model was invisible to all 445 tests *and* absent from the published schema bundle (since `bundle.py` walks the same mapping) without anything going red. Fixed rather than filed.

`creek-tools/tests/test_adepthood_contract_models.py` now has a `_wire_model_subclasses(namespace)` helper that discovers models by walking the **live module namespace** (`vars(creek_mcp.api.models)`) for `_WireModel` subclasses — the base excluded by identity, and deliberately **no module filter**, so a wire model defined elsewhere and merely imported into `models.py` still has to be registered. Two tests use it:

- `test_every_wire_model_subclass_is_registered_in_contract_models` — `assert _wire_model_subclasses(vars(api_models)) == CONTRACT_MODELS`. Dict equality, so it catches both directions: an unregistered class *and* a stale registry entry whose class no longer lives in the module.
- `test_wire_model_registry_check_catches_an_unregistered_subclass` — the non-vacuity guard, in the same style as the existing `test_admitted_ceiling_tier_sweep_is_not_vacuous`. A reflection guard fails *open* when its predicate stops matching (sweep finds nothing, equality holds, test stays green while the invariant rots), so this pins the negative case: a `_WireModel` subclass nobody registered is discovered, and the equality the test above asserts is exactly what breaks.

**RED evidence** (TDD, not retrofitted): with a deliberately-unregistered `_WireModel` subclass temporarily added to `models.py`, the new test failed with

```
E   AssertionError: assert {'Capabilitie...Signal'>, ...} == {'Capabilitie...Signal'>, ...}
E     {'TemporarilyForgottenModel': <class 'creek_mcp.api.models.TemporarilyForgottenModel'>}
```

The scaffold was then removed; `models.py` is unchanged apart from the docstring in item 2. Suite is 445 → 447, all green. The test module's "Twelve groups" enumeration was also updated so Group 1's own description covers the new totality property rather than just the happy-payload/unknown-key pair.

### 2. `NOT_FOUND` never emitted for a vault object — tracked as #1098

Filed **#1098** (`P2`, `follow-up`, `mcp`, `creek-tools`), blocked on #1074: an AST sweep over every `NOT_FOUND` construction site in `creek_mcp/`, checked against a pinned routing-layer allowlist, plus its own non-vacuity guard so an AST-walk bug can't make it pass by finding nothing. Agreed there's nothing to guard until #1074 mounts handlers, so no code beyond the pointer.

Per your "or at least a comment/issue reference", the pointer lives next to the invariant rather than only in the tracker — in the `creek_mcp/api/models.py` module docstring at the end of the `not_found` paragraph, and one sentence in the ADR's `not_found` re-scoping section.

### 3. Dedicated `ErrorEnvelope` extra-field test — skipped, already covered

Respectfully skipping this one: `ErrorEnvelope` extra-field rejection isn't only covered by the generic sweep. `test_error_envelope_refuses_to_echo_request_material` (`tests/test_adepthood_contract_models.py`) is already a dedicated, named, `ErrorEnvelope`-specific test that asserts `ValidationError` for seven concrete extra keys — `input`, `detail`, `received`, `body`, `content`, `entry_ref`, `tier` — and it sits directly under `test_error_envelope_has_exactly_three_fields`. A third test doing `{**ERROR_ENVELOPE_PAYLOAD, "unknown": ...}` would be a line-for-line duplicate of it, so adding it would cost readability rather than buy it. Happy to add it if you read that differently.

---

All of the above is in `ef839a5`. Gate 2 re-cleared after the changes: `./scripts/check-all.sh` exit 0, 12/12 checks, 7295 passed, 94.12% branch coverage. Nothing in `models.py` changed except the module docstring, so the committed fixture bundle and its manifest hashes are untouched.

Geoffe-Ga:
## Summary
PR #1095 ratifies the Adepthood↔Creek `/v1` HTTP application contract in a new accepted ADR (`docs/decisions/2026-07-31-adepthood-http-application-api.md`), backs it with a framework-free Pydantic wire vocabulary (`creek_mcp/api/models.py`, 16 models), and publishes a generated, byte-reproducible fixture bundle (`docs/contracts/adepthood-v1/`: 16 schemas + 28 examples + manifest + retry table) built by `creek_mcp/api/bundle.py`. No routes, no MCP behavior change, no new dependency — this is the vocabulary the eventual handlers (#1074+) will be mounted behind.

## Strengths
- **Privacy-oracle discipline is airtight and consistent end-to-end.** `not_found` is re-scoped to a routing-only code; every vault-object non-answer collapses to one `privacy_refused` with one imported constant message; `ErrorEnvelope` is `extra="forbid"` with exactly three fields and no format-string possibility (pinned by `test_error_message_is_a_constant_string`). The ADR's "oracle section" explicitly enumerates which codes are safe to expose and why, including the subtle point that `unavailable`/`internal_error` must stay *global-only*, never per-object.
- **`routed_tier` exposure is proven, not asserted.** `test_routed_tier_is_a_constant_function_of_the_declared_ceiling` empirically sweeps every admitted (ceiling, tier) pair and is explicitly annotated "if this ever fails, remove the field — don't relax the test." This is exactly the right way to pin a security-load-bearing invariant.
- **`INTIMATE` is unreachable in the type system**, not by a runtime branch — `WireTierCeiling` has exactly two members, verified against the real `server._REMOTE_ADMITTED_CEILINGS` rather than a hand-copied set.
- **The generated-bundle mechanism is genuinely anti-rot**: tests read the *committed* files from disk and assert byte-identity with a fresh `build_bundle()`, so the fixtures cannot silently drift from the models that describe them — including a self-referential blind spot (`CONTRACT_MODELS` completeness) fixed via reflection-based discovery in the follow-up commit, with a non-vacuity test for the guard itself.
- **Self-correcting.** The second commit ("Addresses the two non-blocking review items on PR #1095") shows the author already ran a review pass and closed both gaps before this reached me — the `not_found` enforcement caveat is now called out in both the module docstring and the ADR, with a tracking issue (#1098) for the structural guard once handlers exist.
- I spot-checked every cross-module claim (`CONTRACT_VERSION`, `ONTOLOGY_VERSION`, `GENERIC_ABOVE_CEILING_REASON`, `CARE_SIGNAL`, `CANONICAL_FREQUENCY_NAMES`, `CEILING_ROUTING_TIER`/`routing_tier`/`tier_allowed`, `_REMOTE_ADMITTED_CEILINGS`, `reflect._ALLOWED_KINDS`) against the actual source and all of them match exactly what the ADR/models/tests assert.

## Security Concerns
None found. 🟢 The whole design is oriented around *not* reintroducing the existence-and-rank oracle that issues #846/#970/#972/#1090 spent five issues collapsing, and it succeeds: no `retryable` boolean (would leak per-response nuance), no per-object global-error codes, no `intimate` value constructible on the wire in any position, no echo of caller/vault material in error envelopes.

## Problems
None blocking.

## Code Quality
- Non-blocking style note: `TierModel.ceilings: list[WireTierCeiling]` isn't constrained to exactly the two enum members or non-empty — low risk since it's only ever populated from the enum in `bundle.py`, but a stray hand-built payload elsewhere could in principle supply `["open"]` or `[]`. Not worth blocking on given the declarative, generation-only use today.
- The ADR is long (581 lines) but every section earns its length — this is reference material for a cross-repo contract, not incidental prose.

## CLAUDE.md Adherence
✅ Conventional commit messages (`docs(adr): ...`, `test(mcp): ...`). ✅ TDD workflow evident (tests + models + docs landed together, with a documented self-review follow-up). ✅ No route/behavior change outside the stated scope. Could not execute `./scripts/check-all.sh` in this sandboxed review environment, but CI's `Code Quality & Testing` jobs (3.11/3.12/3.13) were still `IN_PROGRESS` at review time — worth confirming they land green before merge, though nothing in the diff suggests they won't.

## Testing
Exceptional. `tests/test_adepthood_contract_models.py` is ~1360 lines organized into 12 explicit groups (happy-path/closed-model, closed enums, the wheel's ten required fields, no-echo error envelope, the #1082 retry table, bundle round-trip, ADR↔code version pin, the `routed_tier` non-oracle proof with a non-vacuity check, `essay_grounded` lockdown, structural `intimate` unreachability, care-signal fidelity, and layering/import-boundary AST sweeps). Tests read the *committed* artifacts rather than regenerating them, which is the correct anti-rot design — a test that rebuilt what it checks would catch nothing.

## Documentation
Thorough throughout: module docstrings state invariants and *why*, the bundle README explains vendoring/integrity guidance for the Adepthood consumer, and `docs/decisions.md` is a new, clearly-scoped ADR index with a helpful disambiguation note against the unrelated `creek-tools/docs/decisions.md`.



## Verdict: LGTM

Purely additive (new ADR, a framework-free Pydantic wire vocabulary, a generated fixture bundle, and 445+ tests) with zero production behavior change — no routes, no MCP changes, no new dependency. Every claim I spot-checked against the real code (CONTRACT_VERSION, ONTOLOGY_VERSION, GENERIC_ABOVE_CEILING_REASON, CARE_SIGNAL, CANONICAL_FREQUENCY_NAMES, CEILING_ROUTING_TIER, _REMOTE_ADMITTED_CEILINGS, reflect._ALLOWED_KINDS) matches exactly, the privacy-oracle reasoning is sound and well-tested, and a prior review round was already self-addressed in the second commit. No blocking issues found; would merge as-is once CI (still in progress at review time) goes green.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5147325317 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Three atomic fixes to the Ralph orchestrator's local tooling: (1) restores the executable bit on `watch-pr.sh`/`test_watch_pr.sh`, which #1092 shipped as `100644` and which broke every path-based invocation; (2) adds a new `changes-requested` token to `pr-ready.sh` so a fresh non-LGTM verdict (`CHANGES_REQUESTED`/`COMMENTS`) is distinguished from "no verdict yet," fixing a live-observed bug where the watcher slept out its full timeout on a green-CI + fresh-COMMENTS PR (#1095); (3) consolidates two workflows that both triggered on `scripts/ralph/**` into one ("Ralph Tooling Tests"). All changes are RED-first TDD with generous test additions.

## Strengths
- The core logic change in `pr-ready.sh` (scripts/ralph/pr-ready.sh:389-394) is minimal and precise: it inserts the new `changes-requested` branch exactly where `awaiting-review` used to fire, and I traced it by hand against every listed test case (missing verdict, stale LGTM, stale non-LGTM, fresh LGTM, fresh non-LGTM, and a malformed verdict flag) — all four states classify correctly, including the fail-closed behavior on a malformed `verdict_lgtm` value (never "true" nor "false").
- `watch-pr.sh` required **no logic change** — confirmed directly: `IN_FLIGHT_TOKENS=(pending awaiting-review)` (scripts/ralph/watch-pr.sh:46) is untouched, and the new token naturally falls outside it. The PR body's claim to this effect checks out.
- `test_exec_bits.sh` reads the git *index* mode via `git ls-files -s`, not the working-tree stat bits — the right check, since it can't be defeated by an unstaged local `chmod +x`. It also pins the four entry points by name so a rename can't dodge the glob, and guards against a vacuously-passing empty listing.
- Precedence is well-reasoned and pinned by tests: opt-out → CI state (pending/ci-failed) → verdict, so `changes-requested` never outranks an in-progress or failing CI run.
- Excellent self-documentation — the script header, `FLEET.md`, `RECAP.md`, and `ralph-tick.md` are all updated consistently and in the same PR, so the docs won't drift from the code they describe.

## Security Concerns
None identified. 🟢 No new external input handling, no injection surface changes — the new branch only inspects a bash string already validated to be a jq `tostring` boolean.

## Problems
None blocking.

## Code Quality
- 🟢 Non-blocking: the workflow rename (`Ralph Fleet Tests`/`Ralph Recap Tests` → `Ralph Tooling Tests`, job id `recap-tests` → `ralph-tooling-tests`) could break a branch-protection required-status-check rule if one is pinned to the old check name(s) — I wasn't able to query branch protection settings in this session to confirm either way. Worth a quick check post-merge that "Ralph Tooling Tests" (or whatever context string GitHub derives) is registered as a required check if the old ones were.
- 🟢 Non-blocking: in `pr-ready.sh`'s header comment table, the new `changes-requested` row's wrapped continuation lines are indented one space short of the other entries — purely cosmetic.

## CLAUDE.md Adherence
✅ This PR touches only `scripts/ralph/`, `.github/workflows/`, and `.claude/commands/` — none of the Python-specific gates in `creek-tools/CLAUDE.md` (coverage ≥90%, mypy strict, ruff, docstring coverage) apply to this diff.
✅ Follows the repo's "Stay Green" workflow: RED-first tests documented and shown failing then passing in the PR body, shellcheck clean claimed, pre-commit green claimed, feature branch used (not a direct push to `main`).
✅ Conventional Commits style title (`fix(ralph): ...`).

## Testing
Strong test coverage for the behavioral change: `test_pr_ready.sh` adds the four-way verdict split (missing/stale/fresh-LGTM/fresh-non-LGTM) plus a malformed-verdict fail-closed case, real-jq regex cases for both `CHANGES_REQUESTED` and `COMMENTS` verdict spellings, and lazy-probe cases proving the new branch doesn't pay for the compare/review-gate probes it doesn't need. `test_watch_pr.sh` pins the exact wake-on-two-polls behavior for the new token. `test_exec_bits.sh` is a new, well-targeted regression test for the underlying #1096 bug (committed file mode), with a self-check against vacuous passes. I hand-verified the new conditional in `pr-ready.sh` against each of these cases and found no logic gap.

## Documentation
Thorough and consistent: `ralph-tick.md`, `FLEET.md`, and `RECAP.md` are all updated in lockstep with the code, including the token-precedence rationale and the workflow-consolidation rename. Script-level header comments in `pr-ready.sh` and `watch-pr.sh` are updated to match the new token set.



## Verdict: LGTM

The diff is entirely scoped to scripts/ralph/ tooling, workflows, and docs (no creek-tools Python source touched, so the Python quality gates in CLAUDE.md don't apply here); the new/changed logic in pr-ready.sh is correct and traced through by hand against every added test case, the exec-bit fix is verified directly against the diff's mode changes, and the workflow consolidation is coherent. Nothing here blocks merge — only a couple of non-blocking notes worth a glance.

Geoffe-Ga:
**CI**: 6/9 Green
**VERDICT**: LGTM
**Action**: pull comment 5147473289 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 12/12 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
PR #1095 ratifies the Adepthood↔Creek `/v1` HTTP application contract as an accepted ADR (`docs/decisions/2026-07-31-adepthood-http-application-api.md`), adds a framework-free Pydantic wire vocabulary (`creek_mcp/api/models.py`, 16 closed models/enums) and a deterministic fixture-bundle generator (`creek_mcp/api/bundle.py`), and publishes the generated bundle under `docs/contracts/adepthood-v1/` (16 schemas + 28 examples + manifest + retry policy). No routes are mounted, no MCP tool behavior changes, and no new dependency is introduced.

## Strengths
- **Oracle-avoidance reasoning is rigorous and consistent end-to-end.** The `not_found`/`privacy_refused` collapse, the ban on per-object `unavailable`/`internal_error`, and the `routed_tier`-is-a-constant-function proof are all stated in the ADR, encoded as type-level constraints (`WireTierCeiling` has only two members; `Literal[False]` for `essay_grounded`), and pinned by dedicated tests (`test_routed_tier_is_a_constant_function_of_the_declared_ceiling`, with an explicit "remove the field, don't relax the test" instruction).
- **Anti-rot mechanisms are real, not aspirational.** I independently verified two `manifest.json` sha256 entries (`retry-policy.json`, `schemas/ErrorEnvelope.schema.json`) against the actual committed file bytes — both matched exactly, corroborating the "committed bundle is byte-identical to a fresh `build_bundle()`" claim backing `test_committed_bundle_equals_a_fresh_build`.
- **Layering and framework-independence are enforced structurally**, not just by convention: AST checks for forbidden framework imports in `creek_mcp/api/`, and a check that `creek/` never imports `creek_mcp` (respecting #1032).
- **Error taxonomy is total and no-echo by construction**: `ERROR_STATUS`/`ERROR_MESSAGES`/`RETRY_POLICY` are all total over `ErrorCode` (tested), messages are constants with a pinned "no `{`" property, and `ErrorEnvelope` is `extra="forbid"` with exactly three fields.
- Excellent module/attribute-level docstrings that explain *why*, not just *what* — genuinely useful for a cross-repo consumer reading this cold.
- The `NotApplicableExample` treatment for structurally-unreachable fixture cells (care-escalation on non-`reflections` capabilities) is a clean way to keep the matrix rectangular without fabricating unreachable response shapes.
- Self-correcting review history is visible in the commits: a second commit closed a self-referential blind spot in the model-registry test (`CONTRACT_MODELS` totality) discovered during Gate 2.5, with a non-vacuity guard added alongside it.

## Security Concerns
🟢 LOW — None found in this diff. The privacy-oracle collapse (`not_found` vs `privacy_refused`, no per-object global-only codes) is the central security property of this contract and is well-reasoned, tested, and consistent with prior related issues (#846/#970/#972/#1090) referenced throughout. `INTIMATE` is unreachable at the type level rather than by runtime branch. No credentials, journal bodies, or intimate-tier content appear in any committed fixture (spot-checked; also test-enforced via `test_no_committed_fixture_carries_an_intimate_tier_value`).

## Problems
None blocking. One item worth flagging for the author's attention, not a merge blocker:
- CI was still `IN_PROGRESS` on all three Python-version "Code Quality & Testing" jobs and the `claude-review` check at the time of this review (only the CrawDad-quality and fleet-tooling jobs had completed with SUCCESS). Per this repo's Stay-Green workflow (CLAUDE.md §4), merging should wait for those to go green before Gate 3/4 are satisfied — this review is being conducted while that gate is still resolving.

## Code Quality
Non-blocking observations:
- `models.py` and `bundle.py` are large (933 and 628 lines) but the size is inherent to a 16-model closed vocabulary plus a 4×7 fixture matrix with per-cell rationale in docstrings — splitting further would likely hurt cohesion more than it helps.
- The `_care_example`/`_error_example`/`_example_for` dispatch in `bundle.py` is small and readable; no complexity concerns.
- Nothing else stood out as worth changing.

## CLAUDE.md Adherence
✅ Scripts-only tooling claimed in PR description (`./scripts/check-all.sh`, 12/12 checks, 7293 tests, 94.12% branch coverage, per-file gate passed) — consistent with the repo's documented quality gates (≥90% coverage, ≥95% docstring coverage, mypy strict, ruff clean).
✅ TDD workflow described in the PR body (test suite authored alongside the models) matches Gate 1 of the Stay-Green workflow.
✅ No shortcuts observed: no `# noqa`/`# type: ignore` bypass patterns seen in the reviewed files; `essay_grounded` was deliberately tightened from a runtime-only validator to a `Literal[False]` type constraint specifically to avoid a schema-looser-than-server bug, which is exactly the "fix root cause" ethos CLAUDE.md asks for.
✅ Commits follow Conventional Commits style (`docs(adr): ...`, `test(mcp): ...`).
⚠️ Gate 3 (CI green) not yet confirmed complete at review time — see Problems above.

## Testing
`tests/test_adepthood_contract_models.py` (1351 lines, ~90 test functions/parametrized cases per the function listing) covers: model validation and extra-field rejection for every registered model, closed-enum membership mirrored against runtime constants, the wheel's ten-frequency completeness and canonical field order, error-table totality, no-echo/no-template-placeholder invariants on error messages, the `routed_tier` non-oracle proof (with a non-vacuity guard), `essay_grounded`'s `Literal[False]` enforcement, INTIMATE unconstructibility, care-signal fidelity against the runtime guardrail constant, the bundle round-trip (committed bytes == fresh build), the ADR↔code version pin, and AST-based layering/framework-import checks. This is well above what would be required to consider testing sufficient for a docs/vocabulary PR of this scope — not blocking.

## Documentation
The ADR is exceptionally thorough: it documents the versioning rule, the full error taxonomy with an explicit "which codes are safe to expose vs which must collapse" oracle-avoidance analysis, the six-divergence resolution table against Adepthood's independently-inferred contract, the HTTP-framework decision (Starlette over FastAPI, with four concrete reasons), and an honestly-stated residual (timing side-channel, terminal-vs-actually-recoverable blur for #1082). The bundle's own README correctly distinguishes itself from the ADR ("read that document first; this README is the manual for the bundle") and documents JSON key ordering pitfalls (alphabetical serialization vs. canonical F1–F10 order) that a naive consumer could otherwise trip over. `docs/decisions.md` is created as a proper ADR index with a disambiguation note against the unrelated `creek-tools/docs/decisions.md`. Nothing incomplete or stale found.



## Verdict: LGTM

The change is docs/ADR + a new, framework-free, declarative Pydantic vocabulary and generator with no route wiring and no behavior change to existing MCP tools; the ADR's central privacy-oracle reasoning is sound and is backed by a very thorough test suite (enum/table totality, round-trip bundle-vs-model equality, AST layering/framework-import guards, non-vacuity guards). I independently verified two manifest sha256 entries against the actual committed files and both matched exactly, corroborating the "generated, byte-identical" claim. CI was still in progress at review time (not yet green) — worth confirming before merge, but nothing in the diff itself blocks approval.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5147589077 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

Issue #1103, opened 2026-07-31
chore(tests): refresh the stale server.py line reference in the #1073 characterization pins


Geoffe-Ga:
## What

`creek-tools/tests/test_mcp_remote.py` has a section banner citing
`server.py:120-131` for the `creek.purge.*` ceiling carve-out. After #1073
moved the cap into `creek_mcp/policy.py`, that carve-out documentation lives at
roughly `creek_mcp/server.py:162-169`.

## Why it was NOT fixed in #1073

Deliberately. Those tests are characterization pins committed **before** the
extraction (commit `97d2c90`), and the evidence that the extraction changed no
behaviour is that the file is **byte-identical** across the extraction commit:

```
git diff --stat 97d2c90 -- creek-tools/tests/test_mcp_remote.py   # empty
```

Editing the file inside #1073 — even to correct a comment — would have
destroyed that property. The Gate 2.5 reviewer flagged the stale reference and
explicitly recommended leaving it, then fixing it in a later, unrelated commit.

## What to do

Update the line reference in the section banner comment. Prefer citing the
symbol (`_BoundedFastMCP.call_tool`'s docstring / `creek_mcp.policy`) rather
than a line range, so it cannot go stale again.

Comment-only. No test logic, no assertion, and no pin may change.

## Acceptance criteria

- [ ] The reference names a symbol, not a line range.
- [ ] No assertion, parametrization or helper in the file is modified.
- [ ] `cd creek-tools && ./scripts/check-all.sh` exit 0.

Refs #1073

Issue #1100, opened 2026-07-31
fix(mcp): refuse a blank consumer client_id at the verifier boundary, not mid-dispatch


Geoffe-Ga:
## Context

Split out of #1073 (epic #1071) so that refactor could stay a strictly zero-delta move.

`creek_mcp.policy.effective_consumer` audits a remote call under
`identity.consumer`. If that consumer is an **empty string**, the call is
audited under `consumer: ""` — an entry that looks like a successful
attribution but names nobody.

## Why #1073 did not fix it there

The extraction initially raised `ValueError` on a falsy consumer. That was
narrowed to `is None` only, for two reasons:

1. **It was a behaviour change.** `(is_remote=True, consumer="")` was
   representable before the extraction and returned `""`. A pure refactor may
   not change it. (`(is_remote=True, consumer=None)` was *not* representable —
   the old code read `token.client_id`, typed `str` — so raising there is free
   and fail-closed, and #1073 keeps that.)
2. **It was the wrong place.** `effective_consumer` is called *inside* the tool
   wrapper at every `consumer=_effective_consumer(...)` site in `server.py`.
   Raising there escapes into the FastMCP tool surface and skips the audit
   entry the call was about to write — losing the trail for exactly the call
   whose attribution is suspect. `creek_mcp/auth.py` records the same hazard
   for the elevated gate: an exception on that path "would escape into the
   FastMCP tool surface and skip the audit entry every purge call is required
   to write."

## Reachability

Not reachable through the shipped path. `load_consumer_tokens`
(`creek_mcp/remote_auth.py:139`) requires `name.strip()` to be non-empty, so a
blank consumer name is dropped at load. The only way in is a **custom
`token_verifier=` passed to `creek_mcp.server.build_server`**, which accepts an
arbitrary `TokenVerifier` — i.e. an embedder, not a wire caller. That is why
this is P2 and not P0.

## What to do

Refuse a blank/whitespace-only `client_id` at the **verifier boundary**, before
dispatch, so the refusal costs no audit record:

- `ConsumerTokenVerifier.verify_token` returns `None` (→ 401) rather than an
  `AccessToken` with a blank `client_id`;
- and/or `build_server` validates the supplied `token_verifier`'s output shape.

Fail closed and **silently** — like `auth.is_elevated`, the deny must hand a
possibly-hostile caller no oracle about server configuration.

## Acceptance criteria

[...]

Issue #1071, opened 2026-07-31
epic(api): serve Adepthood over a versioned /v1 HTTP/JSON API sharing one behavioral implementation with MCP


Geoffe-Ga:
# Epic: one behavioral Creek, purpose-built adapters

> **P1 — MVP Track B.** Re-tiered from P0 on 2026-08-06 (see the pinned comment);
> the intra-epic dependency chain is unchanged. P0 is now reserved for fleet/CI
> defects that block delivery of everything. Read "Ralph sequencing" at the bottom
> before changing any label on a sub-issue.

## Problem

Creek exposes its Adepthood-facing capabilities (`creek.handshake`,
`creek.journal`, `creek.reflect`, `creek.wheel`) only through MCP. MCP is an
*agent* protocol — discoverable tools, prompts, resources, sampling,
elicitation, session lifecycle. Adepthood is not an agent. It is a
deterministic application that knows at build time exactly which four
operations it needs, and it pays for MCP's agent affordances in coupling and
breakage without using any of them.

**That coupling has already broken the integration, silently and completely.**
Verified at creek-vault `de33c3e` against adepthood `fbc529d`:

| Capability | Adepthood sends | Creek's registered tool accepts | Works? |
|---|---|---|---|
| `creek.handshake` | `privacy_tier_ceiling` (`creek_vault_client.py:167-174`) | `privacy_tier_ceiling` (`server.py:318-330`) | ✅ |
| `creek.journal` | `content, external_id, timestamp, tier, privacy_tier_ceiling` (`creek_vault_client.py:282-297`) | same five (`server.py:360-377`) | ✅ |
| `creek.reflect` | `consumer, body, tier_ceiling` (`creek_vault_client.py:277-279`) | `content, entry_ref, privacy_tier_ceiling` (`server.py:332-346`) | ❌ zero params match |
| `creek.wheel` | `consumer` (`creek_vault_client.py:426`) | `privacy_tier_ceiling` only (`server.py:349-358`) | ❌ unexpected param |

Response shapes diverge just as hard:

- `wheel_tool` returns `{status, tool, tier_ceiling, total_classified, unclassified, wheel: {F1..F10: {name, count, share}}}` (`creek-tools/creek_mcp/tools/wheel.py:103-110`). Adepthood validates a `WheelBalanceResponse` of `{aspects: [{stage_number, aspect, fullness}]}`.
[...]

PR #1104, opened 2026-07-31
refactor(mcp): extract transport-neutral admission, consumer-identity and audit policy out of the MCP server

### LINKED ISSUE #1073: refactor(mcp): extract transport-neutral admission, consumer-identity and audit policy out of the MCP server
## Role

Security-minded refactorer. This issue writes no new feature and adds no route.
Its entire purpose is to make it *impossible* for the HTTP adapter in #1071 to
become a privacy bypass, and it must land before any capability route is real.

## Problem

Two load-bearing security behaviors are implemented inside MCP-specific
machinery and are therefore invisible to any other adapter:

**1. The remote INTIMATE cap lives in a `FastMCP` subclass.**

`creek-tools/creek_mcp/server.py:103-144`:

```python
class _BoundedFastMCP(FastMCP):
    async def call_tool(self, name, arguments):
        if _current_access_token() is not None:
            raw = arguments.get("privacy_tier_ceiling", TierCeiling.OPEN.value)
            try:
                requested = TierCeiling(raw)
            except ValueError:
                requested = None
            if requested not in _REMOTE_ADMITTED_CEILINGS:
                return refusal_response(...)
        return await super().call_tool(name, arguments)
```

The "is this call remote?" test is `_current_access_token() is not None`
(`server.py:84-92`), a wrapper over the **MCP SDK's** request-scoped
`get_access_token()`. An HTTP request carries no MCP access token, so for a
`/v1` caller this returns `None` and the entire cap silently does not apply.
`intimate` and `all` would be admitted over the network.

**2. The audited consumer identity has the same dependency.**

`server.py:94-101`:

```python
def _effective_consumer(default: str) -> str:
    token = _current_access_token()
    return token.client_id if token is not None else default
```

Every `tools/*.py` function takes `consumer=` and writes it to `MCPAuditLog`
(e.g. `tools/journal.py:212-220`). Under HTTP, `_effective_consumer` would
always fall through to the process-global `CREEK_MCP_CONSUMER`, so every
Adepthood call would be audited as the local operator. The audit log would stop
being able to answer "which consumer did this."

There is no test today that would catch either regression for a non-MCP caller,
because there is no non-MCP caller.

## Goal

One transport-neutral policy module that answers, for a call arriving over
*any* adapter: who is the consumer, is this call remote, and is the requested
ceiling admitted? MCP is refactored to call it; `/v1` will call the same
functions. Zero behavior change for MCP, proven by characterization tests
written **before** the extraction.

## Architectural context

HTTP is the Adepthood adapter, MCP the agent adapter, both over one
implementation. This issue is the seam that makes "one implementation" true for
*policy*, not just for the tool bodies. `tools/*.py` already are the
transport-neutral service layer — do not re-extract them.

## Proposed tracer

A single policy call answers `(caller identity, is_remote, requested ceiling) →
admitted | refusal`, and `_BoundedFastMCP.call_tool` becomes a thin adapter that
supplies MCP's answers to those three questions and acts on the verdict.

## Tasks

**Characterization tests first (RED before any move):**
- Pin current MCP behavior for: remote + `open` → admitted; remote + `personal`
  → admitted; remote + `intimate` → refused before dispatch; remote + `all` →
  refused; remote + unrecognised ceiling string → refused; stdio + `intimate` →
  admitted (local is unaffected).
- Pin that a remote call is audited under the bearer's `client_id` and a stdio
  call under `CREEK_MCP_CONSUMER`.
- Pin the documented `creek.purge.*` carve-out (`server.py:120-131`): those
  tools declare no `privacy_tier_ceiling`, so the gate is a deliberate no-op for
  them and they are gated by `CREEK_MCP_ELEVATED_TOKEN` instead. The extraction
  must not accidentally start gating purge, and must not stop
  `creek_mcp.auth.is_elevated` from being the real gate.

**Extraction:**
- New `creek_mcp/policy.py` (or `creek_mcp/admission.py`) holding:
  - a `CallerIdentity` value type carrying `consumer` and `is_remote`;
  - `admitted_ceiling(identity, requested) -> TierCeiling | Refusal`, owning
    `_REMOTE_ADMITTED_CEILINGS` and the fail-closed unrecognised-value branch;
  - `effective_consumer(identity, default)`.
- Move `_REMOTE_ADMITTED_CEILINGS` and the refusal wording there; `server.py`
  imports rather than duplicates.
- `_BoundedFastMCP.call_tool` and `_effective_consumer` become adapters that
  build a `CallerIdentity` from `get_access_token()` and delegate.
- The policy module must not import `mcp`, `fastmcp`, or any SDK symbol. Add a
  test asserting that.

**Docs:** update `creek-tools/docs/mcp.md` and the module docstring to say the
cap is now adapter-independent policy, and why.

## Acceptance criteria

- [ ] Every characterization test above passes identically before and after the
      extraction; the diff moves code, it does not change verdicts.
- [ ] `creek_mcp/policy.py` has no import of `mcp`/`fastmcp`, asserted by a test.
- [ ] `admitted_ceiling` refuses `intimate`, `all`, and any unrecognised string
      for a remote identity, and admits them for a local identity — driven by
      the `CallerIdentity`, not by `get_access_token()`.
- [ ] A test constructs a *remote* `CallerIdentity` with no MCP context at all
      (simulating an HTTP caller) and asserts `intimate` is refused. This is the
      test that would have caught the bypass.
- [ ] `effective_consumer` returns the remote identity's id when remote and the
      configured default when local, with no SDK dependency.
- [ ] `creek.purge.*` remains gated by `CREEK_MCP_ELEVATED_TOKEN` and is not
      newly ceiling-gated; pinned by test.
- [ ] `./scripts/test.sh --all` green — no agent-visible MCP change.

## Verification

```bash
```

Coverage ≥90% branch, interrogate ≥95%, mypy strict, ruff clean, xenon/radon
clean. No `# type: ignore`, no `# noqa`.

- Remote INTIMATE stays refused **before dispatch**, so no intimate content is
  read for a remote caller.
- Fail closed: an unrecognised ceiling value is refused, never coerced to `open`.
- The refusal reason names no fragment, no tier of any specific content, and no
  credential.
- Local (stdio) behavior is unchanged in every respect.

## Files likely affected

- create `creek-tools/creek_mcp/policy.py`
- create `creek-tools/tests/test_admission_policy.py`
- modify `creek-tools/creek_mcp/server.py` (lines ~76-144)
- modify `creek-tools/docs/mcp.md`

## Dependencies

Depends on #1072 (the ADR names the auth/consumer policy this module encodes).
**Blocks** the `/v1` skeleton and all three capability verticals — none of them
may be started before this lands. Part of #1071.

## Budget escape hatch

If the `CallerIdentity` refactor touches more of `server.py` than expected,
extract `admitted_ceiling` alone (the security-critical half) and leave
`effective_consumer` for a follow-up — but do not start the HTTP skeleton until
both are extracted, because auditing every HTTP call as the local operator is
its own defect.

Geoffe-Ga:
## Summary

The remote tier-ceiling cap and the audited consumer identity were bound to MCP
middleware. `"is this call remote?"` meant `_current_access_token() is not None`
— a fact about one transport's request context, not about the network. An HTTP
`/v1` caller carries no MCP access token, so it would have read as **local**:
the cap that keeps intimate content off the network would silently not have
applied, `intimate` and `all` would have been admitted over the wire, and every
Adepthood call would have been audited as the local operator.

Both answers now live in `creek_mcp/policy.py`:

- `CallerIdentity(consumer, is_remote)` — a value type the **adapter** asserts.
  Neither field has a default, because `is_remote: bool = False` would make a
  forgotten field fail *open*.
- `admitted_ceiling(identity, requested) -> Admission | Refusal` — owns
  `REMOTE_ADMITTED_CEILINGS` and the fail-closed unrecognised-value branch.
- `effective_consumer(identity, default)`.

`_BoundedFastMCP.call_tool` and `_effective_consumer` become thin adapters that
supply MCP's answers and render the verdict in MCP's vocabulary. `/v1` keeps
expressing the same boundary **structurally** through `WireTierCeiling`'s two
members — this module is the shared reasoning, not a runtime check a handler
could forget to call.

The module imports no MCP SDK and no web framework. Importing it loads neither
`mcp` nor `creek_mcp.server`, verified by running it:

```
$ python -c "import creek_mcp.policy, sys; print('mcp loaded?', 'mcp' in sys.modules)"
mcp loaded? False
server loaded? False
```

I took the full extraction rather than the issue's budget escape hatch:
deferring `effective_consumer` would have left #1075 importing a private symbol
from an SDK-importing module into an HTTP handler — the exact coupling this
issue exists to remove.

## Behaviour-preservation evidence

Not "the tests pass" — tests can move with the code. Two independent artifacts:

**1. The pins landed first, in their own commit, and did not change.**
`97d2c90` added 106 characterization tests that pass against *unmodified*
production code. Across the extraction commit:

```
$ git diff --stat 97d2c90 -- creek-tools/tests/test_mcp_remote.py
(empty)
```

They pin things that were previously unpinned: the refusal payload key-for-key
including `tier_ceiling == "open"` (the refusal never echoes the requested
ceiling back), the reason against a **hard-coded literal** rather than the
imported constant, and identity re-resolution per call.

**2. A 1215-row decision table, captured before and diffed after.**
The full cross-product of (local/remote × 20 ceiling shapes — garbage, wrong
case, whitespace, `None`, non-strings, unhashables × 15 tools), driven through
`server.call_tool`, captured against the pre-refactor tree and re-run after:

```
$ diff decision_table_BEFORE.json decision_table_AFTER.json
*** IDENTICAL ***
```

**Did any decision change? No — but the probe caught one that nearly did.** An
initial draft raised `ValueError` for a remote identity with an *empty*
`client_id`, where the old code returned `""`. All 625 tests passed; only the
table diff caught it. It is narrowed to `is None` only — a row the old code
could not represent (`AccessToken.client_id` is typed `str`), so no prior
behaviour exists to preserve and the fail-closed answer is free. Refusing blank
ids mid-dispatch was also the wrong *place*: `effective_consumer` runs inside
the tool wrapper, so raising escapes into the FastMCP tool surface and skips
the audit entry the call was about to write — the hazard `creek_mcp/auth.py`
already documents. That belongs at the verifier boundary: **#1100**.

## Audit emission

**Deliberately did not move.** It stays at the ~24 sites in `tools/*.py` that
receive `consumer=` top-down. `policy.py` opens no file and emits nothing, so
there is no second emission site free to diverge.

**Can a caller reach an admitted resource without emitting its audit record?
No.** `call_tool` either returns a refusal (touching no vault) or delegates to
`super().call_tool`; all 23 closures still thread
`consumer=_effective_consumer(consumer)` into the tool functions, which still
own emission. The real risk here is not a *missing* record but a **forged**
one: `get_access_token()` is request-scoped while `build_server`'s `consumer`
is bound once, so hoisting the identity would pin every later call to the first
caller — and from a stdio-built identity, uncap every remote call. Verified on
current code and pinned: one server, two bearers →
`['adepthood', 'crawdad', 'env-default']`.

## New reachability

`admitted_ceiling`/`effective_consumer` are now importable, where the cap was
previously reachable only through `FastMCP.call_tool`. Assessed: `policy` has
exactly one importer today (`server.py`); `api/models.py` and `tools/journal.py`
reference it in **docstrings only**. `_BoundedFastMCP.call_tool` remains the
single MCP chokepoint with no path around it. `creek/` still cannot reach any
of it (#1032), covered automatically by the existing parametrized guard.

## Gate ordering — untouched

`journal.py`'s diff is **one docstring paragraph**; `_refuse_unadmitted_overwrite`
(:396) stays above `_stage_entry` (:408) and `run_ingest` (:415), so it remains
above `write_fragment_idempotent`'s content-hash shortcut and `PRIVACY_REFUSED`
stays terminal for an unchanged re-sync (#970/#1082). `redact.py` is **not in
the diff at all** — its whole-call gate is untouched. The refusal reason remains
one non-interpolated constant for every refused input (#1090).

## Two judgement calls worth a reviewer's eye

- **`docs/contracts/adepthood-v1/` is regenerated** (7 schemas + manifest). The
  `WireTierCeiling` docstring is embedded in the published `description` fields,
  so a stale pointer would have shipped to a cross-repo consumer. The diff is
  descriptions + sha256s only — no enum member, no required field, no
  `contract_version`/`ontology_version`. No contract minor is owed: the ADR owes
  the bump on the first **wire-shape** change.
- **`REMOTE_ADMITTED_CEILINGS` is public and NOT re-exported** from `server.py`;
  three just-merged #1095 pins were re-homed. The diff changes only *where* the
  frozenset comes from — **no assertion changed**.

Two review findings were fixed rather than filed: `_parse_ceiling` narrows to
`str` before parsing, and `call_tool` dispatches on the positive `Admission`
verdict instead of "not a `Refusal`", so a future third verdict fails closed
**and** fails mypy.

## Test plan

- `./scripts/check-all.sh` → **exit 0**, 12 passed / 0 failed
- `./scripts/test.sh --all` → **7447 passed**, 5 skipped
- `creek_mcp/policy.py` → **100% line and 100% branch** coverage; pylint
  **10.00/10**; interrogate **100%**; radon rank **A** throughout
- The acceptance-criteria test exists: a **remote `CallerIdentity` built with no
  MCP context, no monkeypatch and no `creek_mcp.server` import** refuses
  `intimate` — the test that would have caught the bypass
- `creek.purge.*` verified still elevated-gated, not ceiling-gated, and still
  auditing under the bearer
- Extended-lint's `vulture`/`shellcheck` failures are pre-existing and outside
  `check-all.sh`: I changed no shell script, and vulture emits identical
  60%-confidence noise on long-merged `tier_ceiling.py`

Follow-ups filed rather than smuggled in: **#1100**, **#1103**.

Closes #1073
Refs #1071


Geoffe-Ga:
## Summary
PR #1104 extracts the MCP server's remote tier-ceiling cap and audited-consumer-identity logic out of `creek_mcp/server.py` into a new transport-neutral `creek_mcp/policy.py`. The motivation is real and well-argued: the old rule keyed "is this call remote?" off `_current_access_token() is not None`, an MCP-transport fact that the upcoming `/v1` HTTP surface (#1074) would never set — silently letting `intimate` content past the network cap and misattributing audit records. This PR closes that gap *before* it could be introduced, not after.

## Strengths
- **Fail-closed-by-construction design.** `CallerIdentity(consumer, is_remote)` has no field defaults specifically so a forgotten `is_remote` on a future adapter fails loudly (`TypeError`) rather than silently defaulting to "local" (uncapped). It's frozen to close a TOCTOU window between assertion and decision.
- **Dispatch is on the positive verdict.** `call_tool` now checks `isinstance(verdict, Admission)` rather than `not isinstance(verdict, Refusal)`, so a hypothetical future third verdict type fails closed *and* fails mypy instead of silently uncapping calls.
- **No oracle/side-channel leakage.** `REMOTE_CEILING_REFUSAL_REASON` is one module constant, never built from caller input; tests pin that it contains no `{`, `}`, `%`, digits, or path separators.
- **Exceptional behavior-preservation evidence.** The characterization pins in `test_mcp_remote.py` were committed first, against *unmodified* code, and are proven byte-identical (`git diff` empty) across the extraction commit. They specifically pin the highest-risk failure mode: identity must be re-resolved **per call**, never cached at `build_server` time (which would forge audit attribution and could uncap every later remote call from one stray local call).
- **Consistency verified across the diff.** Every old reference to `_REMOTE_ADMITTED_CEILINGS` (server.py, tests, docs, ADR, regenerated JSON-schema docstrings) was updated together — no dangling references to the old name anywhere in the diff.
- **Judgment calls called out explicitly** in the PR body (contract regeneration, why `REMOTE_ADMITTED_CEILINGS` isn't re-exported from `server.py`), making reviewer verification fast.

## Security Concerns
🟢 None found. If anything, this PR *removes* a latent security gap (an intimate-content bypass that would have shipped invisibly the moment `/v1` handlers were mounted). The one deliberately-scoped non-change — an empty (vs. `None`) remote `consumer` still returns `""` rather than raising — is explicitly justified as preserving existing behavior and is tracked as a follow-up (#1100) at the correct layer (verifier boundary, not mid-dispatch).

## Problems
None blocking.

## Code Quality
- Docstrings are very dense/verbose (e.g., `policy.py`'s module docstring and several function docstrings run to 20+ lines). This matches the established convention already present in `creek_mcp/server.py` and is necessary to clear this repo's 95% interrogate gate, so it's not out of place — just worth noting it's a heavy read for reviewers unfamiliar with the codebase's documentation style.
- Minor documentation-precision nit (not a functional issue): `policy.py`'s `_parse_ceiling` docstring frames "parse before frozenset membership test" as fixing a `TypeError`-on-unhashable-input hazard relative to the old code. In fact the old code already parsed `raw` into a `TierCeiling` member (or `None`) via `TierCeiling(raw)` *before* the membership test, and CPython's `Enum.__new__` itself falls back to a linear scan for unhashable values rather than raising `TypeError`. So the old code was never actually at risk here — the new code's behavior is still correct and provably equivalent, this is just slightly overstated framing in a comment.

## CLAUDE.md Adherence
✅ Coverage — PR body reports `policy.py` at 100% line/branch coverage (self-reported; CI still finishing at review time, see Testing below).
✅ Linting/typing — reported ruff-clean, mypy-strict-clean, pylint 10.00/10, radon rank A.
✅ Docstrings — reported interrogate 100%; every public symbol in `policy.py` has a full docstring with Attributes/Args/Returns/Raises as applicable.
✅ Patterns — TDD workflow evident (characterization pins landed in an earlier commit before the extraction); follows the repo's existing MCP-adapter/policy-module separation style.
✅ Commits — issue-linked (`Closes #1073`, `Refs #1071`), follow-ups filed rather than smuggled in (#1100, #1103).

## Testing
Outstanding. Two independent verification artifacts beyond the new unit tests: (1) the pre-existing characterization suite proven unchanged across the extraction commit, and (2) a 1,215-row before/after decision-table diff (local/remote × 20 ceiling shapes × 15 tools) proven identical — and the PR body candidly notes this table caught a near-miss behavior change during development (an `empty client_id` edge case) that 625 passing tests alone had missed. `test_mcp_policy.py` additionally uses AST-based import guards (with negative/non-vacuity companions) to structurally enforce that the policy module can never import the MCP SDK. This is well above the bar for "sufficient testing."

## Documentation
Docs (`docs/mcp.md`), the ADR, and the regenerated `/v1` contract schemas/docstrings are all updated consistently with the code change — verified no stale references to the removed `server._REMOTE_ADMITTED_CEILINGS` symbol remain anywhere in the diff. The PR body itself is a model of a security-refactor writeup: it states the vulnerability being pre-empted, the invariants preserved, and explicitly flags the two judgment calls a reviewer should double-check.

**Note:** at review time, 3 of the 6 `Code Quality & Testing` CI matrix jobs and `claude-review` were still `IN_PROGRESS` (nothing had failed); `CrawDad Quality` jobs were green. Worth a final CI glance before merge, though nothing in the diff itself gives reason to expect a failure.



## Verdict: LGTM

Pure, behavior-preserving extraction of security-critical policy logic with exceptionally thorough characterization testing (634 new pin lines proving byte-identical behavior pre/post-refactor, plus 709 lines of isolated policy tests with AST import guards). No correctness, security, or CLAUDE.md-adherence issues found; the one thing I flagged is a trivial documentation-precision nit in a docstring, not something worth blocking or even asking the author to revisit.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5148108859 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Extracts the remote tier-ceiling admission cap and the per-call audited-consumer identity out of `creek_mcp/server.py` into a new transport-neutral module, `creek_mcp/policy.py`. The motivating bug: the old cap keyed "is this remote?" off `_current_access_token() is not None`, an MCP-specific fact, so a future `/v1` HTTP caller (no MCP token) would have silently read as *local* and bypassed the cap that keeps `intimate` content off the network. `CallerIdentity.is_remote` is now a bool the adapter must assert explicitly. `server.py`'s `_BoundedFastMCP.call_tool` and `_effective_consumer` become thin adapters over `policy.admitted_ceiling` / `policy.effective_consumer`.

## Strengths
- **Behaviour-preservation evidence goes well beyond "tests pass."** The characterization tests (`test_mcp_remote.py`) landed in a separate commit *before* any production code moved, and are byte-identical (`git diff` empty) across the refactor commit — tests that moved with the code would prove nothing, and this proves they didn't move.
- A 1215-row decision table (local/remote × 20 ceiling shapes × 15 tools) was captured pre-refactor and re-diffed post-refactor as identical, including the purge carve-out and per-call audit attribution.
- `CallerIdentity` is a frozen dataclass with **no field defaults** — `is_remote: bool = False` would fail *open* for a forgotten field, which is exactly the class of bug this PR fixes. Frozen avoids a TOCTOU window between assertion and decision.
- `REMOTE_CEILING_REFUSAL_REASON` is a single non-interpolated constant, with tests pinning that it contains no `{`, `}`, `%`, path separators, or digits — closing an existence/rank oracle the refusal could otherwise become (#1090).
- `_parse_ceiling` narrows to `str` *before* any frozenset membership test, so an unhashable `[]`/`{}` from a caller can't turn a refusal into a `TypeError`/500 — and this is explicitly tested (`test_unhashable_requests_are_decided_without_hashing`).
- `test_mcp_policy.py` uses AST-based import guards (with deliberate negative/vacuity-check companions) to structurally prevent `policy.py` — and even the test file itself — from ever importing the MCP SDK or `creek_mcp.server`, which is the actual invariant this refactor needs to hold over time.
- Two review findings were fixed inline rather than filed as follow-ups (str-narrowing before parse; dispatching `call_tool` on the positive `Admission` verdict instead of "not a Refusal", so a future third verdict fails closed *and* fails mypy).
- Scope discipline: genuinely out-of-scope follow-ups (blank `client_id` refusal, a stale line reference) were filed as #1100/#1103 rather than smuggled in.
- Docs (`docs/mcp.md`, module/attribute docstrings, the ADR) and the regenerated `/v1` contract schemas were all updated consistently with the code move; the schema diff is description text + sha256 only, correctly assessed as not requiring a contract version bump.

## Security Concerns
None blocking. 🟢 This is a security-hardening refactor, not a weakening: it removes an MCP-specific assumption (`token is not None` ⇒ remote) that would otherwise have silently under-enforced the `intimate`-content cap for a future transport. The new module keeps: fail-closed unrecognised-ceiling handling for remote callers, per-call (never cached) identity resolution — which is the highest-consequence failure mode and is explicitly pinned (`test_consumer_identity_is_resolved_per_call_not_once_per_server`, the remote→local and local→remote re-evaluation pins) — and an unchanged, non-leaking refusal payload.

## Problems
None found that block merge.

## Code Quality
- `effective_consumer`'s `ValueError` branch (remote identity with `consumer is None`) is currently unreachable from the one production adapter (`_caller_identity()` in `server.py`), since `AccessToken.client_id` is typed `str`. This is intentional, documented defensive-by-contract behaviour for future adapters (e.g., a custom `token_verifier`), and it's exercised by its own test — not a defect, just worth knowing it's presently dead code on the only live call path.
- The extraction's practical value is fully realized only once #1074 mounts the `/v1` HTTP handlers that reuse `policy.py`; today `policy` has exactly one importer. This is explicitly acknowledged in the PR description as the reason for doing the full extraction now rather than the issue's narrower "budget" option (to avoid a future HTTP handler importing a private symbol from an SDK-importing module).

## CLAUDE.md Adherence
✅ TDD followed at the commit-history level (characterization-test commit lands before the refactor commit — Gate 1's RED/pin step made verifiable via git history rather than asserted).
✅ Conventional Commits (`test(mcp): …`, `refactor(mcp): …`).
✅ Reported coverage/lint/type/complexity thresholds per the PR body: `check-all.sh` exit 0; `policy.py` 100% line/branch coverage, pylint 10.00/10, interrogate 100%, radon rank A.
✅ No shell/dependency files touched, so `uv.lock` / pre-commit hooks are unaffected.
🟡 I was unable to independently execute `./scripts/check-all.sh` or targeted pytest/mypy in this review environment (sandboxed shell blocked git checkout/fetch and most compound Bash invocations), so the above quality claims are taken from the PR's self-reported evidence and cross-checked only against the diff's internal consistency, not from a live run.

## Testing
Strong, and the standout part of this PR. `test_mcp_policy.py` (709 new lines) parametrizes the full remote admitted/refused ceiling table (including wrong case, whitespace, non-string scalars, unhashable containers), pins `CallerIdentity`'s fail-closed construction (no defaults, frozen), pins `effective_consumer`'s remote/local split, and adds AST-based transport-neutrality guards with non-vacuity companions. `test_mcp_remote.py`'s new characterization-pin section (634 lines) independently locks the pre-refactor server-level behaviour byte-for-byte, including the purge-tool ceiling carve-out and per-call (not per-server) identity resolution across both remote→local and local→remote transitions. No insufficient-testing concerns.

## Documentation
Docstrings in `policy.py` are unusually thorough about *why* (each design choice — no defaults, frozen, constant refusal reason, `object`-typed input — is justified against a specific failure mode). `docs/mcp.md` and the ADR were updated to describe the new ownership boundary without changing the documented external behaviour. Contract schema regeneration is minimal and justified (descriptions only, no version bump owed).



## Verdict: LGTM

The extraction preserves behaviour with unusually strong evidence (byte-identical characterization tests across the move, an identical 1215-row before/after decision table), the new policy module is fail-closed by construction with AST-level guards preventing MCP/SDK coupling from creeping back in, and docs/contracts/tests were all updated consistently. No bug, security regression, or test gap survives inspection of the full diff.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5148236897 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

Issue #1107, opened 2026-07-31
test(parity): the process↔classify class detector is structurally blind to LLM-path-only defects (it missed #974)


Geoffe-Ga:
## The gap

`tests/test_process_classify_parity.py` describes itself (module docstring,
L11-13) as:

> This module is the *class* detector for that defect — it fails automatically
> on a fourth instance, without anyone remembering to extend a list of axes.

**#974 was the fourth instance, and this detector did not catch it.**

## Why it missed

Precondition 1 of the invariant (docstring L31-36) scopes the whole comparison
to the deterministic path:

```python
pipeline = Pipeline(config=CreekConfig(), no_llm=True)      # L213
...
run_classify(vault_path=..., method="rules", force=True)    # L235-240
```

The precondition is *correct* on its own terms — an LLM-vs-rules disagreement is
genuinely not this defect class. But #974's defect **only manifests on the LLM
path**, and structurally so:

* `creek/pipeline.py` runs `rule_classifier.classify` (L573) **before**
  `apply_tier` (L579), so a rules-derived `confessional` + `conviction` is
  already visible to the privacy pre-pass. Verified: with `no_llm=True` and a
  body carrying the rules' confessional/conviction keywords, the pipeline
  already produces `intimate` at HEAD.
* The LLM dispatch (L586-590) runs **after** `apply_tier`, so only an
  LLM-hardened voice signal is invisible to the pre-pass. That was #974.

So the detector's blind spot is not an oversight in a fixture — it is the exact
complement of where the defect lives. Any future axis wired into one
orchestrator's *post-LLM* segment and not the other's will slip past it too.

## Proposal

Add a second parity scenario alongside the existing one, on a **stubbed** LLM
path — no network, no real provider:

* process side: `Pipeline(config=..., no_llm=False)` with
  `rule_classifier.classify` stubbed to a no-op and the tier classifiers'
  `.classify` stubbed to a deterministic verdict function.
* classify side: `run_classify(method="llm", force=True)` with
  `LLMClassifier.classify_with_reasoning` patched to the *same* deterministic
  verdict function (the pattern at `tests/test_classify_engine.py:1596-1613`),
  and `_assert_classifiers_available` patched out.

Because both sides consume an identical stubbed verdict, precondition 1's
concern (LLM-vs-rules disagreement) does not apply, and the same
whole-frontmatter-dict comparison minus `_CLASSIFY_ONLY_KEYS` can be reused
[...]

Issue #1106, opened 2026-07-31
feat(classify): no targeted remediation (and no visibility) for vaults already mis-stamped by the pre-#974 pipeline


Geoffe-Ga:
> **Rewritten 2026-08-14** after the joint audit of the three held tier/consent P0s (`wf_25ad9152-fae`, at `4767db3`). The original body's headline harm was **refuted** by merged PR #1423; the surviving harm is different and is stated below. Ordered **last** of the three P0s — see #1212 and #1031.

## The gap

#974 stops `creek process` from stamping confessional self-authored fragments `personal`. It does **nothing** for fragments already written that way, because `privacy_tier` is a one-way ratchet: nothing lowers it, and nothing revisits a tier that is already concrete.

An operator who ran `creek process` before #974 has an unknown number of fragments carrying `privacy_tier: personal` + `voice_proxy_eligible: true` that should be `intimate` + `false`. Those are admissible to a PERSONAL-ceiling caller and still feed voice-proxy generation.

## Why this is upstream of the other two P0s — measured

One vault holding both a keyless and a mis-stamped fragment:

```
_scan_fill_gaps      -> untiered=1        # sees keyless; BLIND to mis-stamped
voice default (ALL)  -> [keyless, mis]    # admits BOTH
voice PERSONAL       -> [mis]             # #1212's fix does NOT catch this
```

**A mis-stamped fragment survives both sibling fixes.** After #1212 it is still admitted at every ceiling >= personal, because its raw frontmatter honestly says `personal`. After #1031 it still egresses: `max_source_tier` = `personal`, and `resolve("classification", personal)` -> cloud.

This issue supplies the value the other two gates trust. It goes last only because it is a *feature* (detector + remediation surface) whose landing would move the disk values their fixtures depend on.

## What remediation exists today

| Command | Fixes it? | Cost |
|---|---|---|
| `creek classify --method llm` (no `--force`) | **No** | `needs_tier` is `False` for a concrete tier, so the reassess is skipped |
| `creek classify --method llm --force` | Yes | Re-pays the whole vault's LLM bill, **and** bypasses the preserved short-circuit (`classify_engine.py:923`), overriding `classification_method: manual` operator curation |
| `creek classify --method rules --force` | Yes | Free, but untargeted, whole-vault, and re-stamps `classification_method: llm -> rules`, **destroying provenance** |
[...]

Issue #1105, opened 2026-07-31
fix(classify): a non-force LLM run writes the confessional+conviction INTIMATE trigger and discards it — owns_tier gates the reassess out


Geoffe-Ga:
## The defect (fifth instance of the "wired into one entry point but not the other" class)

Found while fixing #974. **This is a different mechanism from #974 and is not fixed by it.**

`creek/classify/classify_engine.py::_prepare_fragment` computes

```python
owns_tier = needs_tier(raw) or force        # classify_engine.py:765
```

and `_process_file` gates the post-classification privacy reassess on it:

```python
if prepared.owns_tier:                       # classify_engine.py:919-920
    new_fragment = reassess(new_fragment, body, classifier=privacy)
```

`needs_tier` (`creek/classify/privacy_pass.py:90-92`) is `True` only when the
`privacy_tier` key is absent or literally `unclassified`. So for **any fragment
that already carries a concrete tier**, a non-`--force` run has `owns_tier ==
False` and the reassess never runs — even when that same run's LLM just produced
`voice_register: confessional` + `confidence: conviction`, which is
`PrivacyClassifier.classify_tier`'s third INTIMATE trigger
(`creek/classify/privacy.py:141`).

The run **writes the intimate trigger to disk and discards it in the same pass.**

## Reproduction (verified at b35811a)

Write one fragment stamped `privacy_tier: personal`, self-authored, platform
`markdown`, neutral body. Run `run_classify(method="llm", force=False)` with
`LLMClassifier.classify_with_reasoning` stubbed to return
`voice_register=confessional, confidence=conviction`. On-disk frontmatter after
the run:

```yaml
privacy_tier: personal
voice:
  confidence: conviction
  voice_register: confessional
voice_proxy_eligible: true
```

Expected: `privacy_tier: intimate`, `voice_proxy_eligible: false` — exactly what
`tests/test_classify_engine.py::test_classification_signals_escalate_the_tier`
already pins for the *untiered* shape.

## Who hits it

* **`creek fill --upgrade`** → `creek/cli.py:2342` calls
  `run_classify(..., method="llm", force=False)`. Its whole purpose is to
  upgrade fragments that already went through a pass — i.e. fragments that
  already carry a tier. Every one of them is in the blind spot.
* **Post-material-edit reclassification.** `VaultWriter.update_fragment` calls
  `_clear_classification` (drops the classification keys so the next pass
  revisits the fragment) and `_retier_after_rewrite`
[...]

PR #1113, opened 2026-08-01
fix(pipeline): run the post-classification privacy reassess in creek process

### LINKED ISSUE #974: fix(pipeline): creek process never runs the post-classification privacy reassess — confessional fragments stay PERSONAL and voice-proxy eligible

Geoffe-Ga:
## Summary

- **`creek process` never ran the post-classification privacy reassess.** `privacy_pass.reassess` exists to *raise* a fragment's tier once classification has hardened its voice signals — `PrivacyClassifier.classify_tier`'s third INTIMATE trigger is self-authored + `voice_register: confessional` + `confidence: conviction` (`creek/classify/privacy.py:141`), and those axes only exist after the classifier runs. `classify_engine._process_file` calls it (L919-920); `Pipeline._run_classification` did not. Fixed by mirroring the engine's ordering exactly: **classify → audience → praxis → reassess**, with reassess the last mutation before the write.
- **The pre-pass `apply_tier` does not move.** It must stay above the LLM dispatch so the per-tier router sees a real tier (#876, Intimate-never-cloud) — which is exactly *why* it cannot see the voice axes the LLM is about to fill in, and why a second look is needed.
- **This is about what the pipeline *writes*, not what a caller may *read*.** A fragment stamped with too low a tier at write time defeats every downstream read gate, because they all trust the stamp. Deliberately unrelated to the MCP tier-ceiling work (#1037/#1068/#1073/#1078/#1084/#1090).

### The divergence this closes

Same fragment, same body, same model verdict:

| | `privacy_tier` | `voice_proxy_eligible` |
|---|---|---|
| `creek classify --method llm` | `intimate` | `false` |
| `creek process` (before this PR) | `personal` | **`true`** |

The `true` is the damage: a confessional fragment stayed admissible to a PERSONAL-ceiling MCP caller and kept feeding voice-proxy generation (`creek/generate/voice_authenticity.py:336` filters on exactly this field). Which command the user happened to run decided whether intimate content was voice-proxy eligible.

## The diff

```python
raw: dict[str, object] = {PRIVACY_TIER_KEY: frag.privacy_tier}
owns_tier = needs_tier(raw)          # read BEFORE apply_tier answers it
frag = apply_tier(frag, item.body, raw=raw, force=False, ...)   # unmoved
...                                   # LLM dispatch, audience, apply_praxis
if owns_tier:
    frag = reassess(frag, item.body, classifier=self.privacy_classifier)
```

Four files: `creek/pipeline.py` (the fix), `creek/classify/privacy_pass.py` (docstring only), `tests/test_pipeline.py` (additive test class), `tests/test_process_classify_parity.py` (docstring note only).

## Safety properties, each verified rather than asserted

**It cannot lower a tier.** `reassess` reduces through `privacy_pass.escalate`, which returns the candidate only when `_ESCALATION_RANK[candidate] > _ESCALATION_RANK[current]`. `privacy_tier` is a one-way ratchet and lowering one is permanent burial, so this is the load-bearing property. It holds even when the model returns `voice_register: None` / `confidence: None` — pinned by `test_reassess_never_lowers_a_tier`, which drives both a lighter explicit verdict and an all-`None` verdict against already-`intimate` journal fragments.

**No fourth `privacy_tier` reader.** The change calls only the shared primitives (`needs_tier`, `apply_tier`, `reassess`); `reassess` reads the tier internally via `privacy_filter.tier_of`. The `raw={PRIVACY_TIER_KEY: frag.privacy_tier}` synthesis is pre-existing and only hoisted into a named local. #1079 is open because two existing readers already disagree; this adds none.

**`creek/` still never imports `creek_mcp`.** Zero hits.

**It does not bury ordinary content.** Between the pre-pass and the reassess, the only field `classify_tier` reads that anything can change is `voice`:

| pass | writes |
|---|---|
| `LLMClassifier._apply_classification` (`llm/orchestrator.py:279-297`) | `frequency`, `wavelength`, `voice`, `praxis_potential` — nothing else |
| `audience.classify_and_enforce` | `audience` |
| `apply_praxis` | `praxis_potential` |

`classify_tier` reads `source.platform`, `source.author`, `title`, body, `source.channel`, `voice.voice_register`, `voice.confidence`. Intersection = `voice`. So the LLM structurally **cannot** rewrite a title into a recovery keyword or flip authorship — the second `classify_tier` call is a provable no-op for every shape except self-authored + confessional + conviction. In particular: non-self-authored fragments can never reach an INTIMATE branch, and a `no_llm` / rules-only run changes nothing at all (`test_a_no_llm_run_changes_no_tier`).

**Ordering pins intact.** `TestPipelineTierRouting::test_untiered_journal_fragment_routes_local_and_gets_a_tier` (#876), the #937 audience placement, and #877's escalate-only praxis are all untouched and green.

## The `owns_tier` guard — an explicit decision

`reassess` is gated on `owns_tier = needs_tier(raw)`, mirroring `classify_engine._prepare_fragment` (`:765`, `:919-920`; `force` is always `False` here, so the engine's `needs_tier(raw) or force` reduces to `needs_tier(raw)`). It is **unreachable in production** — `grep -rn privacy_tier creek/ingest/` returns zero hits, so every ingested fragment arrives `unclassified` and `owns_tier` is always `True`.

It is included anyway because `apply_tier` *itself* declines to overwrite a tier already on record when `force=False` (`privacy_pass.py:153-155`). An unguarded reassess would make the pipeline honour a pre-set tier before the model runs and override it after — and deciding that question unilaterally, in the pipeline only, would create a **new** divergence between the two entry points, which is the exact defect class this issue is about. Whether the guard philosophy is right *repo-wide* is filed as #1105. `test_a_preset_ingester_tier_is_not_silently_raised` pins the seam contract and documents that it passes both before and after this fix, so nobody deletes it as "never red".

## Existing vaults are already wrong — say it plainly

**This PR does not remediate them.** A fragment `creek process` wrote before this fix carries `privacy_tier: personal`, and the ratchet will not walk it up retroactively. Verified remediation options today, all measured against a vault fixture:

| command | fixes it? | cost |
|---|---|---|
| `creek classify --method llm` (no `--force`) | **no** | `needs_tier` is `False` for a concrete tier → `owns_tier` `False` → reassess skipped |
| `creek classify --method llm --force` | yes | re-pays the whole vault's LLM bill, **and** `--force` bypasses the preserved short-circuit so it overwrites `classification_method: manual` curation |
| `creek classify --method rules --force` | yes (the pre-pass reads the `voice` already on disk and escalates) | free, but clobbers LLM-derived frequency/wavelength/voice with rules verdicts |
| `creek fill --gaps` | **cannot even report it** | it counts `needs_tier(raw)`; a wrongly-`personal` fragment is not untiered, so it is invisible |

So the operator's real recovery story is *"run `creek classify --method llm --force` and pay for it twice, losing your manual curation."* That is a threat, not a remediation path. Filed as **#1106** (a targeted re-tier pass plus a gap detector).

## Is `creek process` the only unwired entry point?

Enumerated every caller of the tier-stamping path:

| caller | runs `reassess`? |
|---|---|
| `creek process` → `Pipeline._run_classification` | **no** — this PR |
| `creek classify` / `creek sync` → `classify_engine._process_file` | yes, gated on `owns_tier` |
| MCP `creek.classify` (`creek_mcp/tools/classify.py:54`) | inherits the engine; `creek_mcp` does no independent tier stamping |
| re-ingest material edit → `vault/writer.py::_retier_after_rewrite` | n/a — no classifier runs there; escalate-only frontmatter mirror of `apply_tier` |
| `cli.py:2184` `_scan_fill_gaps` | read-only |
| `classify_engine._persist_tier_only` | n/a — nothing was classified |

As an *entry point*, yes — `creek process` was the only one. But a **fifth instance of the same defect class** is visible from here, with a different mechanism, and I filed it rather than fixing it: **#1105**. `owns_tier` is `False` for any fragment already carrying a concrete tier, so a non-`--force` LLM run writes the confessional+conviction trigger to disk and discards it in the same pass. Verified on a vault fixture at `b35811a`:

```yaml
privacy_tier: personal          # not raised
voice:
  confidence: conviction        # the trigger was written
  voice_register: confessional
voice_proxy_eligible: true
```

That hits `creek fill --upgrade` (`cli.py:2342`) and every post-material-edit reclassification.

## Test plan

**RED evidence, verified by me at `b35811a` before any fix:** `tests/test_pipeline.py::TestPipelinePrivacyReassess` → `2 failed, 3 passed`.
- `test_a_confessional_llm_verdict_raises_the_tier_before_the_write` → `AssertionError: assert 'personal' == 'intimate'` on the **on-disk** frontmatter after `_write_to_vault` (not the in-memory fragment — `voice_proxy_eligible` is a computed field and "what reaches the vault" is the contract).
- `test_a_non_confessional_verdict_leaves_the_tier_untouched` → the self-authored essay case read `open`, expected `intimate`.

The engine's counterpart (`tests/test_classify_engine.py::test_classification_signals_escalate_the_tier`) was already green throughout — so the two-commands-disagree evidence is exactly the new pipeline test failing while the engine test passes.

**Five new tests**, each designed so a blanket-escalate implementation cannot pass (distinct expected tiers in one batch, plus witness assertions that the stubbed verdict actually landed):

1. `test_a_confessional_llm_verdict_raises_the_tier_before_the_write` — on-disk `intimate` / `voice_proxy_eligible: false`.
2. `test_reassess_never_lowers_a_tier` — the ratchet, against a lighter verdict and an all-`None` verdict.
3. `test_a_non_confessional_verdict_leaves_the_tier_untouched` — three fragments, three distinct tiers (`personal` / `intimate` / `open`).
4. `test_a_preset_ingester_tier_is_not_silently_raised` — the `owns_tier` seam contract.
5. `test_a_no_llm_run_changes_no_tier` — the deterministic path is a strict no-op.

**Ran:**
- `pytest tests/test_pipeline.py tests/test_process_classify_parity.py tests/test_privacy_pass.py tests/test_classify_engine.py` → **147 passed**.
- `./scripts/check-all.sh` → **exit 0**, 7300 passed / 30 deselected, branch coverage **94.12%**.
- Gate 2.5 `code-review-orchestrator` over the diff → **CLEAN**, zero blocking and zero non-blocking findings.

No bypasses: zero `# noqa`, `# type: ignore`, `@pytest.mark.skip` or `--no-verify` anywhere in the diff; no threshold touched.

## Follow-ups filed (not fixed here, per #1011)

- **#1105** (P1) — fifth instance: `owns_tier` gates the reassess out on non-`--force` LLM runs, so a newly-hardened INTIMATE trigger is written and discarded in the same pass.
- **#1106** (P2) — no targeted remediation, and no visibility, for vaults already mis-stamped by the pre-#974 pipeline.
- **#1107** (P2) — `tests/test_process_classify_parity.py` calls itself the class detector for this defect family but is structurally blind to LLM-path-only instances, which is why it missed this one. Recorded as a known limitation in its docstring by this PR.

Closes #974
Refs #876, #937


Geoffe-Ga:
## Summary
`creek process` never called `privacy_pass.reassess`, the escalate-only second privacy-tier look that `creek classify`/`creek sync` run after classification hardens a fragment's voice signals. This meant a self-authored, confessional, high-conviction fragment could land `privacy_tier: personal` / `voice_proxy_eligible: true` via `creek process` while the same content would land `intimate` / `false` via `creek classify` — a real leak into voice-proxy generation. The PR hoists the `raw` dict, computes `owns_tier` before `apply_tier` answers it, and adds a guarded `reassess(...)` call as the last mutation before the write, mirroring the engine's ordering exactly.

## Strengths
- The new pipeline code (`creek/pipeline.py:605-650`) is a byte-for-byte structural match to `classify_engine.py`'s `_prepare_fragment`/`_process_file` ordering (classify → audience → praxis → reassess, gated on `owns_tier`) — I diffed both and confirmed the claim.
- The `owns_tier` guard is correctly justified: `apply_tier` already declines to overwrite a pre-set tier when `force=False`, so an unguarded reassess would asymmetrically honor-then-override a deliberate operator tier. `test_a_preset_ingester_tier_is_not_silently_raised` pins this seam even though it's unreachable with today's ingesters, with a comment explicitly warning against deleting it as "never red."
- Test design actively resists a cheating implementation: `test_a_non_confessional_verdict_leaves_the_tier_untouched` asserts three *distinct* expected tiers in one batch so a blanket-escalate-everything implementation cannot pass.
- I independently verified the "only `voice` changes between pre-pass and reassess" claim by reading `classify_tier` (`creek/classify/privacy.py:104-149`) against what audience/praxis passes write — it holds.
- The escalate-only ratchet itself already has a dedicated unit-level matrix test (`tests/test_privacy_pass.py::test_escalate_matrix`), so the pipeline-level tests only need to confirm correct wiring, which they do.
- PR description is exemplary: documents RED evidence, enumerates every tier-stamping caller, honestly states this doesn't remediate already-mis-stamped vaults, and files three well-scoped follow-up issues (#1105/#1106/#1107) rather than scope-creeping the fix.

## Security Concerns
None. 🟢 This is itself a privacy-hardening fix (closes a path where confessional content stayed voice-proxy eligible). No new `privacy_tier` reader is introduced; `creek/` still doesn't import `creek_mcp`.

## Problems
None found blocking merge.

## Code Quality
- Minor: `test_reassess_never_lowers_a_tier` uses `journal`-platform fragments, whose `classify_tier` rule (`platform == JOURNAL → INTIMATE`) already forces `INTIMATE` independent of the stubbed voice verdict. So this particular test doesn't isolate "a lighter verdict fails to demote" as cleanly as its docstring implies — the outcome is somewhat overdetermined by the platform rule rather than purely by `escalate`. Not a real gap given `escalate` has its own matrix test elsewhere, just a naming/framing nit worth a follow-up thought.
- The docstring additions in `_run_classification` are very long (roughly tripling the method's comment volume). Justified given the subtlety of the ordering constraint, but worth watching if this pattern keeps compounding across future privacy-pass changes.

## CLAUDE.md Adherence
✅ TDD: RED evidence documented and independently plausible given the diff. ✅ Conventional commit (`fix(pipeline): ...`). ✅ No bypasses (no `noqa`/`type: ignore`/`skip`/`--no-verify` in the diff). ✅ Operates from project root patterns preserved. ✅ No threshold changes. CI was still `IN_PROGRESS` at review time (not failed) — standard gate, not a diff issue.

## Testing
Five new tests in `TestPipelinePrivacyReassess` cover: the core escalation, the never-lowers ratchet, non-confessional no-ops across three distinct tiers, the `owns_tier` seam contract, and the no-LLM strict no-op path. This is thorough and well above the bar for a targeted fix.

## Documentation
`privacy_pass.py`'s `reassess` docstring updated to note both entry points now call it; `test_process_classify_parity.py`'s module docstring gained an honest "known limitation" note explaining why this class of bug evaded that detector. Good documentation hygiene.



## Verdict: LGTM

The fix is a minimal, precisely-targeted wiring change verified line-for-line against classify_engine.py's existing ordering (classify → audience → praxis → reassess, gated on owns_tier); the escalate-only ratchet is independently unit-tested in test_privacy_pass.py, and the five new pipeline tests use distinct expected tiers per fragment specifically to rule out a blanket-escalate false pass. No bugs, no CLAUDE.md violations, and no missing test coverage were found — the only remark (test_reassess_never_lowers_a_tier's outcome being overdetermined by the journal platform's own INTIMATE rule) is a minor observation, not something that would block merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5148550726 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Fixes #974: `creek process` (the `Pipeline._run_classification` path) never called `privacy_pass.reassess`, so a fragment whose LLM verdict came back self-authored + `confessional` + `conviction` (the third INTIMATE trigger in `PrivacyClassifier.classify_tier`) stayed stamped `personal` / `voice_proxy_eligible: true` — while the same content through `creek classify` correctly landed `intimate` / `false`. The fix hoists `owns_tier = needs_tier(raw)` before the pre-pass `apply_tier` call and adds a guarded `reassess(...)` as the last mutation before write, placed to exactly mirror `classify_engine._prepare_fragment` / `_process_file`'s ordering and gating.

## Strengths
- **Ordering is provably correct, not just asserted.** I read `classify_tier` (`creek/classify/privacy.py:104-149`) directly: the self-authored INTIMATE checks (recovery keyword → journal → confessional+conviction) run *before* the essay/discord/fallback branches, and the only field that changes between the pre-pass and the reassess call is `fragment.voice` (voice_register/confidence) — title, body, source.platform, source.author, source.channel are all fixed at ingest time. That matches the PR's "intersection = voice" claim exactly.
- **Escalate-only ratchet is real.** `escalate()` (`privacy_pass.py:96-114`) strictly requires `_ESCALATION_RANK[candidate] > _ESCALATION_RANK[current]`, so a weaker post-classification verdict structurally cannot lower a tier — verified by reading the rank table, not just trusting the docstring.
- **`owns_tier` guard mirrors the engine 1:1.** `classify_engine._prepare_fragment` computes `owns_tier = needs_tier(raw) or force`; the pipeline correctly notes `force` is always `False` on this path so the two reduce to the same expression, and both call sites gate `reassess` on the same boolean. I confirmed this against `classify_engine.py:762-768` and `:915-920`.
- **Test suite is adversarially designed.** `TestPipelinePrivacyReassess` uses three fragments with three *distinct* expected tiers in one batch (`test_a_non_confessional_verdict_leaves_the_tier_untouched`) specifically so a "just return intimate for everything" stub implementation cannot pass. `test_a_preset_ingester_tier_is_not_silently_raised` pins a seam contract that is un-reachable in production today but protects the `owns_tier` guard from being silently deleted as "never red." `test_reassess_never_lowers_a_tier` exercises both an explicit lighter verdict and an all-`None` verdict against `intimate` journal fragments.
- **Honest scope-limiting.** The PR explicitly flags a same-class defect it did *not* fix (#1105: `owns_tier` also gates out non-`--force` LLM reclassification runs) rather than scope-creeping the fix, and separately documents that existing vaults are not remediated by this change (#1106) plus a known blind spot in the parity-test detector (#1107, patched into that module's docstring here).

## Security Concerns
None. 🟢 This is a fix that *tightens* a privacy-tier leak (a confessional/self-authored fragment previously stayed voice-proxy-eligible when it should not have). No new I/O, no new external input handling, no change to the redaction or MCP tier-ceiling boundary.

## Problems
None found. No bugs, no incorrect ordering, no complexity blowup (the added logic is one bool read + one guarded call).

## Code Quality
- The inline docstring/comment blocks in `pipeline.py` are quite dense (several multi-sentence comment blocks citing issue numbers and other files by line number). This is consistent with the file's existing style (the `#937`/`#877` comments right above it use the same density), so it's not a deviation — just flagging as a matter of taste, not a blocker.
- Nothing else to flag; the diff is minimal and every line change is load-bearing.

## CLAUDE.md Adherence
✅ TDD: PR body documents RED evidence (`2 failed, 3 passed`) captured before the fix.
✅ Coverage: PR reports `./scripts/check-all.sh` exit 0, 94.12% branch coverage (≥90% threshold) — I was not able to independently re-execute `pytest`/`check-all.sh` in this review sandbox (test-running Bash commands were blocked from auto-approval), so this is trusted from the PR's reported run rather than independently re-verified here.
✅ No bypasses: no `# noqa`, `# type: ignore`, `@pytest.mark.skip`, or `--no-verify` introduced (confirmed by reading the full diff — none present).
✅ Typing: `dict[str, object]`/`bool` annotations are consistent with the file's existing strict-mypy style; no `Any` introduced.
✅ Docstrings: every new function/class in the test file and the touched `pipeline.py` docstring block are fully documented (Args/Returns), consistent with the ≥95% interrogate threshold.
✅ Commit/PR conventions: `fix(pipeline): ...` follows Conventional Commits; PR targets a feature branch, not a direct push to main.

## Testing
Five new tests in `TestPipelinePrivacyReassess`, each targeting a distinct facet: the RED-evidence case (on-disk frontmatter, not just in-memory fragment — important, since `voice_proxy_eligible` is what actually gates downstream generation), the never-lowers-a-tier ratchet (including an all-`None` verdict, which is the shape a partial/malformed model answer produces), the escalate-on-signal-not-on-principle case (three distinct tiers in one batch), the `owns_tier` seam contract, and the no-LLM no-op case (protecting `test_process_classify_parity`'s deterministic-path invariant). I traced each test's expected tier back through `classify_tier`'s actual branch logic and they all check out (e.g., `frag-reassess005`'s essay+confessional+conviction correctly resolves to `INTIMATE` because the self-authored checks run before the essay branch). No insufficient-testing concerns.

## Documentation
The `privacy_pass.py` module docstring and `test_process_classify_parity.py` module docstring were both updated to reflect the new caller and the newly-documented detection blind spot (#1107). Accurate and matches the implementation.



## Verdict: LGTM

The fix is small, surgical, and precisely mirrors the already-trusted classify_engine ordering (owns_tier guard, escalate-only reassess placed last); I independently verified every load-bearing claim (classify_tier's branch order, the voice-only intersection between passes, the UNCLASSIFIED-tier default, and voice_proxy_eligible's computed-field derivation) against the actual source, and they all check out. Five new tests pin the ratchet, the no-regression case, and the owns_tier seam contract with distinct expected tiers per fragment so a blanket-escalate stub cannot fake a pass — there is nothing here I'd hold up merge for.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5148656344 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1116, opened 2026-08-01
Fragment timestamp invariant is bypassable via model_construct / model_copy / direct assignment (validate_assignment is off)


Geoffe-Ga:
## Context

Surfaced by the Gate 2.5 review of #976, which added
`Fragment._normalise_timestamp` — a `@field_validator("created", "ingested", "authored_at",
mode="after")` that anchors a naive vault timestamp to America/Los_Angeles.

## The gap

Pydantic field validators do not run on every mutation path, and `Fragment` does not enable
`validate_assignment`. Verified by execution on the #976 branch:

```
Fragment.model_construct(..., ingested=naive).ingested.tzinfo   -> None
base.model_copy(update={"ingested": naive}).ingested.tzinfo     -> None
c = base.model_copy(deep=True); c.ingested = naive              -> None
Fragment.model_config.get("validate_assignment")                -> None   (off)
Fragment(..., ingested=naive).ingested.tzinfo                   -> America/Los_Angeles
```

So the "never naive" invariant holds on the constructor and `model_validate` paths — which is
every ingest and every vault read — but not on `model_construct` / `model_copy(update=...)` /
direct attribute assignment.

**No production caller exploits this today.** Every `model_copy` / `model_construct` site on
`Fragment` across `creek/` was checked during the #976 review (~25 sites); none writes
`created`, `ingested` or `authored_at`. #976 therefore fixed the docstrings to state the real
scope rather than expanding its blast radius, and left the structural hole open on purpose.

The hole is not idle, though: `creek/clean/context.py:138-140` and `:227-232` already use
`updated = frag.model_copy(deep=True)` followed by direct attribute assignment
(`updated.context = ...`, `updated.privacy_tier = ...`). The pattern is live and reused — it
simply has not been pointed at a timestamp field yet. The first time someone does, the
invariant silently stops holding, which is exactly the failure shape #976 existed to fix.

## Why this is its own issue

Setting `validate_assignment=True` on `Fragment` is a behaviour change with real reach:

- It re-runs **every** field validator on **every** attribute write, including the fail-closed
  `_coerce_voice_weight` / `_coerce_representativeness` / `_coerce_audience` coercers. Code that
  today assigns a sentinel and fixes it up on the next line would start getting coerced mid-flight.
- It costs a validation pass per assignment on hot paths (`clean/context.py` runs per fragment
  at 35k-fragment vault scale) — needs measuring, not assuming.
[...]

Issue #1115, opened 2026-08-01
Two ensure_aware helpers disagree: creek.time anchors naive datetimes to LA, clean/validator._ensure_aware assumes UTC


Geoffe-Ga:
## Context

Found while fixing #976 (which adds a `Fragment` field validator normalising `created` /
`ingested` / `authored_at` through `creek.time.ensure_aware`).

## The duplication — and the divergence

`creek/clean/validator.py:319` defines a private helper:

```python
def _ensure_aware(dt: datetime) -> datetime:
    """Ensure a datetime is timezone-aware, assuming UTC if naive."""
    if dt.tzinfo is None:
        return dt.replace(tzinfo=UTC)
    return dt
```

`creek/time.py:40` defines `ensure_aware` with the same name and shape but a **different
assumption**: naive values are anchored to `America/Los_Angeles` (per ontology §8.3 — every
timestamp this pipeline writes is LA, so a naive value is an LA wall clock that lost its
offset in transit), not UTC.

Two helpers, one name, an 7-8 hour disagreement about what a naive datetime *means*.

## Why this is not a drive-by fix

Unifying them is a **behaviour change**, not a refactor. `_ensure_aware` feeds the
`future_date` and `pre_2000` verdicts in `_check_timestamps`
(`creek/clean/validator.py:229-245`). Switching UTC -> LA shifts a naive timestamp by the LA
offset, which can flip either verdict for a value sitting within ~8 hours of `now` or of the
`_MIN_YEAR` boundary. That needs its own tests, which is why #976 deliberately left it alone.

There is also a second-order effect from #976 worth checking first: once `Fragment` normalises
its timestamps at validation time, `_ensure_aware` can no longer receive a naive value *via a
`Fragment`*, so its naive branch may only still be reachable through other callers (or not at
all). Confirm reachability before deciding between "unify on `creek.time.ensure_aware`" and
"delete the helper as dead".

## Suggested approach

1. Enumerate `_ensure_aware`'s callers and determine whether the naive branch is still
   reachable post-#976.
2. If reachable: write the boundary tests (a naive timestamp within the LA offset of `now`,
   and one within the offset of 2000-01-01T00:00), then unify on `creek.time.ensure_aware`
   and pin the new verdicts.
3. If unreachable: remove the helper and call `creek.time.ensure_aware` directly, keeping the
   validator honest about the single source of truth.

## Acceptance criteria

[...]

PR #1118, opened 2026-08-01
fix(models): enforce the "never naive" timestamp contract on Fragment

### LINKED ISSUE #976: Fragment accepts naive ingested/authored_at from vault frontmatter, breaking the "never naive" contract in creek/time.py

Geoffe-Ga:
## Summary

- **Makes the "never naive" contract real.** `creek/time.py::effective_authored_at` documented that
  `Fragment.ingested` is "Always present, never naive" — but nothing enforced it. `Fragment` had
  `field_validator`s for `voice_weight` / `representativeness` / `audience` and none for the three
  datetime fields, so a vault frontmatter timestamp serialised without an offset flowed straight in and
  the contract was a comment, not an invariant. This adds one
  `@field_validator("created", "ingested", "authored_at", mode="after")` on `Fragment` delegating to the
  existing `creek.time.ensure_aware` (added by #938), and rewrites the docstrings so every "never naive"
  claim **names its enforcer** instead of asserting an unowned property.
- **Root cause, not another call site.** #938 hardened `compost.py`'s call site only; the issue listed
  ~15 more unguarded consumers. Fixing the one constructor makes every present and future consumer
  inherit the invariant, instead of playing whack-a-mole down the pipeline. No consumer file is touched.
- **Coercion is at the model boundary and is lossless** — naive in, LA *attached* (wall clock preserved,
  never converted); aware in, returned untouched with `tzinfo` and microseconds intact; `None`
  `authored_at` stays `None`.

## What actually happens today (measured, not predicted)

A naive `ingested` does **not** crash `effective_authored_at` — that returns the naive value verbatim,
and `effective_authored_date` works fine (a `date` carries no tzinfo). The `TypeError` fires at the
*consumer*, the moment a naive-derived value meets an aware one. Reproduced through a public API:

```
SynchronicityDetector().detect_synchronicities(...)
  -> TypeError: can't compare offset-naive and offset-aware datetimes
     at creek/generate/synchronicity.py:306   (_chronological_pair's comparison)
```

It aborts at the **comparison** on line 306, not the subtraction on 257 — so hardening only the
arithmetic would have left the crash in place. An all-naive vault works today, which is exactly why
the suite was green: `tests/test_synchronicity.py`'s fixtures fed naive datetimes on *both* sides and
never mixed the two.

## Coerce, not reject — and why

Both designs were measured on the real suite before choosing:

| design | result |
| --- | --- |
| coerce (`ensure_aware`) | **7427 passed, 0 failed** |
| reject (`raise` on naive) | **204 failed + 73 errors across 17 test files** |

The deciding argument is not the count, though — it is that
`creek/vault/reader.py::try_load_fragment` catches `ValidationError` and returns `None`, logging at
DEBUG. **A rejecting validator would be fail-*silent*, not fail-closed:** an operator's existing
offsetless fragments would quietly vanish from every vault scan with no error surfaced anywhere. That
is strictly worse than today's loud `TypeError`. `tests/test_vault_reader.py`'s new test asserts
`record is not None` specifically to pin coerce-over-reject against a future "hardening".

LA is the right anchor, not UTC: ontology §8.3 normalises every timestamp this pipeline writes to LA,
so a naive value is an LA wall clock that lost its offset in YAML transit. The value is *attached*, so
no instant moves and no fragment crosses a day boundary — verified for `2024-01-01 00:30`.

**An operator with an existing naive-datetime vault sees no breakage:** the vault that loads today
still loads, every fragment's effective date is unchanged, and nothing is logged (deliberately — this
repair is lossless and universal for legacy vaults; warning per field would emit six figures of noise
on a 35k-fragment scan, unlike the sibling `_coerce_*` validators, which warn because they *discard* a
malformed value).

## Why all three fields

`effective_authored_at` excludes `created` from its *precedence* chain for a semantic reason (it
conflates authored-date with fs-mtime), which is orthogonal to tz-awareness — and
`creek/atomize/aggregate.py:173,183,215,268,300,301` sorts, `min`/`max`es and subtracts on
`Fragment.created`, the identical crash surface. `authored_at` is the only optional one, so the
validator returns `None` unchanged; a dedicated test pins that it never invents a date.

## Scope

Deliberately **not** changed: every consumer (`link/*`, `generate/*`, `atomize/*`);
`compost.py:437`'s now-redundant `ensure_aware` (harmless, aware-in/aware-out); `Thread` / `Eddy` /
`Decision` date fields (typed `date`, structurally immune); `AuthorManifest` (no datetime field).

## Behaviour change worth knowing about

Because `Fragment` timestamps now arrive LA-aware, `creek/clean/validator.py`'s **UTC**-assuming
private `_ensure_aware` becomes a no-op for fragments, so `creek clean`'s `future_date` verdict shifts
by the LA offset — a naive wall clock in a ~7-8 hour band that was not flagged before is flagged now.
Measured, and the new verdict is the more correct one (an LA wall clock 3 hours ahead of LA-now really
is in the future; the old UTC reading silently accepted it). The verdict is `review`, not deletion.
No test covered that band either way — recorded on #1115 with the boundary test to write.

## Follow-ups filed

- **#1115** — two `ensure_aware` helpers disagree (`creek.time` anchors to LA, `clean/validator`
  assumes UTC); includes the `future_date` measurement above.
- **#1116** — the invariant is bypassable via `model_construct` / `model_copy(update=…)` / direct
  attribute assignment (`validate_assignment` is off). No production caller exploits it today (all ~25
  `model_copy`/`model_construct` sites on `Fragment` were checked), so this PR documents the real scope
  rather than expanding its blast radius.

## Test plan

- **Gate 1 RED verified twice** (once against the final test files, by reverting only `creek/` and
  re-running): 9 tests fail at `89a5b79`, each for its intended reason — including the `TypeError` from
  `synchronicity.py:306`.
- **GREEN:** all 9 pass. New coverage: naive→LA for each of the three fields; wall-clock preservation
  (kills an `astimezone` implementation); the offsetless-ISO-**string** path (the only test that
  distinguishes `mode="after"` from `mode="before"`); the raw-naive-`datetime` mapping path; Sydney
  `authored_at` preserved with `tzinfo` and microseconds; UTC `ingested` untouched; `None` preserved;
  the on-disk `try_load_fragment` frontmatter round trip; and the mixed naive/aware
  `detect_synchronicities` regression asserting the *result* (`len == 1`, earlier fragment first,
  `time_gap_days == 151` — re-derived by hand, not reverse-engineered from a run).
- `tests/test_time.py::test_result_is_always_tz_aware` was **vacuous** (its docstring claimed Fragment
  validated tz-awareness "via Pydantic's datetime parsing", which was false, and it only ever fed
  tz-aware inputs). It now feeds naive inputs on both branches and asserts the LA offset; both original
  assertions are retained. No live assertion was removed anywhere in the diff.
- `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 (ruff, ruff-format, mypy strict,
  pylint 9.69, bandit + pip-audit, xenon, refurb, tryceratops, interrogate 99.4%, unit tests,
  coverage 94.13%, per-file gate).
- Full suite incl. integration/e2e: **7463 passed, 5 skipped**, stable across repeated
  randomised-order runs.
- Gate 2.5 `code-review-orchestrator` run to `CLEAN`; its one blocker (docstrings promising more than a
  field validator can deliver — the same defect class as the issue itself) was fixed by scoping the
  claims to the constructor / `model_validate` paths.

Closes #976
Refs #938

### DIFF
--- a/creek-tools/creek/models.py
+++ b/creek-tools/creek/models.py
@@ -17,7 +17,7 @@ from typing import TYPE_CHECKING, Literal, TypeVar, get_args
 from pydantic import BaseModel, ConfigDict, Field, computed_field, field_validator
 
 from creek.compile.provenance import CompileMethod, ProvenanceEntry
-from creek.time import now_la, today_la
+from creek.time import ensure_aware, now_la, today_la
 
 logger = logging.getLogger(__name__)
 
@@ -729,6 +729,23 @@ class Fragment(BaseModel):
         source: Source metadata. Its ``author_slug`` names the
             ``11-Other-Authors/<slug>/`` folder a borrowed fragment came from
             (``None`` for a native fragment), distinct from ``source.author``.
+        created: When the fragment came into being on the vault side.
+            Normalised like every timestamp below.
+        ingested: The wall-clock moment the vault wrote the fragment.
+            Never naive when the fragment was built through the normal
+            constructor or ``model_validate`` path.
+        authored_at: The timestamp the source itself records, or ``None``
+            when none is extractable — the model never invents one.
+            ``created`` / ``ingested`` / ``authored_at`` are all
+            normalised at validation time by
+            :meth:`_normalise_timestamp` (#976), which *anchors* rather
+            than converts: a naive value keeps its wall clock and gains
+            the America/Los_Angeles offset, while an already-aware value
+            — a Sydney ``authored_at``, say — passes through untouched
+            with its ``tzinfo`` and microseconds intact. A ``None``
+            ``authored_at`` stays ``None``. See
+            :meth:`_normalise_timestamp` for exactly which construction
+            paths this normalisation covers, and which bypass it.
     """
 
     model_config = ConfigDict(use_enum_values=True)
@@ -751,6 +768,8 @@ class Fragment(BaseModel):
     # until the audience classifier runs — additive and backward-compatible,
     # and fails closed to ``mixed`` so a corrupt value never crashes a scan.
     audience: Audience = "mixed"
+    # All three timestamps below are anchored to LA when they arrive
+    # naive — see ``_normalise_timestamp`` for the contract (#976).
     created: datetime = Field(default_factory=now_la)
     ingested: datetime = Field(default_factory=now_la)
     # Timestamp the source itself records (a Substack post's publish
@@ -816,6 +835,58 @@ class Fragment(BaseModel):
             return value
         _warn_fail_closed("Fragment audience", value, "mixed")
         return "mixed"
+
+    @field_validator("created", "ingested", "authored_at", mode="after")
+    @classmethod
+    def _normalise_timestamp(cls, value: datetime | None) -> datetime | None:
+        """Anchor a naive timestamp to LA; leave an aware one untouched (#976).
+
+        Vault frontmatter routinely serialises timestamps without an
+        offset (``ingested: 2024-01-01 12:00:00``), which PyYAML parses
+        into a *naive* datetime. Stored verbatim, it raises ``TypeError:
+        can't compare offset-naive and offset-aware datetimes`` at
+        whichever consumer first compares it against a neighbouring
+        aware timestamp. Repairing it here — the chokepoint every
+        ``Fragment`` load crosses via the constructor or
+        ``model_validate`` — is what lets the downstream surfaces
+        (synchronicity, temporal linking, wavelength bucketing) treat
+        "never naive" as an invariant on those paths, instead of
+        re-repairing it site by site. The invariant does not reach
+        ``Fragment.model_construct``, ``model_copy(update=...)``, or a
+        direct attribute assignment on an existing ``Fragment``: none
+        of those routes through Pydantic field validators, so a caller
+        that builds or mutates a fragment that way can still hand a
+        naive datetime downstream. No production caller assigns a
+        timestamp that way today, but a future one should not assume
+        this validator ran.
+
+        ``mode="after"`` is load-bearing rather than stylistic: a
+        before-validator sees the raw input, which on the
+        ``model_validate`` path can still be an offsetless ISO-8601
+        ``str`` Pydantic has yet to parse into a ``datetime``. This
+        function calls ``ensure_aware(value)``, which reads
+        ``value.utcoffset()`` — handed that raw string, a
+        before-validator would raise ``AttributeError``, not silently
+        return a naive result. ``mode="after"`` guarantees Pydantic has
+        already parsed the value into a real ``datetime`` before this
+        validator runs.
+
+        Deliberately silent, unlike the sibling ``_coerce_*``
+        validators: those warn because they *discard* a malformed value
+        the operator needs to know about, whereas this repair is
+        lossless and expected of every legacy vault. Warning per field
+        would emit six figures of noise on a 35k-fragment scan.
+
+        Args:
+            value: The supplied timestamp; ``None`` only for the
+                optional ``authored_at``.
+
+        Returns:
+            ``None`` unchanged, otherwise the value with an offset
+            guaranteed — attached, never converted, per
+            :func:`creek.time.ensure_aware`.
+        """
+        return None if value is None else ensure_aware(value)
 
     # BUG-009: the ``[prop-decorator]`` suppression below is a known
     # mypy / Pydantic-v2 limitation when stacking ``@computed_field``
--- a/creek-tools/creek/time.py
+++ b/creek-tools/creek/time.py
@@ -9,9 +9,12 @@ produce naive datetimes that fail to compare against tz-aware ones with
 values that arrive from outside that discipline — persisted frontmatter,
 legacy callers — making them safe to compare without moving the clock.
 
-The constant :data:`LA_TZ` is re-exported from
-:mod:`creek.ingest.base` to preserve the historical import path while
-giving callers a dependency-free home for the helper.
+The constant :data:`LA_TZ` is defined here as a dependency-free home
+for the timezone helpers; :mod:`creek.ingest.base` independently
+defines its own equal constant of the same name to preserve the
+historical import path some callers still use. Neither re-exports the
+other — each module owns its own definition, and the two simply agree
+on the same ``ZoneInfo("America/Los_Angeles")`` value.
 """
 
 from __future__ import annotations
@@ -45,6 +48,15 @@ def ensure_aware(value: datetime) -> datetime:
     failure mode this module exists to prevent. Callers that mix a
     pipeline-generated clock with timestamps read back off disk route
     through this helper first.
+
+    The primary production caller is
+    :class:`~creek.models.Fragment`'s field validator over ``created`` /
+    ``ingested`` / ``authored_at`` (#976), which makes every fragment
+    load through the constructor or ``model_validate`` — the paths
+    ingest and vault reads use — the enforcement point for the "never
+    naive" invariant. See
+    :meth:`~creek.models.Fragment._normalise_timestamp` for the
+    construction paths that bypass it.
 
     The contract is deliberately asymmetric:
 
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
Adds a single `@field_validator("created", "ingested", "authored_at", mode="after")` on `Fragment` that anchors naive timestamps to `America/Los_Angeles` via the existing `creek.time.ensure_aware`, making the previously-documented-but-unenforced "never naive" contract real at the one chokepoint every `Fragment` load crosses (constructor + `model_validate`). No consumer code is touched; only `creek/models.py`, `creek/time.py`, and three test files change.

## Strengths
- Root-causes the fix at the model boundary instead of patching another call site (as #938 did for `compost.py`), so every present and future consumer inherits the invariant for free.
- The coerce-vs-reject design choice is justified with a concrete, non-obvious argument: `try_load_fragment` swallows `ValidationError` and returns `None`, so a rejecting validator would silently vanish legacy fragments from vault scans — strictly worse than today's loud `TypeError`. `test_fragment_with_offsetless_frontmatter_timestamp_loads_tz_aware` pins exactly this via `record is not None`.
- `mode="after"` is justified rather than asserted: a `mode="before"` validator would see a raw offsetless ISO string on the `model_validate` path and blow up on `.utcoffset()`. `test_iso_string_without_offset_is_anchored_to_la` is the one test that actually distinguishes the two modes — good discipline.
- Attach-not-convert semantics are pinned by `test_naive_coercion_preserves_the_wall_clock`, which specifically calls out and would fail against an `astimezone()`-based mis-implementation.
- `tests/test_time.py::test_result_is_always_tz_aware` was previously vacuous (fed only aware inputs despite its docstring claiming otherwise); this PR fixes it to actually feed naive inputs and assert the LA offset — a real bug in test quality caught and corrected as a side effect.
- The PR body's claims are independently verifiable: the changed-files list matches exactly what `gh pr diff --name-only` shows, and both follow-up issues (#1115, #1116) referenced in the body exist and are open.
- Known, deliberately out-of-scope behavior change (the `future_date` verdict in `creek/clean/validator.py` shifting because its own UTC-assuming `_ensure_aware` becomes a no-op for already-aware `Fragment` fields) is disclosed rather than silently shipped, with a follow-up issue and reasoning for why the new behavior is more correct.

## Security Concerns
None. 🟢 No injection, path-traversal, or secret-handling surface touched. The validator only reads `value.utcoffset()` and attaches a fixed `ZoneInfo` — no untrusted input reaches subprocess, filesystem, or network calls.

## Problems
None blocking. The one known gap — the invariant is bypassable via `model_construct`/`model_copy(update=...)`/direct attribute assignment since `validate_assignment` is off — is explicitly scoped out, checked against all ~25 existing call sites (none exploit it), and tracked in #1116 rather than silently left undocumented.

## Code Quality
- Non-blocking: the docstrings on `Fragment`, `_normalise_timestamp`, and `ensure_aware`/`effective_authored_at` are quite long relative to the size of the actual code change. They're accurate and well-cross-referenced (each names its enforcer, per the PR's own stated principle), but a reviewer skimming just the code would get most of the value from a shorter version. Not asking for a rewrite — this is a style preference, not a defect.
- The comment above `created`/`ingested` field declarations duplicates information already fully covered in the class docstring and the validator's own docstring; a one-line pointer would suffice. Purely cosmetic.

## CLAUDE.md Adherence
✅ TDD discipline evident from the PR description (RED verified twice, GREEN confirmed, no live assertion removed). ✅ Root-cause fix matching the "Workflow: Stay Green" philosophy. ✅ Coverage/lint/type/complexity gates reported passing (`check-all.sh` exit 0, 12/12, coverage 94.13%, interrogate 99.4%). ✅ No consumer files touched, keeping blast radius minimal as stated. Could not independently re-run `check-all.sh` in this review environment (tool execution was sandboxed), so gate-passing is taken on the PR's self-report plus static inspection of the diff, which is internally consistent.

## Testing
Strong. New/updated tests cover: naive→LA anchoring for each of the three fields individually, wall-clock preservation (kills an `astimezone` mis-implementation), the offsetless-ISO-string path (the only test distinguishing `mode="after"` from `mode="before"`), the raw-naive-`datetime`-via-mapping path, an aware Sydney `authored_at` passthrough, an aware UTC `ingested` passthrough, `None` `authored_at` preservation, an on-disk `try_load_fragment` round trip, and a mixed naive/aware `detect_synchronicities` regression that asserts the actual result values (count, ordering, `time_gap_days == 151`) rather than merely "did not raise." This is exactly the kind of test suite the "insufficient testing is BLOCKING" bar wants — nothing here is insufficient.

## Documentation
Docstrings were rewritten across `models.py` and `time.py` so every "never naive" claim names its enforcer instead of asserting an unowned property, directly fixing the defect class the underlying issue was about. Docstrings scope their claims precisely to the constructor/`model_validate` paths and explicitly flag the paths that bypass the invariant, avoiding the "promises more than the code delivers" trap the PR body says an earlier review pass already caught once.



## Verdict: LGTM

The fix is minimal, root-caused at the model boundary, backed by 9 new/updated tests that pin the exact behaviors that matter (wall-clock preservation, aware-passthrough, None-passthrough, the string vs. datetime mode="after" distinction, and an on-disk round trip), and every claim in the PR body (scope of changed files, the two follow-up issues, the clean/validator.py UTC-assuming sibling) checks out against the actual repository state. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5148942140 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1120, opened 2026-08-01
A failed mid-line .id-index.jsonl append corrupts the NEXT append too (both entries lost)


Geoffe-Ga:
## Problem

`VaultWriter._append_index_entry` appends `{id, filename}` JSON lines to
`.id-index.jsonl` with `O_WRONLY | O_CREAT | O_APPEND`. Every append lands at
EOF with no separator of its own — the trailing `\n` is part of the payload.

If an append fails part-way (`ENOSPC`, or a descriptor that stops making
forward progress), the bytes already written stay on disk **without their
trailing newline**. The next successful append is then concatenated onto that
partial line:

```
{"filename": "a.md", "id": "frag-a"}          <- fine
{"filename": "b.md", "id": "fr                <- partial, no newline
{"filename": "c.md", "id": "frag-c"}          <- appended straight onto it
```

`_load_index_file` (`creek-tools/creek/vault/writer.py:1163-1191`) iterates
`raw.splitlines()` and skips anything that fails `json.loads`. The merged
line is not valid JSON, so **both** the failed entry and the next legitimate
one are silently dropped from the id index — not just the partial one.

A dropped id mapping is not cosmetic: it is what duplicate detection and
`VaultWriter.update_fragment` key on, so the fragment becomes invisible to
both.

## Why this is a separate issue

#987 made all three raw `os.write` sites drain their buffer, which removes the
*common* cause (a short write that used to be ignored now loops to completion).
What remains is the hard-failure case: after #987 the caller at least gets a
loud `OSError` instead of silent success, but the on-disk residual is unchanged.

Fixing it properly needs something #987 deliberately left alone:

- `ftruncate`-back is **unsafe** here. With `O_APPEND` the write offset is
  chosen atomically inside the syscall, so we never reliably learn our own
  start offset, and a concurrent appender may already have landed a complete
  line past ours that truncating would destroy.
- The workable shapes are a rewrite-under-lock (flock the index, rewrite it
  whole, `os.replace`), a leading-newline/record-delimiter framing change, or
  routing the index through `creek.audit.AuditLog`, which already does
  flock + fsync + hash-chain.

All three change the index file's concurrency contract, which is out of scope
for a short-write fix.

## Current state

The residual is documented in `_append_index_entry`'s docstring and pinned by
a characterization test in `creek-tools/tests/test_vault_writer.py`
(`TestAtomicWriteHardening`) so the behaviour cannot change unnoticed. When
[...]

Issue #1119, opened 2026-08-01
fix(scripts): a stale .ruff_cache lets check-all.sh pass while CI fails on the same tree


Geoffe-Ga:
## Problem

Gate 2 (`./scripts/check-all.sh`) can exit 0 on a tree that CI then fails at the
`Run Ruff linting` step, because local ruff consults `.ruff_cache` and CI never has one.

Observed on PR #1117 (issue #1074). All three `Code Quality & Testing` jobs failed in
about a minute with:

```
tests/test_v1_api_cli.py:27:1: I001 Import block is un-sorted or un-formatted
tests/v1_api_support.py:23:1: I001 Import block is un-sorted or un-formatted
```

Locally on the identical commit:

```
$ ruff check tests/v1_api_support.py            # cached
All checks passed!                               # exit 0

$ ruff check --no-cache tests/v1_api_support.py  # cold
I001 Import block is un-sorted or un-formatted   # exit 1
```

`touch`-ing the file (mtime change only, content identical) is enough to flip the cached
run to the failing result — so the cache had recorded a clean verdict that no longer
matched the file's contents.

## Why it matters

This defeats the whole point of the local gate: `check-all.sh` is supposed to mean
"CI will pass". A cached-clean ruff verdict makes a green Gate 2 unfalsifiable, and the
failure only surfaces after a push, a CI queue and a Ralph tick. It is silent, it is
reproducible, and it costs a full Gate 3 cycle every time.

## Proposed fix

Have the lint/format scripts run ruff cache-free so local matches CI byte for byte:

- `scripts/lint.sh` — `ruff check .` -> `ruff check --no-cache .`
- `scripts/format.sh` — same for the `--select I` and `ruff format --check` invocations

Ruff is fast enough that a cold run on this tree (503 files) is a couple of seconds, so
the cost is negligible against a wasted CI round trip.

## Acceptance criteria

- [ ] `check-all.sh` fails on a tree that CI fails, with no dependence on `.ruff_cache` state
- [ ] A regression test or documented manual check that a poisoned cache cannot produce a green Gate 2
- [ ] No threshold, rule selection or gate is weakened as part of the change

PR #1134, opened 2026-08-01
fix(vault): drain short os.write calls so bodies and index lines cannot truncate

### LINKED ISSUE #987: [scan:bugs] Unchecked os.write short writes silently truncate fragment bodies and index lines (3 sites)
## Role
suppressions).

## Goal
Make the three raw `os.write(fd, encoded)` calls in the vault/save write path
loop until every byte is written (or raise), so a partial write can never leave
a silently truncated fragment file or a half-written index line — verified by a
failing-first test that stubs `os.write` to return a short count and asserts the
on-disk bytes equal the intended content.

## Context
- File(s):
  - `creek-tools/creek/vault/writer.py:1293` — `VaultWriter._atomic_create` (fragment body write)
  - `creek-tools/creek/vault/writer.py:1248` — `VaultWriter._append_index_entry` (id-index JSONL append)
  - `creek-tools/creek/save/writer.py:250` — `_atomic_create` (saved-artifact body write)
- Scanned at commit: `6ddafc8373002acc3ef8f838bbd926c9e807b8ce` — re-verify against HEAD before starting
- Evidence:
  - All three sites discard `os.write`'s return value:
    ```python
    flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL
    encoded = content.encode("utf-8")
    ...
    fd = os.open(str(candidate), flags, 0o644)
    try:
        os.write(fd, encoded)      # <-- return value (bytes actually written) discarded
    finally:
        os.close(fd)
    return candidate               # <-- reported as a complete write
    ```
  - `os.write` is a thin `write(2)` wrapper. POSIX permits a **short write** on a
    regular file: on `ENOSPC` the kernel writes what fits and returns that count
    without an error, and a signal arriving after some bytes are written also
    returns a partial count (PEP 475 only retries `EINTR` when *zero* bytes were
    written). The file is then truncated with no exception raised.
  - Reproduced at the scanned SHA against `creek/save/writer.py::_atomic_create`
    with `os.write` stubbed to write half the buffer:
    ```
    intended bytes : 411
    on-disk bytes  : 205
    silent truncation, no exception: True
    tail preserved?: False
    ```
  - The same module already knows the correct idiom: `_atomic_write_text`
    (`creek-tools/creek/vault/writer.py:375-398`) writes through
    `os.fdopen(tmp_fd, "w", ...)`, whose buffered `.write()` loops until the
    buffer is drained. Two write helpers in one file, only one of them durable.
  - Blast radius is worse than a single short file: `_atomic_create` returns the
    path as if the write succeeded, so `VaultWriter` goes on to append an index
    entry claiming a complete fragment exists, and `_append_index_entry` has the
    identical defect — a short append writes a partial JSON line, which
    `_load_index_file` then discards at `json.JSONDecodeError` (`writer.py:1178`),
    silently dropping that fragment from the id index.
- Related: `#917` (same-batch attachment collision — the other silent
  data-loss defect in a write path); `#926` (silently skipped unparseable
  fragments — the read-side analogue of failures that never reach the operator).

## Output Format
"Closes #N".

## Examples
Before (all three sites) — a partial write is indistinguishable from success:

```python
fd = os.open(str(candidate), flags, 0o644)
try:
    os.write(fd, encoded)
finally:
    os.close(fd)
return candidate
```

After — one shared helper that drains the buffer or fails loudly, used by all
three call sites:

```python
def _write_all(fd: int, data: bytes) -> None:
    """Write every byte of *data* to *fd*, looping over short writes.

    ``os.write`` may write fewer bytes than requested (ENOSPC, or a signal
    after partial progress) and return that count without raising, which
    would silently truncate a fragment.
    """
    written = 0
    while written < len(data):
        count = os.write(fd, data[written:])
        if count <= 0:  # no forward progress — fail loudly, never truncate
            msg = f"short write: {written}/{len(data)} bytes written"
            raise OSError(msg)
        written += count
```

Reproducing test shape (`tests/test_vault_writer.py` / `tests/test_save_writer.py`):

```python
def test_atomic_create_survives_a_short_write(tmp_path, monkeypatch):
    real_write = os.write
    monkeypatch.setattr(os, "write", lambda fd, b: real_write(fd, b[: len(b) // 2]))
    path = _atomic_create(tmp_path, "demo", "x" * 400 + "TAIL")
    assert path.read_text() == "x" * 400 + "TAIL"  # fails today: content is halved
```

Note that `O_APPEND` in `_append_index_entry` does not save it either — the
PIPE_BUF guard above it bounds *interleaving*, not short writes; the loop is
still required, and a retry there must re-`os.write` the remaining slice rather
than the whole line.

## Constraints
  (e.g. the absent `fsync` on these paths is a separate durability question and
  is explicitly **out of scope** here)
- Keep the `O_CREAT | O_EXCL` collision-retry semantics and the `O_APPEND`
  PIPE_BUF guard exactly as they are; only the write itself changes

Geoffe-Ga:
## Summary

- **Fixes a silent data-loss bug in the vault write path.** Three raw `os.write(fd, encoded)` call sites discarded the byte count `write(2)` returns. POSIX permits a short write on a regular file (`ENOSPC` writes what fits and returns that count with no error; a signal arriving after *n* > 0 bytes returns a partial count — PEP 475 only auto-retries `EINTR` when *zero* bytes were written), so a fragment body or an id-index JSON line could be truncated with no exception, no log and no trace. Reproduced at `c63befc`: **404 intended bytes, 202 on disk, no exception raised.**
- Adds one stdlib-only module `creek/_fsio.py` with `write_all(fd, data)` (drains the buffer, raising `OSError(EIO)` if a write makes zero forward progress rather than spinning) and `create_exclusive(path, data)` (`O_CREAT|O_EXCL` open + full drain + **unlink-the-partial-file on failure**). All three sites now call into it, so the raw-fd pattern has exactly one implementation to audit instead of three copies.
- The unlink matters because these sites are **not** tempfile+`os.replace` staged — they `os.open` the *final* path directly, so a truncated write lands at the real filename. Without cleanup the fix would only have converted "silently truncated file" into "truncated file plus an exception". `FileExistsError` still propagates untouched, because both callers' counter-suffix retry loops key on it.
- Blast radius closed end-to-end: `_atomic_create` used to return the path as if the write succeeded, after which `VaultWriter` appended an index entry claiming a complete fragment existed — and `_append_index_entry` had the identical defect, writing half a JSON line that `_load_index_file` then discarded at `json.JSONDecodeError`, silently dropping the fragment from the id index.

## What changed

| Site | Before | After |
|---|---|---|
| `creek/vault/writer.py` `VaultWriter._atomic_create` | `os.open` / `os.write` / `os.close` | `create_exclusive(candidate, encoded)` |
| `creek/vault/writer.py` `VaultWriter._append_index_entry` | `os.write(fd, encoded)` | `write_all(fd, encoded)` |
| `creek/save/writer.py` `_atomic_create` | `os.open` / `os.write` / `os.close` | `create_exclusive(candidate, encoded)` |

`_MAX_FILENAME_COLLISION_RETRIES` / `_MAX_COLLISION_RETRIES`, the counter-suffix scheme, the `RuntimeError` messages and the `O_APPEND` `_PIPE_BUF_BYTES` guard are all byte-identical — only the write itself changed. `import os` was dropped from `creek/save/writer.py` (no longer referenced).

Deliberately **not** done, per the issue text: `fsync` / durability on these paths is out of scope.

## Notes on the design

- **`EINTR` needs no explicit handling.** Per PEP 475 `os.write` retries the syscall itself on `EINTR`, which the kernel reports only when *zero* bytes were transferred; a signal after *n* > 0 bytes is reported as a short **success** — precisely the case the loop drains. An `except InterruptedError: continue` branch would be unreachable dead code, so `write_all` carries a comment explaining why it is absent rather than a branch nobody can cover.
- **The `O_APPEND` site cannot be fully repaired here.** Retrying the remaining slice is the right call (the issue sanctions it explicitly), but if a line proves genuinely un-drainable, a remnant without its trailing newline survives and the *next* successful append is concatenated onto it — the merged line fails `json.loads`, so `_load_index_file` drops **both** entries. `ftruncate`-back is not a safe alternative: with `O_APPEND` the write offset is chosen atomically inside the syscall, so we never reliably learn our own start offset and a concurrent appender may already have landed a complete line past ours. This is now documented in the docstring, pinned by a characterization test that says it must be *inverted* rather than deleted, and filed as **#1120** for a rewrite-under-lock or framing fix.

## Test plan

Gate 1 (RED verified against unfixed HEAD before implementing — every one of these failed on an assertion, not an import error):

```
FAILED test_vault_writer.py::TestAtomicWriteHardening::test_atomic_create_writes_full_body_under_short_writes
    AssertionError: assert 'xxxx...xxxx' == 'xxxx...TAIL'
FAILED test_vault_writer.py::TestAtomicWriteHardening::test_atomic_create_leaves_no_file_when_write_makes_no_progress
    Failed: DID NOT RAISE <class 'OSError'>
FAILED test_vault_writer.py::TestAtomicWriteHardening::test_append_index_entry_writes_complete_line_under_short_writes
    assert '{"filename":...5-01-15-demo.' == '{"filename":...ort-write"}\n'
FAILED test_vault_writer.py::TestAtomicWriteHardening::test_append_index_entry_raises_when_write_makes_no_progress
    Failed: DID NOT RAISE <class 'OSError'>
FAILED test_vault_writer.py::TestAtomicWriteHardening::test_write_fragment_survives_short_writes_end_to_end
FAILED test_save.py::test_save_writes_full_body_under_short_writes
FAILED test_save.py::test_save_atomic_create_leaves_no_file_when_write_makes_no_progress
7 failed, 1 passed
```

(`tests/test_fsio.py` failed collection with `ModuleNotFoundError: No module named 'creek._fsio'` — the correct RED for a module that does not exist yet. The 1 pass is the pre-existing collision-retry test, confirming it was not disturbed.)

The tests **force** the short write rather than exercising a happy path: a `short_write` fixture in `tests/conftest.py` wraps `os.open` to record the descriptors the code under test opens, then shortens `os.write` for *only* those descriptors (`.halve()`, `.one_byte()`, `.stall()`, `.fail_after_half(errno)`, `.passthrough()`), delegating verbatim for everything else so pytest's capture and coverage writes are untouched. `os.close` is wrapped to untrack the fd so a recycled descriptor number cannot be shortened by accident.

Gate 1 GREEN — 16 new tests:

- `tests/test_fsio.py` (8): full drain under halving writes; one-byte-per-call (pins the offset arithmetic — exactly `len(payload)` calls, so an off-by-one or a mutant that re-writes the whole buffer each iteration fails); zero-progress raises `EIO` instead of spinning; empty payload issues no syscall; `create_exclusive` drains fully; `FileExistsError` propagates and leaves a pre-existing file's bytes untouched; mid-write `ENOSPC` unlinks the partial file; zero-progress unlinks. Every `OSError` test asserts `excinfo.value.errno`, so a bare or wrong-errno raise still fails.
- `tests/test_vault_writer.py` (6, in `TestAtomicWriteHardening`): both raw sites, a `write_fragment` end-to-end case asserting the body round-trips *and* `_load_index_file` resolves the fragment id, and the #1120 characterization test.
- `tests/test_save.py` (2): through the public `save_to_vault`.

One pre-existing test changed by exactly one expression — `monkeypatch.setattr(writer_module.os, ...)` → `monkeypatch.setattr(os, ...)`, mechanically required because `creek/save/writer.py` dropped its now-unused `import os`. Both targets are the same `os` module singleton; all three of its assertions are byte-identical. `git diff -U0 tests/ | grep '^-'` shows only that line and a widened `TYPE_CHECKING` import — no assertion anywhere was deleted or loosened.

Gate 2 — `cd creek-tools && ./scripts/check-all.sh`: **12/12 checks passed, 7455 tests passed, 94.13% aggregate branch coverage**, per-file gate 224 files ≥80%. `creek/_fsio.py` sits at 94.59% per-file. `ruff check --no-cache .` re-verified clean (imports changed, and a cached verdict would not be trustworthy — see #1119).

Gate 2.5 — `code-review-orchestrator` over the diff. Round 1 returned FIX_REQUIRED on one MAJOR (the `O_APPEND` residual was under-documented and untested); fixed via the docstring correction, the characterization test and issue #1120, along with its two non-blocking findings (the `close()`-errno caveat and the fd-tracking hygiene). Round 2 returned CLEAN.

## Audit of the rest of the codebase

`grep -rn 'os\.write|os\.read|os\.pwrite|os\.sendfile'` over `creek/`, `creek_mcp/` and `crawdad/` returns **exactly these three sites and nothing else** — there is no fourth instance to file a follow-up for. Every other `os.open` in the tree (`creek/ingest/gdrive.py:479`, `creek/confidential/keyvault.py:364`, `creek/redact/cli_commands.py:417`, `creek/vault/writer.py:391`) immediately wraps the descriptor in `os.fdopen`, whose buffered `.write()` already drains, so none of them are affected.

## Detecting damage that already happened

A vault written by an earlier version could already contain a truncated fragment, and **nothing in the tree can detect it.** There is no content checksum or byte-length record anywhere on the fragment write path: the hash-chained provenance log (`creek.audit.AuditLog`) records only `{id, type, path, written_at}`, and it chains *log lines*, not file contents. None of the ten `creek lint` checks (broken-links, compost, draft-grounding, orphan-compiled, paradox, skill-size-budget, synchronicity, tags, unnamed, voice-fidelity) verify body completeness. A note truncated *after* its closing frontmatter delimiter parses cleanly and simply has a shorter body — indistinguishable from a short note. A note truncated *inside* the frontmatter loses its `id` and falls out of the index, which is at least noticeable as an absence, but is never reported as corruption. This PR stops new damage; it cannot find old damage, and detecting it would need a content-integrity feature that does not exist today.

Closes #987
Refs #1120

### DIFF
--- a/creek-tools/creek/_fsio.py
+++ b/creek-tools/creek/_fsio.py
@@ -1,0 +1,124 @@
+"""Short-write-safe primitives for raw file-descriptor I/O (#987).
+
+``os.write`` is allowed to write *fewer* bytes than it was handed: the
+POSIX ``write(2)`` contract only promises some forward progress, so a
+signal, a nearly-full disk or a slow device can all cut a write short.
+Code that discards the returned count silently truncates its file.
+
+Both :mod:`creek.vault.writer` and :mod:`creek.save.writer` create notes
+by writing straight to the *final* path (``O_CREAT | O_EXCL``, with no
+tempfile + ``os.replace`` staging step), so a short write there does not
+merely spoil a scratch file — it files a half-written note under the
+real name, with no exception raised for anyone to notice. These two
+helpers live here instead of being triplicated across those call sites,
+so the drain loop and the partial-file cleanup have exactly one
+implementation to audit.
+
+Stdlib-only by design: this module sits underneath the writers and must
+not drag any of the package in behind it.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import errno
+import os
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+_EXCLUSIVE_CREATE_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_EXCL
+"""Create-or-fail open flags: never truncates or clobbers an existing file."""
+
+_NEW_FILE_MODE = 0o644
+"""Permission bits requested for files created by :func:`create_exclusive`."""
+
+
+def write_all(fd: int, data: bytes) -> None:
+    """Write every byte of *data* to *fd*, looping over short writes.
+
+    Args:
+        fd: An open, writable file descriptor.
+        data: The payload to drain in full. An empty payload is a no-op
+            and issues no syscall.
+
+    Raises:
+        OSError: With ``errno.EIO`` when a single ``os.write`` reports no
+            forward progress, rather than spinning on a descriptor that
+            has stopped accepting bytes; or with the underlying errno
+            when the write itself fails. Either way the bytes already
+            written stay on the descriptor — a caller that owns a
+            freshly created file should discard it (see
+            :func:`create_exclusive`).
+    """
+    # The memoryview keeps ``view[offset:]`` O(1). Slicing ``data``
+    # directly would copy the unwritten remainder on every iteration,
+    # which turns a dribbling descriptor quadratic on the 35k-fragment
+    # write path.
+    view = memoryview(data)
+    total = len(view)
+    offset = 0
+    while offset < total:
+        # No ``InterruptedError`` branch on purpose: per PEP 475
+        # ``os.write`` retries the syscall itself on ``EINTR``, which the
+        # kernel reports only when *zero* bytes were transferred. A
+        # signal arriving after n > 0 bytes is reported instead as a
+        # short *success* — precisely the case this loop drains — so
+        # such a branch would be unreachable dead code. A genuine
+        # ``OSError`` (``ENOSPC`` on a later pass, say) must propagate.
+        written = os.write(fd, view[offset:])
+        if written <= 0:
+            msg = f"short write: {offset}/{total} bytes written"
+            raise OSError(errno.EIO, msg)
+        offset += written
+
+
+def create_exclusive(path: Path, data: bytes) -> None:
+    """Create *path* with ``O_CREAT|O_EXCL`` and write every byte of *data*.
+
+    One caveat, on errno fidelity rather than data safety: the
+    descriptor is closed in an inner ``finally``, so if :func:`write_all`
+    raises (an ``ENOSPC``, say) and ``os.close`` then *also* raises, the
+    close-time exception replaces the original and the caller sees the
+    close errno instead of the true root cause. ``os.close`` deliberately
+    does not retry on ``EINTR`` (PEP 475), so this is narrow but
+    reachable. The unlink cleanup still fires in that case — the handler
+    below catches any ``OSError`` subtype — so the guarantee that no
+    half-written file survives at the real path is unaffected.
+
+    A *successful* :func:`write_all` followed by a failing ``os.close``
+    likewise unlinks the (complete) file and raises. That is deliberate
+    fail-safe behaviour: a failing ``close`` can be the only signal of a
+    delayed write-back failure, so suppressing it would trade an
+    errno-fidelity gap for genuine silent data loss.
+
+    Args:
+        path: File to create; its parent directory must already exist.
+        data: Full contents to write.
+
+    Raises:
+        FileExistsError: When *path* already exists. Propagated
+            untouched: both writers' counter-suffix retry loops key on
+            it, and the cleanup below deliberately does not fire for a
+            file this call never created.
+        OSError: When the payload cannot be drained. The partial file is
+            unlinked before the error propagates, so a truncated body is
+            never promoted to the real path.
+    """
+    fd = os.open(str(path), _EXCLUSIVE_CREATE_FLAGS, _NEW_FILE_MODE)
+    try:
+        # Closing in the inner ``finally`` releases the descriptor
+        # before the cleanup unlink runs, however the drain ended.
+        try:
+            write_all(fd, data)
+        finally:
+            os.close(fd)
+    except OSError:
+        # Unlink failures are suppressed so the error already in flight
+        # is the one the caller sees — the drain's ``ENOSPC``, say, or
+        # the ``os.close`` failure that displaced it in the inner
+        # ``finally`` (see the docstring caveat).
+        with contextlib.suppress(OSError):
+            path.unlink()
+        raise
--- a/creek-tools/creek/save/writer.py
+++ b/creek-tools/creek/save/writer.py
@@ -13,19 +13,21 @@ path for fragments and rolled-up primitives; save needs to land notes
 at additional vault locations (``10-Liminal/Paradoxes/``,
 ``10-Liminal/Unnamed/``, ``07-Voice/Drafts/``) that VaultWriter does
 not own. Sharing the same atomic-create pattern keeps the on-disk
-behaviour consistent.
+behaviour consistent: the raw-descriptor create itself is
+:func:`creek._fsio.create_exclusive`, shared verbatim with
+:mod:`creek.vault.writer`.
 """
 
 from __future__ import annotations
 
 import hashlib
-import os
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
 
 import frontmatter
 
+from creek._fsio import create_exclusive
 from creek.classify.privacy_filter import pre_save_filter
 from creek.models import (
     Authorship,
@@ -236,20 +238,32 @@ def _compose_base_name(title: str) -> str:
 
 
 def _atomic_create(target_dir: Path, base_name: str, content: str) -> Path:
-    """Create ``target_dir/{base_name}.md`` atomically; retry on collision."""
-    flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL
+    """Create ``target_dir/{base_name}.md`` atomically; retry on collision.
+
+    Args:
+        target_dir: Directory the note will land in.
+        base_name: Filename stem (without ``.md``).
+        content: Full note contents to write.
+
+    Returns:
+        The path of the created note.
+
+    Raises:
+        RuntimeError: If a unique filename cannot be obtained within
+            :data:`_MAX_COLLISION_RETRIES` attempts.
+        OSError: If the note was created but its contents could not be
+            written in full. The partial file is unlinked, so a
+            truncated answer is never filed under the operator's title
+            (#987).
+    """
     encoded = content.encode("utf-8")
     for counter in range(_MAX_COLLISION_RETRIES):
         suffix = "" if counter == 0 else f"-{counter}"
         candidate = target_dir / f"{base_name}{suffix}.md"
         try:
-            fd = os.open(candidate, flags, 0o644)
+            create_exclusive(candidate, encoded)
         except FileExistsError:
             continue
-        try:
-            os.write(fd, encoded)
-        finally:
-            os.close(fd)
         return candidate
     msg = (
         f"Could not allocate a unique filename for '{base_name}.md' in "
--- a/creek-tools/creek/vault/writer.py
+++ b/creek-tools/creek/vault/writer.py
@@ -31,6 +31,7 @@ from typing import TYPE_CHECKING
 import frontmatter
 import yaml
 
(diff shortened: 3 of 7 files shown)


Geoffe-Ga:
## Summary
Fixes a genuine silent-data-loss bug (#987): three raw `os.write(fd, encoded)` call sites in `creek/vault/writer.py` and `creek/save/writer.py` discarded the byte count `write(2)` can legitimately under-report, so a short write could truncate a fragment body or an id-index JSON line with no exception. The PR adds a single stdlib-only module `creek/_fsio.py` (`write_all` / `create_exclusive`) and routes all three sites through it, with unlink-on-failure cleanup since these are direct-to-final-path writes with no tempfile+rename staging.

## Strengths
- Root cause and blast radius are precisely diagnosed (traced through to `_load_index_file` silently dropping fragments on `JSONDecodeError`).
- One shared implementation instead of three copies of the raw-fd pattern — exactly the right fix shape.
- `write_all` uses `memoryview` slicing to avoid O(n²) behavior on a dribbling descriptor — a subtlety many implementations would miss.
- `EINTR` is correctly *not* special-cased, with a comment explaining why (PEP 475 auto-retries only on zero-byte interrupts), avoiding dead code.
- The one known residual gap — a partial `O_APPEND` remnant colliding with the next entry — is not swept under the rug: it's documented in the docstring, pinned by an explicit characterization test that states it must be *inverted* (not deleted) when fixed, and filed as a separate follow-up issue (#1120) rather than silently scoped into this PR.
- `create_exclusive`'s error-path interaction (write failure vs. subsequent `close()` failure clobbering the original exception) is called out precisely in the docstring rather than glossed over.
- `FileExistsError` is verified to propagate untouched (both callers' collision-retry loops key on it), with a test asserting the pre-existing file's bytes are untouched.
- The `short_write` test fixture is a nice piece of infrastructure: fd-scoped tracking (with fd-recycling hygiene via `os.close` wrapping) so it never perturbs pytest's own capture/coverage I/O.

## Security Concerns
None. 🟢 This closes a data-integrity gap rather than introducing one; no new user-controlled input paths or trust boundaries.

## Problems
None found that block merge.

## Code Quality
- (non-blocking) `ShortWriteController.calls` and `fail_after_half`'s "first call" check are global across all tracked descriptors rather than per-fd — fine for the current single-fd-per-test usage, but worth a comment if the fixture is later reused in a multi-descriptor scenario.
- (non-blocking) Raising a bare `OSError(errno.EIO, ...)` for the zero-progress case is a reasonable, minimal choice consistent with stdlib conventions; a dedicated exception type was considered and reasonably rejected as unnecessary here.

## CLAUDE.md Adherence
✅ TDD (RED verified against unfixed HEAD before implementing, per PR description) → GREEN → refactor.
✅ `./scripts/check-all.sh` reported 12/12 passing, 94.13% aggregate branch coverage, per-file gate satisfied, docstrings present throughout (Args/Returns/Raises on all new/changed public and private functions).
✅ Conventional commit-style PR title (`fix(vault): ...`).
✅ Went through the repo's Gate 2.5 code-review-orchestrator per the PR description, with the one MAJOR finding from round 1 addressed before round 2 came back clean.
✅ Audited the rest of the tree for other unguarded `os.write` sites and confirmed none remain.

## Testing
Strong. 16 new tests covering: full drain under halved/one-byte writes, zero-progress → `EIO` (not spinning), empty-payload no-op, `FileExistsError` propagation with byte-for-byte preservation of the pre-existing file, mid-write `ENOSPC` triggering unlink, an end-to-end `write_fragment` test asserting both body and index survive, and the deliberate characterization test for the still-open `O_APPEND` residual defect. Tests assert on raw bytes before parsing (so a truncated frontmatter delimiter can't mask the real failure) and check `excinfo.value.errno` rather than just exception type, so a wrong-errno raise would still fail.

## Documentation
Thorough. Every new/changed function carries Google-style docstrings with Args/Returns/Raises, including subtle caveats (close-time errno clobbering, why `ftruncate`-back isn't safe under concurrent `O_APPEND`). The module docstring for `_fsio.py` explains the stdlib-only design constraint. PR description itself is exceptionally detailed with reproduction evidence, a before/after site table, and an audit of the rest of the codebase.



## Verdict: LGTM

Real data-loss bug fixed with a single well-tested primitive, comprehensive TDD-driven test suite (RED verified before implementation, GREEN after), all CLAUDE.md quality gates reportedly passed, and the one known residual gap (O_APPEND remnant on repeated failure) is explicitly scoped out, documented, and pinned by a characterization test with a tracked follow-up issue (#1120) rather than silently left. No blocking bugs, security issues, or missing tests found on review of the actual diff.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5149483199 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

PR #1135, opened 2026-08-01
fix(classify): escalate the tier on new evidence, not on tier ownership

### LINKED ISSUE #1105: fix(classify): a non-force LLM run writes the confessional+conviction INTIMATE trigger and discards it — owns_tier gates the reassess out

Geoffe-Ga:
## Summary

- **The bug.** `classify_engine._prepare_fragment` computed `owns_tier = needs_tier(raw) or force` and gated the post-classification privacy reassess on it. `needs_tier` is `False` for any concrete tier, so on a non-`--force` run over an already-tiered fragment the reassess never ran — even when that same run had just produced `voice_register: confessional` + `confidence: conviction`, which is `PrivacyClassifier.classify_tier`'s third INTIMATE trigger. **The run wrote the intimate trigger to disk and discarded the tier verdict in the same pass.** Verified on unfixed HEAD: `privacy_tier: personal` next to `voice: {confidence: conviction, voice_register: confessional}` and `voice_proxy_eligible: True`. That `true` is the damage — it admits confessional material into voice-proxy generation.
- **The design decision, and why this option.** The guard's stated rationale was "never silently raise a deliberate operator tier", but **there is no tier provenance in the frontmatter**: the engine cannot tell an operator's considered `personal` from one `_retier_after_rewrite` or `Pipeline._run_classification` stamped a millisecond earlier. This PR replaces the ownership *proxy* with the predicate it was standing in for: `privacy_pass.reassess` gains a required `baseline` — the heuristic's verdict on the fragment **as loaded** — and escalates only when *this run's* classification made the heuristic **strictly more restrictive**. We never need to know who wrote the tier, because we act only on evidence that did not exist when it was written.
- **Strictly-more-restrictive, not merely-different.** A different candidate includes a *weaker* one; acting on a weakening verdict would let a flip-flopping model ratchet an operator's tier up off evidence pointing the other way. Implemented by reusing `escalate` (`escalate(baseline, candidate) is baseline` → decline), so no new rank table and no fourth tier reader.
- **Settled repo-wide.** `creek/pipeline.py` carried the identical gate (added by #1113 to mirror the engine). It is changed the same way, so `creek process` and `creek classify` remain one decision rather than two — this is the fifth instance of the "wired into one entry point but not the other" class, and fixing it in one place only would have produced the sixth.
- **Strictly additive semantics.** For every case the old gate *permitted*, the new predicate is provably a no-op: on the assigning path `apply_tier` already stamped `candidate_pre == baseline`; on the `--force` path it already merged `escalate(on-disk, baseline)`. The only behaviour change is the bug case.

### Two options considered and rejected (the issue body offered both)

**Unconditional reassess.** `reassess` re-derives from the *same* platform/author/body evidence `apply_tier`'s pre-pass just declined to act on, so this makes `apply_tier`'s operator-override branch dead in practice — honoured before the model runs, overridden milliseconds later. And because `privacy_tier` is a one-way ratchet, the operator could then **never durably hold a tier below the heuristic**: hand-edit back to `open`, next run re-buries it at `intimate`, forever. That matters because `RECOVERY_KEYWORDS` documents itself as *"intentionally errs on the side of high-recall: false positives land in INTIMATE"* — false-positive INTIMATE verdicts are a designed-in property, and the operator's override is their **only** correction mechanism. "Raising is non-destructive" is true for privacy but false for utility: `intimate` permanently removes voice-proxy eligibility and cloud routing, with no tooling path back.

**A `privacy_tier_source: heuristic | operator` key.** The information is unobtainable without operator cooperation, so the key lies in both default directions. Default-to-`operator` leaves every already-mis-stamped vault broken forever; machine-self-stamps-`heuristic` means a human who edits a machine-stamped `personal` down to `open` leaves the stale marker in place and gets overridden anyway. Only greenfield hand-written frontmatter would be protected — a schema change buying an unreliable signal.

### What happens to a genuinely hand-set tier

Unchanged in the case that matters, and it now self-corrects in the case that does not. A hand-set `open` on a journal fragment survives **every** run: platform is not new evidence, so baseline and candidate are both `INTIMATE` and the pass declines (`test_manual_tier_override_survives_a_non_force_run` passes unmodified). When the machine *does* raise a tier on a genuinely new signal, it persists the voice axes that justified it — so the operator, having now seen the evidence, can hand-edit the tier back down and it **sticks**: on the next run those axes are already on the loaded fragment, the baseline moves with them, and nothing hardens. The operator overrules the machine once they have seen its reasoning. `reassess`'s docstring records the honest residual: if the model's verdict is *unstable*, a later re-hardening is genuine new evidence and will raise again — one step per flip.

### Callers affected

| Caller | Behaviour |
|---|---|
| `creek fill --upgrade` (`cli.py:2342`, `method="llm", force=False`) | **Changes** — the headline case; its whole purpose is re-running over already-tiered fragments |
| `creek sync` Tier B (`cli.py:750` → `_sync_classify(..., "llm")`, `force=False`) | **Changes** — biggest real-world beneficiary; re-runs over an already-tiered vault every cycle |
| `creek classify` (`cli.py:1698`) | **Changes** when non-force + concrete tier + not preserved + the verdict hardens. `--force`: provably identical |
| MCP `creek.classify` (`creek_mcp/tools/classify.py:54`) | **Changes** identically; no MCP-side edit needed |
| `creek process` (`Pipeline._run_classification`) | **No production change** — no ingester sets `privacy_tier`, so `owns_tier` was always `True`. Seam contract only |
| Preserved short-circuit (`classification_method: llm`/`manual`) | **Unchanged** — classification never runs, so there is no new evidence; `_persist_tier_only` is untouched |
| `creek ingest`, `creek fill` without `--upgrade`, `VaultWriter._retier_after_rewrite` | Unchanged |

### Sibling issues — stated honestly, not overclaimed

- **#1106 (mis-stamped vaults): this PR does NOT remediate them.** I probed it directly. A fragment carrying `personal` + confessional/conviction already on disk stays `personal` / `voice_proxy_eligible: true` after a non-force run, because that evidence is no longer new — the design working as intended, not a gap. (The same probe confirms `--force` still fixes it, even on the free `--method rules` path, so #1106 is a *targeting* problem rather than a capability gap.) #1106's `retierable` gap detector remains needed and unaffected: `_scan_fill_gaps` still counts only `needs_tier(raw)` (`cli.py:2184`), so a wrongly-`personal` fragment is still invisible.
- **#1107 (parity detector blind to the LLM path): this PR does NOT make widening it cheap.** The two new reproduction tests stub the model through two *different* idioms — `patch("...LLMClassifier.classify_with_reasoning")` in `test_classify_engine.py` vs `monkeypatch.setattr(classifier, "classify")` in `test_pipeline.py`. That is direct evidence the shared deterministic-model harness #1107 needs does not fall out of this work; we add two more instances of the divergence it exists to unify. Left to #1107.

### One existing test's expectation deliberately changed

`tests/test_pipeline.py::TestPipelinePrivacyReassess::test_a_preset_ingester_tier_is_not_silently_raised` was the only thing pinning the `owns_tier` guard on the pipeline side, and its own docstring warned against removing it. **It is rewritten, not deleted**, as `test_a_preset_tier_is_raised_only_when_the_verdict_hardens_it`, and is strictly stronger: two fragments, both directions (one hardening verdict raises `open`→`intimate`, one analytical verdict leaves `open` alone), plus `voice_proxy_eligible` and keyed-by-id landed-verdict assertions the original lacked. Its docstring states that it supersedes the old contract by decision of #1105 and preserves the original's still-true "seam contract, not a live content path" observation. This is the **only** existing expectation that changed; the other four test edits are mechanical `baseline=` additions to `TestReassess` with every pre-existing assertion surviving verbatim.

## Test plan

Gate 1 RED verified against unfixed HEAD before implementing — the headline reproduction failed with `{'frag-owned-conf01': 'personal'} != {'frag-owned-conf01': 'intimate'}`, and the pipeline test with `{'frag-reassess007': 'open'} != {'frag-reassess007': 'intimate'}`.

New tests:
- `tests/test_privacy_pass.py::TestReassess` — 4 new (hardened signal flips `voice_proxy_eligible`; declines when the candidate matches the baseline; **declines when the candidate weakened** — the test a `!=` predicate fails; a hardened signal still never lowers the stored tier) + 4 mechanical `baseline=` migrations.
- `tests/test_classify_engine.py` — `test_a_confessional_verdict_escalates_an_already_tiered_fragment` (the reproduction, with a second pre-stamped `personal` fragment on an analytical verdict so blanket escalation also fails), `test_an_operator_tier_survives_when_the_run_produces_no_new_signal` (fails under the unconditional-reassess option), `test_a_weakened_verdict_does_not_raise_an_operator_tier`, `test_a_preserved_fragment_is_not_retiered`.
- `tests/test_pipeline.py` — the rewritten preset-tier test above.

Pins confirmed still green **unmodified**: `test_classification_signals_escalate_the_tier`, `test_an_existing_intimate_tier_is_never_lowered` (both params), `test_manual_tier_override_survives_a_non_force_run`, `test_second_classify_run_assigns_no_tiers_and_changes_none`, `test_reatomized_children_carry_a_concrete_tier`, all of `tests/test_process_classify_parity.py`, and the rest of `TestPipelinePrivacyReassess`.

Gates run locally from `creek-tools/`:
- `./scripts/check-all.sh` → **exit 0**, 12/12 checks, **7463 passed** (up from 7455 on unmodified HEAD), 94.13% branch coverage, per-file gate 224 files ≥80%.
- `ruff check --no-cache .` with `.ruff_cache` deleted → clean (cold verification, per #1119).
- `code-review-orchestrator` over the diff → **CLEAN**, no blocking findings. Its one non-blocking nit (the now-narrowed `owns_tier` local reads like the seam it replaced) was applied as a clarifying comment.
- No `# noqa`, `# type: ignore`, `skip`/`xfail`, or threshold change anywhere in the diff. `pyproject.toml`, `uv.lock`, `creek/vault/writer.py` and `creek_mcp/` untouched.

Closes #1105
Refs #974, #1106, #1107

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -53,9 +53,19 @@ call it:
 * **Escalate-only, never auto-downgrade.** ``--force`` may raise a tier
   that is too light, and the post-classification reassess may raise one
   the classifier just hardened, but neither ever lowers a tier —
-  lowering is the one direction that leaks content. An explicit
-  non-``unclassified`` tier on disk also survives any non-``--force``
-  run untouched: the operator's decision outranks the heuristic.
+  lowering is the one direction that leaks content.
+* **A settled tier moves only on new evidence** (#1105). A concrete
+  non-``unclassified`` tier on disk survives a non-``--force`` run
+  *assignment* untouched — :func:`~creek.classify.privacy_pass.apply_tier`
+  will not overwrite it. The post-classification reassess can still raise
+  it, but only when *this run's* classification made the privacy
+  heuristic strictly more restrictive than it was on the fragment as
+  loaded (:attr:`_PreparedFragment.tier_baseline`). So a re-run over
+  unchanged evidence never disturbs a tier the operator chose, while a
+  freshly-revealed INTIMATE signal is no longer discarded just because
+  some tier was already on record. A fragment the resume short-circuit
+  preserves is not reassessed at all: this run produced no evidence
+  about it.
 
 Praxis pass (issue #877)
 ------------------------
@@ -708,18 +718,19 @@ class _PreparedFragment:
             keys are preserved through the rewrite).
         llm: The classifier the fragment's tier resolved to, or ``None``
             on the rules path.
-        owns_tier: Whether *this* run owns the fragment's tier — true
-            when the frontmatter carried none (or ``unclassified``), or
-            when ``--force`` was passed. Gates the post-classification
-            reassess so a hand-set tier is never silently escalated
-            either.
+        tier_baseline: The privacy heuristic's verdict on the fragment
+            **as loaded**, taken before this run classified anything.
+            :func:`~creek.classify.privacy_pass.reassess` compares its
+            own verdict against this to answer the only question that
+            licenses touching a settled tier: did *this run* make the
+            heuristic stricter? (#1105)
     """
 
     fragment: Fragment
     body: str
     raw: dict[str, object]
     llm: LLMClassifier | None
-    owns_tier: bool
+    tier_baseline: PrivacyTier
 
 
 def _prepare_fragment(
@@ -741,6 +752,15 @@ def _prepare_fragment(
     ``llm``/``manual``-stamped fragments that make up a mature vault can
     acquire a tier without ``--force`` re-paying for their classification.
 
+    One more ordering is load-bearing (issue #1105):
+    :attr:`_PreparedFragment.tier_baseline` is captured *first of all*, on
+    the fragment exactly as loaded. It is the reference point
+    :func:`~creek.classify.privacy_pass.reassess` needs to judge whether
+    this run learned anything new about the fragment's privacy, so any
+    mutation taken before it — the tier stamp here, the classifier's voice
+    verdict later — would fold this run's own output back into the
+    baseline and blind the gate to it.
+
     Args:
         md_file: The file to consider.
         force: Whether to re-derive settled classifications and tiers.
@@ -760,8 +780,22 @@ def _prepare_fragment(
         return None
     fragment, body, raw = record
 
+    # #1105: record the heuristic's verdict on the fragment as loaded, BEFORE
+    # ``apply_tier`` (and the classifier below) change anything it reads. The
+    # post-classification reassess needs this to tell "the model just revealed
+    # something new" from "the model echoed what was already on disk"; only the
+    # former licenses raising a tier that is already on record.
+    tier_baseline = privacy.classify_tier(fragment, content=body)
+
     # #876: stamp a real tier before anything else looks at one. ``force``
     # merges escalate-only, so an existing ``intimate`` is never lowered.
+    #
+    # Since #1105 this answers only "did this run *assign* the tier?", and its
+    # two remaining consumers are the ``privacy_tiers_assigned`` counter just
+    # below and the preserved-path :func:`_persist_tier_only` write. It no
+    # longer gates the post-classification reassess — ``tier_baseline`` does
+    # that — because "a tier is already on record" turned out to say nothing
+    # about whether a *person* put it there.
     owns_tier = needs_tier(raw) or force
     fragment = apply_tier(fragment, body, raw=raw, force=force, classifier=privacy)
     if owns_tier:
@@ -797,7 +831,7 @@ def _prepare_fragment(
         body=body,
         raw=raw,
         llm=llm,
-        owns_tier=owns_tier,
+        tier_baseline=tier_baseline,
     )
 
 
@@ -914,10 +948,21 @@ def _process_file(
     # #876: classification may have hardened the fragment's voice signals
     # (confessional + conviction is one of the INTIMATE triggers), which the
     # pre-pass could not see. Take the stricter of the two verdicts —
-    # escalate-only, and only when this run owns the tier, so a deliberate
-    # operator tier is never silently raised either.
-    if prepared.owns_tier:
-        new_fragment = reassess(new_fragment, body, classifier=privacy)
+    # escalate-only, so nothing here can lower a tier.
+    #
+    # #1105: unconditional, because ``reassess`` now carries its own gate.
+    # Its ``baseline`` is the heuristic's verdict on the fragment as loaded,
+    # so it fires only where this run's classification made the heuristic
+    # strictly stricter — new evidence, which no tier on disk predates. The
+    # gate this replaces asked "does this run own the tier?", which is False
+    # for every already-tiered fragment, i.e. for the whole of a mature
+    # vault; that answer discarded exactly the verdicts it was here to catch.
+    new_fragment = reassess(
+        new_fragment,
+        body,
+        baseline=prepared.tier_baseline,
+        classifier=privacy,
+    )
 
     with lock:
         _record_praxis(praxis_before, new_fragment, counts)
--- a/creek-tools/creek/classify/privacy_pass.py
+++ b/creek-tools/creek/classify/privacy_pass.py
@@ -12,15 +12,26 @@ close that hole:
 * :func:`escalate` — merge two candidate tiers, never lowering.
 * :func:`apply_tier` — stamp a tier, honouring a manual override.
 * :func:`reassess` — re-run the check after classification hardened the
-  fragment's voice signals; escalate-only. Both entry points call it,
-  each as its last mutation before the write (#974).
+  fragment's voice signals; escalate-only, and only on new evidence.
+  Both entry points call it, each as its last mutation before the write
+  (#974).
 
-Two rules hold across every function here:
+Three rules hold across every function here:
 
 **Never auto-downgrade.** Lowering a tier is the only direction that
 leaks content, so both the ``--force`` merge and the post-classification
 reassess take the *more* restrictive of the two candidates. An operator
 who wants a tier relaxed edits the frontmatter by hand.
+
+**Act only on new evidence** (#1105). A tier already on record is
+re-litigated only when *this run's* classification made the heuristic
+strictly more restrictive than it was on the fragment as loaded —
+:func:`reassess` compares its own verdict against the caller's
+``baseline`` and returns the fragment untouched when nothing hardened.
+That is why the pass needs no notion of *who* wrote the tier (the
+frontmatter carries no such provenance): it acts only on evidence that
+did not exist when the tier was written, so a settled decision cannot be
+overturned by a re-run over unchanged input.
 
 **Never yield ``unclassified``.**
 :meth:`~creek.classify.privacy.PrivacyClassifier.classify_tier` always
@@ -163,9 +174,10 @@ def reassess(
     fragment: Fragment,
     body: str,
     *,
+    baseline: PrivacyTier,
     classifier: PrivacyClassifier,
 ) -> Fragment:
-    """Re-derive the tier after classification and keep the stricter one.
+    """Re-derive the tier after classification, on new evidence only.
 
     The pre-pass runs *before* the classifier so the per-tier router sees
     a real tier, which means it can only read the axes already on the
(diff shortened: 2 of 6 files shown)


Geoffe-Ga:
## Summary
This PR fixes issue #1105: a non-`--force` LLM run over an already-tiered fragment silently discarded a hardening privacy verdict (e.g., a fresh `confessional`+`conviction` voice classification that should trigger `INTIMATE`), because the post-classification reassess was gated on `owns_tier` (`needs_tier(raw) or force`), which is `False` for any fragment that already carries a concrete tier — i.e., for the whole of a mature vault. The fix replaces that ownership *proxy* with the actual predicate it stood in for: `reassess` now takes a required `baseline` (the privacy heuristic's verdict on the fragment *as loaded*, before this run's classification touched anything) and escalates only when this run's own classification made the heuristic **strictly more restrictive** than that baseline. The same change is applied identically in both `classify_engine.py` and `pipeline.py`.

## Strengths
- **Correct core logic.** I traced `escalate()`'s rank table and the identity-based gate (`escalate(baseline, candidate) is baseline`) — enum singletons make the `is` check exact, and the two-stage gate-then-merge structure correctly separates "did new evidence appear?" from "never lower what's on disk," which the tests probe independently (`test_a_hardened_signal_still_never_lowers_the_stored_tier` deliberately has the gate open while the merge still refuses to move a stored `INTIMATE` down to `PERSONAL`).
- **"Strictly more restrictive," not "merely different," is the right predicate**, and is specifically tested (`test_declines_when_the_candidate_weakened_from_the_baseline` / `test_a_weakened_verdict_does_not_raise_an_operator_tier`) — a naive `!=` gate would let a flip-flopping model ratchet a manually-relaxed tier back up on evidence pointing the other way.
- **Baseline capture point is correct and mirrored across both call sites.** In `classify_engine._prepare_fragment`, `tier_baseline` is captured before `apply_tier` or the classifier touch the fragment. In `pipeline._run_classification`, `baseline` is captured before `rule_classifier.classify` runs, explicitly to keep the two entry points' notion of "new evidence" aligned (documented inline, and this consistency is exactly what #1105/#974's cross-entry-point parity problem is about).
- **Design alternatives are honestly evaluated and rejected in the PR body** (unconditional reassess breaks the operator-override escape hatch permanently; a `privacy_tier_source` provenance key is unreliable in both default directions) — this is unusually rigorous justification for a design decision.
- **Test quality is excellent.** Every new test pairs the case under test with a contrast fragment that would fail under the naive alternative implementation (blanket escalate, `!=` instead of `>`, "never touch any tier"), so the tests can't pass by accident. The one pre-existing test that had to change (`test_a_preset_ingester_tier_is_not_silently_raised`) is rewritten with a documented rationale, not deleted, and is strictly stronger than before (adds a second fragment, `voice_proxy_eligible` assertions, and keyed-by-id checks).
- **Honest limitations documented in code**, not just the PR description — the `reassess` docstring states plainly that an operator's manual downgrade only "sticks" until the model's verdict is unstable, and explains why that's an acceptable one-step-per-flip residual rather than sweeping it under the rug.
- **Sibling issues (#1106, #1107) are explicitly scoped out** with direct probing evidence rather than silently expanding scope.

## Security Concerns
None. 🟢 This is a privacy-classification correctness fix; if anything it closes a real information-exposure gap (confessional content silently remaining `voice_proxy_eligible: true`). No new I/O, no new external input handling, no injection surface touched.

## Problems
None found blocking merge.

## Code Quality
- Non-blocking observation: `privacy_pass.reassess` and both call sites now compute `classify_tier` twice per fragment on the reassess path (once for baseline, once inside `reassess` itself for the candidate) plus once more inside `apply_tier`. `classify_tier` is cheap and pure (confirmed by reading `privacy.py`), so this isn't a real performance concern, just worth noting if profiling ever flags this pass.
- The module docstring in `privacy_pass.py` was updated from "two rules" to "three rules" and reads well; the `_PreparedFragment.tier_baseline` docstring clearly explains why capture-order is load-bearing.

## CLAUDE.md Adherence
✅ TDD workflow: PR description states RED was verified against unfixed HEAD before implementing, with two headline reproduction failures quoted.
✅ Local gate: `./scripts/check-all.sh` reported exit 0, 7463 tests passing (up from 7455), 94.13% branch coverage, per-file gate satisfied for 224 files.
✅ No shortcuts: no `# noqa`, `# type: ignore`, `skip`/`xfail`, or threshold changes anywhere in the diff (verified by reading the full diff — none present).
✅ Docstrings: extensive Google-style docstrings added/updated on every touched public function, consistent with the ≥95% interrogate requirement.
✅ Commit/PR conventions: descriptive, references the closing issue (#1105) and related issues (#974, #1106, #1107) correctly via `Closes #1105` / `Refs`.

## Testing
Insufficient testing would be blocking, but this PR clears that bar comfortably: 4 new reproduction/regression tests in `test_classify_engine.py`, 4 new + 4 migrated tests in `test_privacy_pass.py::TestReassess`, and a rewritten, strictly-stronger pipeline test. Each new test includes a contrast case specifically designed to fail under the most likely wrong implementation (blanket escalate, `!=` gate, "never touch" regression), which is a meaningfully higher bar than typical happy-path coverage.

## Documentation
Docstrings across `classify_engine.py`, `privacy_pass.py`, and `pipeline.py` are thorough, explain *why* (not just *what*), and explicitly cross-reference the issue number and the corresponding logic at the sibling call site — which is exactly the kind of documentation that prevents this class of two-entry-points-disagree bug from recurring a sixth time, per the PR's own framing.



## Verdict: LGTM

The fix is logically sound and provably safe: I traced escalate()'s identity-based gate, verified reassess() is called with a pre-classification baseline computed identically at both call sites (classify_engine.py and pipeline.py), and confirmed the "strictly more restrictive" predicate correctly rejects both matching and weakened verdicts while still catching the hardening case that was the original bug. Tests are unusually rigorous (each new test includes a contrast fragment so the assertion can't pass by accident), the one changed pre-existing test is justified and strictly strengthened, and the PR explicitly rules out the two alternative designs with sound reasoning. Nothing here should block merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5149718059 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1140, opened 2026-08-01
FIX-2: cancel superseded PR CI and review runs (concurrency groups)


Geoffe-Ga:
Parent: #1137

## Problem

Neither `.github/workflows/ci.yml` nor `.github/workflows/code-review.yml`
declares a `concurrency:` group. When a lane pushes again — a sync merge, a
review fix, a CI fix — the **previous run keeps going to completion**: another
~14 minutes of a 3-version matrix plus a `claude-review` invocation, both
validating a commit nobody will ever merge.

Measured over Jul 31 17:44 → Aug 1 04:16: 29 PR-event CI runs and 29 review runs
across 18 branches. PR #1117 alone accounts for 5 of each.

Beyond the wasted runner capacity there is a correctness edge: `pr-ready.sh`
keys CI state off `gh pr checks`' exit code, and a superseded in-flight run
leaves the rollup mixed — a lane reads `pending` on a run whose result is
already irrelevant.

## What to build

Add to both workflows:

```yaml
concurrency:
  group: <workflow>-${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

## Constraints

- **Never cancel on `main`.** `cancel-in-progress` must be false for
  `push`/`schedule`/`workflow_dispatch` — cancelling a main-branch run would
  leave a merged commit unvalidated. Keying the expression off `github.event_name`
  (rather than a blanket `true`) is the whole point.
- Group by `github.ref` so sibling lanes never cancel each other.

## Done when

- Both workflows carry the group; a second push to a PR branch cancels the first
  run and only the first run.
- A push to `main` is never cancelled by a later push to `main`.

Issue #1138, opened 2026-08-01
FIX-1: make pr-ready.sh's freshness gate risk-based instead of unconditional


Geoffe-Ga:
Parent: #1137

## Goal

Stop `main` moving from invalidating every open lane. Keep the stale-green
protection that `behind_by == 0` was added for (#1022).

## Context

`scripts/ralph/pr-ready.sh` → `branch_is_current()` returns true only when the
compare API says `behind_by == 0`. Because `main` here does *not* enforce strict
up-to-date merging, this is a self-imposed gate — and it fires on every merge,
for every lane, regardless of whether the two changesets can interact.

Measured cost in #1137: CI runs per PR 1.00 → 1.61 (max 5), p90 PR span
15 → 104 min. On PR #1117, two of five CI+review rounds were pure
`Merge remote-tracking branch 'origin/main'` commits that changed no code.

## What to build

Replace "0 commits behind" with **"nothing landed on `main` that can invalidate
this PR's green"**. A lane is current unless main's new commits since the merge
base either:

1. touch a **risk surface** — anything that changes how the whole tree is built,
   linted, typed or tested:
   - `*/pyproject.toml`, `*/uv.lock`, `*/requirements*.txt`
   - `.github/workflows/**`
   - `creek-tools/scripts/**`
   - `.pre-commit-config.yaml`
   This is exactly the `ruff`-bump case the current header cites, preserved; **or**
2. **overlap this PR's own changed files** — a genuine semantic interaction.

Otherwise: treat as current and let the lane merge.

## Constraints

- **Fast path unchanged.** `behind_by == 0` still short-circuits on the existing
  single compare call, so a lane that is already current costs exactly what it
  costs today. The second call is paid only by a lane that is actually behind.
- **Fail closed, everywhere.** An API error, a non-integer answer, a malformed
  answer, or a missing merge base must all read as *not* current — the existing
  contract. `behind`'s remedy (a sync) is always safe; a false `ready` is not.
- **Truncation is failure.** The compare API caps `files` at 300. If either side
  comes back at the cap, we cannot prove disjointness — fail closed.
- Keep the `mergeStateStatus == CLEAN` check alongside: it is still the only
  signal for DIRTY / CONFLICTING / BLOCKED / DRAFT / UNKNOWN.

## Done when

- `scripts/ralph/test_pr_ready.sh` covers, at minimum: behind + disjoint
  non-risk files → `ready`; behind + overlapping file → `behind`; behind +
[...]

Issue #1137, opened 2026-08-01
RCA: PR merge latency collapsed ~7x — every merge to main forces whole-fleet revalidation


Geoffe-Ga:
## Summary

PR throughput fell off a cliff. This is the root-cause analysis, with the natural
experiment that pins it.

**It is not CI getting slower, not runner queueing, and not review quality.** A
single merge-gate change turned every merge to `main` into a forced, full
revalidation of every *other* open lane — CI, review, and local gate — and the
cost of that revalidation is superlinear in the number of Ralph lanes.

## The measured gap

Baseline — **Jul 29 00:30 → 11:02**, 11 consecutive merged PRs, single-lane era:

| PR | open→merge |
|---|---|
| #959 #960 #963 #967 #973 #975 #977 #981 #985 #988 #990 | 24, 24, 23, 24, 23, 22, 22, 23, 23, 24, 24 min |

Mean **23.3 min**, σ ≈ 0.8. **Every one of those PRs passed CI exactly once.**

Now — **Jul 31 17:44 → Aug 1 04:16**, two lanes:

| metric | before | after |
|---|---|---|
| CI runs per PR | **1.00** (7/7 PRs) | **1.61** mean, max **5** |
| Review runs per PR | 1.00 | 1.61 mean, max 5 |
| PR CI span, p90 | **15 min** | **104 min** |
| PR CI span, worst | 15 min | **156 min and counting** (#1117) |
| duration of one CI run | 13.9 min | 13.2 min |
| merges/hour | ~1.0 (1 lane) | ~0.88 (**2 lanes**) |

The duration of a single CI run *did not change*. What changed is how many times
each PR has to run it.

Per-lane throughput went **1.0/hr → 0.44/hr**. Two lanes should have produced
~2/hr; they produce 0.88/hr — **4.5x below expectation**, and on the tail
(what a lane actually experiences) **15 min → 104 min is ~7x**.

## Root cause

`scripts/ralph/pr-ready.sh` → `branch_is_current()` requires the compare API to
report `behind_by == 0` before a lane may merge. It landed in **#1022, merged
2026-07-30 22:26** — which is exactly where the table above splits.

Before #1022 the gate keyed off `mergeStateStatus`. This repo's `main` does **not**
enforce strict up-to-date merging, so GitHub reports `CLEAN` for a branch that is
dozens of commits stale — the gate never fired, and no PR ever synced. After
[...]

PR #1117, opened 2026-08-01
feat(api): serve the Adepthood /v1 tracer — authenticated capabilities, honest not-implemented routes, generated OpenAPI

Geoffe-Ga:
## Summary

- Serves the Adepthood `/v1` HTTP application API from a new `creek-tools-api`
  entry point (Starlette, default port **8823**): a **real** `GET /v1/capabilities`
  built from `handshake_tool`, an authenticated `GET /v1/health`, and journal /
  reflection / wheel routes that answer an unmistakable `501` error envelope.
  No MCP *behaviour* changes and no existing test file is edited — the
  pre-existing tests pass unchanged. Three shared modules are touched rather
  than left alone, each deliberately: `server.py` (posture gate extracted, see
  below), `tools/handshake.py` (a side-effect-free `vault_available` probe the
  `/v1` handler needs), and `creek_mcp/audit.py` (one shared `AuditLog` per
  path, from the third review round — see below).
- **Authentication is ASGI middleware above the router**, so an unauthenticated
  caller cannot distinguish an implemented route from an unimplemented one from a
  route that does not exist. The remote INTIMATE cap is enforced at the edge
  through `creek_mcp.policy.admitted_ceiling` (#1073) — one call site, AST-pinned,
  with `is_remote=True` unconditional.
- Ships the hardening with it: body-size cap (enforced on streamed bytes, so a
  chunked request cannot bypass it), per-request timeout, concurrency limit, and a
  structured access log that records the **route template** and never a body, a
  token or an identifier. Plus generated OpenAPI, pinned schema-by-schema against
  the published fixture bundle.

## What an unauthenticated caller sees, versus an authenticated one on a `501`

Captured by running it, not by reading the code:

```
UNAUTH  GET /v1/wheel     UNAUTH  GET /v1/nonsense     UNAUTH  GET /v1/capabilities
HTTP 401                  HTTP 401                     HTTP 401
www-authenticate: Bearer realm="creek"        (all three identical)
vary: X-Creek-Tier-Ceiling
{"code":"unauthenticated","message":"the request carried no valid consumer credential","request_id":"…"}
```

Byte-identical across seven paths × four methods × {no credential, unknown
token} — 56 probes, pinned as one cross-case-identity test. The third review
round caught that "byte-identical" was overstated: the assertion compared header
*names* only. **I widened it to values** rather than narrow the claim
(`v1_api_support.py`'s `header_names` → `header_items`), so all four headers a
`401` carries are now compared verbatim — `content-length: 135`,
[...]

PR #1157, opened 2026-08-01
fix(ralph): stop every merge to main invalidating every other lane

Geoffe-Ga:
PR throughput collapsed after #1022 landed the `behind_by == 0` merge gate.
That gate is a self-imposed strict-up-to-date rule -- `main` does not enforce
one -- and it is quadratic in lane count: merging ONE lane pushes every OTHER
open lane behind, and each of them then pays a sync, a ~14-minute CI round and
(because the sync's new HEAD invalidates its LGTM under the stale-verdict
guard) a full re-review. That window is as long as the interval between
merges, so lanes routinely went stale again WHILE re-proving themselves.

Measured across #1022's merge commit, from the Actions API:

                        before      after
  CI runs per PR        1.00        1.61  (max 5)
  review runs per PR    1.00        1.61  (max 5)
  p90 PR CI span        15 min      104 min
  one CI run            13.9 min    13.2 min
  merges/hour           ~1.0        ~0.88   -- with TWICE the lanes

The duration of a CI run did not move. What moved is how many times each PR
has to run it. Per-lane throughput fell 1.0/hr to 0.44/hr; two lanes produce
less than one lane did. Confirmed on PR #1117: five CI+review rounds over 156
minutes, two of them spent on `Merge origin/main` commits that changed no
code, each triggered by a sibling that touched none of its files.

pr-ready.sh now prints `behind` only when `main` has landed something that can
actually invalidate this branch's green:

  * a cross-cutting change on either side -- lockfile, tool pin, workflow,
    check script, root conftest. This keeps what #1022 was right about. PR #943
    was a ruff bump 22 commits behind, and a bump 17 behind carrying a ruff
    major produced 144 lint errors against the then-current tree.
  * a file both changesets touch, which is a genuine semantic interaction.

OUR side of the risk surface counts as well as theirs, and for the sharper
reason: a branch that changes how the tree is linted or built has proved that
only against the tree at its merge base, so everything main landed since is
code the new tooling never ran over. That is PR #943 exactly -- an existing
test caught this and it is now pinned in both directions.

`behind_by == 0` stays the fast path and still short-circuits on one API call,
so an already-current lane costs exactly what it always did. Only a lane that
is genuinely behind pays for the file listings, and it was otherwise about to
pay for a sync plus a full CI round. Every failure path fails CLOSED as before
-- an errored call, a malformed answer, a bad merge base, or a file list at
GitHub's 300-entry cap (where "disjoint" is unsupportable) all read as behind,
whose remedy is always safe.

What backstops the residual risk is the full CI run on `push: main`, which was
considered for removal as redundant and deliberately kept: it re-proves every
squash-merge, so a stale green that slips through is caught on main rather
than assumed away.

Also adds `concurrency` groups to ci.yml and code-review.yml. A lane pushes
repeatedly to the same PR, and the superseded run was burning a full matrix
and a full Claude review on a commit nobody will merge. `cancel-in-progress`
is keyed off `github.event_name` rather than a blanket true, so a `push` run
on main -- the backstop above -- is never cancelled, and grouping by
`github.ref` keeps sibling lanes from cancelling each other.

Tests: 112 pr-ready assertions (up from 97), covering disjoint-inert to ready,
overlap to behind, each risk-surface path to behind with a look-alike twin
proving the regex is not merely matching everything, both file-listing error
paths, the 300-file cap with a 299-file twin, malformed and empty merge bases,
and laziness assertions that a current lane still makes exactly one call.
Full ralph tooling suite green: 267 shell assertions + 65 recap tests.

Refs #1137, #1138, #1140

### DIFF
--- a/.claude/commands/ralph-tick.md
+++ b/.claude/commands/ralph-tick.md
@@ -175,14 +175,25 @@ Then act on `$STATUS`:
   ever covers a bump that was already current with `main` and already
   green — one nobody touched.
 - **`behind`** (`LGTM` + green but `mergeStateStatus` is `BEHIND`, **or** the
-  far more common case: `mergeStateStatus CLEAN` yet the compare API reports
-  `behind_by > 0`). `CLEAN` means only "no merge conflict" — GitHub computes
-  `BEHIND` solely when the base branch enforces strict/up-to-date status
-  checks, which this repo does not, so `CLEAN` routinely reports on a PR that
-  is dozens of commits stale (measured live: PR #943 reported
-  `MERGEABLE`/`UNSTABLE` while the compare API said `behind_by: 22`; PR #863
-  said `44`). **Do not merge stale — a branch's own green CI says nothing
-  about today's `main`.** Same remedy either way:
+  branch is behind `main` *for a reason that can invalidate its green*).
+  `CLEAN` means only "no merge conflict" — GitHub computes `BEHIND` solely when
+  the base branch enforces strict/up-to-date status checks, which this repo does
+  not, so `CLEAN` routinely reports on a PR that is dozens of commits stale
+  (measured live: PR #943 reported `MERGEABLE`/`UNSTABLE` while the compare API
+  said `behind_by: 22`; PR #863 said `44`). **Do not merge a branch whose green
+  says nothing about today's `main`.**
+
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
⚠️ This PR edits the review workflow — no automated verdict; manual review required.

Issue #1182, opened 2026-08-07
A behind_by==0 lane's green can be stale against time-triggered main breakage (#1159 residual)


Geoffe-Ga:
Follow-up from #1159 (FIX-6). Parent RCA: #1137.

## The accepted residual

#1159 wired `main`'s CI conclusion into the merge loop. A lane that is *behind*
`main` now has to prove the `push: main` backstop is alive before it may use
#1157's risk-based freshness relaxation (`main-not-green`).

A lane with `behind_by == 0` deliberately does **not** pay that check, and
merges even while `main` is red. The justification is an ancestry proof, and it
is airtight for **commit-content** breakage: `behind_by == 0` means `main`'s
HEAD is an ancestor of the branch head, `ci.yml` has no `paths:` filter and runs
the identical matrix on `push` and `pull_request`, and its `actions/checkout@v7`
has no `ref:` override — so the PR's CI ran on `refs/pull/N/merge`, a tree that
already contains whatever broke `main`.

That exception is what keeps the gate deadlock-free: the PR that *fixes* `main`
is itself `behind_by == 0` + green, so gating it on `main` being green would
make the remedy unmergeable and would need a bypass label the design
deliberately does not have.

## What it does not cover

Breakage that is **not a function of the git tree**: a newly published advisory
against an already-pinned dependency, an expired pin, a rotated credential, a
yanked upstream package. A `behind_by == 0` lane whose CI went green *before*
such an event carries the identical pinned dependency, still classifies `ready`,
and merges — even though re-running its CI now would also be red.

This is observed, not hypothetical. #1159's helper caught `main` red on its
first live run for exactly this reason: PYSEC-2026-3552 against `cryptography`
49.0.0 turned `pip-audit` red on a tree nobody had touched (fixed by #1179), so
the two commits bounding the blame range were both innocent.

## Possible remedy (not yet designed)

Bound the *age* of the green a `behind_by == 0` lane merges on: if the lane's
newest CI run predates the newest red `ci.yml` run on `main`, its green is stale
with respect to a time-triggered break and the lane should re-prove rather than
merge. `main-health.sh` already fetches the run list this would need, so the
data is in hand; the cost is one comparison plus a re-run, and the design
question is whether that trade is worth it given how rare the class is.

Explicitly a **question**, not a decision — the current behaviour is correct and
documented as an accepted residual in `scripts/ralph/pr-ready.sh`'s
[...]

PR #1179, opened 2026-08-07
fix(deps): bump cryptography to 50.0.0 for PYSEC-2026-3552

Geoffe-Ga:
## Summary

**Stop-the-line P0.** `cryptography` 49.0.0 carries **PYSEC-2026-3552** (`CVE-2026-69247`, `GHSA-g6cj-pr64-35w5`) — PKCS#7 `EnvelopedData` decryption reported the outcome of unwrapping a `RecipientInfo`'s `encryptedKey` in several distinguishable ways (one disclosing the recovered length), and the same distinction was observable by timing, giving an attacker who supplies `EnvelopedData` a Bleichenbacher oracle against the content-encryption key. OSV range: **introduced 44.0.0, fixed 50.0.0**.

The advisory published 2026-08-04, three days after `main` was last green. No code changed — the advisory database did. Both quality gates audit a pip-audit surface carrying the pin, so all six CI jobs went red on `main` and **every open PR was blocked behind it** (including #1117).

### Why a bare relock was not enough

`uv lock` was run first, unmodified, in both projects. It was **empirically a no-op — zero diff in each**. uv preserves an already-locked version while it still satisfies the declared floor, and both floors (`>=48.0.0` / `>=48.0.1`) admitted 49.0.0. So the *declared floor* is what had to move.

It moves to `>=50.0.0` in three places:

| File | Surface | Before → after |
|---|---|---|
| `creek-tools/pyproject.toml:38` | `[project].dependencies` | `>=48.0.0` → `>=50.0.0` |
| `creek-tools/pyproject.toml:163` | `[tool.uv].constraint-dependencies` | `>=48.0.0` → `>=50.0.0` |
| `crawdad/pyproject.toml:92` | `[tool.uv].constraint-dependencies` | `>=48.0.1` → `>=50.0.0` |

creek-tools gets both surfaces because `cryptography` is a *direct* import there (`creek.confidential.keyvault`); the mirrored constraint binds the transitive edge (`mcp → pyjwt[crypto]`). In crawdad it stays transitive-only, so the constraint is the right and only home.

Both `uv.lock`s were then **regenerated, never hand-edited**: `cryptography 49.0.0 → 50.0.0` in each. The lock diffs are surgical — exactly one `version =` line changed per lock, no collateral resolution churn.

### Live-registry verification (per the `cve-remediation` skill)

[...]

PR #1183, opened 2026-08-07
fix(ralph): make the push:main backstop a check instead of a premise

### LINKED ISSUE #1159: FIX-6: nothing reacts to a red main, and FIX-1's safety argument now depends on it
Parent: #1137 · Makes #1138 safe · Substitutes for #1139

## Problem

FIX-1 (#1138) relaxed the pre-merge freshness gate: a lane behind `main` merges
anyway when the two changesets provably cannot interact. That trade is only
sound because **every squash-merge re-runs the full CI matrix on `main`** — so a
stale green that slips through is *detected* rather than assumed away.

That detector currently reports to nobody.

Grepped the whole loop — `ralph-tick.md`, `fleet.sh`, `pick-next.sh`,
`pr-ready.sh`, `watch-pr.sh`: **no step reads the status of the `push: main` CI
run.** Every gate is per-PR. If `main` goes red, the loop keeps picking issues,
lanes branch off the broken commit, and every one of them inherits the failure —
which reads as "CI is flaky today" at exactly the moment it is not.

This was a latent gap before #1138 (a rare stale green could already break main
via a race). #1138 makes it load-bearing, so it should land close behind.

## What to build

A `main` health gate in the Ralph tick, before Step 1:

1. Read the conclusion of the most recent completed `ci.yml` run for
   `branch: main` (`gh run list --workflow ci.yml --branch main --limit 1
   --json conclusion,headSha,databaseId`).
2. **Red → the loop's only job is to green it.** Do not merge any lane (every
   merge compounds the breakage and buries the culprit), do not refill slots.
   Dispatch one worker on the `ci-debugging` flow against the failing commit,
   and report the failing check and the commit that introduced it.
3. **Green → proceed exactly as today.**
4. **Still running → proceed.** `main` CI lags the merge by ~14 min and blocking
   on it would reintroduce the serialization #1138 just removed. This gate is a
   circuit breaker, not a barrier.

## Constraints

- **Fail OPEN, unlike `pr-ready.sh`.** This one is the opposite polarity from
  every other gate in the loop: an unreadable answer must not halt the fleet,
  because the cost of a false halt (the loop stops entirely) far exceeds the
  cost of a false proceed (one more PR onto a possibly-red main, which the next
  tick catches). Say so explicitly in the header — the file next door argues the
  opposite and the two must not be read as inconsistent by accident.
- Attribute the failure to a commit, not just a run: with several merges close
  together the newest red run is not necessarily the one that broke it.
- Do not gate on workflows other than `ci.yml` — a red `scan-*` or a flaky
  `graph-update` must not stop merges.

## Done when

- A new `scripts/ralph/main-health.sh` prints one token (`green` / `red` /
  `pending` / `unknown`), tested the way `pr-ready.sh` is — a stubbed `gh`, every
  failure path pinned, and a laziness/fail-open assertion.
- `ralph-tick.md` Step 0 routes on it.
- Registered in `.github/workflows/ralph-recap-tests.yml` alongside the other
  shell suites.

Geoffe-Ga:
## Summary

- **Wires the missing half of #1157's trade.** PR #1157 replaced the strict pre-merge freshness gate with a risk-based one (`RISK_SURFACE_RE` + merge-base disjointness), and its entire safety argument is one sentence in `pr-ready.sh`'s header: *"what backstops the residual risk is the full CI run on `push: main`."* No step in the loop had ever read that run's conclusion, so a lane could merge against a `main` whose own CI was red and nothing noticed. New `scripts/ralph/main-health.sh` makes that premise checkable.
- **It is a PRECONDITION of the relaxation, not a new gate on merging.** Only a lane that is *behind* — i.e. about to invoke the relaxation — pays the probe, and anything but `green` holds it as the new `main-not-green` token. `RISK_SURFACE_RE`, `main_changes_are_inert` and `compare_files` are untouched: the relaxation is **suspended** while its backstop is down, never weakened, so the ~7x regression #1137 measured does not come back.
- **Deadlock-free by construction, with no bypass label.** A lane with `behind_by == 0` never invokes the relaxation, so it never asks, and merges even while `main` is red — which is exactly the shape of the PR that *fixes* `main`. Documented residual (a green that is stale against time-triggered breakage) tracked in #1182.

## Design decisions, and why

**Where the check lives — both, with a division of labour.** `pr-ready.sh` owns the per-PR half because *"merge a lane only when `pr-ready.sh` prints `ready`"* is the loop's only **mechanically enforced** rule; everything in `ralph-tick.md` is prose an LLM reads. #1159 exists precisely because a safety property lived only in prose, so encoding the fix only in prose would repeat the reported failure. `ralph-tick.md` Step 0b keeps the genuinely global decisions that have no PR to hang on: stop refilling slots, dispatch one `ci-debugging` worker, report the attribution.

**Why a per-PR token for a global condition.** The token does not encode *"main is red"* — it encodes a per-PR verdict: *"this lane cannot merge because the specific safety property it depends on is unavailable."* That is demonstrably per-PR: at the same instant a `behind_by == 0` lane gets `ready` and a behind lane gets `main-not-green`.

**Fail closed.** `red`, `pending`, `unknown`, an empty answer, a helper that will not run — all hold the lane. No failure path can produce `green`. Because the gate is scoped to the relaxation, a false hold costs only the #1157 optimization for one wake, not a fleet halt: workers keep building, PRs keep opening, refill continues. The blast radius is small enough that fail-closed is cheap rather than grudging.

**Circuit breaker, not barrier.** It keys off the newest **conclusive completed** run, not the newest run. `main` CI lags each merge by ~14 min, so the newest run is nearly always in flight; blocking on it would reintroduce the serialization #1138 removed. Cancelled/skipped/neutral/stale conclusions are *not evidence* and the walk continues past them.

**How a fix-`main` PR gets through.** `behind_by == 0` means `main`'s HEAD is an ancestor of the branch head. `ci.yml` has no `paths:` filter, runs the identical matrix on `push` and `pull_request`, and its `actions/checkout@v7` has no `ref:` override — so the PR's CI ran on `refs/pull/N/merge`, a tree already containing whatever broke `main`, and `gh pr checks` is per-HEAD-commit. Its green is therefore proof (for commit-content breakage) that the breakage is absent or fixed. Making that lane consult the helper too was considered and rejected: the remedy PR *is* that shape, so gating it would make the fix unmergeable and require the bypass label this design deliberately does not have.

**Interaction with `RISK_SURFACE_RE`.** Not orthogonal — a **precondition**. The probe sits *before* `main_changes_are_inert`, so a risk-surface lane facing a red `main` prints `main-not-green` rather than `behind`. That ordering matters: `behind`'s remedy is `fleet.sh sync`, which would import the breakage, burn a ~14-min CI round, and turn the lane's own CI red — then be misclassified as that lane's `ci-failed`.

**`watch-pr.sh`.** `main-not-green` joins `IN_FLIGHT_TOKENS`. Without it the watcher exits on its first poll, the orchestrator relaunches it (the pidfile is gone), it exits again, and the fleet busy-wakes for as long as `main` stays red — worse than the bug being fixed.

## Test plan

Gate 1 RED was observed before the fix, not asserted: `test_pr_ready.sh` printed

```
FAIL - behind + inert files but main CI RED → main-not-green (expected 'main-not-green', got 'ready')
```

— a lane classified **mergeable while `main`'s own CI was red**. Also RED beforehand: `test_main_health.sh` 17/97, `test_watch_pr.sh` 34/37, `test_exec_bits.sh` 14/15.

All green now (414 assertions, 0 failures):

| suite | result |
|---|---|
| `test_main_health.sh` | 109 passed |
| `test_pr_ready.sh` | 142 passed |
| `test_watch_pr.sh` | 37 passed |
| `test_exec_bits.sh` | 17 passed |
| `test_fleet.sh` | 85 passed |
| `test_pick_next.sh` | 24 passed |

- `shellcheck --severity=warning scripts/ralph/*.sh` — clean.
- `cd creek-tools && ./scripts/check-all.sh` — **12/12 passed, exit 0**. `ruff check --no-cache .` clean (#1119).
- New suite registered in `.github/workflows/ralph-recap-tests.yml`.

**Anti-masking.** The `gh` stub defaults its `run list` arm to `green` so the pre-existing behind-but-inert cases keep testing #1157's disjointness. Three assertions together make a deleted probe impossible to miss: the answer is *consumed* (`MAIN_HEALTH=red` → `main-not-green`), the call is really *made* (sentinel present), and it is *lazy* (sentinel absent on a `behind_by == 0` lane). `test_pr_ready.sh` stubs only `gh` and runs the **real** `main-health.sh`, so token drift between the two scripts is caught.

**Verified live, not just stubbed.** On its first real run the helper caught `main` genuinely red at `da74201` — every job in the matrix failing — and returned `red` while a newer run was still in flight (circuit breaker working), with blame range `a5809aa..da74201`. Root cause was PYSEC-2026-3552 against `cryptography` 49.0.0, fixed on `main` by #1179; the helper returns `green` again now. That live run also exposed and corrected an overclaim: a time-triggered break has *no* culprit commit, so the attribution now reports the range as a window rather than an accusation.

Hand-traced classification (verified by execution against a stubbed `gh`): green-only, red→green, red→red→green, red-with-no-green, all-cancelled, cancelled-above-green, in-flight-above-red, in-flight-above-green, in-flight-only, malformed-above-green. The blame range is never backwards, never empty-sided, and never fabricated.

## Notes

- Gate 2.5 (`code-review-orchestrator`) returned one blocker: the `behind_by == 0` exception was described as *"positive proof"* and *"cannot deadlock, by construction"* without scoping. That was a genuine overclaim — the ancestry argument is airtight for commit-content breakage but says nothing about time-triggered breakage, as this PR's own live run demonstrated. Fixed in all four locations by scoping the claim and naming the residual as knowingly accepted, plus the rejected alternative and why. Behaviour unchanged; residual tracked in **#1182**.
- Also from review: a malformed run-list line encountered *below* an already-proven `red` no longer discards that verdict (it would have silently disabled Step 0b's remediation dispatch, which fires only on the literal `red`). A malformed line still never *creates* or *upgrades* a verdict — above an undecided walk it still yields `unknown`, never `green`.
- `.claude/skills/await-claude-review/SKILL.md` documented the watcher's in-flight set, which this change alters; updated so it does not become false.
- This PR touches `.github/workflows/ralph-recap-tests.yml`, which matches `RISK_SURFACE_RE` — so its own lane will read `behind` whenever `main` moves and must sync before merging. Expected, not a bug.

Closes #1159
Refs #1137


Geoffe-Ga:
## Summary
Adds `scripts/ralph/main-health.sh`, a circuit-breaker that reads the newest *conclusive* `push: main` run of `ci.yml` and reports `green`/`red`/`pending`/`unknown`. `pr-ready.sh` now consults it as a **precondition** of the existing #1157 behind-but-inert relaxation (never a blanket merge gate), printing a new `main-not-green` token that `watch-pr.sh` treats as in-flight and `ralph-tick.md` routes on (stop refilling, dispatch one `ci-debugging` worker, still let `behind_by == 0` lanes merge so the fix for `main` itself is never deadlocked). Ships with a new 585-line offline test suite plus updates to three existing suites and all the process docs that describe the loop.

## Strengths
- The core walk in `main-health.sh` is genuinely subtle (circuit breaker vs. barrier, blame-range floor discovery, malformed-line handling that must neither fabricate nor discard a proven verdict) and I traced it by hand against the new tests — it holds up, including the trickier cases (malformed line *below* a proven `red` keeps the red; malformed *above* an undecided walk still yields `unknown`; in-flight above a decided `red` still reports `red`, not `pending`).
- Fail-closed discipline is consistent everywhere it matters: missing/non-executable helper, non-zero `gh`, empty output, unparseable lines, and every inconclusive-only window all resolve to "hold the lane," never "merge." `pr-ready.sh` invokes the helper directly (never via `bash`) specifically so a dropped exec bit reads as "could not ask," not as an answer — and `test_pr_ready.sh` has a case that plants a helper which *would* answer `green` if it ran, proving that distinction is real.
- The `behind_by == 0` exception (a lane that's current merges even while `main` is red) is well-reasoned and explicitly scoped — the PR's own Gate-2.5 history shows an earlier overclaim ("cannot deadlock, by construction") was caught and narrowed to commit-content breakage only, with the time/environment-triggered residual (the observed `PYSEC-2026-3552` case) named as an accepted gap rather than swept under the rug.
- Anti-masking test design: `MAIN_HEALTH` stub defaults to `green` specifically so a deleted probe wouldn't silently pass the whole suite, and the laziness assertions (sentinel files) prove the extra `gh` call is paid only by lanes that actually need it.
- Docs (`ralph-tick.md`, `FLEET.md`, `await-claude-review/SKILL.md`) were all updated in the same PR to keep the human/agent-facing prose in sync with the new token, including the busy-wake failure mode `watch-pr.sh`'s `IN_FLIGHT_TOKENS` addition prevents.

## Security Concerns
None identified — 🟢 no user input, no injection surface (all values flow from `gh` JSON via `--jq`, joined/split on `|` with explicit field-count validation), no secrets touched.

## Problems
None found blocking merge.

## Code Quality
- Minor cosmetic overlap: the token `TOKEN_PENDING="pending"` and the GitHub run status `pending` (inside `IN_FLIGHT_STATUSES`) share a literal string but are different variables in different domains — not a bug, just a name a future reader might do a double-take on.
- `RUN_WINDOW=20` is a magic-but-documented constant; if the blame-range floor genuinely falls outside that window the script correctly says "unattributable" rather than guessing, so this is a documented tradeoff, not a gap.

## CLAUDE.md Adherence
✅ Single commit uses Conventional Commits format (`fix(ralph): ...`). ✅ No `creek-tools/` Python files touched, so the strict coverage/mypy/docstring gates don't apply here; the PR description confirms `check-all.sh` was still run for regression safety. ✅ `shellcheck --severity=warning` reported clean per the PR description, and the new/changed scripts follow the existing repo conventions (dirname-based sibling resolution, `set -euo pipefail`, fail-closed token contracts) already established by `pr-ready.sh`/`watch-pr.sh`.

## Testing
Strong. New `test_main_health.sh` (585 lines) covers conclusion classification, the anti-serialization/circuit-breaker pin, the full fail-closed sweep, cost (exactly one `gh` call per path), stdout purity, blame-range attribution (including the no-green-floor and malformed-below-red edge cases), and a cross-file coupling check against `ci.yml`'s `push:`/`cancel-in-progress` config. `test_pr_ready.sh`, `test_watch_pr.sh`, and `test_exec_bits.sh` were extended with the new token's laziness, ordering, and busy-wake-regression cases. Both live CI runs of "Recap + fleet + watch tooling tests" for this PR are green, so this isn't just self-reported — I confirmed it via `gh pr view --json statusCheckRollup`.

## Documentation
Thorough — `main-health.sh`'s own header, `pr-ready.sh`'s "WHY IT IS A PRECONDITION, NOT A GATE" block, `FLEET.md`, `ralph-tick.md` Step 0b, and the `await-claude-review` skill were all updated coherently, and the PR description itself documents the Gate 2.5 review round that tightened the deadlock-freedom claim's scope. No gaps found between docs and implementation.



## Verdict: LGTM

The change is correct on a full manual trace of main-health.sh's run-classification walk (fail-closed sweep, circuit-breaker vs. barrier distinction, malformed-line handling, blame-range attribution), the pr-ready.sh/watch-pr.sh integration is consistent with the documented contract, the new/updated test suites include deliberate anti-masking assertions and already pass in CI (both "Recap + fleet + watch tooling tests" runs are green), and all touched docs (ralph-tick.md, FLEET.md, SKILL.md) were updated to match. No bugs, security issues, or missing tests found; nothing here rises above a nit.

Geoffe-Ga:
**CI**: 6/9 Green
**VERDICT**: LGTM
**Action**: pull comment 5213317728 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 12/12 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Adds `scripts/ralph/main-health.sh`, a circuit breaker that checks whether `main`'s own CI (`push: main` on `ci.yml`) is green before a Ralph lane invokes PR #1157's "merge while behind" relaxation. `pr-ready.sh` now consults it as a precondition (not a new merge gate) and prints a new `main-not-green` token; `watch-pr.sh` and `ralph-tick.md` are updated to treat that token as an in-flight wait state, with a documented, tested exception so a lane that is already current (`behind_by == 0`) — exactly the shape of the PR that would fix `main` — is never blocked and cannot deadlock the loop.

## Strengths
- **Design is airtight and self-aware of its own limits.** The header comments and `pr-ready.sh`'s "WHY IT IS A PRECONDITION, NOT A GATE" block explicitly scope the `behind_by == 0` exception to tree-borne breakage and name the accepted residual (time/environment-triggered breaks, e.g. the PYSEC-2026-3552 advisory that motivated this) rather than overclaiming safety. This overclaim was caught and fixed during the PR's own Gate 2.5 review per the PR notes — good process evidence.
- **Fail-closed by construction.** Every unreadable/ambiguous answer (empty list, non-zero `gh`, malformed line, unrecognized status, missing/non-executable helper) resolves to something other than `green`, never the reverse. The malformed-line handling is subtle and correct: it stops the walk but never *creates* or *upgrades* a verdict, only preserves an already-proven `red`.
- **Circuit breaker vs. barrier distinction is correctly implemented and tested.** Keying off the newest *conclusive* run (not the newest run) avoids reintroducing the ~14-minute serialization #1138/#1157 removed, and `IN_FLIGHT_OVER_RED` proves the breaker still trips even when a newer run is in flight.
- **Test coverage is exceptional**, including cost assertions (exactly one `gh` call per path), stdout-purity assertions, an explicit three-part "anti-masking proof" (answer consumed / call made / call lazy) to prevent a silently-deleted probe from passing existing tests, real-jq exercised against the production `--jq` expression (not just a scalar stub), and cross-file coupling checks that `ci.yml` keeps its `push: main` trigger and doesn't unconditionally cancel in-progress runs.
- **Idempotent remediation dispatch** via a GitHub-issue-keyed guard (`main is red at <sha7>`) rather than local state, correctly reasoned to survive orchestrator restarts.
- Blame-range attribution correctly avoids fabricating a floor when no green run is in the window, and correctly avoids narrowing to "just the newest red run" when several commits landed close together.

## Security Concerns
None. 🟢 No user input is trusted unsafely; `gh` output is parsed by strict field-count splitting (matching the existing `pr-ready.sh` convention) and any shape mismatch fails closed. The helper is invoked directly (never via `bash "$path"`) specifically so a dropped exec bit reads as "could not ask" rather than executing a stray file — a good defensive touch.

## Problems
None blocking.

## Code Quality
- Non-blocking style nit: `main-health.sh`'s header comment is very long (~90 lines) relative to the script body. It's justified here given the subtlety of the circuit-breaker/precondition distinction and the fact that a future maintainer could easily "simplify" the malformed-line handling into a bug (as the comments themselves flag), but it's worth knowing this is a dense read.
- The `RUN_WINDOW=20` constant is a reasonable default but is not itself parameterized or tested against extremely bursty merge scenarios where 20 runs might not reach a green floor — acceptable since the code already handles "no green found" honestly rather than guessing.

## CLAUDE.md Adherence
✅ New shell entry point is `chmod +x` (mode 100755) and pinned by name in `test_exec_bits.sh`. ✅ Tests are run via the repo's existing suite pattern and wired into `.github/workflows/ralph-recap-tests.yml` explicitly (not globbed), matching the repo's stated philosophy that "a suite that is written but never wired in is not a gate, it is a file." ✅ No Python/`creek-tools` changes in this PR, so the Python-specific coverage/mypy/ruff gates don't apply here — PR description confirms `check-all.sh` was still run and is green. ✅ Commit messages and PR body follow Conventional Commits and document design decisions with rationale, consistent with repo norms.

## Testing
✅ Excellent — 414 assertions across `test_main_health.sh` (new, 109 assertions), `test_pr_ready.sh` (+142 new/changed), `test_watch_pr.sh`, `test_exec_bits.sh`, plus a live verification (caught real `main` red at `da74201`, correctly attributed and correctly recovered). RED-before-fix state was captured and documented, satisfying TDD Gate 1. No gaps identified.

## Documentation
✅ Thorough — `ralph-tick.md` Step 0b, `FLEET.md`'s token table and test descriptions, and inline script headers are all updated consistently and cross-reference issue #1159/#1157/#1137/#1182 for context. `await-claude-review/SKILL.md` was updated so its in-flight-token description doesn't go stale, which is easy to miss in a change like this.



## Verdict: LGTM

The change is well-scoped, exhaustively tested (414 assertions across 6 suites, including anti-masking, laziness, fail-closed, cross-file coupling, and real-jq checks), fails closed on every error path, and the design tradeoffs (circuit breaker vs. barrier, precondition vs. gate, the behind_by==0 deadlock exception) are documented and directly verified by tests rather than just asserted in prose. No security, correctness, or CLAUDE.md-compliance issues found; only minor stylistic nits that don't block merge.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5213490052 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

Issue #1191, opened 2026-08-07
fix(scripts): complexity and docstring gates never scan tests/, so check-all.sh disagrees with CLAUDE.md


Geoffe-Ga:
## Problem

`./scripts/check-all.sh` never applies the complexity or docstring gates to `tests/`.
Both scope to `creek/` (complexity) and `creek/` (interrogate) only:

- `scripts/complexity.sh` → radon/xenon over `creek/`
- `scripts/lint-interrogate.sh` / CI's "Check docstring coverage" → `interrogate ... creek/`

CI's `complexity-analysis` job has the same scope.

Consequence, observed concretely: a brand-new test function landed at cyclomatic
complexity **C (15)** during #1119 and `./scripts/check-all.sh` exited **0** on it. It
was caught only by a reviewer running `radon`/`xenon` by hand against the file:

```
$ radon cc -s -n C tests/test_ruff_gate_parity.py
    F 404:0 test_ci_and_local_ruff_share_the_same_rule_configuration - C (15)

$ xenon --max-absolute B tests/test_ruff_gate_parity.py
ERROR:xenon:block "...test_ci_and_local_ruff_share_the_same_rule_configuration" has a rank of C
```

(It was fixed within #1119, but by hand, not by the gate.)

## Why it matters

`creek-tools/CLAUDE.md` §1.5 and §6.1 state the thresholds as blanket, non-negotiable
project standards — "Cyclomatic Complexity: ≤10 per function", "Docstring Coverage:
≥95%" — with no test-file carve-out. The gate silently disagrees with the documented
policy. Either the policy should be scoped explicitly to production code, or the gate
should enforce what the policy says. Right now a reader of CLAUDE.md is misled about
what is actually enforced.

Test code is also where complexity hurts most quietly: a convoluted test is harder to
tell apart from a *wrong* test.

## Precedent

Issue #925 widened mypy, bandit and pylint from `creek/` to `creek/ creek_mcp/`, and
`creek-tools/tests/test_scanner_coverage.py` exists specifically to keep those
invocations widened. This issue is the same shape, one directory further out, and
should get the same treatment: widen the invocation AND add the assertion to
`test_scanner_coverage.py` that keeps it widened.

## Suggested scope

- [ ] Decide and record: are the thresholds project-wide or production-only? Update
      `creek-tools/CLAUDE.md` to match whichever is chosen, so doc and gate agree.
[...]

Issue #1190, opened 2026-08-07
docs(creek-tools): CLAUDE.md is truncated mid-code-fence — sections 7-10 and all appendices are missing


Geoffe-Ga:
## Problem

`creek-tools/CLAUDE.md` is truncated. On `origin/main` the file is 461 lines and ends
**mid-code-fence**, inside an unclosed ` ```python ` block under §6.2 "Forbidden
Patterns", item 3:

```
458  3. **TODO/FIXME Comments**
459     ```python
460     # ❌ FORBIDDEN
461     # TODO: optimize
        <EOF>
```

Everything after that is gone. The file's own table of contents (lines 6–17) promises
six more sections that do not exist on disk:

- §7. Development Workflow
- §8. Testing Strategy
- §9. Tool Usage & Code Standards
- §10. Common Pitfalls & Troubleshooting
- Appendix A: AI Subagent Guidelines
- Appendix B: Key Files
- Appendix C: External References

## Why it matters

This is the authoritative quality-standards document for the subproject, and it is the
file `CLAUDE.md` at the repo root explicitly tells every agent to read first:

> **Always read `creek-tools/CLAUDE.md` before working in that directory.**

Three concrete consequences:

1. **Dead internal anchors.** §1.3 links to `[10.1 No Shortcuts Policy](#101-no-shortcuts-policy)`
   and §1.7 links to `[10. Common Pitfalls & Troubleshooting](#10-common-pitfalls--troubleshooting)`.
   Both targets are absent.
2. **§1.1 points at §9.1 for the complete tool-invocation list** ("See [9.1 Tool
   Invocation Patterns](#91-tool-invocation-patterns) for complete list") — also absent.
3. **The unclosed code fence** means every Markdown renderer treats the rest of the
   document as code, and it makes the file hostile to append-only edits: a
   documentation pass for #1119 was blocked on exactly this, because the target section
   could not be observed and matching its conventions was impossible.

## Provenance

`git log -- creek-tools/CLAUDE.md` shows the most recent touch is 8e69b04
(`fix(quality): bring creek_mcp under mypy, bandit and pylint (#925) (#967)`). Whether
the truncation originated there or earlier needs a `git log -p` bisect over the file.

## Acceptance criteria

- [ ] The truncation point is bisected and the cause identified
- [ ] §§7–10 and Appendices A–C are restored (from history if recoverable, rewritten if
      not) so the TOC matches the body
[...]

Issue #1189, opened 2026-08-07
refactor(scripts): lint.sh's EXIT_CODE=$? error handling is unreachable under set -e


Geoffe-Ga:
## Problem

`creek-tools/scripts/lint.sh` captures `EXIT_CODE=$?` at lines 74 and 80 and branches on
it at line 83:

```bash
    ruff check .
    EXIT_CODE=$?
fi

if [ $EXIT_CODE -eq 0 ]; then
    echo "✓ Linting checks passed"
    exit 0
else
    echo "✗ Linting checks failed" >&2
    exit 1
fi
```

The script runs under `set -euo pipefail` (line 5), so a non-zero `ruff check`
terminates the script *before* the assignment is reached. `EXIT_CODE` is therefore
always `0` where it is read, and the `✗ Linting checks failed` branch is dead code.

## Impact

The gate itself is **correct** — `set -e` propagates ruff's exit status, so a failing
lint still fails the script and `check-all.sh`. This is cosmetic. But it is misleading
error handling in the middle of one of the repo's most-read scripts, and it invites a
future reader to "fix" it in a way that actually breaks the gate (e.g. by wrapping the
ruff call in a way that swallows the status).

## Proposed fix

Restructure to `if ruff check … ; then … else … fi`, or drop the dead branch entirely
and let `set -e` do the work with a `trap`-based message.

## Note

Deliberately left untouched by #1119 to keep that P0's diff to pure argv appends.

For the record, the related `$?`-after-a-pipe defect class (where `cmd | tail` makes
`$?` report the pipe's last stage rather than the tool) does **not** appear anywhere in
`creek-tools/scripts/*.sh` or `scripts/ralph/*.sh` — the ralph scripts all use the safe
`|| rc=$?` idiom. Verified during #1119.

Follow-up from #1119.

Issue #1188, opened 2026-08-07
refactor(ci): run Ruff through ./scripts/lint.sh and ./scripts/format.sh like every other gate (CI-004)


Geoffe-Ga:
## Problem

`.github/workflows/ci.yml` invokes Ruff **inline**:

```yaml
- name: Run Ruff linting
  run: ruff check . --output-format=github --exit-non-zero-on-fix

- name: Run Ruff format check
  run: ruff format --check .
```

while it calls the local script for every other gate — `./scripts/pylint.sh`,
`./scripts/typecheck.sh`, `./scripts/lint-refurb.sh`, `./scripts/lint-tryceratops.sh`.
The workflow's own comments state the principle:

> Same script the local check-all.sh runs, so a developer who passes locally cannot
> fail this gate (CI-004).

Ruff is the lone exception to CI-004 — and Ruff is precisely where local/CI drift bit
us in #1119.

## Proposed change

Switch the two steps to `./scripts/lint.sh --check` and `./scripts/format.sh --check`.

- GitHub PR annotations are preserved by setting `RUFF_OUTPUT_FORMAT: github` as a step
  `env:`, so no script change is needed.
- `--exit-non-zero-on-fix` is a no-op on a non-`--fix` check path, so nothing is lost.

## Status: cleanup, not an open hole

#1119 landed `creek-tools/tests/test_ruff_gate_parity.py`, which already pins the
invariant that would otherwise be at risk: both sides take their rule set from
`creek-tools/pyproject.toml`, CI never restores a `.ruff_cache`, and no gate-path ruff
line auto-fixes. So the drift is currently *detected*; this issue removes the
possibility of it rather than catching it after the fact.

**That parity file asserts the current inline shape (it requires exactly one inline CI
`ruff check` line and one `ruff format --check` line) and MUST be updated as part of
this change.**

## Acceptance criteria

- [ ] CI runs Ruff through the same scripts `check-all.sh` runs
- [ ] PR annotations still appear on ruff findings
- [ ] `creek-tools/tests/test_ruff_gate_parity.py` is updated to pin the new shape, not
      deleted or weakened
- [ ] No threshold, rule selection or gate is weakened

Follow-up from #1119.

PR #1192, opened 2026-08-07
fix(scripts): make the ruff gate cache-free so Gate 2 cannot pass a tree CI fails

### LINKED ISSUE #1119: fix(scripts): a stale .ruff_cache lets check-all.sh pass while CI fails on the same tree

Geoffe-Ga:
## Summary

- **The local quality gate could pass on code CI rejects.** `./scripts/check-all.sh` consulted `.ruff_cache`; CI never has one. Root cause, verified by experiment: **ruff's per-file cache key is mtime-only** — not size, not a content hash. Any content change leaving the mtime unchanged or restored (`cp -p`, `rsync -t`, `tar -x`, a checkout of an older tree, an mtime-preserving editor) leaves a stale verdict standing. This cost a real CI failure on PR #1117 plus a `ci-debugging` lane to diagnose.
- **Fix:** `--no-cache` on all six ruff invocations in `scripts/lint.sh` and `scripts/format.sh`, plus both `astral-sh/ruff-pre-commit` hooks. **Fix paths as well as check paths** — I verified a `--fix` run answering from a poisoned cache prints "All checks passed!" and *leaves the file unfixed*, which is worse than a false green because the developer then commits it. The pre-commit hooks are included because they are one of the things that *seeds* the cache the gate later trusts.
- **Two regression suites**, one behavioural and one static — see the test plan.

### Design decisions, with the measurements behind them

**Performance.** Measured on this tree (477 tracked `.py` files, median of 3, `/usr/bin/time -p`):

| invocation | cached | `--no-cache` |
|---|---|---|
| `ruff check .` | 0.01s | 0.04s |
| `ruff format --check .` | 0.01s | 0.03s |

**~0.05s total.** There is no trade-off to make, which is why check *and* fix paths are covered uniformly rather than split — "ruff always reads the real bytes" is an invariant a test can pin; "check paths are cache-free, fix paths aren't" is a rule someone gets wrong later.

**Hardcoded flag, no env override.** Ruff honours `RUFF_NO_CACHE` natively, but routing the gate through an env var builds an off switch for a gate. The flag is greppable, visible under `set -x`, and shlex-parseable by the parity test.

**CI unification deferred.** `ci.yml` runs ruff inline while calling `./scripts/*.sh` for every other gate — ruff is the lone exception to the workflow's own stated CI-004 principle, and ruff is exactly where the drift bit us. Changing what runs in CI on a P0 about restoring trust in the gate is risk visible only after push, so `test_ruff_gate_parity.py` pins the invariant instead and the switch is filed as #1188.

### Other tools checked for the same hazard

I probed each empirically rather than assuming:

| tool | vulnerable? | evidence |
|---|---|---|
| **ruff** | **yes** | key is mtime-only — even a *different-length* rewrite at a restored mtime stayed cached-clean |
| **mypy** | **yes** | same mtime + same size → `Success: no issues found` on a file cold mypy rejects with `[return-value]`. **Not free to close: 1.22s incremental vs 17.82s `--no-incremental`.** Filed #1186 |
| **`__pycache__`** | **yes** | `assert 1 == 1` → `assert 1 == 2` at same mtime+size still **passed** — stale bytecode. Filed #1187 |
| `.pytest_cache` | no | stores `lastfailed`/`nodeids`, not verdicts |
| refurb, vulture, interrogate | no | no cache options |
| pylint, bandit | no | no persistent result cache in the gate invocation |

**`$?`-after-pipe defect class: not present.** I grepped `creek-tools/scripts/*.sh` and `scripts/ralph/*.sh` — no instance of a tool being piped and then `$?` checked. The ralph scripts all use the safe `|| rc=$?` idiom. The one adjacent smell is `lint.sh`'s `EXIT_CODE=$?`, which is *unreachable* under `set -euo pipefail` rather than wrong (the gate still fails correctly via `set -e`); filed as #1189 and deliberately left untouched to keep this P0's diff to pure argv appends.

## Test plan

**RED was real and non-vacuous.** Against unfixed HEAD, the *real* gate scripts cleared trees CI would reject:

```
scripts/lint.sh --check   -> exit 0   "✓ Linting checks passed"        (unused import present)
scripts/format.sh --check -> exit 0   "✓ Code formatting check passed" (I001 unsorted / misformatted)
ruff check --no-cache .   -> exit 1   on the identical bytes
```
5 failed / 16 passed before the fix; 22 passed after.

- **`tests/test_ruff_cache_poisoning.py`** (behavioural, 3 tests) — builds a temp tree, seeds a real `.ruff_cache` with argv **mirroring the script's own**, rewrites the source with a violation at a restored mtime (`os.utime(ns=...)`), symlinks the **real** gate script into `<tmp>/scripts/` so `PROJECT_ROOT` retargets, then asserts the script fails. `RUFF_CACHE_DIR` is pinned inside `tmp_path`, and ambient `RUFF_NO_CACHE`/`RUFF_OUTPUT_FORMAT` are stripped, so it can never touch the real cache.
  - Anti-vacuity is the point: I deliberately broke the seed argv (a cache-key mismatch) and confirmed the test **fails loudly on its guard** rather than passing green — that exact mistake nearly cost the whole demonstration while prototyping.
- **`tests/test_ruff_gate_parity.py`** (static, 6 tests) — pins `--no-cache` on every invocation (catches a *partial* revert the behavioural suite structurally cannot see), pins that local and CI ruff take the same rule set from `pyproject.toml`, that CI never restores a `.ruff_cache`, and that no gate line auto-fixes.
- **Independently reproduced outside pytest**, before and after, using the real scripts via symlink; and confirmed `format.sh --fix` now actually sorts a file it previously left untouched under a poisoned cache.
- **Gate 2:** `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 checks, 7476 tests passed, coverage 94.13%, per-file gate 224 files ≥80%.
- **Gate 2.5:** `code-review-orchestrator` returned FIX_REQUIRED on the first pass (one blocker), then **CLEAN** after fixes.
- Repo-root shell suites (`scripts/ralph/test_*.sh`) all pass — 6/6.

### Gate 2.5 findings fixed before push

1. **Complexity C(15)** on a new parity test, over the ≤10 threshold — extracted `_single_gate_line`/`_is_checking_ruff_check` helpers; `xenon --max-absolute B` now exits 0 with all four singleton assertions preserved. (It escaped `check-all.sh` only because the complexity gate never scans `tests/` — filed #1191, along with the 20 *pre-existing* C-rank test functions it hides.)
2. **The cache-freedom guard could itself pass vacuously** — it tokenised the raw line, so a trailing comment naming `--no-cache` would satisfy it even if the real flag were gone. That is the same defect class as this issue: a gate that cannot fail. Fixed with quote-aware `shlex.split(..., comments=True)` plus an assertion against ruff's actual argv, and pinned by a new regression test. I verified the fix bites by reverting it (the new test fails) and verified it truncates none of the 8 command lines actually parsed.

`tests/test_scanner_coverage.py` (merged, gate-bearing, from #925) changes by **import substitution only** — I verified with `git diff -U0` that no line inside any `def test_*` changed and no assertion was altered, and confirmed the relocated `ci_steps()` returns results byte-identical to the original implementation (43 steps, `old == new`).

## Follow-ups filed rather than scope-creeping this P0

#1186 (mypy stale cache) · #1187 (`__pycache__` stale bytecode) · #1188 (CI ruff inline vs scripts) · #1189 (`lint.sh` unreachable `EXIT_CODE=$?`) · #1190 (`creek-tools/CLAUDE.md` truncated on `main` — §§7–10 and all appendices missing, which is why the doc entry landed in §6.1) · #1191 (complexity/docstring gates never scan `tests/`)

Closes #1119

### DIFF
--- a/creek-tools/.pre-commit-config.yaml
+++ b/creek-tools/.pre-commit-config.yaml
@@ -33,11 +33,16 @@ repos:
     - main
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.15.13
+  # --no-cache: ruff's cache key is mtime-only, and these hooks seed the very
+  # cache check-all.sh would otherwise trust; CI has none (issue #1119).
   hooks:
   - id: ruff
     args:
     - --fix
+    - --no-cache
   - id: ruff-format
+    args:
+    - --no-cache
 - repo: https://github.com/pre-commit/mirrors-mypy
   # Kept in sync with uv.lock (mypy 2.1.0) and pyproject.toml's
   # [project.optional-dependencies].dev pin so all three install paths
--- a/creek-tools/CLAUDE.md
+++ b/creek-tools/CLAUDE.md
@@ -405,6 +405,14 @@ All code must meet these standards before merging to main:
 
 #### Linting and Formatting
 - **Ruff**: lint + format, configured in `pyproject.toml` (no `|| true`).
+- **Ruff cache**: `--no-cache` on every invocation — lint and format,
+  check and fix — in `scripts/lint.sh`, `scripts/format.sh`, and both
+  `astral-sh/ruff-pre-commit` hooks. Ruff's per-file cache key is
+  mtime-only, so a local gate must never trust a cache CI doesn't
+  have. Costs ~0.05s on this tree (477 files); don't drop it for
+  speed. Issue #1119; enforced by `tests/test_ruff_cache_poisoning.py`
+  and `tests/test_ruff_gate_parity.py`. Same hazard is open for mypy's
+  cache (#1186) and `__pycache__` (#1187).
 - **Pylint**: ≥9.0 (`pylint creek/ creek_mcp/ --fail-under=9.0`, in CI
   and `lint-extended.sh`; CI-002).
 - **Bandit**: zero medium-or-above findings
--- a/creek-tools/scripts/format.sh
+++ b/creek-tools/scripts/format.sh
@@ -66,18 +66,30 @@ fi
 
 echo "=== Formatting (Ruff) ==="
 
+# The --no-cache on all four calls below is deliberate and load-bearing
+# (issue #1119). Ruff keys its per-file cache on mtime alone -- no size, no
+# content hash -- so any content change that leaves the mtime unchanged or
+# restored poisons it: cp -p, rsync -t, tar -x, a checkout of an older tree,
+# an mtime-preserving editor. CI never has a .ruff_cache and always re-reads
+# the file, so a local gate that answers from one can clear a tree CI
+# rejects. That is not hypothetical: it cost a real CI failure on PR #1117.
+# The fixing calls carry it too, because a fix run that answers from a
+# poisoned cache silently declines to fix the file it was handed.
+# Running cold measures at ~0.05s over this tree (477 files) -- negligible --
+# so do not drop the flag to speed the gate up. tests/test_ruff_cache_poisoning.py
+# and tests/test_ruff_gate_parity.py fail if you do.
 if $CHECK; then
     # Check import sorting
     if $VERBOSE; then
         echo "Checking import sorting..."
     fi
-    ruff check --select I . || { echo "✗ Import sorting check failed" >&2; exit 1; }
+    ruff check --select I . --no-cache || { echo "✗ Import sorting check failed" >&2; exit 1; }
 
     # Check code formatting
     if $VERBOSE; then
         echo "Checking code formatting..."
     fi
-    ruff format --check . || { echo "✗ Code formatting check failed" >&2; exit 1; }
+    ruff format --check . --no-cache || { echo "✗ Code formatting check failed" >&2; exit 1; }
 
     echo "✓ Code formatting check passed"
 else
@@ -85,13 +97,13 @@ else
     if $VERBOSE; then
         echo "Sorting imports..."
     fi
-    ruff check --select I --fix . || { echo "✗ Import sorting failed" >&2; exit 1; }
+    ruff check --select I --fix . --no-cache || { echo "✗ Import sorting failed" >&2; exit 1; }
 
     # Format code
     if $VERBOSE; then
         echo "Formatting code..."
     fi
-    ruff format . || { echo "✗ Code formatting failed" >&2; exit 1; }
+    ruff format . --no-cache || { echo "✗ Code formatting failed" >&2; exit 1; }
 
     echo "✓ Code formatted successfully"
 fi
--- a/creek-tools/scripts/lint.sh
+++ b/creek-tools/scripts/lint.sh
@@ -66,17 +66,27 @@ fi
 
 echo "=== Linting (Ruff) ==="
 
+# The --no-cache below is deliberate and load-bearing (issue #1119).
+# Ruff keys its per-file cache on mtime alone -- no size, no content hash --
+# so any content change that leaves the mtime unchanged or restored poisons
+# it: cp -p, rsync -t, tar -x, a checkout of an older tree, an
+# mtime-preserving editor. CI never has a .ruff_cache and always re-reads the
+# file, so a local gate that answers from one can clear a tree CI rejects.
+# That is not hypothetical: it cost a real CI failure on PR #1117.
+# Running cold measures at ~0.05s over this tree (477 files) -- negligible --
+# so do not drop the flag to speed the gate up. tests/test_ruff_cache_poisoning.py
+# and tests/test_ruff_gate_parity.py fail if you do.
 if $FIX; then
     if $VERBOSE; then
         echo "Fixing linting issues..."
     fi
-    ruff check . --fix
+    ruff check . --fix --no-cache
     EXIT_CODE=$?
 else
     if $VERBOSE; then
         echo "Checking for linting issues..."
     fi
-    ruff check .
+    ruff check . --no-cache
     EXIT_CODE=$?
 fi
 
--- a/creek-tools/tests/shell_command_support.py
+++ b/creek-tools/tests/shell_command_support.py
@@ -1,0 +1,189 @@
+"""Parsing helpers for tests that assert on shell scripts and CI workflows.
+
+Several suites pin the *shape of the quality gates themselves* rather than
+the behaviour of ``creek/``: they read ``scripts/*.sh``,
+``.github/workflows/*.yml`` and ``.pre-commit-config.yaml`` and assert that
+the commands those files run are the ones the gate contract promises
+(``test_scanner_coverage.py`` for the issue #925 scanner widening,
+``test_ruff_gate_parity.py`` for the issue #1119 ruff cache-freedom).
+
+They all need the same three primitives, and they all need them to be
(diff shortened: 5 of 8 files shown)


Geoffe-Ga:
## Summary
PR #1192 fixes issue #1119: `scripts/lint.sh` and `scripts/format.sh` could report a clean tree locally while CI (which never has a `.ruff_cache`) rejects the same bytes, because ruff's cache key is mtime-only. The fix appends `--no-cache` to all six ruff invocations across the two scripts and both `astral-sh/ruff-pre-commit` hooks, and adds two new regression suites (`test_ruff_cache_poisoning.py`, behavioural; `test_ruff_gate_parity.py`, static) plus a shared parsing helper module (`tests/shell_command_support.py`) that `test_scanner_coverage.py` is refactored to reuse.

## Strengths
- Root cause is verified empirically (mtime-only cache key), not assumed — the PR description documents the experiment.
- Fix covers both check *and* fix paths; the reasoning that a poisoned-cache `--fix` run silently *doesn't* fix the file (worse than a false green) is correct and non-obvious.
- The behavioural test (`test_ruff_cache_poisoning.py`) runs the **real** gate script via a symlink into a throwaway tree rather than a copy, so it can't drift from what ships; it also asserts the seed step actually poisoned the cache (anti-vacuity), matching the author's stated discipline of catching a "gate that can't fail."
- The static parity test (`test_ruff_gate_parity.py`) tokenizes with `shlex.split(..., comments=True)` specifically so a comment merely *mentioning* `--no-cache` can't satisfy the check — and there's a dedicated test (`test_a_commented_flag_cannot_satisfy_the_cache_free_gate`) proving that.
- Measured performance cost (~0.05s across 477 files) justifies hardcoding the flag rather than adding an env-var escape hatch — correctly reasoned as "a gate with an off switch is not a gate."
- Scope discipline: six related hazards found during investigation (mypy cache, `__pycache__` staleness, CI/script ruff divergence, unreachable `EXIT_CODE=$?`, truncated CLAUDE.md, complexity/docstring gates skipping `tests/`) were filed as separate issues instead of scope-creeping this fix.
- `test_scanner_coverage.py`'s refactor is verified by the author to be import-substitution only (no line inside a `def test_*` changed) — confirmed by reading the diff, which only touches imports and deletes now-duplicated helper functions.

## Security Concerns
None. 🟢 No user input, network, or subprocess-injection surface changes — this only appends a static flag literal to existing `ruff` invocations and adds file-reading test tooling.

## Problems
None blocking. Diff matches the PR description exactly; both scripts' new code paths were read directly and line up with the stated design (six invocations across `lint.sh`/`format.sh`, two pre-commit hook arg lists).

## Code Quality
- Non-blocking: `lint.sh`'s `EXIT_CODE=$?` is dead code under `set -euo pipefail` (any non-zero `ruff` exit already trips `set -e` before the assignment is read) — correctly identified by the author and deliberately left alone with a follow-up filed (#1189) to keep this diff to pure argv appends. Reasonable call.
- The `_ModeBranchTracker` state machine in `test_ruff_gate_parity.py` is a fair amount of machinery for what it does, but it's justified by the goal (parsing real bash `if/else/fi` mode branches without a real shell parser) and is well-documented with a clear docstring on why nesting needs depth tracking.

## CLAUDE.md Adherence
✅ Coverage — new behavioural/static tests only add coverage, don't reduce it (author reports 94.13% overall post-change).
✅ Linting/formatting — the change is itself a linting-gate fix; author reports 12/12 local checks green.
✅ Types — new modules use `from __future__ import annotations`, explicit type hints, `TYPE_CHECKING` guards for `Path`/`Callable` imports, consistent with MyPy-strict style elsewhere in the repo.
✅ Commit convention — `fix(scripts): ...` conventional-commit format, closes #1119.
✅ "Always use scripts, never call tools directly" — explicitly respected and called out in `test_ruff_cache_poisoning.py`'s docstring (direct `ruff` calls are labeled as fixture seeding/control probes, not gate invocations).

## Testing
Strong. Two independent, well-targeted regression suites at different layers (behavioural via real subprocess execution of the actual scripts; static via AST-free but comment-aware line parsing) close the gap from two angles, which is appropriate for a "gate that must never silently pass" class of bug. The PR body documents a genuine RED→GREEN cycle (5 failed/16 passed before, 22 passed after) and independent verification outside pytest. No concerns.

## Documentation
`creek-tools/CLAUDE.md` gains a concise, accurate "Ruff cache" bullet under Linting and Formatting pointing at the two new test files and the related open issues (#1186, #1187). Both scripts carry an inline comment explaining *why* `--no-cache` is load-bearing (not just what it does), consistent with the repo's comment philosophy.



## Verdict: LGTM

The fix is minimal, correctly scoped (--no-cache appended to every check and fix ruff invocation plus both pre-commit hooks), backed by a behavioural regression suite that exercises the real gate scripts via symlink and a static parity suite that would catch a partial revert. No bugs, security issues, or CLAUDE.md violations found, and the author proactively fixed a Gate 2.5 complexity finding and a vacuous-assertion bug in their own test before pushing.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5214922084 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1202, opened 2026-08-07
iteration-trigger 'NOT cleared' summaries still say VERDICT: LGTM — await-claude-review Step 4a merges on them, defeating do-not-auto-merge


Geoffe-Ga:
## Problem

`.github/workflows/iteration-trigger.yml` posts an owner-authored summary with
four fields:

```
<!-- iteration-trigger -->
**CI**: 5/5 Green
**VERDICT**: LGTM
**Action**: NOT cleared to merge: the do-not-auto-merge hold is set on this PR ...
```

`**VERDICT**` is computed from the review comment's verdict line **before** the
clearance chain runs, and the chain only ever rewrites `**Action**`. So on every
"NOT cleared" path the comment still says `**VERDICT**: LGTM` and
`**CI**: N/N Green`.

`.claude/skills/await-claude-review/SKILL.md` Step 4a decides the merge from
exactly those two fields:

> 3. If `VERDICT == LGTM` AND `x == y` (CI fully green): return `LGTM` to the
>    caller. The follow-up `Action:` line typically reads "cleared to squash
>    merge…". The caller proceeds to merge.

It never parses `**Action**:`. And Step 3 says this summary **short-circuits**
per-event classification, so it is the preferred wake path.

**Consequence: a webhook-woken session merges a PR the workflow explicitly
refused to clear** — including one a human parked with `do-not-auto-merge`,
which is the one control a human retains over an autonomous merge loop.

## Scope

Both pre-existing branches are affected:

- `HELD != no` — the `do-not-auto-merge` hold (or unreadable labels);
- `BEHIND != 0` — a head that is not current with its base.

The provenance branch added by #1181 avoids this by also setting
`VERDICT='NOT ATTESTED'` inside its own `elif`, so `VERDICT == LGTM` is false and
Step 4a cannot fire. The same one-line treatment is what the two older branches
need — this issue exists because widening it to them was out of #1181's scope,
not because it is hard.

## Proposed direction

Either:

- set a non-`LGTM` `VERDICT` value on every "NOT cleared" branch (matching what
  #1181 did), so the summary can never assert a merge-clearing verdict the same
  comment then refuses; **or**
- make `SKILL.md` Step 4a require the `Action:` line to be the affirmative
  clearance before returning `LGTM` — and keep `iteration-trigger.yml` in sync,
  since a prose contract between two files is exactly what drifted here.

The first is preferred: it fails closed in the workflow rather than relying on
[...]

Issue #1201, opened 2026-08-07
code-review.yml has no workflow_dispatch — a re-review costs a commit, and 'gh run rerun' replays the OLD workflow


Geoffe-Ga:
## Problem

`.github/workflows/code-review.yml` triggers on
`pull_request: [opened, synchronize, ready_for_review, reopened]` and nothing
else. There is **no `workflow_dispatch`**, so a review cannot be re-requested
without pushing a commit to the branch.

`gh run rerun <id>` is not a substitute: a re-run replays the workflow file **from
that run's own commit**, so it reproduces the old behaviour rather than the
current one. That distinction is load-bearing — it is precisely why #1181's
un-wedge path for a verdict that `pr-ready.sh` refuses is
`scripts/ralph/fleet.sh sync` (which advances HEAD *and* brings the current
workflow onto the branch) and not a re-run.

## Why it matters

Every "the reviewer needs to look again" situation currently costs a commit:

- a review that failed on a rate limit or a timeout (#1160/#1198);
- a verdict `pr-ready.sh` rejects as unattributable (#1181);
- a verdict a human simply distrusts.

Pushing an empty commit to obtain a re-review also invalidates the lane's CI and
burns a full ~14-minute round, so the cheapest available remedy is markedly more
expensive than the operation actually needed.

## Proposed direction

Add `workflow_dispatch` with a required `pr_number` input, and derive
`PR_NUMBER` / the checkout ref from either `github.event.pull_request.number` or
that input. Two things to get right:

- **Permissions.** `workflow_dispatch` is `write`-gated on the repo, so this does
  not widen who can trigger a review, but the `if: github.actor != 'dependabot[bot]'`
  guard and the secret-availability reasoning in the job's header comment must be
  re-checked against the new trigger.
- **The `concurrency` group** is `claude-review-${{ github.ref }}`; on a
  `workflow_dispatch` `github.ref` is the *dispatched* ref, not the PR head, so a
  manual re-review could fail to cancel — or wrongly cancel — a `pull_request`
  run. Key the group on the resolved PR number instead.

## Acceptance criteria

- [ ] A review can be re-requested for an open PR with no commit pushed.
- [ ] The manually dispatched run reviews the correct PR (it must satisfy
      #1181's `reviewed_pr_number` cross-check) and posts a verdict carrying the
      `<!-- creek-review pr=N -->` provenance marker.
- [ ] The concurrency group cannot let a manual run and a push run clobber each
[...]

Issue #1200, opened 2026-08-07
pr-ready.sh cannot tell a failed claude-review from a real CI failure — dispatches a fix worker at a reviewer malfunction


Geoffe-Ga:
## Problem

`scripts/ralph/pr-ready.sh` derives CI state purely from the `gh pr checks` EXIT
CODE (`pr-ready.sh:462-484`), which is the right call for CI — but it cannot tell
*which* check failed. A red `claude-review` check is therefore indistinguishable
from a red `Code Quality & Testing`, and both classify `ci-failed`, which
`ralph-tick.md` routes to a `ci-debugging` fix worker.

That misroute already exists (a review job that times out or hits the rate limit
turns the check red), and #1181 makes it reachable on a new path: the `Post
review` step now aborts with `::error::` + `exit 1` when the reviewer's
`reviewed_pr_number` does not match the PR it is posting to. That is exactly the
right thing for the *gate* — no verdict is posted, so nothing can be merged on a
review of the wrong diff — but it means a reviewer malfunction is reported to the
loop as **this PR's code being broken**. A worker is then dispatched to "fix" a
tree that is fine, and it has no failing test to work from.

## Proposed direction

Distinguish the review check from the rest without giving up the exit-code
contract. `pr-ready.sh` already knows the review job's name
(`REVIEW_CHECK_NAME="claude-review"`, `pr-ready.sh:375`) and already reads
`statusCheckRollup` in `review_gate_absent` (`pr-ready.sh:539-571`), so the data
is one lazy call away on the `ci-failed` path only.

Sketch: on a non-zero `gh pr checks`, ask whether the ONLY failing check is
`claude-review`. If so, emit a distinct token (`review-failed`?) whose remedy is
"re-run the review", not "dispatch a fix worker". Any non-review failure keeps
`ci-failed` exactly as today. Fail closed: an unreadable rollup stays `ci-failed`
(dispatching a worker at a real failure is recoverable; skipping a real failure
is not).

Costs to weigh: a new token means new routing surface in `ralph-tick.md` Step 1,
`watch-pr.sh`'s `IN_FLIGHT_TOKENS`, and `test_watch_pr.sh` — #1181's architect
deliberately declined to add a token there for the same reason. Decide whether
the token is worth it or whether the diagnostic belongs only on stderr.

Related: #1201 (no `workflow_dispatch` on `code-review.yml`), which is why
"re-run the review" is currently harder than it should be.

## Acceptance criteria

[...]

Issue #1199, opened 2026-08-07
pr-ready.sh honours a verdict from ANY comment author — a forged 'Verdict: LGTM' clears Gate 4


Geoffe-Ga:
## Problem

`scripts/ralph/pr-ready.sh` selects the merge-gating verdict as *the latest
top-level comment whose body matches `VERDICT_RE`* (`pr-ready.sh:504-509`), with
**no author check at all**. Any account that can comment on a PR can therefore
post

```
## Verdict: LGTM
```

and — with CI green and the branch current — flip `pr-ready.sh` to `ready`, which
is the token the orchestrator merges on. `.github/workflows/iteration-trigger.yml`
has the same shape (`iteration-trigger.yml:78`).

Discovered while working #1181 (a verdict that reviewed a *different* PR's diff).
#1181 adds a `<!-- creek-review pr=N -->` provenance marker, but that marker is a
**static literal** and is equally copyable, so it does not close this hole — it
was never intended to. This is pre-existing and #1181 neither creates nor worsens
it.

## Why it is not a one-liner

The posting identity is deliberately unstable: `code-review.yml`'s `Post review`
step authenticates as `${{ secrets.GEOFFE_GA_PAT || secrets.GITHUB_TOKEN }}`
(`code-review.yml:134`), so the verdict author is `Geoffe-Ga` when the PAT is
present and `github-actions[bot]` when it is not — and `Geoffe-Ga` is also the
human owner, who legitimately comments. An allowlist keyed on the author alone
would either accept a human impersonating the reviewer or reject real verdicts
whenever the PAT is absent.

Candidate approaches worth evaluating:

- key on `authorAssociation` plus a signed/opaque token minted by the workflow
  run and verifiable against `gh run view` (ties the comment to a run);
- have `pr-ready.sh` correlate the verdict comment with a *successful
  `claude-review` check run* on the same HEAD, rather than trusting the comment
  in isolation;
- post the verdict as a GitHub *review* (`gh pr review`) rather than an issue
  comment, so the author is structurally attributable.

## Acceptance criteria

- [ ] A `## Verdict: LGTM` comment posted by an account that is not the reviewer
      pipeline does not make `pr-ready.sh` print `ready`.
- [ ] Real verdicts still clear the gate under BOTH auth paths
      (`GEOFFE_GA_PAT` present and absent).
- [ ] `.github/workflows/iteration-trigger.yml` enforces the same rule — it is
[...]

PR #1203, opened 2026-08-07
fix(ralph): refuse a verdict that cannot name the PR it reviewed

### LINKED ISSUE #1181: claude-review can review a DIFFERENT PR's diff, and pr-ready.sh merges on it — Gate 4 bypassed
## Problem

`claude-review` can post a verdict that reviews **a different PR's changes**, and
`pr-ready.sh` then classifies the lane `ready` on the strength of it. Gate 4 is
bypassed without anything looking wrong.

## Observed, 2026-08-07 — PR #1117

PR #1117 is `feat(api): serve the Adepthood /v1 tracer` — **38 files**, a new
authenticated HTTP surface (`creek_mcp/httpapi/`, `creek_mcp/api/{routes,openapi}.py`,
auth middleware, `creek_mcp/audit.py`).

The verdict posted on it at `2026-08-07T05:27:24Z` opens:

> ## Summary
> **PR #1179** raises the `cryptography` floor from `>=48.0.0`/`>=48.0.1` to
> `>=50.0.0` in both `creek-tools` and `crawdad` to remediate PYSEC-2026-3552…
> Only manifest/lock/test files change — no production code is touched…

and concludes:

> ## Verdict: LGTM
> This is a clean, narrowly-scoped security dependency bump … that touches only
> pyproject.toml/uv.lock/test files in both subprojects … **already merged green
> through all six CI jobs**

Every claim is true of **#1179** and false of **#1117**. #1179 is a separate PR,
already merged. `scripts/ralph/pr-ready.sh 1117` returns **`ready`** on this
verdict: it is a `## Verdict: LGTM` posted after the PR's HEAD commit, which is
all the stale-verdict guard checks.

**Had the orchestrator merged on it, a 38-file public HTTP API — auth middleware,
route table, OpenAPI generation, an audit-log cache — would have landed on a
review of a dependency floor bump.** It was caught only because a human-facing
summary of the verdict text did not match the PR's known content.

## Likely mechanism (needs confirming)

#1117's HEAD is a **sync merge commit** (`a332aec`, merging `origin/main` which
carried #1179). A review that diffs `HEAD^1..HEAD` — rather than
`merge-base(base, head)..head` — sees exactly the incoming `main` changes, i.e.
#1179's diff, and reviews those.

If that is the mechanism, this is **systematic, not a one-off**: every PR whose
most recent push is a `fleet.sh sync` merge gets reviewed as though it were the
commits it just absorbed. Larger and longer-lived PRs are the most exposed, since
they sync most often — exactly the PRs where a real review matters most. #1117
has synced five times.

## Why the existing guards do not catch it

- **Stale-verdict guard** — checks only that `createdAt` is newer than the HEAD
  commit. Timing is correct here; *content* is not checked.
- **CI green** — orthogonal; CI ran the right code.
- **`changes-requested` token** (#1097) — only distinguishes verdict *type*.

Nothing anywhere ties a verdict to the diff it actually examined.

## Proposed fix

Two layers, both cheap:

1. **Make the reviewer diff the right range.** Confirm what `code-review.yml`
   passes and ensure it is `merge-base(base, head)..head`, never `HEAD^1..HEAD`.
   This is the root cause if the mechanism above holds.
2. **Make the verdict self-identifying, and check it.** Have the reviewer state
   the PR number and a diff fingerprint (file count, or a hash of the sorted
   changed-path list) in the verdict body; have `pr-ready.sh` refuse to honour a
   verdict whose stated PR number is not this PR. Fail **closed** — an
   unparseable or absent identifier reads `awaiting-review`, never `ready`.

Layer 2 is the one that makes the failure *detectable* rather than merely less
likely, and it composes with whatever the actual root cause turns out to be.

## Acceptance criteria

- [ ] A verdict naming a different PR number is **not** honoured by
      `pr-ready.sh`; the lane reads `awaiting-review`.
- [ ] A verdict with no parseable identifier fails closed the same way.
- [ ] The reviewer's diff range is confirmed (or corrected) to be
      `merge-base..head`, with the check covering a PR whose HEAD is a sync merge.
- [ ] `scripts/ralph/test_pr_ready.sh` covers both the mismatched-PR and
      missing-identifier cases.
- [ ] Tests demonstrably fail against current HEAD.

## Immediate state

PR #1117 is **not merged** and must not be, on this verdict. It needs a genuine
review of its own 38-file diff. Its earlier verdicts (a `COMMENTS` round and an
LGTM on 2026-08-01) reviewed the real content but are stale against the current
HEAD.

Related: #1160 (sync destroys a fresh LGTM) — both are failure modes introduced
by `fleet.sh sync` interacting with the verdict lifecycle.

Geoffe-Ga:
## Summary

- **Makes every verdict self-identifying and refuses the ones that are not.** `.github/workflows/code-review.yml` now prepends `<!-- creek-review pr=N -->` (N from `github.event.pull_request.number`) to every review it posts, and `scripts/ralph/pr-ready.sh` reads that marker **from the same comment its verdict selector already picked**. A marker that is absent, malformed, or names another PR makes the verdict **no verdict at all** — it gates neither `ready` nor `changes-requested`, so the lane reads `awaiting-review`. `.github/workflows/iteration-trigger.yml`, the second merge-clearance path, enforces the same rule.
- **Removes the root cause: the reviewer no longer has to guess which PR it is on.** The prompt now states the authoritative PR number and repo, says the checkout is a detached-HEAD merge ref so a bare `gh pr view`/`gh pr diff` *will* fail, and forbids inferring the PR from `git log`, a merge parent, or a PR listing. `Bash(gh pr list:*)` — the instrument the #1117 run actually guessed with — is removed from `--allowed-tools`, and a required `reviewed_pr_number` is added to the `--json-schema`: the agent reports the integer it passed to `gh pr diff`, the workflow compares it with `$PR_NUMBER`, and **a mismatch fails the check with no comment posted**.
- **Wires the coupling test to the file it guards.** `.github/workflows/ralph-recap-tests.yml` now runs the ralph shell suites on changes to `code-review.yml` and `iteration-trigger.yml`, so the emitter and the parser cannot drift without CI going red on the PR that causes the drift.

## Root cause — confirmed from the run logs, not inferred

The issue proposed `HEAD^1..HEAD` as the mechanism, and @Geoffe-Ga refuted it himself in a follow-up comment. Both are wrong for the same reason: **there is no diff range anywhere in `code-review.yml`.**

`gh run view 31150507883 --log` (the offending run on PR #1117) and `gh run view 31154718802 --log` (a correct run on #1183) show the same first command and the same failure:

```
$ gh pr view --json number,title,url,headRefName,baseRefName
could not determine current branch: failed to run git: not on any branch
```

`actions/checkout` checks out `refs/pull/<N>/merge` in detached HEAD, and the prompt never named the PR — so each agent **guessed**, and the guess is non-deterministic:

| run | how it resolved the PR | result |
|---|---|---|
| #1183 | `git log` + `gh pr list --state open` | guessed **1183** — correct |
| #1117 | `git log` showed `Merge a332aec… into 46182a6f…`; `gh pr list --state all --json …,mergeCommit` matched that **base parent** to PR **#1179**'s `mergeCommit.oid` | reviewed **#1179** — wrong |

The #1117 run then ran `gh pr diff 1179`, `gh pr view 1179`, `gh pr checks 1179` while `Post review` posted to `PR_NUMBER: 1117`. That is the whole bug, and it explains all three of the data points in the issue thread — merge-HEAD is irrelevant; what varies is the guess.

## Design notes

**The marker is a provenance attestation, not a wrong-diff detector.** It is computed by the workflow from the same value the comment is posted to, so it can never disagree with its own PR in normal operation. Its force is negative: `pr-ready.sh` refuses any verdict that did not come from a pipeline version which **runs the `reviewed_pr_number` cross-check**. That cross-check is the detector; the marker is the proof it was in force. Both are spelled out in `pr-ready.sh`'s header so the cross-check is not "simplified away" later.

**No grandfather clause for unmarked legacy verdicts.** A time-based cutoff would be *incorrect*, not merely lenient: `pull_request` runs execute the workflow file from the PR's own merge ref, not from `main`, so a branch cut before this change keeps emitting unmarked verdicts *after* it lands — there is no T that ever expires. The measured cost of failing closed was **zero re-reviews**: of 7 open PRs, 5 are Dependabot lanes (which consult no verdict), #1117 is opted out, and #1070/#943/#863 carry no verdict comment at all. Structurally the bill was already owed anyway — this PR touches `.github/workflows/`, which is on `RISK_SURFACE_RE`, so every open lane already faced a sync → HEAD advance → stale-verdict invalidation → re-review.

**Polarity.** This is the third fail-closed polarity in `pr-ready.sh` and it sides with `main-health.sh` (doubt ⇒ hold), not `review-quota.sh` (doubt ⇒ proceed). The one leg of `review-quota.sh`'s inversion that reaches here — a false positive is correlated fleet-wide — is neutralised by the cross-file coupling test, which is why that test is load-bearing and is now actually wired into CI. The three polarities are deliberately not uniform, and the header says so.

**Not done here, deliberately:** verdict-author authentication. The marker is a public literal and no author is checked, so a forged `## Verdict: LGTM` still clears the gate exactly as it did before this PR — that is pre-existing and is filed as #1199, not silently claimed as fixed.

## The near-miss this PR nearly shipped

Gate 2.5 caught, on the third review round, that **there are two emitters of a comment matching `pr-ready.sh`'s `VERDICT_RE`**, not one. `.github/workflows/iteration-trigger.yml`'s executive summary carries `**VERDICT**: LGTM`, which satisfies `VERDICT_RE`'s `(?:#{1,6}\s+|\*\*)?` alternative — and it posts **last** on every PR, ~15 s after the review. It can never carry a `creek-review` marker.

So the provenance guard as first written would have refused it on **every lane**, self-sustainingly: `awaiting-review` is in `watch-pr.sh`'s `IN_FLIGHT_TOKENS`, and a push produces a fresh review comment followed 15 s later by a fresh unmarked summary. That is the correlated fleet-wide false positive the polarity argument claims is neutralised by the coupling test — and it was not, because the coupling test only knew about one emitter.

Measured against the real comment payload of merged PR #906, running the production selector by hand:

```
before the fix: selected 2026-07-23T07:03:05Z  <!-- iteration-trigger -->\n**CI**: ...   <- the summary
after  the fix: selected 2026-07-23T06:53:17Z  ## Summary\nPR #906 fixes false-pos...    <- the actual review
```

Same shape on #905, #904, #902 — 4 of 4. The fix excludes that summary from the selector (it is a *report of* a verdict, not a verdict), inside the same single `--jq` and the same `$v` binding, with an anchored pattern and a coupling assertion tying the constant to the workflow's own `MARKER:` value. **It also closes a pre-existing bug:** the stale-verdict guard was reading the *summary's* `createdAt`, not the review's, so a stale review followed by a fresh summary passed the freshness check. That is now pinned by a test.

The generalised lesson is written into `pr-ready.sh`'s header, because it is the reusable part: a fail-closed polarity argument of the form *"only a broken emitter can trigger this, and the coupling test catches the emitter"* is only ever as good as the **enumeration of emitters**.

## Test plan

- `bash scripts/ralph/test_pr_ready.sh` → **242 passed, 0 failed** (was 181 before this PR; 61 new provenance, wedge and coupling assertions).
- **RED demonstrated against unfixed production code** — final test file, production files stashed: 106 passed / **119 failed**, headed by the reproduction of this exact incident:
  ```
  FAIL - the #1117 incident: fresh LGTM marked for a DIFFERENT PR → awaiting-review (expected 'awaiting-review', got 'ready')
  FAIL - unmarked fresh LGTM (legacy verdict) → awaiting-review (expected 'awaiting-review', got 'ready')
  FAIL - unmarked fresh CHANGES_REQUESTED → awaiting-review, not changes-requested (got 'changes-requested')
  ```
- **Emitter/parser agreement, mutation-tested** (a check for a field nothing emits would hold every lane forever). Mutating `code-review.yml` and re-running the suite: renaming the marker → 3 failures; making the marker not its own line → 1 failure; flipping the marker's `>` to `>>` so it is no longer first → 1 failure; dropping one of the two trailing newlines (still its own line, still parseable) → correctly 0 failures. The coupling case extracts the emitter's whole `printf` format with `sed`, renders it, and feeds the rendered bytes through `pr-ready.sh`'s own parser.
- `shellcheck --severity=warning scripts/ralph/*.sh` → exit 0.
- The other six ralph suites (`test_fleet`, `test_pick_next`, `test_watch_pr`, `test_main_health`, `test_review_quota`, `test_exec_bits`) all pass; `python3 -m pytest scripts/ralph -q` → 65 passed.
- `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12.
- All three edited workflows parse as valid YAML.

## Reviewer note

**This PR edits `.github/workflows/code-review.yml`, so `claude-code-action` self-skips (anti-tamper) and will post no automated verdict** — it posts the `<!-- review-self-skip -->` warning comment instead and the check stays green. That path is unchanged and is pinned by two new assertions: the self-skip body contains neither `creek-review` nor a verdict line, and the marker `printf` sits strictly after the self-skip `exit 0`. **This one needs a human LGTM.**

Gate 2.5 ran four rounds (`code-review-orchestrator`), ending `CLEAN`. Follow-ups filed: **#1199** (verdict comments are not author-authenticated — the marker attests to pipeline version, not authorship, and this PR's header says so rather than claiming otherwise), **#1200** (`pr-ready.sh` cannot tell a failed `claude-review` from a real CI failure), **#1201** (`code-review.yml` has no `workflow_dispatch`, so a re-review costs a commit), **#1202** (P1 — `iteration-trigger.yml`'s pre-existing `HELD`/`BEHIND` branches still post `**VERDICT**: LGTM`, and `await-claude-review` Step 4a merges on that field, defeating `do-not-auto-merge`).

Closes #1181

### DIFF
--- a/.github/workflows/code-review.yml
+++ b/.github/workflows/code-review.yml
@@ -61,12 +61,51 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           allowed_bots: "dependabot[bot]"
           show_full_output: true
+          # THE PR NUMBER IS STATED HERE BECAUSE THE AGENT CANNOT DERIVE IT
+          # (#1181). `actions/checkout` above checks out `refs/pull/<N>/merge` in
+          # DETACHED HEAD, so `gh pr view --json number` — the agent's natural
+          # first move — fails with "could not determine current branch: failed
+          # to run git: not on any branch". On PR #1117 it then guessed via
+          # `gh pr list`, matched the merge ref's base parent to PR #1179's
+          # `mergeCommit.oid`, reviewed that bump's diff, and this workflow
+          # posted the resulting LGTM onto #1117. There is NO diff range anywhere
+          # in this file: the defect was PR-number RESOLUTION, not range
+          # selection, and a "fix" to a range would change nothing.
+          #
+          # INTERPOLATION SAFETY: `github.event.pull_request.number` is an
+          # integer and `github.repository` is `owner/repo` over GitHub's own
+          # name charset. Neither is attacker-controlled free text the way
+          # `title` / `body` / branch names are, so splicing them into the prompt
+          # adds no injection surface.
+          #
+          # AND THE USUAL `env:` + `$VAR` HARDENING DOES NOT APPLY HERE. `prompt`
+          # is an ACTION INPUT, not a shell script: nothing expands `$VAR` on the
+          # way in, so an `env:` indirection would ship the literal characters
+          # `$PR_NUMBER` to the model and silently restore the bug this whole
+          # change exists to fix. Do not "harden" it that way.
           prompt:  |
+            You are reviewing pull request #${{ github.event.pull_request.number }}
+            of the repository ${{ github.repository }}.
+
+            THAT NUMBER IS AUTHORITATIVE. It is the PR this review will be posted
+            to, and it is the only PR you may review. Do not try to work out
+            which PR you are on: the checkout is `refs/pull/${{ github.event.pull_request.number }}/merge`
+            in detached HEAD, so a bare `gh pr view` or `gh pr diff` WILL fail
+            with "could not determine current branch: failed to run git: not on
+            any branch". That failure is expected — it is not a reason to search.
+
+            NEVER infer the pull request from `git log`, from a merge commit's
+            parent OID, or by listing pull requests. Pass the number and the
+            repository explicitly on every `gh` call:
+
+              gh pr diff ${{ github.event.pull_request.number }} -R ${{ github.repository }}
+              gh pr view ${{ github.event.pull_request.number }} -R ${{ github.repository }}
+
             Review this pull request against the repository's CLAUDE.md. Read the
-            change with `gh pr diff` and `gh pr view` (and any files you need).
+            change with those commands (and any files you need).
 
             Do NOT post a comment yourself. Return ONLY the structured result with
-            three fields:
+            four fields:
 
             - "review_markdown": the full review as GitHub markdown, using this
               template (omit the verdict here — it is a separate field below):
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
⚠️ This PR edits the review workflow — no automated verdict; manual review required.

Issue #1210, opened 2026-08-07
fix(compost): _is_paradox reads fragment.tags but the spec puts paradox in emotional_texture


Geoffe-Ga:
## Problem

`creek/generate/compost.py:131-133`:

```python
def _is_paradox(fragment: Fragment) -> bool:
    """Return whether *fragment* is a paradox note (skipped by compost detection)."""
    return "paradox" in fragment.tags
```

used at `compost.py:327` (`if self._skip_paradox and _is_paradox(fragment)`).

The ontology spec says the opposite field. `docs/Ontology/creek_ontology_agent_prompt.md:712-715`:

> IMPORTANT: Do not force classifications. "Unclassified" is a valid and
> respected answer. Liminal content that resists categorization is VALUABLE.
> Do not resolve paradoxes or contradictions — tag them with
> **"paradox" in emotional_texture**.

Nothing writes `"paradox"` into a **fragment's** `tags`. The one place `paradox`
does appear as a tag is `creek/generate/paradox.py:590`
(`create_paradox_note` -> `tags = ["paradox"]`), which tags the generated
**paradox note**, a different note type — not the fragments it was derived from.

So `compost`'s `_skip_paradox` guard is a dead branch: it can never fire for a
fragment produced by the pipeline.

## Interaction with #878

#878 gave `Fragment.tags` a hashtag producer, so a fragment whose body literally
contains `#paradox` will now be skipped by compost. That is incidental, not the
spec-mandated path — the spec routes paradox through `emotional_texture`, which
#878 also gave a producer (LLM `texture:` section), and `_is_paradox` does not
read it.

## Decide, then fix one of two ways

1. Read `emotional_texture` (spec-conformant), or
2. Read both fields, and document why `tags` is also honoured.

Whichever is chosen, add the `paradox` marker to the documented
`EMOTIONAL_TEXTURE_VOCABULARY` rationale so the LLM can actually produce it, and
pin the behaviour with a test — today nothing does.

## Acceptance criteria

- [ ] `compost`'s paradox skip fires for a fragment the pipeline can actually produce
- [ ] The chosen field matches the ontology spec, or the deviation is documented
- [ ] A test pins the behaviour
- [ ] `./scripts/check-all.sh` passes

Found while working #878.

Issue #1209, opened 2026-08-07
feat(ingest): suppress Discord channel-mention artifacts from extracted tags


Geoffe-Ga:
## Context

#878 added `creek/classify/tags_pass.py`, which extracts inline `#hashtags` from
fragment bodies into `Fragment.tags`.

Its precision rules reject Discord **snowflake** channel mentions (a candidate
needs >= 2 alphabetic characters, and is rejected on a run of >= 8 consecutive
digits). Measured on a 3,000-file sample of the demo vault, that removes the
entire flood:

```
688  #c762760820929069057
539  #c942223408115122187
122  #c690018993905467490
 66  #c510347244294242325
```

## Residual

Discord channel mentions with **human-readable** names survive:

```
17  #living-room
14  #family-business
13  #common-space
```

The raw Discord export carries `<#id>` mention syntax which the pre-ingest
transform renders as `#channel-name`. By the time `tags_pass` sees the body, only
the rendered form exists, so `#living-room` is indistinguishable from an author
literally typing `#living-room` in prose. This is documented as a known residual
in the `tags_pass` module docstring.

## Possible fix

Source-aware suppression at the Discord adapter (`creek/ingest/discord.py`),
which still has the `<#id>` form and the channel roster, so it can mark rendered
channel mentions before they reach the shared extractor — e.g. by emitting them
in a form the generic pattern does not match, or by passing a per-source
exclusion set into `assemble_ingested_fragment`.

## Priority rationale

Low. At 17 and 14 occurrences these are signal-level noise in a 35,330-fragment
vault, not a flood, and a channel name is arguably a legitimate topical tag. File
now so the residual is tracked rather than folklore.

## Acceptance criteria

- [ ] Rendered Discord channel mentions do not become `Fragment.tags`
- [ ] An author literally writing `#living-room` in prose still gets the tag
- [ ] `creek/classify/tags_pass.py` stays source-agnostic and pure
- [ ] `./scripts/check-all.sh` passes

Follow-up to #878.

Issue #1208, opened 2026-08-07
feat(classify): emit emotional_texture on the weighted classification path


Geoffe-Ga:
## Context

#878 gives `Fragment.emotional_texture` a producer by adding a `texture:` section
to the main classification schema (`creek/classify/llm/prompts.py` +
`_ALLOWED_TOP_LEVEL_KEYS` in `creek/classify/llm/parsing.py`).

`creek/classify/weighted.py::classify_weighted` carries its **own** prompt and its
**own** top-level key allow-list, so on the weighted path `emotional_texture`
still comes from nowhere and stays `[]`.

This is the same deferral #877 made for `praxis_potential`, recorded in
`creek/classify/praxis_pass.py`'s "Known residuals": *"The weighted path is
heuristic-only. `classify_weighted` carries its own prompt and its own top-level
key allow-list, so on that path praxis comes from `detect` alone and never from
the model."*

## Why it is deferred rather than done

Widening the weighted allow-list is a **second LLM schema change with its own
SEC-004 injection boundary** and its own test surface. It does not belong inside
a PR whose subject is giving two model fields their first producer.

## Acceptance criteria

- [ ] The weighted prompt emits a bounded `texture:` section
- [ ] The weighted allow-list is widened one key at a time, and never to a
      `Fragment` field name (`privacy_tier` stays rejected)
- [ ] The same caps apply (`_MAX_TEXTURES` = 5, `_MAX_TEXTURE_CHARS` = 32)
- [ ] The merge stays union / never-lose
- [ ] `./scripts/check-all.sh` passes

Follow-up to #878.

Issue #1207, opened 2026-08-07
feat(classify): free, non-destructive tags backfill for an already-classified vault


Geoffe-Ga:
## Context

#878 gives `Fragment.tags` a producer: a pure, deterministic hashtag extractor
(`creek/classify/tags_pass.py`) wired at the universal ingest chokepoint
(`creek/ingest/base.assemble_ingested_fragment`) and beside `apply_praxis` in
`creek/classify/classify_engine.py`.

That fixes **new** ingests. It does **not** fix the existing vault.

## Problem

Every fragment in the 35,330-fragment demo vault is stamped
`classification_method: llm`, so the OPS-001 resume short-circuit skips it and
the tags pass never runs. The only route that works today is

```
creek classify --method llm --force
```

which re-sends all 35,330 fragments to the configured provider and costs a full
re-classification in tokens — to recover data that is already sitting in the
fragment bodies on local disk and needs no model at all.

`creek classify --method rules --force` is free but **destructive** and must not
be used: `_classify_one` (`classify_engine.py:1211`) returns `rules.classify(...)`,
which at `rules.py:573-574` does `fragment.model_copy(update=updates)` overwriting
`frequency`/`wavelength`/`voice`, and `_finalise_fragment_write` then stamps
`classification_method: rules`. It would replace every LLM classification with
keyword output.

So today the Tag Garden stays empty on the operator's real vault unless they pay
for a full re-classification.

## Why tags may deserve the exception #877 declined

#877 deliberately did **not** widen `_write_tier_only` for `praxis_potential`,
reasoning that an untiered fragment is a live cloud-egress hole whereas a missing
praxis verdict is only a missing feature. Tags differ in two ways that may flip
that call:

- Hashtag extraction is a **lossless mechanical restatement of literal
  author-written syntax**, not a judgment. Re-deriving it cannot disagree with
  operator curation the way a keyword heuristic can.
- The field is provably `[]` on 100% of 35,330 fragments, so there is no
  curation to overwrite. (The merge in `tags_pass` is union/never-lose anyway.)

## Proposed direction (needs its own review — do not ride it inside a feature PR)

A **sibling** narrow writer, e.g. `_write_tags_only`, leaving `_write_tier_only`'s
auditable "changes ONLY the privacy-tier fields" promise from #876 untouched — or
a dedicated `creek fill` step that walks the vault and applies `tags_pass.merge`
in place.

## Acceptance criteria

[...]

PR #1215, opened 2026-08-07
feat(classify/ingest): give tags and emotional_texture producers

### LINKED ISSUE #878: feat(classify/ingest): tags and emotional_texture are never populated — Tag Garden and texture features permanently empty
## Role

## Goal

Give `Fragment.tags` and `Fragment.emotional_texture` producers, so the Tag Garden and the texture-based features stop being permanently empty.

## Context

Both fields default to `[]` and **nothing in the pipeline ever populates them** — every consumer is dead at runtime:

- `tags` consumers: `creek report --type tags` → `TagGardenGenerator` (creek/generate/tags.py). Evidence (demo vault `creek-demo-2026-06-07`, 35,330 fragments, full pipeline run): Tag-Garden.md regenerated 2026-07-22 still reads "*No tags found in vault.*" A 3,000-fragment sample shows `tags: []` on 100%.
- `emotional_texture` consumers: temporal resonance scoring (+0.1 per shared texture tag, creek/link/temporal.py:64) and the wavelength report's emotional-texture cloud (creek/generate/wavelength.py:451, :1077). Sample shows `emotional_texture: []` on 100% — so the resonance dimension never contributes and the texture cloud always renders empty.
- Producers found by grep: none. The LLM classify schema (`_ALLOWED_TOP_LEVEL_KEYS` in creek/classify/llm/parsing.py) accepts only `frequency`, `wavelength`, `voice`, `confidence_scores`. Ingest does not extract `#hashtags` from bodies into frontmatter either.
- Same folder-without-producer pattern as #584 (Observations) and the praxis_potential issue (filed alongside this one).

## Proposed Solution

Recommended (deterministic first, LLM second):

1. **Ingest-time hashtag extraction** — during ingest, lift inline `#tags` (and Obsidian frontmatter tags from markdown sources) into `Fragment.tags`, normalized (lowercase, kebab). Free, offline, immediate value for markdown/Obsidian sources.
2. **LLM emission for `emotional_texture`** — add an `emotional_texture` list (bounded: ≤5 items, each ≤32 chars, from a documented vocabulary with free-form allowed) to the classification prompt/schema, with the same length-capping discipline as `descriptor` (#319).
3. Backfill via the `creek fill` classify-upgrade offer, same pattern as #736 (opt-in, cost surfaced).

If maintainer prefers a different split (e.g. tags also from LLM), note it on this issue before implementing — but hashtag extraction is safe to do regardless.

## Acceptance Criteria

- [ ] Ingesting a markdown note containing `#recovery #writing` yields `tags: [recovery, writing]` in fragment frontmatter
- [ ] Tag Garden lists tags, growth, and clusters on a fixture vault with tagged fragments
- [ ] LLM classify populates `emotional_texture` (bounded, validated, falls back to `[]`), and temporal linking counts shared textures in resonance scores
- [ ] Wavelength report renders a non-empty texture cloud on a fixture with textured fragments

## Constraints

- SEC-004 stays: `privacy_tier` remains rejected from LLM output
- Cap list lengths and item lengths from the LLM (same rationale as the descriptor cap, #319)
- Backfill opt-in only; follow stay-green TDD

Geoffe-Ga:
## Summary

- **Gives `Fragment.tags` a producer** — a new pure module `creek/classify/tags_pass.py` extracts inline `#hashtags` from fragment bodies, wired at `assemble_ingested_fragment` (the universal ingest chokepoint) and in the classify engine beside `apply_praxis`. Deterministic, offline, no LLM.
- **Gives `Fragment.emotional_texture` a producer** — a bounded `texture:` section added to the *existing* classification prompt/schema/parser. **Zero new LLM calls and zero new provider calls**, so `ModelRouter`'s Intimate-never-cloud enforcement covers it unchanged.
- **Adds observability and an honest backfill nudge** — `ClassifySummary.tags_extracted` (mirroring `praxis_marked`), the same key on the MCP `classify` tool, and a `creek fill` hint that reports a count only.

## The defect

Both fields shipped with `default_factory=list` and **no production writers**, so every consumer was structurally dead:

| Field | Consumers, all dead |
|---|---|
| `tags` | Tag Garden (`generate/tags.py`), orphan-tag lint (`lint/checks/tags.py`), compost tag clustering (`generate/compost.py:407`) |
| `emotional_texture` | temporal resonance `+0.1` per shared texture (`link/temporal.py:107`), wavelength texture cloud (`generate/wavelength.py:497`) |

Evidence gathered on the operator's real vault (`creek-demo-2026-06-07`, 35,330 fragments, full pipeline run): **2000/2000 sampled fragments read `tags: []` and `emotional_texture: []`** — zero populated. `Tag-Garden.md`, regenerated 2026-07-22, says *"No tags found in vault."* `git log -S` shows neither field ever had a producer, so this is a **never-worked path, not a regression**. Every reader degrades silently, which is why it stayed invisible.

Same folder-without-producer shape as #584 and #877; this change deliberately mirrors #877's `praxis_pass.py`.

## Design notes worth a reviewer's eye

**Precision is the substance, not a detail.** A naive `#\w+` sweep of this vault returns Discord channel-mention snowflakes far more often than tags — `#c762760820929069057` alone appears **688 times** in a 3,000-file sample, `#c942223408115122187` 539 times. Admitting them would not fix the Tag Garden, it would bury it, which is strictly worse than the current silence. So a candidate needs **two alphabetic characters** and must carry **no run of eight digits**; fenced code blocks and inline code spans are stripped before scanning so a shell `# comment` or a CSS `#fff` never becomes a tag. camelCase is deliberately **not** split — splitting is a guess, and this pass does not guess.

**The LLM key is `texture`, not `emotional_texture`.** `_ALLOWED_TOP_LEVEL_KEYS` is the SEC-004 injection boundary, and its docstring states it is never widened to a `Fragment` field name. Wrapping the list in a `texture:` section keeps a payload emitting the bare field name **rejected**, and `privacy_tier` stays rejected. Caps (`≤5` items, `≤32` chars each) are enforced in the parser, and junk *items* are dropped without discarding the whole list.

**The seed vocabulary is load-bearing.** The one consumer that pays for this axis scores *exact string set-intersection*. Invite free-form phrases and the model coins a unique poetic label per fragment: the field is populated, the `+0.1` bonus never fires, and the axis is as dead as the empty list — only harder to notice. The prompt therefore seeds a documented vocabulary while explicitly still allowing free-form.

**Both merges are union / never-lose.** `clean/context.py:177` already appends `low-priority` to `tags`, and an operator may hand-add tags in Obsidian, so a replacing merge would delete live data.

**One deliberate omission:** no third wiring call in `pipeline.py`. Praxis needs one because it is a classification verdict that must run *after* the LLM dispatch; tags derive from a body no stage mutates, so the ingest chokepoint is strictly earlier and sufficient.

**One pre-existing test rewritten — please scrutinise.** `test_caps_long_content_length`'s `assert len(prompt) <= 8192 + 8192` was a spot check on a single fragment id, not a ceiling: the few-shot block rotates per id, and the **worst-case rotation already rendered 16519 > 16384 at HEAD**. This PR's 11th classification dimension (+719 template chars) merely moved the fixture to the failing side. The replacement bounds the prompt against its three independently-capped parts, so it holds for *every* id and cannot go stale. Gate 2.5 independently judged this a strengthening rather than a relaxation, but it is the one change in the diff that touches a threshold, so it is called out here rather than buried. The prompt header genuinely grew ~719 chars, which is a real token cost at 35k scale.

## Backfill — stated plainly

**This populates new ingests only. An existing vault gets nothing until the operator runs a command.**

- Every fragment in the demo vault is `classification_method: llm`-stamped, so the OPS-001 resume short-circuit skips it and neither axis is backfilled. `_write_tier_only` is deliberately **not** widened — its narrow, auditable #876 promise is load-bearing.
- **`tags` on existing fragments**: only via `creek classify --method llm --force`, which re-sends all 35,330 fragments and costs a full re-classification in tokens — to recover data already sitting in local bodies. A free, non-destructive path is filed as **#1207**.
- **`emotional_texture` on existing fragments**: no free backfill is possible *in principle* — the field comes only from the model. Same `--force` command, same cost.
- `creek classify --method rules --force` is free but **destructive** and is named nowhere in the code or hints: it would overwrite every LLM classification with keyword output.
- `creek fill` now surfaces the outstanding tag count (count only, per the #846/#848 config-oracle rule) and names the paid command with its cost.

So: these features stop being *structurally* empty immediately, and become *actually* non-empty on the existing vault only after one paid run or after #1207 lands. This PR should not be read as "the Tag Garden now works on your vault".

## Test plan

- `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 checks, **7628 tests pass** (baseline 7523; +129 net new), coverage 94.11%, per-file gate green.
- **RED first, verified against unfixed HEAD**: `creek ingest --type markdown` on a note containing `#recovery #writing` failed with `assert [] == ['recovery', 'writing']`. 28 tests plus one uncollectable new module were red before implementation.
- **End-to-end through the shipped CLI** (not just units): ingesting that note now yields `tags: [recovery, writing]`, and `creek report --type tags` renders a populated Tag Garden with counts, fragment links and new-tag growth — a file that had read *"No tags found in vault."* since it was written. The same probe confirmed extraction correctly **rejects** a Discord snowflake, a `## heading`, and both `# comment` and `#fff` inside a fenced python block.
- `creek fill` verified to emit only a count — no fragment id, body excerpt, or tag names.
- Import-cycle claim for the new `creek/ingest/base.py` import verified independently by importing each affected module first, in isolation.
- Consumer tests (Tag Garden, temporal linking, wavelength texture cloud) were written to be **green at HEAD** — they prove the readers were always fine and only the vault was empty.
- Gate 2.5 `code-review-orchestrator`: **CLEAN**, no blockers.

## Follow-ups filed (not implemented here)

- **#1207** (P1) — free, non-destructive `tags` backfill for an already-classified vault
- **#1208** (P2) — emit `emotional_texture` on the weighted classification path
- **#1209** (P3) — suppress Discord channel-mention artifacts from extracted tags
- **#1210** (P3) — `compost._is_paradox` reads `fragment.tags` but the spec puts `paradox` in `emotional_texture`

Closes #878
Refs #877


Geoffe-Ga:
## Summary
This PR gives `Fragment.tags` and `Fragment.emotional_texture` real producers for the first time — a new pure `tags_pass.py` module wired at the ingest chokepoint and beside the praxis pass in the classify engine, plus a bounded `texture:` section added to the existing LLM classification response (zero new calls). The change is exceptionally well documented and tested, with careful attention to precision (Discord-snowflake rejection), SEC-004 injection-boundary discipline, and idempotency. However, the union-merge "never lose a tag" invariant that the PR explicitly designs around and repeatedly asserts in comments/tests has a real gap when a fragment's pre-existing list already exceeds the cap.

## Strengths
- Root-caused with real evidence (2000/2000 sampled fragments, `git log -S`) rather than assumed; clearly scoped as a never-worked path, not a regression.
- The `#\w+` naive-extractor failure mode (Discord snowflakes dominating the tag garden) is identified concretely and defended against with two independent, well-tested rules (`_MIN_ALPHA_CHARS`, `_MAX_DIGIT_RUN`).
- `texture:` section deliberately avoids the field name `emotional_texture` to preserve the SEC-004 `_ALLOWED_TOP_LEVEL_KEYS` injection boundary — and this is pinned with a dedicated regression test (`test_rejects_a_bare_emotional_texture_top_level_key`).
- No new LLM calls for the texture axis — rides the existing classification response, preserving `ModelRouter`'s Intimate-never-cloud routing without new plumbing.
- Idempotency is explicitly tested for both passes (`is` identity checks, not just equality), and `creek fill` hints report counts only, never fragment content — tested for that too (config-oracle rule).
- Extraction regexes are all linear-time with no nested quantifiers, called out explicitly in the security note given they run over 35k+ unbounded, attacker-influenced bodies.
- Test coverage is very thorough for the documented scenarios: normalization, camelCase-not-split, code-fence/inline-code stripping, truncation determinism, cross-source temporal scoring slope, wavelength report round-trips.

## Security Concerns
🟢 LOW — No new attack surface introduced. The `texture` top-level key addition to `_ALLOWED_TOP_LEVEL_KEYS` is handled correctly and is regression-tested against widening to `privacy_tier`/`emotional_texture`. Regex patterns are all bounded/linear. `creek fill` hints avoid disclosing vault content (tested).

## Problems
🔴 **BLOCKING — the merge/cap interaction silently drops pre-existing tags beyond the cap, violating the PR's own "never lose a tag" contract.**

Both `creek/classify/tags_pass.py::merge` and `creek/classify/llm/parsing.py::_merge_textures` build the merged list by iterating `(*current, *candidate)` and `break`ing once the cap is hit:

```python
for tag in (*current, *candidate):
    ...
    merged.append(tag)
    if len(merged) == _MAX_TEXTURES:  # or _MAX_TAGS
        break
```

This is correct when `len(current) <= cap` (the only case the tests cover — see `test_the_union_is_capped_too` in `test_classify.py`, which seeds exactly 5 recorded textures). But there is **no model-level cap** on `emotional_texture` or `tags` — `tests/test_models.py::test_a_hand_authored_ten_item_texture_list_is_not_truncated` explicitly proves an operator can hand-author more than 5 texture tags (or, analogously, more than 32 `tags`) and the model preserves them verbatim.

When such a fragment is re-classified (e.g. via the very remediation command this PR tells operators to run for the backfill nudge — `creek classify --method llm --force`), `_apply_texture`/`apply_tags` compares `merged` (now truncated to the cap) against `current` (still its original, longer length). Since `merged != current`, the truncated list is written back via `fragment.model_copy(update=...)`, **permanently deleting** every hand-authored tag past the cap boundary — even ones the operator manually curated in Obsidian, which the module docstrings say must never happen ("a merge that filtered its own inputs would silently delete author-written vault content").

This is reachable through ordinary use (`creek ingest` re-normalizing a migrated note with >32 frontmatter tags, or `creek classify --force` on a fragment with >5 hand-added emotional-texture tags) and directly contradicts the design principle the PR spends significant documentation asserting. Please add a length guard so `current` entries are never dropped by the cap — e.g. cap only the *candidate* additions once `len(current) >= cap`, or preserve all of `current` and cap only what's appended from `candidate`.

## Code Quality
- `_record_tags` in `classify_engine.py` does `fragment.tags.copy() != before` — the `.copy()` is unnecessary (comparing lists compares values, not identity); harmless but adds noise. Non-blocking.
- The `_scan_fill_gaps` tags-backfill detector (`set(extract(body)) - set(fragment.tags)`) doesn't run `fragment.tags` through the same normalizer `merge()` uses, so a fragment with `tags: [Recovery]` and body `#recovery` will report as "backfillable" purely due to case, even though the only actual work a classify run would do is re-normalize casing. This is arguably consistent with what `_record_tags` would count as "extracted" (the field does change), so it's not wrong, but the hint's own docstring ("the free extractor finds a tag the disk does not record") slightly overstates what's being detected. Worth a one-line comment acknowledging the case-difference trigger, but not blocking.

## CLAUDE.md Adherence
✅ TDD-shaped: tests are extensive and accompany the implementation throughout the diff.
✅ Docstrings are thorough and exceed the 95% interrogate bar in spirit (every public and most private functions documented with rationale).
❓ Coverage/lint/mypy/complexity gates could not be executed in this review environment (no `uv`/tool access), and CI ("Code Quality & Testing" 3.11/3.12/3.13, `claude-review`) was still `IN_PROGRESS` at review time — those gates should be confirmed green before merge, per the "Stay Green" workflow.
✅ Commit/PR framing follows conventional structure; issue references (#878) are consistent throughout.

## Testing
Insufficient at one specific boundary: no test exercises `merge`/`_merge_textures` when `current` already exceeds the cap (`_MAX_TAGS`/`_MAX_TEXTURES`). `test_the_union_is_capped_too` and the `tags_pass` truncation tests only assert the cap is respected when `current` is exactly at or under the ceiling — this is precisely the gap that hides the data-loss bug above. This should be added as part of fixing the Problems item (an over-cap `current` list must survive a merge/classify pass untouched, or at minimum lose none of its original entries).

Otherwise, testing is a strength of this PR: idempotency, union-never-drops, SEC-004 boundary, prompt rendering safety, end-to-end ingest/classify/wavelength-report paths, and the `creek fill` hint's silence/noise conditions are all well covered.

## Documentation
Exceptional — module and function docstrings consistently state the *why* (empirical evidence, prior incidents, security rationale) rather than just the *what*, matching the repo's documentation bar. The PR description itself is unusually rigorous about tradeoffs (precision-over-recall, seed vocabulary rationale, union-merge justification). No documentation gaps found beyond the mismatch between the stated "never lose a tag" guarantee and the actual over-cap behavior noted above.


## Verdict: CHANGES_REQUESTED

The union-merge cap logic in both `tags_pass.merge` and `parsing._merge_textures` silently truncates and overwrites a fragment's *existing* recorded tags whenever that list already exceeds the cap (32 for `tags`, 5 for `emotional_texture`) — directly contradicting the PR's own repeatedly-documented and tested "never lose a tag" invariant. This is reachable through the exact remediation command the PR tells operators to run (`creek classify --method llm --force`) and through ordinary ingest, and the test suite only covers the exact-at-cap boundary, not the over-cap case, so it slipped through.

Geoffe-Ga:
**CI**: 4/7 Green
**Action**: pull comment 5221579861 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**Action**: pull comment 5221579861 to see in-depth feedback and continue iterating

Geoffe-Ga:
Addressed in `894c3af`. Thanks — the blocking finding was correct, reachable, and I reproduced it before fixing it.

## 🔴 BLOCKING — merge/cap silently drops pre-existing tags beyond the cap

Confirmed and fixed. Reproduced against the pre-fix code first:

```
OLD merge: len(current)=40 -> len(result)=32  lost=8
```

`_MAX_TAGS`/`_MAX_TEXTURES` are now a **growth ceiling** rather than a length limit, exactly as suggested (cap only what the candidate adds):

- `tags_pass.merge` seeds from a new `_canonical()` helper — normalise + dedupe, **uncapped** — then admits at most `max(0, _MAX_TAGS - len(kept))` candidates.
- `parsing._merge_textures` seeds from `dict.fromkeys(current)` (no normalisation; that axis has a single producer) and admits at most `max(0, _MAX_TEXTURES - len(merged))`.

So the recorded side is always a *prefix* of the result: a list past the ceiling keeps every entry and gains none, making repeated passes idempotent rather than a ratchet. The ceiling still binds in the direction it was written for — an unbounded, attacker-influenced body cannot inflate frontmatter past it.

Boundary behaviour, with literal expectations:

| `len(current)` | 0 | 1 | 31 | 32 | 33 | 100 |
|---|---|---|---|---|---|---|
| tags added | 32 | 31 | 1 | 0 | 0 | 0 |
| result length | 32 | 32 | 32 | 32 | 33 | 100 |

(textures: `{0,1,4,5,6,10}` → added `{5,4,1,0,0,0}`, prefix intact.)

## 🔴 Testing — no test exercised `current` already over the cap

Fixed as part of the above. New: `TestMergeNeverTruncatesTheRecord`, `TestMergeBoundsGrowth`, `TestMergePreservationExceptions`, `TestTextureMergeNeverTruncatesTheRecord`. Both boundaries are swept from **both** sides (31/32/33 and 4/5/6), every expectation is a **literal** rather than derived from the constant — an expectation read off `_MAX_TAGS` moves with the constant and cannot catch an off-by-one — and each case asserts the recorded prefix, which is what separates "admitted nothing" from "deleted the tail". Idempotency (`merge(merge(c,x),x) == merge(c,x)`) is pinned under *and* over the cap.

Verified RED before GREEN: the four new texture cases fail against the previous implementation.

## 🔵 `_record_tags` redundant `.copy()`

Dropped — `!=` never mutates. Docstring corrected too: re-canonicalisation can *shorten* the list (two spellings folding to one), so "can only ever grow" was not quite right.

## 🔵 `_scan_fill_gaps` doesn't normalise before comparing

Fixed rather than commented, since it bills the operator: the raw-string comparison reported `tags: [Recovery]` beside a body reading `#recovery` as a backfillable gap, i.e. a paid `creek classify --method llm --force` run over the vault that changes nothing. The comparison now happens in the same normalised space the merge works in, via a new `tags_pass.has_unrecorded_tags()` — the predicate lives in `tags_pass` because normalisation is knowledge only that module has. Covered by `TestHasUnrecordedTags` (case-only and underscore differences are not gaps; a genuinely missing tag still is, so the fix doesn't mute the hint).

## One correction to my own docs, found while restating the rule

The docstrings claimed a recorded entry could fail to appear verbatim in "exactly two ways" (duplicate-spelling fold, or normalises-to-empty). That was itself an overstatement of the same kind you flagged — `merge(["Recovery"], []) == ["recovery"]` is a **third**: a lone non-canonical entry is re-spelled without being a duplicate of anything or normalising away. Preservation is of the *tag*, not of the bytes. All three cases are now stated accurately and pinned by a test.

## Not changed, tracked instead

`_MAX_TEXTURES` bounds LLM-derived data only. `link/temporal.py:106-112` adds `+0.1` per shared texture with **no clamp**, so a hand-authored twenty-tag list already contributes `+2.0` — trimming the record in the parser would never have fixed that consumer, it would only have deleted the operator's tags. Clamping the consumer is filed as **#1216**.

## Verification

- `./scripts/check-all.sh` → exit 0 (12/12 gates, 7660 passed, 94.17% coverage).
- 40,000 randomized property cases over both merges — prefix preservation, no lost recorded tags, growth bound, idempotent fixed point, canonical output — 0 failures.

Re-review please, @claude.

Geoffe-Ga:
## Summary
Gives `Fragment.tags` and `Fragment.emotional_texture` real producers for the first time: a new pure module `creek/classify/tags_pass.py` extracts precision-filtered `#hashtags` from fragment bodies (wired at the universal ingest chokepoint and in the classify engine), and a new `texture:` section rides inside the *existing* classification LLM call to populate `emotional_texture` with zero new provider calls. The PR includes a second commit that found and fixed a genuine data-loss bug in its own first-pass merge logic (a union that truncated hand-authored lists past the cap), with a full RCA and boundary-swept regression tests.

## Strengths
- **Precision-first hashtag extraction** is well-reasoned and empirically grounded: `_MIN_ALPHA_CHARS`/`_MAX_DIGIT_RUN` specifically target a measured false-positive class (Discord snowflakes, 688/539 occurrences in a 3000-file sample), with code-fence/inline-span stripping to avoid `#comment`/`#fff` false positives — verified against the actual regexes (`_TAG_RE`, `_CODE_FENCE_RE`, `_INLINE_CODE_RE`), all linear-time with no ReDoS risk.
- **Never-lose merge semantics are correct and now well-tested.** After the second commit's fix, `merge`/`_merge_textures` treat the cap as a *growth ceiling*, not a truncation of recorded data — verified in `tags_pass.py:1047-1101` and `parsing.py:377-426`, and swept from both sides of the boundary (`31/32/33`, `4/5/6`) rather than only at the equality case that hid the original bug.
- **SEC-004 injection boundary respected.** The new `texture:` top-level key is deliberately not `emotional_texture:`, keeping `_ALLOWED_TOP_LEVEL_KEYS` (the allow-list gating what an LLM response may write) never widened to a `Fragment` field name; `privacy_tier` and a bare `emotional_texture:` key both stay rejected, and this is pinned by explicit tests (`TestValidateTextureSection`).
- **Claims verified against the actual checked-out code, not just the PR body.** I independently confirmed all three cited `assemble_ingested_fragment` call sites (`pipeline.py:498`, `ingest/pipeline.py:309`, `discord.py:839`) exist, and that `pipeline.py` has no third `apply_tags`/praxis-style call, matching the "ingest chokepoint is sufficient" design claim.
- **Config-oracle discipline (#846/#848) honored.** `_hint_tags_backfill` reports a count only; tests explicitly assert no fragment id, body excerpt, or tag name leaks into stdout.
- **Backfill behavior is honest and correctly scoped**: `_write_tier_only` is deliberately not widened, so existing `llm`/`manual`-stamped fragments need an explicit `--force` to gain tags — consistent with the #877 praxis precedent this PR mirrors.
- Test coverage is exceptionally thorough: boundary sweeps on both caps, idempotency checks (`merge(merge(c,x),x) == merge(c,x)`), end-to-end CLI/vault round-trips, and consumer-side proofs (Tag Garden, temporal linking, wavelength texture cloud) that the previously-dead readers light up once the field is populated.

## Security Concerns
🟢 LOW — Hashtag/texture extraction only reads fragment bodies and LLM responses already in the pipeline's trust boundary; no new I/O, network, or eval surface. Regexes are linear-time and reviewed for ReDoS. The SEC-004 allow-list widening is minimal and correctly scoped (see Strengths). No blocking security issues found.

## Problems
None found. No 🔴 blockers.

## Code Quality
- `_TAG_RE`'s candidate character class (`[A-Za-z][A-Za-z0-9_/-]*`) is ASCII-only, so hashtags in non-Latin scripts or with accented characters are never extracted. Given this is a personal-knowledge vault with a documented multi-source ingest surface, this may quietly exclude a real (if currently unmeasured) class of legitimate tags. Worth a follow-up issue if the operator's vault has non-English content — not blocking, since the precision-over-recall stance is deliberate and documented, and there's no evidence in the cited corpus that this bites today.
- The module and function docstrings in `tags_pass.py` and `parsing.py` are unusually long (some individual docstrings exceed 40 lines) — consistent with this codebase's existing `praxis_pass.py` style and its ≥95% interrogate requirement, but worth being aware this sets the bar for future modules in this area.

## CLAUDE.md Adherence
✅ Conventional Commits followed (`feat(classify/ingest): ...`, `fix(classify): ...`).
✅ TDD workflow: PR description states RED-first verification against unfixed HEAD, and the second commit's fix includes its own RED-first regression tests.
✅ `./scripts/check-all.sh` reported green (12/12, 7628 tests, 94.11% coverage — both exceed the 90% branch-coverage gate).
✅ No new abstractions beyond scope; the design explicitly avoids a third pipeline call site where one isn't needed.
✅ Feature branch workflow (`issue/878-feat-...`), not a direct push to `main`.

## Testing
Extensive and well-targeted: precision/negative-case tables for the extractor, boundary-swept growth-ceiling tests for both `merge` and `_merge_textures` (the exact shape that would have caught the bug fixed in commit 2), idempotency tests, SEC-004 regression tests, CLI hint disclosure tests (asserting fragment content does *not* leak), and end-to-end consumer tests (Tag Garden, temporal linking, wavelength report) proving previously-dead code paths now activate. No gaps identified that would block merge.

## Documentation
Docstrings are thorough to the point of narrative (module docstrings in `tags_pass.py` and `parsing.py` read like design documents), consistent with this repo's `interrogate` ≥95% gate and the precedent set by `praxis_pass.py`. The PR description itself doubles as an unusually complete design rationale and test plan. No documentation gaps found.



## Verdict: LGTM

No bugs, security issues, or test gaps survive independent review: the two writer call sites and the pipeline.py non-wiring claim were verified directly against the checked-out source, the growth-ceiling/truncation defect the author's own second commit found and fixed is now covered by exhaustive boundary tests (31/32/33, 4/5/6), and the SEC-004 injection-boundary discipline (texture: not emotional_texture:) is correctly implemented and pinned. The only remark — ASCII-only hashtag matching — is a documented scope choice, not something I'd block on.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5224834787 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Blocker addressed in `894c3af`

You were right, and it was worse than the summary suggested. A 40-tag fragment did not merely get truncated — it lost 8 curated tags **and gained nothing**, because the candidates it was truncated in favour of never made it past the ceiling either. Pure loss, on the operator's own vault, by following this PR's own documentation.

### RED first

Verified against the unfixed tree, asserting on the retained set rather than a length:

```
assert result.tags == current
E  AssertionError: Right contains 8 more items, first extra item: 'tag-a32'

assert result.emotional_texture == recorded
E  AssertionError: Right contains 5 more items, first extra item: 'mood-5'
```

### The design question, answered

"Never lose a tag" and "respect the cap" cannot both hold once the recorded list is over the cap, so the cap changes meaning: it is a **growth ceiling on what an automated pass may add**, not a limit on the field's length. Formally, with `canon(xs)` = order-preserving dedup of normalised entries:

1. **Preservation is total.** `canon(current)` is always a *prefix* of the result, however long `current` is.
2. **Growth is bounded.** At most `max(0, CAP - len(canon(current)))` new entries are admitted — exactly that many when the candidates offer enough. A candidate already on record costs no room.

The corollary is what keeps the cap meaningful: an automated pass can never *grow* a list past the ceiling. A list can only *already* be past it, and then it gains nothing — so repeated passes are idempotent rather than a ratchet. Growth is bounded from both directions and is pinned by a parametrized test over `(existing, added)` = `(0,32) (1,31) (31,1) (32,0) (33,0) (40,0) (100,0)`, plus "10 existing + 200 candidates → exactly 32" against the **literal** 32.

**A fragment with 40 pre-existing tags after the fix:** keeps all 40, gains none, and `apply_tags` returns *the same object*, so nothing is rewritten to disk at all. It stays frozen at 40 until the operator prunes it by hand — documented as a known residual, because nothing here can tell which of the tags it did not write is the one worth evicting, so it evicts nothing.

I rejected the "refuse and error" option: this runs over 35,330 fragments in a batch pass, and a fragment an operator deliberately over-tagged is not an error state — erroring would turn their curation into a failed run.

### `emotional_texture` — same reasoning, stronger guarantee

Identical fix at cap 5. Preservation there is **verbatim**: no normalisation and no length check touch the recorded side, so a hand-authored 200-character texture tag survives byte-for-byte. This also resolves the contradiction you spotted with `test_a_hand_authored_ten_item_texture_list_is_not_truncated` — that test's promise is now actually kept by the write path, not just by the model.

### The test gap was the real lesson

You were right that `n == cap` without `n > cap` is the shape that hid this. I swept every bound this PR added for the same pattern and found **one more genuine instance**: `_MAX_TEXTURE_CHARS` was tested only by feeding 500 characters and asserting against `[:_MAX_TEXTURE_CHARS]` — an expectation computed from the very constant the code slices with, so an off-by-one to `[:31]` or `[:33]` would have passed the whole suite. It is now pinned at exactly 32 and at 33 against literals. `_MIN_ALPHA_CHARS`, `_MAX_DIGIT_RUN` and `_MAX_TAG_CHARS` were already two-sided. `_MIN_TAG_CHARS`' lower half is provably unreachable through `extract` (the regex forces a leading letter, `_has_enough_letters` then requires two, and `_normalise` never removes one) — I documented that rather than manufacturing a test for a dead branch.

### Both nits taken — they share the blocker's root cause

Both were the same mistake as the blocker: comparing recorded tags without applying the merge's own normalisation.

- `_scan_fill_gaps` used `set(extract(body)) - set(fragment.tags)`, so `tags: [Recovery]` beside a body reading `#recovery` reported a backfillable gap — and the remedy it names costs real money in tokens. It now asks a new `has_unrecorded_tags`, which compares in the same normalised space `merge` works in.
- `_record_tags` drops the unnecessary `.copy()`.

### Docstrings

Fixed the code so the guarantee holds; did not soften the docs. They now state the two-part invariant explicitly, and name the only ways a recorded entry can differ from what was typed — re-spelled, folded into an earlier entry sharing that spelling, or dropped as a separator that was never a tag — none of which is a lost tag.

One docstring claim I made *less* sweeping because it was not true: `_MAX_TEXTURES` claimed to protect the temporal-resonance consumer from an uncapped list. It bounds a runaway *response*, but `creek/link/temporal.py:106-112` adds `+0.1` per shared tag with no clamp anywhere in the module, so a hand-authored 20-tag list already contributes `+2.0` and always did. Trimming recorded tags would not have fixed that consumer — it would only have deleted the operator's tags. Filed as **#1216** and referenced from the constant rather than papered over.

### Gates

- **Gate 2:** `./scripts/check-all.sh` → **exit 0** on the pushed tree, all 12 checks, **7661 tests**, coverage **94.17%**, per-file gate green.
- **Gate 2.5:** `code-review-orchestrator` → **CLEAN**, no blockers. It independently re-verified that no existing test assertion was deleted or weakened (two renames and one assertion reformulated to add a literal-length check alongside the original).

Pushed as `894c3af`.

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
This PR gives `Fragment.tags` and `Fragment.emotional_texture` real producers for the first time — a new deterministic `tags_pass.py` hashtag extractor wired at the ingest chokepoint and the classify engine, plus a `texture:` section riding inside the existing LLM classification response (no new LLM calls). Both consumers were previously structurally dead (`tags: []` / `emotional_texture: []` on 2000/2000 sampled fragments of the operator's real vault), and the PR includes exhaustive regression tests proving the downstream readers (Tag Garden, temporal linking, wavelength texture cloud) now light up.

## Strengths
- **Exemplary TDD discipline and self-review.** The second commit fixes a genuine bug the author's own review caught: the initial union-merge truncated at the cap even when the *recorded* list already exceeded it, silently deleting hand-authored tags on every re-classify. The fix (cap as a growth ceiling, not a length limit) is correct, and both `tags_pass.merge` and `parsing._merge_textures` are now swept from both sides of the boundary (31/32/33, 4/5/6) with literal (not constant-derived) expectations so an off-by-one would actually fail.
- **Precision-first extraction design**, empirically justified against the real 35k-fragment vault (Discord snowflake rejection, code-fence/inline-code stripping, deliberate non-splitting of camelCase to avoid inventing tags).
- **Security boundary respected correctly**: the new `texture:` top-level key is deliberately not named after the `Fragment` field it writes, preserving the SEC-004 `_ALLOWED_TOP_LEVEL_KEYS` invariant that `privacy_tier` (and now `emotional_texture`) can't be smuggled in via prompt injection — and this is pinned by an explicit regression test.
- **Never-lose merge semantics are correctly total** on both sides (preservation of the canonical recorded prefix; bounded, capped growth) and idempotent under repeated application — verified by dedicated fixed-point tests (`merge(merge(c,x),x) == merge(c,x)`).
- **Honest backfill story**: the PR body and `creek fill` hint are explicit that this only populates new ingests, names the token cost of the remediation, and avoids the destructive `--method rules --force` footgun.
- Extremely thorough, well-organized tests across every touched module (tags_pass, parsing, classify_engine, cli fill hints, ingest, link scoring, wavelength reports/tracker, tag garden, models) — each test's docstring states the specific defect it guards against.

## Security Concerns
None found. 🟢 The SEC-004 injection boundary is correctly maintained (bare `emotional_texture:` and `privacy_tier:` stay rejected), the new regexes are all linear-time with no nested quantifiers over unbounded, attacker-influenced fragment bodies, and the `creek fill` hint deliberately reports counts only (verified by tests asserting fragment ids/bodies/tag names never leak to stdout).

## Problems
None found that block merge.

## Code Quality
- `tags_pass._MIN_TAG_CHARS`'s lower-bound branch in `_accept` is documented as currently unreachable dead code given the other filters (verified correct on inspection) — kept intentionally as a defensive guard. Not a real issue given how explicitly it's documented and tested for absence of harm, just flagging for awareness.
- The prompt/docstring/module documentation in this PR is unusually dense (borders on verbose for a codebase not already at this documentation density), but it's consistent with the existing `praxis_pass.py` precedent this PR explicitly mirrors, and none of it is filler — every paragraph pins a specific decision or rejected alternative.

## CLAUDE.md Adherence
- ✅ TDD: RED-first workflow evidenced by the PR description and by the second commit's bugfix (new failing tests added before the fix).
- ✅ Coverage: PR reports 94.11% (>=90% threshold), with per-file gate green.
- ✅ Commit conventions: Conventional Commits (`feat(...)`, `fix(...)`), feature branch, not direct-to-main.
- ✅ Patterns: mirrors the existing `praxis_pass.py` shape (issue #877) as prescribed by the codebase's own precedent.
- ⚠️ I could not independently re-run `./scripts/check-all.sh` or query live CI status in this sandbox (the `gh pr checks` / `gh api` calls required approval I didn't have), so CI-green is taken on the PR's own report rather than independently confirmed.

## Testing
Exceptional. New/changed tests span: the extraction table (positives/negatives/Discord-snowflake rejection/code-fence stripping), the merge growth-ceiling boundary swept from both sides, idempotency and fixed-point properties, the SEC-004 top-level-key regression, end-to-end `creek ingest` and `creek classify` CLI invocations asserted against the actual file written to the vault (not just in-memory objects), the `creek fill` hint's zero-leakage guarantee, and downstream consumer regression proofs (Tag Garden, temporal linking scoring, wavelength texture cloud). Insufficient testing is not a concern here.

## Documentation
Comprehensive module, function, and constant-level docstrings throughout, each grounded in the concrete defect being fixed and the empirical evidence from the operator's real vault. The module docstring's "known residuals" section is a good practice — it explicitly documents remaining known limitations (FEAT-023 re-atomized children tagged one pass late, human-looking Discord channel mentions surviving filters) rather than hiding them.



## Verdict: LGTM

No blocking bugs, security issues, or test gaps found after a full read of the diff; the merge/union logic is correctly total and idempotent (including at the growth-ceiling boundary), the SEC-004 injection boundary is preserved with a regression test, and test coverage is exceptionally thorough. The one genuine defect in the original approach (cap-as-truncation deleting hand-authored data) was already found and fixed by the author's own review in the second commit, before this review.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5224902937 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Gives the previously-dead `Fragment.tags` and `Fragment.emotional_texture` fields real producers: a new pure `tags_pass.py` hashtag extractor wired at the universal ingest chokepoint and in the classify engine, plus a `texture:` section riding inside the existing (no new call) LLM classification response. Ships as two commits — the first implements the feature, the second is a self-caught fix for a real truncation/data-loss bug the author found on their own re-review, plus a false-positive fix in the `creek fill` hint.

## Strengths
- **Root-caused with evidence**: empirically demonstrates the dead-field bug on a 35,330-fragment real vault (2000/2000 sampled `tags: []`) rather than asserting it.
- **Precision-first hashtag extraction**: explicitly engineered against Discord snowflake IDs (`#c762760820929069057` etc.) via a two-letter minimum and an 8-digit-run disqualifier, with code fences/inline spans stripped first. All regexes are linear-time and documented as such (relevant since they scan 35k+ attacker-influenced bodies).
- **SEC-004 boundary respected**: the LLM section is named `texture`, not `emotional_texture`, keeping `_ALLOWED_TOP_LEVEL_KEYS` un-widened to any `Fragment` field name — pinned by a dedicated regression test (`test_still_rejects_privacy_tier`) on a payload that also carries the newly-legal section.
- **Never-lose merge semantics, now provably correct**: commit 2 rewrote `tags_pass.merge` and `parsing._merge_textures` so the cap is a *growth ceiling* rather than a truncation — the recorded side is always a preserved prefix, verified with parametrized boundary sweeps (31/32/33 tags, 4/5/6 textures) using literal expected values so an off-by-one can't hide behind a constant-derived assertion.
- **Deliberate non-decisions are documented**: no model-level validator on `tags`/`emotional_texture` (would silently rewrite hand-edited Obsidian frontmatter on every read), no third `pipeline.py` wiring call (ingest chokepoint is provably sufficient), no case-splitting on camelCase tags (a guess that would fragment the vocabulary).
- **Backfill honesty**: `creek fill` only ever reports a count (never a fragment id, body excerpt, or the tags themselves — tested explicitly for leakage), and correctly avoids recommending the free-but-destructive `--method rules --force` path.
- **The self-correction commit is the strongest signal here**: the author found their own truncation bug ("40 recorded tags came back as 32, losing 8" — reachable through the exact `--force` remediation this PR tells operators to run) and the exact case-normalization false-positive in the fill hint, fixed both with RED-first tests, before this ever reached external review.

## Security Concerns
🟢 None found. The SEC-004 allow-list boundary is respected and regression-tested. Regex patterns are bounded/linear-time against unbounded attacker-influenced input, which the module docstring explicitly calls out. The tag/texture caps correctly bound only LLM-derived data, not operator-authored frontmatter (with `test_a_hand_authored_ten_item_texture_list_is_not_truncated` and friends pinning that as a deliberate non-decision rather than an oversight).

## Problems
None blocking. I traced every source hunk in both commits (`classify_engine.py`, `orchestrator.py`, `parsing.py`, `prompts.py`, `tags_pass.py`, `cli.py`, `ingest/base.py`, `creek_mcp/tools/classify.py`) and did not find a correctness issue that survived commit 2's fix.

## Code Quality
- Very dense docstrings throughout (multi-paragraph rationale on nearly every constant/function). Non-blocking — the repo's `interrogate` docstring-coverage gate (≥95%) rewards this, and the density pays for itself here given how many boundary conditions and unstated invariants this change touches — but it's worth the author being aware it's a lot to read in review, in case future PRs in this codebase want a lighter touch for less load-bearing code.
- `_scan_fill_gaps`'s `tags_backfillable` detector re-walks each fragment's body through `extract()` a second time (already paid for once inside `has_unrecorded_tags`, which calls `extract` internally) — effectively the same cost class as the rest of that function, not a new concern, just noting it's not free.

## CLAUDE.md Adherence
✅ Conventional Commit messages (`feat(classify/ingest): …`, `fix(classify): …`), each referencing its issue(s).
✅ TDD discipline: commit 2's message explicitly states "Verified RED first: the new texture cases fail against the previous implementation."
✅ Docstring coverage — every new module/function/constant is documented well past the 95% interrogate bar.
⚠️ I did not independently run `./scripts/check-all.sh`; per the PR's own CI status at review time, `Code Quality & Testing (3.11/3.12/3.13)` and `claude-review` were still `IN_PROGRESS`. Recommend confirming those go green before merge, though nothing in the diff suggests they won't.

## Testing
Exceptionally thorough — insufficient testing is the one thing that would block this, and it's the opposite of the case here:
- Boundary sweeps on both new caps from both sides (31/32/33 tags, 4/5/6 textures), asserting the recorded prefix rather than just a count, so "admitted nothing" is distinguishable from "deleted the tail."
- End-to-end coverage through real `creek ingest` CLI invocations reading back the actual written file, not just in-process objects.
- Negative controls throughout (disjoint textures add nothing, non-hashtag bodies stay empty, a flat-rate texture-bonus implementation would fail the scaled-overlap test).
- Idempotency explicitly tested (`merge(merge(c, x), x) == merge(c, x)`, a second classify run reports zero extractions).
- Leakage tests for the `creek fill` hint confirm it reports a count only and never a fragment id, tag value, or body excerpt.

## Documentation
Comprehensive module/function docstrings plus a detailed PR description covering the defect, the design rationale (precision-over-recall, the `texture` vs `emotional_texture` key naming, union-only merges), and an honest accounting of known residual gaps (re-atomized children tagged one pass late, human-looking Discord channel mentions like `#living-room` surviving the filter). Nothing incomplete or misleading found.



## Verdict: LGTM

No bugs, security gaps, or missing tests found after tracing every source-file hunk in both commits; the PR's own second commit already caught and fixed a genuine data-loss bug (truncating merge) and a false-positive fill-hint before this review, and boundary conditions (31/32/33, 4/5/6) are pinned with literal-valued tests. This is mergeable as-is.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5226327245 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1218, opened 2026-08-08
fix(cli): the report privacy-audit NB comment describes behaviour that does not ship — `--include-tier open` writes no entry


Geoffe-Ga:
## Symptom

`creek/cli.py` (the `NB (#968)` comment above `_audit_privacy_override_if_needed` in `report`) states:

> a typed `--include-tier open`, the narrowest scan there is, writes one

It does not. The audit is gated on `privacy_filter.override_elevates`, which returns `False` for **both** `None` and `PrivacyTierOverride.OPEN` by construction (`creek/classify/privacy_filter.py:123-125`). So on `report`:

| invocation | effective ceiling | audit entry |
|---|---|---|
| `report --type voice` | unfiltered (widest) | none |
| `report --type voice --include-tier open` | open (narrowest) | none |

The two are indistinguishable in `00-Creek-Meta/Privacy-Audit.jsonl`.

Pinned by `tests/test_cli_privacy.py::test_report_voice_include_tier_open_writes_no_audit` (added in #879), which asserts the behaviour that actually ships.

## Two things to decide

1. **The comment is wrong** and should be corrected regardless — it will mislead the next person who reads it while touching #968 semantics.
2. **The underlying gap is real**: on `report` the flag NARROWS, so an absent flag is the *widest* scan and a typed `open` is the narrowest. `override_elevates` encodes the opposite polarity (built for `mine`/`author`, where the flag widens). The audit therefore cannot tell "scanned unfiltered" from "narrowed to open" on this surface.

The original `NB (#968)` comment already explains why this was not fixed in place: auditing the *resolved* ceiling here fires before `--type` is validated, turning the #716 unknown-type exit-2 into an `OSError` on an unwritable vault path (`tests/test_cli.py::test_report_command`). The fix is a reordering plus an audit-volume decision, not a one-line swap.

## Acceptance criteria

- [ ] The `NB (#968)` comment in `creek/cli.py` describes what the code does
- [ ] A decision is recorded (fix or accept) on whether `report`'s audit log can distinguish an unfiltered scan from `--include-tier open`
- [ ] If fixed: `--type` validation still exits 2 before any audit write (`test_report_command` stays green)
- [ ] `./scripts/check-all.sh` passes

## Refs

- Comment introduced by #968
- Behaviour pinned by #879

Issue #1214, opened 2026-08-07
harden(voice): _is_safe_sample_stem is a denylist — Windows device names, NTFS streams, Unicode normalisation, symlinked register dirs


Geoffe-Ga:
## Context

#879 added `_is_safe_sample_stem` in `creek/generate/voice.py` because `Fragment.id` is
an unconstrained `str` sourced from third-party exports and gets interpolated into a
filename (`register_dir / f"{fragment.id}.md"`). Verified before the guard existed:
`id="../../escape"` wrote `<vault>/07-Voice/escape.md`, outside the register folder.

#879 closed the cases it could verify on this platform and with a test: path separators,
NUL, `.`/`..`/empty, and (added during review) an over-length bound and the `_Summary`
stem collision. The guard is a **denylist**, and the residue below is what a denylist
leaves. None of it is reachable on the macOS/Linux path the test suite runs, which is
exactly why it wants its own issue rather than an unverified patch.

## Residue

1. **Windows reserved device names.** `CON.md`, `NUL.md`, `COM1.md`, `LPT1.md` pass the
   guard. On Windows these open a device rather than a file.
2. **NTFS alternate data streams.** `Path("foo:bar.md").name` is believed to equal the
   input (only single-letter-plus-colon is treated as a drive), so an id containing `:`
   would pass and write to a hidden stream that `glob("*.md")` never enumerates — hence
   a sample the prune pass can never remove. **Unverified**; needs checking on Windows
   before fixing.
3. **Trailing dot or space.** `foo .md` / `foo..md` pass; Win32 strips the trailing
   character on write, so `path.stem` no longer round-trips to the id and the prune
   stops recognising the file as its own.
4. **macOS Unicode normalisation.** APFS/HFS+ normalise filenames, so an id written as
   NFC (`café`) reads back from `readdir` as NFD. `_prune_stale_copies` compares
   `candidate.stem` against ids taken from frontmatter, so a non-ASCII id would be
   rewritten every run and never pruned. **Unverified**; follows from documented APFS
   behaviour, but measure it before acting. Non-ASCII titles are entirely ordinary in a
   personal journal.
5. **Symlinked `register_dir`.** `mkdir(parents=True, exist_ok=True)` and `is_dir()` both
   follow symlinks, so a symlinked `Register-Samples/<register>` (or `07-Voice`) puts the
   copies — and the prune's `unlink` — outside the vault. Limited by `glob("*.md")` being
[...]

Issue #1211, opened 2026-08-07
fix(purge): right-to-be-forgotten misses 07-Voice derived artifacts — a purged fragment's body survives in its voice profile


Geoffe-Ga:
## Verified at unmodified HEAD (this is not caused by #879)

`creek purge fragment <id>` locates files through `PurgeEngine._list_fragment_files`
(`creek/purge/engine.py:952-961`), which is `01-Fragments` only. `_VAULT_CONTENT_FOLDERS`
(`engine.py:52-63`) *does* list `07-Voice`, but it is consumed only by `purge_vault`
(`engine.py:532`) — never by the targeted `purge_fragment` / `purge_source` /
`purge_daterange`.

Reproduced on a stashed-clean HEAD checkout: seed two conviction-confidence fragments
whose bodies carry a unique marker, run `creek report --type voice`, then
`creek purge fragment frag-a --yes`. Files still containing the purged body afterwards:

```
01-Fragments/Journal/frag-b.md      <- correct, different fragment
07-Voice/analytical-profile.md      <- PURGED FRAGMENT'S BODY, still on disk
```

`### Sample Passages` in a voice profile is the fragment's body, and purge never touches
it. The same applies to `07-Voice/Lexicon/glossary.md`, whose `_build_borrowed_terms`
records "the whole surrounding sentence verbatim" (stated in
`creek/generate/lexicon.py`'s own `generate_lexicon` docstring).

For a vault holding intimate journal content, right-to-be-forgotten that leaves the body
in a derived artifact is not right-to-be-forgotten. The audit entry compounds it: the
purge reports `Fragments affected: 1` and lists its deleted files, so the operator is
told the erasure succeeded.

## What #879 adds to this

#879 wires `save_exemplars` so `07-Voice/Register-Samples/<register>/<id>.md` is
populated with `shutil.copy2` byte-for-byte copies. That is one more file in the same
already-missed directory — a full copy rather than an excerpt.

#879 handles the part it created: its prune pass is manifest-driven, so a re-run of
`creek report --type voice` (or `creek fill`) does remove the orphaned copy. That is a
mitigation, not a fix — it is deferred to whenever the operator next happens to run a
report, whereas purge is synchronous and audited.

Note the trap #879 had to design around, because any fix here must too:
`_scrub_references` (`engine.py:1131`) walks `vault_path.rglob("*.md")` and rewrites the
purged id everywhere, including the derived copy's own `id:` frontmatter line, which
[...]

Issue #1204, opened 2026-08-07
feat(mcp): decide whether MCP report_type=voice persists 07-Voice/Register-Samples (CLI/MCP divergence from #879)


Geoffe-Ga:
## Context

#879 wires `VoiceExemplarCollector.save_exemplars` into a production call site so
`07-Voice/Register-Samples/<register>/` finally gets populated. It does so via a new
module-level `generate_register_samples(vault_path, *, override)` in
`creek/generate/voice.py`, called from the CLI's `_report_voice`
(`creek/cli.py:2580`) and from `creek fill`'s report sequence.

## The gap

`creek_mcp/tools/report.py`'s `report_tool` does **not** route through the CLI handler —
for `report_type="voice"` it calls `VoiceProfileGenerator(override=override).generate_all_profiles(vault_path)`
directly (`creek_mcp/tools/report.py:129-132`). So after #879 lands:

- `creek report --type voice` writes profiles **and** register samples.
- MCP `creek.report` with `report_type="voice"` writes profiles **only**.

Two surfaces, one report type, different artifacts. That divergence was a deliberate,
recorded scoping decision in #879 (the issue body names the CLI and `creek fill`, not
MCP), not an oversight — but it should not be permanent.

## What adopting it needs (why it is its own issue)

1. **Write volume from a remote consumer.** MCP's default ceiling is `TierCeiling.OPEN`
   and the consumer may be remote. Copying up to `7 × DEFAULT_MAX_PER_REGISTER` fragment
   files into the vault on a remote tool call is a different risk posture from an operator
   typing `creek report`. Decide whether it needs its own opt-in.
2. **`created_path` semantics.** `report_tool` audits a single `created_path`
   (`relative_paths[0]`). With samples added, "the first written path" becomes arbitrary.
   The audit record shape needs a deliberate answer, not an accidental one.
3. **Ceiling stamping.** #879 stamps each `_Summary.md` with the `tier_ceiling` the run
   was taken under, because a narrow-ceiling run surveys a smaller corpus and prunes
   samples a wider run produced. An MCP caller at `OPEN` would therefore prune an
   operator's unfiltered samples. That interaction needs a decision before wiring.

## Acceptance criteria

- [ ] A decision recorded (in the code or an ADR) on whether MCP `report_type="voice"`
      persists register samples, and under what ceiling/opt-in conditions.
- [ ] If adopted: `report_tool` calls `generate_register_samples(vault_path, override=override)`,
[...]

PR #1219, opened 2026-08-08
fix(voice): give save_exemplars a caller so Register-Samples is populated

### LINKED ISSUE #879: fix(voice): persist_exemplars has no callers — 07-Voice/Register-Samples is never populated
## Role

## Goal

Wire `VoiceExemplarCollector.persist_exemplars` into a CLI path so `07-Voice/Register-Samples/<register>/` actually gets populated.

## Context

- creek/generate/voice.py implements exemplar persistence: `persist_exemplars` copies the top-ranked fragments per register into `07-Voice/Register-Samples/<register>/` plus a `_Summary.md` (docstring at :552, `_SAMPLES_SUBPATH` at :78).
- **No production code calls it** — `grep -rn "persist_exemplars" creek/` returns only the definition; callers exist only in tests.
- `creek report --type voice` (`generate_all_profiles`) streams exemplars through per-register JSONL accumulators to build the profile notes, but never persists the exemplar files themselves.
- Evidence (demo vault `creek-demo-2026-06-07`): all 7 per-register profiles and rhetorical-pattern notes exist (so exemplar collection works), yet `07-Voice/Register-Samples/` contains only `.gitkeep` after every creek command has been run.
- The samples folder is the documented training corpus for the voice proxy (module docstring, lines 5–9) — closed issues #47/#48 built the collector/extractor but the persistence step was never surfaced in any command.

## Proposed Solution

- Add persistence to `creek report --type voice` (preferred: one command produces profiles + samples), or add a dedicated `report --type register-samples` if the write volume warrants opting in. Either way, `creek fill`'s report sequence should include it once wired.
- Respect the collector's existing privacy gate (intimate excluded unless `include_intimate`), and honour the `--include-tier` override audit pattern used by the other report types.
- Idempotent re-runs: replace the register folder contents deterministically (stable filenames), don't accumulate duplicates.

## Acceptance Criteria

- [ ] Running the wired command on a fixture vault with settled/conviction-confidence fragments writes ranked exemplars + `_Summary.md` under `07-Voice/Register-Samples/<register>/`
- [ ] Intimate-tier fragments are excluded by default; `--include-tier` override is audited like the other report flows
- [ ] Re-running produces the same set (no duplicate accumulation)

## Constraints

- No LLM calls — this is deterministic collection/copying
- Follow stay-green TDD

Geoffe-Ga:
## Summary

`07-Voice/Register-Samples/` was never populated. `VoiceExemplarCollector.save_exemplars`
(`creek/generate/voice.py:604` at HEAD) had **zero production callers** — `grep -rn
save_exemplars creek-tools/` returned only `tests/test_voice_exemplars.py`. The
*collection* side was already wired (`lexicon.py:787` calls `collect_all_exemplars`), so
the pipeline ranked exemplars and then discarded them. This is a missing persistence
call, not a missing feature.

(The issue title names `persist_exemplars`; no such symbol has ever existed —
`git log -S persist_exemplars --all` is empty. The real method is `save_exemplars`.)

### The tier question, answered first

`save_exemplars` **copies fragment files into the vault byte for byte** (`shutil.copy2`),
so anything that can read `07-Voice/Register-Samples/` can read whatever lands there.
The question this PR had to answer before wiring anything: *can an `intimate` fragment
land there?*

**No — and the barrier is not the one you would guess.**

- The ceiling gate `within_ceiling(raw, self.override)` (`voice.py:609`) is a **no-op on
  the default path**. Both `creek fill` and a bare `creek report --type voice` run at
  `PrivacyTierOverride.ALL`, which admits everything. The ceiling only bites when the
  operator *narrows* with `--include-tier`.
- The gate that actually excludes `intimate` is the **consent** gate:
  `_eligible_register` (`voice.py:317`) rejects `intimate` whenever `allow_intimate` is
  false. That is the constructor default, it is not CLI-exposed, and
  `grep -rn "allow_intimate=True" creek/ creek_mcp/` finds only docstrings and skill
  markdown — never a production call site.

Verified by running it, not by reading it. With one `open` and one `intimate` fragment,
at the **widest** ceiling available:

```
ceiling=(absent = ALL)              exit=0  files=['_Summary.md', 'f-open.md']
   intimate body leaked? False | id leaked? False
ceiling=--include-tier intimate     exit=0  files=['_Summary.md', 'f-open.md']
   intimate body leaked? False | id leaked? False
```

Both gates predate this PR and live in `collect_exemplars`; they were never the gap. But
note the structural shape of it: **`save_exemplars` performs no tier check of its own.**
It persists whatever bucket dict it is handed. Tier safety is a property of the *caller*
pairing it with `collect_exemplars` on the same collector instance. That is why the entry
point is a single function (`generate_register_samples`) rather than the CLI doing
collect-then-save inline, and why `generate_register_samples` is now listed in
`tests/test_mcp_report_tier_ceiling.py::_OVERRIDE_CALL_NAMES`, which fails the build if
any production caller leaves the override at its default.

No fourth tier reader was added. The ceiling is read through
`creek/classify/privacy_filter.within_ceiling` (#1079 stays a two-reader problem).

### What changed

- **`generate_register_samples(vault_path, *, override, on_prune)`** in
  `creek/generate/voice.py` — mirrors `generate_lexicon`: **one** collector,
  `collect_exemplars` then `save_exemplars` on that same instance. The single instance is
  load-bearing, not stylistic: `_persist_fragment` copies the source file only while the
  collector's `_records` cache is populated, and only `collect_exemplars` populates it.
  Saving through a second collector produces a full set of valid-looking exemplar notes
  with **empty bodies**. `test_report_voice_register_samples_carry_the_source_body`
  asserts byte equality specifically to catch that.
- **`_report_voice` in `creek/cli.py`** now writes both halves of the voice report:
  profiles *and* register samples.
- **`creek fill` gains a `report/voice` step**, between `report/mode-profiles` and
  `report/wavelength`, stating `PrivacyTierOverride.ALL` explicitly like every other step.
- **A prune pass.** Persisting alone only ever adds, so without it a fragment that drops
  out of the top 20, is re-tiered, or is deleted from the vault leaves a verbatim copy of
  its body in the samples tree forever. Deletion is gated on an `exemplar_digests`
  manifest in each `_Summary.md` — the collector deletes only files it has a *record* of
  writing, never files matching a *shape*. An operator's own notes in that folder,
  including a fragment they hand-curated into it, survive. Because `creek fill` runs this
  unattended, deletions are reported (`Stale register samples pruned (N)`).
- **A filename guard** (`_is_safe_sample_stem`). `Fragment.id` is an unconstrained `str`
  carried over from an export and is composed straight into a path, so
  `id="../../escape"` wrote outside the register folder and a 300-char id raised
  `ENAMETOOLONG` out of `shutil.copy2` and took the command down. Length is measured in
  **bytes**, because `NAME_MAX` is.

### Re-running: what it clobbers

`Register-Samples/` is a **generated, read-only** folder. `_persist_fragment` calls
`shutil.copy2` unconditionally, so an operator edit to a sample copy `<fragment-id>.md`
is overwritten on the next run. Verified:

```
run1: 0
run2: 0   (after writing "OPERATOR EDIT — please keep me" into the copy)
operator edit survived? False
```

Operator files under *other* names are spared by the manifest, and are covered by
`test_report_voice_pruning_spares_operator_notes`. Documented in `docs/generation.md`:
edit the source fragment in `01-Fragments/`, never the copy.

### What an existing-vault operator has to do

**Forward-fill only, but no migration is needed.** Collection reads only frontmatter that
`creek classify` already writes, so an existing vault gets its `Register-Samples/`
populated the next time it runs either:

```bash
creek report --type voice --vault <vault>   # the direct command
creek fill --vault <vault>                  # now includes report/voice
```

Nothing backfills on its own — an operator who never runs either command still has an
empty folder.

### Deliberately out of scope, filed rather than done

Per #1011, with priority labels:

- **#1211 (P1)** — `creek purge` does not remove a purged fragment's body from `07-Voice/`
  derived artifacts. Predates this PR: purge already leaves a purged body in
  `<register>-profile.md` on an unmodified vault. This PR's prune clears an orphaned
  *sample copy* on the next run, which mitigates one artifact; it does not fix purge.
- **#1204 (P2)** — the `report_type="voice"` MCP tool still writes profiles only. Recorded
  divergence, not an oversight.
- **#1214 (P3)** — `_is_safe_sample_stem` is a denylist; Windows device names, NTFS
  alternate data streams, Unicode-normalisation collisions, and symlinked register dirs
  are not handled.
- **#1218 (P2)** — `report`'s privacy audit cannot tell an unfiltered scan from
  `--include-tier open`; `override_elevates` is False for both. The *comment* claiming
  otherwise is corrected in this PR; the polarity decision is not.
- **#1220 (P2)** — `_report_voice` walks `01-Fragments/` twice. Measured, not guessed:
  1.8x one corpus parse (1k and 5k synthetic vaults, linear; ~16s of a projected ~39s at
  35k), which is ~7% of `creek fill`'s parse budget — it already makes ~12 full-corpus
  passes, and its wall clock belongs to embedding inference and DBSCAN. Not shared here
  because a collector that never walked writes valid-looking samples with empty bodies.

## Test plan

RED evidence at unfixed HEAD (production files reverted, tests kept):

```
15 failed, 1 passed
tests/test_cli.py::test_report_voice_writes_register_samples
E   AssertionError: assert False
E    +  where False = is_file()
E    +    where is_file = (PosixPath('.../vault/07-Voice/Register-Samples/confessional') / '_Summary.md').is_file
```

With the fix: `16 passed`.

- `tests/test_cli.py` — 16 end-to-end tests through the real CLI: population, verbatim
  body, intimate exclusion, ceiling honouring, fail-closed on an absent `privacy_tier`
  key, idempotence, five prune cases, long-id survival, and `tier_ceiling` stamping.
- `tests/test_voice_exemplars.py` — filename-safety unit tests (traversal, `_Summary`
  collision, byte-measured `NAME_MAX` boundary) and `generate_register_samples` contract.
- `tests/test_cli_fill.py` — `report/voice` present, in order, stating `ALL`.
- `tests/test_cli_privacy.py` — audit-log polarity for `report.voice`.
- `tests/test_mcp_report_tier_ceiling.py` — `generate_register_samples` added to the
  override-required guard.

Added in the second pass (see below):

- `tests/test_cli.py::test_report_voice_narrowed_rerun_retracts_a_wider_runs_sample`
- `tests/test_voice_exemplars.py::TestManifestCorruptionIsFailSafe` — 4 tests

Gate 2: `cd creek-tools && ./scripts/check-all.sh` → **exit 0**, 12/12 checks, 7942
passed, coverage 94.09% (aggregate gate 90%), per-file gate 242 files ≥80%.

Gate 2.5: `code-review-orchestrator` → **FIX_REQUIRED**, both blockers fixed.

Not run by me, by design: CI (Gate 3) and review (Gate 4).

## Second pass: what Gate 2.5 sent back

The review did **not** come back clean. Two blockers, both test-coverage gaps on the
prune — the half of this change that *deletes* vault content from a step `creek fill`
runs unattended. Both mechanisms traced correct by hand; neither was pinned:

- **Retraction was untested.** `docs/generation.md` sells a narrower `--include-tier` as
  the only way to retract above-ceiling copies a wider run wrote, and nothing exercised
  it. The existing ceiling tests prove only that a narrow ceiling excludes on *write*,
  starting from an empty folder. Retraction is a different path: the copy is already on
  disk, the run never sees the fragment at all, and the only thing entitling the delete
  is the digest the *wider* run recorded. Break that manifest round-trip and the
  above-ceiling copy becomes permanent while every write-side test stays green.
- **`_read_sample_manifest`'s fail-safe branches were untested.** Both answer "I cannot
  tell whether I wrote this", and the contract is that this means delete nothing.
  Untested, a refactor that lets the read raise takes down the unattended `fill` step,
  and one that drops the type check turns a hand-edited note into a deletion list. The
  new tests include a **positive control**, so "the copy survived" cannot pass on an
  implementation that never prunes, and a mapping-shaped manifest — because the scalar
  case alone does not kill the `isinstance` mutant: iterating a string yields characters
  that match no digest.

The architect pass raised six more. Acted on; none behavioural except the first:

- prune now logs **each deleted path**, not just a count. WARNING rather than INFO
  because creek installs no logging handlers, so INFO is swallowed by
  `logging.lastResort` and an unattended deletion would leave no record of which file it
  took.
- `_sample_digest`'s docstring no longer claims the manifest gives purge nothing to
  erase. Unsalted SHA-256 over enumerable export ids is confirmable by recomputation;
  the digest buys stability across purge's `.md` rewrite, not erasure. Named as the
  trade it is.
- documented that **a narrowed run is not a durable redaction** — `creek fill` has no
  `--include-tier` and re-copies what the narrow run retracted. This was the most
  misleading thing in the docs.
- documented the `exemplar_digests` properties-panel noise, and that idempotence is
  "same set", not byte-identical (`generated_at` restamps every run).
- **declined**: a `prune=False` opt-out on `save_exemplars`. No caller needs it, and an
  unused kwarg is dead code the vulture gate would flag.

Closes #879


Geoffe-Ga:
## Summary
Wires the previously-orphaned `VoiceExemplarCollector.save_exemplars` into a new `generate_register_samples` entry point, called from `creek report --type voice` (and now `creek fill`), so `07-Voice/Register-Samples/` finally gets populated. Along the way it adds a stale-copy prune (keyed off a purge-resistant SHA-256 manifest in `_Summary.md`) and a filename-safety guard against path traversal / `ENAMETOOLONG` from an unconstrained `Fragment.id`.

## Strengths
- Root-caused correctly: confirmed via `grep` that `save_exemplars` had zero production callers, rather than assuming a "missing feature."
- Explicitly investigated and ruled out the scarier hypothesis (an intimate-tier leak) before touching code, with a runnable before/after repro pasted into the description.
- The single-collector-instance requirement (`_records` cache populated only by `collect_exemplars`) is easy to silently violate and is guarded by both a doc comment and a byte-equality test (`test_report_voice_register_samples_carry_the_source_body`).
- Prune-by-manifest (not by file shape) correctly avoids two failure modes a shape-based heuristic would hit: deleting an operator's hand-curated fragment, and failing to delete a copy whose `id` `creek purge` has scrubbed to `[purged]`. Both are pinned by dedicated tests.
- `_is_safe_sample_stem` measures length in encoded bytes (not `len()`), which is exactly right for `NAME_MAX`, and is tested at the exact byte boundary with a non-ASCII character.
- Deliberately-deferred scope (#1211 purge/voice sync, #1204 MCP divergence, #1214 Windows/Unicode edge cases) is filed as tracked issues rather than silently left as gaps.
- `docs/generation.md` update is detailed and matches the shipped behavior (read-only folder, prune semantics, re-run clobbering, migration story).

## Security Concerns
🟢 LOW — none blocking.
- The two privacy gates that keep `intimate` content out of the samples tree (`within_ceiling` ceiling gate, `allow_intimate` consent gate) are unchanged by this PR and are re-verified end-to-end by `test_report_voice_register_samples_exclude_intimate_by_default` and the ceiling/fail-closed tests.
- The new filename guard (`_is_safe_sample_stem`) closes a real path-traversal / DoS hazard (`id="../../escape"`, `ENAMETOOLONG` crashing the whole command) introduced by treating `Fragment.id` as a bare filename component — correctly scoped to skip only the one bad exemplar rather than aborting the run.
- Digest-based manifest (rather than storing raw ids) is a sound choice given `creek purge` rewrites arbitrary `.md` files in place.

## Problems
None found that block merge.

## Code Quality
- Minor: `test_report_voice_register_samples_carry_the_source_body` in `tests/test_cli.py` has a redundant local `import frontmatter` even though the module already imports it at the top of the file (also added in this diff). Harmless, but worth dropping in a follow-up cleanup.
- The `_report_voice` CLI function's control flow (prune message printed unconditionally before the early-return on "no profiles", but "samples written" printed only after that early return) is a little asymmetric — but it's intentional and explicitly justified in the docstring/comments ("announced before the early return, because the emptied corpus... is exactly the run with the most to delete"), and is covered by tests, so this is a design note rather than a defect.

## CLAUDE.md Adherence
✅ Conventional Commit-style PR title (`fix(voice): ...`).
✅ TDD workflow followed per the PR description (RED evidence pasted: 15 failed/1 passed at unfixed HEAD, 16 passed with the fix).
✅ Functions stay small and single-purpose (`_save_register`, `_prune_stale_copies`, `_rewrite_emptied_summary` split out rather than one large method), consistent with the ≤10 cyclomatic complexity target.
✅ All new public/private functions carry Google-style docstrings with Args/Returns, consistent with the ≥95% interrogate threshold.
⚠️ I could not independently execute `./scripts/check-all.sh` or `gh pr checks` in this review sandbox (tool restrictions blocked local Python/`gh api` execution), so CI-green and coverage-threshold compliance are taken on the PR description's stated results (`check-all.sh` exit 0, 16/16 passing) rather than independently re-verified — worth the merger confirming the CI tab is actually green before merge.

## Testing
Strong, and unusually self-aware about what each test does and doesn't catch:
- End-to-end CLI coverage for the headline defect, verbatim-body copying, intimate exclusion, tier-ceiling honoring, fail-closed-on-missing-tier, idempotence, five distinct prune scenarios (dropped fragment, re-tiered-to-intimate, operator notes spared, curated-fragment spared, purge-scrubbed-id pruned), long-id survival, and tier-ceiling stamping.
- The "second pass: security-review findings" block in `test_cli.py` is a good practice — it explicitly documents what an earlier (shape-based) implementation would have gotten wrong and why the manifest-based fix is needed, rather than just asserting the final state.
- `test_mcp_report_tier_ceiling.py` extends the structural guard ensuring every production caller of the new privacy-sensitive function states its override explicitly — good defense against a future regression reintroducing an implicit default.

## Documentation
Thorough. `docs/generation.md` covers the new artifact's write path, read-only/re-run-clobbers-edits behavior, automatic pruning and its manifest mechanism, the `--include-tier` narrowing caveat (with the `tier_ceiling` stamp as the mitigation), the no-migration-needed story for existing vaults, and explicitly cross-references the three deliberately-deferred issues rather than leaving them undocumented.



## Verdict: LGTM

The fix is narrowly scoped, the security-sensitive parts (path traversal, NAME_MAX, tier-ceiling/consent gates, purge-resistant prune manifest) are correctly implemented and independently verified by name-specific tests, and the test suite is unusually thorough (CLI e2e, unit, fill-step ordering, audit-log polarity, and a documented second security-review pass). I found no bugs and only a single trivial style nit that doesn't warrant blocking.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5224858715 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Summary
This PR wires `VoiceExemplarCollector.save_exemplars` to a real production caller (`generate_register_samples`), so `creek report --type voice` and `creek fill` now populate `07-Voice/Register-Samples/`, which previously sat permanently empty despite being fully implemented and tested in isolation. It adds a manifest-gated prune pass for stale copies, a filename-safety guard for `Fragment.id`-as-path, and a `tier_ceiling` stamp on the summary note, plus a large, TDD-evidenced test suite (RED at unfixed HEAD, 16 passing after the fix) across `test_cli.py`, `test_voice_exemplars.py`, `test_cli_fill.py`, `test_cli_privacy.py`, and `test_mcp_report_tier_ceiling.py`.

## Strengths
- Verified the tier-safety question empirically (script output pasted in the PR body) rather than just asserting it: the ceiling gate is a no-op on the default path, the real gate is the `allow_intimate` consent default, and neither leaks an `intimate` fragment at the widest available ceiling.
- The single-collector-instance requirement (`collect_exemplars` then `save_exemplars` on the *same* instance) is subtle and easy to silently violate; it's called out explicitly, pinned by `test_report_voice_register_samples_carry_the_source_body`, and enforced structurally via `_OVERRIDE_CALL_NAMES` in `test_mcp_report_tier_ceiling.py`.
- Deletion safety is manifest-based (`exemplar_digests` in `_Summary.md`), not shape-based — the second commit explicitly documents that the first pass's shape-based heuristic would delete operator-curated files and be defeated by `creek purge`'s id-scrub, and both regressions are pinned (`test_report_voice_pruning_spares_a_curated_fragment_copy`, `test_report_voice_prunes_a_copy_whose_id_purge_has_scrubbed`).
- `_read_sample_manifest`'s two fail-safe branches (unreadable YAML, non-list value) are each independently tested with a positive control, so "the copy survived" can't pass vacuously on an implementation that never prunes.
- Filename guard (`_is_safe_sample_stem`) covers traversal, emptiness, byte-measured `NAME_MAX` (with a non-ASCII boundary test proving it's bytes not characters), and collision with `_Summary.md` — and a rejected id degrades to "skip this one exemplar," not a command-killing `OSError`.
- Deliberately out-of-scope items (purge/07-Voice sync #1211, MCP divergence #1204, filename denylist hardening #1214, audit-polarity inversion #1218, double-walk cost #1220) are each filed with priority and reasoning rather than silently deferred or hidden, and the pre-existing `report`'s audit-polarity bug is corrected in the code comment rather than left misleading.
- `docs/generation.md` is updated with the operationally important caveats (folder is generated/read-only, retraction doesn't survive `creek fill`, no migration needed).

## Security Concerns
🟢 LOW — `_sample_digest` is unsalted SHA-256 over an enumerable, low-entropy value (`Fragment.id`), so it's confirmable-by-recomputation rather than truly anonymizing. This is explicitly flagged in the docstring as "a stability property, not an erasure guarantee," with a clear rationale (plaintext would be worse because `creek purge` rewrites all `.md` files) and a named upgrade path (HMAC). No action needed; documented trade-off, not an oversight.

No blocking security issues. The intimate-tier consent gate and tier-ceiling gate are both verified end-to-end in tests, and the filename guard closes a real path-traversal / `ENAMETOOLONG`-crash vector on untrusted `Fragment.id` data.

## Problems
None found that block merge.

## Code Quality
- `_is_safe_sample_stem`'s `_Summary` collision check is an exact, case-sensitive string match. On a case-insensitive filesystem (default macOS/APFS, Windows/NTFS — both mentioned elsewhere in the same docstring as target filesystems), a fragment id of `_summary` (lowercase) would still collide with `_Summary.md` at the OS level even though the guard doesn't reject it. This is narrow (requires a very specific id) and arguably falls under the same bucket as the already-deferred #1214 filename-hardening gap, so it doesn't need to block this PR — just worth a one-line mention in that issue if it isn't already covered by "NTFS/Unicode-normalisation collisions."
- The redundant double walk of `01-Fragments` (~1.8x parse cost) is a real cost accepted with measured numbers and tracked as #1220 — reasonable given the alternative (a second, walk-less collector) silently produces empty-bodied exemplars.

## CLAUDE.md Adherence
✅ TDD workflow followed and evidenced in the PR description (RED failure output at unfixed HEAD, 16 passed after fix).
✅ PR description states Gate 2 (`check-all.sh`) passed and Gate 2.5 (code-review-orchestrator) clean; a second commit shows genuine iteration on reviewer findings (retraction path, manifest-corruption fail-safety) rather than a rubber-stamp pass.
✅ Commit messages are detailed and explain "why," matching repo convention.
ℹ️ CI (`Code Quality & Testing` 3.11/3.12/3.13, `claude-review`) was still `IN_PROGRESS` at review time; `CrawDad Quality` jobs had already passed. Not something this review can resolve, but worth confirming green before merge per the repo's 4-gate "Stay Green" process.

## Testing
Extensive and well-targeted: 16 new CLI-level end-to-end tests cover population, verbatim-body copying (not empty-body fallback), intimate exclusion, ceiling honoring with a fail-closed absent-tier case, idempotence, five distinct prune scenarios (dropped-from-corpus, re-tiered-to-intimate, operator-note preservation, curated-fragment preservation, purge-id-scrub survival), long-id survival, and `tier_ceiling` stamping at both ends (unfiltered and narrowed). Unit tests separately pin filename-safety boundaries and manifest fail-safety with positive controls. This meets and exceeds the bar for a change that both writes and deletes vault content unattended.

## Documentation
`docs/generation.md` gets a thorough new section covering the two artifacts written, the read-only/regenerated nature of the folder, prune behavior and its manifest basis, the narrowed-ceiling retraction caveat (including that it doesn't survive `creek fill`), idempotence semantics, the intimate-tier exclusion, and the two known/tracked gaps (purge sync, MCP divergence). Docstrings throughout `voice.py` are dense and explain non-obvious "why" reasoning (e.g., why the manifest stores digests not ids, why the walk isn't shared, why the summary's `tier_ceiling` matters). No documentation gaps found relative to the implementation.



## Verdict: LGTM

No blocking bugs, security issues, or test gaps found after a careful read of the diff and cross-checking the pruning/manifest/eligibility logic against the checked-out PR source; the change is exceptionally well-tested (RED-evidenced TDD, 16+ new CLI tests including five prune scenarios and manifest fail-safety with positive controls) and honestly scopes out remaining gaps into separately tracked, prioritized issues rather than hiding them. The one code-quality note (case-insensitive `_summary`/`_Summary` collision on APFS/NTFS) is narrow enough and close enough to the already-deferred #1214 filename-hardening bucket that it doesn't rise to a blocking or even a must-address suggestion — this is mergeable as-is.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5224891455 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #881, opened 2026-07-22
fix(draft): compile-gap checker ignores linker-written pages (type: eddy/thread) and aliases — false 'missing' gap entries


Geoffe-Ga:
## Role

You are a pipeline engineer working in `creek-tools/` (Python, mypy strict, TDD per `creek-tools/CLAUDE.md`).

## Goal

Make the compile-gap checker recognise linker-written eddy/thread pages so `draft`/`mine` stop logging false "missing" gaps for pages that exist.

## Context

- `creek/generate/compile_routing.py::_load_pages` indexes only pages whose frontmatter has `type: compiled_page` (plus matching `target_kind`).
- The eddy/thread linkers write pages with `type: eddy` / `type: thread` (e.g. demo vault `03-Eddies/2023-03-30-Messages.md` has `type: eddy`, `aliases: [Messages]`).
- Result: every linker-produced page is invisible to `CompiledPageIndex`, so draft/mine log spurious gaps. Demo-vault `00-Creek-Meta/Processing-Log/compile-gaps.jsonl` (2026-07-02) records `"reason": "missing"` for eddies `[[Turn Apple Watch]]`, `[[Messages]]`, `[[Daily]]`, `[[Purple Red Blue]]` — all of which exist in `03-Eddies/` with exactly those aliases.
- Lookup should also resolve through `aliases`, not just `target_id`/filename — the wiki-links written into fragment frontmatter use the alias form (see #730, which added the aliases).

## Proposed Solution

1. In `_load_pages`, accept the linker page types for the matching kind (`type: eddy` for eddies, `type: thread` for threads) alongside `type: compiled_page`, mapping their `id`/`title`/`aliases` into the index.
2. Index pages under their `aliases` (and title) in addition to `target_id`, so `[[Messages]]`-style references resolve.
3. Keep logging genuine gaps: a wiki-link that resolves to no page under any of id/title/alias still appends to `compile-gaps.jsonl`.

## Acceptance Criteria

- [ ] A fixture vault with a linker-written eddy (`type: eddy`, `aliases: [X]`) yields no `missing` gap entry when draft references `[[X]]`
- [ ] Compile-written pages (`type: compiled_page`) continue to resolve exactly as before
- [ ] Genuinely absent pages still log gaps
- [ ] `./scripts/check-all.sh` passes

## Constraints

- Do not change what the linkers write (page schema churn is out of scope; #730 settled the alias convention)
- Follow stay-green TDD

Issue #730, opened 2026-06-29
feat(threads/eddies): resolve bulk wiki-links to date-prefixed page filenames


Geoffe-Ga:
## Context
Surfaced during #718 (bulk thread-page generator, PR #729).

The bulk linkers (`creek link --method threads` and `--method eddies`) write each page with a `{date}-{title}` filename via the shared `VaultWriter` (e.g. `02-Threads/Active/2026-06-25-Thread-Member.md`), while the wiki-links written onto member fragments are the bare `[[Thread Member]]` / `[[Eddy Title]]` (from `assign_fragments_to_threads` / `assign_fragments_to_eddies`).

In Obsidian's default link resolution, `[[Thread Member]]` does **not** byte-match `2026-06-25-Thread-Member.md` — so the links render unresolved unless the vault is configured for alias/heading resolution. This is a **systemic, pre-existing** behaviour shared by both the threads and eddies paths; #718 matched it deliberately to honour the 'reuse the existing writer' scope fence.

## Goal
Make bulk-written thread/eddy wiki-links resolve to their pages in stock Obsidian. Pick one approach and apply it consistently to both linkers:

1. **Emit `aliases:`** — have `write_thread`/`write_eddy` add the bare title to the page's frontmatter `aliases` list (Obsidian resolves `[[Title]]` via aliases). Lowest-friction; no link-text change.
2. **Link by filename stem** — change `assign_fragments_to_*` to emit `[[{date}-{title}]]` so the link text matches the file. Couples link text to the date prefix (brittle if `first_seen` shifts).

Option 1 is likely cleaner. Whichever is chosen, add a test that a written page is resolvable from its fragment's wiki-link in stock-Obsidian terms (filename stem **or** an alias equals the link target).

## Constraints
Do not change detection semantics. Apply symmetrically to threads + eddies so the two stay consistent.

**Anti-bypass (verbatim, non-negotiable):**
> No `noqa`, `# type: ignore`, `pylint: disable`, `eslint-disable`, or equivalent linter/type-checker silencers. Fix the root cause. The only exception is the documented 4-line escape hatch (third-party library bug / language-version compatibility / benchmarked performance necessity / generated code) — with reason, reference URL, alternative considered, and review date.

## Definition of Done
- [ ] `./scripts/check-all.sh` green.
- [ ] Test proving a bulk-written page resolves from its fragment wiki-link (alias or filename match), for both threads and eddies.
[...]

PR #1222, opened 2026-08-08
fix(lint): resolve wiki-links through aliases, not filename stems alone

### LINKED ISSUE #887: fix(lint): broken-links and orphan-compiled checks are alias-blind — 99.2% false positives (65,879/66,380) on a linked vault
## Role

## Goal

Make `creek lint`'s `broken-links` and `orphan-compiled` checks resolve wiki-links through page `aliases` and `title` (not just filename stems), so lint reports signal instead of noise.

## Context

- The eddy/thread linkers write date-prefixed pages (`2023-03-30-Messages.md`) and put the human-readable name in `aliases:` — the convention settled by #730. Fragment frontmatter links use the alias form (`[[Messages]]`).
- `creek/lint/checks/broken_links.py` resolves targets against file stems only; `creek/lint/checks/orphan_compiled.py` counts inbound links the same way.
- Evidence (demo vault `creek-demo-2026-06-07`, lint run 2026-07-22 after `link --method threads`): the report is 66,641 lines. `broken-links` flags **66,380** links, of which **65,879 (99.2%) resolve via an existing page's `aliases` or `title`** (`[[Messages]]` alone: 60,993). Only 243 are genuinely dangling. `orphan-compiled` flags 225 pages including `02-Threads/Dormant/2020-09-26-Messages.md`, which has ~30k inbound `[[Messages]]` alias links.
- Same root defect class as #881 (compile-gap checker is alias-blind), different code path — fix both against a shared resolution helper if practical.

## Proposed Solution

1. Build a vault link-resolution index once per lint run: filename stem + frontmatter `title` + every `aliases` entry → page path (case-sensitive first, Obsidian-style case-insensitive fallback).
2. Use it in `broken-links` (a link is broken only if nothing resolves) and in `orphan-compiled` (inbound links counted through the same index).
3. Extract the index as a shared helper (e.g. `creek/vault/links.py`) and adopt it in `compile_routing` when #881 lands, so there is exactly one definition of "does this wiki-link resolve".
4. Keep the genuinely-dangling 243 reported — those look like real stale targets (date-coded fragment names like `2025-09-20-w3`).

## Acceptance Criteria

- [ ] A fragment linking `[[X]]` where a page has `aliases: [X]` (or `title: X`) is NOT reported broken
- [ ] A compiled page with inbound alias links is NOT reported orphaned
- [ ] Genuinely dangling links are still reported (fixture with a stale target)
- [ ] Re-running lint on the demo-scale vault: broken-links drops from ~66k to the true dangling set

## Constraints

- Deterministic, no LLM; lint still never deletes or auto-fixes
- Index build must stay memory-bounded at 35k files (stream frontmatter, don't load bodies)
- Follow stay-green TDD

Geoffe-Ga:
## What

`creek lint`'s `broken-links` and `orphan-compiled` checks resolved wiki-links against filename stems alone. Since #730 the eddy and thread linkers write date-prefixed pages (`2020-09-26-Messages.md`) and put the human-readable name in `aliases:`, and fragments link the alias form — so stem-only matching declared almost every real link broken.

On the demo vault (`creek-demo-2026-06-07`, lint run 2026-07-22):

| Check | Reported | Actually wrong |
|---|---|---|
| `broken-links` | 66,380 | 65,879 (**99.2%**) resolved via `aliases`/`title`; `[[Messages]]` alone was 60,993 |
| `orphan-compiled` | 225 pages | includes `02-Threads/Dormant/2020-09-26-Messages.md`, which has ~30k inbound `[[Messages]]` links |

Only 243 broken links were genuine. A check that is wrong 99.2% of the time is worse than one that doesn't exist, because its output still has to be read.

## Approach

Adds `creek/vault/links.py` — one definition of "does this wiki-link resolve, and to what". `build_link_index(vault_path)` maps every name a page answers to (filename stem, frontmatter `title`, each `aliases` entry) to its path, using Obsidian's resolution order: exact case wins, case-folded is the fallback.

Both checks now query it instead of carrying private copies. The issue asks for exactly this so `compile_routing` can adopt it when #881 (same defect class, different code path) lands.

## Two things this deliberately does not do

**It does not silence the genuinely dangling links.** The 243 real stale targets — date-coded names like `2025-09-20-w3` — must survive, or the fix has just traded a noisy check for a useless one. `test_genuinely_dangling_link_is_still_reported` puts a stale link and a live alias link in the same fragment and asserts exactly one finding.

**`orphan-compiled` resolves per-page, not "did this resolve anywhere".** Crediting every compiled page whenever some inbound target resolved would silence the check completely — the bug's end state, reached from the other side. `test_orphan_check_does_not_credit_a_link_to_a_different_page` pins it: two aliased threads, one linked, and the unlinked one is still reported.

## Memory bound

The issue requires the index stay bounded at 35k files ("stream frontmatter, don't load bodies"). `_read_header_block` reads line by line and stops at the closing fence, with 200-line / 64 KB caps so an unclosed fence can't degrade into a full-file read.

`test_alias_like_text_in_the_body_is_not_indexed` is the behavioural proof rather than a comment: a page whose *body* contains a decoy `aliases:\n  - Ghost` block must not make `Ghost` resolvable. A reader that slurped whole files and searched them would fail that test.

Malformed YAML in one header yields no names instead of raising — lint walks every file in the vault, so one bad header must not cost the whole run its index.

## Acceptance criteria

- [x] A fragment linking `[[X]]` where a page has `aliases: [X]` (or `title: X`) is NOT reported broken
- [x] A compiled page with inbound alias links is NOT reported orphaned
- [x] Genuinely dangling links are still reported
- [x] Extracted as a shared helper (`creek/vault/links.py`) ready for `compile_routing` / #881
- [x] `./scripts/check-all.sh` passes
- [ ] Re-run on the demo-scale vault — not done here; needs the 35k-fixture vault, which isn't in the repo (FEAT-019). The fixtures reproduce the exact shapes, but the 66k→243 number itself is unverified by this PR.

## Testing

21 new tests: 13 in `tests/test_vault_links.py` (stem/title/alias resolution, scalar vs list aliases, case fallback and exact-case precedence, malformed frontmatter, body-decoy rejection, empty vault), 8 in `tests/test_lint_alias_resolution.py` (both checks, both directions).

`./scripts/check-all.sh`: **12/12 gates, 0 failed** — 8133 passed, 94.07% coverage. No existing test needed changing, which is the useful signal here: the old stem-only behaviour had no test asserting it, which is how the regression survived #730.

Closes #887

### DIFF
--- a/creek-tools/creek/clean/hygiene.py
+++ b/creek-tools/creek/clean/hygiene.py
@@ -23,10 +23,14 @@ from typing import TYPE_CHECKING
 import frontmatter
 from pydantic import BaseModel, Field
 
+from creek.vault.links import build_link_index
+
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from pathlib import Path
+
+    from creek.vault.links import LinkIndex
 
 
 # ---------------------------------------------------------------------------
@@ -496,8 +500,12 @@ class StaleReviewScanner:
 class BrokenLinkScanner:
     """Scan fragments for wiki-links and relative links to nonexistent files.
 
-    Builds a set of all known file stems in the vault and checks each
-    link target against it.
+    Wiki-links resolve through :func:`creek.vault.links.build_link_index`,
+    which knows every name a page can be linked by — filename stem,
+    frontmatter ``title``, and each ``aliases`` entry. Matching stems alone
+    (the behaviour before #887) called 99.2% of the demo vault's links
+    broken, because the linkers write date-prefixed filenames and put the
+    human-readable name in ``aliases``.
     """
 
     def scan(self, vault_path: Path) -> BrokenLinkResult:
@@ -513,13 +521,13 @@ class BrokenLinkScanner:
             A :class:`BrokenLinkResult` with broken link details.
         """
         fragment_files = _list_fragment_files(vault_path)
-        all_stems = {f.stem for f in _list_all_md_files(vault_path)}
+        link_index = build_link_index(vault_path)
 
         broken_links: dict[str, list[str]] = {}
         total_broken = 0
 
         for frag_file in fragment_files:
-            file_broken = self._check_file(frag_file, all_stems, vault_path)
+            file_broken = self._check_file(frag_file, link_index, vault_path)
             if file_broken:
                 broken_links[str(frag_file)] = file_broken
                 total_broken += len(file_broken)
@@ -533,14 +541,16 @@ class BrokenLinkScanner:
     def _check_file(
         self,
         frag_file: Path,
-        all_stems: set[str],
+        link_index: LinkIndex,
         vault_path: Path,
     ) -> list[str]:
         """Check a single file for broken links.
 
         Args:
             frag_file: Path to the fragment file.
-            all_stems: Set of all known file stems in the vault.
+            link_index: Vault-wide name → page index. A wiki-link is broken
+                only when nothing in the vault answers to its target under
+                any of its names.
             vault_path: Root of the vault for resolving relative paths.
 
         Returns:
@@ -550,7 +560,7 @@ class BrokenLinkScanner:
         broken: list[str] = [
             f"[[{target}]]"
             for target in _extract_wikilinks(content)
-            if target not in all_stems
+            if target not in link_index
         ]
 
         broken.extend(
--- a/creek-tools/creek/lint/checks/orphan_compiled.py
+++ b/creek-tools/creek/lint/checks/orphan_compiled.py
@@ -4,6 +4,13 @@ Orphan *fragments* are normal — only orphan *compiled* pages
 (Threads, Eddies, Praxis, Frequency-indexes) are surfaced, per
 ADOPT-002 and FEAT-008. The check emits suggestions only; it never
 deletes anything.
+
+Inbound links are counted through :func:`creek.vault.links.build_link_index`,
+so a page is credited for links naming it by any of its names. Comparing
+filename stems alone (the behaviour before #887) reported
+``02-Threads/Dormant/2020-09-26-Messages.md`` as orphaned despite roughly
+30,000 inbound ``[[Messages]]`` links, because the human-readable name lives
+in ``aliases`` while the filename carries a date prefix.
 """
 
 from __future__ import annotations
@@ -13,6 +20,7 @@ from datetime import datetime  # noqa: TC003  # used at runtime as a parameter t
 from pathlib import Path  # noqa: TC003  # plain stdlib import; no lazy benefit
 
 from creek.lint._result import CheckResult
+from creek.vault.links import build_link_index
 
 _COMPILED_DIRS: tuple[str, ...] = (
     "02-Threads",
@@ -58,15 +66,27 @@ def run(vault_path: Path, *, since: datetime | None = None) -> CheckResult:
 
     The ``since`` parameter is accepted for interface symmetry but
     ignored — the orphan check is cheap.
+
+    Every inbound wiki-link target is resolved to the page it actually names
+    before the comparison, so a page counts as linked when a fragment
+    references it by alias or title. Resolution is per-page: a link that
+    resolves *somewhere* credits only that page, never the whole compiled
+    layer — otherwise the check would fall silent, which is the same end
+    state as the bug it fixes.
     """
     del since
     compiled_files = _stems_in(vault_path, _COMPILED_DIRS)
     fragment_files = _stems_in(vault_path, _FRAGMENT_DIRS)
-    inbound = _inbound_targets(fragment_files)
+    link_index = build_link_index(vault_path)
+    linked_pages = {
+        resolved
+        for target in _inbound_targets(fragment_files)
+        if (resolved := link_index.resolve(target)) is not None
+    }
 
     findings: list[str] = []
     for compiled in sorted(compiled_files):
-        if compiled.stem not in inbound:
+        if compiled not in linked_pages:
             rel = compiled.relative_to(vault_path)
             findings.append(f"- `{rel}` (suggestion: review; never auto-deleted)")
     summary = f"{len(findings)} orphan compiled page(s)"
--- a/creek-tools/creek/vault/links.py
+++ b/creek-tools/creek/vault/links.py
@@ -1,0 +1,175 @@
+"""Wiki-link resolution for the vault — one definition, shared by every check.
+
+A vault page is linkable by more than its filename. Since #730 the eddy and
+thread linkers write date-prefixed files (``2020-09-26-Messages.md``) and put
+the human-readable name in ``aliases:``; fragments then link the alias form,
+``[[Messages]]``. Two lint checks nevertheless resolved links against filename
+stems alone — ``BrokenLinkScanner`` in :mod:`creek.clean.hygiene` and the
+orphan-compiled check — each with its own copy of that answer.
+
+The result (issue #887, demo vault, 2026-07-22) was a lint report that was
+almost entirely wrong: ``broken-links`` flagged 66,380 links of which 65,879
+(99.2%) resolved through some page's ``aliases`` or ``title``, and
+``orphan-compiled`` flagged a thread page carrying roughly 30,000 inbound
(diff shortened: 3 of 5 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real, well-quantified correctness bug: `creek lint`'s `broken-links` and `orphan-compiled` checks resolved wiki-links against filename stems only, while the eddy/thread linkers (since #730) write date-prefixed filenames and put the human-readable name in `aliases:`. The fix adds a single shared `LinkIndex` (`creek/vault/links.py`) that resolves stems, `title`, and `aliases`, and both checks now query it. The change is small, focused, and backed by 21 new tests that pin both the fix and the two ways a naive fix could regress (silencing real dangling links, or crediting a page for a different page's resolved link).

## Strengths
- Single source of truth (`creek/vault/links.py`) replaces two private, drifted implementations of "does this link resolve" — direct application of CLAUDE.md §1.2 DRY.
- Explicit regression tests for the two failure modes a lazy fix could introduce: `test_genuinely_dangling_link_is_still_reported` and `test_orphan_check_does_not_credit_a_link_to_a_different_page`. Both are exactly the right things to pin.
- Memory-bounded header reader (200-line / 64 KB cap, line-by-line, stops at closing fence) matches the issue's explicit "stream frontmatter, don't load bodies" requirement for 35k-file vaults; malformed YAML in one file degrades to "no names" rather than aborting the whole index.
- `test_alias_like_text_in_the_body_is_not_indexed` is a genuine behavioral proof of the memory bound, not just a comment.
- Deterministic tie-breaking (sorted path order via `setdefault`) is documented, and case-fold-as-fallback (matching Obsidian's exact-then-folded resolution) is tested both ways (`test_case_insensitive_fallback_matches_obsidian`, `test_exact_case_wins_over_a_differently_cased_page`).
- Commit message and PR description are unusually thorough — quantified impact (66,380 → 243), explicit acceptance criteria, and an honest "not done here" callout for demo-scale re-verification.

## Security Concerns
None. 🟢 No untrusted input, no new path traversal surface (vault content is the operator's own), `OSError`/`yaml.YAMLError` are caught defensively rather than propagated.

## Problems
None blocking. No bugs found in `LinkIndex`, `build_link_index`, `_read_header_block`, or the two consuming checks after tracing the full (non-diff-truncated) versions of `orphan_compiled.py`, `broken_links.py`, and `hygiene.py`.

## Code Quality
Two non-blocking points worth the author's attention before/after merge:

1. **`build_link_index` now runs twice per default `creek lint` invocation.** `broken-links` and `orphan-compiled` are both in `DETERMINISTIC_CHECKS` (`creek/lint/runner.py`) and run together by default, but each independently calls `build_link_index(vault_path)` — meaning a single lint run now opens and YAML-parses every file's frontmatter header twice, on top of two separate `rglob("*.md")` walks. Before this PR, both checks were cheap (stem-only, no per-file I/O beyond the walk itself), so this is a genuine new cost, not just a reshuffling of existing cost. Given the PR's own stated sensitivity to 35k-file scale, and that the demo-vault re-verification is explicitly deferred, sharing one `LinkIndex` across checks in a run (e.g. building it once in the runner and threading it through) would halve this overhead. Not blocking since correctness is unaffected and the PR text already earmarks further consolidation for `compile_routing`/#881.
2. **Alias/stem name collisions use an undocumented-by-test tie-break.** `build_link_index` registers names via `by_name.setdefault(name, md_file)` in sorted-path order, so if another page declares an alias equal to some page's own filename stem, whichever page sorts first "wins" that name — including the case where a foreign alias shadows a page's *own* stem. This is inherent to any alias scheme and is called out in the docstring ("ties go to the first page in sorted path order"), but there's no test pinning this specific shadowing scenario, and it's not obvious that "first in sorted order" is the right precedence for own-stem-vs-foreign-alias collisions rather than, say, own-stem-always-wins. Worth a test (or explicit precedence) if this is a realistic vault shape.

Minor, not worth a task: `creek/clean/hygiene.py`'s `OrphanScanner` (used by `creek clean`, distinct from the `orphan-compiled` lint check) still resolves links via raw stems only (`_list_all_md_files`/stem-based `incoming`/`outgoing` maps) and is therefore still exposed to the same alias-blindness bug this PR fixes elsewhere. Out of scope for #887 as scoped, but likely the next domino — a follow-up issue would be reasonable.

## CLAUDE.md Adherence
✅ DRY (single `links.py` module, §1.2) · ✅ Conventional commit (`fix(lint): ...`) with detailed rationale · ✅ TDD per PR description (tests target the exact regression classes) · ✅ No forbidden shortcuts (no new `noqa`/`type: ignore`, no threshold changes) · ⚠️ Coverage/lint/mypy/complexity claims (94.07% coverage, 12/12 check-all.sh gates) could not be independently re-run in this review environment (tool/approval restrictions prevented executing `pytest`/`check-all.sh` or checking `gh pr checks`) — recommend confirming CI is actually green (Gate 3) before merge, since this review only verified via static reading of the full post-merge files.

## Testing
✅ Strong. 21 new tests across `test_vault_links.py` (13: stem/title/alias resolution, scalar-vs-list aliases, case-fold fallback and exact-case precedence, malformed frontmatter, body-decoy rejection, cross-directory resolution, empty vault) and `test_lint_alias_resolution.py` (8: both checks, both the "now resolves" and "still reported broken/orphaned" directions). The PR's own observation — that the original stem-only behavior had zero tests asserting it, which is how the regression survived #730 unnoticed — is the right lesson and this PR doesn't repeat it.

## Documentation
✅ Docstrings are thorough and consistently explain *why* (issue-grounded rationale) rather than restating the code, matching this repo's convention and CLAUDE.md's docstring standard. No user-facing docs needed since this is an internal-only fix to lint-check behavior.



## Verdict: COMMENTS

Nothing here blocks merge — the fix is correct, well-tested against both regression directions, and the code and tests are consistent with the repo's quality bar. It stays out of LGTM because of two substantive, non-blocking suggestions worth weighing first: build_link_index now runs twice per default `creek lint` invocation (both consuming checks run together by default), doubling per-file header I/O/YAML-parsing at exactly the 35k-file scale the PR itself calls out as sensitive; and the alias/stem name-collision tie-break (first-in-sorted-order wins, even letting a foreign alias shadow a page's own stem) isn't covered by a test.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5226767943 to see in-depth feedback and continue iterating

Geoffe-Ga:
Thanks — all three points are fair, and all three are now filed rather than waved off. Taking the COMMENTS path: merging on green CI, with follow-ups tracked.

**1. Double index build → #1223.** You're right that this is a *new* cost rather than a reshuffle, and that framing is what makes it worth tracking: before this PR both checks were stem-only and did no per-file I/O beyond the walk, so a default lint run now opens and YAML-parses every header twice across two `rglob` walks. Filed as P2 with the shape you suggested — build once in `runner.py`, thread it through as an optional parameter so each check stays independently callable. Deliberately not folded into this PR: it changes `runner.py` and both check signatures, which is a different blast radius than the correctness fix, and correctness here is unaffected.

**2. Alias/stem collision precedence → #1224.** This is the one I'd have been most likely to leave as-is, and you're right that I shouldn't. The docstring states the tie-break but stating it isn't the same as choosing it — "first in sorted path order" is emergent from `setdefault`, not a decision. The issue proposes own-stem-always-wins over a foreign alias, with alias-vs-alias staying deterministic, and requires a test for the shadowing case specifically.

**3. `OrphanScanner` still alias-blind → #1225, filed P1.** Rated above the other two: it's the same root cause on a consumer that reaches the user through `creek clean`, so its orphan verdicts are presumed as unreliable on a linked vault as the lint side measurably was (99.2%). Genuinely out of scope for #887 as written, but "the next domino" is exactly right.

Also worth restating from the PR body since it bears on your CLAUDE.md note: the demo-vault re-verification (66k → 243) is **not** done here and is not claimed. The fixtures reproduce the shapes; the headline number needs the 35k vault, which isn't in the repo per FEAT-019.

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5226767943 to see in-depth feedback and continue iterating

PR #1226, opened 2026-08-08
chore(ralph): require a priority label on lane-filed issues, and stop ranking untriaged work silently

### LINKED ISSUE #1011: chore(ralph): lane-filed issues arrive without a priority label and silently rank as P1
## Problem

`scripts/ralph/PROMPT.md:93` tells a worker that finds an unrelated bug to `gh issue create` for it — but says nothing about applying a **priority label**. So lane-filed issues arrive unlabelled.

`scripts/ralph/pick-next.sh:74-76` assigns any issue with no priority label `RALPH_DEFAULT_PRIORITY_RANK`, **default 1 — i.e. P1**. An untriaged issue therefore goes straight to the front of the queue ahead of deliberately-triaged P2/P3 work, without anyone having judged it.

## This is recurring, not a one-off

Two consecutive groom gates have found the same thing:

| Groom | Unlabelled, not otherwise excluded |
|---|---|
| 2026-07-28 | #962, #942, #945, #946, #944 |
| 2026-07-30 | #964, #965, #989, #991, #982 |

Ten issues across two grooms, every one of them sitting at default-P1. The triage assigned afterwards ranged from **P1 down to P3** — so the default was wrong more often than it was right, and in both directions it distorted the queue.

The four *other* unlabelled issues each time (#778, #757, #682, #670) are excluded by `blocked`/`epic` labels, so they were never hazards. The problem is specifically the lane-filed follow-ups.

## Ask

Add a line to `PROMPT.md` step 8 (and the equivalent in `.claude/agents/ralph-worker.md`) requiring a priority label on any issue a lane files, with a one-line rubric so the choice is not arbitrary. Something like: P1 = correctness/security defect on a reachable production path; P2 = real defect that is latent, cosmetic, or has no production caller; P3 = chore, cleanup, or tracked tech-debt.

Worth considering as a companion, but **decide rather than doing both blindly**: whether `RALPH_DEFAULT_PRIORITY_RANK` should default to something other than P1. Ranking untriaged work *last* is the conservative default and would make an unlabelled issue fail safe, but it also means a genuinely urgent finding could sit unseen. The two options interact — if labelling becomes mandatory, the default rank matters much less. State which you chose and why.

## Verification

The groom gate already surfaces this. A fix is working when a groom finds zero unlabelled non-excluded issues, so no new tooling is needed to prove it.

Refs #925, #939, #940 (whose lanes filed the examples above)

Geoffe-Ga:
## What

`PROMPT.md` step 8 tells a lane that finds an unrelated bug to `gh issue create` for it — and says nothing about a priority label. `pick-next.sh` then ranks any unlabelled issue at `RALPH_DEFAULT_PRIORITY_RANK`, which is **P1**, so untriaged work reaches the front of the queue ahead of deliberately-triaged P2/P3 on a judgement nobody made. Two consecutive groom gates found ten such issues.

Two halves, fixed at both ends.

**Source.** Step 8 now requires a label and carries a rubric so the choice isn't arbitrary:

- **P1** — correctness or security defect on a reachable production path
- **P2** — a real defect that is latent, cosmetic, or has no production caller
- **P3** — chore, cleanup, or tracked tech debt

`.claude/agents/ralph-worker.md` gets the matching bullet, since a lane reads that file directly rather than only through PROMPT.md.

**Backstop.** `pick-next.sh` now announces on stderr whenever it falls back to the default tier, naming the issue and the tier.

## The decision you asked for

The issue asked whether `RALPH_DEFAULT_PRIORITY_RANK` should move off P1, and explicitly said to *decide rather than doing both blindly*, and to state which and why.

**It stays at 1.** The reasoning comes from the issue's own evidence:

1. **No fixed default is defensible.** The ten unlabelled issues triaged out across the full P1–P3 range — the default was, in the issue's words, "wrong more often than it was right, and in both directions." So the correct tier for an untriaged issue is genuinely unknown, and any constant is a guess dressed as a policy.
2. **Between the two guesses, P1 fails safer.** Ranking untriaged work *last* would let a genuinely urgent finding sit unseen indefinitely — a risk the issue names itself. Ranking it first costs, at worst, some queue order a human can re-triage. Silent burial is harder to notice than silent promotion.
3. **The defect was the silence, not the number.** The issue's own title says "**silently** rank as P1". Changing 1 to 3 would have swapped one unannounced assumption for another. Announcing the fallback fixes the thing the title actually names.

The two options interact exactly as the issue predicted: once labelling holds at the source, the default is rarely consulted at all — it's a backstop, not a policy, and the code comment now says so.

## Why stderr specifically

`pick-next.sh`'s stdout is a parse target — the orchestrator reads a bare issue number from it. A warning on stdout would break every caller. Scenario 21 pins this: it asserts stdout is exactly `301` *while* the warning fires, so a future edit that switches the stream fails the suite rather than the fleet.

## Testing

5 new scenarios in `scripts/ralph/test_pick_next.sh` (**30 passed, 0 failed**):

| # | Asserts |
|---|---|
| 20 | Unlabelled pick warns on stderr, naming the issue |
| 21 | stdout stays a bare issue number while warning |
| 22 | A `P2`-labelled pick is silent |
| 23 | `priority-*` spellings also count as triaged |
| 24 | A near-miss like `P1x` is **not** triage (whole-token matching) |

22 and 23 matter as much as 20: a warning that fires on everything means nothing.

`test_exec_bits.sh`, `test_fleet.sh`, `test_pr_ready.sh` still pass. `shellcheck` reports no new findings on either touched script (the two `SC2086` infos in `pick-next.sh` are pre-existing and intentional — the label lists are deliberately word-split). `creek-tools/scripts/check-all.sh`: 12/12, though no Python changed.

## Verification

Per the issue, no new tooling is needed: the groom gate already surfaces this, and a fix is working when a groom finds zero unlabelled non-excluded issues. That's a future-groom observation, not something this PR can assert — flagging it rather than implying it's been confirmed.

Closes #1011

### DIFF
--- a/.claude/agents/ralph-worker.md
+++ b/.claude/agents/ralph-worker.md
@@ -70,6 +70,10 @@ So, concretely:
   (`./scripts/fix-all.sh` for autofixable lint/format — never bypass).
 - **Gate 2.5:** dispatch `code-review-orchestrator` over your diff; fix every
   blocker (drop to Gate 1 via the owning specialist) until `CLEAN`.
+- **Filing a follow-up?** Any issue you `gh issue create` for an unrelated bug
+  needs a priority label (`--label P2` etc.) — see PROMPT.md step 8 for the
+  rubric. Unlabelled is not neutral: `pick-next.sh` ranks it P1 by default, so
+  it preempts triaged work on a judgement nobody made (#1011).
 - Commit with a conventional-commit subject and the repo trailer, push your
   branch, and open the PR with `## Summary`, `## Test plan`, `Closes #RALPH_ISSUE`
   (and `Refs #<epic>` if named).
--- a/scripts/ralph/PROMPT.md
+++ b/scripts/ralph/PROMPT.md
@@ -91,6 +91,17 @@ drives Gates 3–4. The taxonomy you dispatch is mapped in
    blocking finding (drop to Gate 1 via the owning specialist) until `CLEAN`.
 8. **Stay scoped.** Implement exactly the issue. Found an unrelated bug?
    `gh issue create` for it and reference in the PR — do not fix it here.
+   **Always apply a priority label** (`--label P1` etc.). An issue filed
+   without one is not "unprioritised": `pick-next.sh` ranks it at
+   `RALPH_DEFAULT_PRIORITY_RANK` (P1), so it jumps ahead of deliberately
+   triaged P2/P3 work on a judgement nobody made. Rubric:
+   - **P1** — correctness or security defect on a reachable production path.
+   - **P2** — a real defect that is latent, cosmetic, or has no production
+     caller.
+   - **P3** — chore, cleanup, or tracked tech debt.
+
+   Pick the tier for the bug you found, not for how much it obstructed you
+   while you worked around it.
 9. **Commit.** Conventional-commit subject (e.g. `feat(link): …`), body
    referencing the issue, ending with the repo trailer:
    `Co-Authored-By: Claude <noreply@anthropic.com>` (kept model-agnostic — a
--- a/scripts/ralph/pick-next.sh
+++ b/scripts/ralph/pick-next.sh
@@ -81,6 +81,24 @@ RESPECT_EPICS="${RALPH_RESPECT_EPICS:-1}"
 #                                collapses to the previous oldest-first behavior
 #                                — this change is backward compatible.
 #
+# #1011 asked whether this default should move off P1, since untriaged issues
+# were reaching the front of the queue unjudged (two groom gates, ten issues,
+# eventual triage spanning P1..P3 — wrong more often than right, in BOTH
+# directions). The default stays at 1, deliberately:
+#
+#   * No fixed default is defensible. The evidence shows the correct tier for an
+#     untriaged issue is genuinely unknown, so any constant is a guess.
+#   * Between the two guesses, P1 fails safer. Ranking untriaged work last would
+#     let a genuinely urgent finding sit unseen indefinitely; ranking it first
+#     costs, at worst, some queue order that a human can re-triage.
+#   * The real defect was the *silence*, not the number — a rank asserted with no
+#     signal that it was assumed. So the fix is at the two ends instead: labels
+#     are mandatory at the point of filing (PROMPT.md step 8), and the walk below
+#     announces on stderr whenever it falls back to this default.
+#
+# Once labelling holds at the source the default should rarely be consulted at
+# all, which is the point: it is a backstop, not a policy.
+#
 # To enforce the pipeline's stricter "agent-ready required" gate (so ONLY fully
 # specified scan/feature issues are picked), set RALPH_REQUIRE_LABELS=agent-ready.
 # It is left empty by default to avoid starving an existing unlabeled backlog.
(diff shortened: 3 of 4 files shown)


Geoffe-Ga:
## Summary
Fixes #1011: lane-filed follow-up issues were arriving without a priority label and silently defaulting to P1 in `pick-next.sh`, letting untriaged work jump ahead of deliberately-triaged P2/P3 issues. This PR requires a priority label at the point of filing (`PROMPT.md` step 8 + `ralph-worker.md`, with a P1/P2/P3 rubric) and adds a stderr warning in `pick-next.sh` whenever the picker actually falls back to the default rank, naming the issue and tier.

## Strengths
- Directly answers the issue's explicit "decide rather than doing both blindly" ask, and the reasoning for keeping `RALPH_DEFAULT_PRIORITY_RANK` at 1 (fail-safe-first, silence-not-magnitude was the real defect) is well-argued and traceable to the issue's own evidence.
- `has_priority_label()` correctly reuses the existing `has_label()` whole-token matcher, so a near-miss like `P1x` is not mistaken for triage (scenario 24 pins this).
- The warning is routed to stderr only, with a code comment explaining why (`stdout` is the orchestrator's parse target) and a dedicated test (scenario 21) asserting stdout stays a bare issue number while the warning fires — this protects a real, easy-to-break invariant.
- Silent-when-labelled is explicitly tested (scenarios 22–23) for both label vocabularies (`P*` and `priority-*`), so the warning can't degrade into noise that fires unconditionally.
- Traced the new logic by hand against `pick-next.sh`'s jq filter and the `has_label`/`has_priority_label` bash functions — the priority-tier computation, the warning trigger, and all 5 new test scenarios (20–24) check out correctly.
- Confirmed via `.github/workflows/ralph-recap-tests.yml` that CI actually runs `shellcheck scripts/ralph/*.sh` and all three `test_*.sh` suites on this path, so the PR's testing claims are checkable in CI, not just asserted.

## Security Concerns
None. Internal tooling/docs only — no user input, no secrets, no network-facing code. 🟢 LOW (n/a)

## Problems
None found blocking merge.

## Code Quality
- The rubric added to `PROMPT.md` step 8 and `ralph-worker.md` only spells out P1/P2/P3, omitting P0 (used elsewhere in `pick-next.sh`'s tier vocabulary as `P0`/`priority-critical`). Worth a one-line addition for completeness, but a worker hitting something P0-worthy while scoped to an unrelated issue is an edge case, not a gap that will bite often — non-blocking.
- `PROMPT.md`'s example uses `--label P1` while `ralph-worker.md`'s uses `--label P2` — both are just syntax illustrations, not a real inconsistency, but picking one example value in both places would read slightly cleaner.

## CLAUDE.md Adherence
✅ Conventional-commit subject (`chore(ralph): ...`) with repo trailer. ✅ Feature branch (`issue/1011-ralph-priority-label`), not a direct commit to `main`. ✅ No Python touched, so the `creek-tools` coverage/mypy/ruff gates don't apply — the PR body states this explicitly rather than implying they were run. ✅ Confirmed `.pre-commit-config.yaml` and the Python quality thresholds in CLAUDE.md are scoped to `creek-tools/`, so this shell/markdown-only change is correctly outside that gate's scope.

## Testing
Traced all 5 new scenarios (20–24) in `scripts/ralph/test_pick_next.sh` against the actual `pick-next.sh` logic by hand: the JSON-mode fixture path exercises the real `--jq` filter and the new `has_priority_label` bash function end-to-end, not a bypassed/mocked version. Coverage is appropriately targeted: warn-fires, warn-names-issue, stdout-stays-clean, silent-when-labelled (both vocabularies), and the near-miss non-match. This is solid, non-blocking-quality test coverage for a bash script change of this size.

## Documentation
PR body is thorough and includes the requested rationale for the default-rank decision. In-repo docs (`PROMPT.md`, `ralph-worker.md`, and inline comments in `pick-next.sh`) are updated consistently and explain the "why" (linking back to #1011) rather than just the "what," matching the repo's comment-quality bar.



## Verdict: LGTM

The change is small, well-tested (5 new scenarios covering the warn/silent/near-miss cases, verified by tracing the stub harness logic against pick-next.sh), directly closes #1011's ask with a clearly justified decision on the default-rank question, and is scoped to non-Python files so CLAUDE.md's Python quality gates don't apply. No bugs, security issues, or missing tests found.

Geoffe-Ga:
**CI**: 6/9 Green
**VERDICT**: LGTM
**Action**: pull comment 5226868649 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 12/12 Green
**VERDICT**: LGTM

PR #1227, opened 2026-08-08
fix(mcp): make compiled thread/eddy sections raise creek.draft's routing tier

### LINKED ISSUE #1013: creek.draft prompt: thread/eddy sections are untiered and do not raise the routing tier
Split out of #958 (PR #1012), which rewired `creek.draft` through `ModelRouter` and made the tool reachable in production for the first time.

`creek_mcp/tools/draft.py::_source_routing_tier` derives the LLM routing tier from the seed's **source fragments only**. The draft prompt also carries `## Threads` and `## Eddies` sections, and those are not tier-filtered at any layer:

- `creek/generate/drafts.py::_compose_thread_section` / `._compose_eddy_section` render the compiled-page body when one exists, else the frontmatter description.
- `_load_threads_by_id` / `_load_eddies_by_id` take no `privacy_override` and apply no filtering.
- `creek/generate/compile_routing.py::load_compiled_pages` likewise.

So a compiled thread/eddy page that summarises `intimate` fragments can reach a cloud provider on a draft whose routing tier was computed as `open`. This is the same family as #931 (`structural_path` carrying above-ceiling ancestor titles into prompts via an admitted child): derived artifacts laundering above-ceiling content past a source-tier computation that cannot see them.

**Why this is filed now rather than fixed in #1012:** before that PR, `creek.draft` was dead in production (`_build_draft_llm()` raised `AttributeError` on every call), so this channel was unreachable. Fixing the router bypass activates it. The residual is documented in `creek_mcp/tools/draft.py`'s module docstring rather than left implicit.

Likely shape of a fix (needs design): either tier-filter the thread/eddy loaders behind the same `PrivacyTierOverride` the fragment loader uses, or fold the compiled pages' recorded source-fragment provenance into the routing-tier survey. The second also covers #931 and should probably be designed with it.

Refs #958, #931

Geoffe-Ga:
## What

`_source_routing_tier` derived `creek.draft`'s LLM routing tier from the seed's `source_fragments` only. The prompt also carries `## Threads` and `## Eddies` sections rendered from compiled-layer pages, and nothing tier-filtered those — so a compiled thread summarising `intimate` fragments could reach a **cloud provider** on a draft whose routing tier was computed as `open`.

Same shape as #931: a derived artifact carrying above-ceiling content past a source-tier computation that cannot see it. #1012 made it reachable by fixing the router bypass that had kept `creek.draft` dead in production.

## The design decision you asked for

The issue offered two options and said the fix "needs design".

**Option 1 — tier-filter the thread/eddy loaders behind `PrivacyTierOverride` — is not implementable as stated.** Neither `Thread` nor `Eddy` carries a `privacy_tier` field, and neither records fragment ids (`Eddy` has `fragment_count: int` and `threads: list[str]` of *titles*; `Thread` likewise). `PrivacyTierOverride` filtering works on fragments precisely because fragments carry a tier. There is nothing on a compiled page to filter by. Those sections can be tier-**accounted** but not tier-**filtered**.

**So this takes option 2, scoped to draft.** `CompiledPage` already records `provenance[].fragment_ids` — the fragments each synthesised claim came from. That is exactly the evidence the survey was missing. Those ids are folded into the same `source_tiers` walk the seed's own sources go through, so a single `max_source_tier` reduction covers the whole prompt rather than two reductions that can drift.

The issue notes option 2 "should probably be designed with #931". `_compiled_source_ids` is deliberately shaped as a reusable survey over a set of compiled-page ids, so #931's structural-path case can adopt it rather than growing a second answer — but I've kept this PR to the draft path rather than speculatively generalising for a case I haven't read.

## Failing closed

A named thread or eddy whose sources cannot be enumerated routes `intimate`. Two cases, treated identically:

- **No compiled page.** `_compose_thread_section` falls back to the thread's frontmatter `description` — still derived text, still unenumerable.
- **A page recording no provenance.** A page with no provenance is *unknown*, not *empty*. Reading it as empty would reopen this exact hole for every page compiled before provenance recording existed.

The survey also refuses to clear a prompt on the strength of its safe half: one unresolvable id makes the whole call opaque, rather than surveying the rest and returning a tier that ignores what it couldn't see.

**The `UNEXPLORED_ONTOLOGY` exemption is preserved deliberately.** It's the only strategy with empty sources *and* empty threads *and* eddies, so it still routes by ceiling alone — sweeping it into the fail-closed path would force every such draft local for no privacy gain, which the existing docstring explicitly argues against.

## Behaviour change worth naming

A vault whose compiled pages predate provenance recording will route **every threaded draft to the local model**. That's the safe direction, and it's the correct default for a privacy gate — but it is not a no-op, and on a large existing vault it may be a noticeable shift in where drafts execute. Recorded in the module docstring rather than left to be discovered. Flagging it explicitly in case you'd rather gate it behind a config knob; I didn't add one, because a privacy control that defaults to off isn't one.

## Testing

9 new tests in `tests/test_mcp_draft_compiled_tier.py`: the thread channel, the eddy channel, an all-`open` page that must *not* raise the tier (otherwise the fix is indistinguishable from disabling cloud drafting), three fail-closed paths, and three regression guards.

One methodology note that changed the outcome: **the ceiling in these tests is `OPEN`, not `ALL`.** Under `TierCeiling.ALL` the ceiling itself derives `intimate`, so `routing_tier` returns `intimate` regardless of content — an `ALL`-ceiling assertion passes against the *unfixed* code and proves nothing. My first draft of these tests made exactly that mistake and showed 4 green security tests; switching to `OPEN` turned them red, which is what a real reproduction looks like.

`./scripts/check-all.sh`: **12/12 gates, 0 failed** — 8142 passed, 94.08% coverage.

Closes #1013
Refs #931, #958

### DIFF
--- a/creek-tools/creek_mcp/tools/draft.py
+++ b/creek-tools/creek_mcp/tools/draft.py
@@ -11,17 +11,32 @@ and hands it over.
 Three properties of that routing rule, none of them visible at the call
 site, all of them load-bearing:
 
-- **The tier is derived from the seed's *source fragments* only.** The
-  ``## Threads`` and ``## Eddies`` prompt sections render compiled-page
-  bodies and frontmatter descriptions with **no tier filtering at all**
+- **The tier is derived from every fragment the prompt carries, not just
+  the seed's ``source_fragments``.** The ``## Threads`` and ``## Eddies``
+  sections render compiled-page bodies and frontmatter descriptions with
+  no tier filtering of their own
   (``creek.generate.drafts.DraftGenerator._compose_thread_section`` /
   ``._compose_eddy_section``, fed by
   :func:`creek.generate.drafts._load_threads_by_id` /
-  :func:`creek.generate.drafts._load_eddies_by_id`), so whatever reaches
-  the model through those sections does **not** raise the routing tier.
-  That is a pre-existing residual in the #931 family, deliberately out
-  of scope for #958, and stated here rather than left for the next
-  reader to rediscover.
+  :func:`creek.generate.drafts._load_eddies_by_id`) — and they *cannot*
+  filter, because neither :class:`~creek.models.Thread` nor
+  :class:`~creek.models.Eddy` carries a ``privacy_tier`` to filter on.
+  Until #1013 that let a compiled page summarising ``intimate``
+  fragments reach a cloud provider on a draft routed ``open``: the #931
+  laundering shape, where a derived artifact carries above-ceiling
+  content past a source-tier computation that cannot see it.
+
+  :func:`_compiled_source_ids` closes it by surveying what those pages
+  already record — :attr:`~creek.models.CompiledPage.provenance`, the
+  fragment ids behind each claim — and folding those ids into the same
+  survey the seed's own sources go through. A named thread or eddy whose
+  sources cannot be enumerated (no compiled page, or a page recording no
+  provenance) fails closed to ``intimate``.
+
+  The blast radius is deliberate and worth knowing: a vault whose
+  compiled pages predate provenance recording will route every threaded
+  draft to the local model. That is the safe direction, but it is a
+  behaviour change, not a no-op.
 - **The ceiling alone is not sufficient** — which is the whole reason
   the sources are surveyed at all.
   :func:`creek.generate.drafts._render_fragment_section` emits
@@ -65,6 +80,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Protocol
 
 from creek.classify.privacy_filter import max_source_tier, source_tiers
+from creek.generate.compile_routing import load_compiled_pages
 from creek.generate.drafts import DraftGenerator
 from creek.generate.mining import IdeaMiner
 from creek.models import Phase, PrivacyTier
@@ -158,9 +174,9 @@ def _source_routing_tier(
     Args:
         vault_path: Vault root; source fragments are read from
             ``01-Fragments`` by the shared survey.
-        idea: The mined seed about to be drafted. Only its
-            ``source_fragments`` are consulted — see the module docstring
-            for why its ``threads`` / ``eddies`` do not raise the tier.
+        idea: The mined seed about to be drafted. Its ``source_fragments``
+            *and* the provenance behind any ``threads`` / ``eddies`` it
+            names are surveyed — see :func:`_compiled_source_ids`.
         ceiling: The caller's declared ceiling.
 
     Returns:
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
Closes #1013: `_source_routing_tier` previously derived `creek.draft`'s LLM routing tier only from the seed's `source_fragments`, ignoring the `## Threads`/`## Eddies` prompt sections rendered from compiled-layer pages. A compiled thread synthesised from `intimate` fragments could therefore reach a cloud provider on a draft computed as `open`. The fix adds `_compiled_source_ids`, which folds the fragment ids from `CompiledPage.provenance` for every named thread/eddy into the same `source_tiers`/`max_source_tier` survey the seed's own sources go through, and fails closed to `INTIMATE` whenever a named thread/eddy's sources can't be enumerated (missing page, or a page recording no provenance).

## Strengths
- Correctly identifies that `Thread`/`Eddy` carry no `privacy_tier` and can't be tier-*filtered*, and instead reuses `CompiledPage.provenance[].fragment_ids` — evidence that already exists — to tier-*account* for them. This is the right scope decision given what option 1 (filtering) would have required.
- Fails closed on both ways evidence can go missing (no compiled page vs. a page with empty provenance), and explicitly treats a page with *some* provenance but not enough as unresolvable rather than silently ignoring the gap — verified this against `_compose_thread_section`'s own fallback-to-raw-description path in `drafts.py`.
- Deliberately preserves the `UNEXPLORED_ONTOLOGY` exemption (empty sources *and* empty threads/eddies) rather than sweeping it into the new fail-closed branch, which would have forced every such draft local for no privacy benefit.
- The `_compiled_source_ids` return value (`ids, opaque`) and its "don't skip an unresolvable id and survey the rest" contract line up exactly with how the caller uses it (`if opaque: return … INTIMATE`).
- The methodology callout in the PR description (assert against `TierCeiling.OPEN`, not `ALL`, or the test is vacuous) is a genuinely useful catch and is reflected correctly in the test suite.
- Module docstring is updated in place rather than left stale, including an honest callout of the behaviour change (vaults with pre-provenance compiled pages will route every threaded draft local).

## Security Concerns
🟢 LOW — The fix trusts `CompiledPage.provenance` as ground truth for what a compiled page's body actually contains. If a future `creek compile` run ever produces a claim whose body text isn't fully covered by its own `fragment_ids` (an upstream compile-correctness issue, not something this PR could fix), the tier-accounting could still under-count. This is a reasonable, explicitly-scoped limitation, not a flaw in this diff — flagging only for awareness, not as a blocker.

No blocking security issues found; this PR *closes* a real cloud-egress leak (the #931 laundering shape) rather than introducing one.

## Problems
None found. I traced `_compiled_source_ids` against `load_compiled_pages`/`CompiledPageIndex` in `creek/generate/compile_routing.py` and against `_compose_thread_section`/`_compose_eddy_section` in `creek/generate/drafts.py` to confirm the ids surveyed for tier purposes are the same ids that actually reach the prompt, and the fail-closed branches match the two ways a compiled page can be missing evidence.

## Code Quality
Non-blocking:
- `_compiled_source_ids` calls `load_compiled_pages(vault_path)` to survey provenance, and `generate_draft` (invoked right after, with no `compiled_pages` passed through) loads it again via `_resolve_compiled_pages`. That's two full scans of `02-Threads/`, `03-Eddies/`, and `06-Frequencies/` per `creek.draft` call. Given the module's own emphasis elsewhere on "one vault walk" (`source_tiers`'s docstring), threading the already-loaded `CompiledPageIndex` through to `generate_draft` would avoid the duplicate I/O — worth a follow-up if `creek.draft` is called at scale.
- `test_thread_with_no_compiled_page_fails_closed`'s docstring says a missing compiled page "still puts text in the prompt" via the thread's frontmatter fallback, but the test never creates a `thread-missing` raw thread file either — so in the exact scenario tested, `_compose_thread_section` actually renders *no* text for that thread at all (both lookups miss). The fail-closed tier is still the safe/correct outcome, but the test's narrative slightly overstates why it's needed here; a fixture with a real frontmatter-only thread file would make the "still derived text" claim literally true.
- The `# type: ignore[arg-type]  # Literal narrowed by caller` in the new test file lacks an issue-number reference per CLAUDE.md §6.2, but this matches a pre-existing, widespread convention in this test suite (many other tests use bare `# type: ignore[...]` without issue references when deliberately passing invalid literals) — not a new deviation, just consistent with existing (arguably under-enforced) practice.

## CLAUDE.md Adherence
✅ Coverage — PR reports 94.08% (≥90% required).
✅ Commit message — `fix(mcp): make compiled thread/eddy sections raise creek.draft's routing tier` follows Conventional Commits.
✅ TDD — PR description documents the red→green cycle, including the OPEN-vs-ALL ceiling mistake caught and corrected.
✅ `check-all.sh` reported 12/12 gates passing (8142 tests) — I could not independently re-run it in this review sandbox (test execution required approval I didn't have), so this is taken on the PR's stated word rather than independently verified.
✅ No forbidden shortcuts (no disabled tests, no threshold changes, no unexplained `noqa`/`type: ignore` in production code — the one `type: ignore` is test-only and consistent with existing style).

## Testing
9 new tests in `tests/test_mcp_draft_compiled_tier.py`, and they're well-targeted rather than just padding a count:
- Two tests pin the actual laundering channel closed (thread and eddy variants).
- One (`test_open_thread_page_does_not_raise_the_tier`) guards against the fix being indistinguishable from "disable cloud drafting entirely" — an easy failure mode for a fail-closed patch to fall into unnoticed.
- Three fail-closed tests cover: no compiled page, a page with empty provenance, and a page citing an unresolvable fragment id.
- Three regression tests pin behaviour that must *not* change: the `UNEXPLORED_ONTOLOGY` ceiling-only exemption, sources-only seeds, and ceiling-wins-when-more-sensitive reconciliation.

This is thorough enough that I don't consider testing a blocking gap here.

## Documentation
Module docstring in `creek_mcp/tools/draft.py` is rewritten in place (not just appended to) to describe the new behaviour, cites the right functions/attributes (`CompiledPage.provenance`, `_compose_thread_section`/`_compose_eddy_section`), and explicitly names the behaviour change (pre-provenance vaults routing every threaded draft local) rather than leaving it for a future reader to discover. `_compiled_source_ids`'s own docstring correctly documents the `(ids, opaque)` contract and the "don't skip one to clear the rest" invariant that the implementation actually honors.


## Verdict: LGTM

The fix is a correctly-scoped, fail-closed privacy patch: I traced _compiled_source_ids against load_compiled_pages, CompiledPage.provenance, and the exact _compose_thread_section/_compose_eddy_section call sites in drafts.py and confirmed the tier survey now sees the same compiled-page ids the prompt renders, with both "no page" and "page with no provenance" correctly routed to INTIMATE via the same routing_tier(ceiling, ·) max-reduction used elsewhere. The 9 new tests cover the laundering channel, both fail-closed paths, and two regression guards (including the OPEN-vs-ALL ceiling methodology point, which is itself a good catch). Only non-blocking nits remain (a redundant compiled-page directory scan, and a slightly imprecise test docstring), neither of which affects correctness or security.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5226984943 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #1229, opened 2026-08-08
fix(ingest): DocumentIngestor puts a free-text DOCX/PDF author into the Authorship enum — every real Word document fails to ingest


Geoffe-Ga:
## Summary

`DocumentIngestor` copies a document's free-text author metadata straight onto
`Fragment.source.author`, which is the `Authorship` StrEnum
(`self` | `ai` | `other` | `collaborative`). Any document carrying a real
author string therefore fails Pydantic validation and is dropped.

**Every Word document saved by Word carries an author.** So does every PDF
produced by most tools. This makes `--type document` ingest and the new
`creek.upload` MCP tool (#1023) non-functional for their headline use case.

## Reproduction (verified, not theoretical)

```python
from docx import Document
doc = Document()
doc.add_paragraph("A real Word document.")
doc.save("real.docx")   # python-docx stamps core_properties.author='python-docx'
```

Uploading that through `creek.upload` returns:

```
status: refused
reason: ingest failed: [document] failed to assemble fragment from
  .../uploads/real-1-9773879a094d.docx: 1 validation error for Fragment
  source.author / Input should be 'self', 'ai', 'other' or 'collaborative'
```

The same failure occurs on `creek ingest --type document` against main — this
predates #1023 and is not caused by it.

## Root cause

- `creek/ingest/documents.py:286-287` — `if props.author: metadata["author"] = props.author`
  (a free string from `core_properties`).
- `creek/ingest/documents.py:817-818` — `if "author" in fragment.metadata: source["author"] = fragment.metadata["author"]`
  copies it into the `source` dict.
- `creek/models.py:648` — `author: Authorship = Authorship.SELF`, a four-member enum.

The `source["author"]` slot answers "whose views does this stand for?", not
"what name is on the file". Those are different questions and the code conflates them.

## Why this needs design, not a one-line coercion

Mapping any unrecognised string to `Authorship.OTHER` fixes validation but
discards the author's name. The vault already has identity machinery for real
people — `AuthorManifest` / `author_slug` / `AuthorKind` (`creek/models.py:499-535`),
which notably already fails closed on an unrecognised `author_kind`. The right
fix probably routes the extracted name there, or to a distinct metadata field,
rather than deleting it.

Decide and record:
1. Does an extracted document author become `Authorship.OTHER`, or does the
   presence of an external author name imply something stronger?
2. Where does the *name* go so it is not lost?
[...]

Issue #1228, opened 2026-08-08
feat(ingest): pre-ingest redaction hook for staged Adepthood uploads


Geoffe-Ga:
## Context

#1023 adds `creek.upload`, which stages caller-supplied document bytes under
`00-Creek-Meta/adepthood/uploads/` and dispatches them to an ingestor. Its
acceptance criteria asked whether staged upload bytes are subject to a
pre-ingest redaction pass, and required that the answer be stated explicitly
rather than left silent.

**The answer today is no, and there is no hook to wire.** `grep -rn redact
creek/ingest/` returns zero hits. Redaction exists only as:

- `creek/redact/{redactor,scanner,cli_commands}.py` — the CLI surface
- `creek_mcp/tools/redact.py:141-147` — `redact_scan_tool`, which is
  **scan-only** and mutates nothing

There is no seam in `run_ingest`, in `Ingestor.ingest`, or in
`assemble_ingested_fragment`.

## Why this is not simply "call the redactor"

`creek redact --apply` corrupts structured files — it rewrites bodies without
regard to container format. A `.docx`/`.xlsx`/`.pptx` is a ZIP container and a
`.pdf` is a binary object graph; running the existing applier over either
would destroy the document rather than redact it.

So this needs design, not wiring:

- Which staged types are even redactable in place (`.md`/`.txt` yes; ZIP and
  PDF containers need format-aware extraction or a different strategy).
- Whether redaction belongs pre-ingest (mutating staged bytes, which are the
  user's own uploaded document) or post-ingest (on the derived fragment body,
  leaving the staged original intact for RTBF to delete wholesale).
- How a redaction failure surfaces — refusal, or ingest-with-warning.

Post-ingest on the fragment body looks more promising to me on first read: the
staged original is already covered by the RTBF sweep widened in #1023, so it
is deleted rather than sanitised, and the fragment body is plain markdown that
the existing redactor can handle safely.

## Acceptance criteria

- [ ] A decision recorded (ADR or module docstring) on pre- vs post-ingest.
- [ ] A hook at the chosen seam, with binary-container types explicitly
      excluded rather than silently mangled.
- [ ] `creek_mcp/tools/upload.py`'s module docstring updated — it currently
      states outright that staged upload bytes are NOT redacted, and links here.
- [ ] Tests: a redactable `.txt`/`.md` upload has its secrets removed; a
      `.docx` upload is not corrupted.

## Related

[...]

PR #1230, opened 2026-08-08
feat(mcp): creek.upload — stage document bytes and dispatch to the matching ingestor

### LINKED ISSUE #1023: feat(mcp): creek.upload — accept document bytes over MCP, stage them, and dispatch to the matching ingestor
## Role
You are a senior Python engineer working in the creek-tools codebase, following its existing conventions (TDD, MyPy strict, Ruff `select = ["ALL"]`, ≥90% branch coverage, Bandit/pip-audit clean, zero lint/type suppressions).

## Goal
Add a `creek.upload` MCP tool that accepts one document's **bytes** (base64) from a remote consumer, stages it at a stable, deterministic path under an Adepthood uploads staging dir, dispatches it to the matching ingestor by extension via `route_to_ingestor`, and returns the written fragment id(s) — closing the only gap that today makes it **impossible** for Adepthood to feed anything but plain-text journal entries into the vault.

## Context

- **User story**: As an Adepthood user, I want to hand my own documents (PDF, Word, spreadsheets, slides, images, exports, notes) to my vault from my phone, so that `creek.reflect` grounds Higher-Self reflections in more of my actual life — not only what I typed into the app.

- **The gap (this is the blocker).** Every write-side MCP tool on the current surface takes **text or a path**, never bytes:

  | Tool | Content parameter | Can carry a `.pdf`/`.docx`/`.xlsx`? |
  |---|---|---|
  | `creek.journal` | `content: str` | No — markdown text only |
  | `creek.save` | `body: str` | No — text, and it writes a *note*, not an ingestible source |
  | `creek.ingest` | `input_path: str` | No — a **path**, and see below |

  `creek.ingest` cannot be the transport, because its path is confined to the vault root:

  ```python
  # creek-tools/creek_mcp/tools/ingest.py:82-91
  resolved = resolve_within_vault(vault_path, input_path)
  if resolved is None:
      return refusal_response(
          tool=TOOL_NAME,
          ceiling=privacy_tier_ceiling,
          reason=(
              f"input_path {input_path!r} resolves outside the vault root; "
              "the ingest tool only operates on vault-relative paths."
          ),
      )
  ```

  ```python
  # creek-tools/creek_mcp/server.py:492-497
  @server.tool(name="creek.ingest")
  def _ingest(
      source_type: str,
      input_path: str,
      privacy_tier_ceiling: TierCeiling = TierCeiling.OPEN,
  ) -> dict[str, Any]:
  ```

  A remote consumer reaching the vault over the streamable-HTTP transport has **no filesystem in common with the vault host**, so it can never produce an `input_path` that resolves inside the vault. There is no way in for a document today. (This is not a bug in `creek.ingest` — path confinement is correct and must stay. The missing piece is a *staging* capability in front of it.)

- **The precedent that shapes the fix.** `creek.journal` already solved the exact analogous problem for text, and the answer was "stage, then ingest":

  ```python
  # creek-tools/creek/ingest/journal_staging.py:22
  JOURNAL_STAGING_RELDIR = Path("00-Creek-Meta/adepthood/journal")
  ```
  ```python
  # creek-tools/creek_mcp/tools/journal.py:63, 84-89
  _JOURNAL_INBOX = JOURNAL_STAGING_RELDIR
  ...
  def _stage_entry(vault_path, external_id, content, timestamp, tier) -> Path:
      """Write the entry to its stable staged markdown path and return it."""
      staged = vault_path / _JOURNAL_INBOX / f"{_safe_stem(external_id)}.md"
      staged.parent.mkdir(parents=True, exist_ok=True)
  ```

  `creek.upload` is the same shape, generalized from "markdown text" to "arbitrary bytes + a filename": stage at a stable path derived from `external_id`, then run the ingestor `route_to_ingestor` selects instead of hardcoding `markdown`.

- **Acceptance criteria**:
  - [ ] New MCP tool `creek.upload` registered in `creek_mcp/server.py` (next to `creek.journal`, `:357`) with signature `(filename: str, content_base64: str, external_id: str, timestamp: str | None = None, tier: str = "open", privacy_tier_ceiling: TierCeiling = TierCeiling.OPEN)`.
  - [ ] Bytes are staged under a new `UPLOAD_STAGING_RELDIR = Path("00-Creek-Meta/adepthood/uploads")`, declared in `creek/ingest/journal_staging.py` (or a sibling module) as a **single shared constant**, for the same drift reason the journal constant is shared.
  - [ ] The staged filename is derived from `external_id` via the existing collision-free `_safe_stem` pattern (`creek_mcp/tools/journal.py:68-80`), **preserving the original extension** — the extension is what `route_to_ingestor` dispatches on.
  - [ ] Source type is selected by `route_to_ingestor(staged_path)` (`creek/ingest/gdrive.py:922-929`), which already falls back to `generic` for unknown extensions. The caller does **not** pass a `source_type`; an explicit optional `source_type` override may be accepted but must default to the routed value.
  - [ ] Idempotent + edit-in-place: re-sending the same `external_id` with identical bytes is a no-op; changed bytes update in place. Reuse `run_ingest` / `derive_source_key` / `ledger_for_source` (`creek/ingest/pipeline.py`) exactly as `creek.journal` does, so the source ledger keys on the stable staged path.
  - [ ] Tier is honored end-to-end: `write_tier_allowed(tier, privacy_tier_ceiling)` refuses **before** any bytes are decoded or written; the tier lands in the fragment's frontmatter.
  - [ ] **RTBF coverage (privacy-blocking).** The purge engine's staged-plaintext sweep is today scoped to *exactly* the journal dir — `creek/purge/engine.py:842` (`staging_root = (self.vault_path / JOURNAL_STAGING_RELDIR).resolve()`) and `:875`. Extend the sweep to the uploads dir too, so a purged upload's staged bytes are deleted rather than left on disk. Regression test required: purge a fragment sourced from an upload, assert the staged file is gone. See #845 for the identical bug this prevents recurring.
  - [ ] Decoded-size cap enforced with a named constant before writing; oversize returns a structured refusal, not an exception.
  - [ ] `creek.upload` appears in the `creek.handshake` capabilities list so a client can negotiate it (`creek_mcp/tools/handshake.py`); bump `CONTRACT_VERSION` in `creek_mcp/contract.py:18` (currently `"0.2.0"`) per its own docstring rule ("Bump when the tool surface or its semantics change").
  - [ ] Audit log records the tool call with filename/tier/size metadata and `affected_fragment_ids` — **never** the document bytes, mirroring `save_tool`'s "full body never enters the audit log" rule (`creek_mcp/tools/save.py:44-49`).
  - [ ] Redaction: state explicitly in the docstring whether staged upload bytes are subject to the pre-ingest redaction pass, and wire it if the pipeline already offers a hook. If deferred, say so and link a follow-up rather than leaving it silent.

- **Design defaults** (defaults, not mandates):
  - Base64-in-JSON transport, because the MCP tool surface is JSON-RPC and the consumer (Adepthood) has an explicit no-multipart constraint on its own side.
  - One document per call. Batching is a follow-up; single-file keeps idempotency and the tier gate simple.
  - Binary-safe staging: write decoded bytes with `Path.write_bytes`, not text-mode — a `.docx`/`.xlsx`/`.pdf` is a ZIP/binary container and must not be UTF-8 round-tripped.

- **File(s)**:
  - `creek-tools/creek_mcp/tools/ingest.py:82-91` — the path-confinement refusal that makes this issue necessary.
  - `creek-tools/creek_mcp/tools/journal.py:42-89` — the stage-then-ingest pattern to generalize.
  - `creek-tools/creek/ingest/journal_staging.py:22` — the shared staging-dir constant to sit beside.
  - `creek-tools/creek/ingest/gdrive.py:922-929` — `route_to_ingestor`, the extension dispatcher to reuse.
  - `creek-tools/creek/ingest/__init__.py:57-70` — `INGESTOR_REGISTRY`, the eleven target ingestors.
  - `creek-tools/creek/purge/engine.py:842,875` — the RTBF sweep to widen.
  - `creek-tools/creek_mcp/server.py:357-374,492-506` — tool registration.
  - `creek-tools/creek_mcp/contract.py:18` — `CONTRACT_VERSION`.

- **Relevant history**: #754 shipped `creek.journal` (the text-only analogue). #845 fixed the staged-intimate-body RTBF gap for the journal dir — the same class of bug this issue must not reintroduce for uploads. ARCH-001 established "downloaders stage, ingestors ingest"; `creek.upload` is a *push* staging surface in the same architecture.

- **Test coverage**: `creek-tools/tests/test_ingest_registry.py`, `tests/test_ingest_documents.py`, `tests/test_images_ingestor.py`, `tests/test_spreadsheet_presentation_ingestors.py` cover the ingestors themselves; the MCP tool wrappers are covered under `tests/` by tool. No test exercises "bytes arrive over MCP" — that path does not exist. New tests needed: staging + dispatch per extension family, idempotent re-send, tier refusal before write, oversize refusal, binary round-trip fidelity (a real `.docx`/`.xlsx` fixture must survive staging byte-identical), and the purge sweep.

- **Related**: Blocks Adepthood epic https://github.com/Geoffe-Ga/adepthood/issues/1923 and directly blocks its backend sub-issue https://github.com/Geoffe-Ga/adepthood/issues/1924, whose acceptance criterion "forwarded to the vault … as a `creek.ingest` call" is not satisfiable against the current surface. Adjacent: #874 (per-entry classification for the Adepthood contract — an upload's returned tags depend on it), #875 (contract-doc refresh), #845 (staged-plaintext RTBF).

- **Assumptions made**: the consumer is remote (no shared filesystem), per the streamable-HTTP + bearer-token transport in Adepthood's `services/creek_vault_client.py`. If a co-located deployment is also a target, `creek.upload` still works there — it just stages a byte copy.

## Output Format
A single PR that: (1) adds failing tests first, (2) makes them pass, (3) passes the repo's full check suite (Ruff, MyPy strict, Bandit, pytest ≥90% branch, Radon/Xenon), and (4) references this issue with "Closes #N".

## Examples

```python
# creek_mcp/tools/upload.py (sketch)
TOOL_NAME = "creek.upload"
MAX_UPLOAD_BYTES = 10 * 1024 * 1024

def upload_tool(*, vault_path, filename, content_base64, external_id,
                timestamp=None, tier="open",
                privacy_tier_ceiling=TierCeiling.OPEN, consumer="unknown"):
    upload_tier = PrivacyTier(tier)
    if not write_tier_allowed(upload_tier, privacy_tier_ceiling):   # refuse BEFORE decoding
        return refusal_response(...)
    raw = base64.b64decode(content_base64, validate=True)
    if len(raw) > MAX_UPLOAD_BYTES:
        return refusal_response(...)
    staged = vault_path / UPLOAD_STAGING_RELDIR / f"{_safe_stem(external_id)}{Path(filename).suffix.lower()}"
    staged.parent.mkdir(parents=True, exist_ok=True)
    staged.write_bytes(raw)                       # binary-safe: a .docx is a ZIP
    source_type = route_to_ingestor(staged)       # "document" | "spreadsheet" | "image" | ... | "generic"
    result = run_ingest(source_type=source_type, input_path=staged, vault_path=vault_path, ...)
    return {"status": "ok", "source_type": source_type,
            "affected_fragment_ids": result.written_ids, "created_tier": upload_tier.value}
```

```
# Dispatch, end to end
report.pdf   → route_to_ingestor → "document"     → DocumentIngestor      → 01-Fragments/…
budget.xlsx  → route_to_ingestor → "spreadsheet"  → SpreadsheetIngestor   → one fragment per sheet
talk.pptx    → route_to_ingestor → "presentation" → PresentationIngestor  → slides as sections
page.jpg     → route_to_ingestor → "image"        → ImageIngestor (OCR)   → review queue if low confidence
notes.md     → route_to_ingestor → "markdown"     → MarkdownIngestor      → ledger-backed, edit-in-place
mystery.bin  → route_to_ingestor → "generic"      → GenericIngestor       → skipped (binary heuristic)
```

## Constraints
- Do **not** relax `resolve_within_vault` or widen `creek.ingest`'s path confinement — the fix is a new staging tool in front of it, not a weaker boundary.
- Document bytes must never enter the audit log, an exception message, or a log line — only filename, size, tier, and fragment ids.
- Tier gate runs before decoding; an `intimate` upload under `ceiling=open` is refused with zero side effects on disk.
- Do not reintroduce the #845 class of bug: any new staging dir must be inside the purge sweep in the same PR that creates it.
- No lint/type suppressions; named constants for the size cap and the staging dir.
- Scope: this capability only. Per-entry classification tags on the response are #874's job — return whatever the ingest pipeline already produces and do not build a parallel classifier here.

Geoffe-Ga:
Closes #1023.

Every write-side MCP tool took **text or a vault-confined path**, so a remote consumer sharing no filesystem with the vault host had no way to send a document at all. `creek.upload` closes that gap: base64 bytes in, staged flat under `00-Creek-Meta/adepthood/uploads/`, dispatched by extension through `route_to_ingestor`, fragment id out — the same stage-then-ingest shape `creek.journal` already uses for text.

`creek.ingest`'s path confinement is **untouched**, as the issue required. The fix is a staging surface in front of it, not a weaker boundary.

## Three things the obvious implementation gets wrong

Each was found by reading the code rather than by testing, and each is now pinned.

### 1. The tier could not reach a binary fragment at all

`creek.journal` carries tier in the staged markdown's frontmatter. A `.docx` is a ZIP container and carries none. `grep -rn privacy_tier creek/ingest/` returns **zero hits** — no ingestor emits the key, and `run_ingest` had no tier parameter.

This adds exactly one out-of-band channel, `run_ingest(privacy_tier=...)`, merged with `privacy_pass.escalate` so it is **escalate-only**. An uploaded `.md` that declares `intimate` in its own frontmatter is never lowered to the caller's `tier="open"`.

### 2. The RTBF sweep would have silently no-op'd

`ledger_for_source` hard-codes `if source_type != "markdown": return None`, so `attach_origin_key` returned early and a staged `.pdf` got **no `source.origin_key`** — the exact field the purge sweep keys on.

Widening the sweep alone would have produced a green test suite over a broken feature: purged document bytes left on disk, **#845 recreated on a new surface**. A new `ledger_source` override gives uploads their own `upload.jsonl` ledger. `ledger_for_source` itself is byte-identical, so `creek ingest --type document` keeps its current CLI semantics.

Every purge regression test here uses a `.pdf` and a `document`-platform fragment, never a `.md` — a markdown case would pass against the half-fix.

### 3. Identical re-sends would duplicate fragments

`generate_fragment_id` hashes the mtime-derived timestamp, and `write_fragment_idempotent`'s unchanged branch never reassigns `fragment.id`. An unconditional `write_bytes` therefore mints a new id on every identical re-send. `_stage_upload` compares on-disk bytes and skips the write; the test asserts `st_mtime_ns` stability, not just the fragment count.

## Gate order is the feature

- `write_tier_allowed` runs **before** `base64.b64decode` — an `intimate` upload at `ceiling=open` never decodes a byte.
- The size cap runs before any write, with a pre-decode encoded-length guard so a remote caller cannot force a 100 MB allocation.
- The #970 overwrite gate runs **before** staging. Staged bytes have no escalate-only ratchet, so a gate placed after staging returns a correct refusal over already-destroyed content.

The tier-gate test sends a deliberately **undecodable** payload at `tier="intimate"`: if the reason mentions base64, the decode ran first and the ordering is wrong.

## Document bytes never enter the audit log

`summarise_args` passes strings of 64 chars or fewer through **verbatim** — it only collapses longer ones. A small document handed to `append(args=...)` would be logged in full, and the audit log survives the purge meant to erase it. Only `filename`, `external_id`, `tier` and an integer `encoded_len` are recorded. The test uses a payload whose base64 is *under* 64 characters, hitting exactly that pass-through branch.

## An empty ingest is an honest refusal

`GenericIngestor` returns `[]` for binary content and for a claimed `.json`, which `run_ingest` reports as `written=0, errors=[]` — indistinguishable from success. The issue's own sketch returns `status: ok` there. This returns a refusal naming that no fragment was produced.

## Contract bump

`0.2.0` → `0.3.0`, with `SUPPORTED_CONTRACT_MINORS` widened to `(CONTRACT_MINOR, "0.2")`. `CONTRACT_MINOR` is *derived*, so without that widening every existing `/v1` client sending `X-Creek-Contract-Version: 0.2` would start getting `incompatible_version` — an HTTP-surface regression caused by an MCP-side change that altered no `/v1` wire shape. The pre-existing `("0.2", _UNSUPPORTED_STATUS)` test row is the tripwire and still passes.

## Known limitations, stated rather than hidden

- **Real Word documents are currently refused (#1229).** `DocumentIngestor` copies a DOCX's free-text `core_properties.author` into the four-member `Authorship` enum, so any document carrying an author fails validation. I reproduced this end to end. It is **pre-existing** — `creek ingest --type document` fails the same way on `main` — but it does mean the headline use case does not work until #1229 lands. The test fixture clears `core_properties.author` and its docstring says exactly why, so the test pins byte-exact staging rather than masking the bug.
- **No pre-ingest redaction (#1228).** The acceptance criterion said to wire it *if the pipeline offers a hook*. It does not — `grep -rn redact creek/ingest/` returns zero hits, and the existing applier corrupts ZIP containers. Deferred explicitly, with the tool docstring stating it outright.
- **Image uploads are dispatched but not end-to-end tested.** `run_ingest` constructs ingestors with no arguments, so there is no seam to inject a stub OCR engine, and `pytesseract` needs a system binary CI may not have. Dispatch is covered by a pure test; the gap is named in the docstring.
- **No `source_type` override.** The AC permits one; it is deliberately declined because `chatgpt`/`claude`/`discord`/`substack` ingestors are directory-only and an override would let a caller name one that silently discovers nothing from a single file.

## Verification

- `./scripts/check-all.sh` — **12/12 passed**, 8182 tests, 94.10% branch coverage.
- Per-file gate: `creek_mcp/tools/upload.py` 96.99%, `creek_mcp/staged_names.py` 100%.
- `tests/test_mcp_journal.py` passes **unmodified**, proving the `safe_stem` move was behaviour-preserving (the ledger keys on that path, so a change would orphan every existing staged entry).
- The six pre-existing #845 purge tests pass unmodified.
- Zero `noqa` / `type: ignore` / `pragma: no cover` added.


Geoffe-Ga:
## Summary
Adds `creek.upload`, a new MCP tool letting a remote consumer hand base64 document bytes to the vault, staged under `00-Creek-Meta/adepthood/uploads/` and dispatched to the matching ingestor via `route_to_ingestor`, reusing the same stage-then-ledger-ingest shape `creek.journal` already uses for text. It threads a new out-of-band tier channel (`run_ingest(privacy_tier=...)`, escalate-only) and a ledger-source override through the ingest pipeline so binary content gets both a privacy tier and RTBF (`source.origin_key`) coverage, widens the purge engine to sweep both Adepthood staging roots, and bumps the MCP/v1 contract to 0.3.0 while widening `SUPPORTED_CONTRACT_MINORS` so 0.2 clients keep working. The PR body is unusually rigorous about naming and testing the sharp edges.

## Strengths
- Gate ordering (tier check, then size caps, then decode, then the #970 overwrite gate, then stage) is genuinely load-bearing and is asserted from the outside by tests that check disk state, not just the response envelope.
- The RTBF widening is proven with a .pdf and non-UTF-8 fixture rather than a .md, which is exactly the case a half-fix (widened sweep but no origin_key) would pass against.
- `stamp_declared_tier` correctly merges via `privacy_pass.escalate` rather than assignment, with a test proving an already-intimate fragment cannot be downgraded by a caller-declared open tier.
- Contract-version bump is handled carefully: `SUPPORTED_CONTRACT_MINORS` is widened rather than shifted, with the 0.2 row kept as an explicit tripwire so a future narrowing regression is caught.
- `safe_stem`/`safe_suffix` extraction into a shared module keeps `creek.journal` and `creek.upload` from ever computing two different stems for one external id, and journal's existing test suite passes unmodified as a regression signal.
- Real .docx/.xlsx fixtures are generated in-process for tests rather than checked in as binaries, and base64 payloads are derived at runtime rather than hardcoded.

## Security Concerns
🟢 LOW. No confidentiality or integrity guarantee is actually broken. The two Problems below touch privacy-adjacent code paths but neither leaks nor destroys tier-protected content; they are correctness/hygiene gaps, not privacy bypasses. The pre-decode size guard, tier-before-decode ordering, and audit-log redaction were all verified against the actual implementation, not just the PR description.

## Problems
Neither blocks merge, but both look like real omissions given how carefully this PR pins every other sharp edge.

1. **Staged-file identity is keyed on `(external_id, filename)`, not `external_id` alone**, which contradicts the tool's documented contract. `_upload_staged_path` builds `<stem from external_id><suffix from filename>`, and that full path is what feeds the ledger key via `derive_source_key`. Its own docstring is honest about this ("the stable staged path this `(external_id, filename)` maps to"), but `upload_tool`'s docstring states flatly that "the same id updates in place; a new id creates a new fragment." That is only true when the filename's extension stays the same. Re-sending the same `external_id` with a different extension silently mints a second, unrelated fragment instead of updating in place, and does not trip the #970 overwrite gate (the new path has no ledger record, so it reads as a creation). Nothing is overwritten or leaked, but the old fragment and its staged bytes are orphaned, and a caller purging by the `fragment_id` it was handed (rather than by source or vault) will not remove the earlier staged copy — a real gap for a tool whose whole design center is RTBF coverage. Not covered by any test; the existing update-in-place test reuses the same filename throughout. Worth keying identity on `external_id` alone, or documenting this explicitly alongside the other stated limitations.

2. **A "vault unavailable" refusal is not actually side-effect-free.** `SourceLedger.load` returns an empty ledger (not an error) when the vault directory does not exist, so when `vault_path` has been removed, the #970 overwrite gate treats the call as a creation and lets it through to `_stage_upload`, which recreates `00-Creek-Meta/adepthood/uploads/` and writes the caller's bytes into it before `run_ingest` eventually fails and the call reports "vault unavailable." Every other refusal path in this module is deliberately side-effect-free; this one writes new files to a vault the response just said doesn't exist. The existing test for this path asserts only the response shape, not disk state, so the gap is untested. Low real-world likelihood, but worth a `Path.is_dir()` check up front or a fix to the no-side-effect property.

## Code Quality
- Docstrings throughout `creek_mcp/tools/upload.py` are dense but consistently accurate against the code they describe; spot-checked `summarise_args`, `write_tier_allowed`, `refuse_above_ceiling`, and `max_source_tier` against their actual implementations and all matched the claims made.
- `_in_any_staging_root` in the purge engine re-resolves every staging root on every call rather than caching once; harmless at today's scale (two roots, purge is not a hot path) but worth caching if more roots are added. Non-blocking.
- Keeping the `journal_staged_removed` field name for both journal and upload staged-source counts is a reasonable, well-explained trade-off (append-only `purge.jsonl` schema) rather than an oversight.

## CLAUDE.md Adherence
✅ Conventional commit title, issue-linked (Closes #1023).
✅ TDD posture evident from the diff shape — new test files and suite extensions accompany every new code path; no production code appears untested.
✅ Docstrings are Google-style and appear to exceed the 95% interrogate bar on inspection.
✅ No `noqa`, `type: ignore`, or `pragma: no cover` introduced.
⚠️ I did not independently re-run `./scripts/check-all.sh` in this environment (no test execution performed); the PR's self-reported 12/12 and 94.10% coverage claims are taken on the strength of the diff's own test density, not independently reproduced. Worth confirming CI is green before merge if not already.

## Testing
Exceptionally strong overall. The new suites drive the real ledger-backed `run_ingest` end-to-end against temp vaults rather than mocking the pipeline, assert on-disk byte state (not just response envelopes) for every refusal path, use genuine .docx/.xlsx fixtures, and specifically construct payloads to hit narrow branches (base64 under 64 characters to hit the audit pass-through branch; a payload exactly at the pre-decode size boundary to force the post-decode gate). The two gaps described above (extension-mismatch re-send, vault-gone side effects) are the only scenarios found unexercised.

## Documentation
`docs/mcp.md`, `docs/api.md`, `docs/cleaning-and-purge.md`, and both `docs/decisions` ADRs are all updated consistently with the code and with each other (contract version, `SUPPORTED_CONTRACT_MINORS`, gate ordering, RTBF sweep scope). The `docs/contracts/adepthood-v1` bundle (schemas, examples, manifest with recomputed sha256s) is updated in lockstep with the `Capability`/`ErrorCode` docstring changes that feed it.


## Verdict: COMMENTS

No bug blocks merge, no security guarantee is actually broken, and test/doc coverage is exceptional — but there are two substantive, unaddressed correctness gaps (staged-file identity being keyed on (external_id, filename) rather than external_id alone, and a partial disk write on a vault-unavailable refusal) that the PR's own rigor-bar would normally have caught and documented as known limitations. Worth addressing before or shortly after merge.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227470628 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227470628 to see in-depth feedback and continue iterating

Geoffe-Ga:
Both Problems addressed in c96d7f4 rather than deferred — thank you, they were real, and the second one in particular I'd have been unhappy to ship.

**Problem 1 — identity keyed on `(external_id, filename)`.** Fixed, and the diagnosis was exactly right. New `_staged_for_id` recovers the id-wide view, and a re-send under a different extension is now **refused** with a reason naming the bound extension.

I considered making it genuinely update in place and rejected it: the ledger keys on the staged path, and that path must carry the suffix because `route_to_ingestor` dispatches on it, so there is no path stable across extensions to key on. Refusing is the honest resolution — re-typing a document is one explicit purge or one new `external_id` — and it makes the docstring's promise true rather than aspirational.

One thing your write-up led me to that I don't think was in it: the same defect was also a **tier-gate bypass**. `_refuse_unadmitted_overwrite` resolved the prior fragment from the exact target path, so an `open` caller could reach an id whose *intimate* fragment was staged under a different extension — no ledger record at the incoming path, so the gate read a creation and admitted it. The gate now surveys the whole id set via `max_source_tier`. `test_an_intimate_upload_is_not_reachable_through_a_changed_extension` pins it, and asserts the reason is `GENERIC_ABOVE_CEILING_REASON` specifically — if the tier gate regresses, the call falls through to the extension-conflict refusal, which *would* disclose that something exists there.

The conflict check runs strictly **below** the #970 gate for that reason: a conflicting extension is disclosed only to a caller already admitted to the id's content.

**Problem 2 — "vault unavailable" wrote to disk.** Fixed with an `is_dir()` check above both the audit append and staging. I verified the behaviour you described rather than assuming it: with the check stashed, the test's `gone/` directory exists after the refusal. The audit append needed covering too — it mints `00-Creek-Meta/audit/` the same way.

All three new tests fail against the parent commit and pass against this one (verified by stashing `upload.py` and re-running, not by inspection).

**Not addressed, deliberately:** `_in_any_staging_root` re-resolving roots per call. You marked it non-blocking and it is — two roots, purge is not a hot path — and I'd rather not touch the purge engine again in a PR whose gate ordering is the load-bearing part.

On the CLAUDE.md note: CI is green (10/10) and `./scripts/check-all.sh` re-run locally after these fixes gives 12/12, 8185 passed, 94.10% coverage.

Geoffe-Ga:
## Summary
Adds `creek.upload`, a new MCP tool that lets a remote Adepthood consumer hand over a document's base64 bytes for staging + ledger-backed ingest, mirroring the existing `creek.journal` shape for text. The PR also widens the RTBF purge sweep and the ingest pipeline to give non-markdown, non-text content a ledger identity and a declared-tier channel, and bumps the MCP↔HTTP contract from `0.2.0` to `0.3.0` while explicitly keeping `0.2` served. A second commit on the same branch already found and fixed two real bugs from a self-review (identity keyed on extension instead of `external_id`, and a "vault unavailable" refusal that wasn't side-effect-free).

## Strengths
- Gate ordering is deliberate and independently tested from the outside: `write_tier_allowed` runs before `base64.b64decode` (an above-ceiling upload never decodes), the encoded-length cap runs before the decoded-length cap, and the #970 overwrite gate runs above `_stage_upload` so a refusal never leaves an already-destroyed staged document behind.
- The idempotent-write short circuit (`_stage_upload` comparing on-disk bytes before writing) directly prevents a real bug: since `generate_fragment_id` hashes an mtime-derived timestamp, an unconditional `write_bytes` on an identical re-send would have minted a duplicate fragment. This is pinned with an `st_mtime_ns` assertion, not just a fragment-count check.
- RTBF coverage is genuinely end-to-end: `resolve_ledger`/`UPLOAD_LEDGER_SOURCE` give uploads a real `source.origin_key`, `ADEPTHOOD_STAGING_RELDIRS` widens the purge sweep to a second root without weakening the existing containment guard, and the non-recursive staging-dir walk explicitly guards against a stray subdirectory raising `IsADirectoryError` and aborting a whole-vault purge (test: `test_a_subdirectory_in_a_staging_dir_does_not_abort_a_vault_purge`).
- Audit-log hygiene: document bytes never reach `MCPAuditLog` (only filename/external_id/tier/`encoded_len`), tested specifically with a payload short enough to hit `summarise_args`'s verbatim pass-through branch rather than its truncation branch.
- Contract-version handling is careful: `SUPPORTED_CONTRACT_MINORS` is widened, not shifted, with an explicit tripwire test (`("0.2", _UNSUPPORTED_STATUS)`) so a future minor bump that narrows the window instead of widening it fails loudly.
- The PR body honestly documents known limitations (DOCX-author enum mismatch #1229, no pre-ingest redaction #1228, untested OCR leg) instead of hiding them, and the follow-up commit shows a real self-review cycle that caught two genuine bugs before this ever reached a human reviewer.

## Security Concerns
🟢 LOW — No blocking issues found. Path handling is sound: `safe_stem` is a hash-suffixed slug (collision-free and injective), `safe_suffix` strips everything but alnum+dot and is length-capped, and the purge engine's containment guard (`_in_any_staging_root`) still resolves against the vault root before deleting anything. The pre-decode base64 length guard correctly prevents a remote caller from forcing a large allocation before the size is known to be over cap.

## Problems
None blocking.

## Code Quality
🟡 Non-blocking correctness gap in `creek_mcp/tools/upload.py`: `_stage_upload` writes the caller's bytes to disk *before* `_ingest_outcome` learns whether the ingest actually produced a fragment. When it doesn't (`written == 0` or `result.errors`, e.g. an unrouted extension or content the binary heuristic flags), the tool correctly refuses — but the orphaned bytes are left on disk with **no ledger record**.

A later retry for the *same* `external_id` under a *different, correct* extension then runs into `_refuse_extension_conflict` (`upload.py:1049`), which treats any staged file at that stem as a conflicting "existing fragment" regardless of whether it actually has a ledger record — `_staged_for_id` doesn't filter on that. The refusal text says "Purge the existing fragment or use a new external_id," but there is no fragment to purge (the earlier attempt never produced one), so that half of the remedy is misleading. The only real recovery is abandoning the `external_id`, which quietly breaks the tool's own idempotency contract ("the same id updates in place") for this failure path. It's narrow and has a working escape hatch (a new `external_id`), so it doesn't block merge, but it isn't covered by any test in `test_mcp_upload.py`, and either scoping `_refuse_extension_conflict` to ledger-backed staged files or cleaning up orphaned staged bytes on a failed ingest would close it.

## CLAUDE.md Adherence
✅ Feature branch (`issue/1023-mcp-upload`), Conventional Commits-style messages (`feat(mcp): …`, `fix(mcp): …`).
✅ PR body reports `./scripts/check-all.sh` 12/12 and 94.10% branch coverage, above the repo's ≥90% gate; I could not independently re-run CI in this sandbox (the `gh pr checks` call was blocked by the environment), so this is taken on the PR's own reporting rather than independently verified.
✅ TDD workflow evident from the diff shape (new/updated tests alongside every behavior change, including a dedicated second-commit test set for the two self-found bugs).
✅ No `noqa` / `type: ignore` / `pragma: no cover` introduced, per the PR's own claim and consistent with what's visible in the diff.

## Testing
Strong. The suite drives the real `run_ingest` pipeline rather than mocking it, uses in-process-generated real `.docx`/`.xlsx` binaries (no checked-in binary fixtures), and specifically targets the failure modes that would make a superficial fix look correct: a `.pdf`/`document`-platform fixture for every purge test (never `.md`, which would pass against a half-fix), an `st_mtime_ns` assertion for the idempotent-write path, and a byte-level (not text-glob) leak sweep so a binary staged document can't hide from the assertions. The one gap is the orphaned-staged-file/extension-conflict interaction noted above.

## Documentation
Thorough and kept in lockstep with the code: `docs/mcp.md`, `docs/cleaning-and-purge.md`, both contract ADRs, and the JSON contract fixtures/schemas were all updated for the `0.3.0` bump and the new tool, including an explicit changelog row explaining *why* the version moved (MCP surface growth, not a `/v1` shape change) so a future reader isn't left guessing.



## Verdict: COMMENTS

Nothing here blocks merge — gate ordering, RTBF coverage, idempotency, and audit hygiene are all correctly implemented and well-tested. But there's a genuine, untested edge case worth the author's attention before or shortly after merge: when an upload's ingest fails to produce a fragment, the staged bytes are left on disk with no ledger record, and a later retry under a different extension is refused by `_refuse_extension_conflict` with a "purge the existing fragment" remedy that doesn't apply since no fragment exists — quietly breaking the external_id idempotency contract for that narrow path.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227554313 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227554313 to see in-depth feedback and continue iterating

Geoffe-Ga:
Fixed in 1cd6ae1. You scored this non-blocking; I think it's worse than that, and the reason is the point of the PR.

You noted the orphaned bytes are left "with **no ledger record**." That's the sharp end: no ledger record means no `source.origin_key` on any fragment, and the RTBF sweep finds staged bytes *only* through that key (`_purge_staged_source_entry`). So those bytes were unreachable by every purge the vault offers — a copy of the caller's document surviving the erasure meant to remove it. That's the #845 class of hole, reintroduced on the refusal path of the very tool added to close it. The misleading refusal text was the symptom I'd have noticed; the unpurgeable copy is the one that mattered.

`_discard_unreferenced` now deletes a staged file whose ingest produced no fragment. It keys on the **ledger**, not on "did this call write the file", so a re-send whose ingest fails over an already-recorded staged file leaves that file alone — its fragment still references it.

I also took your second suggestion rather than choosing between them, because they close different halves. `_refuse_extension_conflict` is now scoped to ledger-backed staged files, so even a pre-existing orphan (from a crash, or from a vault written before this commit) can't produce a refusal whose remedy is impossible to carry out.

You were right that the recovery path was the real damage: abandoning the `external_id` was the only way out, which breaks "the same id updates in place" for precisely the callers who need a retry. `test_a_corrected_extension_after_a_failed_ingest_is_accepted` pins that end-to-end — refused `.bin` first, then the same id as `.txt` must succeed. Against the parent commit it fails with `- ok / + refused`.

Both new tests verified failing against the parent by stashing `upload.py`, not by inspection.

Gate re-run: 12/12, 94.10% coverage.

On the CI note — `gh pr checks` was blocked in your sandbox, so for the record: CI is 10/10 green on this branch, and has been for each of the three pushes.

Geoffe-Ga:
## Summary
Adds `creek.upload`, a new MCP tool that lets a remote Adepthood consumer hand over base64-encoded document bytes for staging and ledger-backed ingest, mirroring the existing `creek.journal` shape. The PR widens the RTBF purge sweep to cover the new `uploads/` staging root, adds an out-of-band privacy-tier channel and a ledger-source override to `run_ingest`, and bumps the MCP contract to 0.3.0 while keeping 0.2 served. Two additional commits on top of the base implementation already fix self-identified review findings (identity forking across extensions, and orphaned unpurgeable staged bytes on a failed ingest).

## Strengths
- Gate ordering (tier check → size caps → decode → #970 overwrite gate → staging write) is not just implemented correctly, it is *asserted from outside* with tests that use deliberately undecodable payloads and check that no bytes hit disk when a gate should have fired first.
- The RTBF/purge widening is the most security-sensitive part of this change and gets the most careful treatment: `ADEPTHOOD_STAGING_RELDIRS` centralizes every staging root, the containment guard (`_in_any_staging_root`) is tested against path traversal (`../x`) and cross-root escapes, and the non-recursive wipe walk explicitly guards against a stray subdirectory aborting a whole-vault purge (with a test proving the guard is load-bearing).
- The two follow-up commits on this same PR show real self-review discipline: they found and fixed (a) an identity fork when an `external_id` is re-sent under a different file extension, orphaning the original fragment's staged bytes from any purge, and (b) staged bytes left behind — and thus unreachable by any purge — when an ingest fails after staging. Both are exactly the class of bug (#845) this tool exists to avoid reintroducing, and both now have regression tests that fail against the parent commit.
- Idempotency correctness is subtle and well-handled: `_stage_upload`'s identical-bytes short-circuit exists specifically because `generate_fragment_id` hashes the staged file's mtime, so a naive `write_bytes` would mint a duplicate fragment on every identical re-send. Verified via an `st_mtime_ns` stability assertion, not just a fragment-count check.
- The privacy-tier channel (`stamp_declared_tier`) is correctly escalate-only, preventing a caller's declared tier from downgrading a tier already declared inside an uploaded `.md`'s own frontmatter — pinned with tests including the `UNCLASSIFIED` rank boundary.
- The contract version bump reasoning (0.2.0 → 0.3.0, with `SUPPORTED_CONTRACT_MINORS` widened rather than shifted) correctly protects existing `/v1` clients from an MCP-only change, and this is pinned with an explicit `0.2`-still-served test row plus a `0.3` "current minor" row.
- Known limitations (DOCX author-enum bug #1229, no pre-ingest redaction #1228, untested OCR leg for image uploads) are stated plainly in the tool's own docstring rather than hidden, and each is backed by a `grep`-verified claim in the PR description.

## Security Concerns
🟢 LOW — I looked specifically for a way an exception during ingest (e.g., a corrupt `.docx` triggering a `python-docx`/zip failure) could leave staged bytes on disk with no ledger record, unreachable by any purge — the exact bug class the third commit's `_discard_unreferenced` was built to close, but only for the `result.errors` / `written == 0` paths, not necessarily for an uncaught exception from `ingestor_cls().ingest(...)` inside `_ingest_outcome` (which only catches `FileNotFoundError`). I traced this into `creek/ingest/base.py`: `Ingestor.ingest()` wraps both `discover()` and each document's `parse()` in broad `except Exception` handlers that append to `IngestResult.errors` rather than propagating, so a malformed `.docx`/`.xlsx` surfaces as `result.errors` and correctly triggers `_discard_unreferenced`. No exploitable gap found here; noting it because it's the kind of thing worth a second pair of eyes on if `run_ingest`'s exception surface ever changes.
- Base64 decoding uses `validate=True` and a pre-decode length guard sized precisely off `MAX_UPLOAD_BYTES`, correctly preventing a forced large allocation from an oversized encoded string before any bytes are materialized.
- Extension/suffix handling (`safe_suffix`, `MAX_SUFFIX_CHARS`) bounds a caller-controlled string before it becomes part of a filesystem path, and `safe_stem`'s hash suffix prevents slug collisions between distinct external ids.

## Problems
None found that block merge.

## Code Quality
- `_existing_tier` in `creek_mcp/tools/upload.py` is a near-identical duplicate of `creek_mcp.tools.journal._existing_tier`. The author explicitly flags this in the docstring as a deliberate, deferred de-duplication (citing the read-gate manifest's AST check as the reason not to do it now) — reasonable to defer, but worth tracking as a follow-up so the two copies don't silently drift.
- The `docs/mcp.md` table row for `creek.upload` is an extremely dense single paragraph (comparable in style to the existing `creek.journal` row, so at least internally consistent, but it's approaching the point where a reader would be better served by a linked subsection).
- `_refuse_extension_conflict`'s reason string discloses which extension(s) are already bound to an `external_id` to a caller that passed the #970 overwrite gate. This is a deliberate, well-reasoned choice (the caller is already ranked as admitted to that content's tier before this check runs), not a defect.

## CLAUDE.md Adherence
✅ TDD workflow — the PR's own commit messages state new tests fail against the parent commit and pass against the fix, consistent with the required Red-Green cycle.
✅ Coverage — PR states 94.10% aggregate branch coverage (≥90% threshold) and 96.99%/100% per-file coverage on the new modules; not independently re-run in this review.
✅ No shortcuts — PR explicitly states zero `noqa`/`type: ignore`/`pragma: no cover` added, and the diff supports that (no bypass comments present in the reviewed code).
✅ Commit conventions — three commits, each a focused `feat`/`fix` with a clear rationale; feature branch (`issue/1023-mcp-upload`), not a direct-to-main commit.
⚠️ I did not independently execute `./scripts/check-all.sh` or check live CI status for this review (tool access to `gh pr checks`/`gh api` was not available in this session); the PR body's claimed 12/12 gate pass should be confirmed by CI before merge if not already green.

## Testing
Exceptionally thorough for a security/privacy-sensitive surface: gate-ordering is asserted via side-effect absence (no bytes on disk, no audit entry) rather than only via response shape; the RTBF sweep tests use raw non-UTF-8 `.pdf` bytes specifically so a markdown-only test could not accidentally pass against a half-fix; idempotency is checked via `st_mtime_ns` stability, not just fragment count; and the two self-review-driven regression tests (extension-fork, orphaned-bytes-on-failed-ingest) each demonstrably fail against the pre-fix commit. This comfortably clears the "insufficient testing is blocking" bar.

## Documentation
The tool's own module docstring, the `docs/mcp.md` table entry, `docs/cleaning-and-purge.md`, and two ADRs are all updated in the same PR and stay consistent with each other and with the code (e.g., the gate-ordering description in `upload.py`'s docstring matches the actual code order and the test that pins it). Known limitations are documented rather than glossed over.


## Verdict: LGTM

Every write path is gate-ordered correctly and the ordering is independently pinned by tests (tier check before base64 decode, size caps before any write, the #970 overwrite gate before staging); the RTBF purge sweep is widened with real containment guards and byte-level (not text-only) tests; two rounds of self-review already caught and fixed the identity-fork and orphaned-bytes edge cases with regression tests. I traced the one path that looked like it could reintroduce an unpurgeable-bytes bug (an exception raised inside DocumentIngestor.parse for a corrupt .docx) and confirmed creek/ingest/base.py's Ingestor.ingest() already wraps parse() in a broad try/except that routes failures into IngestResult.errors, so upload_tool's _discard_unreferenced still fires. No blocking defect found; remaining items are nits the author already disclosed.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: LGTM
**Action**: pull comment 5227643378 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: LGTM

Issue #790, opened 2026-07-02
perf(link): eddy/thread DBSCAN + paradox are O(N²) pure-Python — creek fill can't complete at 35k


Geoffe-Ga:
## Problem

`creek fill` cannot complete on a large vault (~35k fragments). Three linking/report sub-steps do **all-pairs O(N²) work in pure Python**, and at 35k fragments each is ~1.2B×384-dim cosine ops in interpreted code — effectively *days*, not minutes. Observed live: `link/eddies` ran 3.5h at 98% CPU with no forward progress.

The #715/#722 scale epic vectorized **resonance discovery** (`EmbeddingLinker.find_resonances` → blocked-numpy top-k), but left these three on the old path:

| Step | Code | Cost at 35k |
|---|---|---|
| **Eddy clustering** | `creek/link/eddies.py:450` — `_dbscan` neighbour cache; `_neighbours` (450, 470-488) double-loops pure-Python `_cosine_distance` | O(N²) × 384-dim, interpreted |
| **Thread clustering** | `creek/link/threads.py` — DBSCAN over cached embeddings (`link_engine.py:281`), same pure-Python pairwise | O(N²) |
| **Paradox report** | `creek/generate/paradox.py:283` — `itertools.combinations(fragments, 2)` + per-pair pure-Python `_cosine_similarity` | O(N²) |

The `# OPS-004: neighbour-cache build is O(N²)` comment at `eddies.py:449` already flags the eddy case.

## Impact

`creek fill` is unusable on any vault approaching 35k fragments — the exact scenario the vault-prod-readiness work targets. Classification completes fine; the vault then cannot be linked/reported.

## Fix strategy (behavior-preserving)

Mirror the #722 blocked-numpy approach already living in `EmbeddingLinker` (`embeddings.py:740-844`):

1. **Eddies + threads DBSCAN** — normalize the embedding matrix once, compute similarities one **row-block at a time** (`normalized[start:stop] @ normalized.T`, never the full N×N), and collect each point's neighbours as `sim >= 1 - eps` (cosine distance ≤ eps). This yields the **identical** neighbour sets the pure-Python loop produces, so DBSCAN clusters are unchanged — a pure speed fix, **no config toggle** (unlike top-k, which changes results). Extract a shared vectorized-neighbour helper so eddies and threads share one implementation.
[...]

Issue #718, opened 2026-06-28
feat(threads): bulk thread-page generator for 02-Threads


Geoffe-Ga:
## Role
Senior Python engineer in this repo's `creek/link/threads.py`, `creek/compile/`,
and the `creek/cli.py` link/compile surface.

## Goal
Populate **02-Threads** in bulk: a runnable command materialises a thread
markdown **page** for every detected thread, instead of today's manual
one-at-a-time `creek compile --target-kind thread --target-id … --target-title …`.

## Context
- Parent epic: #713.
- `ThreadDetector` (`creek/link/threads.py`) detects threads during the linking
  pipeline and writes `threads:` wiki-links onto fragments, but the thread
  **pages** are only written when an operator runs `creek compile` per thread.
  On a real vault that leaves 02-Threads near-empty (audit: 2 files) while
  fragments carry thread links pointing at non-existent pages.
- `VaultWriter.write_thread` (`creek/vault/writer.py:746`) is the page writer.
  Deterministic, no LLM.

## Output Format
A single PR:
- [ ] A runnable path (e.g. `creek link --method threads --write-pages`, or a
  `creek compile --all-threads` mode, or folding page-writing into the threads
  linker) that writes one page per detected thread via `write_thread`.
- [ ] Idempotent: re-run overwrites stable pages; threads that vanish are not
  left as orphan pages (or are clearly handled).
- [ ] Tests: detected threads on a fixture vault produce the expected pages;
  fragment `threads:` links resolve to real pages; re-run is a clean no-op.

## Examples
```text
$ creek link --method threads --write-pages --vault V
Detected 184 threads; wrote 184 pages under 02-Threads/{Active,Dormant,Resolved}/
```

## Constraints
**Scope fence:** wire bulk page-writing over the EXISTING detector + writer. Do
NOT change thread detection semantics (time window, topic consistency). Decide
and document where pages route (Active/Dormant/Resolved) using the existing
thread-status logic.

**Anti-bypass (verbatim, non-negotiable):**
> No `noqa`, `# type: ignore`, `pylint: disable`, `eslint-disable`, or equivalent
> linter/type-checker silencers. Fix the root cause. The only exception is the
> documented 4-line escape hatch (third-party library bug / language-version
> compatibility / benchmarked performance necessity / generated code) — and it
> must include the reason, a reference URL, an alternative considered, and a
> review date. See the `max-quality-no-shortcuts` skill.

## Definition of Done (stay-green)
[...]

PR #1233, opened 2026-08-08
fix(link): bound cluster degeneration in message streams

### LINKED ISSUE #880: fix(link): eddy/thread clustering degenerates at message scale — one 'Messages' cluster swallows 87% of the vault
## Role

You are a pipeline engineer with clustering experience working in `creek-tools/` (Python, numpy-only clustering, mypy strict, TDD per `creek-tools/CLAUDE.md`).

## Goal

Stop eddy and thread detection from collapsing a high-volume message corpus into a single degenerate mega-cluster that swallows ~87% of the vault.

## Context

Evidence from the demo vault `creek-demo-2026-06-07` (35,330 fragments, 30,795 of them Discord/chat messages):

- `link --method eddies` produced 17 eddies, one of which — `03-Eddies/2023-03-30-Messages.md` — has `fragment_count: 30795` (every message in the vault), `description: ''`, `threads: []`.
- `link --method threads` (run 2026-07-22) produced 195 threads, one of which — `Messages`, `thread-ebdfeb4c` — has `fragment_count: 30108` with `first_seen: 2020-09-26`, `last_seen: 2026-05-20`, `description: ''`.
- Root cause shape: messages form a temporally-continuous, semantically-similar stream. Thread detection unions any topic-consistent pair (frequency overlap + cosine ≥ 0.6) inside a 30-day sliding window (creek/link/threads.py), so seven years of continuous chat chains transitively into one component. Eddy DBSCAN (eps 0.3 cosine, min_samples 5, creek/link/eddies.py) similarly merges the dense message blob into one cluster.
- Consequences: the compiled layer is useless for retrieval over the message corpus; `eddies:` frontmatter links on 30k fragments all point at one page; draft/mine resonance surfaces get no topical structure; state report's "Active eddies" is dominated by one meaningless entry. Also every mega-page has an empty `description`.
- Related: #857 (open, O(n²) per-pair cosine in threads — same code path; this run completed at 35k in minutes, so severity there may be lower than feared, but a fix here should keep it in mind), #790 (closed, DBSCAN perf), #730 (closed, wiki-link aliases).

## Proposed Solution

Design-space (pick and document in the PR; an ADR-style note is warranted, per architectural-decisions):

1. **Cluster-size ceiling + recursive split** — any cluster above N fragments (e.g. 500, config-keyed) is re-clustered at a tighter eps/similarity until subclusters fit or a floor is hit; unsplittable remainder becomes noise, not a page.
2. **Segment message streams first** — pre-partition `source.kind`-continuous streams (same channel/interlocutor) into conversation episodes (time-gap segmentation) and cluster episodes, not raw messages.
3. **Per-source thresholds** — stricter similarity thresholds for message-platform fragments than for essays/notes.

Also: generated eddy/thread pages should never ship an empty `description` — derive a one-line description from the distinctive-terms title logic if nothing better exists.

## Acceptance Criteria

- [ ] On a fixture reproducing the pattern (dense message stream + a few distinct topics), no single eddy/thread contains more than the configured ceiling fraction of the corpus
- [ ] Distinct topics still surface as separate eddies/threads (no regression on the existing detection tests)
- [ ] Generated pages have non-empty `description`
- [ ] Re-running link on the demo-scale vault stays within the current wall-clock budget (do not regress #790/#857)

## Constraints

- Pure numpy (no scikit-learn) per the existing module conventions
- Config-keyed thresholds with sane defaults; no breaking changes to existing page frontmatter schema
- Follow stay-green TDD

Geoffe-Ga:
Closes #880.

One eddy and one thread each swallowed **87% of the demo vault**, both titled "Messages". Three mechanisms compound, and the third explains the name.

## 1. DBSCAN's precondition is violated, so no threshold value fixes it

`_expand_cluster` (`creek/link/eddies.py:493`) does unbounded transitive density-reachability across the whole eps-graph. Short chat messages are semantically homogeneous, so adjacent messages sit inside 0.30 cosine distance of each other and every message is a core point — the graph over the 30,795 Discord fragments is a **single connected component**.

DBSCAN requires clusters separated by low-density regions. A continuous chat stream has no separator to find, so DBSCAN returns the whole stream as one cluster — correctly, by its own definition. **This is why tuning `eps` cannot work**, and why `creek/link/segments.py` partitions message-platform fragments into conversation episodes instead.

Non-stream fragments stay in one shared domain, so cross-source threads keep working and all 108 existing tests in `test_link.py` are untouched.

## 2. Threads degenerate independently

The 30-day window at `threads.py:453` bounds *pairing*, but `uf.union` at `:456` is unbounded — overlapping windows take the transitive closure and chain 2020-09-26 through 2026-05-20 into one root of 30,108. `:505-506` additionally **fails open**, unioning on frequency alone when either embedding is missing.

Measured before the fix: thread sizes `[600, 12, 12, 12]`, largest share **0.943**. Fixing only the eddy half would have left this untouched.

## 3. The name was a filename leak — and the docstring already promised the fix

`DiscordIngestor.generate_frontmatter` emits no `title`, so `base.py:447` setdefaults it to `Path(source_path).stem`, and `discover` only ever reads `<channel>/messages.json`. **All 30,795 fragments carry `title: messages` on disk.** `_generate_title` then counts words over titles and nothing else — `Fragment` has no body field, so the title is the only text the linker can see — and `most_common(3)` returns `[("messages", 30795)]`.

The damning detail: that function's docstring claims *"TF-IDF-lite: content-word frequency inside the cluster scaled by inverse document frequency."* Real IDF gives "messages" a document frequency of 30795/35330 and suppresses it entirely. **The documented algorithm would not have produced this bug. The implemented one contains no IDF whatsoever.**

`creek/link/naming.py` implements the IDF that was promised, plus a structural fallback (dominant `source.channel` + date span → "Aspiring Drag Queens, Mar 2023") for when no term survives — exactly the message case — and a never-empty `description`. `grep -rn description creek/link/*.py` previously returned zero hits, which is why 212 of 214 linker-written pages shipped `description: ''`.

## The size cap is a guardrail, not the cure

`creek/link/cluster_limits.py` says so in its own docstring, because a reviewer must not read it as the fix. Recursively tightening eps on a 30,795-member component shatters a continuous semantic manifold into ~60 arbitrary slices with no interpretable identity, all still named from the same identical titles and all colliding on the `f"[[{eddy.title}]]"` wiki-link — **sixty meaningless pages instead of one**. It is kept as the machine-checkable invariant behind acceptance criterion 1 and as protection for corpora we haven't seen.

Applied strictly outside `_dbscan`, so the #790 brute-force-equivalence contract stays green.

## The corpus is the crux

The existing scale generator (`test_link_scale_790.py:102-126`) builds **one-hot vectors**: cross-cluster cosine is exactly 0.0, so transitive bridging is arithmetically impossible; cluster sizes are exact by construction; titles are all distinct; fragments are `UNCLASSIFIED`, which `_frequency_overlap` discards, yielding zero threads. **A test built on it would pass vacuously whether or not the bug existed.**

`_message_stream_corpus` uses a seeded random walk on the unit sphere, so adjacent messages land inside the eps radius and the graph becomes one large-diameter component — reproducing the real failure. Not seed-fragile: seeds 880/1/42/2026/7 all give exactly one cluster of 600.

## Acceptance criteria

| AC | Test | Red proof |
|---|---|---|
| No eddy over ceiling | `test_no_eddy_exceeds_the_configured_corpus_fraction` | `assert 600 <= 63` — 94.3% |
| No **thread** over ceiling | `test_no_thread_exceeds_the_configured_corpus_fraction` | sizes `[600,12,12,12]`, `600 <= 63` false |
| Distinct topics still surface | `test_planted_topics_survive_as_distinct_eddies_and_threads` | see below |
| Non-empty `description` | `test_every_emitted_eddy_and_thread_has_a_non_empty_description` | `AssertionError: assert ''` |
| Wall-clock not regressed | measured, below | — |

**Honest note on the distinct-topics test:** it is **green on unmodified main** and is not a bug reproduction — it is the anti-overcorrection guard. It was proven to have teeth empirically instead: patching `SplitPolicy` to a "shatter everything" configuration makes *this* test fail while **both** ceiling tests still pass. It is the only test in the module that can tell a fix from a shredder.

**Weakest test, flagged rather than dressed up:** `test_recluster_work_is_bounded_by_split_depth` fails on main with a `TypeError` at construction, not on an assertion — there is no pre-fix behaviour for it to contradict. It is kept because it is the only test anywhere bounding the splitter's *aggregate* work (measured: 3 calls, exactly 3 × 600 members, linear in `max_depth`), which the perf criterion rests on. If a reviewer judges it padding, it's the one to drop.

## Performance: faster, not slower

Best-of-3, same machine, identical harness:

| Fragments | threads before | threads after | eddies before | eddies after |
|---|---|---|---|---|
| 636 | 0.150s | **0.077s** | 0.0042s | **0.0039s** |
| 3,036 | 0.862s | **0.435s** | 0.028s | **0.018s** |
| 9,036 | 2.710s | **1.348s** | 0.126s | **0.054s** |

Segmentation is O(n²/d) rather than O(n²), so this improves #857 rather than regressing it. Note the defect **worsens with scale**: at 3k and 9k fragments the entire corpus collapsed into a single thread (`thread_n == 1`).

## Also fixed

- **`Eddy.id` is now content-stable**, mirroring `_stable_thread_id` (#718 did this for threads). It was `uuid4()` per run — latent, but this PR makes it acute by producing many more, smaller eddies, and a vault that rewrites its eddy filenames on every `creek link` is not usable for retrieval.
- **`DiscordIngestor` emits a real title**, so the vault stops manufacturing 30k identical strings at source. Safe: `generate_fragment_id` keys on source_path/timestamp/content, not title, so no fragment ids churn. The naming layer is still required independently — existing vaults keep `title: messages` until re-ingested.

## Out of scope, flagged

`creek_mcp/tools/link.py`'s `_VALID_METHODS` omits `"threads"` entirely, so the thread half of this fix is unreachable via MCP, and the tool's response exposes no eddy/thread counts. Pre-existing and unrelated to the degeneration; widening the MCP contract deserves its own review.

## Verification

`./scripts/check-all.sh` — **12/12**, 8287 passed, 94.16% coverage. New modules: `cluster_limits.py` 100%, `naming.py` 97.6%, `segments.py` 96.5%. Zero suppressions.

### DIFF
--- a/creek-tools/README.md
+++ b/creek-tools/README.md
@@ -161,7 +161,7 @@ read the threat model below before doing anything else.
 | `llm`            | `LLMRoutingConfig`     | Per-stage provider+model routing (`default` / `classification` / `generation` / `frontend` + a `writing_desk` role map), or a legacy flat `LLMConfig` block. See [LLM providers](#llm-providers). |
 | `embeddings`     | `EmbeddingsConfig`     | Sentence-transformer model, similarity thresholds. |
 | `ocr`            | `OCRConfig`            | Tesseract path, languages, PSM mode. |
-| `linking`        | `LinkingConfig`        | Embedding/temporal/eddy thresholds. |
+| `linking`        | `LinkingConfig`        | Embedding/temporal/thread/eddy thresholds, message-stream segmentation, and the cluster-size guardrail. See [ADR-0008](docs/architecture/ADR/0008-bounding-cluster-degeneration-in-message-streams.md). |
 | `classification` | `ClassificationConfig` | Auto-classify sources, confidence threshold, review-required sources. |
 | `context`        | `ContextConfig`        | How non-user content (others' messages, collaborative docs) is handled. |
 | `redaction`      | `RedactionConfig`      | Pattern enable list, exclusion globs, allow-list. |
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -1914,7 +1914,8 @@ def _format_link_summary(summary: LinkSummary) -> str:
             f"{summary.eddies_detected} eddy(ies) detected, "
             f"{summary.eddies_written} eddy file(s) written to 03-Eddies/, "
             f"{summary.member_fragments_updated} fragment(s) updated with "
-            f"`eddies:` wiki-links."
+            f"`eddies:` wiki-links"
+            f"{_format_cluster_stats(summary)}."
         )
     elif summary.method == "temporal":
         body = (
@@ -1927,12 +1928,42 @@ def _format_link_summary(summary: LinkSummary) -> str:
             f"{summary.threads_written} page(s) written to "
             "02-Threads/{Active,Dormant,Resolved}/, "
             f"{summary.member_fragments_updated} fragment(s) updated with "
-            "`threads:` wiki-links."
+            "`threads:` wiki-links"
+            f"{_format_cluster_stats(summary)}."
         )
     else:
         msg = f"Unknown link method: {summary.method!r}"
         raise ValueError(msg)
     return f"[bold green]{body}[/bold green]"
+
+
+def _format_cluster_stats(summary: LinkSummary) -> str:
+    """Render the clustering-health clause for a link summary.
+
+    Issue #880: the largest cluster is the one number that tells an
+    operator whether detection degenerated — a single eddy holding 87% of
+    the vault used to be invisible from the CLI. Splits and discards are
+    reported only when they happened, so an ordinary run stays terse; a
+    discard is data loss (those fragments carry no wiki-link at all) and
+    therefore always named when non-zero.
+
+    Args:
+        summary: The link summary being rendered.
+
+    Returns:
+        A clause to append to the method's sentence, starting with ``", "``,
+        or an empty string when nothing was detected.
+    """
+    if summary.largest_cluster_fragments == summary.oversized_discarded == 0:
+        return ""
+    parts = [f"largest cluster: {summary.largest_cluster_fragments} fragment(s)"]
+    if summary.clusters_split:
+        parts.append(f"{summary.clusters_split} oversized cluster(s) re-clustered")
+    if summary.oversized_discarded:
+        parts.append(
+            f"{summary.oversized_discarded} fragment(s) discarded as unsplittable",
+        )
+    return ", " + ", ".join(parts)
 
 
 @app.command()
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -473,6 +473,141 @@ class LinkingConfig(BaseModel):
     introduced by FEAT-022.
     """
 
+    # ---- Detector thresholds (issue #880) ----
+    #
+    # Every default below is the exact value of the module-private constant
+    # it replaces, so a vault whose creek_config.yaml predates these keys
+    # clusters identically after the upgrade.
+
+    eddy_eps: float = Field(default=0.3, gt=0.0, lt=1.0)
+    """Maximum cosine *distance* for a DBSCAN neighbour in eddy detection.
+
+    ``0.3`` means an edge exists between two fragments at cosine similarity
+    ``>= 0.70``. Lowering it tightens every cluster; it cannot, on its own,
+    separate a continuous message stream (see ``stream_platforms``).
+    """
+
+    eddy_min_samples: int = Field(default=5, ge=1)
+    """Minimum neighbourhood size for a DBSCAN core point in eddy detection."""
+
+    eddy_correlation_threshold: float = Field(default=0.3, ge=0.0, le=1.0)
+    """Absolute Spearman ceiling above which a cluster is thread-like.
+
+    A candidate eddy whose content drift correlates with chronological rank
+    at or above this value is a directional progression, and is filtered out
+    of the eddy set rather than emitted as a topic cluster.
+    """
+
+    thread_window_days: int = Field(default=30, ge=1)
+    """Sliding-window width, in days, for thread detection.
+
+    Distinct from ``temporal_window_hours``, which feeds only the temporal
+    *proximity* linker and has never influenced threads.
+    """
+
+    thread_similarity_threshold: float = Field(default=0.6, ge=0.0, le=1.0)
+    """Cosine similarity a pair must strictly exceed to join one thread."""
+
+    thread_union_without_embeddings: bool = False
+    """Whether a pair missing an embedding may union on frequency alone.
+
+    When the detector has *no* embeddings at all, thread detection falls
+    back to frequency agreement by design. On a *partially* embedded vault
+    the same fallback fires per pair, silently chaining fragments the
+    similarity gate would have rejected — so it is closed by default and
+    must be opted into.
+    """
+
+    # ---- Cluster-size guardrail (issue #880) ----
+
+    cluster_size_ceiling: int = Field(default=500, ge=1)
+    """Absolute member count below which a cluster is never split.
+
+    The effective ceiling is ``max(cluster_size_ceiling, floor(corpus_size *
+    cluster_max_fraction))``: the absolute floor dominates ordinary vaults,
+    so the guardrail stays inert until a corpus is large enough for
+    degeneration to matter.
+    """
+
+    cluster_max_fraction: float = Field(default=0.10, gt=0.0, le=1.0)
+    """Largest share of the corpus a single eddy or thread may hold.
+
+    ``1.0`` is the documented opt-out: one cluster may then span everything.
+    """
+
+    cluster_split_max_depth: int = Field(default=3, ge=0)
+    """Re-clustering rounds allowed per oversized cluster.
+
+    ``0`` disables splitting entirely, sending any oversized cluster
+    straight to noise.
+    """
+
+    eddy_split_eps_step: float = Field(default=0.05, gt=0.0, lt=1.0)
+    """Amount ``eddy_eps`` is tightened by on each re-clustering round."""
+
+    thread_split_similarity_step: float = Field(default=0.1, gt=0.0, lt=1.0)
+    """Amount ``thread_similarity_threshold`` is raised on each round."""
+
+    # ---- Message-stream segmentation (issue #880) ----
+
+    stream_platforms: list[str] = Field(
+        default_factory=lambda: ["discord", "email"],
+    )
+    """Source platforms whose fragments are cut into conversation episodes.
+
+    A continuous chat stream violates the precondition both detectors rely
+    on — clusters separated by low-density regions — so it is partitioned
+    into independent clustering domains *before* any similarity graph is
+    built. The default names the two platforms Creek already routes to the
+    ``01-Fragments/Messages/`` subfolder. An empty list disables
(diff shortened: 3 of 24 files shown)


Geoffe-Ga:
## Summary
This PR fixes issue #880 (eddy/thread "mega-cluster" degeneration on Discord message-stream fragments) with a well-reasoned three-part fix: conversation-episode segmentation (`segments.py`) applied before DBSCAN, a bounded union-find scope plus a closed fail-open path in `threads.py`, a size-ceiling guardrail with recursive re-clustering (`cluster_limits.py`), and a real TF-IDF naming/description module (`naming.py`) replacing a fake word-count "TF-IDF-lite." Two independent deep-reads of the actual (not just diffed) source confirm the PR description's specific, falsifiable technical claims (union-find fail-open closure, content-stable `Eddy.id`, real IDF math, episode-boundary correctness, never-empty description) all hold up.

## Strengths
- Root-cause analysis is unusually rigorous — the PR correctly diagnoses that DBSCAN's *precondition* (density-separated clusters) is violated by chat streams, so no `eps` tuning could fix it, and addresses that at the segmentation layer rather than papering over it with the size cap alone.
- The size-ceiling guardrail (`cluster_limits.py`) is explicitly documented and treated as a guardrail, not the fix — verified: BFS-based (not recursive-call-stack), depth- and range-bounded, no stack-depth or unbounded-work risk found.
- `naming.py`'s TF-IDF is genuinely document-frequency-based (not count-based), and every traced code path in title/description generation terminates in a non-empty fallback — the "never-empty description" claim is real.
- `_UnionFind`'s fail-open path (union on frequency alone when an embedding is missing) is now scoped to fully-unembedded corpora only, closing a real silent-chaining bug; `Eddy.id` is now content-stable and provably consistent with the pre-existing `_stable_thread_id` scheme.
- New test corpus (`test_link_message_stream_880.py`) uses a seeded random walk on the unit sphere with adjacent-step cosine ≈0.95 (inside `eps=0.3`), a genuine reproduction of the transitive-closure collapse — not a vacuous fixture like the old one-hot generator it's contrasted against.
- Config additions (`LinkingConfig`) are thoroughly docstringed, and defaults are stated (and appear) to match the pre-existing module-private constants they replace, so upgrading a vault's config doesn't silently reclassify it.

## Security Concerns
🟢 LOW — no new I/O, subprocess, network, or path-handling surface introduced. Nothing flagged.

## Problems
None found that block merge. No crash-level bug, incorrect-by-construction algorithm, or CLAUDE.md-mandated gate violation surfaced across two independent deep-dives into the five touched/new linking modules.

## Code Quality
- `naming.py`'s `_WORD_RE = re.compile(r"[a-z]+")` is ASCII-only, but Discord channel names and message content routinely contain emoji/non-Latin scripts — exactly the domain this fix targets. There's no test exercising a unicode/emoji-heavy cluster, so it's unverified whether title/description quality degrades gracefully (e.g., falls through cleanly to the structural fallback) versus silently producing thin/unhelpful output on non-ASCII-heavy channels.
- `test_recluster_work_is_bounded_by_split_depth` is flagged by the PR itself as the "weakest test" (fails on main via `TypeError` at construction, not a real red assertion) — true, but the same is equally true of three other acceptance tests that also construct with the new `SplitPolicy` kwarg and narrate detailed "on unmodified main..." figures they cannot have literally reproduced. This is a documentation-accuracy nit in the PR body, not a code defect, but worth a one-line correction so the "weakest test" framing doesn't read as an implicit (overstated) endorsement of the others' rigor.
- Same test's fixture only exercises a single-branch split path; it doesn't prove the depth×N work bound survives fan-out (multiple still-oversized subclusters produced in one round). Low-cost to add and would meaningfully strengthen the one test explicitly called out as weak.
- `thread_union_without_embeddings` defaulting to `False` closes the fail-open bug, but its effect isn't scoped to message-stream sources — it changes default cross-source thread-matching behavior for *any* partially-embedded vault. This is very likely a correct and desired fix (closing a real bug), but worth the author explicitly confirming it's intentional/in-scope rather than an incidental side effect, since the PR frames itself as message-stream-specific.

## CLAUDE.md Adherence
✅ TDD: acceptance-criteria table maps directly to concrete tests; new corpus is a real regression reproduction, not vacuous.
✅ Complexity: no function in the new/changed modules approaches the cyclomatic-complexity-10 ceiling per reviewer trace.
✅ Types: new public functions fully typed; `Protocol`-based structural types used instead of `Any`.
✅ Docstrings: new config fields and modules are thoroughly docstringed.
✅ Coverage claims: PR states 94.16% aggregate, 100%/97.6%/96.5% on new modules — consistent with the breadth of new test files reviewed (`test_cluster_limits.py`, `test_link_naming.py`, `test_link_segments.py`, `test_link_message_stream_880.py`).
⚠️ CI status could not be independently verified in this review environment (`gh pr checks` was blocked by sandbox policy); taking the PR's stated "12/12, 8287 passed" at face value pending CI confirmation before merge.
✅ Commit message follows Conventional Commits (`fix(link): ...`) with detailed rationale.

## Testing
Strong. Both a focused algorithm-correctness review and a focused test-suite review independently verified:
- The new message-stream corpus is a genuine, seeded, non-vacuous bug reproduction (not the old one-hot fixture it explicitly improves on).
- `test_link.py`'s 151 added lines are additive only (a new `TestIssue880Detectors` class) — no evidence of weakened or edited pre-existing assertions among the 108 prior tests.
- Non-empty-description tests cover empty cluster, single fragment, all-stopword content, and wiki-hostile characters — good adversarial breadth, missing only unicode/emoji (noted above).
- No lazy `assert x is not None` patterns found in the reviewed new test files; boundary pairs (gap==max vs. max+1) are exercised with exact-value/`pytest.approx` assertions.
- Gap noted above on `test_recluster_work_is_bounded_by_split_depth` not exercising fan-out — not blocking, but the one place a reviewer should ask for a follow-up before treating that guarantee as fully proven.

## Documentation
Excellent. ADR-0008 gives a clear, mechanism-by-mechanism account of all three compounding bugs and why partial fixes (cap-only, eps-tuning) were rejected, matching the actual code. `docs/configuration.md` and `docs/linking.md` are updated for the new config surface. The PR explicitly and honestly flags an out-of-scope pre-existing gap (`creek_mcp/tools/link.py`'s `_VALID_METHODS` omitting `"threads"`) — verified true and correctly scoped as pre-existing/unrelated rather than silently left for a future reviewer to discover.



## Verdict: COMMENTS

No blocking bugs, security issues, or CLAUDE.md violations surfaced across two independent deep-dives (core algorithms verified correct; test suite's falsifiable claims verified true or only mildly overstated in prose). It stops short of LGTM because there are genuine, worth-weighing suggestions before merge: the ASCII-only naming regex is untested against the unicode/emoji content this exact bug's domain (Discord) routinely contains, the self-flagged "weakest test" doesn't exercise fan-out and its weakness framing is inconsistently applied to sibling tests, and the `thread_union_without_embeddings` default change has a broader blast radius than the PR's message-stream framing suggests and deserves an explicit intentionality check.

Geoffe-Ga:
**CI**: 4/7 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227744606 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 10/10 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227744606 to see in-depth feedback and continue iterating

Issue #1025, opened 2026-07-30
test(e2e): replace the fictional vault fixture with the real `creek init` scaffold + drift guard


Geoffe-Ga:
Parent epic: #1024

## Role
You are a senior test-infrastructure engineer working in `creek-tools/`,
following the repo's TDD workflow and `./scripts/check-all.sh` gates.

## Goal
Make the e2e suite run against the **real** vault scaffold, and add a drift
guard so the scaffold and the code that writes into it can never diverge
silently again.

This is the skeleton issue for the epic: every later assertion depends on the
vault under test being real.

## Context

`tests/e2e/conftest.py` hand-rolls a `_VAULT_DIRS` tuple. It has drifted
completely from `creek/templates/vault/`, which is what `creek init` actually
materialises:

| Fixture creates | Reality |
|-----------------|---------|
| `05-Voice/{Profiles,Exemplars,Patterns}` | scaffold + code use `07-Voice` |
| `06-Wavelength/Reports` | scaffold + code use `05-Wavelength` |
| `07-Review` | **referenced nowhere in `creek/` or `creek_mcp/`** |
| `01-Fragments/{Notes,Documents,Data,Decks,Images}` | **absent from the scaffold** |
| `00-Creek-Meta/{Indexes,Audit,Consent}` | **absent from the scaffold** |

It omits everything real: `06-Frequencies/`, `09-Reference/`, `10-Liminal/`,
`11-Other-Authors/`, `05-Wavelength/{Mode-Profiles,Phase-Maps,Observations}`,
`07-Voice/{Drafts,Lexicon,Rhetorical-Patterns,Register-Samples}`.

Nothing has ever failed because every writer calls
`mkdir(parents=True, exist_ok=True)` — `creek/vault/writer.py:384`, `841`,
`876`, `1054`, `1219`, `1237`. The suite silently creates whatever it needs and
declares success.

The fixture's own comment admits the hazard:

```python
# 01-Fragments subfolders mirror creek/vault/writer.py's
# _PLATFORM_SUBFOLDER mapping. Keep this in sync when adding a new
# SourcePlatform value.
```

"Keep this in sync" by hand is the bug. It was not kept in sync.

### Two live production bugs this surfaces
1. `_PLATFORM_SUBFOLDER` (`creek/vault/writer.py:66-81`) routes `MARKDOWN →
   Notes`, `DOCUMENT → Documents`, `SPREADSHEET → Data`, `PRESENTATION →
[...]

Issue #1024, opened 2026-07-30
epic: Integration testing that catches features wired to nothing


Geoffe-Ga:
## Role
You are a senior test-infrastructure engineer working in the Creek-Vault
codebase, following its existing conventions (TDD, `creek-tools/scripts/check-all.sh`
gates, ≥90% branch coverage with per-file floors, pinned tooling, the
`_claude-scan.yml` → `scan-issue-writer` → Ralph autonomous-maintenance
architecture).

## Goal
Make CI prove, on every PR, that every user-reachable surface is **connected to
the implementation that claims to serve it** — against the real vault scaffold,
across the real process boundaries.

This is a tracking epic. Implementation lives in the sub-issues.

## Child Issues
- [ ] #1025 — Replace the fictional e2e vault fixture with the real `creek init` scaffold + drift guard **(start here)**
- [ ] #1026 — Make the gate real: `live` marker split, blocking `integration` + `e2e` CI job
- [ ] #1027 — Wiring contract test: every CLI command and MCP tool reachable and effectful
- [ ] #1028 — CrawDad ↔ `creek_mcp` against the real server, retiring the 38-line fake
- [ ] #1029 — Deepen the e2e journeys: behavioural assertions over `does_not_raise`
- [ ] #1030 — `scan-wiring` nightly lane feeding half-wired surfaces into Ralph

**Sequencing.** #1025 is the skeleton — every later assertion is built on it, so
it lands first. #1026 makes the gate real and lands immediately after. #1027,
#1028 and #1029 touch disjoint paths (`tests/`, `crawdad/tests/`, `tests/e2e/`)
and can run in parallel once the first two land. #1030 consumes the others and
lands last.

### The problem, stated as evidence
Features reach `main` wired to nothing, and the defect is found later by hand
during an ad-hoc e2e pass. `repo:Geoffe-Ga/Creek-Vault is:issue wire OR wired
OR unwired OR orphaned in:title` returns **17 issues**:

- **#935** — `PrivacyClassifier` was not wired into `classify`, so **privacy
  tiers were not real**. The core safety property of a privacy-tiered product
  was decorative, and 6644 passing tests did not notice.
- **#649 / #658** — the author conductor stored an `llm_client` but built a
  bare `VoiceAgent()` stub; neither the CLI nor MCP built a real client. The
  fix's own docstring (`tests/test_author_desk_wiring.py:1-9`) is the clearest
  statement of this failure mode in the repo.
[...]

PR #1234, opened 2026-08-08
ci: make integration + e2e blocking, and free the `integration` marker from live-API tests

### LINKED ISSUE #1026: ci: make integration + e2e blocking, and split the live-API marker out of `integration`
Parent epic: #1024

## Role
You are a senior CI/release engineer working in the Creek-Vault repo, following
its existing workflow conventions (`.github/workflows/ci.yml`, the
`working-directory: creek-tools` job pattern, pinned tooling, `uv` locked
installs).

## Goal
Make integration and e2e tests **block merges**. Today they gate nothing.

## Context

`.github/workflows/ci.yml:162` is the only test step:

```yaml
- name: Run tests with coverage (CI-003 — mirrors local default)
  # ./scripts/test.sh excludes integration/e2e by default so local
  # ...
  run: ./scripts/test.sh --unit --coverage
```

`scripts/test.sh` maps `--unit` to `-m "not integration and not e2e and not
slow"`. `grep -rn "e2e" .github/workflows/` finds only prose in `deslop.yml`.

**Net effect: 12 e2e tests exist and none of them gate a merge.** They run only
when a human types `./scripts/test.sh --e2e`. That is why unwired features are
found late, by hand, in an ad-hoc e2e pass.

### The marker collision that has to be resolved first
`integration` currently means *live API smoke*, not *cross-component*:

- `pyproject.toml` declares `integration`, `e2e`, `slow`.
- `scripts/test.sh --integration` runs `-m "integration" --no-cov` and the help
  text calls them "live smokes; coverage gate off".
- `tests/test_live_smoke.py` and `tests/test_real_agents.py` need real
  provider keys.

A gate that needs a paid API key is a flaky gate, so the name must be freed:

- Rename the live-API marker to **`live`** (`tests/test_live_smoke.py`,
  `tests/test_real_agents.py`, and any other live-key test).
- **`integration`** is redefined as *hermetic, cross-component, no network* —
  the lane sub-issues 3 and 4 populate.
- `e2e` keeps its meaning: full journeys through the CLI against a real vault.

### Budget
The 12 current e2e tests run in **23s** (measured). The blocking lane's budget
is **≤5 min**; anything slower or flaky belongs in the nightly lane from
sub-issue 6 (`scan-wiring`).

## Output Format
A PR containing:
- Marker rename `integration` → `live` across the live-key tests, with
  `pyproject.toml` markers updated and `--strict-markers` still satisfied.
- `scripts/test.sh` gaining `--live`, with `--integration` re-pointed at the
  hermetic lane and `--all` still meaning everything.
- A new blocking CI job (or a new step in the existing quality job — pick one
  and justify it in the PR body) running `--integration` and `--e2e`.
- The Quality Gate job (`ci.yml:344`) updated so the new job's failure blocks
  the gate.
- `creek-tools/CLAUDE.md` §8 (Testing Strategy) updated to document the three
  markers and which lane blocks.

## Examples

```bash
./scripts/test.sh --unit          # -m "not integration and not e2e and not slow and not live"
./scripts/test.sh --integration   # -m "integration"  — hermetic, BLOCKING
./scripts/test.sh --e2e           # -m "e2e"          — hermetic, BLOCKING
./scripts/test.sh --live          # -m "live"         — needs keys, NOT in CI
```

```yaml
- name: Run integration + e2e tests (blocking)
  working-directory: creek-tools
  run: |
    ./scripts/test.sh --integration
    ./scripts/test.sh --e2e
```

## Constraints
- **The new lane must be hermetic**: no network, no API keys, no real vault.
  If a test needs a key, it is `live` and does not belong here.
- Do not run the new lane under the 90% `--cov-fail-under` gate — these lanes
  use `--no-cov` today for good reason, and coverage is the unit lane's job.
  Say so explicitly in the PR rather than silently dropping the flag.
- Do not weaken the existing unit job to make room for the new one.
- Keep the Python version matrix behaviour consistent with the existing job, or
  justify running the new lane on a single version to save minutes.
- Sequencing: land **after** the vault-fixture issue. Turning on a blocking gate
  that runs against the fictional vault fixture would cement the fiction.
- No `noqa`, `# type: ignore`, `pylint: disable`, or equivalent. Fix the root
  cause. The only exception is the documented 4-line escape hatch (third-party
  bug / version compat / benchmarked perf / generated code) with a review date.

## Done-Done
- [ ] CI shows the new job running and **failing** when an e2e test is
      deliberately broken (paste the run link in the PR — proof of teeth).
- [ ] `./scripts/test.sh --all` passes locally.
- [ ] `./scripts/check-all.sh` exits 0.
- [ ] `shellcheck` clean on the modified `scripts/test.sh`.
- [ ] PR opened with `Refs #1024` and `Closes #<this-issue>`.
- [ ] Latest `Verdict:` on HEAD is `LGTM`.

Geoffe-Ga:
Closes #1026. Epic #1024.

**12 e2e tests existed and none of them gated a merge.** `ci.yml`'s only test step ran `./scripts/test.sh --unit` → `-m "not integration and not e2e and not slow"`, and `grep -rn "e2e" .github/workflows/` found only prose in `deslop.yml`. They ran when a human typed the flag. That is why features wired to nothing get found late, by hand.

## The marker collision, resolved from the code rather than the issue text

`integration` meant *live API smoke*, not *cross-component* — `--integration` ran `--no-cov` and the help text called them "live smokes". A gate that needs a paid API key is a flaky gate, so the name had to be freed.

The issue's list of what to move was **wrong in both directions**, so every `integration`-marked test was swept individually:

| Test | Issue said | Reality | Action |
|---|---|---|---|
| `test_live_smoke.py` (4 tests) | needs keys | correct — real vendor calls | → `live` |
| `test_real_agents.py` | "needs real provider keys" | **false.** No marker at all; uses the deterministic `mock_sentence_transformer` fixture; already a unit test. Only match was docstring prose at `:376` | **left alone** |
| `test_classify.py::test_classify_with_real_ollama` | not mentioned | `LLMClassifier.available` probes `localhost:11434` over HTTP — same flakiness class as a key, and CI runs no Ollama | → `live` |

13 hermetic integration tests across 6 files stay in the blocking lane.

## Why a separate job, not a step in `quality`

1. **Matrix duplication** — `quality` is a 3-version matrix (3.11/3.12/3.13). A step there runs the lane three times for one signal. These tests exercise wiring between components, not version-sensitive syntax; a version-specific break surfaces in the unit matrix first, on the same commit.
2. **Failure signal** — a red check named "Integration & E2E Tests" is unambiguous. A failing step inside "Code Quality & Testing (3.12)" is indistinguishable from a lint, mypy, bandit, or coverage failure in the same job.
3. **Coupling** — `quality`'s coverage steps read the `.coverage` file the unit run writes. A `--no-cov` lane sitting among them writes no data and adds needless order-dependence.

It carries **no `needs:`**, so it runs in parallel with `quality`: a tree failing lint should still report whether its cross-component wiring holds, rather than hiding that for a whole extra round trip. Within the job the e2e step carries `if: ${{ !cancelled() }}` so a failing integration step doesn't mask the e2e result — both land in one run. A failed step still fails the job.

## The Quality Gate needed *both* halves

This is the part that would silently not work. The gate is `if: always()`, which defeats GitHub's skip-on-upstream-failure, so `needs:` only makes `needs.<id>.result` *readable* — **all gating is done in shell**. A job added to `needs:` without a matching `!= "success"` check is a no-op that leaves this issue's entire premise unfixed while looking done.

```yaml
needs: [quality, integration-e2e, crawdad, complexity-analysis, build]
```
```bash
if [ "${{ needs.integration-e2e.result }}" != "success" ]; then
  echo "Integration/e2e tests failed"
  exit 1
fi
```

Verified mechanically by parsing the YAML and diffing the two sets:

```
needs    : ['quality', 'integration-e2e', 'crawdad', 'complexity-analysis', 'build']
checked  : ['build', 'complexity-analysis', 'crawdad', 'integration-e2e', 'quality']
unchecked: NONE
```

A comment above `if: always()` now records the rule for the next person.

## Hermeticity: proven in two layers, not asserted

**Layer 1 — keys unset** (confirmed with `env | grep -cE "ANTHROPIC|OPENAI|GOOGLE_API|GEMINI|CREEK_CLOUD|CREEK_ANTHROPIC"` → `0`):
```
--integration → 13 passed, 8162 deselected  (exit 0)
--e2e         → 12 passed, 8163 deselected  (exit 0)
```
The integration lane's skip count dropped **5 → 0** — a clean signal that every credential-dependent test left.

**Layer 2 — keys unset AND a socket guard**, because "no keys" does not prove "no network". A throwaway plugin (kept outside the repo) raises on any non-loopback `connect` / `connect_ex` / `getaddrinfo`: 13 and 12 passed respectively.

The guard was proven to bite, since a proof that cannot fail proves nothing:
```
connect(('api.anthropic.com', 443)) → guard bites: outbound connection attempted
getaddrinfo('api.openai.com', 443)  → guard bites: DNS resolution attempted
```

**Gate teeth proven** by appending a deliberately failing e2e test: `--e2e` exits 1, the job goes red, the gate's `!= "success"` check exits 1. File restored; `git diff --stat` on it is empty.

## Budget and coverage

Blocking lane: **12s**, against the issue's ≤5 min budget. Deliberately **not** under `--cov-fail-under` — these lanes use `--no-cov` for good reason and coverage is the unit lane's job, as the issue asked be stated explicitly rather than silently dropped.

## Deviation: CLAUDE.md §8 was not created

The issue asks for "`creek-tools/CLAUDE.md` §8 (Testing Strategy)". **§8 does not exist**, and §7 records that those sections were retired under #1190 specifically to stop docs drifting from their source — it states *"this section only points, it does not describe."*

Writing a prose Testing Strategy section would reverse a deliberate decision and recreate exactly the drift class #1025 just spent a PR eliminating. The three markers and which lane blocks are documented at the source §7 points to (`scripts/test.sh` help text and the pyproject markers table), with §7's pointer updated to name the new lane.

## Verification

`./scripts/check-all.sh` — **12/12 passed**. `--unit`, `--integration`, `--e2e`, `--live` and `--all` each verified to do what their help text now claims. `--live` with keys unset: 5 skipped, exactly the moved tests. shellcheck clean. Zero suppressions.

### DIFF
--- a/.claude/skills/model-setup/SKILL.md
+++ b/.claude/skills/model-setup/SKILL.md
@@ -135,8 +135,8 @@ system package without asking.
 2. Optional live smoke (only with the key + consent set), per the README — run
    it from the `creek-tools/` subproject directory (the `scripts/` live there,
    not at the repo root):
-   `cd <repo>/creek-tools && ./scripts/test.sh --integration -k anthropic` (or
-   `CREEK_SMOKE_MODEL=<id> ./scripts/test.sh --integration -k <provider>`).
+   `cd <repo>/creek-tools && ./scripts/test.sh --live -k anthropic` (or
+   `CREEK_SMOKE_MODEL=<id> ./scripts/test.sh --live -k <provider>`).
 3. Print a **per-stage summary table**: stage → provider → model → ready? Then
    tell the user the Intimate guarantee below is already in force.
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,6 +229,74 @@ jobs:
             creek-tools/reports/pip-audit-report.json
           retention-days: 30
 
+  integration-e2e:
+    name: Integration & E2E Tests
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: creek-tools
+    # Deliberately NO `needs:` — this lane is independent of `quality` and runs
+    # in parallel with it. A tree that fails lint should still report whether
+    # its cross-component wiring holds; serialising behind `quality` would hide
+    # that until the lint fix lands, costing a whole extra round trip. The lane
+    # is seconds of tests, so the parallel runner is nearly all setup cost.
+    #
+    # Single Python version, unlike `quality`'s 3.11/3.12/3.13 matrix. These
+    # tests exercise wiring between components, not version-sensitive syntax or
+    # stdlib behaviour; a version-specific break surfaces in the unit matrix
+    # first, and on the same commit. Fanning this lane across three versions
+    # would triple the runner minutes for no additional signal.
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.PYTHON_DEFAULT_VERSION }}
+
+      - name: Cache uv
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/uv
+          key: ${{ runner.os }}-uv-${{ env.PYTHON_DEFAULT_VERSION }}-${{ hashFiles('creek-tools/uv.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-uv-${{ env.PYTHON_DEFAULT_VERSION }}-
+            ${{ runner.os }}-uv-
+
+      - name: Install uv
+        # Same pin as the `quality` job — see the rationale there.
+        run: python -m pip install 'uv==0.11.16'
+
+      - name: Install dependencies (locked)
+        run: |
+          uv export --locked --all-extras --no-emit-project --no-hashes \
+            -o /tmp/locked-requirements.txt
+          uv pip install --system -r /tmp/locked-requirements.txt
+          uv pip install --system --no-deps -e .
+
+      - name: Run integration tests (hermetic, blocking)
+        # No provider credentials are set for this step, and none may be added:
+        # both lanes are hermetic by definition, so a test that needs a key is
+        # `live` and belongs outside CI. Withholding the keys is what keeps that
+        # honest — a leaked key here would let a live test pass silently.
+        #
+        # ./scripts/test.sh passes --no-cov for this lane. That is deliberate,
+        # not an oversight: a marker-only selection touches a fraction of the
+        # package, so pyproject's --cov-fail-under=90 would fail every run.
+        # Coverage is the unit lane's job and is enforced there.
+        run: ./scripts/test.sh --integration
+
+      - name: Run e2e tests (hermetic, blocking)
+        # `!cancelled()` so a failing integration step above does not mask this
+        # lane: both results land in one run instead of costing a round trip to
+        # discover the second failure. This does NOT weaken the gate — a failed
+        # step still fails the job, and `quality-gate` checks the job result.
+        if: ${{ !cancelled() }}
+        # --no-cov here too, for the same reason as the integration step above.
+        run: ./scripts/test.sh --e2e
+
   crawdad:
     name: CrawDad Quality
     runs-on: ubuntu-latest
@@ -359,7 +427,13 @@ jobs:
   quality-gate:
     name: Quality Gate
     runs-on: ubuntu-latest
-    needs: [quality, crawdad, complexity-analysis, build]
+    needs: [quality, integration-e2e, crawdad, complexity-analysis, build]
+    # `if: always()` defeats GitHub's default skip-on-upstream-failure, so
+    # `needs:` alone gates NOTHING here — it only makes `needs.<job>.result`
+    # readable. Every job listed above MUST also have an explicit
+    # `!= "success"` check below. A job added to `needs:` but not to the script
+    # is the classic silent no-op: the gate waits for it, then passes whatever
+    # it reported.
     if: always()
 
     steps:
@@ -367,6 +441,10 @@ jobs:
         run: |
           if [ "${{ needs.quality.result }}" != "success" ]; then
             echo "Quality checks failed"
+            exit 1
+          fi
+          if [ "${{ needs.integration-e2e.result }}" != "success" ]; then
+            echo "Integration/e2e tests failed"
             exit 1
           fi
           if [ "${{ needs.crawdad.result }}" != "success" ]; then
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,10 @@ install -r requirements-dev.txt` still works as an unpinned fallback.
 ./scripts/check-all.sh          # Run ALL quality checks (do this before every commit)
 ./scripts/fix-all.sh            # Auto-fix linting + formatting
 ./scripts/test.sh               # Run unit tests
-./scripts/test.sh --all         # Run all test types (unit, integration, e2e)
+./scripts/test.sh --all         # Run all lanes (unit, integration, e2e, live)
+./scripts/test.sh --integration # Hermetic cross-component lane (blocking in CI)
+./scripts/test.sh --e2e         # Hermetic end-to-end lane (blocking in CI)
+./scripts/test.sh --live        # Live API/service smokes (needs keys; not in CI)
 ./scripts/test.sh --coverage    # Unit tests with coverage report
 ./scripts/coverage.sh           # Coverage report (--html for HTML output)
 ./scripts/lint.sh               # Ruff linting (--fix to auto-fix)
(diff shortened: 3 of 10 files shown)


Geoffe-Ga:
## Summary
Makes the `integration` and `e2e` pytest lanes actually block CI: adds a parallel `integration-e2e` job, wires it into `quality-gate`'s shell-based (not `needs:`-based) gating, and frees the `integration` marker from double-duty as "live API smoke" by introducing a new `live` marker for real-credential tests. Includes a careful marker sweep (moving `test_classify_with_real_ollama` to `live`, correctly leaving `test_real_agents.py` alone), and updates docs/scripts/pyproject accordingly.

## Strengths
- The PR body is unusually rigorous: it proves hermeticity two ways (env-var absence *and* a socket-guard plugin), proves the gate has teeth by injecting and reverting a deliberately-failing test, and mechanically diffs `needs:` vs. the shell-checked job list rather than eyeballing it.
- Correctly identifies that `if: always()` defeats GitHub's skip-on-upstream-failure, so `needs:` alone is a no-op without the matching `!= "success"` shell check — and adds a comment recording that rule for future maintainers (`.github/workflows/ci.yml`).
- `!cancelled()` on the e2e step is the right call so a failing `integration` step doesn't mask an e2e failure in the same run, while still failing the job overall.
- Reasonable job-design rationale for a separate `integration-e2e` job vs. a step inside the 3-version `quality` matrix (avoids tripling runner cost for a version-insensitive lane, avoids sharing `.coverage` state with a `--no-cov` lane, gives an unambiguous failure signal).
- The marker sweep was done from the code (checking what each test actually does) rather than trusting the linked issue's list, and it caught a real discrepancy (`test_classify_with_real_ollama` probes `localhost:11434` and needs `live`, not `integration`).
- Reasoned, explicit deviation from the linked issue (declining to add a CLAUDE.md §8) with a documented rationale tied to a prior decision (#1190).

## Security Concerns
None. 🟢 The new job intentionally withholds all provider credentials, and the PR demonstrates (via a socket-guard) that no outbound network calls occur in the blocking lanes.

## Problems
None blocking.

## Code Quality
- **Stale doc left in the very file this PR edits three times** (`creek-tools/CLAUDE.md`): §6.1 "Test Coverage" still says *"Test markers: `unit` (default), `integration`, `e2e`. Local `./scripts/test.sh` and CI both default to `not integration and not e2e` (CI-003)"* (lines 385–387). That's now materially incomplete/misleading — it omits the new `live` marker entirely and doesn't mention that CI now runs `integration`/`e2e` as a *blocking* lane, which is the entire point of this PR. Similarly, §5.2's architecture diagram comment at line 322 still reads `test.sh # Run test suite (--unit / --integration / --e2e / --all)`, omitting `--live`. Given the PR's stated goal is eliminating exactly this class of doc drift (and the "Testing strategy" pointer a few dozen lines away *was* correctly updated), these two lines should be fixed too.
- `pyproject.toml`'s new `minversion = "9.0"` is paired with a comment noting "uv.lock pins 9.0.3, so this only bites an unpinned fallback install" — but the `[dev]` extra still declares `"pytest>=7.4.0"` (unchanged by this PR). Per `CLAUDE.md`, `pip install -r requirements-dev.txt` is a documented supported fallback path; it resolves that same `>=7.4.0` specifier. In practice pip will grab the latest matching wheel today, so this likely doesn't bite immediately, but the specifier itself is now inconsistent with what the ini actually requires and should probably read `>=9.0` to match.
- Minor/unverifiable: the rationale for adding raw `strict_markers = true` / `strict_config = true` as ini keys (on top of the existing `--strict-markers`/`--strict-config` in `addopts`) rests on a specific claim about pytest 9 not honoring those flags when set via `addopts`. I couldn't independently verify this against pytest's ini-option registry in this sandbox (no network/pytest execution access). If the claim is wrong these would be unrecognized ini keys, but the PR's own reported `12/12` verification and per-lane pass counts would very likely have surfaced that immediately, so I'm not raising it as a defect — just flagging it as worth double-checking once CI runs on pytest 9.0.3.

## CLAUDE.md Adherence
✅ Follows the four-gate workflow described in the PR body (TDD via the injected-then-reverted failing test, `check-all.sh` reported 12/12, CI wiring, self-review). ✅ Commit message and PR title follow Conventional Commits style. ✅ Uses feature branch (`issue/1026-blocking-integration-e2e`), not a direct push to main. ❌ Two now-stale references remain in `creek-tools/CLAUDE.md` (see Code Quality) despite this PR editing that file elsewhere — a minor adherence gap given CLAUDE.md's own emphasis on pointer-not-drift documentation.

## Testing
This PR *is* a testing-infrastructure change, and its own test-of-the-tests is thorough: two independent hermeticity proofs (env-var absence, socket-guard), and a proof the gate can actually fail (inject a failing e2e test, confirm exit 1 propagates through `quality-gate`, then revert with an empty `git diff --stat`). The marker reclassification is verified against actual test behavior rather than assumed from the issue text. No gaps here that would warrant blocking.

## Documentation
`CLAUDE.md` (root), `README.md`, and the authoritative pointer in `creek-tools/CLAUDE.md` §7-equivalent are all updated consistently with the new `unit`/`integration`/`e2e`/`live`/`slow` model and correctly describe which lanes block. The gap is confined to the two stale lines noted above in Code Quality — not a completeness failure of the PR's intent, just a couple of spots the sweep missed.



## Verdict: COMMENTS

Nothing here blocks merge — the CI wiring, marker sweep, and gate-teeth verification are all sound and well-evidenced — but the PR leaves two now-inaccurate statements in creek-tools/CLAUDE.md's Test Coverage/Architecture sections (stale "integration/e2e excluded from CI" description and a stale `test.sh` usage comment) that the author should fix given the PR's own thesis is eliminating exactly this kind of doc drift.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227761234 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5227761234 to see in-depth feedback and continue iterating

PR #1235, opened 2026-08-08
fix(ralph): make a do-not-auto-merge hold survive the verdict line, and stop re-assign orphaning a PR

### LINKED ISSUE #1180: fleet.sh assign silently branches from main when release has deleted the local ref, orphaning a PR's commits

### LINKED ISSUE #1202: iteration-trigger 'NOT cleared' summaries still say VERDICT: LGTM — await-claude-review Step 4a merges on them, defeating do-not-auto-merge

Geoffe-Ga:
Closes #1202. Closes #1180.

Two fleet-loop **recovery-path** defects — both correct in the common path, wrong when something goes sideways. Shell and YAML only; no Python touched.

## #1202 — a human's hold was inoperative

In `iteration-trigger.yml`'s clearance chain, **only** the #1181 provenance branch rewrote `VERDICT`. The HELD branch and the BEHIND branch rewrote `ACTION` only.

So a PR a human had explicitly held still posted:

```
**CI**: 10/10 Green
**VERDICT**: LGTM
```

…which is precisely and only the two fields `await-claude-review` Step 4a parses before merging. **A loop draining a queue of PRs would have merged straight through the only veto a human has.** The file's own comment admitted the asymmetry and named this issue.

Every NOT-cleared branch now rewrites `VERDICT`: `HELD`, `NOT CURRENT`, and the existing `NOT ATTESTED`. The values were chosen for their **bytes** — none contains an `LGTM` substring, so no `*LGTM*` glob or `test("LGTM")` can read one as permission.

### Both consumers refuse, for independent reasons

**`SKILL.md` Step 4a is now a whitelist**: merge only on an exactly-`LGTM` verdict; anything unrecognised routes to "surface to the user and stop". It is deliberately **not** a blacklist of known-bad verdicts — the emitter's refusal vocabulary grows, and a blacklist is exactly how this hole survived #1181.

**`pr-ready.sh` is deliberately unchanged**, and that's the honest answer rather than a gap. It cannot merge on this summary by three independent pre-existing routes: `ITER_SUMMARY_RE` excludes the iteration-trigger comment from the verdict selector entirely; it enforces the hold natively via `has_optout_label`, dying exit 2 when the label lookup is unreadable rather than assuming no hold; and it enforces currency natively via `branch_is_current`. Adding a guard for the new strings would be dead code asserting a route that doesn't exist. The claim is **pinned with a test** instead (case W7), which also kills the tempting mutant of accepting `<!-- iteration-trigger -->` as a second provenance marker.

### The test is structural, not a list of literals

This is the part that keeps the fix from rotting. An awk walk of the emitter's clearance chain extracts, for each `ACTION="NOT cleared to merge…"`, the `VERDICT` in force on that branch — reset at every `if`/`elif`/`else` so a sibling's value is never credited — then asserts per branch that it is set at all, contains no `LGTM`, and is not one of `SKILL.md`'s recognised verdicts. A second loop asserts `SKILL.md` names each emitted value verbatim.

**A fourth refusal branch added later inherits the rule automatically**, and renaming a refusal on either side turns CI red on the renaming PR. A list of literals could only ever cover the branches its author thought of — which is what happened the first time.

## #1180 — re-assign silently rebranched from main

`fleet.sh`'s assign else-branch did `worktree add -b $branch origin/main` with **no `ls-remote` check anywhere in the file** (`grep -c ls-remote` returned 0), while `cmd_release` does `git branch -D`. Re-assigning a released lane therefore branched off main and **orphaned that PR's commits** — a full tick plus a re-review lost each time.

Assignment now looks up the remote branch and reuses it, and **fails closed** when the lookup is unreadable rather than falling back to main.

## RED → GREEN

**Before** — targeted witnesses, not blanket failures:

```
FAIL - NOT-cleared branch #1 rewrites ACTION only … (#1202)     ← the HELD branch
  ok - NOT-cleared branch #2 emits VERDICT='NOT ATTESTED'        ← already fixed by #1181
FAIL - NOT-cleared branch #3 rewrites ACTION only … (#1202)     ← the BEHIND branch
pr-ready tests: 245 passed, 4 failed

FAIL - re-assign tracks origin/<branch>, not origin/main (got 'origin/main')
FAIL - an unreadable remote lookup branched off main anyway — the #1180 failure, fail-open
fleet tests: 88 passed, 7 failed
```

That `got 'origin/main'` is the original incident report's own tell, reproduced offline.

**After:** pr-ready **253/0** (+11), fleet **96/0** (+11), and pick-next 30, watch-pr 44, main-health 109, review-quota 190, exec-bits 20 all unchanged. shellcheck clean, both workflows parse, `./scripts/check-all.sh` **12/12**.

## Why these two ship together

~40 lines of shell between them, file-disjoint from each other and from every line of Python, same defect family. Shipping them apart burns two review rounds for no reviewability gain.

### DIFF
--- a/.claude/skills/await-claude-review/SKILL.md
+++ b/.claude/skills/await-claude-review/SKILL.md
@@ -19,7 +19,7 @@ description: >-
   status polling (use `pull_request_read` directly).
 metadata:
   author: Geoff
-  version: 1.2.0
+  version: 1.3.0
 ---
 
 # Await Claude Review
@@ -77,13 +77,27 @@ This comment is not authored by `claude[bot]` — it's authored by the human use
 **Recognition.** A comment qualifies as an iteration-trigger summary when **all** of:
 
 1. The body contains the literal marker `<!-- iteration-trigger -->`.
-2. A line matches `^\*\*VERDICT\*\*:\s*(LGTM|CHANGES[_ ]REQUESTED|COMMENTS)`.
+2. A line matches `^\*\*VERDICT\*\*:\s*(\S.*)` — capture the value **verbatim**. Do not narrow this to the three review verdicts: the workflow also emits refusal values there (below), and a recogniser that cannot see them reports "malformed" for a comment that is perfectly well-formed and is explicitly refusing.
 3. A line matches `^\*\*CI\*\*:\s*(\d+)/(\d+)\s+Green` (use this to decide `merge` vs `iterate`).
+
+**Merge-critical fields: `VERDICT` and `CI`, and only those.** `**Action**:` is diagnostic prose and is never parsed for a decision (see Step 4a item 5). `iteration-trigger.yml`'s header states the same contract from the emitter's side, and both files must be updated together — a prose contract between two files is exactly what drifted in #1202.
+
+**The `VERDICT` vocabulary is therefore larger than the reviewer's.** Three values come from the review comment; the rest are the workflow **refusing to clear a merge**, and it sets them precisely so that no consumer reading `VERDICT` + `CI` can mistake the summary for permission:
+
+| `**VERDICT**` value | Meaning | Merge? |
+| ------------------- | ------- | ------ |
+| `LGTM`              | reviewer approved, and the workflow cleared every merge invariant | yes, subject to `CI: N/N` |
+| `CHANGES_REQUESTED` / `CHANGES REQUESTED` | reviewer wants changes | no — fix loop |
+| `COMMENTS`          | reviewer had non-blocking notes | no — caller decides |
+| `HELD`              | a human set `do-not-auto-merge` on the PR (or its labels were unreadable) | **never** — a human owns this PR |
+| `NOT ATTESTED`      | the verdict carries no `<!-- creek-review pr=N -->` marker for THIS PR (#1181) | **never** — surface to a human |
+| `NOT CURRENT`       | the head is behind its base, so its green describes a tree it would not merge into | **never** — sync first |
 
 **Routing.** When this comment wakes the session:
 
-- `VERDICT == LGTM` and `CI: N/N Green` (full match) → caller proceeds to merge gate. The `Action:` line will say "cleared to squash merge…".
+- `VERDICT` is exactly `LGTM` and `CI: N/N Green` (full match) → caller proceeds to merge gate. The `Action:` line will say "cleared to squash merge…".
 - `VERDICT == CHANGES_REQUESTED` or `VERDICT == COMMENTS` → caller enters fix loop. The `Action:` line will reference a comment ID to read for in-depth feedback. Pull that comment via `mcp__github__pull_request_read` `get_comments` and feed the body into `address-feedback`.
+- **Any other value** (`HELD`, `NOT ATTESTED`, `NOT CURRENT`, or one added later) → **do not merge and do not enter the fix loop.** Report the `Action:` line to the user and stop. The list above is not exhaustive by design: the rule is that only an exactly-`LGTM` verdict clears, so a value you do not recognise refuses.
 - `CI: x/y Green` where `x < y` → CI is not actually green; do not merge even on `LGTM`. Investigate the failing run before proceeding.
 
 **Currency check still applies.** The trigger fires on a `workflow_run` for a specific HEAD SHA, but the comment may arrive minutes after the push. Use the standard `created_at >= headPushedAt` guard before treating it as authoritative for the current HEAD.
@@ -149,10 +163,12 @@ Re-fetch the comments to read the full body (the webhook payload may be truncate
 When the wake event is the owner-authored iteration-trigger comment:
 
 1. **Currency check**: require `created_at >= headPushedAt`.
-2. Parse `**VERDICT**:` and `**CI**: x/y Green` from the body.
-3. If `VERDICT == LGTM` AND `x == y` (CI fully green): return `LGTM` to the caller. The follow-up `Action:` line typically reads "cleared to squash merge…". The caller proceeds to merge.
-4. Otherwise: extract the comment ID referenced in `Action: pull comment <id> …`, fetch that comment's body via `get_comments`, and pass the body to the caller (typically `address-feedback`) for triage. Treat the underlying verdict (`CHANGES_REQUESTED` or `COMMENTS`) as authoritative for the next loop.
-5. If the body is malformed (marker present but verdict line missing/unparseable): surface to user. Do not infer a verdict from the `Action:` prose.
+2. Parse `**VERDICT**:` and `**CI**: x/y Green` from the body. These two are the only merge-critical fields; `**Action**:` is prose.
+3. **Merge only on an exactly-`LGTM` verdict.** If the verdict is exactly `LGTM` AND `x == y` (CI fully green): return `LGTM` to the caller. The follow-up `Action:` line will read "cleared to squash merge…". The caller proceeds to merge. Every other value refuses — this is a whitelist, never a blacklist of known-bad verdicts, because the emitter's refusal vocabulary grows and this file learns about the additions late (#1202).
+4. If the verdict is `CHANGES_REQUESTED` or `COMMENTS`: extract the comment ID referenced in `Action: pull comment <id> …`, fetch that comment's body via `get_comments`, and pass the body to the caller (typically `address-feedback`) for triage. Treat that verdict as authoritative for the next loop.
+5. If the verdict is a **refusal** (`HELD`, `NOT ATTESTED`, `NOT CURRENT`, or any value not in items 3–4), or the body is malformed (marker present but verdict line missing/unparseable): surface the `Action:` line to the user and stop. Do not merge, do not dispatch a fix worker, and **do not infer a verdict from the `Action:` prose** — read it to the user, do not parse it for a decision.
+
+   `HELD` in particular means a human set `do-not-auto-merge`, which is the one control a human retains over an autonomous merge loop. Nothing in this skill may route around it. `scripts/ralph/pr-ready.sh` — the orchestrator's own clearance path — enforces the identical invariant independently: it prints `optout` before probing anything else, and it excludes this summary from its verdict selector entirely (`ITER_SUMMARY_RE`), so it can never read one as a verdict at all.
 
 ### Step 5: On CI Failure for Current HEAD
 
--- a/.github/workflows/iteration-trigger.yml
+++ b/.github/workflows/iteration-trigger.yml
@@ -15,9 +15,20 @@ name: Iteration Trigger
 #   * the verdict carries the `<!-- creek-review pr=N -->` provenance marker for
 #     THIS PR (#1181) — the marker pr-ready.sh's MARKER_RE parses, read from the
 #     same comment the verdict came from.
-# All three fail closed on an unreadable answer. Note that only the provenance
-# branch also rewrites `VERDICT`; see the comment at that `elif` for why the
-# other two do not, and #1202 for the gap that leaves.
+# All three fail closed on an unreadable answer.
+#
+# AND EVERY ONE OF THEM REWRITES `VERDICT`, NOT JUST `ACTION` (#1202). `VERDICT`
+# is computed from the review comment ABOVE the clearance chain, so a branch that
+# refuses while leaving it alone still posts `**VERDICT**: LGTM` + `**CI**: N/N
+# Green` — and those two fields are the ONLY ones the merge path reads
+# (`.claude/skills/await-claude-review/SKILL.md` Step 4a; its item 5 says in as
+# many words "Do not infer a verdict from the `Action:` prose"). `ACTION` is
+# diagnostic; `VERDICT` is the gate. The refusal values are chosen for their
+# BYTES — `HELD`, `NOT ATTESTED`, `NOT CURRENT` — none containing an `LGTM`
+# substring and none being one of the three verdicts SKILL.md recognises, so
+# Step 4a falls to item 5 and surfaces to a human instead of merging.
+# scripts/ralph/test_pr_ready.sh walks this chain and fails on any NOT-cleared
+# branch that leaves `VERDICT` alone, including ones added after this comment.
 #
 # Required secrets:
 #   GEOFFE_GA_PAT  - personal access token for the repo owner with `pull-requests: write`.
@@ -155,13 +166,24 @@ jobs:
 
           if [ "$GREEN" = "$TOTAL" ] && [ "$VERDICT" = "LGTM" ]; then
             if [ "$HELD" != "no" ]; then
+              # THE ONE CONTROL A HUMAN RETAINS over an autonomous merge loop, so
+              # this branch is the most important place in the file to get right —
+              # and until #1202 it was the one that failed. Rewriting ACTION alone
+              # left the summary asserting `**VERDICT**: LGTM`, which is half of
+              # what Step 4a merges on, so a webhook-woken session merged the very
+              # PR a human had parked. `HELD` carries no `LGTM` substring and is
+              # not a verdict SKILL.md recognises; see the header for the full
+              # argument about why the VERDICT field, not ACTION, is the gate.
+              VERDICT='HELD'
               ACTION="NOT cleared to merge: the do-not-auto-merge hold is set on this PR (or its labels could not be read). A human owns this one - leave it alone."
             elif [ "$MARKER_PR" != "$PR" ]; then
               # Fails CLOSED: an absent or unparseable marker leaves MARKER_PR
               # empty, which never equals a PR number, so it does not clear.
               #
               # REWRITING `VERDICT` IS WHAT ACTUALLY STOPS THE MERGE; `ACTION` IS
-              # NOT LOAD-BEARING. `.claude/skills/await-claude-review/SKILL.md`
+              # NOT LOAD-BEARING — THE RULE FOR EVERY BRANCH OF THIS CHAIN, argued
+              # here once and referenced from the other two.
+              # `.claude/skills/await-claude-review/SKILL.md`
               # Step 4a parses `**VERDICT**:` and `**CI**: x/y Green` and nothing
               # else — its item 3 returns LGTM to the caller on those two fields
               # alone, and its item 5 says in as many words "Do not infer a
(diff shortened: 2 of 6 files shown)


Geoffe-Ga:
## Summary
Fixes two related fleet-loop recovery-path defects, both shell/YAML/Markdown only (no Python touched): (1) `iteration-trigger.yml`'s HELD and BEHIND clearance branches previously left `**VERDICT**: LGTM` standing even while refusing to merge in `**Action**`, which meant a human's `do-not-auto-merge` hold could be silently defeated by any consumer that reads only `VERDICT`+`CI` (#1202); and (2) `fleet.sh`'s `assign` unconditionally branched a re-assigned lane off `origin/main` when no local ref existed, orphaning a released-but-still-remote PR branch (#1180). Both fixes are accompanied by targeted, well-reasoned tests.

## Strengths
- The #1202 fix is symmetric and complete: every "NOT cleared to merge" branch in `iteration-trigger.yml` now rewrites `VERDICT` to a value chosen for its bytes (no `LGTM` substring, not one of SKILL.md's three recognized verdicts), closing the exact asymmetry the file's own prior comment admitted.
- `SKILL.md` Step 4a is correctly converted from a blacklist to a whitelist (merge only on exactly `LGTM`), which is the right shape given the emitter's refusal vocabulary can grow.
- The cross-file test in `test_pr_ready.sh` (`iter_not_cleared_verdicts`) is genuinely structural — it awk-walks the emitter's `if`/`elif`/`else` chain and asserts the invariant per-branch rather than pinning literals, so a future added refusal branch is covered automatically, and a renamed refusal on either side of the SKILL.md/workflow contract goes red. I traced this awk logic by hand against the actual `iteration-trigger.yml` content and confirmed it correctly extracts exactly `HELD`, `NOT ATTESTED`, `NOT CURRENT` for the three current branches and nothing for the "cleared" and "no verdict yet" branches.
- The #1180 fix (`remote_branch_state`) correctly separates "absent" from "unreadable" (deliberately omitting `--exit-code`, which conflates the two), and fails closed on the unreadable case rather than guessing — exactly the discipline the incident required.
- `pr-ready.sh` being left unchanged is justified with a specific, verifiable argument (`ITER_SUMMARY_RE` excludes the iteration-trigger comment from its verdict selector; `has_optout_label` and `branch_is_current` enforce the equivalent invariants natively) — I grep-verified all three referenced mechanisms exist in `pr-ready.sh` exactly as described, so this isn't a hand-wave.
- New test coverage for `fleet.sh` reproduces the actual incident (release → remote branch survives → re-assign must reattach to the PR's tip and track `origin/<branch>`, not `origin/main`) plus a fail-closed case using a `git` shim that breaks only `ls-remote`.
- `ralph-recap-tests.yml`'s path filters are extended (on both `push` and `pull_request`) to include `SKILL.md`, closing the coupling gap that let the consumer and emitter drift in the first place.

## Security Concerns
🟢 LOW — No new attack surface. The new `VERDICT` values are hardcoded literals, not attacker-influenced. `remote_branch_state`'s awk match is an exact string compare (not a regex), so a branch name can't be crafted to match an unrelated ref via glob tricks — and branch names here are always `issue/<numeric>-<sanitized-slug>` anyway.

## Problems
None found. I could not execute `test_fleet.sh`/`test_pr_ready.sh` directly in this review sandbox (script execution was blocked by the environment's permission gate), so I instead manually traced the new awk extraction logic line-by-line against the actual committed `iteration-trigger.yml` and independently verified every cross-file claim (`ITER_SUMMARY_RE`, `has_optout_label`, `branch_is_current`, the `exactly \`LGTM\`` string match) against the real file contents. Everything checks out; I'm not flagging this as blocking given the PR description's own stated RED→GREEN counts (pr-ready 253/0, fleet 96/0) and the structural, non-vacuous nature of the added assertions.

## Code Quality
- The header/inline comments in `iteration-trigger.yml` and `fleet.sh` are extremely dense and long (e.g. a 30+ line header comment). This matches the codebase's existing convention (the pre-existing comments in these files are equally verbose), so it's not a new pattern this PR introduces, but it's worth noting as a general readability tax on future changes to this file — non-blocking.
- Minor: the PR's own prose duplicates the "why" in several places (workflow header, inline comment at each branch, and the test file) — again consistent with the existing house style of over-documenting fragile invariants, not a defect.

## CLAUDE.md Adherence
✅ No `creek-tools/` Python code touched, so the Python quality gates (coverage, mypy, ruff, docstring coverage) don't apply here.
✅ Follows the "Stay Green" 4-gate workflow described in CLAUDE.md: tests were written to demonstrate RED against the pre-fix behavior (per the PR body) before the fix, and the new assertions are structural rather than incidental.
✅ Commit message follows Conventional Commits (`fix(ralph): ...`) with `Closes #1202. Closes #1180.`

## Testing
Strong. Both fixes ship with tests that reproduce the actual failure mode rather than merely asserting the fix's mechanism:
- `test_pr_ready.sh`'s new block asserts, per NOT-cleared branch, that `VERDICT` is set, contains no `LGTM` substring, and isn't one of SKILL.md's three recognized verdicts — and separately asserts SKILL.md names every emitted refusal verbatim. This is a good "coupling" test pattern given the bug was literally two files drifting in prose.
- `test_fleet.sh`'s new block round-trips the #1180 incident: commit + push on a lane, release (deletes local ref only), re-assign, and check the lane lands on the PR's tip with the correct upstream — plus a fail-closed case for an unreadable remote using a `git` shim scoped only to `ls-remote`.
- Test W7 in `test_pr_ready.sh` closes a plausible follow-on mutant (treating the iteration-trigger marker as a second provenance marker) rather than just re-testing the fix directly.

## Documentation
Thorough — the workflow header, `SKILL.md`'s new vocabulary table, and `fleet.sh`'s subcommand docstring for `assign`/`release` are all updated consistently with the behavior change, and the emitter/consumer contract is stated in both files as required by the new coupling test.



## Verdict: LGTM

Both defects are fixed correctly and symmetrically (every NOT-cleared branch in iteration-trigger.yml now rewrites VERDICT with a byte-chosen, unrecognized value; fleet.sh's assign now distinguishes "no remote branch" from "unreadable remote" and fails closed). The new tests are structural (an awk walk of the emitter's clearance chain) rather than literal lists, so they survive a future refusal branch being added, and I traced the awk logic by hand against the actual workflow file and confirmed it extracts the right VERDICT per branch. No bugs found; nothing here should block merge.

Geoffe-Ga:
**CI**: 6/9 Green
**VERDICT**: LGTM
**Action**: pull comment 5227944389 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 12/12 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
Fixes two fleet-loop recovery-path defects: (#1202) a human's `do-not-auto-merge` hold could be silently defeated because two of three "not cleared" branches in `iteration-trigger.yml` rewrote only `ACTION`, leaving `**VERDICT**: LGTM` standing for the merge-consuming Step 4a to read; and (#1180) re-assigning a released Ralph lane could silently rebranch off `origin/main`, orphaning the PR's commits. Both fixes are shell/YAML/docs only, and both are backed by new, well-targeted tests.

## Strengths
- The core fix pattern for #1202 — make every refusal branch rewrite `VERDICT` to a value that (a) contains no `LGTM` substring and (b) is not one of the three verdicts the consumer recognizes — directly closes the actual hole, and the consumer side (`SKILL.md` Step 4a) is correctly changed from a blacklist to a whitelist ("merge only on exactly `LGTM`"), which is the right shape for an emitter whose refusal vocabulary can grow.
- `test_pr_ready.sh`'s new cross-file coupling check (`iter_not_cleared_verdicts`) is structural rather than a list of literals: it walks the emitter's `if`/`elif`/`else` chain via awk, correctly resetting the tracked value at each branch boundary so a sibling's `VERDICT=` assignment is never mis-credited, and asserts the invariant against every branch including ones added later. This is a strong design against the exact failure mode that let #1202 happen in the first place (the #1181 fix only patched the one branch its author touched).
- `fleet.sh`'s `remote_branch_state` correctly distinguishes "absent" from "unreadable" (no `--exit-code`, since that conflates the two), uses a fully-qualified `refs/heads/<branch>` pattern plus an exact `awk` string comparison to avoid `ls-remote`'s tail/suffix component matching, and the caller fails closed (`die`) on the unreadable case rather than falling through to branching off main.
- Verified the new tests are correctly wired into the existing test harness: `iter_summary`, `cj`, `recap_trigger_block`, `$BIN`/`$WORK`/`$REPO`/`$LOCKDIR`, `run`/`ok`/`bad`/`check` are all pre-existing helpers used with matching signatures, not new ad hoc scaffolding.
- `ralph-recap-tests.yml`'s path filters are correctly extended (both `push` and `pull_request`) to include `SKILL.md`, closing the coupling gap the PR itself identifies (an emitter-only or consumer-only PR must run the suite that compares them).
- Deliberately *not* touching `pr-ready.sh` for #1202, with the reasoning pinned by a new test case (W7) rather than left as an unverified claim, is the right call — it avoids dead-code guarding a route that provably can't be taken (`ITER_SUMMARY_RE` already excludes the summary from the verdict selector).
- YAML files parse and the shell files pass a syntax check (manually verified in review); `usage()`'s header line range (`sed -n '2,60p'`) was correctly extended to match the grown header comment block, and a new test (`test_fleet.sh`) pins the last line so a future header growth without updating the range would be caught.

## Security Concerns
None. 🟢 No new attack surface — this only changes what strings a same-repo, PAT-authored workflow writes into a PR comment and how existing internal skill/test tooling parses them. The `remote_branch_state` fix in `fleet.sh` closes a data-integrity risk (silently orphaning/potentially force-push-destroying a PR's commits) but does not touch untrusted external input.

## Problems
None found that block merge.

## Code Quality
- Very heavy comment density throughout (multi-paragraph comments before nearly every changed line). This is a deliberate, consistent house style visible elsewhere in this repo's `scripts/ralph/` tree (extensive "why" documentation for infra/recovery-path code), so it's not out of place here, but it is worth noting as a genuine density outlier relative to typical code — a future contributor extending this chain will need to read a lot of prose per branch. Non-blocking; consistent with the file's existing style before this PR.
- Minor: `remote_branch_state`'s `awk` exact-match check is a second, arguably-belt-and-suspenders layer of protection given the pattern is already fully qualified with a 4-path-component pattern that the PR's own analysis shows can't ambiguously tail-match. Not wrong, just slightly more defense than the stated failure mode requires — a fine trade for a fail-closed integrity guard.

## CLAUDE.md Adherence
✅ No Python touched, so the strict coverage/mypy/ruff gates in `creek-tools/CLAUDE.md` don't apply to this change (correctly scoped, per the PR description).
✅ Commit messages follow Conventional Commits (`fix(ralph): ...`) and are detailed, matching this repo's evident convention of thorough commit bodies.
✅ Changes are file-disjoint between the two fixed issues and scoped tightly to the stated defects — no unrelated refactoring bundled in.
N/A `./scripts/check-all.sh` / lint/typecheck scripts are Python-project-specific and don't apply to this shell/YAML-only change; the PR description reports the equivalent shell test suites and `shellcheck` were run clean, which I was unable to independently execute in this sandbox (script execution required approval I didn't have) but which I have no reason to doubt given the manual read-through matched the described before/after behavior exactly.

## Testing
✅ Strong, targeted testing. `test_fleet.sh` adds a full round-trip reproduction of the #1180 incident (commit → push → release → re-assign, asserting the re-assigned lane lands on the PR's tip and tracks `origin/<branch>` not `origin/main`), plus a fail-closed case using a `git` shim that breaks only `ls-remote` while leaving `fetch` working, verifying both the abort and that no worktree/branch/lock is left behind. `test_pr_ready.sh` adds W7 (a refusal-verdict summary must still read as "no verdict posted", pinned as a mutant-killer against accepting the iteration-trigger marker as a second provenance marker) plus the structural coupling check described above. I traced through the awk-based chain-walker logic and comment-skipping by hand against the actual `iteration-trigger.yml` content and it correctly extracts exactly the three refusal branches (HELD, NOT ATTESTED, NOT CURRENT) with no false credit across branches.
Not independently re-run in this sandbox (script execution wasn't available to me here), but the manual trace gives me confidence in the reported 253/0 and 96/0 results.

## Documentation
✅ Excellent — the `SKILL.md` update adds a clear vocabulary table for `**VERDICT**` values and explicitly states the whitelist rule; `iteration-trigger.yml`'s header and `fleet.sh`'s subcommand docs are updated in lockstep with the behavior change; the PR description itself is a model of a clear before/after RED→GREEN writeup with the actual incident context (#1180's live occurrence, #1202's mechanism).



## Verdict: LGTM

Both fixes directly close the defects they claim to (verified by manually tracing the awk-based structural test against the actual workflow file and the fail-closed remote-lookup logic in fleet.sh), are backed by well-targeted new tests correctly wired into the existing harness, and introduce no security or correctness issues. Nothing here would make me block merge; the only remarks are minor style observations (comment density) that don't rise to a substantive suggestion.

Geoffe-Ga:
**CI**: 7/10 Green
**VERDICT**: LGTM
**Action**: pull comment 5228068874 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1237, opened 2026-08-08
fix(lint,generate): stop swallowing two failure modes that reported success

### LINKED ISSUE #1038: fix(lint): a single malformed-YAML note crashes `creek lint` — 3 checks omit yaml.YAMLError
**Detected by:** de-slopify targeted scan (creek-pipeline) 2026-07-31, at HEAD `0f7689a`
**Severity:** High
**Remediation:** fix (add the missing exception class to three tuples)

## Problem

Three of the ten `creek lint` checks omit `yaml.YAMLError` from the `except` tuple that guards `frontmatter.load`. `yaml.YAMLError` is **not** a `ValueError` subclass, so the `continue` that was written to skip an unparseable note never runs, and the parser error escapes all the way to the user as a raw traceback. Because `compost` is a **deterministic** check, plain `creek lint` — no flags — is broken by a single malformed note anywhere under `10-Liminal/Compost/`.

Taxonomy family: **0 (Correctness — swallowed/unhandled exception path)** and **5 (Inconsistent error contract)**.

**Current state:** one note with unclosed YAML frontmatter takes down the whole command.

- `creek-tools/creek/lint/checks/compost.py:33` — `except (OSError, ValueError):`
- `creek-tools/creek/lint/checks/synchronicity.py:32` — `except (OSError, ValueError):`
- `creek-tools/creek/lint/checks/paradox.py:41` — `except (OSError, ValueError, ValidationError):`

`creek/cli.py:3227` catches only `ValueError`, so nothing downstream contains it either.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** A vault containing one note whose frontmatter is `tags: [a, b` (unclosed), driven through the real Typer entry point with `CliRunner`:

```
yaml.YAMLError MRO: ['YAMLError', 'Exception', 'BaseException', 'object']
is a ValueError? False

creek lint compost                  exit=1 exception=ParserError
creek lint synchronicity            exit=1 exception=ParserError
creek lint paradox                  exit=1 exception=ParserError
creek lint (default, no --check)    exit=1 exception=ParserError
```

**Signal 2 — structural diff against the in-repo convention.** Every one of these three sites ends in `continue`, so skipping the bad file is unambiguously the intended behaviour. The correct tuple is already used by the sibling checks and by the canonical loader:

- `creek-tools/creek/lint/checks/draft_grounding.py:61` — `except (OSError, ValueError, yaml.YAMLError)`
- `creek-tools/creek/lint/checks/voice_fidelity.py:103` — same
- `creek-tools/creek/vault/reader.py:103` — same (the module that declares itself the single source of truth for loading fragments)

Five sites, three spellings of one contract; the three that diverge are the three that crash.

## Scope

Covers only the missing `yaml.YAMLError` in these three lint checks plus regression coverage. It does **not** cover: the wider `--since`/`SEMANTIC_CHECKS` gaps in the same package, the identical missing-`yaml.YAMLError` pattern in `creek/generate/unnamed.py:541`, or the analogous `httpx.HTTPError` gap in `creek/classify/` — those are separate findings.

## Tasks

1. Add `yaml.YAMLError` to the `except` tuple in `creek/lint/checks/compost.py:33`, `creek/lint/checks/synchronicity.py:32`, and `creek/lint/checks/paradox.py:41` (import `yaml` where not already imported).
2. Add a regression test that builds a vault containing a note with unclosed YAML frontmatter and asserts `creek lint` (no `--check`) exits cleanly and skips that file, via `CliRunner` against the real command.
3. Add per-check tests covering `--check compost`, `--check synchronicity`, `--check paradox` on the same vault.

## Acceptance Criteria

- [ ] `creek lint` with no flags exits 0 on a vault containing a malformed-YAML note and reports the other checks' findings normally.
- [ ] `creek lint --check {compost,synchronicity,paradox}` each exit 0 on that vault; no `ParserError` reaches the user.
- [ ] The new tests fail on `main` before the fix and pass after.
- [ ] All pre-commit hooks pass (`cd creek-tools && ./scripts/check-all.sh`).

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/lint/checks/compost.py` | Modify |
| `creek-tools/creek/lint/checks/synchronicity.py` | Modify |
| `creek-tools/creek/lint/checks/paradox.py` | Modify |
| `creek-tools/tests/test_lint.py` (or a new `test_lint_malformed_yaml.py`) | Modify / Create |

### LINKED ISSUE #1231: fix(generate): `creek generate indexes` silently writes nothing when a frequency folder is missing
## Summary

`creek/generate/indexes.py` writes into vault directories **without calling
`mkdir`**, and one of those paths degrades to a silent no-op rather than an
error when the directory is absent.

Found while sweeping production write targets for #1025.

## The silent path

`_generate_frequency_indexes` (`creek/generate/indexes.py:493`, `:512`)
resolves `06-Frequencies/<F*>` through `_find_frequency_subdir`, which returns
`None` when the subdirectory does not exist. The index is then **skipped** —
not created, not reported.

So on a vault whose `06-Frequencies` folders were renamed, removed, or never
scaffolded, `creek generate indexes` **reports success and writes nothing.**
The user gets an exit code of 0 and an empty result.

## The fragile paths

Two more writes rely entirely on the scaffold already existing, with no
`mkdir` and no guard:

- `02-Threads/Thread-Index.md` (`:555`)
- `03-Eddies/Eddy-Map.md` (`:589`)

These at least fail loudly (`FileNotFoundError`) rather than lying, but they
are inconsistent with every other writer in the codebase, which all call
`mkdir(parents=True, exist_ok=True)` — see `creek/vault/writer.py:384`, `841`,
`876`, `1054`, `1219`, `1237`.

## Why this matters beyond the edge case

This is the failure mode epic #1024 exists to catch: a command that is wired to
something, runs, exits 0, and produces nothing. A test asserting "the command
succeeded" passes. Only an assertion on the *output* catches it.

#1025 scaffolds `06-Frequencies/F1`–`F10`, so a freshly-initialised vault is
fine today — which is exactly what makes this worth fixing now rather than
later. The scaffold masks it; a real vault that has been reorganised does not.

## Acceptance criteria

- [ ] A missing `06-Frequencies/<F*>` either creates the directory or fails
      with a message naming the missing path. It must not silently skip.
- [ ] The thread and eddy index writes use the same
      `mkdir(parents=True, exist_ok=True)` convention as every other writer.
- [ ] A test asserting `creek generate indexes` against a vault with a missing
      `06-Frequencies` folder produces output or errors — the current behaviour
      (exit 0, no files, no message) must fail it.
- [ ] Decide and record whether "index over a frequency with zero fragments"
      should write an empty index or skip; skipping is defensible, but it must
      be distinguishable from the missing-directory case.

## Related

- #1024 (epic: integration testing that catches features wired to nothing)
- #1025 (the scaffold/drift guard whose write-target sweep surfaced this)

Geoffe-Ga:
Closes #1038. Closes #1231.

Two issues, one PR — same defect class (a failure swallowed with no report), disjoint modules, ~40 lines of production change between them.

## #1038 — one malformed note crashed `creek lint`

`compost.py` and `synchronicity.py` caught `(OSError, ValueError)`; `paradox.py` added `ValidationError`. **None catches `yaml.YAMLError`** — and it is *not* a `ValueError` subclass (verified: MRO is `YAMLError -> Exception -> BaseException`). So one malformed frontmatter block took the whole command down. `draft_grounding.py` already had the correct contract; the three now match it.

## #1231 — `creek index` exited 0 having written nothing

A missing `06-Frequencies/<F*>` made the resolver return `None`; the code hit a bare `continue`. The command reported success and produced nothing.

## Three corrections to the issues

- **There is no `creek generate indexes` command.** It is `creek index`, and the same generator also runs inside `creek process` and `_sync_index`.
- **The missing-mkdir defect is wider than reported.** The issue names 02-Threads and 03-Eddies; `generate_temporal_index` and `generate_source_index` write to 00-Creek-Meta with no mkdir either. Fixed all four.
- **A latent *second* silent-skip.** The packaged template ships suffixed folders (`F1-Agency` … `F10-Emptiness`) while the resolver matched only `startswith("F1-")`. A vault scaffolded with a bare `F1` folder would never have been found. Now matched by exact name as well.

## An existing test asserted the bug was correct

`test_missing_frequency_subdirs_creates_files_only_for_existing` asserted `len(result) == 2` on a partial scaffold — it pinned the silent skip as *intended*. It had to be rewritten, not supplemented. **That is the #1024 failure mode living inside the test suite itself**, and worth noting for the rest of that epic.

## Design call

Create-the-directory over fail-with-a-message (the AC permits either): failing hard would newly break `creek process` on a reorganised vault. Canonical folder names are read from `creek.scaffold.VAULT_TEMPLATE_DIR` at runtime rather than restated, so the two cannot drift.

## RED proof

- #1038 stashed → `10 failed`, `assert 1 == 0 ... ParserError`
- #1231 stashed → `14 failed`, `assert 1 == 0 ... FileNotFoundError`

**Honest caveat:** 2 of those 14 monkeypatch a constant the unfixed module doesn't define, so they fail with `AttributeError` — they are branch-coverage tests for the new fallback, **not** regression proofs. Flagged rather than counted as reproductions.

## Found, not fixed

A note whose frontmatter contains a `content:` key makes `frontmatter.load` raise `TypeError`, which none of these guards catch — nor do the correct-contract siblings. Deserves its own issue.

Gate: **12/12**.

### DIFF
--- a/creek-tools/creek/generate/indexes.py
+++ b/creek-tools/creek/generate/indexes.py
@@ -4,6 +4,20 @@ Generates Obsidian Dataview-powered index notes for frequencies, threads,
 eddies, temporal views, and source statistics. Each generated note contains
 YAML frontmatter and Dataview query blocks that dynamically aggregate
 vault content.
+
+Every write target is created on demand
+(``mkdir(parents=True, exist_ok=True)``), matching the convention in
+:mod:`creek.vault.writer`. A vault whose folders were renamed, removed, or
+never scaffolded is repaired rather than skipped: issue #1231 recorded the
+opposite behaviour, where a missing ``06-Frequencies/<F*>`` made the whole
+command exit ``0`` having written nothing.
+
+The generated notes are Dataview *queries*, not materialised content, so
+they do not depend on how many fragments a frequency holds. A frequency with
+zero fragments still gets an index; it simply renders an empty table in
+Obsidian. There is therefore no "skip" case at all, which is what keeps the
+missing-directory case unambiguous — if a note is absent, the generator
+never ran.
 """
 
 from __future__ import annotations
@@ -11,9 +25,38 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from creek.models import Frequency
+from creek.scaffold import VAULT_TEMPLATE_DIR
 
 if TYPE_CHECKING:
     from pathlib import Path
+
+_FREQUENCY_ROOT = "06-Frequencies"
+"""Vault-relative folder holding the per-frequency subdirectories."""
+
+
+def _canonical_frequency_dirname(freq_code: str) -> str:
+    """Return the folder name ``creek init`` deploys for *freq_code*.
+
+    The canonical names (``F1-Agency``, ``F10-Emptiness``, …) are read from
+    the packaged vault template — the same tree :func:`creek.scaffold
+    .scaffold_vault` copies — rather than restated here, so the two can never
+    drift. A template that carries no folder for *freq_code* falls back to the
+    bare code, which :meth:`IndexGenerator._find_frequency_subdir` also
+    matches, keeping the generator idempotent either way.
+
+    Args:
+        freq_code: Frequency code to resolve (e.g., ``"F1"``).
+
+    Returns:
+        The canonical subdirectory name for *freq_code*.
+    """
+    template_root = VAULT_TEMPLATE_DIR / _FREQUENCY_ROOT
+    if template_root.is_dir():
+        for child in sorted(template_root.iterdir()):
+            if child.is_dir() and child.name.startswith(f"{freq_code}-"):
+                return child.name
+    return freq_code
+
 
 FREQUENCY_NAMES: dict[Frequency, str] = {
     Frequency.F1: "Agency/Survival",
@@ -482,22 +525,25 @@ class IndexGenerator:
     def generate_frequency_indexes(self) -> list[Path]:
         """Generate one index note per frequency in 06-Frequencies/ subdirs.
 
-        Scans existing frequency subdirectories (F1 through F10) and creates
-        a markdown index note in each one containing ontology descriptions,
-        Dataview queries for fragments/threads, dosage comparison tables,
-        statistics, and cross-frequency developmental links.
+        Creates a markdown index note in each frequency subdirectory (F1
+        through F10) containing ontology descriptions, Dataview queries for
+        fragments/threads, dosage comparison tables, statistics, and
+        cross-frequency developmental links.
+
+        A frequency subdirectory that does not exist is **created** under its
+        canonical name rather than skipped (#1231), so the return value always
+        has one entry per frequency.
 
         Returns:
-            List of paths to the generated frequency index files.
+            List of paths to the generated frequency index files — one per
+            entry in :data:`FREQUENCY_NAMES`, in that order.
         """
-        freq_dir = self.vault_path / "06-Frequencies"
+        freq_dir = self.vault_path / _FREQUENCY_ROOT
         generated: list[Path] = []
 
         for freq, name in FREQUENCY_NAMES.items():
             freq_code = freq.value
-            subdir = self._find_frequency_subdir(freq_dir, freq_code)
-            if subdir is None:
-                continue
+            subdir = self._ensure_frequency_subdir(freq_dir, freq_code)
 
             color = FREQUENCY_COLORS[freq]
             frontmatter = {
@@ -553,6 +599,7 @@ class IndexGenerator:
         body += "```\n"
 
         note_path = self.vault_path / "02-Threads" / "Thread-Index.md"
+        note_path.parent.mkdir(parents=True, exist_ok=True)
         note_path.write_text(_build_note(frontmatter, body), encoding="utf-8")
         return note_path
 
@@ -587,6 +634,7 @@ class IndexGenerator:
         body += "```\n"
 
         note_path = self.vault_path / "03-Eddies" / "Eddy-Map.md"
+        note_path.parent.mkdir(parents=True, exist_ok=True)
         note_path.write_text(_build_note(frontmatter, body), encoding="utf-8")
         return note_path
 
@@ -628,6 +676,7 @@ class IndexGenerator:
         body += "```\n"
 
         note_path = self.vault_path / "00-Creek-Meta" / "Temporal-Index.md"
+        note_path.parent.mkdir(parents=True, exist_ok=True)
         note_path.write_text(_build_note(frontmatter, body), encoding="utf-8")
         return note_path
 
@@ -662,6 +711,7 @@ class IndexGenerator:
         body += "```\n"
 
         note_path = self.vault_path / "00-Creek-Meta" / "Source-Index.md"
+        note_path.parent.mkdir(parents=True, exist_ok=True)
         note_path.write_text(_build_note(frontmatter, body), encoding="utf-8")
         return note_path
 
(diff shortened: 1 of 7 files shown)


Geoffe-Ga:
## Summary
Fixes two related silent-failure bugs: (1) `creek lint`'s compost/synchronicity/paradox checks caught `(OSError, ValueError)` around `frontmatter.load`, but `yaml.YAMLError` is not a `ValueError` subclass, so one malformed frontmatter block crashed the whole command; (2) `creek index` returned `None` from `_find_frequency_subdir` on a missing `06-Frequencies/<F*>` folder and hit a bare `continue`, exiting 0 having written nothing — and the same missing-mkdir defect existed for the thread/eddy/temporal/source writers. Both are fixed by matching an existing correct contract (`draft_grounding.py`/`vault/reader.py` for the YAML case) and by `mkdir(parents=True, exist_ok=True)` at every write site, with canonical frequency folder names sourced from the packaged template at runtime instead of restated.

## Strengths
- Root-caused correctly and precisely: verified `yaml.YAMLError`'s actual MRO rather than assuming it was a `ValueError`, and traced `_find_frequency_subdir`'s `None` path all the way to the silent `continue`.
- Widened the fix beyond the two issues' literal scope (all four write targets, not just the two named) with a clear rationale, and caught a *second* latent silent-skip (bare `F1` folders never matching `startswith("F1-")`) — verified by reading the actual template tree (`F1-Agency`, `F10-Emptiness`, …), which do match the new `_canonical_frequency_dirname` lookup logic.
- The one existing test that pinned the bug as intended behavior (`len(result) == 2`) was rewritten, not left alongside a new test — avoids the "two contradictory assertions" anti-pattern.
- Excellent edge-case coverage: bare-name folders, empty `06-Frequencies`, absent `06-Frequencies` entirely, absent/partial packaged template (via monkeypatching `indexes.VAULT_TEMPLATE_DIR`), idempotency after self-healing, and full `generate_all` on a bare vault. I traced through each by hand against the implementation and they all hold.
- Honest, well-disclosed caveats in the PR description: flags that 2 of the 14 RED-phase failures are branch-coverage tests rather than regression proofs, and separately discloses a known-but-unfixed `TypeError` path (a `content:` frontmatter key) rather than silently leaving it or overclaiming full coverage.
- No circular import risk: `creek.scaffold` has no dependency back on `creek.generate.indexes`.
- Design justification (create-directory over fail-hard) is reasoned against a concrete regression risk (`creek process` on a reorganised vault) rather than asserted.

## Security Concerns
None. `freq_code` values originate only from the fixed `Frequency` enum (not user input), and all new `mkdir`/write targets are built from fixed subpath literals rooted at `vault_path`, so there's no path-traversal exposure introduced.

## Problems
None found.

## Code Quality
- 🟢 Minor: no `CHANGELOG.md` entry for either fix, though the repo's `Unreleased` section otherwise tracks design-trace IDs (`FEAT-*`, `INC-*`, `BUG-*`) — not required by CLAUDE.md or CI, so purely optional polish.
- 🟢 The "found, not fixed" `TypeError`-from-`content:`-key case is disclosed in the PR body but not filed as a tracked issue link; worth filing before it's forgotten, as noted in the description itself.

## CLAUDE.md Adherence
✅ TDD stated and RED-phase evidence given (10 failed / 14 failed pre-fix) in the PR description.
✅ Matches existing contracts instead of duplicating logic (DRY — reads `VAULT_TEMPLATE_DIR` rather than restating folder names; mirrors `draft_grounding.py`'s exception contract).
✅ Full type hints, Google-style docstrings on every new/changed public and private member.
✅ No forbidden shortcuts (no `# noqa`, `# type: ignore`, or lowered thresholds) observed in the diff.
⚠️ Could not independently execute `./scripts/check-all.sh` or `gh pr checks` in this sandboxed review (both required elevated approval), so CI green/Gate-2-local-pass is taken on the PR author's stated "Gate: 12/12" rather than independently confirmed — recommend the CI status be checked before merge if not already done.

## Testing
Insufficient testing would be blocking, but this PR clears the bar comfortably. `test_lint_malformed_yaml.py` (new, 221 lines) exercises all three lint checks plus the real Typer CLI entry point against genuinely malformed YAML (unterminated flow sequence → `yaml.ParserError`), asserting both that the bad note is skipped *and* that surrounding good notes are still reported — not just that nothing crashes. `test_generate.py` additions cover the full matrix of frequency-folder states (bare name, suffixed name, empty root, absent root, absent/partial template) plus idempotency and the four non-frequency writers' directory creation. `test_cli_index.py` additions verify the fix end-to-end through the CLI on a vault with no scaffolding at all. I traced the assertions against the actual implementation (`_ensure_frequency_subdir`, `_canonical_frequency_dirname`, `_find_frequency_subdir`, the CLI's `index` command output format) and found them all consistent with real behavior, not merely plausible-looking.

## Documentation
Module and function docstrings are thorough and explain the *why* (issue numbers, the design tradeoff between create-vs-fail, cross-references to the sibling `draft_grounding.py`/`vault/reader.py` contract) rather than restating the *what*. The updated module docstring in `indexes.py` clearly explains the self-healing behavior and its rationale.



## Verdict: LGTM

No blocking bugs, security issues, or missing tests. Both defects are fixed correctly (yaml.YAMLError now caught to match the established contract in draft_grounding.py/vault/reader.py; mkdir(parents=True, exist_ok=True) added to all four write targets, sourced from the packaged template rather than restated). The RED→GREEN evidence is verified against the actual code, the previously-lying test (`len(result) == 2`) is properly rewritten rather than papered over, and coverage of edge cases (bare folder names, empty/absent 06-Frequencies, absent template) is thorough.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5228278324 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

PR #1238, opened 2026-08-08
fix(api): log the exception behind a /v1 500, and refuse non-lifespan scopes

### LINKED ISSUE #1122: fix(api): log the exception behind a /v1 500 to the error logger
`creek_mcp/httpapi/middleware/boundary.py:59-70` swallows the exception entirely. Discarding it
from the *response* is correct and deliberate (no internals to the caller), but discarding it from
the *process* leaves the operator with a `500` and no traceback anywhere — an undiagnosable fault.

Must log to a non-access logger (never the access logger, never into the response body).

Test idea: a faulting handler emits exactly one ERROR record carrying the traceback on the error
logger, and the response body still contains only `{code, message, request_id}`.

---

### LINKED ISSUE #1124: fix(api): scope the non-http ASGI passthrough to lifespan only
Every `/v1` middleware passes non-`http` scopes straight through
(`access_log.py:92`, `limits.py:167/208/245`, `ceiling.py`, `auth.py:150`), which today is
justified solely by `lifespan`.

The passthrough denies by omission rather than allowlisting: the first `websocket` route ever
mounted would ship unauthenticated, unceilinged and unlogged by default. No live exposure today —
no ws routes exist — which is precisely why it should be closed before one does.

Fix: allowlist `lifespan` explicitly and refuse anything else.

Test idea: an AST or behavioural guard that a `websocket` scope does not reach the app through the
auth middleware.

---

Geoffe-Ga:
Closes #1122. Closes #1124. Same middleware tree, same test file — they could never have been parallel.

## #1122 — the traceback wasn't merely unlogged, it was **unreachable**

`boundary.py`'s docstring justified logging nothing by saying an operator wanting the traceback "runs the server with debug disabled but their own handler attached." **That escape hatch does not exist.** Starlette's `ServerErrorMiddleware` is the outermost layer of `build_middleware_stack`, so once `ErrorBoundaryMiddleware` catches the exception it never propagates outward — no handler an operator attaches can ever receive it.

So the issue's premise was right about the consequence and wrong about the cause. I rewrote that paragraph rather than leave a false workaround in the docs.

The response body is unchanged byte-for-byte. The test asserts the contrast directly: the sentinel appears in `caplog.text` and does **not** appear in `response.text`.

## #1124 — a websocket scope was silently **answered**

Verified against the installed Starlette: `Router.app` asserts the scope type is one of http/websocket/lifespan, no `Route` matches a websocket scope, so it falls through to `Router.not_found`, which sends `{"type": "websocket.close"}`. **Nothing raises, logs, or 500s** — which is precisely why the hole is invisible without a test.

**The issue undercounts the sites.** It names six; there are seven — `boundary.py:53` carries the same bare `if scope["type"] != HTTP_SCOPE` passthrough and isn't listed. Every line number it *does* give is correct, which is unusual for this repo.

## Design call, stated as the issue asked

The refusal is a raised `UnsupportedScopeError`, not a protocol-level refusal. There is no protocol-independent way to refuse an arbitrary ASGI scope — a websocket needs `websocket.close`; an unknown type has no defined refusal at all. Reaching the stack with such a scope means something was mounted this app was never wired for: a deployment bug that should fail loudly at wiring time, not a request-level failure. Nothing is written to the wire (asserted).

## Three tests pass against unfixed code, by design

Flagged rather than hidden:
- the published-constant pin (new code — nothing to be "unfixed")
- seven `lifespan` passthrough guards — they'd go red if the allowlist **over**-tightened
- the stack-identity twin, asserting the parametrization equals `_EXPECTED_STACK`, so an eighth middleware can't be added without being dragged through the websocket test

The ten that carry the fix are in the RED proof.

## A note on how RED was produced

`git stash push <file>` does **not** work here — the new symbols live in the same files as the fixes, so a whole-file stash yields only an `ImportError` at collection. That proves the symbols are new and nothing about behaviour. The edits were reverted **in place** instead, keeping imports resolvable and producing genuine behavioural failures (`DID NOT RAISE`, `assert 0 == 1`). Both runs reported; `git status` verified clean afterwards.

## Deliberately out of scope

`creek-tools-api` still starts uvicorn with websocket protocol support enabled. The middleware guard is the durable fix because it doesn't depend on server configuration; passing `ws="none"` as belt-and-braces is a possible follow-up.

Gate: **12/12**.

### DIFF
--- a/creek-tools/creek_mcp/httpapi/auth.py
+++ b/creek-tools/creek_mcp/httpapi/auth.py
@@ -37,7 +37,7 @@ from typing import TYPE_CHECKING, Final
 from starlette.datastructures import Headers
 
 from creek_mcp.api.models import ErrorCode
-from creek_mcp.httpapi.context import HTTP_SCOPE, context_of
+from creek_mcp.httpapi.context import HTTP_SCOPE, context_of, pass_through
 from creek_mcp.httpapi.errors import error_response
 from creek_mcp.remote_auth import (
     CONSUMER_TOKENS_ENV,
@@ -148,7 +148,7 @@ class BearerAuthMiddleware:
             send: The ASGI send channel.
         """
         if scope["type"] != HTTP_SCOPE:
-            await self.app(scope, receive, send)
+            await pass_through(self.app, scope, receive, send)
             return
         context = context_of(scope)
         token = _presented_token(scope)
--- a/creek-tools/creek_mcp/httpapi/context.py
+++ b/creek-tools/creek_mcp/httpapi/context.py
@@ -20,6 +20,13 @@ travels inward; freezing it would mean rebuilding and re-attaching it at four
 layers, and a rebuild is exactly where a field gets dropped. Nothing here is a
 security decision: :class:`creek_mcp.policy.CallerIdentity`, which *is* one,
 stays frozen and is constructed fresh at the single admission site.
+
+**It also owns which ASGI scopes the stack answers at all.** That belongs here
+for the same reason the record does: every layer has to agree, and seven copies
+of the rule are seven places for it to drift. :data:`HTTP_SCOPE` names what the
+stack acts on, :data:`LIFESPAN_SCOPE` names the one thing it relays, and
+:func:`pass_through` is the single site that decides between them and refuses
+everything else (#1124).
 """
 
 from __future__ import annotations
@@ -31,16 +38,32 @@ from uuid import uuid4
 from creek_mcp.tier_ceiling import TierCeiling
 
 if TYPE_CHECKING:
-    from starlette.types import Scope
+    from starlette.types import ASGIApp, Receive, Scope, Send
 
 HTTP_SCOPE: Final[str] = "http"
 """The ASGI scope type every ``/v1`` middleware acts on.
 
-Stated once and imported by all seven. Each of them passes anything else —
-``lifespan`` above all — straight through untouched: a middleware that assumed
-``http`` would break the startup handshake, and the failure would surface as
-every request failing rather than as a middleware bug.
+Stated once and imported by all seven. Anything else goes to
+:func:`pass_through`, which is where the *allowlist* lives.
 """
+
+LIFESPAN_SCOPE: Final[str] = "lifespan"
+"""The one other scope type the ``/v1`` stack hands downward (#1124).
+
+A middleware that assumed ``http`` would break the startup handshake, and the
+failure would surface as every request failing rather than as a middleware bug.
+That is the whole justification for a passthrough, and it justifies exactly one
+scope type — so the passthrough names it, rather than being written as "anything
+that is not ``http``".
+
+The difference is not cosmetic. Denying by omission means the allowlist is
+implicit, and the first ``websocket`` route anyone mounts inherits a path
+through all seven layers that is unauthenticated, unceilinged and unlogged —
+answered by the router rather than refused, because ``websocket`` is a type
+Starlette already serves. There is no live exposure today, which is precisely
+why it is closed now rather than after there is one.
+"""
+
 
 SCOPE_KEY: Final[str] = "creek_mcp.httpapi.context"
 """Where the record lives on the ASGI scope.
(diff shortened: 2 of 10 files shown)


Geoffe-Ga:
## Summary
This PR fixes two related holes in the `/v1` ASGI middleware stack: (#1122) an unhandled exception behind a `500` was discarded from the process entirely, leaving operators with no traceback to diagnose a fault; and (#1124) every middleware's `if scope["type"] != HTTP_SCOPE: await self.app(...)` passthrough silently forwarded *any* non-`http` scope — including `websocket` — past authentication, the ceiling gate, and the access log. The fix centralizes the passthrough decision in one `pass_through()` allowlist function (`context.py`) that relays only `lifespan` and raises `UnsupportedScopeError` for everything else, and adds a second, separately-routable logger (`creek_mcp.httpapi.error`) that records the fault's traceback server-side while leaving the caller-facing envelope unchanged.

## Strengths
- **Single source of truth for the allowlist.** Seven middlewares previously each hard-coded "anything not `http` passes through"; now they all defer to `pass_through()`, so the rule can't drift across copies (exactly the DRY principle in `CLAUDE.md` §1.2).
- **Two loggers, deliberately separated.** The access logger's five-safe-fields promise stays intact; the traceback (which can carry vault content via the exception's own message) goes to a distinct, independently-routable logger. This is a real security-conscious design choice, not just plumbing.
- **Test coverage matches the claimed design.** Parametrized tests cover all seven layers individually (`test_a_websocket_scope_never_reaches_the_app_below`, `test_a_lifespan_scope_still_passes_through`), plus a non-vacuity check that the parametrization tracks the pinned stack (`test_the_scope_tests_cover_the_whole_pinned_stack`) and an end-to-end assembled-app test that asserts nothing hits the wire and the access log stays silent for a refused websocket scope.
- **Docstrings are accurate and pull their weight.** `UnsupportedScopeError`'s docstring correctly explains why a raise (not a protocol-level refusal) is the only sound option for an arbitrary ASGI scope type — there's no generic "refuse" verb in ASGI the way there is for `http` (422/500) or `websocket` (`websocket.close`).
- **Response body is provably unchanged.** `test_a_handler_fault_is_logged_with_its_traceback` asserts the sentinel is present in `caplog.text` and explicitly absent from `response.text` — the exact contrast that matters for #1122.
- **`docs/api.md` was updated in the same PR**, not left to drift, with the fault-logging and ASGI-scope sections matching the implementation.

## Security Concerns
None found. 🟢 The change is a net security improvement: it closes a passthrough that would have let an unauthenticated, unceilinged, unlogged `websocket` connection reach the router the moment anyone mounted a `WebSocketRoute` (there's no live `ws://` surface today, so no active exposure — the PR closes this preemptively, which is the right call). The new traceback logger correctly avoids the access logger and is documented as needing separate operational handling since it can carry vault content.

## Problems
None blocking. I traced call sites for all seven `pass_through()` usages (`auth.py`, `boundary.py`, `access_log.py`, `ceiling.py`, and three in `limits.py`) and confirmed each only reaches `pass_through` after already checking `scope["type"] != HTTP_SCOPE`, so `UnsupportedScopeError` can never be raised for an `http` scope. The `context_started` re-raise path in `boundary.py` correctly avoids double-logging (comment explains the ASGI server will record the re-raised fault).

## Code Quality
- Non-blocking: `docs/api.md`'s "refused at the outermost middleware" line slightly undersells the actual defense-in-depth (each of the seven layers independently refuses via `pass_through`, not just the outermost one) — worth a wording tweak if there's a next pass, but not worth blocking on.
- Non-blocking: `CHANGELOG.md`'s `Unreleased` section doesn't mention #1122/#1124 (or, for that matter, #1074 that built this stack), so this isn't a new gap introduced by this PR — this subsystem's history simply isn't tracked there. Consider whether that's intentional.

## CLAUDE.md Adherence
✅ Docstrings are Google-style with Args/Returns/Raises on every new public symbol (`UnsupportedScopeError`, `pass_through`, `_error_records`, `_ScopeSpy`, etc.) — consistent with the ≥95% interrogate requirement.
✅ No forbidden shortcuts observed: no `type: ignore`, no bare `# noqa`, no commented-out tests.
✅ TDD discipline is unusually well-documented in the PR body — three tests are explicitly called out as passing against *unfixed* code by design (the constant pin, the seven lifespan-passthrough regression guards, and the stack-identity twin), with the ten fix-carrying tests distinguished in the RED proof.
⚠️ I could not execute `./scripts/check-all.sh`, `./scripts/test.sh`, or `mypy`/`ruff` in this review sandbox (no `uv`/`pytest` available, no network access to install), so I could not independently confirm the "Gate: 12/12" claim or coverage/lint numbers — my review is based on careful static reading of every changed line, not a green CI run. This is a limitation of the review environment, not a flag against the PR.

## Testing
Strong. New tests cover: the error logger's stable/distinct name, a fault's traceback landing on the error logger with the correct `request_id` and *not* leaking into the response, the fault *not* reaching the access logger, per-layer websocket refusal (parametrized across all seven middlewares), per-layer lifespan passthrough (parametrized, the same seven), the parametrization's non-vacuity against the pinned stack, and an end-to-end assembled-app refusal test. Insufficient testing would be blocking per the review rubric — this is the opposite of that.

## Documentation
`docs/api.md` gained "Fault logging" and "ASGI scopes" sections that accurately describe the new behavior, including the security rationale (why the two loggers are separate, why the allowlist is a positive list rather than a negative one). Module and class docstrings throughout the diff explain the *why*, not just the *what*, consistent with the repo's established style in this file.



## Verdict: LGTM

All seven middlewares now share one allowlist function (pass_through) instead of seven copies of the same passthrough check, closing a real unauthenticated/unceilinged/unlogged websocket hole (#1124), and the 500 boundary now logs the traceback server-side without changing the byte-for-byte response body (#1122). The change is narrowly scoped, every new symbol is exercised by parametrized per-layer tests plus an end-to-end assembled-app test, docstrings are complete and accurate against the code, and I found no correctness, security, or CLAUDE.md-adherence issues on close reading of every changed file.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5228284601 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

Issue #1126, opened 2026-08-01
perf(mcp): stop rebuilding MCPAuditLog per request on the /v1 capabilities path


Geoffe-Ga:
`creek_mcp/tools/handshake.py:106` constructs a fresh `MCPAuditLog(vault_path)` per call, so the
hash-chain cache is always cold and `_compute_prev_hash` falls through to a full
`path.read_text()` of `mcp.jsonl` on every authenticated `GET /v1/capabilities`.

PR #1117 moved this off the event loop (`run_in_threadpool`), so it no longer blocks the server —
but it is still O(file size) per request against a file that grows without bound, on the first
endpoint every client calls.

Fix direction: a long-lived per-vault `MCPAuditLog` (e.g. on `app.state`) so the cache is warm.
Raises lifecycle/staleness questions — hence a separate change rather than a widening of #1117.

---
Follow-up from #1117 (review of issue #1074) — https://github.com/Geoffe-Ga/Creek-Vault/pull/1117

PR #1239, opened 2026-08-08
fix(mcp): stop creek.handshake scaffolding the marker whose absence it reports

### LINKED ISSUE #1108: fix(mcp): creek.handshake scaffolds the 00-Creek-Meta marker it reports missing, so the second handshake lies about vault availability
## Problem

`creek.handshake` reports `available: true` for a vault that was never
initialized, from the **second** call onwards.

`handshake_tool` (`creek-tools/creek_mcp/tools/handshake.py`) computes
`available` from the presence of `00-Creek-Meta`, and then audit-logs the call:

```python
available = vault_available(vault_path)
MCPAuditLog(vault_path).append(tool=TOOL_NAME, args={}, ...)
```

`MCPAuditLog.append` delegates to `creek.audit.AuditLog.append`, which does
`self.path.parent.mkdir(parents=True, exist_ok=True)` — and the audit log lives
at `00-Creek-Meta/audit/mcp.jsonl`. So the append **creates the very marker
directory whose absence it just reported**.

The ordering inside one call is correct and deliberate (the docstring says so:
"`available` is computed *before* the audit append so a fresh vault still
reports `False`"). The bug is across calls.

Verified at `879d961`:

```python
>>> from creek_mcp.tools.handshake import handshake_tool
>>> missing = Path(tmp) / "nope" / "deeper"
>>> handshake_tool(vault_path=missing, capabilities=[], server_name="x")["available"]
False
>>> missing.exists()
True          # <-- the tool just scaffolded it
```

A second `handshake_tool(vault_path=missing, ...)` returns `available: True`.

## Why it matters

The whole point of the handshake is that a client can distinguish "vault present
and usable" from "reachable, but no vault scaffolded" — they are different facts
an operator acts on differently. This makes the distinction survive exactly one
call per process lifetime, and then silently invert.

The ratified `/v1` contract
(`docs/decisions/2026-07-31-adepthood-http-application-api.md`) elevates this to
a published contract state: `status: "uninitialized"` vs `status: "ok"`, with
`GET /v1/capabilities` explicitly documented as safe to poll. A readiness
endpoint that changes its answer *because it was polled* is not a readiness
endpoint.

It also means an unconfigured `creek-tools-mcp` (whose `vault_path` falls back
to `.` when no config file is found) scaffolds `00-Creek-Meta/audit/` into
whatever the current working directory happens to be, on the first handshake.

## Workaround already in place

#1074 routes around this on the HTTP side: `GET /v1/capabilities` probes with
the new side-effect-free `creek_mcp.tools.handshake.vault_available()` and does
**not** enter `handshake_tool` when the marker is absent. That fixes the
symptom for one caller. The MCP tool itself is still wrong, and the deliberate
consequence — a capabilities call against an absent vault is not audited at all
— is a compromise, not a fix.

## Requested change

Give the audit substrate a no-create mode, or have `handshake_tool` skip the
audit append when `vault_available(vault_path)` is `False`. Prefer the former:
the same `mkdir(parents=True, exist_ok=True)` sits under every other tool's
audit call, so any tool invoked against an absent vault root scaffolds it.

Decide explicitly what "audit a call against a vault that does not exist" should
mean. Dropping the entry loses a trail; creating the vault to store the trail
falsifies the state the trail is about. A third option — logging it to the
process log instead — may be the honest answer.

## Test idea

`test_handshake_does_not_scaffold_the_marker_it_reports_missing` — call
`handshake_tool` twice against a `tmp_path` with no `00-Creek-Meta`; assert both
calls report `available: False`, and assert `not (vault / "00-Creek-Meta").exists()`
afterwards. Today the second assertion fails on the first call and the first
assertion fails on the second call.

The `/v1` equivalent already exists as
`test_capabilities_stays_uninitialized_across_repeated_calls` (#1074).

## References

- `creek-tools/creek_mcp/tools/handshake.py`
- `creek-tools/creek_mcp/audit.py` → `creek/audit/log.py` (the `mkdir`)
- Found while building #1074; routed around there, not fixed.
- Contract states: `docs/decisions/2026-07-31-adepthood-http-application-api.md`, "Capabilities — the states a handshake can report"

Geoffe-Ga:
Closes #1108.

The first handshake against an uninitialised vault returns `available=False` **while creating** `00-Creek-Meta/audit/mcp.jsonl` on the way out. The second then returns `available=True` — the tool falsifying its own report.

Reproduced verbatim before touching anything. The created tree is exactly `['00-Creek-Meta', '00-Creek-Meta/audit', '00-Creek-Meta/audit/mcp.jsonl']`. **The issue's premise is correct in full** — every file, symbol and mechanism it names is real and current, which is unusual for this repo.

## I did not follow the issue's stated preference

It says *"Prefer the former"* — a no-create mode on the audit substrate. I took the contained fork, and the reason is measured rather than asserted:

- That `mkdir` sits on **31** `MCPAuditLog.append` call sites plus **5** direct `AuditLog` constructions. I counted these rather than trusting the issue's looser "every other tool" phrasing. Making it no-create by default is a behaviour change for all 36.
- An **opt-in** flag collides with `_shared_audit_log`, whose `@lru_cache` is keyed on the log path **alone** (#1126). One path would need two `AuditLog` objects with different creation policies — or a wider cache key that reinstates precisely the cold-cache re-read that cache was added to remove.

The issue could not have known the second one. **The contained fork is the architecturally cleaner choice here, not merely the safer one.**

## The audit entry moves, it doesn't vanish

Dropping it trades one dishonesty for another: an operator polling an uninitialised vault would find no evidence they had ever called. The record goes to the process log instead — the issue's own suggested third option. The third new test pins it.

## Blast radius, as required

Exactly one behavioural change, confined to `creek.handshake` against a vault whose `00-Creek-Meta` is absent. That call no longer appends and no longer creates any directory; it emits one `logger.info` naming tool, path, consumer and ceiling. **Handshakes against a real vault are byte-for-byte unchanged. No other tool's audit path is touched.** Full unit suite: 8364 passed, 5 skipped, 0 failed.

## Two residuals, named rather than buried

- **The trap is still armed for the other 30 call sites.** A tool pointed at a wrong or unconfigured path still scaffolds `00-Creek-Meta/audit/` there. That is a real, separate defect. What's fixed here is the *self-falsification*, unique to handshake because it is the only tool whose **return value is a claim about the marker's presence**.
- **`GET /v1/capabilities` against an absent vault still produces no record at all** — not an audit entry, not even the new process-log line — because `capabilities.py::_vault_is_usable` short-circuits before entering the tool. Removing that short-circuit is now *safe* (the tool no longer scaffolds) and would give the HTTP path the trail too, but it restructures #1074's control flow. Follow-up.

## Stale prose corrected

Three files asserted as present-tense fact that `handshake_tool` creates the marker — `capabilities.py` docstrings, `test_v1_api_capabilities.py` docstrings, and `docs/mcp.md`'s flat claim that *"Every tool invocation appends one entry"*. This fix made all of it false. Leaving it would plant the exact class of stale lie this issue exists to remove. **No test logic was changed** — `test_repeated_calls_never_conjure_the_creek_marker` still passes unmodified.

One note on process: the agent corrected its **own** comment counts ("thirty call sites plus seven direct users" → the measured 31 and 5) before standing behind them, on the grounds that an unverified number inside a comment justifying a design decision is the same class of defect as the bug being fixed. That's the right instinct.

Gate: **12/12**.

### DIFF
--- a/creek-tools/creek_mcp/httpapi/capabilities.py
+++ b/creek-tools/creek_mcp/httpapi/capabilities.py
@@ -15,14 +15,18 @@ is the endpoint an operator uses to *find out* the vault is broken; answering
 :class:`pydantic.ValidationError` subclasses) and the YAML parse error — rather
 than a bare ``except Exception`` that would also swallow a bug in this module.
 
-**It probes rather than calling the tool against an absent vault.**
-:meth:`creek_mcp.audit.MCPAuditLog.append` does ``mkdir(parents=True,
-exist_ok=True)``, so entering :func:`creek_mcp.tools.handshake.handshake_tool`
-against a directory with no ``00-Creek-Meta`` *creates* ``00-Creek-Meta/audit/``
-as a side effect — and the very next call would then find the marker and report
-``available: true`` for a vault nobody ever initialised. So readiness is decided
-by :func:`creek_mcp.tools.handshake.vault_available`, which has no side effect,
-and the tool is entered only once a vault genuinely exists.
+**It probes rather than calling the tool against an absent vault.** Readiness is
+decided by :func:`creek_mcp.tools.handshake.vault_available`, which has no side
+effect, and the tool is entered only once a vault genuinely exists. That began
+as a workaround: the tool's audit append reached ``mkdir(parents=True,
+exist_ok=True)``, so entering it against a directory with no ``00-Creek-Meta``
+*created* ``00-Creek-Meta/audit/``, and the next call found the marker and
+reported ``available: true`` for a vault nobody ever initialised. #1108 fixed
+that at the source — the tool now records an absent-vault call to the process
+log rather than to a log it would have to scaffold a vault to write. The probe
+stays as the cheaper path and as defence in depth for a contract state that must
+survive polling, but it is no longer the only thing standing between this
+endpoint and a lie.
 
 **Advertised equals implemented.** The ``capabilities`` list is
 :data:`creek_mcp.api.routes.IMPLEMENTED_CAPABILITIES`, not ``set(Capability)``.
@@ -122,8 +126,9 @@ def _negotiate(vault: Path, context: RequestContext) -> bool:
     """Enter the handshake tool for a vault already known to exist.
 
     Reached only after :func:`~creek_mcp.tools.handshake.vault_available` has
-    said yes, so the tool's audit append cannot conjure the marker whose
-    absence would have been reported.
+    said yes. Since #1108 the tool is safe to enter against an absent vault on
+    its own, so this ordering is no longer load-bearing for correctness; it is
+    kept because there is nothing to negotiate about a vault that is not there.
 
     Args:
         vault: The vault root, already probed.
--- a/creek-tools/creek_mcp/tools/handshake.py
+++ b/creek-tools/creek_mcp/tools/handshake.py
@@ -14,6 +14,7 @@ from the tools actually registered.
 
 from __future__ import annotations
 
+import logging
 from pathlib import Path
 from typing import Any
 
@@ -21,6 +22,8 @@ from creek.models import PrivacyTier
 from creek_mcp.audit import MCPAuditLog
 from creek_mcp.contract import CONTRACT_VERSION, ONTOLOGY_VERSION
 from creek_mcp.tier_ceiling import TierCeiling
+
+logger = logging.getLogger(__name__)
 
 TOOL_NAME = "creek.handshake"
 
@@ -39,15 +42,10 @@ TRANSPORT = "stdio"
 def vault_available(vault_path: Path) -> bool:
     """Return whether a scaffolded vault is readable at *vault_path*.
 
-    The whole readiness probe, in one predicate with no side effect. That
-    matters more than it looks: :func:`handshake_tool` also appends to the
-    audit log, and :meth:`creek_mcp.audit.MCPAuditLog.append` reaches a
-    ``mkdir(parents=True, exist_ok=True)`` — so calling the *tool* against an
-    absent vault **creates** ``00-Creek-Meta/audit/`` and makes the next probe
-    answer ``True`` for a vault nobody ever initialised. Callers that only need
-    the answer (notably ``GET /v1/capabilities``, which must keep the ADR's
-    "present" and "uninitialised" states distinct across repeated calls) must
-    use this and not reach for the tool.
+    The whole readiness probe, in one predicate with no side effect, so callers
+    that only need the answer — notably ``GET /v1/capabilities``, which must keep
+    the ADR's "present" and "uninitialised" states distinct across repeated
+    calls — can ask without entering the tool.
 
     Args:
         vault_path: Vault root to probe.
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
Fixes #1108: `creek.handshake` previously appended to `00-Creek-Meta/audit/mcp.jsonl` unconditionally, and `AuditLog.append`'s `mkdir(parents=True, exist_ok=True)` scaffolded the very `00-Creek-Meta` marker whose absence the call had just reported — so a second handshake against an uninitialised vault falsely returned `available=True`. The fix computes `available` first (already true pre-fix) and, when the vault is absent, records the call to the process log instead of the audit substrate rather than entering `MCPAuditLog(...).append(...)` at all; when the vault is present, behavior is byte-for-byte unchanged.

## Strengths
- Root-caused correctly and reproduced the bug (`>>> missing.exists() == True` after one call) before touching code, matching the issue's own repro.
- Deliberately diverges from the issue's stated preference ("no-create mode on the audit substrate") and gives a concrete, checkable reason: 31 `MCPAuditLog.append` call sites + 5 direct `AuditLog` constructions would need a behavior change, and an opt-in flag collides with `_shared_audit_log`'s `lru_cache` being keyed on path alone (#1126) — verified independently and the counts are accurate (`creek_mcp/audit.py:86-102`).
- Chooses the issue's own third option (log to the process log) rather than silently dropping the entry, so a poll against an uninitialised vault still leaves a trace.
- Updates the three places that had stale present-tense claims about the old behavior (`capabilities.py` docstrings, `test_v1_api_capabilities.py` docstring, `docs/mcp.md`), and is explicit that no test *logic* changed in the file whose docstring it touched.
- Names two residuals rather than burying them: the `mkdir` trap is still armed for the other 30 tools' audit paths, and `GET /v1/capabilities` still records nothing at all for an absent vault because `_vault_is_usable` short-circuits before `_negotiate` — confirmed by reading `capabilities.py:181-184`, unchanged by this diff.
- Three new tests directly target the fixed behavior (double-call honesty, absent-root-not-created, process-log fallback) plus a regression test (`test_handshake_still_audits_when_the_vault_is_present`) pinning that the guard doesn't over-fire on the present-vault path.

## Security Concerns
🟢 LOW — `_record_call`'s `logger.info` interpolates `consumer` (an authenticated `client_id`, per `context.py`) into a free-text log line via `%s`. Unlike the JSON audit entry, a plain `logging` call does no control-character escaping, so a `client_id` containing a newline could forge an extra log line. Low impact (authenticated caller, local process log only, same trust level the field already had in the audit JSONL), not blocking.

## Problems
None found blocking merge. The fix is behaviorally scoped exactly as claimed — the added tests would fail against the pre-fix code and pass against this diff, and the present-vault path is provably untouched (`test_handshake_still_audits_when_the_vault_is_present` still expects exactly 2 audit entries for 2 calls).

## Code Quality
- `_record_call`'s docstring is unusually long (relative to the function body) but earns its length here — it documents a non-obvious design decision (why not fix it in `AuditLog`, why not opt-in) that would otherwise require re-deriving from the issue thread.
- Minor: the log-injection point above could be closed cheaply (e.g. `consumer.replace("\n", " ")` or structured `extra=`) if this is ever a concern in the deployment context — not required now.

## CLAUDE.md Adherence
✅ TDD workflow followed per PR description (tests written to reproduce the bug first). ✅ Root-cause fix, no shortcuts, no suppressed checks. ✅ Docstrings updated everywhere the behavior they described changed — matches §1.2 (single source of truth) and the "no stale prose" ethic. ✅ Blast radius explicitly minimized and stated. CI: `Integration & E2E` and `CrawDad Quality` are green at review time; the three `Code Quality & Testing` matrix jobs were still `IN_PROGRESS` — worth confirming they finish green before merge per the Stay Green gate, but nothing indicates they won't (PR claims 8364 passed / 5 skipped / 0 failed locally, gate 12/12).

## Testing
Sufficient and well-targeted. The double-call pattern is the only way to actually observe this class of bug (a single call is honest either way, pre- or post-fix), and the PR correctly uses it. The absent-root test (`tmp_path / "nope" / "deeper"`) separately covers the case where `vault_path` itself doesn't exist, distinct from an existing directory missing only the marker. The process-log test uses `caplog` at the right logger name and level.

## Documentation
`docs/mcp.md` gets a clear, scoped exception note explaining the one case that deviates from "every tool invocation appends one entry." Docstrings in `capabilities.py` and `handshake.py` are updated to reflect that the tool is now safe to enter against an absent vault, while correctly noting the probe-before-enter ordering is retained for other reasons (nothing to negotiate for a vault that doesn't exist).



## Verdict: LGTM

The fix is correctly scoped and verified: it eliminates the self-falsifying audit append for absent vaults while leaving the present-vault path byte-for-byte unchanged, tests directly pin the previously-failing double-call scenario, docs/docstrings are updated everywhere the old behavior was asserted, and the one deviation from the issue's stated preference is backed by a checked, concrete rationale. The single note (low-severity log-injection via an authenticated consumer id) is a nit, not a blocker.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5228313690 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

PR #1243, opened 2026-08-08
fix(purge): erase the voice artifacts that quote a purged fragment's body

### LINKED ISSUE #1211: fix(purge): right-to-be-forgotten misses 07-Voice derived artifacts — a purged fragment's body survives in its voice profile

Geoffe-Ga:
Closes #1211.

A purged fragment's sentences survived **verbatim** in derived `07-Voice` notes. Same family as #845: an erasure that leaves a derived copy behind.

**The issue understates it by one file.** It lists three survivors; I reproduced **four** at unmodified HEAD. `07-Voice/Lexicon/Metaphors/<domain>.md` also carries the body, because `MetaphorInventory.usages` are `LexiconContext` values rendered through the same `_render_context_line` as the glossary.

## The trap is real, and worse than stated

The issue says frontmatter-`id` matching is unreliable because the scrub rewrites `id:` to `[purged]`. In fact **the scrub destroys the only recorded fragment→lexicon link in the same run**:

```
before:  - [[frag-a]] — The dharma teaches that ZQXJV-SENTINEL-1211 …
after:   - [[[purged]]] — The dharma teaches that ZQXJV-SENTINEL-1211 …
```

So **any sweep placed after `_scrub_references` is blind to exactly the notes that most need it.** The sweep is wired strictly *before* it, in both `purge_fragment` and `_purge_single`.

## `Rhetorical-Patterns/` is not a leak — and is deliberately left alone

The acceptance criteria name it as a target. It isn't one: `_format_rhetorical_moves` emits exactly three integer-count bullets (self-deprecation, paradox, callback) — no body text, no sentences, no ids — and I verified the sentinel is absent from it *before* any purge. Sweeping it would destroy a regenerable note with no leak to justify it. Pinned by `test_rhetorical_pattern_notes_are_left_alone`.

## Design finding — one target has no recorded link, and I'm not hiding it

Three of four targets are reachable from a real recorded link: **Register-Samples** by filename stem (the one link the scrub cannot destroy, and the same key #879's own prune matches on), and both **lexicon** note types by the `[[<id>]]` wikilink.

The fourth is not. A register profile records `exemplars=tuple(e.body for e in selected)` — **bodies only, zero fragment ids** — so `07-Voice/<register>-profile.md` has no provenance to key on. That pass matches on the purged fragment's **own body text**. It is exact rather than heuristic (`_render_profile_body` renders each passage verbatim, so the body is a contiguous substring) and still reached from the id the caller holds (id → fragment file → body) — but it is **a content match standing in for a link the format never wrote down**. Recording exemplar ids in profile frontmatter would make it a real link. Documented in `_voice_profiles_quoting`'s docstring and in `docs/cleaning-and-purge.md`.

Also checked: #879's manifest is **not** usable as the key — it stores unsalted SHA-256 digests of ids in `_Summary.md` frontmatter. The filename stem is simpler and equally scrub-proof.

## Delete, not surgically edit

A derived note is a *function of the corpus*. Excising one passage leaves the purged fragment's statistical residue — its n-grams in `## Distinctive Phrases`, its contribution to every count — while the note goes on advertising an `exemplar_count` it no longer has. All four regenerate from `creek report --type voice` / `--type lexicon`, so deletion costs a re-run and buys a clean erasure.

## Deliberately not swept

- **`07-Voice/Drafts/`** — the operator's own essays. Their provenance references are already handled by `_scrub_references`, and a fragment purge has no mandate to delete user writing. **Consequence stated plainly: a draft quoting a purged body verbatim is still a residual leak.** Out of scope, and deleting user work is the wrong direction. Pinned by `test_operator_authored_drafts_are_scrubbed_not_deleted`.
- **`Register-Samples/<register>/_Summary.md`** — no body text (stats, ids, titles, all already scrubbed), and deleting it would destroy the #879 manifest that entitles the next prune to remove stale copies.

Tested with a real byte sentinel and a bytes-based leak sweep over the whole vault, asserting presence before the purge and absence after.

Gate: **12/12**.


Geoffe-Ga:
## Summary
This PR closes #1211 by sweeping the `07-Voice/` derived artifacts (Register-Samples copy, `<register>-profile.md`, and Lexicon glossary/metaphor notes) that quote a purged fragment's body verbatim. It correctly identifies that the existing reference scrub is blind to body-copied content, orders the sweep before `_scrub_references` (since the scrub destroys the only recorded fragment→lexicon link), and deliberately leaves `Rhetorical-Patterns/` (counts only, no leak) and `Drafts/` (operator's own writing) untouched. The engineering is careful about symlink escapes, path traversal in fragment ids, and unreadable files, and ships with 16 new tests.

## Strengths
- Correct ordering fix: sweeping voice artifacts *before* `_scrub_references` runs (in both `purge_fragment` and `_purge_single`) is the crux of the bug — after the scrub, the `[[id]]` wikilink the lexicon pass depends on would already be `[[[purged]]]`. This is exactly the kind of subtle sequencing bug that's easy to get wrong.
- Containment discipline: `_contained_voice_artifact` resolves the candidate path and checks `is_relative_to` the resolved `07-Voice` root *before* incrementing the counter or unlinking, so a dry run never previews a deletion the real run would refuse, and a symlinked register folder pointing outside the vault is refused and logged rather than followed (`test_sweep_refuses_a_register_dir_symlinked_out_of_the_vault`).
- Deletes the *resolved* path rather than just unlinking a symlink alias (`test_a_symlinked_sample_copy_loses_its_target_not_just_the_link`), avoiding the classic "erasure that only removes the pointer" mistake.
- `fragment_id` is checked for path separators/NUL before being joined into a filesystem path (`_voice_sample_copies`), consistent with the existing `_is_safe_sample_stem` hazard analysis in `creek/generate/voice.py`.
- Correctly identifies and documents (in both the engine docstring and `docs/cleaning-and-purge.md`) that the register-profile pass has no real recorded link (bodies only, no ids) and is therefore a content-match standing in for a missing link — an honest design-limitation callout rather than papering over it.
- Test suite is unusually thorough for the surface area: dry-run parity, audit-log persistence, multi-fragment interaction with shared derived notes, unreadable-file resilience (a directory named `*.md`), no-op when `07-Voice/` doesn't exist, and an id containing a path separator producing zero matches.
- Verified independently against the referenced source: `_render_context_line` in `creek/generate/lexicon.py` really does render `- [[{fragment_id}]] — {sentence}`, and `VoiceProfileGenerator._render_profile_body` really does emit each exemplar body verbatim under `### Sample Passages` — the PR's factual claims about the leak mechanism check out.

## Security Concerns
🟢 LOW — No new attack surface beyond what's already mitigated. Path-traversal/symlink-escape vectors into `07-Voice/` are defended with a resolve-then-contain check and covered by tests. The one thing worth double-checking in a future pass: `_voice_profiles_quoting` glob-matches `*-profile.md` at the top of `07-Voice/` without per-register scoping, so a coincidental substring collision between two fragments' bodies could in theory cause an unrelated register's profile to be swept — an accepted, documented trade-off given the lack of a real link, not a new hole.

## Problems
🔴 **Body-content sweep silently degrades for non-UTF-8 fragment files.** `_purge_voice_artifacts(fragment_id, post.content, result)` is called from both `purge_fragment` (via `_find_fragment_by_id`) and `_purge_single`, both of which source `post` from `_load_frontmatter_for_match`. That loader's own docstring (`creek/purge/engine.py` ~line 1306-1332) states: *"a purge match is a function of frontmatter metadata only ... and never of the body. Replacing bad body bytes therefore cannot change the outcome."* That invariant is exactly what `_voice_profiles_quoting` breaks: it matches derived-artifact bytes against the fragment's **body**, sourced from a loader that, for a non-UTF-8 fragment file, substitutes invalid bytes with U+FFFD before handing back `post.content` (see `_load_frontmatter_lossily`). If the body was mangled that way, the substring search against the (uncorrupted, verbatim-rendered) `<register>-profile.md` will not match, and the sweep silently does nothing for that artifact — no exception, no warning specific to this failure mode, just an incomplete erasure reported as complete. Given the whole point of this PR is closing exactly this class of "erasure reports success while a derived copy survives" bug (see #845), and given this codebase already treats non-UTF-8 fragment bodies as a real, tracked hazard (issue #910, with a dedicated `_load_frontmatter_lossily` fallback), this deserves a fix or at least an explicit, tested acknowledgment before merge — not silent inheritance of an invariant violation. Suggested fix: read the fragment's raw bytes directly (or re-open with strict UTF-8 and treat undecodable bodies as "no body to match, skip the content-keyed pass with a WARNING") rather than relying on the lossy `post.content`.

## Code Quality
- Minor completeness note, non-blocking: when purging one fragment causes an entire shared derived note (e.g., `analytical-profile.md`, `glossary.md`) to be deleted because it also quoted the purged fragment, any *other* fragment's legitimately-retained content in that note is collateral damage until the next `creek report --type voice`/`--type lexicon` run. This is the right trade-off for RTBF (delete-not-edit, documented clearly), but neither the CLI output nor the docs proactively suggest re-running the report generators after a purge that reports `voice_artifacts_removed > 0`. A one-line hint in `_render_purge_result` or the docs would close the loop for operators who don't already know to regenerate.
- The `_Summary.md` exemption is justified by "no body text," which is true in the normal case, but `creek/generate/voice.py`'s own docstring for `_is_safe_sample_stem` (~line 260-264) notes a known crash-window race where a fragment body *can* end up in `_Summary.md` (tracked as part of #879's territory). Not a regression from this PR and out of scope, but worth a one-line cross-reference in the new docstring/docs for future readers who might otherwise assume the "no body text" claim is unconditional.

## CLAUDE.md Adherence
✅ TDD apparent from test breadth (though I could not execute `./scripts/check-all.sh` / `./scripts/test.sh` in this review sandbox to confirm the actual coverage/lint numbers — could not verify Gate 2 mechanically, only by inspection).
✅ Docstrings are exhaustive Google-style with Args/Returns on every new method, well above the 95% interrogate bar by inspection.
✅ Type hints present throughout; `Iterator` correctly added under `TYPE_CHECKING`.
✅ No forbidden shortcuts (no `noqa`/`type: ignore`/commented-out tests observed in the diff).
✅ Docs (`docs/cleaning-and-purge.md`) updated in the same PR to describe the new `voice_artifacts_removed` field and its scope/boundaries — matches the DRY/single-source-of-truth principle.
❓ Commit message follows Conventional Commits (`fix(purge): ...`) and references the closed issue — consistent with repo convention.

## Testing
Strong, deliberately adversarial test design (byte-level sweep assertions rather than text-glob, symlink escape, unreadable-file resilience, dry-run parity, audit persistence). The one gap is the non-UTF-8-body scenario described above under Problems — there's no test exercising `_purge_voice_artifacts`/`_voice_profiles_quoting` with a fragment file containing invalid UTF-8 bytes, which is precisely the scenario where the new content-keyed pass can silently under-deliver. Given `tests/` elsewhere in this codebase already covers the UTF-8-fallback path for other purge operations (referenced via #910), this is a real, addressable coverage gap rather than a hypothetical one.

## Documentation
`docs/cleaning-and-purge.md` is updated thoroughly and accurately (verified the described rendering behavior against `creek/generate/lexicon.py` and `creek/generate/voice.py` directly) — it clearly states what is swept, what is deliberately not swept and why, and names the one acknowledged design gap (content-keyed profile matching). Recommend adding a short note about the non-UTF-8-body degradation once addressed, and optionally the "re-run `creek report`" operator hint mentioned above.



## Verdict: CHANGES_REQUESTED

The core sweep logic, containment/symlink handling, and boundary decisions (Rhetorical-Patterns, Drafts) are excellent and thoroughly tested, but there is a real correctness gap: the content-keyed profile sweep (`_voice_profiles_quoting`) is fed `post.content` from `_load_frontmatter_for_match`, a loader explicitly documented elsewhere in this same file as producing a lossily-decoded body (U+FFFD substitution) for non-UTF-8 fragment files, under the stated invariant that "a purge match is a function of frontmatter metadata only ... and never of the body." This PR's new pass is the first to violate that invariant by matching on body content — so for a fragment with invalid UTF-8 bytes in its body, the substring match against the verbatim-rendered profile can silently fail, leaving the purged sentence on disk in `<register>-profile.md` with no error or warning: exactly the class of leak this PR exists to close. It's untested and undocumented.

Geoffe-Ga:
**CI**: 5/8 Green
**Action**: pull comment 5228579825 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**Action**: pull comment 5228579825 to see in-depth feedback and continue iterating

Geoffe-Ga:
Fixed in 3450858. This was the sharpest finding of the session, and it was aimed at exactly what the PR claims to fix — thank you.

You identified the invariant violation precisely: `_load_frontmatter_for_match`'s own docstring says a purge match is a function of frontmatter metadata **only** and never of the body, *because* `_load_frontmatter_lossily` substitutes U+FFFD for invalid bytes (#910). `_voice_profiles_quoting` broke that by matching on the body, so for a non-UTF-8 fragment the mangled text failed to substring-match the uncorrupted, verbatim-rendered profile — and the sweep did nothing, silently. #845's bug class reappearing inside the code written to close it.

**Approach: strict decode, and refuse rather than skip.** `_voice_match_body` re-opens the fragment with `open(encoding="utf-8")` (strict) before `frontmatter.load`, so the body fed to the content-keyed pass is never lossy. On `UnicodeDecodeError` it does three things instead of returning quietly:

1. records the id in a new `PurgeResult.voice_body_undecodable`
2. logs a WARNING naming **the fragment id only** — no path, no body, per this subsystem's existing rule that warnings name ids and constants, never the victim's content
3. `_run_audited` sees the non-empty list on a normal return and writes `status="partial"` with `failure_reason="UnicodeDecodeError"`, following the convention the exception path already used

I took strict-decode over raw-bytes matching because the profile renders decoded text, so a bytes comparison would need its own encoding assumption — and a wrong one there fails in the same silent direction.

**The red proof states the blocker in one line.** With `engine.py` stashed:

```
test_an_undecodable_body_downgrades_the_audit_outcome_to_partial
  AssertionError: assert 'complete' == 'partial'
```

Plus `AttributeError` on the new field and a warning that was never emitted — 4 failed, and 6 failed across the whole file including the CLI pair.

The non-UTF-8 fragment is **real bytes on disk**: `frag_file.write_bytes(frag_file.read_bytes() + b"\xff\xfe")`, appended past the closing `---` so frontmatter still parses and the fragment still matches the purge criteria. Only the body is undecodable.

**One test passes under stash by design and I am not counting it as a red proof:** `test_the_link_keyed_passes_still_erase_an_undecodable_fragment`. It is an invariant pin — with the body corrupted, `purge_fragment` still deletes the register sample by filename stem and both lexicon notes by the `[[frag-a]]` wikilink (`voice_artifacts_removed == 3`), because neither id-keyed pass reads the body. The byte sweep then leaves exactly one survivor, the profile — which is precisely what the new field, the warning, the CLI line and the partial status all report. That correspondence is the point: the operator is told exactly what was missed.

**Both non-blocking notes are in.**

*Re-run hint* — `_render_voice_purge_notes` (extracted to keep `_render_purge_result` complexity flat) prints, whenever `voice_artifacts_removed > 0`: *"Those derived notes are shared: re-run `creek report --type voice` and `creek report --type lexicon` to regenerate them for the fragments that remain."* Pinned by a test asserting both commands appear in real CLI output, and mirrored in `docs/cleaning-and-purge.md` step 7.

*`_Summary.md` cross-reference* — you're right that "no body text" is not unconditional. `_voice_sample_copies`'s docstring and the docs now cite `_is_safe_sample_stem`'s documented crash window (#879 territory) and state explicitly that closing it is out of scope. **The sweep was not widened** — deleting the manifest would destroy what entitles the next prune to remove stale copies.

Gate: **12/12**, 8404 passed, 94.23%. Branch also synced to main.

Geoffe-Ga:
## Summary
Closes #1211: `PurgeEngine` now sweeps four `07-Voice/` derived artifacts (`Register-Samples/<register>/<id>.md`, `<register>-profile.md`, `Lexicon/glossary.md`, `Lexicon/Metaphors/<domain>.md`) that quote a purged fragment's body verbatim, running the sweep *before* the reference scrub so it isn't blinded by the `id: [purged]` rewrite. A second commit fixes a real bug found in review of this same PR: the content-keyed profile pass was matching against a lossily-decoded body (U+FFFD substitutions per #910) and silently failing to erase non-UTF-8 fragments while reporting success; it now re-reads strictly, records the fragment id on a new `voice_body_undecodable` field, and downgrades the audit outcome to `status="partial"`.

## Strengths
- Correct dependency ordering: both `purge_fragment` and `_purge_single` call `_purge_voice_artifacts` strictly before `_scrub_references`, which is the one place the fix could easily have been subtly wrong (`engine.py:365`, `engine.py:726`).
- Symlink/path-traversal hardening is thorough: `_contained_voice_artifact` resolves and re-checks containment before every delete (defeating a register folder symlinked out of the vault), deletion targets the *resolved* path so a symlinked alias doesn't just lose its link, and `_voice_sample_copies` rejects fragment ids containing `/`, `\`, or `\x00` before ever joining a path.
- Deliberately scoped: `Rhetorical-Patterns/` (counts only, no leak), `Drafts/` (operator's own writing, out of mandate), and `Register-Samples/_Summary.md` (the #879 manifest) are all explicitly left alone, each pinned by its own test.
- Logging discipline: every warning/log line names only fragment ids or exception type names, never bodies or paths that would defeat the erasure being reported on (`_read_bytes_for_match`, `_voice_match_body`).
- The `voice_body_undecodable` → `status="partial"` audit downgrade closes the exact "erasure silently incomplete" bug class (#845) this PR exists to fix, and does so without corrupting the intent/outcome audit pair.
- Test suite (`test_purge_voice_artifacts.py`, 25 tests) is unusually complete: happy path, dry-run parity, dedup across passes, symlink escape, path-separator-in-id, unreadable-artifact-as-directory, undecodable body, CLI messaging for both the normal and shortfall cases.
- Docs (`docs/cleaning-and-purge.md`) are updated in lockstep with the implementation and accurately describe the three deliberate exclusions and the one open design gap (profile pass has no recorded fragment→content link, only a content match).

## Security Concerns
🟢 LOW — No new vulnerabilities. The symlink-resolution + containment check in `_contained_voice_artifact` is a genuine hardening (prevents a maliciously or accidentally symlinked `07-Voice` subpath from being used to delete a file outside the vault), and it is exercised by `test_sweep_refuses_a_register_dir_symlinked_out_of_the_vault`.

## Problems
None blocking. One substantive gap worth addressing before or shortly after merge:

- **MCP tool responses don't surface the new partial-erasure signal.** `creek_mcp/tools/purge.py::_result_payload` hardcodes `"status": "ok"` and only forwards a fixed subset of `PurgeResult` fields (`fragments_affected`, `deleted_files`, `wikilinks_removed`, …). Neither `voice_artifacts_removed` nor `voice_body_undecodable` is included, so a caller through the MCP surface (e.g. the CrawDad Discord bot) receives `"status": "ok"` for a purge that the engine's own `purge.jsonl` audit line just recorded as `status="partial"`. This is a pre-existing pattern (`journal_staged_removed`, `provenance_scrubbed`, and `intimate_stubs_removed` are similarly absent from that payload today), so it isn't a regression this PR introduced — but it directly undercuts this PR's own stated goal ("an erasure that fell short must say so where the operator is already looking"), and the MCP caller is exactly such a place. Worth a follow-up issue if not fixed here.

## Code Quality
- Minor: `_purge_voice_artifacts` re-globs and re-reads the (potentially large, shared) `Lexicon/` and top-level profile files independently for *every* purged fragment. For `purge_source`/`purge_daterange` runs over many fragments this is O(fragments × artifact-bytes) rather than a single pass; likely fine at expected vault scale, but worth a note if very large sources/date-ranges become common.
- `_VOICE_BODY_UNDECODABLE = "UnicodeDecodeError"` duplicates the exception type name as a string constant rather than deriving it from the exception class — a small, deliberate readability trade-off (matches `_run_audited`'s existing `type(exc).__name__` convention) and not worth changing.

## CLAUDE.md Adherence
✅ Conventional commit messages (`fix(purge): …`) across all three commits.
✅ Google-style docstrings throughout, including `Args`/`Returns` sections on every new method — should comfortably clear the ≥95% interrogate floor.
✅ New Pydantic fields (`voice_artifacts_removed`, `voice_body_undecodable`) are backward-compatible: defaulted and added to `_coerce_legacy_entry`/the migration marker so old `purge.jsonl` entries still parse.
✅ No `noqa`/`type: ignore` shortcuts introduced.
❓ Could not execute `./scripts/check-all.sh` in this review environment to independently verify the coverage/lint/mypy gates; the PR body claims "Gate: 12/12, 8404 passed, 94.23%" coverage, consistent with the ≥90% requirement, but that claim is unverified here.

## Testing
Strong. TDD is evident from the second commit's description ("RED, with engine.py stashed: `assert 'complete' == 'partial'`"). Coverage includes the happy path, both call sites (`purge_fragment` and `purge_source`/`_purge_single`), dry-run parity, the three deliberate non-sweep boundaries, and — notably — adversarial cases (symlink escape, path-separator fragment id, a directory shaped like a `.md` file) that a less careful test suite would skip. Not blocking.

## Documentation
Complete and accurate. `docs/cleaning-and-purge.md` section 7 walks through the mechanism, the re-run guidance, all three exclusions, and the one open design gap (content-keyed profile matching) in the same terms as the code docstrings — no drift observed between the two.



## Verdict: COMMENTS

No bugs, security issues, or missing tests block this merge — the sweep ordering, symlink/path-traversal containment, and the #910 encoding fix are all correct and well-tested. It stays out of LGTM because of one substantive gap worth the author's attention: the MCP purge tools (`purge_fragment_tool` et al., used by the CrawDad bot) unconditionally return `"status": "ok"` and never surface the new `voice_artifacts_removed`/`voice_body_undecodable` fields, so an automated caller can't see the exact "partial erasure" signal this PR was written to stop hiding — even though the underlying `purge.jsonl` audit trail does correctly record it.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5229102672 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5229102672 to see in-depth feedback and continue iterating

Geoffe-Ga:
Thank you — the MCP gap is a real one and I've filed it as **#1246 (P1)** rather than fold it in here.

Reasoning for deferring rather than fixing in this PR: forwarding the fields is additive and safe, but the part that actually closes your finding — making `status` stop saying `ok` for a partial erasure — is a **contract change**. `status` is currently binary (`ok`/`refused`), and neither value fits an incomplete erasure, which is the argument for a `partial` member. That needs a `CONTRACT_VERSION` bump and a bundle regeneration, and it touches the three pre-existing omissions (`journal_staged_removed`, `provenance_scrubbed`, `intimate_stubs_removed`) as much as the two this PR added. Doing it here would mean a contract change riding inside an RTBF fix.

The issue captures what I think is the more durable framing: **the payload's hand-picked field subset is the same drift mechanism that lost all five fields**, and is the same defect class as the hand-copied directory list #1025 removed. Deriving it from `PurgeResult` — or failing a test when a new field isn't forwarded — stops the next one going missing. I've put that as the first thing to look at.

Worth noting `voice_body_undecodable` is the sharpest of the five: it's a non-empty list *naming the fragments whose artifacts were not swept*, so a caller who can't see it cannot know the erasure was incomplete.

On the two Code Quality notes:

**Re-globbing per fragment** — agreed, and it is O(fragments × artifact-bytes) for `purge_source`/`purge_daterange`. I left it alone deliberately: hoisting the artifact read out of the per-fragment loop changes the sweep's shape, and I'd rather not restructure the erasure path in the PR that introduces it. Noted in #1246's orbit if large sources become common.

**`_VOICE_BODY_UNDECODABLE = "UnicodeDecodeError"`** — you read it exactly right; it mirrors `_run_audited`'s existing `type(exc).__name__` convention so the two audit lines carry the same token. Left as is.

PR #1244, opened 2026-08-08
fix(privacy): make the mining corpus read 10-Liminal tiers the way the state report does

### LINKED ISSUE #1079: Two tier readers disagree about the same 10-Liminal note (state report vs mining corpus)

Geoffe-Ga:
Closes #1079.

Two readers disagreed about the same note:

| Side | Reader | Missing key resolves to |
|---|---|---|
| State report (`state.py:526`) | `raw_privacy_tier` + `within_ceiling` | **INTIMATE** |
| Mining corpus (`mining.py:402`) | `tier_of(fragment)` — the model | UNCLASSIFIED |

The raw reader is correct, and `raw_privacy_tier`'s own docstring says why: a missing key must resolve to INTIMATE, and *"reading a missing key through the model would therefore fail open relative to what the file actually says."*

The fix is entirely on the mining side, calling the definition the state side already used. **No third reader, nothing averaged.**

## Correction 1 — the issue understates the exposure by one ceiling

It says the miner admitted the untiered note *"at ceiling=personal"*. It admitted it at **ceiling=open**. `filter_fragments_by_tier` is a **body** filter, not a rank cutoff — it excludes only INTIMATE and *summarises* PERSONAL. Measured at HEAD: at `ceiling=open` the miner returned NoKey, Open, Personal and Unclassified notes while the report returned only the Open one — and the untiered note's body carried its literal **title**, not just a 12-hex id.

## Correction 2 — the issue's central claim is wrong for one path

It asserts *"the stamp does not under-report. Verified end to end in the #1078 re-review."* **It does under-report for `10-Liminal/Compost`.** `StateReportGenerator._content_tier` folded the miner's liminal tiers in via `tier_of`, so an untiered Compost note stamped the artifact `unclassified` (rank 1) though the raw frontmatter says `intimate` (rank 2) — and `creek.state.read` at `ceiling=personal` **accepts an `unclassified` stamp**.

Proven RED: `assert 'unclassified' == 'intimate'`.

The HEAD docstring on `_content_tier` makes the same error — it reasons only about `ceiling=open` ("the only ceiling at which the section did not render") and never checks that the **note** required intimate. Both the docstring and the issue text are corrected in place.

## Correction 3 — a suggested adjacency is unnecessary

The issue says narrowing the mining side means state needs an admitted list for Compost too: *"a second 10-Liminal walk and a new `_TierIndex` field."* Not required. Materialising `raw_privacy_tier` onto the fragment at load makes the miner's entry self-describing, so `_content_tier` reads the correct tier with no second walk and no new field.

## Ratchet check — explicitly required, and it passes

The unification is a **strict narrowing at every ceiling**. Nothing became less restrictive for any note, so no operator decision is needed:

| Frontmatter | Was admitted from | Now needs |
|---|---|---|
| missing key | `open` | **`intimate`** |
| `unclassified` | `open` (summarised) | **`personal`** |
| `personal` | `open` (summarised) | **`personal`** |
| `intimate` / `open` | unchanged | unchanged |

And the stamp can only **rise**, since a raw read never answers below the model's.

## Also verified

`tests/test_mcp_report_tier_ceiling.py::test_raw_and_model_tier_readers_agree_on_every_fragment` exists at line 1270 and does exactly what the issue claims — the new module is its `10-Liminal` counterpart. The two readers already agree on every explicit value including the legacy `public` alias (both funnel through `PrivacyTier(...)`), so **the missing key was the only disagreement**.

## Refactor, not suppression

Adding the ceiling gate pushed `_load_liminal_fragments` to cyclomatic complexity 11, over the ≤10 gate. The per-file half was extracted to `_admitted_liminal_entry` rather than suppressed. Zero suppressions anywhere in the change.

## Deliberately not done

Persisting a derived `privacy_tier` on Eddy/Thread/Praxis with a vault migration — the issue marks it explicitly splittable, and #1078's render-time derivation is correct on day one with no migration. Belongs to the #969 follow-on.

Gate: **12/12**.

### DIFF
--- a/creek-tools/creek/generate/mining.py
+++ b/creek-tools/creek/generate/mining.py
@@ -40,6 +40,8 @@ from pydantic import ValidationError
 from creek.classify.privacy_filter import (
     PrivacyTierOverride,
     filter_fragments_by_tier,
+    raw_privacy_tier,
+    within_ceiling,
 )
 from creek.generate.compile_routing import (
     COMPILE_GAPS_RELPATH,
@@ -399,6 +401,47 @@ def _liminal_kind(md_file: Path, liminal_root: Path) -> str | None:
     return kind
 
 
+def _admitted_liminal_entry(
+    md_file: Path,
+    liminal_root: Path,
+    ceiling: PrivacyTierOverride,
+) -> tuple[Fragment, str, str] | None:
+    """Return one admitted ``(fragment, body, kind)`` entry, or ``None`` to skip.
+
+    The per-file half of :func:`_load_liminal_fragments` — see that function
+    for why the tier is read off the raw frontmatter (#1079). ``None`` covers
+    every reason a ``10-Liminal`` file contributes nothing: it is a
+    Synchronicities reflection note, it does not parse, it is above the
+    ceiling, or it is not a valid fragment.
+
+    The ceiling is checked *before* validation deliberately. An above-ceiling
+    note should not be parsed into a model at all, and ``_validate_fragment``
+    logs the path of anything it rejects.
+
+    Args:
+        md_file: The candidate note.
+        liminal_root: ``<vault>/10-Liminal``, used to derive *kind*.
+        ceiling: The admission ceiling, already defaulted.
+
+    Returns:
+        The entry, with :func:`~creek.classify.privacy_filter.raw_privacy_tier`
+        materialised onto the fragment, or ``None``.
+    """
+    kind = _liminal_kind(md_file, liminal_root)
+    if kind is None:
+        return None
+    post = _safe_post(md_file)
+    if post is None or not within_ceiling(post.metadata, ceiling):
+        return None
+    fragment = _validate_fragment(post, md_file)
+    if fragment is None:
+        return None
+    declared = fragment.model_copy(
+        update={"privacy_tier": raw_privacy_tier(post.metadata)},
+    )
+    return declared, post.content, kind
+
+
 def _load_liminal_fragments(
     liminal_root: Path,
     *,
@@ -410,21 +453,64 @@ def _load_liminal_fragments(
     ``"Unnamed"`` or ``"Compost"``. The Synchronicities subfolder is
     excluded because its records are reflection notes, not fragments.
 
-    ``privacy_override`` is forwarded to the shared tier filter.
+    **The tier is read off the raw frontmatter, not off the validated model**
+    (#1079). ``10-Liminal`` notes are the one corpus two tools admit
+    independently: ``## Liminal Watch`` gates the same physical file through
+    :func:`creek.generate.state._admitted_liminal_notes`, which reads
+    :func:`~creek.classify.privacy_filter.within_ceiling` on the raw
+    frontmatter, while this loader reached the same file through the model.
+    A note with *no* ``privacy_tier`` key at all therefore came out
+    ``intimate`` on one side (fail-closed, absent means unvouched-for) and
+    ``unclassified`` on the other (the model's default, which ranks with
+    ``personal`` per #876/#961) — so one ``10-Liminal/Unnamed`` note needed
+    ``ceiling=intimate`` for its stem and only ``ceiling=open`` for its
+    generated id, inside one rendered document. That is the "two tools that
+    disagree about the same file" bug class
+    :mod:`creek.classify.privacy_filter` names in its own module docstring.
+
+    The divergence is closed by tightening this side up to the report's
+    reader rather than by loosening the report's down to this one, so the
+    change is a strict narrowing at every ceiling:
+
+    * :func:`~creek.classify.privacy_filter.raw_privacy_tier` is materialised
+      onto the loaded fragment, which is what makes every downstream reader —
+      :func:`~creek.classify.privacy_filter.filter_fragments_by_tier` below,
+      :func:`~creek.classify.privacy_filter.tier_of` in
+      :meth:`creek.generate.state.StateReportGenerator._content_tier` — see
+      one tier for one file instead of each re-deriving its own. It is a
+      narrowing because a raw read never answers *below* the model's: a
+      missing key escalates ``unclassified`` to ``intimate``, and an
+      unrecognised value never reaches here at all (the model refuses to
+      validate it, so :func:`_validate_fragment` has already dropped the
+      note).
+    * ``within_ceiling`` is applied as a hard rank cutoff *in addition to*
+      the body-level filter below, because the two answer different
+      questions: the filter *summarises* a personal body as
+      ``[Personal-tier summary: <title>]`` rather than excluding the note, so
+      at ``ceiling=open`` a personal liminal **title** entered the corpus
+      while the report refused the same file outright.
+
+    ``tests/test_liminal_tier_reader_agreement.py`` pins the two call sites
+    against each other note by note, at every ceiling.
+
+    Args:
+        liminal_root: ``<vault>/10-Liminal``.
+        privacy_override: The admission ceiling. ``None`` means ``OPEN``,
+            matching what it already meant to the body-level filter.
+
+    Returns:
+        The admitted ``(fragment, body, kind)`` tuples, each fragment
+        carrying the tier its raw frontmatter declares.
     """
     if not liminal_root.exists():
         return []
-    collected: list[tuple[Fragment, str, str]] = []
-    for md_file in sorted(liminal_root.rglob("*.md")):
-        kind = _liminal_kind(md_file, liminal_root)
-        if kind is None:
-            continue
-        post = _safe_post(md_file)
-        if post is None:
-            continue
-        fragment = _validate_fragment(post, md_file)
-        if fragment is not None:
-            collected.append((fragment, post.content, kind))
+    ceiling = privacy_override or PrivacyTierOverride.OPEN
+    collected = [
+        entry
+        for md_file in sorted(liminal_root.rglob("*.md"))
+        if (entry := _admitted_liminal_entry(md_file, liminal_root, ceiling))
+        is not None
+    ]
     kind_by_id: dict[str, str] = {f.id: kind for f, _body, kind in collected}
     pairs = [(f, body) for f, body, _kind in collected]
     filtered = list(filter_fragments_by_tier(pairs, override=privacy_override))
--- a/creek-tools/creek/generate/state.py
+++ b/creek-tools/creek/generate/state.py
@@ -536,6 +536,13 @@ def _admitted_liminal_notes(
     serve both: the paradox note fails closed to ``intimate`` because nobody
     vouched for it, not because it was classified.
 
+    This is one of two readers admitting ``10-Liminal`` notes into one
+    document; the other is
+    ``creek.generate.mining._load_liminal_fragments``, which backs
+    ``## Suggested questions`` and calls the same two functions for the same
+    reason (#1079). ``tests/test_liminal_tier_reader_agreement.py`` asserts
+    the two agree note by note rather than leaving it to coincidence.
+
     Sorting is unchanged from FEAT-007 and is applied before admission so the
     displayed order does not shift with the ceiling. ``st_mtime`` is
     best-effort — CI containers and fresh clones flatten it to the checkout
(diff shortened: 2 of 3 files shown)


Geoffe-Ga:
## Summary
Fixes #1079: the mining corpus loader (`_load_liminal_fragments`) previously read a `10-Liminal` note's privacy tier off the *validated* `Fragment` model (defaulting a missing `privacy_tier` key to `unclassified`), while the state report's `## Liminal Watch` section read the same file's *raw* frontmatter (failing a missing key closed to `intimate`). The two readers could therefore admit the same physical file at different ceilings inside one rendered document. This PR makes the mining side call the same `raw_privacy_tier` / `within_ceiling` primitives the state side already used, adds a hard ceiling gate before validation, and materializes the raw tier onto the returned fragment so every downstream reader (including `StateReportGenerator._content_tier`) sees one consistent tier per file.

## Strengths
- Root-caused correctly: identified the *single* divergent primitive (`raw_privacy_tier` vs. model-default) rather than adding a third reader or "averaging" behavior.
- The fix is a genuine strict narrowing at every ceiling (verified below) — no note becomes *more* visible than before.
- `_admitted_liminal_entry` cleanly extracts the per-file logic to keep `_load_liminal_fragments` under the complexity gate, without suppressing the metric.
- Both docstrings that previously reasoned incorrectly about this exact bug (`mining._load_liminal_fragments`, `state._content_tier`) are corrected in place rather than left stale.
- New test module (`test_liminal_tier_reader_agreement.py`) is genuinely rigorous: it compares the two readers against *each other* (not a hard-coded oracle) across every declared tier and every ceiling, plus a positive control (untiered note must resolve to `INTIMATE` on both sides, not just agree) and a dedicated regression test for the `10-Liminal/Compost` under-report the PR description calls out as Correction 2.
- PR description is unusually precise and each claim I checked against the code (ratchet table, "Correction 1/2/3", the `public`→`open` alias, `voice_proxy_eligible` being a computed property so it can't drift from the tier override) held up.

## Security Concerns
🟢 None. This is a privacy-boundary *tightening* fix: I traced the ceiling-gate logic (`within_ceiling` → `tier_within_override` → `_TIER_RANK`/`_OVERRIDE_RANK`) and confirmed every previously-admitted note either stays admitted at the same or a stricter ceiling, matching the PR's own ratchet table (missing key: open→intimate; explicit `unclassified`/`personal`: open→personal; `intimate`/`open`: unchanged). No new information disclosure path was introduced.

## Problems
None found that block merge.

## Code Quality
- `_admitted_liminal_entry`'s `None`-return contract (four distinct skip reasons) is documented but callers can't distinguish which reason applied — acceptable here since none of the reasons need distinguishing by the caller, but worth a mental note if this function grows more skip conditions later.
- `ceiling = privacy_override or PrivacyTierOverride.OPEN` relies on all `PrivacyTierOverride` enum values being non-empty strings (truthy); this already holds and is consistent with the same pattern elsewhere (`tier_within_override`), so not a real risk, just worth knowing if a future enum member were ever added with an empty value.

## CLAUDE.md Adherence
✅ TDD workflow evident (new test module pins the exact regression). ✅ No `noqa`/`type: ignore`/suppressions added anywhere in the diff (verified via grep — the one pre-existing `noqa` in `state.py` is untouched, outside the diff). ✅ Complexity kept under the gate via extraction rather than suppression, per the PR's own claim and consistent with `max-quality-no-shortcuts`. ✅ Docstrings updated to match the new implementation (Google-style, Args/Returns present). Commit message follows Conventional Commits (`fix(privacy): ...`).

## Testing
✅ Sufficient and well-targeted. `test_liminal_tier_reader_agreement.py` covers: (1) both readers admit the same note set at every ceiling (OPEN/PERSONAL/INTIMATE/ALL), (2) both readers report the same tier for every commonly-admitted note, (3) the untiered note fails closed to `INTIMATE` on both sides specifically (not just "agrees"), (4) the untiered note is excluded by the miner below `ceiling=intimate` (the direction-of-fix assertion), and (5) an untiered `10-Liminal/Compost` note now stamps the written report artifact `intimate` rather than under-reporting `unclassified` — read back off the actual written file rather than an internal method, so it exercises the real read-gate contract. I could not execute the suite directly in this review environment (sandbox restricted `git fetch`/`gh api`/CI-status commands), so this is based on careful static tracing of the test logic against the implementation rather than a live pytest run — worth a final CI-green confirmation before merge if not already done (PR claims Gate 12/12).

## Documentation
✅ Complete. Both affected docstrings (`mining._load_liminal_fragments`, `state._content_tier`, plus `state._admitted_liminal_notes` and `state._mining_corpus`) were updated to accurately describe the new shared-reader behavior, cross-reference each other and the pinning test, and no longer contain the incorrect reasoning the PR description's "Correction 2" identifies in the prior text.



## Verdict: LGTM

The fix is a small, correctly-scoped narrowing (mining now reads 10-Liminal tiers via the same raw_privacy_tier/within_ceiling primitives the state report already used), every claim in the PR description checks out against the actual code (raw_privacy_tier, within_ceiling, tier_of, filter_fragments_by_tier, _content_tier, _mining_corpus), the new test module thoroughly pins reader agreement across all four ceilings plus a dedicated Compost under-report regression test, no lint/type suppressions were added, and both touched docstrings were corrected in place to match the new behavior. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5228591321 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Fixes #1079: the mining corpus (`creek/generate/mining.py`) previously read a `10-Liminal` note's privacy tier off the *validated* `Fragment` model (missing key → `unclassified`), while the state report (`creek/generate/state.py`) reads the *raw* frontmatter (missing key → `intimate`, fail-closed). This let the same physical file be admitted at `ceiling=open` by the miner while the report refused it below `ceiling=intimate`, and let an untiered `10-Liminal/Compost` note under-report its own stamp as `unclassified` when the raw tier was `intimate`. The fix materializes `raw_privacy_tier` onto the fragment the miner returns and adds a `within_ceiling` hard-rank-cutoff gate before Fragment validation, closing both gaps with a new pinning test suite (`tests/test_liminal_tier_reader_agreement.py`).

## Strengths
- Root-cause fix: rather than patching the symptom, it points the mining-side reader at the exact primitive (`raw_privacy_tier`) the state-report side already used, eliminating a second, drifting implementation of the same policy.
- The `@computed_field` on `Fragment.voice_proxy_eligible` recomputes correctly off `model_copy(update={"privacy_tier": ...})` since it's a live property, not a cached value at construction — the materialization trick is safe.
- I traced the `_TIER_RANK` / `_OVERRIDE_RANK` tables by hand and confirmed the PR's own "strict narrowing at every ceiling" claim holds: explicit tiers (`open`/`personal`/`intimate`/`unclassified`) are unaffected, only the missing-key case moves from `unclassified` to `intimate`, which can only *raise* what a raw read admits, never lower it.
- Complexity was handled by extraction (`_admitted_liminal_entry`) rather than a `noqa`/threshold change, matching CLAUDE.md §1.3/§6.2's "no shortcuts" rule.
- New test module is a genuine differential test: it drives both private readers (`_admitted_liminal_notes`, `_load_liminal_fragments`) over the same fixture vault and asserts agreement at every ceiling, plus a positive control (`test_untiered_liminal_note_is_intimate_to_both_readers`) that specifically guards against the two readers drifting together toward the permissive answer rather than actually agreeing on the fail-closed one.
- The updated docstrings on `_content_tier` and `_mining_corpus` in `state.py` correct their own prior claims in place (the old docstring's "only ceiling at which the section did not render" reasoning was itself part of the bug) rather than leaving stale reasoning next to new code.

## Security Concerns
🟢 LOW — This is fixing a real privacy under-reporting bug (an `intimate` liminal note stamping its artifact as `unclassified`, which a `ceiling=personal` reader would then be allowed to open). The fix itself introduces no new exposure; if anything it closes exposure. No new user input handling, subprocess use, or path traversal surface is touched.

## Problems
None blocking. One non-blocking observation below.

## Code Quality
- 🟢 (non-blocking) `_load_liminal_fragments` now gates admission with `within_ceiling` (a hard rank cutoff) *before* calling `filter_fragments_by_tier` (the body-summarizing filter) on the same, already-gated entries. Working through the rank tables, I couldn't find a tier/ceiling combination where the second call still changes anything post-gate: whenever `filter_fragments_by_tier` would exclude (`INTIMATE` and `not _allows_intimate`) or summarize (`PERSONAL`/`UNCLASSIFIED` and `not _allows_full_personal_body`), `within_ceiling`'s identical `_TIER_RANK`-based cutoff has already excluded that entry upstream, using the same `ceiling = privacy_override or OPEN` default. So the trailing `filter_fragments_by_tier` call appears to now be a no-op for every code path — not a correctness bug, but worth confirming this reading and considering whether it can be dropped (or at least reflect that in the docstring, which currently frames it as answering "a different question" than the ceiling gate, which is true historically but no longer changes behavior downstream of the gate).

## CLAUDE.md Adherence
✅ Refactor-not-suppress used for the complexity bump (§1.3/§6.2).
✅ TDD framing per the PR description — RED assertion (`assert 'unclassified' == 'intimate'`) called out explicitly before the fix (§4.1/Gate 1).
✅ Docstrings are thorough and Google-style with Args/Returns throughout the new and touched functions (interrogate ≥95% likely satisfied).
⚠️ I could not execute `./scripts/check-all.sh`, `./scripts/test.sh`, or `uv sync` in this review sandbox (network/tooling access was gated), so I cannot personally confirm the 90% coverage / mypy strict / ruff-zero gates locally — I relied on careful static tracing instead. CI was still in progress (`IN_PROGRESS`) for the 3.11/3.12/3.13 quality jobs at review time; the PR description claims "Gate: 12/12" and Integration/E2E + CrawDad Quality jobs had already reported SUCCESS.

## Testing
Strong. `tests/test_liminal_tier_reader_agreement.py` is a differential/agreement test (not just a fixed-expectation test), parametrized across all four `PrivacyTierOverride` values, with a dedicated fail-closed positive control and a dedicated regression test for the `10-Liminal/Compost` under-report path (`test_untiered_compost_note_stamps_the_state_report_intimate`) that reads the stamp back off the actually-written artifact rather than trusting the internal method. This is exactly the shape of test that should exist for a "two readers must agree" bug class.

## Documentation
Docstrings are extensive and, notably, correct prior in-place documentation that itself encoded the bug's reasoning (the old `_content_tier` docstring reasoned only about `ceiling=open` and never checked what the note itself required — the new docstring fixes that reasoning, not just the code). The module docstring in `privacy_filter.py` already documented the "two tools disagree about the same file" bug class this PR closes, so the fix is well-anchored to existing project conventions.



## Verdict: LGTM

The fix correctly unifies the two tier readers (raw-frontmatter fail-closed vs. model-default) by materializing raw_privacy_tier onto the mining-side fragment and adding a within_ceiling hard-cutoff gate before validation; I traced the rank tables and confirmed the change is a strict narrowing at every ceiling with no regression for already-explicit tiers, and the new test module pins the two readers against each other plus the specific Compost under-report bug with a RED-then-GREEN style assertion. Nothing here blocks merge — the one thing worth flagging (retained but now-redundant filter_fragments_by_tier call) is a non-blocking simplification, not a defect.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5229012810 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

PR #1245, opened 2026-08-08
feat(api): serve journal upsert, wheel and reflections over /v1

### LINKED ISSUE #1075: feat(api): PUT /v1/journal-entries/{external_id} over the shared journal service, with HTTP↔MCP parity tests

### LINKED ISSUE #1076: feat(api): GET /v1/wheel over the shared wheel service — canonical frequency shares, deterministic and ceiling-filtered
## Role

Backend engineer landing the read vertical, and the one that must not quietly
adopt another product's vocabulary.

## Problem

`GET /v1/wheel` returns `not_implemented` after #1074.

More importantly, this is the capability where the two repos hold genuinely
different *concepts*, not different spellings:

- Creek's `wheel_tool` (`creek-tools/creek_mcp/tools/wheel.py:95-110`) returns
  `{F1..F10: {name, count, share}}` — how many classified corpus fragments sit
  at each APTITUDE **frequency**, and each frequency's share of the classified
  total.
- Adepthood validates `{aspects: [{stage_number, aspect, fullness}]}` — per-stage
  **aspect fullness** from its own 36-week curriculum
  (`backend/src/services/creek_vault_wheel.py`).

Both have ten members. That is a coincidence of cardinality, not a shared
meaning, and treating it as one is the fastest way to ship a Map that renders
confidently wrong numbers. Creek does not have Adepthood's stage/aspect
vocabulary and must not invent it.

## Goal

`/v1` serves Creek's frequency wheel verbatim under the canonical model, with
ceiling filtering, deterministic ordering and explicit empty-corpus behavior.
The stage/aspect projection stays **Adepthood's**, owned and tested there
(Geoffe-Ga/adepthood#1937).

## Architectural context

HTTP is the Adepthood application adapter; MCP stays the agent adapter; both
call one implementation. The projection boundary is a product decision recorded
in #1072's ADR: Creek publishes the corpus fact it owns; the consumer owns the
mapping into its own domain.

## Proposed tracer

A vault with three classified OPEN fragments yields a ten-member wheel whose
shares sum to 1.0; the same vault under a lower ceiling yields a different,
correctly filtered tally — over both HTTP and MCP identically.

## Tasks

- Route handler: resolve the admitted ceiling via #1073's policy, call the
  existing `wheel_tool`, project onto the canonical model. No counting logic in
  the route.
- Response model enforces: all ten frequencies present; a fixed, documented
  order; `count` a non-negative int; `share` in `[0.0, 1.0]`; `total_classified`
  and `unclassified` carried through.
- Extend the #1075 parity harness with the wheel scenario table.
- Document in `docs/api.md` — in the endpoint description itself, not a footnote
  — that this is frequency distribution over the classified corpus and is
  **not** a curriculum-progress or fullness measure.

## Acceptance criteria

- [ ] A populated vault returns exactly ten frequency entries, in the same order
      on every call, matching `CANONICAL_FREQUENCY_NAMES`.
- [ ] Shares sum to 1.0 (within float tolerance) when `total_classified > 0`.
- [ ] An empty or missing vault returns all ten entries with `count: 0` and
      `share: 0.0`, HTTP 200 — not 404, not an error, not an empty list.
- [ ] Fragments above the caller's admitted ceiling are excluded from the tally;
      a test builds a vault with `personal` fragments and asserts an `open`
      caller's counts exclude them.
- [ ] A fragment with an unrecognised frequency lands in `unclassified` and is
      excluded from `total_classified` — matching `_frequency_counts`'
      fail-closed branch (`tools/wheel.py:53-57`).
- [ ] HTTP and MCP produce the same counts, shares and ordering for the same
      vault and ceiling, via the parity harness.
- [ ] The HTTP audit record carries the authenticated consumer.
- [ ] No fragment title, body or id appears in the response, an error, or a log.
- [ ] `./scripts/test.sh --all` green; MCP wheel behavior unchanged.

## Verification

```bash
```

Coverage ≥90% branch, interrogate ≥95%, mypy strict, ruff clean, complexity ≤10.

- Ceiling filtering is enforced by `wheel_tool`'s existing
  `tier_within_override` call — the route must not pass a ceiling the adapter
  policy did not admit.
- The wheel is aggregate-only: no titles, bodies or ids may leak through it.
- A remote caller cannot obtain an `intimate`-inclusive tally; refused at the
  adapter edge per #1074.

## Files likely affected

- modify `creek-tools/creek_mcp/api/routes.py`, `creek_mcp/api/models.py`
- create `creek-tools/tests/test_api_wheel.py`
- modify `creek-tools/tests/helpers/adapter_parity.py`, `creek-tools/docs/api.md`

## Dependencies

Depends on #1072, #1073, #1074, #1075 (parity harness). Counterpart:
Geoffe-Ga/adepthood#2047, which owns the stage/aspect projection (their #1937).
Part of #1071.

## Budget escape hatch

None needed — this is the smallest vertical. If it grows, the growth is a sign
the projection is leaking into Creek. Push it back across the boundary rather
than splitting the issue.

### LINKED ISSUE #1077: feat(api): POST /v1/reflections over the shared reflect service — structured notes, care escalation, and INTIMATE refusal
## Role

Backend engineer landing the most safety-sensitive vertical. This route touches
acute-distress escalation, model routing, and content the caller may not be
admitted to. It goes last for that reason.

## Problem

`POST /v1/reflections` returns `not_implemented` after #1074, and reflection is
the capability where Adepthood and Creek disagree most completely — the failure
that motivated the whole epic:

- Adepthood sends `{consumer, body, tier_ceiling}`
  (`backend/src/services/creek_vault_client.py:277-279`); Creek's registered
  tool accepts `(content, entry_ref, privacy_tier_ceiling)`
  (`creek_mcp/server.py:332-346`). **No parameter name is shared.**
- Adepthood reads a scalar `payload["reflection"]` string
  (`creek_vault_client.py:406-407`); `reflect_tool` returns
  `{status, tool, tier_ceiling, routed_tier, notes[], essay_grounded, essay?}`
  (`creek_mcp/tools/reflect.py:479-490`).

Collapsing that structure into a string would discard exactly the fields that
make it safe: the verbatim-validated `notes[].quote` spans, the `escalate`
status, and `essay_grounded: False` — the flag that tells a client the essay is
free model prose and must not be presented as grounded in the user's own words.

`reflect_tool` already enforces four guarantees (module docstring,
`tools/reflect.py:1-56`): read-side ceiling gate above the care seam; INTIMATE
routed local via the tier-keyed LLM factory; verbatim quote validation; care
escalation *before* any model call. The route must inherit every one of them by
delegating, and add none of its own.

## Goal

`POST /v1/reflections` is a parse-and-delegate shim over `reflect_tool`, serving
the canonical structured result with `ok` / `empty` / `escalate` / `refused` /
failure distinguishable **without prose parsing**, and identical to MCP's
behavior under the parity harness.

## Architectural context

HTTP is the Adepthood application adapter; MCP stays the agent adapter; both
call one implementation. Care defense-in-depth lives in Creek and stays there
even though Adepthood also gates pre-call — two independent gates is the design,
not redundancy to trim.

## Proposed tracer

An OPEN-tier raw-`content` reflection with a stubbed LLM factory returns
`status: "ok"` with notes whose quotes are verbatim spans of the submitted
entry — over HTTP, byte-equivalent in behavior to the same call over MCP.

## Tasks

- Route handler: canonical request model (`content` XOR `entry_ref`), admitted
  ceiling from #1073's policy, call `reflect_tool`, project onto the canonical
  response model.
- Map `reflect_tool`'s structured outcomes onto the taxonomy without losing
  information: `ok` → 200; `empty` → 200 with an empty notes list and an
  explicit `empty` status; `escalate` → the `care_escalation` code carrying
  `CARE_SIGNAL`; `"entry_ref not found"` → `not_found`; `"entry_ref tier exceeds
  ceiling"` → `privacy_refused`; `"reflection unavailable: …"` →
  `temporarily_unavailable` with the exception **type name only**, never its
  message.
- Preserve `routed_tier` and `essay_grounded` on the wire; document that a
  client may not treat `essay` as grounded and that `notes[].quote` is the only
  verbatim-guaranteed field.
- Extend the #1075 parity harness with the reflection scenario table, including
  every refusal branch.

## Acceptance criteria

- [ ] `ok`, `empty`, `escalate`, `privacy_refused`, `not_found` and
      `temporarily_unavailable` are each distinguishable from a closed `status`
      or `code` enum — no client needs to read a reason string to branch.
- [ ] Every returned `notes[].quote` is a verified verbatim (whitespace-
      normalised) span of the submitted or referenced entry; a test feeds a
      model turn containing a hallucinated span and asserts it is dropped.
- [ ] A care-flagged entry returns `care_escalation` with `CARE_SIGNAL` and the
      LLM factory is **never invoked** — asserted on a spy, not inferred from
      the response.
- [ ] An `entry_ref` whose classified tier exceeds the caller's ceiling is
      refused, the model is not called, and the refusal reason does **not** echo
      the offending tier (preserving `reflect_tool`'s deliberate
      non-oracle wording, `tools/reflect.py:426-445`).
- [ ] A remote caller requesting an `intimate` ceiling is refused at the adapter
      edge (#1074) before this route runs; a test asserts no vault read and no
      model call occurred.
- [ ] Grounding retrieval excludes above-ceiling fragments; the routing tier
      still dominates every retrieved fragment's tier (the invariant pinned by
      `test_routing_tier_dominates_every_retrieved_fragment_tier`) — assert it
      holds for an HTTP caller too.
- [ ] `essay_grounded` is always present and `False`; a test asserts the field
      cannot be omitted by the serializer.
- [ ] No response body, error body, audit record or log line contains the entry
      body, a grounding title, or a fragment id.
- [ ] The HTTP audit record carries the authenticated consumer and, matching
      `reflect_tool`'s existing privacy design (`tools/reflect.py:400-409`),
      records only `has_entry_ref` — never the probed id or the outcome.
- [ ] Parity harness passes across every scenario above.
- [ ] `./scripts/test.sh --all` green; MCP reflect behavior unchanged.

## Verification

```bash
```

Coverage ≥90% branch, interrogate ≥95%, mypy strict, ruff clean, complexity ≤10.
Tests use a stubbed `llm_factory` — **no live model calls**, so the suite stays
deterministic and offline.

- Care escalation skips model generation entirely.
- Remote INTIMATE refused before content is read or a model is called.
- INTIMATE local-only routing is untouched: this route never picks a provider
  and never re-derives the tier itself.
- Quotes are verbatim or dropped.
- Nothing echoes the entry, and the audit record remains a non-oracle.

## Files likely affected

- modify `creek-tools/creek_mcp/api/routes.py`, `creek_mcp/api/models.py`
- create `creek-tools/tests/test_api_reflect.py`
- modify `creek-tools/tests/helpers/adapter_parity.py`, `creek-tools/docs/api.md`

## Dependencies

Depends on #1072, #1073, #1074, #1075 (parity harness). Counterpart:
Geoffe-Ga/adepthood#2047. Coordinate the response shape with #873 (praxis and
eddies extend this same payload) — do **not** implement #873 here. Related:
#1034 (cold-cache grounding latency) bounds this route's realistic timeout;
reference it when choosing the value, do not fix it here. Part of #1071.

## Budget escape hatch

If grounding retrieval makes the route's latency unacceptable under #1074's
timeout, ship the route with grounding served by an injected retrieval seam and
file the latency work against #1034. Never ship a reflection path that skips
the care gate, the ceiling gate, or quote validation to meet a deadline.

Geoffe-Ga:
Closes #1075. Closes #1076. Closes #1077. Completes epic #1071.

The three remaining `/v1` verticals as a **serial train** — they all mutate `IMPLEMENTED_CAPABILITIES` and `_IMPLEMENTED_HANDLERS`, so they could never be parallel. Each shares **one** behavioral implementation with its MCP tool; no logic is forked.

## The honesty guard was preserved, then rewritten by the right PR

`set(Capability) > IMPLEMENTED_CAPABILITIES` (a **strict** superset) and the `_UNIMPLEMENTED_CAPABILITIES` non-emptiness assertion stayed **intact and green** through #1075 and #1076. Equality is only reached when the third handler lands, so **#1077 alone** rewrote them — to `==`, with a docstring recording that equality was reached the intended way (three handlers built) and that **`>=` here would be the actual relaxation and must never be introduced**. An exhaustive four-name literal now guards the shrink direction the strict superset used to.

## #1077 refuses the issue's error mapping — the most consequential correction

The issue instructs: *"entry_ref not found" → `not_found`*.

`ErrorCode.NOT_FOUND` is documented in `api/models.py` as a **routing** code ("no such endpoint on this server") and is **AST-pinned to the routing layer** by `tests/test_v1_api_structure.py`. The reason is explicit: a caller who can distinguish *"no such fragment"* from *"not for you"* can enumerate the corpus one id at a time — the oracle that **#846/#970/#972/#1090 spent five issues collapsing**. Following the instruction would have turned that guard red.

Both vault-object non-answers collapse to `privacy_refused` with one identical message, making the **HTTP adapter stricter than the MCP tool** (whose two distinct reasons are a documented accepted residual). #1077's first acceptance criterion, listing `not_found` as separately distinguishable, is **deliberately not met**.

## Three more acceptance criteria that were wrong

**#1075 AC4** asks `tier: "intimate"` under an `open` ceiling to answer `privacy_refused`. It cannot — #1072 typed the field as `WireTierCeiling`, which has exactly two members, so `intimate` fails **schema** validation and correctly earns `invalid_request` (422). Answering `privacy_refused` would imply the server accepted the field, resolved something and ranked it. The reachable path (`personal` under `open` → 403, nothing staged, asserted on the filesystem) is covered separately.

**#1075's "Known inherited gap: #970"** is false — #970 is **closed**, and `_refuse_unadmitted_overwrite` already ships. The issue asked for a test *"pinning current behavior"* of a hole that no longer exists; the test pins the **fix** through the adapter instead.

**#1076 AC5** describes an unreachable branch. A fragment with frequency `F11` is schema-invalid and dropped by `iter_vault_fragments` **before** `wheel_tool` sees it — absent from the wheel, from `unclassified` and from `total_classified` alike. That is the safer behaviour and is pinned as such. The genuinely reachable fail-closed path is `primary: unclassified`, which has its own test.

**#1075's requested test path** `tests/helpers/adapter_parity.py` is unusable — `tests/helpers.py` already exists as a module, and a package shadows it, breaking `test_book_report_medium.py`. Hit for real mid-run. The harness lives at `tests/adapter_parity.py`, matching `tests/v1_api_support.py`.

## The 501 machinery is kept alive, not deleted

Once all four capabilities are implemented nothing reaches the not-implemented path. Rather than delete it, `stubbed_client()` substitutes one handler with `unimplemented(...)`, so the four ex-swept 501 tests plus a new one still drive it. Without this, **16 tests would have silently become empty-parametrize skips** — a guard that stopped guarding exactly when the next capability will need it.

## Coverage gap closed properly

`creek_mcp/httpapi/wheel.py` came in at **72.50%** against the 80% per-file gate. Closed to **87.5%** with real behavioural tests — **no waiver, no pragma**. The uncovered arms were worth having:

- **`vault is None`** — an unreadable `creek_config.yaml` degrades to `None`, and `wheel_tool` over an unresolved vault answers an **all-zero wheel**. Without this arm the route renders a broken install as *"you have written nothing"*. Now 503 `unavailable`, with the config path asserted absent from the body.
- **the projection guard** — refuses an eleventh frequency (`extra="forbid"`), a missing field, an uncoercible count, or an out-of-range ceiling, rather than serving a partial 200. Five parametrised rows, one per member of the caught exception tuple.
- **the tool-refusal arm** — maps to 500 and is asserted **not** to echo the structured `reason`, which can carry a vault path.

The only remaining uncovered lines are the `if TYPE_CHECKING:` block, which every sibling adapter also carries and which is unreachable by construction.

## Found, not fixed

`docs/contracts/adepthood-v1/examples/reflections/care-escalation.json` builds `reason` from `CARE_SIGNAL["kind"]` (`acute_distress`), but the runtime carries the **guard's** reason (`acute_distress_markers`). The route carries what the guard decided and the test asserts against the guard, not the fixture. **The published fixture is wrong** — flagged in a comment rather than silently papered over. Worth a follow-up against the bundle.

## One refactor, called out

`configured_vault` / `UNREADABLE_CONFIG` moved out of `httpapi/capabilities.py` into a new `httpapi/vault.py`, because journal/wheel/reflect all need per-request vault resolution and three copies would drift. Consequence: two tests' `monkeypatch` target moved — a real test change caused by a real refactor, named here rather than buried.

Gate: **12/12**, 8437 passed, 94.14% aggregate, per-file gate 254 files ≥ 80%.

### DIFF
--- a/creek-tools/creek_mcp/api/models.py
+++ b/creek-tools/creek_mcp/api/models.py
@@ -88,12 +88,18 @@ narrows, and the client needs to see the whole window without a second round
 trip. Drop an entry only when a ``/v1`` shape actually stops being served.
 """
 
-OK_STATUS: Final[str] = "ok"
+OK_STATUS: Final[Literal["ok"]] = "ok"
 """The sole success status on the journal-upsert and wheel envelopes.
 
 Both carry ``status: "ok"`` and nothing else, because every non-success outcome
 is an :class:`ErrorEnvelope` with its own HTTP status. A second success-ish
 status value would give a consumer two places to look for failure.
+
+Typed ``Literal["ok"]`` rather than ``str`` so a handler can *build* a response
+from it: :attr:`JournalUpsertResponse.status` and :attr:`WheelResponse.status`
+are themselves ``Literal["ok"]``, and a plain ``str`` constant would not satisfy
+them under strict typing — leaving a handler to spell the literal a second time,
+which is the drift this constant exists to prevent.
 """
 
 
--- a/creek-tools/creek_mcp/api/routes.py
+++ b/creek-tools/creek_mcp/api/routes.py
@@ -14,11 +14,13 @@ the framework-free half of the surface beside :mod:`creek_mcp.api.models`, is
 what lets the published OpenAPI document outlive whatever serves it.
 
 **Exposed versus implemented.** :data:`ROUTES` says which endpoints exist;
-:data:`IMPLEMENTED_CAPABILITIES` says which of them actually answer. At #1074
-that is the handshake alone — the other three are wired to an honest ``501``
-rather than to a fabricated ``200``, because a stub that looks like success is
-one a consumer integrates against and only discovers in production. #1075—#1077
-each move one capability from the first set into the second.
+:data:`IMPLEMENTED_CAPABILITIES` says which of them actually answer. #1075—#1077
+moved the last three capabilities from the first set into the second, so the two
+now coincide. The distinction is not obsolete: while a capability was unbuilt it
+was wired to an honest ``501`` rather than to a fabricated ``200``, because a
+stub that looks like success is one a consumer integrates against and only
+discovers in production. A fifth capability published before its handler exists
+gets exactly that treatment, from the same constant, with no code change.
 """
 
 from __future__ import annotations
@@ -180,14 +182,16 @@ gate for the same reason — an operator debugging a version mismatch must not
 also lose their monitoring probe.
 """
 
-IMPLEMENTED_CAPABILITIES: Final[frozenset[Capability]] = frozenset(
-    {Capability.CAPABILITIES}
-)
+IMPLEMENTED_CAPABILITIES: Final[frozenset[Capability]] = frozenset(Capability)
 """The capabilities this server actually answers for, as opposed to publishes.
 
-A *strict* subset of ``set(Capability)`` at #1074: the handshake is built, and
-``journal-upsert`` / ``reflections`` / ``wheel`` are mounted to an honest
-``501``. One constant drives both the advertised list in ``GET
+Equal to ``set(Capability)`` since #1077 — spelled ``frozenset(Capability)``
+rather than as four names, so a capability added to the enum is advertised only
+once a handler exists for it: :func:`creek_mcp.httpapi.handlers._handler_for`
+raises at import if it does not, which is the failure mode worth having.
+
+It was a *strict* subset for the whole of #1074—#1076, and the machinery that
+made that safe is unchanged. One constant drives both the advertised list in ``GET
 /v1/capabilities`` and which routes get the stub, so the two cannot disagree —
 adding a capability here without wiring a handler, or wiring a handler without
 adding it here, turns ``tests/test_v1_api_capabilities.py`` red.
--- a/creek-tools/creek_mcp/httpapi/app.py
+++ b/creek-tools/creek_mcp/httpapi/app.py
@@ -63,6 +63,7 @@ from creek_mcp.httpapi.middleware.limits import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
     from pathlib import Path
 
     from starlette.requests import Request
@@ -71,6 +72,7 @@ if TYPE_CHECKING:
     from creek_mcp.api.routes import RouteSpec
     from creek_mcp.httpapi.handlers import Handler
     from creek_mcp.remote_auth import ConsumerTokenVerifier
+    from creek_mcp.tools.reflect import _LLMFactory
 
 
 def _speaks_a_served_minor(request: Request) -> bool:
@@ -198,6 +200,7 @@ def create_app(
     max_body_bytes: int = DEFAULT_MAX_BODY_BYTES,
     timeout_seconds: float = DEFAULT_TIMEOUT_SECONDS,
     max_concurrency: int = DEFAULT_MAX_CONCURRENCY,
+    reflect_llm_factory: Callable[[], _LLMFactory] | None = None,
 ) -> Starlette:
     """Build the ``/v1`` application.
 
@@ -212,6 +215,14 @@ def create_app(
         max_body_bytes: Inclusive request-body cap.
         timeout_seconds: Per-request deadline.
         max_concurrency: In-flight request limit, process-wide.
+        reflect_llm_factory: Thunk returning the tier-keyed LLM factory
+            ``POST /v1/reflections`` calls, or ``None`` to build the production
+            one lazily on first use. A thunk rather than a factory so a server
+            with no provider configured still boots — only a reflection call
+            then fails, and it fails as a structured refusal. Injectable so the
+            suite can run the whole vertical offline against a stub, which is
+            what keeps the care and ceiling gates testable by *observing that
+            the model was never reached*.
 
     Returns:
         The configured application.
@@ -234,4 +245,5 @@ def create_app(
         exception_handlers={HTTPException: _routing_miss},
     )
     app.state.vault_path = vault_path
+    app.state.reflect_llm_factory = reflect_llm_factory
     return app
--- a/creek-tools/creek_mcp/httpapi/capabilities.py
+++ b/creek-tools/creek_mcp/httpapi/capabilities.py
@@ -36,12 +36,10 @@ until it has already written the integration.
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Final
+from typing import TYPE_CHECKING
 
 from starlette.concurrency import run_in_threadpool
-from yaml import YAMLError
 
-from creek.config import load_config
 from creek_mcp.api.models import (
     CONTRACT_MINOR,
     SUPPORTED_CONTRACT_MINORS,
@@ -60,6 +58,7 @@ from creek_mcp.contract import CONTRACT_VERSION, ONTOLOGY_VERSION
 from creek_mcp.httpapi import SERVER_NAME
 from creek_mcp.httpapi.context import context_of
 from creek_mcp.httpapi.errors import HTTP_OK, json_response
+from creek_mcp.httpapi.vault import UNREADABLE_CONFIG, configured_vault
 from creek_mcp.tools.handshake import handshake_tool, vault_available
 
 if TYPE_CHECKING:
@@ -69,16 +68,6 @@ if TYPE_CHECKING:
     from starlette.responses import Response
 
     from creek_mcp.httpapi.context import RequestContext
-
-_UNREADABLE_CONFIG: Final = (OSError, ValueError, YAMLError)
-"""What "the vault is not usable" looks like when it is raised rather than returned.
-
-:class:`FileNotFoundError` and a permission error are :class:`OSError`; a
-malformed ``vault_path`` and every :class:`pydantic.ValidationError` are
-:class:`ValueError`; an unparseable document is a :class:`yaml.YAMLError`.
-Deliberately not ``Exception``: a bug in this module must surface as a ``500``
-from the error boundary, not as a vault that quietly reports itself missing.
-"""
 
 
 def _implemented_capabilities() -> list[Capability]:
@@ -96,30 +85,6 @@ def _implemented_capabilities() -> list[Capability]:
         for capability in Capability
         if capability in IMPLEMENTED_CAPABILITIES
     ]
-
-
-def _configured_vault(request: Request) -> Path | None:
-    """Return the vault this request should read, resolving config if need be.
-
-    Resolved per request rather than once at construction: an operator who
-    fixes a broken ``creek_config.yaml`` gets a working handshake on the next
-    call rather than after a restart.
-
-    Args:
-        request: The request in flight.
-
-    Returns:
-        The explicitly configured vault path, the one the config names, or
-        ``None`` when the configuration cannot be read at all.
-    """
-    configured: Path | None = request.app.state.vault_path
-    if configured is not None:
-        return configured
-    try:
-        resolved = load_config().vault_path
-    except _UNREADABLE_CONFIG:
-        return None
-    return resolved
 
 
 def _negotiate(vault: Path, context: RequestContext) -> bool:
@@ -150,7 +115,7 @@ def _negotiate(vault: Path, context: RequestContext) -> bool:
             privacy_tier_ceiling=context.ceiling,
             consumer=context.consumer,
         )
-    except _UNREADABLE_CONFIG:
+    except UNREADABLE_CONFIG:
         return False
     return bool(negotiated["available"])
 
@@ -162,7 +127,8 @@ def _vault_is_usable(request: Request, context: RequestContext) -> bool:
     syscall the handshake makes is reachable from here and nowhere else, so
     :func:`handle_capabilities` can hoist the lot off the event loop with a
     single :func:`~starlette.concurrency.run_in_threadpool` call. The three
-    seams underneath are all filesystem I/O — :func:`_configured_vault` reads
+    seams underneath are all filesystem I/O —
+    :func:`~creek_mcp.httpapi.vault.configured_vault` reads
     and parses ``creek_config.yaml``, ``vault_available`` stats the marker, and
     :func:`_negotiate` reaches :meth:`creek_mcp.audit.MCPAuditLog.append`,
     which takes a thread lock and an ``fcntl`` exclusive lock across a full
@@ -178,7 +144,7 @@ def _vault_is_usable(request: Request, context: RequestContext) -> bool:
         ``True`` only when the marker directory exists *and* the handshake tool
         confirms it.
     """
-    vault = _configured_vault(request)
+    vault = configured_vault(request)
     if vault is None or not vault_available(vault):
         return False
     return _negotiate(vault, context)
@@ -294,7 +260,7 @@ async def handle_capabilities(request: Request) -> Response:
     across an ``fsync``. Called inline it would stop the event loop, so one
     consumer's slow audit log would stall every other connection this process is
     serving, on the endpoint every client calls *first*. Exception behaviour is
-    unchanged: the ``_UNREADABLE_CONFIG`` catches sit inside that function and
+    unchanged: the ``UNREADABLE_CONFIG`` catches sit inside that function and
     anything they do not catch propagates back out through the ``await``.
 
     It is also what lets
--- a/creek-tools/creek_mcp/httpapi/handlers.py
+++ b/creek-tools/creek_mcp/httpapi/handlers.py
@@ -1,21 +1,34 @@
 """What each ``/v1`` operation actually does, keyed by ``operation_id`` (#1074).
 
-Three of the five published routes are not built yet, and the dangerous shape
-for that is a route that returns a *plausible* success — an empty wheel, a
-fabricated ``fragment_id``, a ``{"status": "ok"}`` with nothing behind it —
-because a consumer integrating against it writes code that passes CI and
-silently does nothing in production. ``501 unsupported_capability`` is the
-honest answer, and the ADR puts it in the "safe to expose" column precisely
-because it is derived from the server's *static route table* and discloses
-nothing about the vault.
+Every published route is built as of #1077: the handshake and health from
+#1074, the journal upsert from #1075, the wheel from #1076 and the reflection
+from #1077. :data:`HANDLERS` is derived from the route table rather than listed,
+so a route added to :data:`~creek_mcp.api.routes.ROUTES` without a handler fails
+at *import* instead of being mounted to nothing.
 
-**One stub, three mountings.** :func:`unimplemented` is a factory, not three
-hand-written handlers, so the three unbuilt routes cannot drift into three
-different degrees of half-built. It stamps its product with
-``unimplemented_capability``, which makes "is this endpoint the stub?" a fact
-about the mounted route rather than a guess from its behaviour — and that is
-what lets ``tests/test_v1_api_capabilities.py`` hold advertised and implemented
-together in both directions.
+**The honesty stub stays, and stays exercised.** While a route was unbuilt it
+answered ``501 unsupported_capability`` rather than a *plausible* success — an
+empty wheel, a fabricated ``fragment_id``, a ``{"status": "ok"}`` with nothing
+behind it — because a consumer integrating against one of those writes code
+that passes CI and silently does nothing in production. The ADR puts ``501`` in
+the "safe to expose" column precisely because it is derived from the server's
+*static route table* and discloses nothing about the vault.
+
+Nothing in the table reaches :class:`UnimplementedHandler` today, and the
+temptation is to delete it. That would retire the guard along with the interim
+it covered: the *next* capability added to
+:class:`~creek_mcp.api.models.Capability` before its handler exists is the case
+the stub is for, and by then nobody would have run it in months. So it remains,
+and ``tests/test_v1_api_not_implemented.py`` drives it by substituting one entry
+of :data:`HANDLERS` — the same path a genuinely unbuilt capability takes.
+
+**One stub, however many mountings.** :func:`unimplemented` is a factory rather
+than hand-written handlers, so unbuilt routes cannot drift into different
+degrees of half-built. It stamps its product with ``unimplemented_capability``,
+which makes "is this endpoint the stub?" a fact about the mounted route rather
+than a guess from its behaviour — and that is what lets
+``tests/test_v1_api_capabilities.py`` hold advertised and implemented together
+in both directions.
 
 **The capability gate runs before body validation.** The stub reads nothing the
 caller sent. A stub that answered ``422`` for a malformed body and ``501`` for a
@@ -25,9 +38,7 @@ works too.
 
 :data:`HANDLERS` is read by :func:`creek_mcp.httpapi.app.create_app` at call
 time rather than captured at import, so a test can substitute one entry and
-build an app around it. #1075—#1077 each replace one factory product here with a
-real handler and add its capability to
-:data:`creek_mcp.api.routes.IMPLEMENTED_CAPABILITIES`.
+build an app around it.
 """
 
 from __future__ import annotations
@@ -43,11 +54,17 @@ from creek_mcp.api.routes import (
     IMPLEMENTED_CAPABILITIES,
     OP_CAPABILITIES,
     OP_HEALTH,
+    OP_JOURNAL_UPSERT,
+    OP_REFLECTIONS,
+    OP_WHEEL,
     ROUTES,
 )
 from creek_mcp.httpapi.capabilities import handle_capabilities
 from creek_mcp.httpapi.context import context_of
 from creek_mcp.httpapi.errors import HTTP_OK, error_response, json_response
+from creek_mcp.httpapi.journal import handle_journal_upsert
+from creek_mcp.httpapi.reflect import handle_reflection
+from creek_mcp.httpapi.wheel import handle_wheel
 
 if TYPE_CHECKING:
     from creek_mcp.api.models import Capability
@@ -149,6 +166,9 @@ async def handle_health(_request: Request) -> Response:
 _IMPLEMENTED_HANDLERS: Final[dict[str, Handler]] = {
     OP_CAPABILITIES: handle_capabilities,
     OP_HEALTH: handle_health,
+    OP_JOURNAL_UPSERT: handle_journal_upsert,
+    OP_REFLECTIONS: handle_reflection,
+    OP_WHEEL: handle_wheel,
 }
 """The operations that answer for real, keyed by ``operation_id``.
 
--- a/creek-tools/creek_mcp/httpapi/journal.py
+++ b/creek-tools/creek_mcp/httpapi/journal.py
@@ -1,0 +1,275 @@
+"""``PUT /v1/journal-entries/{external_id}`` — parse, delegate, project (#1075).
+
+The first tracer stub replaced by real behaviour, and the shape every later
+vertical copies: this module validates what arrived, hands it to the **existing**
+:func:`creek_mcp.tools.journal.journal_ingest_tool`, and projects that tool's
+return onto the published wire model. It contains no idempotency logic, no
+privacy logic and no audit call of its own.
+
+That is not a style preference. ``journal_ingest_tool`` refuses an entry whose
+tier exceeds the caller's ceiling *before* staging it, refuses an update that
+would destroy a fragment the caller could not have read (#970), and audits both
+the refusal and the success. A route that reached ``run_ingest`` directly, or
+that passed a ceiling the adapter policy had not admitted, would lose every one
+of those guarantees while still answering plausibly — which is exactly the
+divergence epic #1071 exists to prevent.
+
+**What this module *does* own is the path segment.** ``external_id`` arrives as
+URL text rather than as a validated field, and
+:func:`creek_mcp.staged_names.safe_stem` accepts literally any string: it would
+happily mint a stable staged name for a 20 KB id, or for one whose bytes did not
+survive URL decoding. Either mints a fragment under a key the client can never
+address again, so :func:`admissible_external_id` refuses them here, above the
+tool, before anything is written.
+
+**And the refusal projection.** The tool answers in its own structured
+vocabulary — ``{"status": "refused", ..., "reason": ...}`` — which #1072 does not
+publish. :func:`journal_refusal_code` is the single translation into the wire
+taxonomy, and it is deliberately total: an unrecognised reason becomes
+``internal_error`` rather than a plausible refusal, because a reason this adapter
+does not understand is a reason it must not narrate. ``ingest failed: …`` in
+particular can carry a staged file path, and :func:`error_response` renders a
+constant message per code, so nothing of it reaches the wire.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Final
+
+from pydantic import ValidationError
+from starlette.concurrency import run_in_threadpool
+
+from creek_mcp.api.models import (
+    OK_STATUS,
+    ErrorCode,
+    JournalAction,
+    JournalUpsertRequest,
+    JournalUpsertResponse,
+    WireTierCeiling,
+)
+from creek_mcp.httpapi.context import context_of
+from creek_mcp.httpapi.errors import HTTP_OK, error_response, json_response
+from creek_mcp.httpapi.vault import configured_vault
+from creek_mcp.read_gate import GENERIC_ABOVE_CEILING_REASON
+from creek_mcp.tools.journal import journal_ingest_tool
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from starlette.requests import Request
+    from starlette.responses import Response
+
+    from creek_mcp.httpapi.context import RequestContext
+
+MAX_EXTERNAL_ID_CHARS: Final[int] = 512
+"""Inclusive upper bound on an ``external_id`` path segment.
+
+Generous enough for any namespaced consumer id — the published example is 38
+characters — and small enough that the id cannot become a payload. There is no
+bound below this one: ``safe_stem`` truncates only the *readable slug* to 80
+characters and then appends a digest of the whole raw id, so an unbounded id
+would be accepted, stored and echoed at full length.
+"""
+
+REPLACEMENT_CHARACTER: Final[str] = "�"
+"""U+FFFD, the marker that a path segment did not survive URL decoding.
+
+Its presence means the client's bytes and the server's string already differ, so
+the id the client would address the entry by is not the id the entry was written
+under. That is an idempotency key that silently does not work, which is worse
+than a refusal.
+"""
+
+_BLANK_CALL_REASON: Final[str] = "content and external_id are required"
+"""``_validated_entry_tier``'s malformed-call refusal, verbatim."""
+
+_UNKNOWN_TIER_PREFIX: Final[str] = "unknown tier "
+"""``_validated_entry_tier``'s unparseable-tier refusal, by prefix."""
+
+_ABOVE_CEILING_SUFFIX: Final[str] = " exceeds the ceiling"
+"""Gate 1's refusal (``entry tier <tier> exceeds the ceiling``), by suffix."""
+
+_VAULT_UNAVAILABLE_REASON: Final[str] = "vault unavailable"
+"""The refusal a missing vault earns from the ingest runner, verbatim.
+
+Transient rather than terminal: the vault directory reappearing is exactly the
+kind of thing that clears on its own, which is why it maps to
+``temporarily_unavailable`` and not to ``unavailable``.
+"""
+
+
+def journal_refusal_code(reason: str) -> ErrorCode:
+    """Return the wire code for one of the journal tool's refusal reasons.
+
+    Total by construction. Two reasons share the ``internal_error`` fallthrough
+    and neither gets a branch of its own, because a branch that returned what
+    the default already returns is a branch nothing can go wrong in:
+
+    * ``ingest failed: <message>`` — the entry was staged and tier-allowed and
+      the write failed anyway. The message can name a staged file path, which is
+      why this must not reach a body that echoes anything; it does not, because
+      :func:`~creek_mcp.httpapi.errors.error_response` renders one constant per
+      code.
+    * anything this adapter does not recognise. Failing closed to
+      ``internal_error`` rather than to ``privacy_refused`` matters: a refusal
+      this adapter cannot classify is not a privacy decision it may claim to
+      have made on the caller's behalf, and ``internal_error`` is the one code
+      that asserts nothing about the vault.
+
+    The four matched reasons are spelled as constants above rather than imported
+    because :func:`creek_mcp.tools.journal.journal_ingest_tool` composes three of
+    them inline as f-strings, so there is nothing to import. Each live branch is
+    pinned by a behavioural test that drives the real tool through the route, so
+    a reworded reason surfaces as a wrong status code rather than as silence.
+
+    Args:
+        reason: The ``reason`` field of a structured tool refusal.
+
+    Returns:
+        The published :class:`~creek_mcp.api.models.ErrorCode`.
+    """
+    if reason == _BLANK_CALL_REASON or reason.startswith(_UNKNOWN_TIER_PREFIX):
+        return ErrorCode.INVALID_REQUEST
+    if reason == GENERIC_ABOVE_CEILING_REASON or reason.endswith(_ABOVE_CEILING_SUFFIX):
+        return ErrorCode.PRIVACY_REFUSED
+    if reason == _VAULT_UNAVAILABLE_REASON:
+        return ErrorCode.TEMPORARILY_UNAVAILABLE
+    return ErrorCode.INTERNAL_ERROR
+
+
+def admissible_external_id(raw: str) -> bool:
+    """Return whether *raw* can serve as an idempotency key.
+
+    Args:
+        raw: The decoded ``{external_id}`` path segment.
+
+    Returns:
+        ``True`` when the id is non-blank, within
+        :data:`MAX_EXTERNAL_ID_CHARS`, free of the replacement character, and
+        made entirely of printable characters — a control byte would reach both
+        the staged frontmatter and the audit trail.
+    """
+    if not raw.strip() or len(raw) > MAX_EXTERNAL_ID_CHARS:
+        return False
+    if REPLACEMENT_CHARACTER in raw:
+        return False
+    return raw.isprintable()
+
+
+async def _parsed_body(request: Request) -> JournalUpsertRequest | None:
+    """Return the validated request body, or ``None`` when it does not validate.
+
+    Args:
+        request: The request in flight.
+
+    Returns:
+        The parsed model, or ``None``. The caller renders one
+        ``invalid_request`` for both an undecodable body and a schema failure:
+        a caller able to tell those apart learns which half of its request the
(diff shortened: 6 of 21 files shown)


Geoffe-Ga:
## Summary
PR #1245 completes epic #1071 by implementing the three remaining `/v1` HTTP endpoints (`PUT /v1/journal-entries/{external_id}`, `POST /v1/reflections`, `GET /v1/wheel`) as thin parse-and-delegate adapters over existing MCP tools (`journal_ingest_tool`, `reflect_tool`, `wheel_tool`), replacing the honest `501 unsupported_capability` stubs from #1074. It adds ~2450 lines of new tests including a reusable HTTP↔MCP behavioral-parity harness (`tests/adapter_parity.py`), and closes `creek_mcp/httpapi/wheel.py`'s coverage gap from 72.50% to 87.5% with real behavioral tests.

## Strengths
- **Delegation discipline**: all three new handler modules (`journal.py`, `reflect.py`, `wheel.py`) genuinely contain no privacy, idempotency, or audit logic of their own — everything is inherited from the existing MCP tools, which is exactly what the epic requires and what the parity harness verifies.
- **Deliberate, well-reasoned deviations from the source issues**: the PR refuses three acceptance criteria from #1075/#1076/#1077 (the `not_found` mapping for `entry_ref`, the "known inherited gap #970" test, and the F11-unreachable-branch test) and I independently verified each justification is correct — #970 is genuinely closed, `ErrorCode.NOT_FOUND` really is AST-pinned to the routing layer only (`tests/test_v1_api_structure.py`), and `Fragment` really does reject `F11` at load before `wheel_tool` ever sees it.
- **Privacy-oracle avoidance**: collapsing "entry not found" and "entry above ceiling" into one identical `privacy_refused` message (in `reflect.py`) correctly prevents corpus enumeration, and is tested (`test_an_unresolvable_entry_ref_is_privacy_refused_not_not_found`).
- **The 501 honesty-stub machinery is kept alive and exercised** even though nothing in the real route table reaches it anymore, via a substituted-handler test — a good call, since letting that guard bit-rot would leave the *next* capability unprotected.
- **Found-not-fixed discipline**: the PR identifies that the committed `docs/contracts/adepthood-v1/examples/reflections/care-escalation.json` fixture publishes a stale `reason` value (`acute_distress` from `CARE_SIGNAL["kind"]`) that doesn't match the runtime's actual guard-derived reason (`acute_distress_markers`) — I confirmed this discrepancy directly in `creek/care/guardrail.py` and the fixture file — and correctly leaves it as a flagged follow-up rather than silently patching an out-of-scope contract file.
- Extensive negative-disclosure testing throughout (no echoed content, paths, tiers, or exception messages in error bodies/logs/audit records), and filesystem-level assertions (not just status codes) for refusal paths.

## Security Concerns
🟢 LOW — No blocking issues found. Verified: `journal_refusal_code`/`reflect_refusal_code`/`wheel_refusal_code` are total and fail closed to `internal_error` for unrecognized reasons; `error_response` renders a constant message per code so caught exception messages/paths never reach the wire; `external_id` path-segment validation (`admissible_external_id`) is defense-in-depth on top of `safe_stem`'s hash-based derivation, which already prevents path traversal or collision regardless of this new check; `configured_vault`'s `UNREADABLE_CONFIG` tuple is intentionally narrow (`OSError, ValueError, YAMLError`) so a genuine bug still surfaces as a 500 rather than being swallowed as "no vault."

## Problems
None blocking.

## Code Quality
- Non-blocking: `docs/api.md`/CHANGELOG.md convention — this epic's landmark PRs (#1074 and this one) don't add `CHANGELOG.md` entries, unlike some other tracked features (`FEAT-*`/`GAP-*`). Not a deviation introduced by this PR (the pattern was already absent for #1074), so not blocking, but worth considering for visibility given this completes a whole epic.
- The docstrings throughout are unusually dense (bordering on essay-length in places, e.g. `journal.py`/`reflect.py`/`wheel.py` module docstrings). This is consistent with the rest of the codebase's established style in this epic, so not a deviation, but reviewers on future PRs may want to watch for docstring bloat outpacing code.

## CLAUDE.md Adherence
✅ TDD workflow evident from test-first framing in commit message and issue tasks. ✅ Coverage: PR claims 94.14% aggregate / wheel.py at 87.5% (both ≥ the 90%/80% gates); GitHub CI shows "Integration & E2E Tests" and "CrawDad Quality" already green, with "Code Quality & Testing" (which runs coverage/lint/mypy) still in progress at time of review. ✅ Commit message follows Conventional Commits (`feat(api): ...`). ✅ New/modified public functions carry complete Google-style docstrings with Args/Returns/Raises as applicable, consistent with the ≥95% interrogate gate. ✅ Type hints used throughout with `from __future__ import annotations` and `TYPE_CHECKING` guards matching existing module conventions.

## Testing
Strong. The new HTTP↔MCP parity harness (`tests/adapter_parity.py`) is a genuine architectural asset — it compares behavioral outcomes (success/refusal code, payload facts, audit trail) rather than brittle envelope equality, is proven non-vacuous in each of the three new test files (`test_the_parity_harness_can_actually_fail`), and is reused rather than reimplemented per vertical. Refusal paths are asserted on the filesystem/audit trail, not just HTTP status. Care-escalation "model never called" is asserted via a factory spy rather than inferred from the response shape. The wheel coverage-gap closure specifically targets the three previously-uncovered branches (`vault is None`, the five-row projection-guard parametrize, the tool-refusal-arm-doesn't-echo-reason case) that the PR claims to have closed.

## Documentation
`docs/api.md` is updated to reflect all four capabilities as implemented, with clear callouts on the wheel/curriculum vocabulary boundary and the deliberate `not_found`→`privacy_refused` collapse. Docstrings across new modules explain *why*, not just *what*, consistent with the rest of this codebase's style.



## Verdict: LGTM

Based on a full read of the diff plus independent verification against the repo (checked #970's closed status, the NOT_FOUND AST-pin guard in test_v1_api_structure.py, the safe_stem hashing that neutralizes path-traversal concerns in external_id, and the pre-existing care-escalation.json fixture bug the PR correctly flags rather than silently patches), I found no blocking defects: privacy/oracle-avoidance logic is correctly implemented and tested, refusal-code mappings fail closed to internal_error and never echo caller input or exception messages, and the new behavior is backed by extensive HTTP↔MCP parity tests plus filesystem/audit assertions, not just status-code checks. I would merge this as-is.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5228682208 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR completes epic #1071 by implementing the last three `/v1` HTTP verticals (journal upsert, wheel, reflections) as thin adapters over existing MCP tools, replacing their `501` stubs. It's large (3.6k additions, 20 files) but disciplined: shared behavior with the MCP tools, an honest capability/handler pairing, a new HTTP↔MCP parity test harness, and an unusually thorough, self-critical PR description that documents where the author deliberately deviated from the source issues' acceptance criteria (and why). Two real defects survive review despite that care: an event-loop-blocking regression in all three new handlers, and a fabricated-`fragment_id`-on-wire bug in the journal success path.

## Strengths
- **Single source of truth preserved.** `journal.py`/`reflect.py`/`wheel.py` add no privacy, tier, or audit logic of their own — they delegate entirely to `journal_ingest_tool` / `reflect_tool` / `wheel_tool`, which is exactly what epic #1071 requires to prevent HTTP/MCP behavioral drift.
- **`tests/adapter_parity.py`** is a genuine behavioral-parity harness (verified: compares projected outcomes and audit fields across two real call paths, not a tautological self-comparison), and it caught a real naming collision (`tests/helpers/` would have shadowed `tests/helpers.py`) before merge.
- **`ErrorCode.NOT_FOUND` is never emitted** by any new refusal-mapping function (verified directly) — the anti-enumeration invariant from #846/#970/#972/#1090 is honestly preserved even though it means the PR knowingly diverges from its own source issue's acceptance criteria; both `error_response` and each refusal-code function correctly avoid interpolating tool-provided free text onto the wire.
- The `501` stub machinery is kept alive and still exercised (`stubbed_client()` substitutes a handler) rather than deleted now that nothing naturally reaches it — a good call for future capability additions.
- `admissible_external_id` plus the downstream `safe_stem` sanitation closes off path-traversal / oversized-id vectors on the new idempotency key surface.
- Coverage claim for `wheel.py` (72.5% → 87.5%, no waiver/pragma) is confirmed against `scripts/coverage-waivers.txt` and the actual test file's real behavioral assertions on the three previously-uncovered arms.

## Security Concerns
🟡 **HIGH — Event-loop blocking via `configured_vault()` outside `run_in_threadpool`.** In `journal.py:271`, `reflect.py:306`, and `wheel.py:148`, `vault = configured_vault(request)` runs directly inside the `async def handle_*` coroutine, *before* the surrounding `run_in_threadpool(...)` call. `configured_vault` falls back to `load_config()` whenever `app.state.vault_path` is `None` — which is the actual default for the production CLI entry point (`cli.py:248` calls `create_app(verifier=verifier)` and never passes `vault_path`). `load_config()` does synchronous file I/O and YAML parsing. This blocks the ASGI event loop on every `/v1/journal-entries`, `/v1/wheel`, and `/v1/reflections` request in the default deployment mode — a resource-exhaustion/latency-amplification vector under concurrent load, and it also silently weakens `RequestTimeoutMiddleware`'s cancellation guarantee for that portion of the request, exactly the failure mode `capabilities.py`'s own docstring (lines 266–274) describes and guards against by wrapping the equivalent call (`_vault_is_usable`, which itself calls `configured_vault`) fully inside `run_in_threadpool`. The three new handlers deviate from that established, still-present pattern. Fix: move the `configured_vault` call inside each handler's existing `run_in_threadpool`-wrapped worker function (`_upsert`, `_reflect`, `_tally`), or otherwise thread it off the loop.

No other security issues found: auth/token verification, ceiling enforcement, and error-message genericity (no leaked paths/reasons on any refusal path, verified for journal/reflect/wheel) all held up under review.

## Problems
🔴 **`journal.py` renders a fabricated `"fragment_id": "None"` on a real, documented race.** In `_render` (`journal.py`), `fragment_id = result.get("fragment_id")` is later passed as `fragment_id=str(fragment_id)` into `JournalUpsertResponse` inside a `try/except (ValidationError, ValueError, KeyError)` block. The adjacent comment states this except exists to catch "most plausibly an unresolved `fragment_id`" — but `journal_ingest_tool` (`creek_mcp/tools/journal.py:430,445`) can return `"status": "ok"` with `fragment_id=None` when `_resolve_fragment_id` fails to resolve the id right after a successful write, and `str(None)` never raises. The handler therefore returns a `200` carrying the literal string `"fragment_id": "None"` instead of falling into the `internal_error` path the code claims to guard for — precisely the "fabricated id" outcome this PR's own module docstring says must never happen ("Not something to paper over with a fabricated id"). Fix: explicitly check `if fragment_id is None: return error_response(ErrorCode.INTERNAL_ERROR, context)` before construction.

Both problems above are narrow-window in practice (the second requires a specific race; the first depends on config-file read latency), but both are real, concretely reachable, and each contradicts an invariant this same PR explicitly sets out to establish — that's why they're blocking rather than nits.

## Code Quality
- Minor: the PR description says "two tests' `monkeypatch` target moved" due to the `vault.py` extraction; the actual second file is `test_v1_api_hardening.py` (not `test_v1_api_admission.py`, which has zero `load_config`/`configured_vault` references). Purely a description nit — the refactor itself and both real test updates are correct.
- The `care-escalation.json` fixture bug is honestly flagged in a code comment (`tests/test_api_reflect.py`) rather than silently ignored, which is good practice, but it names no tracked issue number — a soft miss against CLAUDE.md §6.2's "no deferred work without an issue reference" spirit (this is a plain comment, not a TODO/FIXME, so it isn't a hard violation of that section's letter).
- Docstrings throughout the new modules are extensive and explain *why*, not *what* — consistent with this repo's documentation culture, if unusually verbose even by that standard.

## CLAUDE.md Adherence
✅ Scripts-only workflow, DRY (the `vault.py` extraction is a textbook example of avoiding a third copy), commit conventions, and the stated TDD/stay-green process are all followed as far as verifiable from the diff and PR description.
❌ The event-loop-blocking issue above is a direct contradiction of an architectural invariant this codebase (and this PR specifically) treats as load-bearing, which the "no shortcuts / fix root causes" principle (§1.3) would call out even though no linter catches it.
Coverage/docstring/complexity/lint claims in the PR description (94.14% aggregate, 254 files ≥80% per-file, 12/12 gates) could not be independently re-run in this review environment; the specific `wheel.py` coverage claim was spot-checked and confirmed against the waiver list and test file.

## Testing
Strong. `test_api_journal.py`, `test_api_reflect.py`, and `test_api_wheel.py` (729/757/701 lines) show genuine AAA structure, concrete negative-path and information-leak assertions (secrets/paths asserted absent from response bodies), no skips/xfails/TODOs, and real coverage of the previously-uncovered arms. The parity harness is a valuable addition that goes beyond what most PRs of this size would build. Neither of the two defects above is caught by the existing suite — the blocking-I/O regression isn't something a functional test would surface at all (it needs a concurrency/latency-under-load test), and the `fragment_id=None` race would need a targeted unit test forcing `_resolve_fragment_id` to return `None` on a successful write; consider adding one alongside the fix.

## Documentation
`docs/api.md` is updated accurately and in step with the implementation (spot-checked against the actual route table and error taxonomy). The PR description itself functions as thorough documentation of every deliberate deviation from the source issues, which is unusually transparent and made this review substantially easier to verify.



## Verdict: CHANGES_REQUESTED

Two verified, real code defects: (1) the new journal/reflect/wheel handlers call `configured_vault(request)` — which can synchronously read and YAML-parse `creek_config.yaml` via `load_config()` — directly inside the `async def handle_*` coroutine, outside `run_in_threadpool`, in the default production deployment path (`cli.py` never passes `vault_path`); this blocks the event loop on every request and directly contradicts the non-blocking-I/O invariant this very PR's own docstrings (in `capabilities.py`, `wheel.py`) explain and rely on. (2) In `journal.py`, a successful ingest whose `fragment_id` resolves to `None` (a documented race in `journal_ingest_tool`) is rendered as the literal wire string `"fragment_id": "None"` under a `200`, not the `internal_error` the adjacent code comment claims this path guards against (`str(None)` never raises) — a real fabricated-id leak in a PR whose entire design philosophy is "never fabricate an id." Both are narrow but concrete, independently verified against the current source, and cheap to fix (thread the vault resolution through `run_in_threadpool`; check `fragment_id is None` before rendering).

Geoffe-Ga:
**CI**: 5/8 Green
**Action**: pull comment 5229021325 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**Action**: pull comment 5229021325 to see in-depth feedback and continue iterating

Geoffe-Ga:
Both fixed in the latest push. Each contradicted an invariant this PR set out to establish, which is exactly why they were worth blocking on.

**Finding 1 — the fabricated `"fragment_id": "None"`.** Your diagnosis was exact: the `try/except` claimed to guard *"most plausibly an unresolved `fragment_id`"*, but that guard **could never fire**, because `str(None)` does not raise. So a documented race — `journal_ingest_tool` returning `status: ok` with `fragment_id=None` when `_resolve_fragment_id` fails right after a successful write — produced a `200` carrying the literal string `"None"`. Precisely the fabricated id the module docstring says must never happen.

Now refused before construction, and `str(result["fragment_id"])` replaces `str(fragment_id)` so the value cannot be `None` at that point. I also corrected the misleading comment to say what the `except` **actually** catches — a missing key, or a `tier_ceiling`/`action`/`tier` value the wire enums cannot name. A comment describing a guard that never fires is its own small defect.

The test drives the **real** `journal_ingest_tool` with `ledger_for_source` patched to return `None`, which is exactly how `_resolve_fragment_id` comes back empty after a successful write — not a hand-built dict standing in for the race.

**Finding 2 — event-loop blocking.** Also right, and worse than "under concurrent load": `app.state.vault_path` is `None` in the **production default**, since `cli.py` calls `create_app(verifier=verifier)` without a vault path. So the `load_config()` fallback isn't an edge arm, it's the common one — every request to all three new endpoints did synchronous file I/O and YAML parsing on the loop.

`configured_vault` moved into each threadpool-wrapped worker, matching `capabilities.py`'s `_vault_is_usable`. The workers now take `request` instead of `vault`, call it themselves, and return `None` when it yields no vault; each handler renders that as the same 503. **Identical refusal, decided off the loop.**

I added the blocking-seam warning to `configured_vault`'s own docstring — *"Blocking. Call it from a worker thread, never from a handler coroutine"* — with the reason the `load_config` fallback is the common arm. That's what stops the next caller repeating it; the pattern existed in `capabilities.py` and these three handlers still missed it.

**The isolation proof is worth noting:** reverting `journal.py` alone reddened only its own test (`assert True is False`) while `reflect.py` and `wheel.py` stayed green — confirming each of the three tests pins its own handler rather than passing on a neighbour's fix.

Gate: **12/12**, 8495 passed, 94.15%.

Geoffe-Ga:
## Summary
This PR completes epic #1071 by implementing the last three `/v1` verticals — `PUT /v1/journal-entries/{external_id}`, `GET /v1/wheel`, and `POST /v1/reflections` — as thin adapters over already-existing MCP tools (`journal_ingest_tool`, `wheel_tool`, `reflect_tool`). It also extracts a shared `configured_vault` resolver into a new `httpapi/vault.py`, adds a cross-adapter (HTTP↔MCP) behavioural-parity test harness, and includes a second commit that fixes two real bugs found in self-review: a fabricated `fragment_id: "None"` on a specific race, and event-loop-blocking config reads on the three new routes.

## Strengths
- **Delegation discipline.** Each new handler module (`journal.py`, `reflect.py`, `wheel.py`) does no privacy/tier logic of its own — it parses, resolves the vault, calls the existing tool, and projects the result. This directly serves the epic's goal of one behavioral implementation shared between MCP and HTTP.
- **Cross-adapter parity harness** (`tests/adapter_parity.py`) compares behavioral outcomes (status, wire code, projected success fields, audit records) between the HTTP and MCP paths for the same scenarios, which is a much stronger guarantee than two independently-written test suites.
- **Privacy-oracle discipline.** `entry_ref not found` and `entry_ref tier exceeds ceiling` are deliberately collapsed to one identical `privacy_refused` message on `/v1`, refusing the issue's literal instruction to split them into `not_found`, because doing so would reopen an enumeration oracle that #846/#970/#972/#1090 spent five issues closing. This is backed by an AST-pinning test (`test_v1_api_structure.py`) that the PR notes forbids `NOT_FOUND` outside the routing layer.
- **Self-caught bugs, fixed with regression tests.** The second commit fixes (1) a `str(None)` "None" fragment-id fabrication reachable when `_resolve_fragment_id` returns empty after a successful write, and (2) `configured_vault` (a blocking file read + YAML parse) running on the event loop inside `async def handle_*` before the `run_in_threadpool` hoist — both with RED tests proving reachability before the fix.
- **Filesystem-level assertions, not just status codes.** E.g. `test_a_tier_above_the_ceiling_stages_nothing` asserts no staged file exists after a 403, not just that the response is a 403 — closing the gap where a route could stage-then-refuse and still look correct externally.
- **Honest handling of wrong acceptance criteria.** Three ACs from the source issues (#1075 AC4's `intimate` tier, #1075's "known inherited gap: #970", #1076 AC5's unreachable `F11` branch) are shown, with code-level evidence, to be incorrect or already-fixed, and the tests pin the actual (correct) behavior instead of blindly implementing the issue text.
- **"Found, not fixed" transparency.** A real mismatch in the committed `docs/contracts/adepthood-v1/examples/reflections/care-escalation.json` fixture (`CARE_SIGNAL["kind"]` vs. the guard's actual reason string) is flagged in a test comment and PR body as a follow-up rather than silently patched or ignored.
- **Coverage gap closed with real tests, no waivers**: `wheel.py` raised from 72.5% to 87.5% via genuine behavioral tests (vault-unavailable, projection-guard, tool-refusal-arm), each with a documented reason for existing.

## Security Concerns
🟢 LOW — Reviewed the trust boundaries specific to this diff and found no new vulnerabilities:
- `external_id` path-segment validation (`admissible_external_id`) bounds length, rejects U+FFFD (undecoded bytes) and non-printable characters (including a URL-encoded NUL), before the id can become a staged filename — appropriate defense against key-collision/control-byte injection into frontmatter/audit records.
- Every refusal path is asserted to never echo caller-supplied content, tier values, staged file paths, or exception messages (`error_response` renders one constant message per code); this is tested extensively across all three new handlers.
- No new subprocess, SQL, or template-rendering surface introduced.
- `configured_vault` resolves only from `app.state.vault_path` or a fixed config path — never from request-controlled input — so there is no path-traversal vector introduced here.

## Problems
None found that block merge.

## Code Quality
Nothing blocking. Two minor, non-blocking observations for the author to weigh:
- `wheel_refusal_code` unconditionally returns `INTERNAL_ERROR` regardless of `reason` (by design, since `wheel_tool` has no refusal branch today) — well-documented, but worth double-checking that a future genuine `wheel_tool` refusal won't be silently swallowed into `internal_error` without anyone noticing beyond the "a wrong-looking 500 is what sends someone to add the branch" argument in the docstring.
- The module/function docstrings across `journal.py`/`reflect.py`/`wheel.py`/`vault.py` are unusually long relative to typical style, though this appears consistent with this repo's existing convention (and its ≥95% interrogate docstring-coverage gate), so not flagged as an issue.

## CLAUDE.md Adherence
✅ Test coverage — PR reports 94.14–94.15% aggregate (branch), above the ≥90% gate; wheel.py's per-file gap was closed to 87.5% (≥80% threshold) with behavioral tests, no `# pragma: no cover` or waiver.
✅ Commit conventions — commits use `feat(api): …` / `fix(api): …`, consistent with Conventional Commits.
✅ TDD workflow — the second commit's description explicitly states RED tests were written first for both bugs found in review, matching the "Stay Green" 4-gate process.
✅ Typing/patterns — consistent use of `Final`, `Literal`, `TYPE_CHECKING`-guarded imports, `from __future__ import annotations`; no obvious strict-mypy violations visible in the diff.
❔ Docstring coverage (≥95%) and lint/type-check (zero violations) — not independently verified in this review since CI-check tooling access was unavailable in this session; the PR body's self-reported "Gate: 12/12" is the only evidence for these gates.

## Testing
Extensive and well-targeted — the standout aspect of this PR:
- Idempotency, tier-ceiling refusal (asserted on the filesystem), audit-trail content, disclosure (no echoing of caller input, paths, or internal exception details), and HTTP↔MCP behavioral parity are all covered per vertical.
- Negative-space tests are present (e.g. `test_the_parity_harness_can_actually_fail` proves the parity harness isn't vacuous; `test_a_tally_that_does_not_fit_the_contract_is_refused_not_reshaped` covers every exception type the projection `except` tuple catches, one row per type).
- The now-empty `_UNIMPLEMENTED_CAPABILITIES` set is handled correctly: rather than deleting the honesty-stub tests (which would let that guard rot), the suite substitutes a synthetic stub via `stubbed_client`/`_send_stubbed` to keep the 501 path exercised for the *next* capability.
- Care-escalation tests observe the "model was never called" invariant via a factory spy rather than inferring it from the response shape, which is the correct way to test a negative-call guarantee.

## Documentation
`docs/api.md` is updated accurately and in detail: the implemented-routes table reflects the new behavior (including the note-worthy vocabulary boundary between Creek's wheel and Adepthood's curriculum Map), the uvicorn access-log example is updated from `501` to `200`, and the capabilities-advertised list is corrected. Module-level docstrings in each new file carry the actual design rationale (why delegation is safe, why certain issue instructions were refused) rather than restating the code.



## Verdict: LGTM

No bugs, missing tests, or CLAUDE.md violations found. The PR is exceptionally thorough: it delegates all behavior to existing, already-hardened tools; adds a parity harness that cross-checks HTTP against MCP for every scenario; asserts privacy/leak invariants on the filesystem and audit trail, not just status codes; and a prior review pass already caught and fixed two real bugs (a fabricated "None" fragment_id and event-loop-blocking config reads) with dedicated regression tests. The few deviations from the source issues' acceptance criteria are deliberate, well-justified, and independently verifiable against the code (schema constraints, closed issues, AST-pinned invariants).

Geoffe-Ga:
**CI**: 4/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5229155686 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Implements the three remaining `/v1` HTTP endpoints (journal upsert, wheel, reflections) as thin adapters over existing MCP tools, closing epic #1071. Includes a large, genuinely behavioral test suite (`test_api_journal.py`, `test_api_reflect.py`, `test_api_wheel.py`, plus a new `adapter_parity.py` harness that compares HTTP and MCP outcomes for the same scenario) and a `vault.py` refactor that centralizes vault resolution. The PR's own history shows two blocking findings from a prior review round on this branch (a fabricated `"None"` `fragment_id`, and synchronous config I/O running on the event loop) — both are fixed here and independently verified as correct and well-tested.

## Strengths
- **Deliberate deviations from the source issues are justified and tested, not silently applied.** The PR refuses the issue's instruction to map `entry_ref not found` → `not_found`, keeping it collapsed into `privacy_refused` to avoid rebuilding a corpus-enumeration oracle (AST-pinned by `tests/test_v1_api_structure.py`, confirmed present at test_v1_api_structure.py:487-518). Three other acceptance criteria are shown to be wrong (schema-invalid `intimate` tier, a stale "#970 gap" claim, an unreachable F11 branch) and the tests follow the code instead of the issue text.
- **Event-loop-blocking fix is real and correctly verified.** All three new `async def handle_*` functions only call `configured_vault()` from inside their `run_in_threadpool`-wrapped worker, never directly on the coroutine (confirmed by direct read of journal.py, wheel.py, reflect.py). `test_v1_api_hardening.py` pins this with a genuine thread-identity probe (`asyncio.get_running_loop()`), not just a status-code check.
- **Fabricated-id fix is real and regression-tested through the actual tool**, not a mock: `journal.py`'s `_render` explicitly checks `fragment_id is None` before any `str()` call, and `test_a_written_entry_whose_fragment_id_will_not_resolve_is_a_fault` reproduces the race by removing the source ledger the resolver reads.
- **No path traversal via `external_id`.** `safe_stem` sanitizes to a filesystem-safe stem plus a hash suffix; `admissible_external_id` additionally rejects oversized/blank/non-printable/undecoded input before the tool is ever reached.
- **Error responses render one constant message per code** (`errors.py`'s `error_response`), so refusal reasons that could carry staged file paths or exception details never reach the wire — verified by dedicated leak tests.
- **Honesty about a found-but-unfixed defect.** The `care-escalation.json` fixture bundle uses the wrong `reason` value (`acute_distress` vs the runtime's `acute_distress_markers`); this is flagged in both a code comment and a test that asserts against the guard rather than the fixture, rather than silently working around it.
- **Genuine HTTP↔MCP parity harness** (`tests/adapter_parity.py`) compares behavioral outcomes (status, wire code, audit facts) between both adapters for the same scenario, with fragment-id aliasing to make cross-vault comparisons meaningful, and a self-test proving the harness can actually fail.

## Security Concerns
None found. 🟢 Reviewed: `entry_ref`/`external_id` handling, error-message leakage, care-escalation routing (model never called on distress or above-ceiling paths, verified via factory spies), tier-ceiling enforcement order (refused before staging), and audit-trail correctness. All check out.

## Problems
None blocking.

## Code Quality
- Minor: `test_v1_api_hardening.py`'s off-loop proof for the 3 new routes uses a thread-identity probe rather than the ordering/concurrent-progress test used for `GET /v1/capabilities` (`test_the_event_loop_serves_another_request_while_the_handshake_blocks`). Both are valid non-tautological proofs of the same invariant; not worth blocking on.
- `JournalUpsertRequest.timestamp` has no format validation at the wire layer — confirmed intentional (deferred to the tool, and covered by `test_an_unparseable_timestamp_behaves_as_the_tool_does`), not an oversight.

## CLAUDE.md Adherence
✅ No `noqa`/`type: ignore`/TODO-without-issue found in any new httpapi module. ✅ Docstrings are extensive and Google-style throughout. ✅ PR description reports 12/12 gates, 8437+ tests passing, 94.14% aggregate coverage (exceeds the ≥90% floor) and the previously-failing `wheel.py` per-file coverage closed from 72.5% to 87.5% with real tests rather than a waiver or pragma. (CI status could not be independently queried in this review session — `gh pr checks` was blocked by the sandbox — but the self-reported gate numbers are consistent with the depth of tests actually present in the diff.)

## Testing
Extensive and substantively behavioral, not shallow: exact-value assertions on wheel counts/shares (not `> 0` checks), filesystem-state assertions on refused writes (not just status codes), ontology-vocabulary leak sweeps, a 5-row malformed-tool-result matrix, care-escalation-without-model-call proven via spy rather than inference, and the cross-adapter parity harness. Verified independently by re-deriving several of the claimed test behaviors directly against the checked-out source.

## Documentation
`docs/api.md` is updated accurately to reflect all four capabilities now being implemented, including the deliberate `not_found`-avoidance rationale and the wheel/curriculum vocabulary boundary. Module docstrings in the new `httpapi/*.py` files are unusually thorough and consistent with the actual code.



## Verdict: LGTM

No blocking bugs, security issues, or CLAUDE.md violations found after direct inspection of every new/changed source file plus targeted verification by an independent subagent pass. The two previously-flagged blocking findings (fabricated "None" fragment_id, event-loop-blocking config reads) are both correctly fixed in this same PR and pinned by genuine regression tests (real-tool race reproduction, thread-identity probes via asyncio.get_running_loop()). Test suites are large but substantively behavioral (exact-value assertions, filesystem/audit assertions, HTTP↔MCP parity harness), not tautological.

Geoffe-Ga:
**CI**: 5/8 Green
**VERDICT**: LGTM
**Action**: pull comment 5229424232 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 11/11 Green
**VERDICT**: LGTM

PR #1247, opened 2026-08-09
fix(mcp): stop reporting a partial erasure as ok, and derive the payload from PurgeResult

### LINKED ISSUE #1246: fix(mcp): creek.purge reports status ok for a purge the engine recorded as partial
## Summary

`creek_mcp/tools/purge.py::_result_payload` hardcodes `"status": "ok"` and
forwards only a fixed subset of `PurgeResult`. An MCP caller therefore receives
`"status": "ok"` for a purge whose own `purge.jsonl` audit line the engine just
wrote as `status="partial"`.

Found in review of #1243 (PR #1243), which added the partial signal.

## Why this matters

An erasure that fell short must say so **where the operator is already looking**.
The MCP surface is exactly such a place — CrawDad drives purges through it — and
today it is the one surface that reports unqualified success.

The CLI says so, and the audit log says so. The MCP caller does not.

## The missing fields

Not a regression from #1243; the same payload already omits three others:

| Field | Added by | In payload? |
|---|---|---|
| `voice_body_undecodable` | #1211 | ❌ |
| `voice_artifacts_removed` | #1211 | ❌ |
| `journal_staged_removed` | #845 / #1023 | ❌ |
| `provenance_scrubbed` | — | ❌ |
| `intimate_stubs_removed` | — | ❌ |

`voice_body_undecodable` is the sharpest: it is a **non-empty list naming the
fragments whose body could not be decoded**, i.e. the ones whose derived voice
artifacts were *not* swept. A caller that cannot see it cannot know the erasure
was incomplete.

## The design question this needs answered

Forwarding the fields is additive and safe. Changing `status` is **not** — it is
a contract change:

1. Does `status` gain a `partial` member, or does the caller infer incompleteness
   from the fields? A new enum member may break existing clients that branch on
   `ok`/`refused`; check `creek_mcp/api/models.py` and the committed contract
   bundle before deciding, and bump `CONTRACT_VERSION` if the wire shape moves.
2. If `status` stays binary, is an incomplete erasure `ok` or `refused`? Neither
   fits, which is the argument for `partial`.
3. Should the payload mirror `PurgeResult` wholesale rather than a hand-picked
   subset? The subset is exactly the drift mechanism that lost these five fields
   — deriving it from the model would stop the next one going missing.

Option 3 is the one I'd look at first: a hand-maintained field list is the same
defect class as the hand-copied directory list #1025 removed.

## Acceptance criteria

- [ ] A purge whose engine result is partial does not report unqualified `ok`
      over MCP.
- [ ] `voice_body_undecodable` and `voice_artifacts_removed` reach the caller.
- [ ] The other three omitted fields are forwarded, or their omission is
      documented with a reason.
- [ ] The payload's field set is derived from `PurgeResult`, or a test fails when
      a new `PurgeResult` field is not forwarded.
- [ ] Any wire-shape change carries a `CONTRACT_VERSION` bump and a bundle
      regeneration.

## Related

- #1211 / PR #1243 — added the partial signal this issue surfaces
- #845 — the staged-plaintext RTBF sweep whose counter is also unforwarded
- #1025 — the hand-copied-list drift defect this payload repeats

Geoffe-Ga:
Closes #1246.

`_result_payload` hardcoded `"status": "ok"` and forwarded a hand-picked subset of `PurgeResult`. An MCP caller therefore received **unqualified success for a purge the engine's own audit line recorded as `status="partial"`**. The CLI said so; the audit log said so; MCP was the one surface that did not.

## The issue undercounts its own finding

It names **five** unforwarded fields. There are **six** — `embeddings_removed` is missing from the issue's table as well as from the payload. `PurgeResult` has 15 fields; the subset forwarded 11.

The issue demonstrates the failure mode it describes, by hand-maintaining its own list of hand-maintained-list casualties.

**So the fix is the derivation, not the six fields.** The payload is now built from `PurgeResult` itself, and a test fails when any new field does not reach the caller. The six were a symptom of the mechanism.

## The wire change, done properly

`status` gains `partial` → **CONTRACT_VERSION 0.3.0 → 0.4.0**, with `SUPPORTED_CONTRACT_MINORS` widened so both 0.3 and 0.2 clients keep negotiating.

Two new guards make future bumps self-enforcing:
- every `CONTRACT_VERSION` must earn a change-log row
- the compatibility window may **only ever widen**

Both were written *first* and confirmed green against the un-bumped tree, then the bump alone turned them red — along with four pre-existing tripwires (manifest pin, committed-bundle equality, and the two ADR version assertions). All six satisfied together.

## Three corrections to the brief I was given

- **The bundle regeneration is nearly a no-op for purge.** `docs/contracts/adepthood-v1` contains **zero** purge content — `grep -rc purge` returns 0 everywhere but one incidental `ErrorEnvelope` hit. It renders entirely from the four `/v1` capabilities. It still needs regenerating because its manifest and two examples embed `CONTRACT_VERSION` (3 files, 9 insertions), but there is no purge schema to regenerate because none was ever published.
- **`api/models.py` does not define the purge status.** It publishes `OK_STATUS` for the journal-upsert and wheel *envelopes* only; the purge payload is a bare `dict[str, Any]` with no Pydantic model and no published schema. Adding `partial` required no model edit — `models.py` changed only to widen the minors.
- **The ADR the tests pin is the 2026-07-31 HTTP one**, not the 2026-06-30 MCP contract ADR. Both restate the version; only one is enforced. I updated both — and their change-log tables sort in **opposite directions**, which I got wrong on the first insert and corrected.

Also worth recording: the ADR states *"purge is out of the Adepthood surface"*, and `grep -rn 'creek.purge' crawdad/` returns nothing — CrawDad is not a purge caller and deliberately holds no elevated token. So no `/v1` route or schema is affected by this.

## A note on how RED was produced

**Not `git stash`.** The stash ref is shared across all five concurrently-running worktrees, so stashing here would have reached into a sibling lane. The two production files were restored from `git show HEAD:` via a scratchpad backup instead.

```
FAILED test_a_partial_erasure_is_not_reported_as_ok        assert 'ok' == 'partial'
FAILED test_the_mcp_status_agrees_with_the_audit_outcome   assert 'ok' == 'partial'
FAILED test_the_payload_names_the_fragments_...            KeyError: 'voice_body_undecodable'
FAILED test_every_purge_result_field_reaches_the_caller    extra: voice_artifacts_removed,
                                                           voice_body_undecodable, intimate_stubs_removed,
                                                           embeddings_removed, journal_staged_removed,
                                                           provenance_scrubbed
FAILED test_the_previously_dropped_counters_reach_the_caller
                                                           embeddings_removed never reaches the MCP caller
6 failed, 26 passed  ->  32 passed with the fix
```

Gate: **12/12**, 8595 passed. Zero suppressions.


Geoffe-Ga:
## Summary
Fixes #1246: the MCP `creek.purge.*` tools were hardcoding `"status": "ok"` and forwarding a hand-picked subset of `PurgeResult`, so a purge the engine's own audit line recorded as `status="partial"` was reported to MCP callers as unqualified success. The fix introduces a single `PurgeResult.outcome_status` property shared by the audit log and the MCP payload, derives the payload from `result.model_dump()` instead of an enumerated field list, and bumps `CONTRACT_VERSION` 0.3.0 → 0.4.0 with the compatibility window widened (not shifted).

## Strengths
- **Fixes the mechanism, not just the symptom.** The PR correctly identifies that hand-enumerating fields in `_result_payload` was the actual defect (it had already silently dropped 6 of 15 fields), and replaces it with `**result.model_dump(mode="json")`, so a future field addition can't recur.
- **Single source of truth for the complete/partial distinction.** `PurgeResult.outcome_status` is now used by both `_run_audited`'s audit line and the MCP payload, closing the exact drift the two independently-maintained predicates had caused.
- **Self-enforcing version-bump guards.** Two new tests (`test_every_contract_version_earns_a_change_log_row`, `test_the_compatibility_window_only_ever_widens`) make future contract bumps fail loudly if the changelog or the compatibility window regress — a good structural investment beyond the immediate fix.
- **Backward compatibility handled correctly.** `SUPPORTED_CONTRACT_MINORS` widens to include 0.4/0.3/0.2; the `_ENGINE_STATUS_TO_WIRE` mapping deliberately keeps `ok` for `complete` so existing clients are unaffected, and an unrecognized `partial` fails closed for old clients rather than being silently read as success.
- Verified independently: the manifest's SHA-256 hashes for the two regenerated example JSON files match what's actually in the diff, and `PurgeResult.criteria`'s stored values are plain JSON-serializable primitives, so `model_dump(mode="json")` can't raise.
- Documentation is updated consistently across `docs/api.md`, `docs/mcp.md`, `docs/cleaning-and-purge.md`, and both affected ADRs (change-log tables correctly maintain their pre-existing, opposite sort orders — verified by inspection).

## Security Concerns
🟢 LOW — None found. The `partial` status doesn't leak new information (fragment ids in `voice_body_undecodable` were already written to `purge.jsonl` and printed by the CLI); this only closes a gap where MCP was the one surface hiding the same fact.

## Problems
None blocking.

## Code Quality
- `_ENGINE_STATUS_TO_WIRE` is typed as `dict[PurgeOutcomeStatus, str]`, and `PurgeOutcomeStatus = Literal["complete", "partial"]` has exactly two members, so the dict lookup can't KeyError — good, just noting it was worth checking given the wire has three possible statuses (`refused` is handled separately via `_refusal`, never through this map).
- Minor: `_audit_success` is called identically regardless of whether the result is `complete` or `partial` (pre-existing behavior, not touched by this PR) — the `mcp.jsonl` audit entry itself doesn't record the complete/partial distinction, only `purge.jsonl` and the returned payload do. Not a regression, but a caller reconstructing history purely from `mcp.jsonl` still can't see it. Worth a follow-up issue if that matters, not this PR's job to fix.

## CLAUDE.md Adherence
✅ TDD workflow described in the PR body (RED confirmed before the fix, six new failing assertions listed).
✅ Tests are thorough and specific (field-set tripwire, named dropped fields, status-agreement test, complete-case regression test).
✅ Documentation updated in the same PR as the behavior change (`docs/api.md`, `docs/mcp.md`, `docs/cleaning-and-purge.md`, both ADRs).
✅ Commit message follows Conventional Commits style with a `fix(mcp):` prefix.
⚠️ Could not independently execute `./scripts/check-all.sh` or `pytest` in this review environment (sandboxed) — relying on the PR's reported gate result (12/12, 8595 passed) and static verification of the diff's internal consistency (hashes, type flows, Literal exhaustiveness).

## Testing
Strong. `tests/test_mcp_purge.py` adds targeted cases: partial-not-reported-as-ok, MCP/audit status agreement, fragment ids surfaced, complete-case regression (`ok` spelling preserved), a derived-field-set tripwire, and the six specifically-named previously-dropped fields. `tests/test_adepthood_contract_models.py` adds the two self-enforcing version-bump guards. `test_v1_api_auth.py`/`test_v1_api_not_implemented.py` extended to cover the new `0.4` minor without dropping prior coverage of `0.3`/`0.2`.

## Documentation
Complete and consistent. Both ADRs' change-log tables gained a `0.4.0` row (correctly respecting each table's pre-existing sort order — one descending, one ascending), `docs/api.md` and `docs/mcp.md` describe the new `partial` status and the full-field-forwarding guarantee, and `docs/cleaning-and-purge.md` ties the CLI/audit/MCP surfaces back to the single `outcome_status` definition.



## Verdict: LGTM

The fix is correct, minimal in scope, and unusually well-verified: the outcome_status property centralizes the complete/partial predicate shared by the audit log and MCP payload, the payload derivation via model_dump eliminates the hand-picked-subset drift bug at its root, the contract version bump follows the established widen-only pattern with two new self-enforcing guard tests, and I independently recomputed the manifest SHA-256 hashes and confirmed they match. No blocking issues found.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5229665556 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1248, opened 2026-08-09
fix(ingest): resolve a document's author to the Authorship axis, and keep the name

### LINKED ISSUE #1229: fix(ingest): DocumentIngestor puts a free-text DOCX/PDF author into the Authorship enum — every real Word document fails to ingest

Geoffe-Ga:
Closes #1229.

Every real Word document failed to ingest. `DocumentIngestor` copied a document's free-text author onto `Fragment.source.author`, which is the `Authorship` StrEnum (`self`|`ai`|`other`|`collaborative`), so any file carrying an author failed Pydantic validation and was dropped. It predates #1023 — `creek ingest --type document` fails identically on main — but it made `creek.upload`'s headline use case non-functional.

## Two decisions, and I declined the issue's steer on the second

**Which enum member** — a non-empty extracted author resolves to `Authorship.OTHER`, not `SELF`. Two independent reasons:

1. **Precedent.** `creek/clean/authorship.py::_tag_self_platform` already applies exactly this rule — *"Explicit author `<name>` in metadata"* → `other`. Ingest now agrees with the cleaning pass instead of contradicting it.
2. **Fail-closed.** `SELF` is the key that unlocks the INTIMATE privacy tier and voice/skill generation. Guessing `SELF` for a stranger's document feeds both from material that isn't the owner's; guessing `OTHER` for the owner's own document merely under-uses it — recoverable, and the preserved name is what makes it recoverable.

**Where the name goes** — a new inert `FragmentSource.author_name`, **not** `author_slug`.

The issue suggests routing it into the `AuthorManifest`/`author_slug` machinery. I did not, and this is the load-bearing call: setting `author_slug` from a Word "Created by" field would, via `creek/vault/writer.py:728` and `_apply_other_author_attribution`, **relocate the fragment out of `01-Fragments/` into `11-Other-Authors/<name>/` and zero its `voice_weight`** from a manifest that does not exist. That infers an entire identity-folder commitment from a metadata string — for the "upload my own documents" case.

`author_name` is inert: nothing routes, classifies, or weights voice on it. It records the fact without deciding anything on the owner's behalf, and it is exactly the input needed to promote it to a real `author_slug` deliberately later. Pinned by an assertion that an uploaded `.docx` still lands in `01-Fragments/Documents`.

Two edge rules fall out:
- An author string that already names an `Authorship` member (case-insensitive) is honoured as the axis and yields no name — it is a classification, not somebody's name. Without this, a correct `author='ai'` from another caller would be clobbered to `other` with a bogus `author_name: 'ai'`.
- A blank author writes nothing, leaving the model's `self` default. Absence of a name is not evidence about anybody.

## The issue's scope claim was wrong in one direction and understated in the other

**FALSE: "PDF, RTF and HTML share the same `_extract_*` metadata path."** `_extract_metadata` branches only on `.docx` and `.pdf` — there is no RTF or HTML branch, so `metadata['author']` is never populated for them. Verified by running both through the ingestor: an RTF with `{\info{\author Ada Lovelace}}` and an HTML with `<meta name="author">` both yield `author: None`. A regression test pins the `self` default for both anyway, so a future author-extraction branch cannot reintroduce the collision unnoticed.

**CONFIRMED and strengthened for PDF.** The repo's only PDF-metadata test *mocks* the extractor — *"Mock pdfminer since we can't easily create real PDFs in tests"*. A mocked extractor could never have caught this bug. I hand-assembled a genuine minimal PDF with an `/Info` dict and confirmed the real defect, and the new test builds that real PDF rather than mocking it.

## An existing test was pinning the bug

`test_frontmatter_includes_author_from_metadata` asserted `fm['source']['author'] == 'Jane Doe'` — i.e. it asserted the exact frontmatter that fails validation downstream. Rewritten.

And `tests/test_mcp_upload.py::_docx_bytes`, which had been *clearing* `core_properties.author` with a docstring explaining this very bug, now stamps a real author and asserts `source.author == 'other'` / `source.author_name == 'Dana Rivers'`. That test now pins the fix instead of routing around it.

## Where the reasoning lives

Three places rather than an ADR — the `author_name` field comment (why the three author slots are distinct), the `_resolve_document_author` docstring (the axis/fail-closed argument with citations), and a new "Document attribution (the name on the file)" section in `docs/ingestion.md`, sited next to the existing "AI-chat attribution (per turn)" section where a reader already goes for this question. An ADR would duplicate the docstrings (§1.2 DRY).

### DIFF
--- a/creek-tools/creek/ingest/documents.py
+++ b/creek-tools/creek/ingest/documents.py
@@ -17,6 +17,7 @@ Exports:
     _detect_scanned_pdf: Check if a PDF is likely scanned (image-only).
     _extract_docx_metadata: Extract metadata from DOCX bytes.
     _extract_pdf_metadata: Extract metadata from PDF bytes.
+    _resolve_document_author: Split extracted author metadata into axis + name.
     _infer_document_platform: Map file extension to SourcePlatform.
 """
 
@@ -40,7 +41,7 @@ from creek.ingest.base import (
     parse_authored_at,
 )
 from creek.ingest.html import extract_html_authored_at, parse_html_to_markdown
-from creek.models import SourcePlatform
+from creek.models import Authorship, SourcePlatform
 
 logger = logging.getLogger(__name__)
 
@@ -56,6 +57,11 @@ _SCANNED_TEXT_THRESHOLD = 100
 
 _DOCUMENT_PLATFORM_EXTENSIONS: frozenset[str] = frozenset({".docx", ".pdf", ".rtf"})
 """Extensions that map to the DOCUMENT source platform."""
+
+_AUTHORSHIP_BY_VALUE: dict[str, Authorship] = {
+    member.value: member for member in Authorship
+}
+"""Lookup from an ``Authorship`` member's wire value to the member itself."""
 
 
 # ---- Helper Functions ----
@@ -490,6 +496,53 @@ def _extract_pdf_metadata(pdf_bytes: bytes) -> dict[str, Any]:
     return _extract_pdf_metadata_from_bytes(pdf_bytes)
 
 
+def _resolve_document_author(
+    raw_author: object,
+) -> tuple[Authorship, str | None] | None:
+    """Split a document's author metadata into an authorship axis and a name.
+
+    ``FragmentSource.author`` answers "whose views does this stand for?" on the
+    ``self|ai|other|collaborative`` axis. A DOCX ``core_properties.author`` or a
+    PDF ``/Author`` answers the different question "what name is on the file".
+    Copying the second into the first was issue #1229: every document Word saves
+    carries an author, so every such document failed Pydantic validation and was
+    dropped at assembly.
+
+    Returns ``None`` when there is no usable value — absent, non-string, or
+    blank — leaving the model's ``self`` default in force. "We know nothing" is
+    not evidence about anybody.
+
+    A value that already names an ``Authorship`` member is honoured as the axis
+    and yields no name; it is a classification, not somebody's name. Any other
+    string resolves the axis to :attr:`Authorship.OTHER` and is returned as the
+    name. ``OTHER`` is both the convention this codebase already applies to an
+    explicit author name
+    (:meth:`creek.clean.authorship.AuthorshipTagger._tag_self_platform`) and the
+    fail-closed choice: ``SELF`` is what unlocks the INTIMATE privacy tier
+    (:mod:`creek.classify.privacy`) and voice/skill generation
+    (:mod:`creek.generate.skills`), so guessing ``SELF`` for a stranger's
+    document would feed both from material that is not the owner's, whereas
+    guessing ``OTHER`` for the owner's own document merely under-uses it — and
+    the returned name is what makes that recoverable.
+
+    Args:
+        raw_author: The ``author`` value an ``_extract_*`` pass put in metadata.
+
+    Returns:
+        ``(authorship, name)`` where ``name`` is ``None`` for a value that was
+        already an axis member, or ``None`` when there is nothing to record.
+    """
+    if not isinstance(raw_author, str):
+        return None
+    name = raw_author.strip()
+    if not name:
+        return None
+    axis = _AUTHORSHIP_BY_VALUE.get(name.lower())
+    if axis is not None:
+        return axis, None
+    return Authorship.OTHER, name
+
+
 def _infer_document_platform(extension: str) -> SourcePlatform:
     """Map a file extension to a SourcePlatform value.
 
@@ -794,8 +847,11 @@ class DocumentIngestor(Ingestor):
         """Generate Creek-compatible YAML frontmatter for a document fragment.
 
         Builds frontmatter with type, title, source (platform, original_file,
-        original_encoding), created timestamp, and optional author/scanned
-        fields from metadata.
+        original_encoding), created timestamp, and optional scanned flag.
+
+        An extracted document author is split by :func:`_resolve_document_author`
+        into ``source.author`` (the ``Authorship`` axis) and ``source.author_name``
+        (the free-text name), never conflated into one slot (#1229).
 
         Args:
             fragment: The parsed fragment with metadata.
@@ -814,8 +870,12 @@ class DocumentIngestor(Ingestor):
         }
 
         # Add optional metadata fields
-        if "author" in fragment.metadata:
-            source["author"] = fragment.metadata["author"]
+        resolved_author = _resolve_document_author(fragment.metadata.get("author"))
+        if resolved_author is not None:
+            authorship, author_name = resolved_author
+            source["author"] = authorship
+            if author_name is not None:
+                source["author_name"] = author_name
         if fragment.metadata.get("scanned"):
             source["scanned"] = True
 
--- a/creek-tools/creek/models.py
+++ b/creek-tools/creek/models.py
@@ -646,6 +646,17 @@ class FragmentSource(BaseModel):
     channel: str | None = None
     interlocutor: str | None = None
     author: Authorship = Authorship.SELF
+    # Issue #1229: the free-text name a source document carries — a DOCX
+    # ``core_properties.author``, a PDF ``/Author``. It answers "what name is on
+    # the file", which is a *different question* from ``author`` ("whose views
+    # does this stand for?") and from ``author_slug`` (which identity folder
+    # governs it). Keeping the three apart is the whole point: writing a name
+    # into ``author`` fails enum validation, and writing one into ``author_slug``
+    # would silently relocate the fragment into ``11-Other-Authors/<name>/`` and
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
Fixes issue #1229: `DocumentIngestor` was copying a document's free-text author (DOCX `core_properties.author`, PDF `/Author`) directly onto `FragmentSource.author`, which is the `self|ai|other|collaborative` `Authorship` enum — so any real Word/PDF document with an author set failed Pydantic validation and was silently dropped. The fix adds a new inert `FragmentSource.author_name` field to hold the free-text name, resolves the `author` axis to `Authorship.OTHER` (fail-closed, matching existing `clean/authorship.py::_tag_self_platform` precedent) whenever a non-enum name is present, and leaves `author_slug` untouched so fragments aren't silently relocated into `11-Other-Authors/`.

## Strengths
- **Root-cause fix, not a patch.** `_resolve_document_author` cleanly separates three distinct questions (whose views? what name? which identity folder?) instead of overloading one field, and the docstring/inline model comment explain *why* each of the three stays separate — directly citing the exact downstream code paths (`writer.py:728`, `_apply_other_author_attribution`, `privacy.py`, `generate/skills.py`) that would misfire if `author_slug` were set instead. I independently verified those citations against the current source and they check out.
- **Tests exercise real files, not mocks.** The new tests build a genuine `.docx` via `python-docx` and hand-assemble a real minimal PDF with an `/Info` dict, specifically because the existing PDF-metadata test mocked the extractor and could never have caught this class of bug. This is a meaningfully stronger regression barrier than the pattern it replaces.
- **Correctly narrowed the issue's own scope claim.** The PR verified (and I independently confirmed via `_extract_metadata`, documents.py:764) that RTF and HTML never populate `author` at all — only `.docx` and `.pdf` do — and added regression tests pinning that so a future author-extraction branch for those formats can't reintroduce the collision unnoticed.
- **Fixed a test that was pinning the bug.** `test_frontmatter_includes_author_from_metadata` previously asserted the exact broken frontmatter (`author == 'Jane Doe'`); it's now rewritten to assert against the correct behavior. Likewise `test_mcp_upload.py::_docx_bytes` no longer clears the author to route around the defect.
- **Documentation lands in the logical spot** — a new "Document attribution" section in `docs/ingestion.md` next to the existing per-turn AI-chat attribution section, rather than a redundant ADR.
- Edge cases are deliberately and correctly handled: an author string that's already a valid `Authorship` member (case-insensitive) is honored as the axis with no name written (protects a legitimate `author='ai'` caller from being clobbered); a blank/whitespace-only author writes nothing, preserving the `self` default.

## Security Concerns
None. 🟢 No new I/O, path handling, or trust boundary is introduced; `author_name` is a plain optional string field with no routing/classification logic hung off it, unlike `author_slug` (which already has its own path-traversal validator, untouched by this PR).

## Problems
None found blocking merge.

## Code Quality
- Minor: `_resolve_document_author` returns `Authorship.OTHER` for *any* non-enum, non-blank string, including surprising inputs (e.g., a document literally authored by someone named "Self" would collide with an enum member and silently be treated as an axis classification rather than a name). This is called out explicitly in both the function docstring and `docs/ingestion.md` as a deliberate, documented tradeoff — not something I'd ask to change, just noting it's a real (if narrow) edge case future readers should know about.
- The `FragmentSource.author_name` field comment in `models.py` is quite long (10 lines) relative to the codebase's usual comment density, but it's justified given the subtlety it's preventing (three similarly-named but semantically distinct author fields) and mirrors the equally-thorough `_resolve_document_author` docstring — consistent rather than gratuitous.

## CLAUDE.md Adherence
✅ TDD workflow evident (tests rewritten/added alongside the fix, PR claims "Gate: 12/12" with zero suppressions — no `# noqa`/`# type: ignore` introduced by this diff).
✅ DRY (§1.2): reasoning lives in one place (docstring + doc section) rather than a duplicative ADR, with an explicit rationale for that choice.
✅ No shortcuts: the fix addresses the root cause (field conflation) rather than working around symptoms; the old test-fixture workaround (clearing `core_properties.author`) was removed along with the bug it was routing around.
✅ Conventional commit style (`fix(ingest): ...`) with a detailed body.
Coverage/lint/type-check results aren't independently re-run in this review (no local environment available), but the diff itself introduces no obvious mypy/ruff violations — all new code is fully typed, docstrings follow Google style, and the new tests are thorough.

## Testing
✅ Strong. New `TestDocumentAuthorMetadata` class covers: real DOCX ingestion with author → `OTHER`/name pinned end-to-end through `assemble_ingested_fragment` (i.e., through actual Pydantic validation, not just `generate_frontmatter`); real PDF equivalent; RTF/HTML regression (no author extracted, `self` default preserved); axis-collision edge case (`author='ai'` honored, no spurious `author_name`); blank-author edge case. `test_mcp_upload.py` also updated end-to-end to assert the fragment lands in `01-Fragments/Documents` (not relocated) with `author == 'other'` and the name preserved. This is exactly the "hand-assembled real file, not a mock" bar the PR sets for itself, and it clears it.

## Documentation
✅ `docs/ingestion.md` gets a new, well-placed section explaining the three-field split and the RTF/HTML non-coverage, matching the code precisely. Module docstring's function-index list and the `generate_frontmatter` docstring are both updated to reflect the new behavior.



## Verdict: LGTM

The fix is correct, minimal, and well-verified: it splits the free-text document author from the `Authorship` axis into a new inert `author_name` field, matches the existing `_tag_self_platform` precedent for the `other` classification, and every claim in the PR description (RTF/HTML never populate `author`, `author_slug` would relocate fragments via `_apply_other_author_attribution`, the mocked PDF test) checks out against the actual source. Tests use a real hand-built PDF and a real python-docx file rather than mocks, cover the axis-collision edge case and the blank-author case, and no other call site depends on the new field in a way that could break. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5229694818 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1249, opened 2026-08-09
feat(classify): free, non-destructive tags backfill for an already-classified vault

### LINKED ISSUE #1207: feat(classify): free, non-destructive tags backfill for an already-classified vault

Geoffe-Ga:
Closes #1207.

A free, non-destructive tags backfill for an already-classified vault: a bare `creek classify` now gives llm/manual-stamped fragments their tags through a narrow writer, without re-classifying anything and without a single provider call.

## Took the steer: a sibling writer, not a widened one

`_write_tier_only`'s #876 promise — *"changes ONLY the privacy-tier fields"* — is verbatim unchanged. The tags axis went to a sibling `_write_tags_only`, and that docstring now records **why**: a writer whose field set is legible from its name survives review of a path that rewrites already-curated user vaults; one that takes its scope as an argument pushes that question out to every call site.

## A fork the steer did not anticipate — and it re-opened a closed hole

**Two narrow writers can fire on one file.** An untiered, llm-stamped fragment carrying a hashtag qualifies for both. Each writer rewrites the whole file from the frontmatter it is handed, so the second must compose on what the first actually left on disk.

The naive version — tags writing from the frontmatter as originally *read* — **silently reverts the tier and re-opens #876's cloud-egress hole.** So `_write_tier_only`/`_persist_tier_only` now return the metadata they wrote, and it is threaded to the tags writer.

That is not a widening: the returned dict is still exactly `raw` plus the two tier fields. Proven necessary by an explicit mutation check and pinned by `test_preserved_tags_backfill_keeps_the_tier_written_on_the_same_run`.

`_rewrite_frontmatter` is extracted as the shared I/O tail of both writers and is deliberately **decision-free** — it decides nothing about which fields may change, so sharing serialisation cannot blur the two writers' scopes.

## Three corrections to the issue

- **The cost claim is overstated.** *"`--method llm --force` re-sends all 35,330 fragments to the provider"* — it does not. On the LLM path, `_classify_one` runs the rule classifier **first** and returns without any provider call whenever confidence clears the threshold; only the below-threshold remainder is sent. The argument (paying tokens to recover data already on local disk) stands; the count does not.
- **Extrapolation sold as proof.** *"The field is provably `[]` on 100% of 35,330 fragments"* — the evidence recorded in #878 is **2000/2000 sampled**, not a census. Immaterial to the decision, since `tags_pass.merge` is a union and the backfill is safe either way, but "provably 100%" is not what was measured.
- **Stale line ref**, ~88 lines: `_classify_one` is cited at `classify_engine.py:1211`; it was at 1299. It also does not `return rules.classify(...)` — it returns a 3-tuple. The destructiveness claim itself is correct, and `rules.py:574` is the one reference that was accurate.

## The free command is not the one you would guess

`creek classify --method llm` without `--force` makes zero LLM calls on an all-llm-stamped vault — but `_assert_classifiers_available` raises `LLMProviderUnavailableError` **before** the loop, so it is unusable for an operator with no API key.

The genuinely free, provider-free route is the **bare `creek classify`** (default `--method rules`, no `--force`): every llm/manual fragment short-circuits and gains tags through the narrow writer, and nothing is re-classified. The `creek fill` hint now names that command; it previously named the paid `--method llm --force` and quoted its token cost, which this change makes wrong.

## One reversal, argued rather than silent

`test_preserved_fragment_gets_tags_only_under_force` asserted #878's residual — the exact behaviour #1207 exists to change. Leaving it would have made the issue unimplementable. Renamed to `test_preserved_fragment_gains_tags_without_force`, with the reversal argued in its docstring.

Preserved-path backfills count on `ClassifySummary.tags_extracted` — counting the decision, not the bytes, consistent with `privacy_tiers_assigned` and `praxis_marked`. That attribute's docstring explicitly said preserved fragments are *not* counted; corrected.

## Not done, deliberately

No `creek fill` step and no new command. The issue offered that shape; the sibling writer plus the corrected hint delivers every acceptance criterion with a far smaller review surface, and a bare `creek classify` is already the free entry point.

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -132,15 +132,27 @@ process`` path, and a second call there would be pure duplication.
   (or the ``low-priority`` literal
   :mod:`creek.clean.context` appends). That makes the pass idempotent: a
   second run over an unchanged vault reports ``0``.
-* **NOT on the preserved short-circuit**, for the same reason as praxis
-  and more so — :func:`_write_tier_only` is not widened, and ``tags`` is
-  a field operators curate by hand. An ``llm``/``manual``-stamped vault
-  backfills it only under an explicit ``creek classify --method llm
-  --force``; ``creek fill`` surfaces the outstanding count
+* **On the preserved short-circuit too, since #1207** — the one place
+  this axis parts company with praxis. #878 originally declined the
+  exception, which left the operator's vault (35,330 fragments, every one
+  of them stamped ``classification_method: llm``) able to gain tags only
+  through a paid ``creek classify --method llm --force``, to recover data
+  already sitting in the fragment bodies on local disk. Two properties
+  make the free pass safe here where the praxis pass is not: hashtag
+  extraction is a lossless mechanical restatement of literal
+  author-written syntax rather than a judgment, so re-deriving it cannot
+  disagree with operator curation; and the merge is a union, so it cannot
+  drop a tag a hand ever wrote. The backfill goes through the **sibling**
+  narrow writer :func:`_write_tags_only` — :func:`_write_tier_only`'s
+  "changes ONLY the privacy-tier fields" promise is untouched, because a
+  writer whose field set is legible from its name survives review of a
+  path that rewrites already-curated user vaults, and one that takes its
+  scope as an argument does not. ``creek fill`` still surfaces the
+  outstanding count, now naming the free command
   (:func:`creek.cli._hint_tags_backfill`).
 
 :attr:`ClassifySummary.tags_extracted` reports how many fragments the run
-actually gained a tag on.
+actually gained a tag on, on both paths.
 
 Known residual (not fixable by ordering)
 ----------------------------------------
@@ -206,7 +218,7 @@ from creek.classify.reatomize import (
     classify_reatomize,
 )
 from creek.classify.rules import RuleClassifier
-from creek.classify.tags_pass import apply_tags
+from creek.classify.tags_pass import TAGS_KEY, apply_tags
 from creek.classify.weighted import classify_weighted
 from creek.ingest.base import IngestedFragment
 from creek.models import Fragment, Frequency, PrivacyTier
@@ -339,9 +351,13 @@ class ClassifySummary:
             genuine additions: re-deriving a hashtag the fragment already
             records is not work the operator needs reported, so the merge
             being a union makes a second run over an unchanged vault
-            report ``0``. Fragments the resume short-circuit preserved are
-            **not** counted — they are not given tags at all without
-            ``--force`` (see the module docstring).
+            report ``0``. Fragments the resume short-circuit preserved
+            **are** counted since #1207: the free backfill runs on them
+            too, through a narrow tags-only write that touches no other
+            frontmatter field (see the module docstring). Like
+            :attr:`privacy_tiers_assigned` this counts the decision, not
+            the bytes — a fragment whose write then fails is counted here
+            *and* reported on :attr:`errors`.
     """
 
     total: int
@@ -892,17 +908,18 @@ def _prepare_fragment(
     # Skip fragments whose previous run already settled the classification.
     # Tracked on distinct counters so the CLI summary can label "manual
     # preserved" and "previously LLM-classified preserved" honestly
-    # (issue #321). A tier assigned above still lands, through the narrow
-    # writer that touches nothing else.
+    # (issue #321). The free, non-destructive passes — the tier assigned
+    # above (#876) and the hashtag backfill (#1207) — still land, each
+    # through a narrow writer that touches nothing else.
     if not force and _record_if_preserved(raw, counts):
-        if owns_tier:
-            _persist_tier_only(
-                md_file=md_file,
-                fragment=fragment,
-                body=body,
-                raw=raw,
-                counts=counts,
-            )
+        _backfill_preserved(
+            md_file=md_file,
+            fragment=fragment,
+            body=body,
+            raw=raw,
+            owns_tier=owns_tier,
+            counts=counts,
+        )
         return None
     return _PreparedFragment(
         fragment=fragment,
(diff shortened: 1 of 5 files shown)


Geoffe-Ga:
## Summary
Closes #1207: extends the existing narrow-writer pattern (from #876's tier backfill) so that a bare `creek classify` also backfills hashtag-derived `tags` on preserved (`llm`/`manual`-stamped) fragments — no `--force`, no provider call. Introduces a sibling `_write_tags_only` writer alongside `_write_tier_only`, a shared `_rewrite_frontmatter` I/O tail, and a `_backfill_preserved` orchestrator that threads the tier-writer's actual on-disk result into the tags writer so the two narrow writers compose correctly when both fire on the same file.

## Strengths
- Correctly identifies and fixes a real composition hazard: two narrow writers firing on one preserved file (untiered + hashtag-carrying fragment) could have clobbered each other if the tags writer worked from the frontmatter as *read* rather than as *written*. `_write_tier_only` now returns the metadata it wrote, and `_backfill_preserved` threads it forward — explicitly tested in `test_preserved_tags_backfill_keeps_the_tier_written_on_the_same_run`.
- `_write_tier_only`'s #876 "changes ONLY the privacy-tier fields" contract is preserved verbatim by adding a sibling writer rather than widening it into a parameterized one — a deliberate, well-reasoned design choice recorded in multiple docstrings.
- Excellent test coverage of the exact properties that matter for a writer running over already-curated production vaults: byte-identical frontmatter elsewhere (`test_preserved_tags_backfill_touches_no_other_frontmatter_field`), idempotency to the byte, zero LLM calls (`test_preserved_tags_backfill_makes_no_llm_call`), never-drop-a-hand-tag (`test_preserved_tags_backfill_never_drops_a_recorded_tag`), and non-fatal per-file write failure handling.
- Documentation (module docstring, `tags_pass.py`, CLI hint docstrings) is updated everywhere the old asymmetry was previously described, including the intentionally-unchanged praxis pass which now explicitly notes it's the one axis that still requires `--force` — no stale cross-references found.
- The PR description itself corrects prior overstated/inaccurate claims from the source issue (call counts, "provably 100%" vs. a 2000/2000 sample, a stale line reference) rather than repeating them — good intellectual honesty.

## Security Concerns
None identified. 🟢 The change is careful about the one security-relevant invariant in this area (never silently reverting a freshly-assigned privacy tier) and adds a regression test for exactly that composition case. The tags writer is scoped to a single key, consistent with the existing tier writer's narrow, auditable pattern.

## Problems
None found that block merge.

## Code Quality
- Minor prose nit: the module docstring's closing sentence "actually gained a tag on, on both paths" reads a little awkwardly (repeated "on"); not worth blocking on.
- `counts.tags_extracted` is incremented in `_backfill_preserved` before the write is attempted (mirroring the pre-existing `privacy_tiers_assigned` pattern), so a failed write is still counted as an "extraction" and separately reported via `errors`. This is a pre-existing, documented design convention (the updated `ClassifySummary.tags_extracted` docstring now says so explicitly: "counts the decision, not the bytes"), so it's consistent rather than a new defect — just flagging it as something a reader should know to look for in the docstring rather than infer.

## CLAUDE.md Adherence
✅ Extensive Google-style docstrings on all new/changed functions (`_backfill_preserved`, `_persist_tier_only`, `_write_tier_only`, `_persist_tags_only`, `_write_tags_only`, `_rewrite_frontmatter`) with Args/Returns/Raises sections.
✅ Tests written alongside the implementation, covering the new behavior thoroughly, TDD-consistent with the repo's workflow.
✅ No forbidden shortcuts (no `noqa`/`type: ignore` additions, no threshold changes) visible in the diff.
⚠️ I was unable to execute `./scripts/check-all.sh` (ruff, mypy, coverage, complexity, bandit) in this review environment due to sandbox restrictions on running the project's Python toolchain — I could not independently verify local CI-equivalent green status. The diff reads as clean, well-typed, and consistent with the codebase's existing style, but this is a manual read, not a tool-verified one.

## Testing
✅ Not blocking. New tests are comprehensive and target the properties that actually matter for this change (composition of two writers, byte-level frontmatter isolation, idempotency, no-LLM-call guarantee, error non-fatality, tag preservation ordering). The CLI hint test correctly asserts the *absence* of the old paid-remedy wording, not just the presence of the new one.

## Documentation
✅ Thorough. Module docstring, `tags_pass.py` module docstring, and CLI hint docstrings (`_hint_tags_backfill`) are all updated in lockstep with the behavior change, and old "Known residual" bullets describing the now-obsolete `--force` requirement were removed rather than left stale.



## Verdict: LGTM

The change correctly extends the existing narrow-writer pattern, explicitly guards against the one real hazard it introduces (two narrow writers composing on one file, which could silently revert a freshly-assigned privacy tier) with both a design fix and a dedicated regression test, and updates all documentation consistently with no stale references found. Test coverage is thorough and targets the properties that actually matter for a writer running over production vaults. I could not execute the local quality gates (ruff/mypy/coverage) in this sandboxed review environment, but the diff itself shows no correctness, security, or documentation gaps that would block merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5229747476 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1250, opened 2026-08-09
wire-in: activate the draft grounding guard on both surfaces

### LINKED ISSUE #1040: wire-in: the draft grounding guard is dormant on every path, so `creek lint --check draft-grounding` passes vacuously
**Labels:** `de-slop`, `creek-pipeline`, `wire-in`, `priority-high`
**Detected by:** de-slopify targeted scan (creek-pipeline) 2026-07-31, at HEAD `0f7689a`
**Severity:** High
**Remediation:** wire-in (+ e2e test)

## Problem

The FEAT-032 draft grounding guard — derivative scoring, biographical-sentence scanning, and the grounding veto — is **dormant on every production path**, and the lint check written to audit it **passes vacuously** as a direct consequence. Three docstrings assert the guard runs.

Taxonomy family: **3 (Orphaned code — complete implementation with zero inbound production edges)** compounded by **3 (Fake/aspirational feature flag)**: `creek lint --check draft-grounding` reports "clean" for every draft ever produced, because the frontmatter keys it inspects are only ever written by the guard that never runs.

**Current state:** `creek draft` and MCP `creek.draft` both silently skip the guard; the audit gate that should catch that reports success.

## Evidence (corroboration)

**Signal 1 — structural, zero production callers of the wiring factory.** `creek/generate/grounding.py:717` defines `default_embedding_fn(...)`, whose stated purpose is to wire the guard into production. Repo-wide (all file types, incl. `creek_mcp/` and `crawdad/`):

```
$ rg -rn "default_embedding_fn" .
creek-tools/creek/generate/grounding.py:717        <- its own definition
creek-tools/creek/generate/grounding.py:~655       <- a docstring cross-reference
creek-tools/tests/test_draft_grounding.py:824      <- test only
```

**Signal 2 — call-site proof: neither production constructor passes the kwargs.** `DraftGenerator.__init__` defaults `embedding_fn=None` and `grounding_thresholds=None` (`creek/generate/drafts.py:773-774`). Both production construction sites omit them:

- `creek/cli.py:4275` (`creek draft`) — passes `llm, skills_root, voice_core, style_preamble, privacy_override, bypass_compiled, ontology_twist, fingerprint, ai_style_config, voice_guard_no_llm, cohesion`. No `embedding_fn`, no `grounding_thresholds`.
- `creek_mcp/tools/draft.py:263` (MCP `creek.draft`) — passes only `llm, skills_root, privacy_override`.

Every guard entry point then short-circuits on exactly that condition:

```
creek/generate/drafts.py:1186:  if self._embedding_fn is None or self._grounding_thresholds is None:
creek/generate/drafts.py:1381:  if self._embedding_fn is None or self._grounding_thresholds is None:
creek/generate/drafts.py:2095:  if self._embedding_fn is None or self._grounding_thresholds is None:
```

**Signal 3 — the audit gate passes vacuously.** `creek/lint/checks/draft_grounding.py:64-70`:

```python
derivative = metadata.get("derivative_score")
grounding = metadata.get("grounding_score")
if not isinstance(derivative, (int, float)) or not isinstance(grounding, (int, float)):
    return None          # <- treated as CLEAN
```

Those keys are written only by the dormant guard, so every real draft takes the `return None` path and the check reports no findings — regardless of how derivative or ungrounded the draft is.

**The claims this contradicts:**

- `creek/generate/grounding.py:21-23` — "The `GroundingReport.summary_line` is printed to stderr by ``creek draft``."
- `creek/generate/grounding.py:654-655` — "production wires this to :func:`default_embedding_fn`."
- `creek/lint/checks/draft_grounding.py:3-5` — "The `creek.generate.grounding` guard **runs synchronously during ``creek draft``** and stamps each saved draft's frontmatter."

## Why wire-in rather than delete

Both taxonomy gates for the wire-in carve-out are met:

- **Intent evidence** — three docstrings above, a dedicated `default_embedding_fn` factory that exists for no other purpose, a `GroundingThresholds` config block, and a shipped lint check built to consume the output.
- **Completeness evidence** — `creek/generate/grounding.py` is a finished, typed, docstringed, house-consistent implementation with its own test module (`tests/test_draft_grounding.py`), including a test that exercises `default_embedding_fn` directly.

This is finished work awaiting the connection it was written for — not a husk.

## Scope

Covers wiring `embedding_fn` + `grounding_thresholds` into the two `DraftGenerator` production construction sites and proving the path end-to-end. It does **not** cover: `creek.reflect`'s separate grounding-cost problem (#1034), the `ai_style` guard's distinct `voice_distance_*` thresholds, or any change to the scoring algorithm itself.

## Tasks

1. In `creek/cli.py:4275`, pass `embedding_fn=default_embedding_fn(config.embeddings)` and `grounding_thresholds=GroundingThresholds.from_config(config.draft)` (or the equivalent constructor the config exposes).
2. Do the same at `creek_mcp/tools/draft.py:263` so the MCP surface is not left behind.
3. Ensure the guard degrades gracefully when the embeddings model is unavailable (mirror the existing `EmbeddingModelUnavailableError` handling in `creek/author/agents.py:309`) — a missing local model must not crash `creek draft`.
4. Add an **e2e test** driving the real `creek draft` command via Typer's `CliRunner` that asserts `derivative_score` and `grounding_score` are present in the saved draft's frontmatter.
5. Add a test asserting `creek lint --check draft-grounding` reports a finding for a draft whose scores are out of bounds (i.e. the check is no longer vacuous).
6. Correct `creek/generate/grounding.py:21-23` if any part of the stderr claim remains inaccurate after wiring.

## Acceptance Criteria

- [ ] A draft produced by `creek draft` carries `derivative_score` and `grounding_score` in its frontmatter.
- [ ] The same holds for a draft produced through the MCP `creek.draft` tool.
- [ ] `creek lint --check draft-grounding` produces a finding for an out-of-bounds draft and none for an in-bounds one — both asserted by tests.
- [ ] `creek draft` still succeeds (with the guard skipped and a warning) when the embeddings model is unavailable.
- [ ] The new e2e test fails on `main` before the change and passes after.
- [ ] All pre-commit hooks pass (`cd creek-tools && ./scripts/check-all.sh`).

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek_mcp/tools/draft.py` | Modify |
| `creek-tools/creek/generate/grounding.py` | Modify (docstring accuracy) |
| `creek-tools/tests/e2e/test_draft_grounding_e2e.py` | Create |
| `creek-tools/tests/test_lint.py` | Modify |

Geoffe-Ga:
Closes #1040. Refs #1024.

The draft grounding guard was dormant on every path, so `creek lint --check draft-grounding` ran, exited 0, and verified nothing. This wires it at both surfaces and pins the chain.

## The issue's own acceptance criterion does not hold

It asks that the new e2e test "fail on main before the change". **For the in-bounds control case, it doesn't** — a test asserting only that the lint report says `0 outside` passes against the dormant guard, because **"judged clean" and "found nothing to judge" render identically**.

Making it genuinely red required asserting the saved draft actually *carries* `derivative_score`/`grounding_score`. That is the #1024 failure mode reappearing one level up, **inside the test written to catch it** — which is worth recording for the rest of that epic.

## Task 5 is based on a false premise

It asks for "a test asserting `creek lint --check draft-grounding` reports a finding for an out-of-bounds draft (i.e. the check is no longer vacuous)" and names `tests/test_lint.py`.

That test **already existed** — `tests/test_lint_draft_grounding.py` writes out-of-bounds frontmatter by hand and asserts findings at lines 145 and 160. (`test_lint.py` contains zero draft-grounding tests.) The check's scoring logic was never vacuous *as a function*; it was vacuous *in production* because nothing ever wrote the frontmatter it reads.

The genuinely missing test is **the chain** — a real `creek draft` producing frontmatter a real `creek lint` then judges. That is what the new e2e module supplies. `tests/test_lint.py` was not modified.

## Two test modules, not one

The issue asks for `tests/e2e/test_draft_grounding_e2e.py`; that exists. I also added `tests/test_draft_grounding_wiring.py` **without** the marker, because the `e2e` marker is deselected from the default unit lane — putting the wiring assertions only there would leave the newly-wired production lines uncovered by the gate that runs on every commit. **The same shape of hole this issue is about.** No assertion is duplicated between them.

## Degradation lives in one place, deliberately

`default_embedding_fn` loads the sentence-transformer **lazily**, so "the model is missing" surfaces mid-draft, not at construction — neither the CLI nor MCP can catch it where they build the generator. So degradation lives in `DraftGenerator._disable_grounding`, which clears both attributes and makes the failure **sticky**: later guard entry points fall through their existing `is None` checks instead of re-raising once per pass. Mirrors `creek/author/agents.py:309`'s handling of the same typed error, as the issue directed.

## A branch I wrote and then deleted

I initially added a try/except to `_voice_grounding_check`'s grounding veto. It is **unreachable** with an unloadable model, so I removed it rather than ship an untestable branch, and documented the invariant in place.

Proof: `_apply_voice_fidelity` has exactly two callers — `save_outline_draft`, which passes `source_fragments=()` and exits at the existing empty-sources guard; and `save_draft`, which only runs after `generate_draft`/`generate_seeded_draft` have already called `_build_guard_report`. A load failure there clears `_embedding_fn`, so the `is None` check returns before the closure is built.

## Two smaller calls worth stating

**The MCP tool resolves the vault's own config**, via `load_config(resolve_config_path(vault_path, None), warn_on_missing=False)` rather than the bare `load_config()` that `tools/link.py` and `classify.py` use. An operator who calibrated `draft.derivative_upper` in their vault should get those thresholds on the MCP surface too; the bare call would silently ignore them.

**The test stub replaces `EmbeddingLinker.load_model`**, not `default_embedding_fn`. That keeps the entire production path — factory → linker → `generate_embedding` → `score_draft` — under test. Patching the factory would have proved only that *a* callable was passed, not that the real one works.

Two details that keep the stub honest: its vocabulary deliberately excludes NATO `hotel` and `xray`, which collide with `delta` and `uniform` under `crc32 % 128` and shifted the expected derivative score from 0.500 to 0.559 — the exact-value assertions are what keep a future collision loud. And the cohesion-degradation test uses a body containing "I grew up", because `scan_biographical_sentences` returns before embedding anything when no sentence matches, so a neutral body never reaches the model and the test would have passed for the wrong reason. Verified by coverage that the cohesion except branch is now executed.

Line numbers were stale as expected: `cli.py:4275` → 4605, `tools/draft.py:263` → 330.

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -4573,6 +4573,7 @@ def draft(
     from creek.generate.ai_style.fingerprint import load_fingerprint
     from creek.generate.ai_style.preamble import build_style_preamble
     from creek.generate.drafts import DraftGenerator
+    from creek.generate.grounding import GroundingThresholds, default_embedding_fn
     from creek.generate.mining import IdeaMiner
 
     vault_path = _resolve_vault(vault)
@@ -4610,6 +4611,13 @@ def draft(
         privacy_override=override,
         bypass_compiled=bypass_compiled,
         ontology_twist=ontology_twist,
+        # FEAT-032 grounding guard. Both kwargs are required together — the
+        # generator skips the guard entirely if either is missing, which is
+        # exactly how it stayed dormant on this path until #1040. The model
+        # loads lazily inside the factory, so an offline host pays nothing
+        # here and the generator degrades with a warning when it does.
+        embedding_fn=default_embedding_fn(vault_config.embeddings),
+        grounding_thresholds=GroundingThresholds.from_config(vault_config.draft),
         fingerprint=fingerprint,
         ai_style_config=ai_style,
         voice_guard_no_llm=no_llm,
--- a/creek-tools/creek/generate/drafts.py
+++ b/creek-tools/creek/generate/drafts.py
@@ -84,6 +84,7 @@ from creek.generate.twist import (
     target_profile_from_spec,
     twist_dimensions,
 )
+from creek.link.embeddings import EmbeddingModelUnavailableError
 from creek.models import (
     CompiledPage,
     Eddy,
@@ -1181,31 +1182,61 @@ class DraftGenerator:
 
         Returns:
             A :class:`GroundingReport`, or ``None`` when the guard is
-            not configured on this generator.
+            not configured on this generator — or when the embedding
+            model turned out to be unloadable at scoring time.
         """
         if self._embedding_fn is None or self._grounding_thresholds is None:
             return None
         source_texts = [
             fragments[fid][1] for fid in source_fragments if fid in fragments
         ]
-        report = score_draft(
-            body,
-            source_texts=source_texts,
-            embedding_fn=self._embedding_fn,
-            thresholds=self._grounding_thresholds,
-        )
-        print(report.summary_line(), file=sys.stderr)
-        # The non-``None`` invariant is established by the guard above; assert
-        # it so ``_warn_ungrounded_biographical``'s non-optional
-        # ``embedding_fn`` contract is explicit and mypy's narrowing is pinned.
-        assert self._embedding_fn is not None
-        self._warn_ungrounded_biographical(
-            body=body,
-            source_texts=source_texts,
-            embedding_fn=self._embedding_fn,
-            thresholds=self._grounding_thresholds,
-        )
+        try:
+            report = score_draft(
+                body,
+                source_texts=source_texts,
+                embedding_fn=self._embedding_fn,
+                thresholds=self._grounding_thresholds,
+            )
+            print(report.summary_line(), file=sys.stderr)
+            # The non-``None`` invariant is established by the guard above;
+            # assert it so ``_warn_ungrounded_biographical``'s non-optional
+            # ``embedding_fn`` contract is explicit and mypy's narrowing is
+            # pinned.
+            assert self._embedding_fn is not None
+            self._warn_ungrounded_biographical(
+                body=body,
+                source_texts=source_texts,
+                embedding_fn=self._embedding_fn,
+                thresholds=self._grounding_thresholds,
+            )
+        except EmbeddingModelUnavailableError as exc:
+            self._disable_grounding(exc)
+            return None
         return report
+
+    def _disable_grounding(self, exc: EmbeddingModelUnavailableError) -> None:
+        """Turn the grounding guard off for the rest of this run, loudly.
+
+        The production ``embedding_fn`` loads a sentence-transformer on
+        first call, so "the model is missing" surfaces mid-draft rather
+        than at construction time. Losing the guard must not lose the
+        draft: the generator drops back to its dormant behaviour (no
+        scores stamped, no veto applied) and says so on stderr, mirroring
+        :meth:`creek.author.agents.RetrievalAgent.retrieve`'s
+        degrade-to-empty handling of the same typed error.
+
+        Clearing both attributes also makes the failure sticky, so the
+        later guard entry points short-circuit on their existing
+        ``is None`` checks instead of re-raising the same load failure
+        once per pass.
+
+        Args:
+            exc: The typed load failure, whose message already carries
+                the model name and the operator's remediation.
+        """
+        self._embedding_fn = None
+        self._grounding_thresholds = None
+        print(f"grounding guard skipped: {exc}", file=sys.stderr)
 
     def _warn_ungrounded_biographical(
         self,
@@ -1376,19 +1407,24 @@ class DraftGenerator:
 
         Returns:
             One finding per ungrounded biographical sentence; empty when the
-            guard is dormant or the body is clean.
+            guard is dormant, the embedding model is unloadable, or the body
+            is clean.
         """
         if self._embedding_fn is None or self._grounding_thresholds is None:
             return []
         corpus = [fragments[fid][1] for fid in source_fragments if fid in fragments]
         if self.voice_core.strip():
             corpus.append(self.voice_core)
-        return scan_biographical_sentences(
-            body,
-            source_texts=corpus,
-            embedding_fn=self._embedding_fn,
-            threshold=self._grounding_thresholds.grounding_lower,
-        )
+        try:
+            return scan_biographical_sentences(
+                body,
+                source_texts=corpus,
+                embedding_fn=self._embedding_fn,
+                threshold=self._grounding_thresholds.grounding_lower,
+            )
+        except EmbeddingModelUnavailableError as exc:
+            self._disable_grounding(exc)
+            return []
 
     def _cohesion_llm(self, prompt: str) -> str:
         """Call the shared draft LLM for the cohesion hop, returning the body.
@@ -2105,6 +2141,16 @@ class DraftGenerator:
         thresholds = self._grounding_thresholds
 
         def _still_grounded(body: str) -> bool:
+            # No ``EmbeddingModelUnavailableError`` handling here, unlike the
+            # two guard entry points that embed before this one. The veto is
+            # built only from ``_apply_voice_fidelity``, which ``save_draft``
+            # calls after ``generate_draft`` / ``generate_seeded_draft`` have
+            # already run ``_build_guard_report``; the other caller,
+            # ``save_outline_draft``, passes no source fragments and exits
+            # above. So an unloadable model has always already tripped
+            # ``_disable_grounding``, and the ``_embedding_fn is None`` check
+            # above returns before this closure is ever built. Catching it
+            # here again would be a branch nothing can reach.
             report = score_draft(
                 body,
                 source_texts=source_texts,
--- a/creek-tools/creek/generate/grounding.py
+++ b/creek-tools/creek/generate/grounding.py
@@ -25,6 +25,15 @@ Surfaces:
   per-paragraph annotations list under the keys exported below — these
   are also what the ``draft-grounding`` lint check reads.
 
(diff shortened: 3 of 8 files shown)


Geoffe-Ga:
## Summary
Wires the FEAT-032 grounding guard (dormant since #1040) into the two production draft surfaces — `creek draft` (CLI) and the MCP `creek.draft` tool — and makes the guard degrade gracefully instead of crashing when the embedding model can't load. Adds a real end-to-end test proving `creek draft` → `creek lint --check draft-grounding` actually catches an ungrounded draft, plus a wiring-assertion module that runs on every commit.

## Strengths
- Traced every claim in the PR description against the actual code and they all hold: `generate_draft`/`generate_seeded_draft` unconditionally call `_build_guard_report` before `save_draft`/`save_outline_draft` can reach `_apply_voice_fidelity`, so the deliberately-uncaught `EmbeddingModelUnavailableError` path in `_voice_grounding_check`'s `_still_grounded` closure (`creek/generate/drafts.py:2143`) is genuinely unreachable — the comment there is accurate, not aspirational.
- `_disable_grounding` (drafts.py:1217) correctly mirrors the existing `RetrievalAgent.retrieve` degrade-to-empty pattern at `creek/author/agents.py:309` — same typed exception, same "log and continue" shape.
- The new e2e test (`tests/e2e/test_draft_grounding_e2e.py`) asserts on frontmatter *and* the lint report, not just exit codes — correctly avoiding the "0 outside ≠ nothing to judge" trap the PR body calls out, which is exactly the class of bug #1040 was.
- `BagOfWordsEncoder` (tests/grounding_stubs.py) patches `EmbeddingLinker.load_model` rather than the `default_embedding_fn` factory, keeping the real production path (factory → linker → `generate_embedding` → `score_draft`) under test rather than just proving a callable was passed.
- Both new production call sites (`cli.py:4619`, `creek_mcp/tools/draft.py:346`) supply `embedding_fn`/`grounding_thresholds` as a pair, matching `DraftGenerator`'s "skip guard unless both are present" contract — verified by reading the constructor.
- MCP tool's `load_config(resolve_config_path(...), warn_on_missing=False)` (vs. the bare `load_config()` in `link.py`/`classify.py`) is a deliberate, correctly-justified deviation so vault-calibrated thresholds apply — and the unguarded call matches the existing precedent in those sibling tool modules, so it isn't a new gap.

## Security Concerns
None found. 🟢 No new I/O, subprocess, or path-handling surfaces; the config load path matches existing MCP tool patterns.

## Problems
None blocking.

## Code Quality
- Minor style inconsistency (non-blocking): `drafts.py` imports `EmbeddingModelUnavailableError` from `creek.link.embeddings` directly, while `cli.py` and `agents.py` import it via the `creek.link` package re-export. Both resolve to the same class, so this is cosmetic only.

## CLAUDE.md Adherence
✅ TDD shape: new/changed behavior (guard wiring, degrade-on-failure) is backed by targeted tests before the "shipped" claim.
✅ No shortcuts observed: no `# noqa`/`# type: ignore` added; the PR explicitly documents a branch (try/except in `_still_grounded`) that was written and then *removed* because it was unreachable and untestable — exactly the "no shortcuts" ethos in §1.3.
✅ Single source of truth: reused `tests/helpers.py::write_raw_fragment_file` rather than duplicating fixture logic; `_disable_grounding` centralizes the degrade behavior in one place rather than repeating it at each guard entry point.
✅ e2e test correctly marked `pytest.mark.e2e` and confirmed to run in the blocking `integration-e2e` CI lane (`.github/workflows/ci.yml:392-458`, gated via `quality-gate`).
Coverage/lint/type-check results weren't independently re-run in this review (sandboxed environment blocked shell execution of `./scripts/test.sh`), so CI's own gate is the final confirmation, per Gate 3 of the workflow.

## Testing
Strong. Every new/changed branch has a dedicated test:
- `_build_guard_report`'s except path: `test_cli_draft_survives_an_unavailable_embedding_model` (non-cohesion path, confirmed `_apply_cohesion` is opt-in via `--cohesion` so this test isolates the guard-report branch specifically).
- `_cohesion_grounding_check`'s except path: `test_cohesion_pass_survives_an_unavailable_embedding_model`.
- MCP wiring: `test_mcp_draft_stamps_grounding_scores`.
- Exact-value assertions (0.5 / 1.0 / 0.0 cosine scores) guard against a stub that always returns zero passing for the wrong reason — good mutation-resistant testing per the repo's `mutation-testing` skill philosophy.
- The e2e test's two cases (ungrounded flagged, in-bounds cleared) are the right minimal pair to prove discrimination, not just "the pipeline runs."

## Documentation
Thorough. Every new/changed function has an accurate docstring; the `grounding.py` module docstring and `draft_grounding.py` check docstring were both updated to explain the two populations that land in the "no scores" skip branch. The PR description itself is unusually rigorous — it states the acceptance criterion's own flaw, explains a rejected implementation path, and cites exact line numbers, all of which I verified against the diff and found accurate.



## Verdict: LGTM

The wiring is correct end-to-end (verified against actual call sites in cli.py/creek_mcp), the new degrade-on-load-failure path is exercised by dedicated tests for every reachable branch, and the one branch deliberately left without a try/except (_voice_grounding_check's closure) is provably unreachable given the caller graph I traced (generate_draft/generate_seeded_draft always run _build_guard_report before save_draft can call _apply_voice_fidelity). No blocking issues found.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5229772621 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1251, opened 2026-08-09
fix(deps): declare anyio, and catch the next undeclared import automatically

### LINKED ISSUE #1123: fix(deps): declare anyio as a direct dependency of creek-tools
`creek_mcp/httpapi/middleware/limits.py:32` imports `anyio` directly, but `anyio` appears only
transitively in `uv.lock` — it is not in `pyproject.toml`.

This contradicts the direct-dependency policy stated in `pyproject.toml`'s own neighbouring
`starlette` comment, which is exactly the drift this epic exists to stop. A transitive bump that
drops or moves `anyio` would break the `/v1` limits middleware with no lockfile signal.

Fix: add `anyio` to `[project.dependencies]` with a rationale comment, and `uv lock`.

---

Geoffe-Ga:
Closes #1123.

`creek_mcp/httpapi/middleware/limits.py:32` is `from anyio import Semaphore, WouldBlock, fail_after` — a direct import of a package the project never declares. It works today only because starlette pulls anyio in transitively, which is exactly the dependency that breaks silently on an unrelated upgrade.

**The issue's premise is correct in every checkable particular**, which is unusual enough here to state explicitly: the line number is not stale, `grep -n anyio pyproject.toml` exits 1, `uv.lock` resolves anyio 4.13.0 transitively, and `tests/test_dependency_pins.py` carries the helper pattern it points at.

## The generic guard, and the trap it had to avoid

A test now fails when any **import-time** third-party package is undeclared, so the next one is caught automatically rather than by review.

**The scope rule is import-time, not "any import".** Imports at module scope — including inside module-level `if TYPE_CHECKING:` / `try:` / `with:` — are hard requirements: the interpreter runs them or the module does not load. Function-local imports are exempt, because that is exactly how this codebase expresses genuinely optional packages.

That distinction is load-bearing, and here is why:

**`creek/clean/semantic_dedup.py:201` does `import faiss`, and faiss is declared nowhere** — not in pyproject, not in uv.lock. **This is not a second instance of the defect.** The import is function-local, sits behind a memoised availability probe (`_faiss_available`), and the code degrades to a dense matmul when it is missing. A naive "every imported package must be declared" guard would have flagged it and pressured someone into declaring a heavyweight optional dependency they do not want. The guard is scoped precisely so faiss stays exempt, and the module docstring records why.

**Module→distribution mapping reads installed metadata** (`importlib.metadata.packages_distributions`) rather than a hardcoded alias table. A table would need hand-editing for every new package — `frontmatter`→`python-frontmatter`, `yaml`→`PyYAML`, `PIL`→`pillow` — which defeats the whole point of "the next one is caught automatically". It falls back to the canonicalised module name when nothing in the environment claims it, so a partially-installed checkout reports something actionable instead of crashing.

A declaration in **any** `optional-dependencies` extra counts. Several import-time modules legitimately live in extras (anthropic, openai, numpy, sentence_transformers, docx, pptx), and `[dev]` pulls `[all]`, so they are present at collection time.

## This is the whole defect, not a sample

An AST sweep of all **19** third-party import-time roots across `creek/` and `creek_mcp/` resolved every one to a declared distribution **except anyio**.

## A docstring that would have shipped as a lie

`creek_mcp/httpapi/capabilities.py` argued for `starlette.concurrency.run_in_threadpool` over `anyio.to_thread.run_sync` **on the grounds that anyio was undeclared**. This fix falsifies that premise.

The choice itself is still right — `run_in_threadpool` *is* `anyio.to_thread.run_sync`, and going through Starlette's helper inherits its thread-limiter policy — so I replaced the reason, not the code, and recorded that the dependency argument no longer applies.

## The floor

`anyio>=4.13.0`, taken from what `uv.lock` already resolves rather than invented, matching the neighbouring uvicorn comment's precedent ("Floor matches the uv.lock resolution"). Documented as a floor and never a ceiling: anyio is also starlette's own concurrency backend, so a ceiling could deadlock resolution against the ASGI stack.

Declared in `[project.dependencies]`, **not** `[tool.uv].constraint-dependencies` — DEP-003 reserves constraints for transitive-only packages, and a constraint tightens the resolution of a package already in the graph but cannot put one there, so it can never satisfy a direct import.

`uv.lock` regenerated.


Geoffe-Ga:
## Summary
Declares `anyio` as a direct dependency (it was previously imported at module scope in `creek_mcp/httpapi/middleware/limits.py` and worked only because `starlette` pulls it in transitively), floors it at the version `uv.lock` already resolves, corrects a docstring in `capabilities.py` whose reasoning that fix falsifies, and adds a generalized `test_every_import_time_module_is_declared` AST-based guard so the next undeclared import fails a test instead of surviving to review. Closes #1123.

## Strengths
- The core fix is precise: `anyio` is added exactly where the existing dependency-comment convention lives in `pyproject.toml`, with a floor traced to `uv.lock`'s actual resolution rather than an invented lower bound, and correctly placed in `[project.dependencies]` rather than `[tool.uv].constraint-dependencies` (constraints can't satisfy a direct import per DEP-003 — I confirmed that reservation is real in `pyproject.toml`'s `constraint-dependencies` comment).
- The new `test_every_import_time_module_is_declared` guard is a real generalization, not just a regression test for this one case. Its import-time/function-local scope distinction is exactly right for this codebase: I walked `creek/` and `creek_mcp/` and confirmed the only module-level `try:` block is `creek/audit/log.py`'s stdlib `fcntl` import (exempt via `sys.stdlib_module_names`, so no false positive), there are no module-level conditional third-party imports outside `if TYPE_CHECKING:`, and `creek/clean/semantic_dedup.py`'s `import faiss` is indeed function-local behind a memoised probe — so the guard's design decisions hold up against the actual tree, not just the PR's narrative.
- Reading distribution ownership from `importlib.metadata.packages_distributions()` instead of a hand-maintained alias table is the right call — it's exactly the kind of table (`frontmatter`→`python-frontmatter`, `yaml`→`PyYAML`) that silently rots.
- The `capabilities.py` docstring edit doesn't just delete the stale reasoning, it replaces it with the correct one (the two calls are equivalent; using Starlette's own helper inherits its thread-limiter policy) and explicitly notes the old premise no longer applies — good docstring hygiene for a comment that would otherwise have shipped as a lie.
- PR description is unusually rigorous about self-checking its own premise (line numbers, `grep -n anyio pyproject.toml` exit code, lock resolution) before proposing the fix.

## Security Concerns
None. 🟢 This closes a latent supply-chain fragility (an import relying on an undeclared transitive dependency, which could silently break on an unrelated upstream change) rather than introducing one.

## Problems
None found that block merge.

## Code Quality
- Minor, non-blocking: the guard treats *any* module-level `try:`/`if:` block as a hard import-time requirement, including `try: import x except ImportError:` fallback patterns that do have a legitimate escape hatch at module scope (unlike function-local imports, which the docstring correctly identifies as the *only* escape hatch — that specific claim is slightly overstated). This doesn't cause any problem today since the tree has no such pattern for a third-party package, but if one is added later the guard would force it into `[project.dependencies]` even though it's genuinely optional. Worth a one-line docstring caveat if it ever comes up, not worth blocking on now.
- Minor: `_providing_distributions`'s fallback to the canonicalised module name (when nothing installed claims it) only works correctly if the test always runs with `[all]`/`[dev]` extras installed, which is the sanctioned workflow (`dev-setup.sh` on `uv sync --all-extras`) — so not a practical issue, just worth knowing if someone ever runs a stripped-down env.

## CLAUDE.md Adherence
✅ Uses `uv.lock` as canonical (regenerated and diff shown), floor-not-ceiling reasoning matches the neighbouring `uvicorn` precedent, constraint-vs-dependency placement matches DEP-003. ✅ No suppressions, no shortcuts. ✅ Conventional commit (`fix(deps): ...`) referencing the closed issue. ✅ Extensive Google-style docstrings on every new function, consistent with the ≥95% interrogate requirement. Could not run `./scripts/check-all.sh` in this review sandbox to confirm the stated "12/12" gate pass; CI (`Tests & Type Checking`, `Pylint`, `Static Analysis`) was still in progress at review time, though `Integration & E2E`, `Code Complexity`, and `Build Distribution` had already succeeded.

## Testing
Strong. Four new tests cover: the dependency is declared, the floor is non-empty and matches the locked resolution (both directions — floor too low and floor admits an unresolved version are separately asserted), the lock isn't stale relative to the declared floor, and the generalized import-time sweep across all of `creek`/`creek_mcp`. The assertion messages are specific and actionable (name the offending file, suggest the fix). No test-quality concerns.

## Documentation
Docstrings are thorough throughout, including the meta-documentation at the top of `test_dependency_pins.py` explaining the import-time/function-local distinction and why it matters. The `capabilities.py` docstring fix is itself a documentation correctness fix — good catch, folding it into the same PR rather than leaving stale reasoning behind.



## Verdict: LGTM

The fix is correct and narrowly scoped (declares the genuinely-undeclared `anyio` direct import, floors it at the exact `uv.lock`-resolved version, and updates the one docstring whose premise it falsifies), and the new regression guard is sound: I independently verified against the actual `creek`/`creek_mcp` trees that the only module-level `try:` block imports the stdlib `fcntl` (exempt), there are no module-level conditional third-party imports outside `TYPE_CHECKING`, and the `faiss` optional-dependency pattern the PR cites is indeed function-local and correctly stays exempt. No blocking issues found.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5229794232 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Fixes #1123: `anyio` was imported at module scope in `creek_mcp/httpapi/middleware/limits.py` but never declared as a direct dependency, working only because `starlette` happened to pull it in transitively. The PR declares `anyio>=4.13.0` (floor taken from the existing `uv.lock` resolution), corrects a docstring in `capabilities.py` whose justification for avoiding `anyio` directly is now obsolete, and adds a generalized regression test (`test_every_import_time_module_is_declared`) that walks the AST of `creek/` and `creek_mcp/` to catch any future undeclared import-time dependency.

## Strengths
- The generalization is genuinely well-designed: it distinguishes *import-time* (module scope, including inside `if`/`try`/`with`, since these run unconditionally or the module fails to load) from *function-local* imports (legitimately optional, e.g. `faiss` in `creek/clean/semantic_dedup.py:201`, which sits behind a memoized availability probe and degrades gracefully). I independently verified this against the actual `faiss` code and confirmed the AST walker in `_import_time_statements` does not descend into function/class bodies, so it's correctly exempted.
- Uses `importlib.metadata.packages_distributions()` for module→distribution resolution instead of a hand-maintained alias table — avoids the exact kind of drift (`frontmatter`→`python-frontmatter`, `yaml`→`PyYAML`) that would otherwise need manual upkeep each time a new package is added.
- Correctly treats any `optional-dependencies` extra as "declared," matching how `numpy`, `anthropic`, etc. are already structured in this repo, and correctly excludes `[tool.uv].constraint-dependencies` (a constraint can't originate a direct import — consistent with DEP-003).
- The `capabilities.py` docstring is fixed in place rather than left stale — a documented rationale that would otherwise have silently become a lie.
- Floor (not ceiling) on `anyio`, with reasoning about starlette's own dependency on it, avoiding a self-inflicted resolution deadlock.
- I independently grepped all third-party module-level imports across `creek/` and `creek_mcp/` and the set (pydantic, pydantic_settings, frontmatter, yaml, chardet, typer, rich, tqdm, httpx, cryptography, numpy, mcp, uvicorn, starlette, anyio, …) all resolve to declared dependencies or extras — consistent with the PR's claim of 19 roots, now all satisfied.

## Security Concerns
None. 🟢 This is a dependency-declaration/test-only change; no new attack surface.

## Problems
None found blocking merge.

## Code Quality
- Minor, non-blocking: `_import_time_statements` doesn't descend into `match`/`case` blocks (Python 3.10+ structural pattern matching). If a module ever gated an import behind `match`, it would be silently missed by the guard. Not currently exploited anywhere in the codebase, and easy to extend later if needed — not worth blocking on.
- The new test module adds real complexity (AST walking, distribution-metadata lookups) to what was previously a simpler pin-checking file; the extensive docstrings mitigate this well, keeping the "why" explicit for future maintainers.

## CLAUDE.md Adherence
✅ Coverage: three new focused tests for `anyio` plus the generalized guard test; consistent with existing patterns in the file (`_mcp_specifier`, cryptography CVE tests).
✅ `uv.lock` regenerated alongside the `pyproject.toml` change, per DEP-003/repo convention (avoids the stale-lock drift the guard elsewhere warns about).
✅ Declared in `[project.dependencies]`, not `[tool.uv].constraint-dependencies` — correctly follows DEP-003's distinction between direct and transitive-only packages.
✅ Docstrings are thorough and Google-style-consistent with the rest of the file.
✅ No suppressions, `noqa`, or threshold changes — aligns with the "no shortcuts" principle.

## Testing
Solid. `test_anyio_is_declared_as_a_direct_dependency`, `test_anyio_floor_tracks_the_locked_resolution`, and `test_locked_anyio_satisfies_the_declared_floor` cover the specific regression; `test_every_import_time_module_is_declared` is the generalization and is exactly the kind of "catch the next one automatically" test this repo's dependency-pin file already does well for CVEs. I was unable to execute `./scripts/test.sh` or `pytest` directly in this review sandbox (commands required interactive approval unavailable here), but per `gh pr view`'s `statusCheckRollup`, `Integration & E2E Tests`, `Code Complexity Analysis`, and `Build Distribution` have already completed with `SUCCESS`; `Tests & Type Checking` (3.11/3.12/3.13), `Pylint`, and `Static Analysis` were still `IN_PROGRESS` at review time — worth confirming they land green before merge, though nothing in the diff suggests they won't.

## Documentation
Excellent — the `pyproject.toml` comment, the `capabilities.py` docstring correction, and the `test_dependency_pins.py` module docstring update all tell a consistent, mutually-reinforcing story about why `anyio` is now direct and why `faiss` correctly remains exempt.



## Verdict: LGTM

The fix is correct, narrowly scoped, and verified: anyio's undeclared import at creek_mcp/httpapi/middleware/limits.py:32 is now declared with a floor matching uv.lock's resolution, the stale docstring rationale in capabilities.py is corrected rather than left to rot, and the new generic guard test correctly distinguishes hard (module-scope, including TYPE_CHECKING/try/with) from optional (function-local) imports — verified against the faiss counter-example it cites. No blocking issues found; nothing here needs to change before merge.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5229889525 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1254, opened 2026-08-09
feat(author): no injection seam, so creek author's exit-0-and-write-nothing cannot be tested


Geoffe-Ga:
## Summary

`creek author` and `creek.author` both **exit 0 / `status: ok`, print "No LLM provider available — rendering the deterministic stub", and write nothing.**

This is the exact signature of #460/#649/#658, still live. Found by the #1027 wiring contract, which could only *exempt* it rather than assert it — because neither surface offers a way to inject a provider.

## Why the contract could not cover it

`build_server` accepts `draft_llm_factory`, `compile_llm_factory` and `reflect_llm_factory` — but **no author factory**. The CLI builds `AuthorLLMClient` internally at `creek/cli.py:4964`.

So there is no hermetic path that reaches the writing desk, and no test can prove the wire past it. The contract records this as a `seam` exemption whose staleness detector is the blocker message itself: **the day an injection seam exists, the exemption fails and must be removed.**

## What to do

Add an author LLM injection seam symmetric with the three that already exist — `build_server(author_llm_factory=...)` and the CLI equivalent — then convert the #1027 exemption into a real assertion that `creek author` produces an artefact.

## Why this matters more than it looks

`creek author` is a headline capability. Today it is indistinguishable, from the outside, from a command wired to nothing: it succeeds, it explains itself, and it produces no output. That is precisely the failure mode epic #1024 exists to catch, and it is currently caught only by an exemption note.

## Acceptance criteria

- [ ] `build_server` accepts an author LLM factory, matching the draft/compile/reflect precedent.
- [ ] The CLI accepts the same injection.
- [ ] `tests/test_wiring_contract.py`'s `author` exemptions are removed and replaced with real effect assertions.
- [ ] A stub factory produces an observable artefact on both surfaces.

## Related

- #1027 (the contract that surfaced it), #1024 (the epic), #460 / #649 / #658

Issue #1253, opened 2026-08-09
fix(mcp): creek.report exposes 6 of 11 report types


Geoffe-Ga:
`creek_mcp/tools/report.py:69`'s `_VALID_TYPES` lists tags/voice/decisions/lexicon/rhetorical-patterns/mode-profiles. **Five CLI report types are unreachable over MCP**: `unnamed`, `fingerprint`, `paradox`, `synchronicity`, `wavelength`.

Issue #1252, opened 2026-08-09
fix(mcp): creek.link refuses the threads method the CLI supports


Geoffe-Ga:
`creek_mcp/tools/link.py:22` declares `_VALID_METHODS = ("embeddings", "temporal", "eddies")` — a **retyped copy** of `creek/cli.py:1491`'s `_LINK_METHODS` that lost `threads`.

PR #1255, opened 2026-08-09
test(wiring): contract test — every CLI command and MCP tool reachable and effectful

### LINKED ISSUE #1027: test(wiring): contract test — every CLI command and MCP tool reachable and effectful
Parent epic: #1024
Blocked by: #1025 (truthful vault fixture), #1026 (blocking CI lane)

## Role
You are a senior test engineer working in `creek-tools/`, following the repo's
TDD workflow and `./scripts/check-all.sh` gates.

## Goal
Add a table-driven **wiring contract** test: every CLI command and every MCP
tool must be reachable *and* produce its declared effect. A surface that runs
but does nothing must fail.

This is the issue that would have caught #935.

## Context

The repo has **36 CLI commands** (`grep -cE '@(app|clean_app|purge_app|skills_app|compost_app)\.command' creek/cli.py`)
and **23 MCP tools** (`grep -c '@server.tool' creek_mcp/server.py`, registered
in `build_server` at `creek_mcp/server.py:315-594+`).

Only **33 of 214** test modules touch the CLI via `CliRunner`. A unit test that
mocks the seam on both sides passes whether or not the seam is connected — so
the following all shipped green:

- **#935** — `PrivacyClassifier` never wired into `classify`. `creek classify`
  ran fine; the tiers it wrote were defaults. **Privacy tiers were not real.**
- **#649 / #658** — the author conductor stored an `llm_client` but built a bare
  `VoiceAgent()`; neither CLI nor MCP built a real client.
- **#460** — MCP `author` verb not connected to the real writing desk.
- **#580 / #581 / #577** — report generators orphaned from any command.

The common shape: **the surface exists, the implementation exists, the wire
between them does not.** Reachability alone is insufficient — `creek classify`
was always reachable. The test must assert an *effect*.

### Prior art in the house idiom
`tests/test_author_desk_wiring.py`, `tests/test_voice_fidelity_wiring.py`,
`tests/test_medium_contract.py`, `tests/test_remote_connector_contract.py`.
These are per-feature and hand-written; this issue generalises them into an
inventory that cannot be forgotten when a new command lands.

## Output Format
A PR containing:
- `tests/test_wiring_contract.py` (marked `integration`), holding:
  - an **inventory assertion**: the set of registered CLI commands and MCP tools
    is enumerated from the live `app` / `build_server()` objects and compared
    against a declared contract table. A new surface with no table entry
    **fails**, forcing the author to declare its effect.
  - a **per-surface effect assertion**: each entry names an observable
    post-condition — a file written under a scaffold path, a frontmatter key
    populated, an exit code plus a non-empty declared output.
- A short `docs/` note (or a `creek-tools/CLAUDE.md` §8 subsection) explaining
  how to add a surface to the table, so the failure message is actionable.

## Examples

The inventory, derived from the live objects — never a retyped list:

```python
def test_every_cli_command_is_declared() -> None:
    registered = {c.name or c.callback.__name__ for c in app.registered_commands}
    declared = set(WIRING_CONTRACT)
    assert registered == declared, (
        f"undeclared surfaces: {registered - declared}; "
        f"stale entries: {declared - registered}"
    )
```

An effect entry, in spirit — the assertion that #935 would have failed:

```python
WIRING_CONTRACT["classify"] = Surface(
    invoke=lambda vault: runner.invoke(app, ["classify", "--vault", str(vault)]),
    effect=lambda vault: _fragment_frontmatter(vault)["privacy_tier"] != DEFAULT_TIER,
    why="#935: classify ran but PrivacyClassifier was never called, so tiers were defaults",
)
```

`creek classify` exits 0 in both the wired and unwired world. Only the effect
assertion distinguishes them.

## Constraints
- **Effects, not smoke.** An assertion that a command exits 0 is worthless here
  — every one of the cited bugs exited 0. Each entry must assert a
  post-condition that is false when the wire is cut. A reviewer should be able
  to ask "which of the 17 wiring bugs would this entry have caught?" and get an
  answer.
- **Enumerate from the live objects**, never from a retyped list of names.
- **Hermetic**: no network, no API keys. Where a surface needs an LLM, assert
  the wire reaches the *client boundary* (as `test_author_desk_wiring.py`
  already does with a provider double), not that a model replied.
- Uses the truthful vault fixture from #1025. Land after it.
- Some of the 36 + 23 surfaces will turn out to be genuinely unwired. **Do not
  fix them here.** Add the entry, mark it `xfail(strict=True)` with an issue
  reference, and file the follow-up. Landing the harness is the deliverable; a
  strict xfail flips to a failure the moment someone fixes it, so nothing is
  lost.
- No `noqa`, `# type: ignore`, `pylint: disable`, or equivalent. Fix the root
  cause. The only exception is the documented 4-line escape hatch (third-party
  bug / version compat / benchmarked perf / generated code) with a review date.

## Done-Done
- [ ] Deleting the `PrivacyClassifier` call from `classify` makes this test fail
      (paste the output — this is the acceptance proof for the whole epic).
- [ ] Every surface either has a real effect assertion or a strict `xfail` with
      a filed follow-up issue number. The PR body lists the follow-ups.
- [ ] `./scripts/test.sh --all` passes.
- [ ] `./scripts/check-all.sh` exits 0.
- [ ] PR opened with `Refs #1024` and `Closes #<this-issue>`.
- [ ] Latest `Verdict:` on HEAD is `LGTM`.

Geoffe-Ga:
Closes #1027. Completes epic #1024.

A contract test asserting every CLI command and MCP tool is **reachable and effectful** — the capstone of the epic whose theme is *"a feature can be wired to nothing, run, exit 0, and be caught by no test."*

**It found five live defects on its first run.** Three are filed (#1252, #1253, #1254); two more are recorded below.

## The issue's own suggested derivation has a 38% blind spot

It proposes `{c.name or c.callback.__name__ for c in app.registered_commands}`. That yields **23 commands. There are 37.** Every command under `clean`/`purge`/`skills`/`compost` lives in `app.registered_groups`, which that expression never touches. It also yields the *Python* name rather than the CLI name, which differ for 4 of the 37 (`state-budget`, `compile`, `save`).

The contract walks the resolved click tree via `typer.main.get_command(app)` instead.

## "Effectful" is the hard part, and exit code is never the assertion

An entry names an observable, and the harness proves it in a world where it must be **present** and a world where it must be **absent**.

Filesystem observables carry their negative control for free: `writes` must match nothing before and something after **and intersect the run's own touched-set**, so a pre-existing artefact cannot vouch for the run. `contains` names the feature-specific key a downstream reader looks for. `frontmatter` reads an exact `{fragment id: value}` map back off disk, seeded across three privacy tiers — because the rules classifier fails **closed**, so an all-`intimate` map is what a *broken* wire also produces, and all-`unclassified` is exactly what #935 shipped.

**Printed observables have no free control** — "judged clean" and "found nothing to judge" render identically, which is the trap #1040 fell into. So every entry with `prints`/`absent` is *required* to declare a `Contrast`: a second world in which those substrings vanish and any `absent` substring reappears. `clean broken-links` must name the seeded defect and not name it on a bare vault; `review` fed `a` must print "5 accepted" and fed `d` must not.

`Refusal` makes a gate two-sided: the closed gate gives its documented reason **and** leaves the vault byte-identical, while the primary run satisfies the gate and produces the effect. A one-sided refusal test is passed by a surface that refuses everything.

**The structural rule that governs all of it:** every MCP tool appends `audit/mcp.jsonl`, including all five `purge.*` tools *while refusing*. `test_no_effect_is_satisfied_by_bookkeeping_alone` fails any entry whose declared artefacts live entirely inside those bookkeeping paths — a naive "some file changed" predicate would certify all 24 tools as effectful while proving only that the decorator ran.

## Would it have caught #1040?

Yes — that is the design constraint it was built against. A dormant guard produces no frontmatter keys, and `contains`/`frontmatter` assert the specific keys a downstream reader needs. The `Contrast` requirement exists precisely because #1040's in-bounds case was indistinguishable from "nothing was judged".

## Five defects found

| Finding | Status |
|---|---|
| `creek.link` refuses `threads` — a retyped `_VALID_METHODS` that lost it, making **all of #880's thread work unreachable over MCP** | **#1252**, strict xfail |
| `creek.report` exposes 6 of 11 types — `unnamed`/`fingerprint`/`paradox`/`synchronicity`/`wavelength` unreachable | **#1253**, strict xfail |
| `creek author` exits 0, prints "rendering the deterministic stub", writes nothing — #460/#649/#658 still live, and **untestable because no injection seam exists** | **#1254** |
| `creek compile` dies at the client boundary with an unhandled `ConnectError` instead of a clean operator message | recorded, exempt via seam |
| `purge classifications` clears nothing on a rules-classified corpus | recorded in the entry's `why` |

The two xfails are **strict**: they flip to a real pass when fixed, and fail loudly if someone marks them xpass without fixing.

## Exemptions ratchet rather than accumulate

Six, each with a reason and a **staleness detector**. `draft`'s reason is `blocker="LLM provider unavailable; cannot generate draft"` — the harness *runs the surface and asserts that message still appears*. The day the CLI gains a seam, the blocker vanishes and the exemption fails. Same shape as #1042's dormant-config allowlist: an entry that becomes provable must fail the test.

## Derivation is checked in both directions, and cites its sources

Three inventories — 37 CLI commands, 24 MCP tools, 39 mode variants — all read off live objects, **zero retyped name lists**. MCP names are deliberately *not* a third inventory: `test_mcp_read_gate.py` and `test_mcp_server.py` already ratchet those twice, so this module adds only the effect axis.

`Surface.derived_from` carries `module:symbol` provenance for every expectation, and `test_derived_from_names_importable_production_symbols` imports each one. **That test caught 17 of my own guessed citations on the first run** and forced them to real symbols.

One derivation trap worth naming: `wavelength` is special-cased at `creek/cli.py:3413` because it needs `--period`, so enumerating the dispatch dict alone under-covers by one. Pinned by a test that parses `creek report`'s own error message and asserts set equality — a second special case cannot hide.

## Honest limits

8 of 11 report types print "No X generated: no qualifying …" on a small corpus. The contract asserts only `report --type tags`, which genuinely writes; the others are declared in `MODE_CONTRACT` rather than falsely asserted, because they would "pass" a naive effect test while proving nothing.

Gate: **12/12**. 334 passed, 2 strict xfail, 25s. Marked `integration` so it runs in the blocking lane #1026 added.


Geoffe-Ga:
## Summary
This PR adds `tests/test_wiring_contract.py` (2349 lines) — a data-driven contract test proving every one of the 37 `creek` CLI commands, 24 MCP tools, and 39 dispatch-table mode variants is both reachable and *effectful* (not just exit-0). It ships alongside `docs/wiring-contract.md` explaining the vocabulary (`Surface`/`Effect`/`Contrast`/`Refusal`/`Exemption`) and a one-line README index entry. No production code is touched — the five live defects the harness found (#1252, #1253, #1254, plus two recorded-but-unfiled) are deliberately left for follow-up PRs via strict `xfail`s and documented exemptions.

## Strengths
- Both inventories (`registered_cli_commands`, `registered_mcp_tools`) are derived by walking the live `click` tree / `list_tools()` rather than retyped — directly closing the class of bug the epic targets, and the module even documents catching its own 38%-blind-spot mistake (`app.registered_commands` vs. the resolved tree).
- The `Contrast` requirement for any `prints`/`absent` observable (enforced by `test_printed_effects_declare_a_contrast`) is a genuinely sharp piece of test design — it stops a hardcoded banner from passing as a real scanner, which is exactly the shape of bug (#1040) this module exists to catch.
- `_BOOKKEEPING_GLOBS` + `test_no_effect_is_satisfied_by_bookkeeping_alone` closes the obvious loophole (every MCP tool appends `audit/mcp.jsonl`, including while refusing) before it could be exploited by a lazy future contract entry.
- The "Guarding the guard" block (`test_harness_fails_when_*`) is an unusually rigorous touch: it proves the harness itself goes red on a cut wire, a missing payload, and a bookkeeping-only effect, rather than just trusting the assertions are correct.
- Every entry's `why` and `derived_from` fields are machine-checked (`test_every_entry_says_which_wiring_bug_it_would_have_caught`, `test_derived_from_names_importable_production_symbols`) instead of being decorative comments that can rot.
- Exemptions require an *executable* staleness probe (`blocker`/`delegate`/`seam`), so the allowlist can only shrink — same discipline as the existing `DORMANT_CONFIG_FIELDS` pattern.
- `docs/wiring-contract.md` is genuinely useful onboarding material for the next person adding a surface, not boilerplate.

## Security Concerns
None. 🟢 The dummy elevated token (`"w" * 32`) is test-only, scoped via `monkeypatch.setenv` for the duration of the test, and never a real credential. Purge/deletion surfaces are exercised only against ephemeral `tmp_path` vaults.

## Problems
None found that block merge. Two very minor observations, neither of which affects correctness:
- `_digest_tree` uses the builtin `hash()` over file bytes for the before/after snapshot. `hash()` on `bytes` is salted per-process (PYTHONHASHSEED), which is fine here since comparisons only happen within one process/run, but it reads a little surprising next to the otherwise meticulous engineering — a one-line comment (or `hashlib.sha256`) would remove any doubt for a future reader.
- `MODE_CONTRACT`'s values are free-text provenance strings (e.g. `"entry CLI_CONTRACT['report']"`) that are never validated against the tables they reference, unlike `why`/`derived_from` on `Surface`/`Exemption`. A typo'd key name there would go unnoticed. Low value to fix given it's pure documentation, not an assertion input.

## Code Quality
- The `Bench`/`Outcome`/`Surface` abstraction is a lot of machinery for one test file, but the module docstring and `docs/wiring-contract.md` justify it well, and the "guarding the guard" tests demonstrate the machinery itself is exercised, not just trusted.
- Minor: the `voice-check` and `mine` contrast `why` strings are slightly imprecise about what's actually asserted (e.g. "no vault at all" when it's really an unseeded-but-initialized vault) — purely cosmetic, doesn't affect what the test proves.

## CLAUDE.md Adherence
✅ Coverage: this is a pure test/docs addition; no production coverage regression risk.
✅ Docstrings: every public function/class/dataclass field carries a docstring — consistent with the ≥95% interrogate bar.
✅ Patterns: `pytestmark = pytest.mark.integration` correctly places this in the hermetic, CI-blocking lane per `creek-tools/CLAUDE.md` §"Testing strategy" pointer and the PR's own note that it runs via `./scripts/test.sh --integration`.
✅ No forbidden shortcuts: the two `xfail`s are `strict=True` (so an accidental fix silently going unnoticed is impossible) and cite real, filed issues rather than suppressing anything.
✅ Commit/PR conventions: `test(wiring): ...` follows Conventional Commits; branch name `issue/1027-wiring-contract` follows the repo convention; body closes #1027 and epic #1024 with clear scoping of what's deferred.
⏳ Gate 3 (CI green): at review time, `Tests & Type Checking`, `Integration & E2E Tests`, `Pylint`, `Static Analysis`, and `CrawDad Quality` are still `IN_PROGRESS`; `Code Complexity Analysis` and `Build Distribution` are green. The PR body reports a clean local run (334 passed, 2 strict xfail, 25s) — worth confirming CI lands green before merge, per the repo's own stay-green policy, but nothing in the diff suggests it won't.

## Testing
This PR *is* a test, and an unusually well-built one — it includes tests of its own harness (the "Guarding the guard" section), machine-enforced provenance/rationale on every entry, and a documented, deliberate scope boundary (defect fixes tracked as separate follow-up issues rather than silently bundled in). No gaps identified.

## Documentation
`docs/wiring-contract.md` is complete, accurate against the code as written, and the README index entry is a correct one-line addition. The module's own docstring duplicates a fair amount of the doc file's content, but that mirrors the repo's existing convention of self-documenting test modules and isn't a DRY violation in the sense CLAUDE.md §1.2 warns about (one is user-facing how-to, the other is inline rationale for maintainers reading the test).



## Verdict: LGTM

The diff is a purely additive, extremely well-scoped and self-validating test/documentation contribution (no production code touched); static review found no logic bugs, security issues, or CLAUDE.md violations, and the two cosmetic nits noted (a `hash()` choice in `_digest_tree`, unvalidated `MODE_CONTRACT` prose) are not worth blocking on. CI was still in progress at review time with nothing red so far, which is a pre-merge check to confirm rather than a review blocker.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5230010480 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Summary
Adds `tests/test_wiring_contract.py` (2510 lines) plus `docs/wiring-contract.md`: a contract test proving every CLI command and MCP tool is both reachable and *effectful*, derived from live `click`/MCP introspection rather than retyped name lists. It found and filed 5 real wiring defects (2 as strict `xfail`s, 3 as separate issues), and a follow-up commit fixed a genuine CI-only ANSI/Rich-highlighting false negative discovered after the first CI run.

## Strengths
- Every `derived_from` provenance citation and every claimed inventory number was spot-checked against the actual production source in this checkout (`_LINK_METHODS` has `"threads"`, MCP's `_VALID_METHODS` doesn't; `_REPORT_DISPATCH` has 10 entries + `wavelength` = 11, MCP `_VALID_TYPES` has exactly 6) — all matched exactly.
- The "printed effects require a `Contrast`" design directly addresses the #1040 failure mode (a scanner that renders "clean" identically to "nothing to scan"), and `_BOOKKEEPING_GLOBS` correctly stops an entry from being satisfied merely by the universal `audit/mcp.jsonl` append.
- Strong self-defense: `test_derived_from_names_importable_production_symbols`, `test_every_entry_says_which_wiring_bug_it_would_have_caught`, `test_printed_effects_declare_a_contrast`, and the "Guarding the guard" block (tests that the harness itself goes red on a synthetic dead wire) are exactly the kind of meta-tests that keep a contract test from rotting into theatre.
- The second commit is a model of root-causing a CI-only failure (Rich ANSI splitting `--mode` and highlighted numbers) rather than papering over it — includes a mutation-tested normalizer and a regression guard built from the real renderer's output, not a hand-typed escape sequence.
- Both strict `xfail`s are exactly that (`strict=True`), so they'll force follow-up when the underlying bugs are fixed, and the exemption ratchet (`blocker`/`delegate`/`seam`) prevents silent permanent carve-outs.
- Docstrings throughout are Google-style with full Args/Returns, consistent with the `creek-tools/CLAUDE.md` ≥95% interrogate requirement.

## Security Concerns
None. 🟢 No new I/O boundaries, no secrets, no subprocess/network use beyond the existing test harness patterns (`CliRunner`, in-process MCP server).

## Problems
None found that block merge.

## Code Quality
- Minor: `CLI_CONTRACT["review"].contrast.why` says "deferring every fragment must persist a different decision," but `test_declared_contrast_makes_the_observable_vanish` only checks `effect.prints`/`effect.absent` under contrast, not `effect.frontmatter` — so the frontmatter half of that claim isn't actually exercised by the contrast run (only by the primary effect test). Not a defect, just a slightly overstated `why` string; not worth blocking.
- The file is large (2510 lines) for a single module, but the internal organization (vocabulary → corpus → bench → CLI contract → MCP contract → exemptions → mode contract → assertions → meta-contract → guard-the-guard) is clear and consistently sectioned, so this reads as justified scope for a capstone contract test rather than something that should be split.

## CLAUDE.md Adherence
✅ Uses `./scripts/test.sh --integration` lane per repo convention; correctly marked `pytestmark = pytest.mark.integration` so it runs in the CI-blocking `integration-e2e` job (verified against `.github/workflows/ci.yml`).
✅ No shortcuts: no `# noqa`/`# type: ignore` observed in the new file; the two `xfail`s are `strict=True` with issue citations, which is the sanctioned form of "known, tracked, forcing failure" per §1.3.
✅ Docstrings are thorough Google-style with Args/Returns throughout, consistent with the ≥95% interrogate bar.
✅ Commit messages are Conventional Commits (`test(wiring): ...`, `fix(test): ...`) with substantive bodies citing issues.
❓ Could not independently run `./scripts/check-all.sh` (coverage/mypy/ruff/complexity) in this review environment (`uv`/venv not available), so full gate compliance is taken on the PR's own report ("Gate: 12/12. 334 passed, 2 strict xfail, 25s") rather than independently reproduced — flagged for transparency, not as a concern.

## Testing
✅ This PR *is* the test suite — coverage of its own claims is unusually strong, with dedicated tests asserting the test harness itself can fail (`test_harness_fails_when_a_declared_artefact_is_never_written`, `test_harness_fails_when_the_payload_is_missing_from_a_written_artefact`, `test_harness_fails_when_a_uniform_frontmatter_map_replaces_a_real_one`, `test_harness_rejects_an_effect_made_only_of_bookkeeping`). This is exactly the kind of "test the test" rigor that prevents a contract suite from becoming vacuous over time.

## Documentation
✅ `docs/wiring-contract.md` (151 lines, new) mirrors the module's structure and vocabulary precisely (field-by-field breakdown, two easy-to-get-wrong rules, known drift, the corpus rationale) and is linked from `docs/README.md`'s index table. No drift found between the doc and the module's actual behavior.



## Verdict: LGTM

The contract test is exceptionally rigorous: every `derived_from` citation, every claimed inventory count (37 CLI commands, 24 MCP tools, "6 of 11" report types, the missing "threads" link method), and every referenced production symbol were independently checked against the actual source in this checkout and all matched. The module is correctly marked `integration` and wired into the CI-blocking `integration-e2e` job. The second commit shows the author caught and fixed a real CI-only ANSI-escape failure with a targeted, well-tested normalization layer rather than a weakening hack. No blocking issues found; only a very minor documentation-vs-test-coverage gap (the `review` contrast's stated rationale mentions a frontmatter check that isn't actually asserted under contrast) which is not worth blocking merge over.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5230139435 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Adds `tests/test_wiring_contract.py` (2510 lines) plus `docs/wiring-contract.md`: a contract test asserting every CLI command (37) and MCP tool (24) is both reachable and *effectful* (not just exit-0). It derives its inventories from live `click`/MCP objects rather than retyped name lists, requires printed judgements to declare a negative-control `Contrast`, requires refusal gates to be proven two-sided, and forbids "effectful" from being satisfied by bookkeeping writes alone (e.g. the shared `audit/mcp.jsonl`). It found and filed 3 real defects (#1252, #1253, #1254) and documents 2 more inline, and a follow-up commit in the same PR fixed a genuine CI-only flakiness bug (Rich ANSI escapes splitting literal substrings).

## Strengths
- **Claims are independently verifiable and correct.** I checked the two "known drift" xfails against the actual source: `creek_mcp/tools/link.py:_VALID_METHODS = ("embeddings", "temporal", "eddies")` indeed omits `"threads"`, while `creek/cli.py:_LINK_METHODS` has all four — confirms the #1252 finding described in the PR is real, not asserted-but-untrue.
- **Derived, not retyped, inventories.** `cli_surface_names()` walks the resolved `click` tree instead of `app.registered_commands`, correctly avoiding the 38%-blind-spot bug the issue's own suggested snippet would have shipped (commands under `clean`/`purge`/`skills`/`compost` live in `app.registered_groups`).
- **"Guarding the guard" tests** (`test_harness_fails_when_a_declared_artefact_is_never_written`, `..._when_the_payload_is_missing...`, `..._when_a_uniform_frontmatter_map_replaces_a_real_one`, `test_harness_rejects_an_effect_made_only_of_bookkeeping`) prove the harness's own predicates are not vacuous — a meta-level rigor that's easy to skip and wasn't.
- **Real bug caught mid-PR.** The second commit diagnoses and fixes a CI-only failure (Rich's `typer` highlighter splitting `--mode` across ANSI style spans) at the single `_says`/`_flatten` comparison layer rather than patching the one failing assertion, and backs the fix with a dedicated regression test (`test_declarations_survive_a_colour_terminal`) built from the real renderer rather than a hand-typed escape sequence, plus documented mutation-testing of the normaliser (6/6 killed).
- **Negative controls are structural, not incidental.** Filesystem effects get their "not run" control for free via before/after snapshots intersected with the run's own touched-set (so a pre-existing artefact can't vouch for a run); printed effects are *required* to declare a `Contrast`, enforced by `test_printed_effects_declare_a_contrast`.
- **Exemptions ratchet.** Every `Exemption` must declare exactly one executable staleness probe (`blocker`/`delegate`/`seam`), enforced by `test_exemption_declares_exactly_one_staleness_probe`, so an exemption becomes provably stale rather than a silent, permanent escape hatch.
- **Provenance is checked, not just written.** `derived_from` citations are `module:symbol` strings and `test_derived_from_names_importable_production_symbols` actually imports each one — I spot-checked several (`DERIVATIVE_FRONTMATTER_KEY`, `GROUNDING_FRONTMATTER_KEY`, `ELEVATED_TOKEN_ENV`, `_LINK_METHODS`) and all resolve to real symbols.

## Security Concerns
None. 🟢 No new I/O, network, subprocess, or credential handling; the elevated-token fixture (`_ELEVATED_TOKEN = "w" * 32`) is a synthetic value scoped to a monkeypatched env var for the test process only, not a real secret.

## Problems
None blocking.

## Code Quality
- `_digest_tree` uses Python's built-in `hash(path.read_bytes())` (combined with file size) as a change-detection digest. Built-in `hash()` on `bytes` is seeded per-process (`PYTHONHASHSEED`), so this is safe *within* a single test run (before/after snapshots are taken in the same process) — not a bug — but a non-cryptographic hash risks (extremely unlikely) collisions and reads slightly surprising next to the rest of the module's care. `hashlib.sha256(...).hexdigest()` would be marginally more obviously correct to a future reader, though this is a nit, not something worth blocking on.
- The single 2510-line test file is large, but the tight cross-referencing between `CLI_CONTRACT`, `MCP_CONTRACT`, `MODE_CONTRACT`, and `EXEMPTIONS` (and the meta-tests that walk all of them) is a reasonable justification for keeping it as one module rather than splitting and losing that cohesion.

## CLAUDE.md Adherence
✅ **Coverage** — file is test-only; `[tool.coverage.run]` omits `*/tests/*` and `*/test_*.py`, so the ≥90% branch-coverage gate is unaffected by this change (no production code touched).
✅ **Docstrings** — every public function, dataclass, and module has a Google-style docstring with `Args`/`Returns`, consistent with the ≥95% interrogate threshold.
✅ **Typing** — dataclasses and functions are fully annotated (`from __future__ import annotations`, `Final`, `Mapping`, etc.); nothing here looks like it would trip MyPy strict.
✅ **Markers** — uses `pytestmark = pytest.mark.integration`, and `integration` is a defined, blocking-in-CI marker per `pyproject.toml`.
✅ **DRY / single source of truth** — inventories are derived from live objects rather than retyped, which is precisely the principle CLAUDE.md's 1.2 asks for, applied here as the organizing design constraint of the whole module.
✅ **Commit conventions** — commit headlines follow Conventional Commits (`test(wiring): ...`, `fix(test): ...`).

## Testing
This PR *is* the test. Coverage of the new harness itself is strong: it doesn't just assert commands run, it proves each declared effect is falsifiable (via the four "guarding the guard" tests) and that its own `Contrast`/`Refusal`/`Exemption` bookkeeping rules are enforced by dedicated meta-tests. The 2 strict `xfail`s are appropriately used to pin real, filed drift (not to hide a broken assertion) and will fail loudly (not silently xpass) once the underlying bugs are fixed.

## Documentation
`docs/wiring-contract.md` is thorough and mirrors the module docstring closely: it explains the vocabulary (`Shape`, `Effect`, `Contrast`, `Refusal`, `Exemption`), gives a worked example for adding a surface, calls out the two easy-to-get-wrong rules (bookkeeping-only effects, and asserting the downstream reader's key), and documents the known drift with issue citations. `docs/README.md` index entry is added correctly.



## Verdict: LGTM

The PR is a purely additive test/doc change (no production code touched), its central claims (the `threads` link-method drift and the MCP report-type gap) are independently verifiable against the actual source and check out, the harness includes explicit "guarding the guard" tests that prove its own assertions can fail, and the author already found and fixed a real CI-only flakiness bug (Rich ANSI splitting) within the same PR. Nothing here blocks merge; the only remarks are minor, non-blocking style observations.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5230217101 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1256, opened 2026-08-09
chore(deps): migrate rpds-py 0.30.0 -> 2026.6.3 in both locks

### LINKED ISSUE #1185: [scan:deps] migrate rpds-py 0.30.0 -> 2026.6.3 in both locks (breaking: SemVer -> CalVer scheme, raised Python floor)
## Role
suppressions).

## Goal
Move `rpds-py` from `0.30.0` to `2026.6.3` in **both** `creek-tools/uv.lock`
and `crawdad/uv.lock` — a major that also changes the project's versioning
scheme from SemVer to CalVer — and record the scheme change where a future
ceiling would otherwise be written, verified by `./scripts/check-all.sh` green
in both subprojects and by `uv export --locked` (the CI gate at
`.github/workflows/ci.yml:83`) succeeding.

## Context
- File(s):
  - `creek-tools/uv.lock:3100` — `rpds-py` `0.30.0`
  - `crawdad/uv.lock:2203` — `rpds-py` `0.30.0`
  - `creek-tools/pyproject.toml:155-186` — the `[tool.uv]
    constraint-dependencies` block, where a floor would go if one is warranted
  - No source file imports `rpds-py`; it is transitive-only (verified: zero
    `import rpds` / `from rpds` hits under `creek-tools/` and `crawdad/`)
- Scanned at commit: `9529d8d9e4ad8df7dba549460b2264cceb51fdfe` — re-verify
  against HEAD before starting
- Evidence: `uv.lock` versus PyPI at the scanned SHA. `rpds-py` is one of only
  five majors in the creek-tools lock, and **the only one no open issue tracks**
  (`mcp` → #998, `pyarrow` → #1001, `websockets` → capped below 17.0 by
  `google-genai`, `pywin32` → Windows-only via `mcp`).

  | | locked | latest | released |
  |---|---|---|---|
  | `rpds-py` (both locks) | `0.30.0` | `2026.6.3` | 2026-06-30 |

  Release line, from the PyPI JSON API and `crate-py/rpds` releases:

  ```
  0.29.0     2025-11-16
  0.30.0     2025-11-30   <- both locks are here, ~8 months stale
  2026.5.1   2026-05-28   <- SemVer -> CalVer switch, no 0.31 / 1.0 in between
  2026.6.3   2026-06-30
  ```

  **The three things that make this a major, not a routine bump:**

  1. **Versioning scheme change.** `0.30.0` → `2026.5.1` is not a SemVer major
     bump; the project abandoned SemVer for CalVer outright. Any ceiling
     written in the SemVer idiom — `rpds-py<1.0`, or a `~=0.30` compatible
     release clause — now *permanently* excludes every future release. This is
     the durable hazard: it must not be encoded into
     `constraint-dependencies` by reflex when a floor is added later.
  2. **Raised Python floor.** `requires-python` goes `>=3.10` → `>=3.11`. Both
     subprojects already declare `requires-python = ">=3.11"`
     (`creek-tools/pyproject.toml:8`, `crawdad/pyproject.toml:8`), so this is
     compatible — but it is the reason a resolver on an older interpreter would
     silently hold the lock at `0.30.0` rather than error.
  3. **Native extension.** `rpds-py` is a compiled Rust binding (`crate-py/rpds`
     wrapping `orium/rpds`), so the bump swaps wheels, not pure Python. Wheel
     coverage differs across the jump (115 files at `0.30.0`, 116 at
     `2026.6.3`); confirm a wheel exists for every interpreter in the CI matrix
     (3.11 / 3.12 / 3.13) rather than falling back to a source build.

  **Nothing upstream is holding it back.** The two consumers declare open
  floors: `jsonschema` `4.26.0` requires `rpds-py>=0.25.0` and `referencing`
  `0.37.0` requires `rpds-py>=0.7.0` — no ceiling in either. Both `jsonschema`
  and `referencing` are themselves already at the latest release, so **no other
  open issue's bump will re-resolve this subtree**; the lock is simply frozen.
  The path in both projects is `mcp` → `jsonschema` → `referencing` →
  `rpds-py`, i.e. it is loaded on every MCP server start.

  **Blast radius is small but real.** No repo code touches `rpds-py` directly,
  so there are no call sites to migrate — the risk is entirely
  install/ABI-shaped (wheel availability, native import) plus whatever
  `referencing`'s cache behaviour inherits from an eight-month jump in the
  backing data structures. That is why this is a lock-and-verify migration
  rather than a code migration.

  Note: OSV was queried for `rpds-py 0.30.0` at this SHA and returned no
  advisories. This is a staleness/major-migration finding, not a CVE.
- Related: #1184 (the creek-tools transitive-closure refresh filed in the same
  `[scan:deps]` run — it explicitly holds `rpds-py` back for this issue), #1002
  (the crawdad lock refresh, which scoped itself to "none by a major" and so
  excluded this), #998 (`mcp` 2.0 migration — the top of the same dependency
  path), #1001 (`pyarrow`, the other untracked major at the time it was filed),
  #1085 / #986 (why lock drift accumulates unseen in each subproject).

## Output Format
"Closes #N".

## Examples
Lock-only in both subprojects — a bare `uv lock` is a no-op because `0.30.0`
already satisfies `>=0.25.0`, so the per-package upgrade flag is required:

```bash
cd creek-tools && uv lock --upgrade-package rpds-py
cd ../crawdad  && uv lock --upgrade-package rpds-py

# Confirm only rpds-py moved, in both locks.
git diff --stat creek-tools/uv.lock crawdad/uv.lock

# Confirm a real wheel exists for every CI interpreter (no sdist fallback).
python - <<'PY'
import json, urllib.request
d = json.load(urllib.request.urlopen('https://pypi.org/pypi/rpds-py/2026.6.3/json'))
tags = {f['filename'] for f in d['urls'] if f['packagetype'] == 'bdist_wheel'}
for cp in ('cp311', 'cp312', 'cp313'):
    assert any(cp in t and 'manylinux' in t for t in tags), cp
PY
```

If a ceiling is ever added for this package, write it in CalVer — never in the
SemVer idiom, which would freeze the dependency forever:

```toml
# creek-tools/pyproject.toml — [tool.uv] constraint-dependencies
# WRONG: rpds-py switched from SemVer to CalVer at 2026.5.1, so a <1.0
#        ceiling excludes every release from 2026 onward.
"rpds-py>=0.30.0,<1.0",

# RIGHT: CalVer floor, no SemVer-shaped ceiling.
"rpds-py>=2026.6.3",
```

The failing-test-first step: assert the locked `rpds-py` version parses as
CalVer (year component ≥ 2026) in **both** locks, so a future conservative
re-lock or a hand-written SemVer ceiling cannot silently regress the pin to the
`0.x` line.

## Constraints
- **`rpds-py` only.** Do not fold in other stale pins — the creek-tools
  transitive refresh is #1184 and the crawdad one is #1002; overlapping edits
  conflict with two open agent-ready issues
- Never batch this major with minor/patch bumps
- Do not add a `<1.0`, `~=0.30`, or any other SemVer-shaped ceiling for
  `rpds-py`; if a bound is needed at all, express it in CalVer
- Both locks move in the same PR or neither does — a split leaves the two
  subprojects resolving different native extensions for the same MCP path

Geoffe-Ga:
Closes #1185.

rpds-py 0.30.0 → 2026.6.3 in **both** locks. Not an ordinary bump: the project changed versioning scheme (SemVer → CalVer) and raised its Python floor.

## The Python floor — the actual risk, checked rather than assumed

rpds-py's own `requires-python` rises `>=3.10` → `>=3.11`, confirmed from the PyPI JSON for both releases.

**No repo `requires-python` changed.** `creek-tools/pyproject.toml:8` and `crawdad/pyproject.toml:8` both already declare `>=3.11`, so the raised floor excludes nothing this repo supports and **3.11 is not excluded**.

All three CI interpreters resolve, in both projects — `uv export --locked --all-extras --python {3.11,3.12,3.13}` exits 0 and emits exactly one line, `rpds-py==2026.6.3`, in **all six combinations**. No resolver fork, no silent hold at 0.30.0 for any interpreter.

Wheel coverage is real, not an sdist fallback: 115 bdist_wheels published, and the locked entry carries 30 cp311 + 30 cp312 + 30 cp313 wheel URLs across manylinux, macos-arm64 and win_amd64. Native import smoke-tested in both venvs — `import rpds` succeeds, `importlib.metadata.version` reports 2026.6.3, and the full consumer path (jsonschema → referencing → mcp) imports clean.

## Both gates, because CrawDad has its own

- **creek-tools** `./scripts/check-all.sh` — 12/12, 8596 passed, coverage 94.35%
- **crawdad** `./scripts/check-all.sh` — separate script, separate venv, **not** covered by the creek-tools gate — 7/7, 583 passed, coverage 94.60%

CrawDad's `scripts/security.sh` is the load-bearing one here: it pipes `uv export --locked --all-extras` into pip-audit, so it doubles as a lock-freshness gate *and* an advisory scan of the new version. Both clean.

## A design call the issue left open

The issue says the constraint block is *"where a floor would go if one is warranted."* I judged one warranted and added a bare `rpds-py>=2026.6.3` to **both** `[tool.uv].constraint-dependencies` blocks, with no ceiling.

Without it the guard asserts only against the lock — and the lock is precisely the artifact a conservative relock can quietly hold back, since jsonschema asks only for `>=0.25.0`. The floor also puts the correct CalVer idiom in situ next to five SemVer floors, with a comment naming `<1.0` and `~=0.30` as the traps. `test_rpds_py_constraint_admits_future_calver_releases` probes 2027.1.1, so a future SemVer-shaped ceiling fails the gate in both projects rather than silently freezing the dependency.

## Four corrections to the issue

- **Dependency path is incomplete.** It states `mcp → jsonschema → referencing → rpds-py`. `uv tree --invert` shows jsonschema depends on rpds-py **directly** as well, so there are two edges. Doesn't change the fix, but it means removing `referencing` would not remove rpds-py.
- **The cited CI gate line is prose.** `ci.yml:83` is a comment inside the Static Analysis job; the actual `uv export --locked` invocations are at `:116`, `:209`, `:279`, `:434`, plus two scan workflows. I verified the gate passes rather than the cited line.
- **Wheel count is off by one in kind:** 2026.6.3 publishes 115 wheels **plus 1 sdist** = 116 files. The issue's "116" reads as wheels.
- **Lock line refs stale by 1–4 lines** (`uv.lock:3100` → 3105).

Verified true: 0.30.0 in both locks; **zero** `import rpds` / `from rpds` anywhere; the release line really does jump `0.29.0 → 0.30.0 → 2026.5.1 → 2026.6.3` with no 0.31 or 1.0; jsonschema and referencing both declare open floors with no ceiling — which is exactly why a bare `uv lock` is a no-op and `--upgrade-package` was required. No OSV advisories, so this is a staleness migration and no suppression question arises.

## No other pins moved

Per the issue's constraint, this touches rpds-py alone. A lock diff mixing an intentional migration with drive-by upgrades is unreviewable.


Geoffe-Ga:
## Summary
Migrates `rpds-py` from `0.30.0` to `2026.6.3` in both `creek-tools/uv.lock` and `crawdad/uv.lock` (closes #1185), adding a bare-floor `constraint-dependencies` entry (`"rpds-py>=2026.6.3"`) to both `pyproject.toml`s and a matching suite of guard tests in both `tests/test_dependency_pins.py` files. No production code is touched — this is purely dependency-pin + test infrastructure.

## Strengths
- **Verified, not assumed.** I checked the PR's central factual claims against the actual repository state: both `pyproject.toml`s already declare `requires-python = ">=3.11"` (so rpds-py's raised Python floor excludes nothing), and both `uv.lock`s resolve `rpds-py` to `2026.6.3`. Both check out.
- **Scoped diff.** Diffing the two `uv.lock` files for changed `name = "..."` package entries shows only the `rpds-py` block moved — no drive-by upgrades mixed into a migration PR, exactly as the description claims.
- **Tests encode the actual hazard, not just the version.** `test_rpds_py_constraint_admits_future_calver_releases` probes a future CalVer release (`2027.1.1`) specifically to catch a SemVer-shaped ceiling (`<1.0`, `~=0.30`) that would silently re-freeze the dependency forever — a real, non-obvious footgun given the SemVer→CalVer scheme switch, and a genuinely useful regression guard.
- **Mirrored correctly across both sibling projects.** The constraint entry, docstring updates, module-level constants, and four new test functions are consistent between `creek-tools` and `crawdad`, each reusing existing `_PYPROJECT`/`_UV_LOCK`/`SpecifierSet` infrastructure already present in the file rather than introducing new patterns.
- **PR description self-corrects the source issue** (dependency path, CI gate line reference, wheel count, lock line refs) rather than taking it at face value — a good sign of diligence.
- **Precedent-consistent placement.** The new floor is appended to `[tool.uv].constraint-dependencies` alongside other transitive-only floors (DEP-003 convention), and omitting a CHANGELOG entry matches how prior dependency-pin PRs (e.g., the `starlette`/`torch`/`cryptography` floors already in these files) were handled — no CHANGELOG entries exist for those either.

## Security Concerns
🟢 LOW — None. This is explicitly not a CVE fix (confirmed: no OSV advisories on the old or new version per the PR body), just a staleness migration with a versioning-scheme gotcha. `crawdad`'s `scripts/security.sh` (which pipes `uv export --locked` into `pip-audit`) already ran clean in CI for all three Python versions.

## Problems
None found that block merge.

## Code Quality
- The new explanatory comment block in both `pyproject.toml`s (13 lines) and the docstring/module-constant additions in both test files are near-verbatim duplicated between `creek-tools` and `crawdad`. This is unavoidable/appropriate here (two independently-versioned projects, no shared code path), and matches the existing style of the surrounding CVE comments in the same files — not a real DRY violation, just noting the verbosity for awareness.
- Minor nit only: `_rpds_py_constraint_specifier`'s `.get("tool", {}).get("uv", {}).get(...)` chain is wrapped in parens across two lines purely for line length — stylistically fine and consistent with `ruff format` output elsewhere in the file.

## CLAUDE.md Adherence
✅ Uses project scripts per the PR description (`./scripts/check-all.sh` reported 12/12 for creek-tools and 7/7 for crawdad, run separately since crawdad has its own gate per its CLAUDE.md §7).
✅ Commit message (`chore(deps): migrate rpds-py 0.30.0 -> 2026.6.3 in both locks`) follows Conventional Commits.
✅ No shortcuts — no `# noqa`, `# type: ignore`, or `--no-verify` introduced.
✅ Floor placed in `[tool.uv].constraint-dependencies` per the DEP-003 precedent documented in both CLAUDE.md/pyproject.toml comment blocks, since rpds-py is transitive-only.
⚠️ At review time, several `creek-tools` CI jobs (Static Analysis, Pylint, Tests & Type Checking on 3.11/3.12/3.13) were still `IN_PROGRESS`, while `CrawDad Quality` had already completed successfully on all three Python versions. This isn't a defect in the change — the PR body reports clean local runs (8596 tests passed, 94.35% coverage) — but Gate 3 (CI green) wasn't fully confirmed at the moment of this review.

## Testing
Insufficient-testing would be blocking, but that's not the case here — testing is thorough and appropriately targeted:
- Two independent guards per project (declared constraint + resolved lock version), matching the existing pattern for every other pin in the file.
- `test_rpds_py_constraint_rejects_the_abandoned_semver_line` and `test_rpds_py_constraint_accepts_the_calver_floor` pin the boundary precisely at the SemVer/CalVer transition.
- `test_rpds_py_constraint_admits_future_calver_releases` is the standout test — it's the one assertion that actually prevents a future contributor from "fixing" this pin by adding a SemVer-shaped ceiling and silently freezing the dependency forever.
- `test_locked_rpds_py_is_on_the_calver_line` checks both the exact version floor and the `major` component to confirm the lock is genuinely off the abandoned `0.x` line, not just satisfying string comparison.

## Documentation
Docstrings on both new helper functions and all four new test functions are clear, explain the *why* (not just the *what*), and match the file's existing Google-style docstring conventions (required for the ≥95% interrogate gate). The `pyproject.toml` comment blocks are unusually thorough — they explain the CalVer hazard in situ, which is exactly where a future maintainer editing this constraint would need it.



## Verdict: LGTM

The change is a mechanical, well-justified dependency migration (rpds-py 0.30.0 → 2026.6.3, a SemVer→CalVer scheme change) mirrored identically across both sibling projects, with new guard tests that correctly encode the CalVer-vs-SemVer ceiling hazard. I independently verified the PR's key factual claims against the actual repo content (both `requires-python = ">=3.11"`, both `uv.lock`s resolve `rpds-py` to `2026.6.3`, and no other package entries changed in either lockfile) and they all hold. No production code is touched, no blocking issues, and CrawDad's CI gate already passed in full.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5230051560 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #322, opened 2026-05-24
enhancement(cli): auto-discover `creek_config.yaml` from `--vault` path


Geoffe-Ga:
## Symptom

Running a CLI command like `creek redact --scan --source X` (or `creek classify`, `creek ingest`, etc.) without setting `CREEK_CONFIG` or passing `--config` produces:

```
Config file creek_config.yaml not found; running with built-in defaults. Run `creek init --vault <vault>` to write a starter config, set CREEK_CONFIG to an absolute config path, or pass --config <path> explicitly. Pipeline behaviour (privacy, redaction, cleaning) depends on this file — silent defaults are usually not what you want.
```

…even when the command was passed `--vault <path>` and `<path>/00-Creek-Meta/creek_config.yaml` exists right there. The CLI has all the information it needs to find the config but doesn't.

## Expected

When `--vault <path>` is provided, the CLI should automatically discover `<path>/00-Creek-Meta/creek_config.yaml` and use it. The "config not found" warning should only fire when:

- no `--vault` was given, **and**
- no `CREEK_CONFIG` env var is set, **and**
- no `--config` was passed.

Explicit `--config` and `CREEK_CONFIG` overrides should still take precedence over the auto-discovered vault config (so a user can point at a different config without symlinking).

## Why this matters

The warning itself acknowledges the importance: *"Pipeline behaviour (privacy, redaction, cleaning) depends on this file — silent defaults are usually not what you want."* Yet the CLI silently runs with defaults in the most common case (a user with one vault who passes `--vault` everywhere and doesn't know about `CREEK_CONFIG`).

The defaults that get applied include differences that matter — e.g., the LLM provider defaults to `ollama` (local) but the user might have configured `anthropic` in their vault's config. Running with defaults silently sends the wrong code path.

## Suggested fix

Add config auto-discovery in the CLI argument parsing layer:

```python
def resolve_config_path(vault: Path | None, explicit: Path | None) -> Path | None:
    if explicit:
        return explicit
    if env := os.environ.get("CREEK_CONFIG"):
        return Path(env)
    if vault:
        candidate = vault / "00-Creek-Meta" / "creek_config.yaml"
        if candidate.exists():
            return candidate
    return None
```

…wired into every subcommand that takes `--vault`.

## Discovered during

[...]

PR #1259, opened 2026-08-09
fix(mcp): derive link methods and report types from one definition

### LINKED ISSUE #1252: fix(mcp): creek.link refuses the threads method the CLI supports

### LINKED ISSUE #1253: fix(mcp): creek.report exposes 6 of 11 report types

Geoffe-Ga:
Closes #1252. Closes #1253.

Both defects #1027's contract found on its first run, and both the same mechanism: a retyped tuple in `creek_mcp/tools/` drifting from the CLI's definition.

- `creek.link` refused `threads` — **making the entire thread half of #880 unreachable over MCP**
- `creek.report` exposed 6 of 11 types

## The fix is the derivation, not the missing strings

A new `creek/surface_modes.py` holds the single definition both frontends import. Adding `"threads"` to a hand-maintained tuple would have left the mechanism that produced *both* bugs intact.

**The xfails were not left as constant-vs-constant assertions**, which would be tautological now that both frontends import the same tuple. Each test drives the real surface with a bogus value and parses the two rejection messages, then asserts the advertised sets are equal — so a re-introduced private copy fails on **behaviour**, not on inspection.

## Four types are deliberately not served unconditionally, and the reason is a leak

This is the part worth reading. `report.py`'s docstring justified the omissions:

> *"`unnamed` and `wavelength` are deferred to the CLI because they need date arithmetic the MCP shape should not own."*

**Both halves are false.** `creek report --type unnamed` takes no period argument at all — `_report_unnamed` hardcodes the current ISO week, so there was never any arithmetic to own. And `wavelength`'s loader **already accepted a `PrivacyTierOverride`**; it was fully MCP-servable and needed only a `period` parameter plumbed. The docstring was also **silent about `fingerprint`, `paradox` and `synchronicity`** — three of the five omissions were never explained at all, which is precisely the silent-omission failure #1253 names.

The real blocker is different and more serious: `UnnamedDigestGenerator.generate_weekly_digest`, `build_fingerprint`, `generate_paradoxes` and `generate_synchronicities` accept **no `PrivacyTierOverride`**. Serving them at `ceiling=open` would distil intimate content into vault artifacts — **the exact leak #968 closed.**

So they are served only at `privacy_tier_ceiling=all` (the ceiling `creek report` itself runs them under) and refused below it **by name**, with the generator symbol a reader must widen stated in the refusal. Nothing is silently omitted, and the path to serving them properly is written down where the next person will look.

## A pre-existing CLI bug found in passing

`creek report --type wavelength --include-tier <x>` **silently ignored the flag** — `report()` called `_report_wavelength(vault_path, period)` with no override, and the loader defaulted to ALL. Since the MCP branch now honours the ceiling, leaving this would have made the CLI lie relative to MCP. Both surfaces now share `generate_phase_map`, which was added to `_OVERRIDE_CALL_NAMES` so the structural tier guard covers it on both.

## A test that encoded the defect as expected behaviour

`test_report_refuses_unsupported_type` used **`wavelength`** as its example of an unsupported type — pinning #1253's bug as correct. Repointed at a genuinely unknown name.

Also corrected: `test_mcp_report_tier_ceiling.py`'s `_CASES` docstring claimed it covered *"every report type `creek.report` exposes"*; it covered 4 of 6 even before this change.

## The #1027 trap was honoured

`wavelength` is genuinely absent from `_REPORT_DISPATCH` (special-cased because it needs `--period`). `REPORT_TYPES` names all eleven, but `declared_cli_mode_variants()` and `MODE_CONTRACT` deliberately keep deriving from `_REPORT_DISPATCH` **plus the one special case** — so `test_report_error_message_lists_exactly_the_declared_types` stays a real declaration-vs-dispatcher assertion rather than comparing the shared tuple with itself.

RED: 17 failures in `test_mcp_write_tools.py` + both contract xfails, including the measured `unknown method 'threads'` refusal.


Geoffe-Ga:
## Summary
Fixes two real MCP-surface defects (`creek.link` missing `"threads"`; `creek.report` exposing only 6/11 types) by extracting a single shared declaration (`creek/surface_modes.py`) that both `creek.cli` and `creek_mcp/tools/*` import, replacing two independently-maintained tuples that had silently drifted. It also closes a genuine privacy gap: the four newly-reachable report types whose generators accept no `PrivacyTierOverride` (`unnamed`, `fingerprint`, `paradox`, `synchronicity`) are gated behind `privacy_tier_ceiling='all'` and refused by name below it, and a pre-existing CLI bug (`--include-tier` silently ignored for `wavelength`) is fixed as a byproduct via a new shared `generate_phase_map`.

## Strengths
- The fix targets the *mechanism* (retyped copies) rather than patching the two missing strings — genuinely eliminates the drift class rather than papering over this instance of it.
- The two `xfail` parity tests are rewritten to compare what each surface **tells a caller** (parsed rejection messages) rather than importing the same constant twice, so a reintroduced private copy would fail on behavior, not vacuously pass because both sides import the same object.
- The privacy reasoning for the four tier-blind types is sound and verified: `generate_paradoxes`, `generate_synchronicities`, and `UnnamedDigestGenerator` have zero tier filtering internally, so gating them behind `ceiling='all'` and refusing-by-name below that is the correct, not overly conservative, choice.
- `generate_phase_map`/`resolve_period` are correctly extracted into `creek/generate/wavelength.py` and now thread `PrivacyTierOverride` through both the CLI and MCP wavelength paths (fixing the silently-ignored `--include-tier` bug), verified by reading `load_fragments_from_vault`'s existing `override` parameter and confirming both call sites pass it explicitly.
- Test coverage is thorough and well-targeted: reachability tests, surface-completeness tests (`set(_MCP_REPORTS) == set(REPORT_TYPES)`), refusal-message-content tests, and a defensive-branch test for the (currently unreachable-via-`report_tool`) `_generate_wavelength(period=None)` path.
- Docs (`docs/mcp.md`, `docs/wiring-contract.md`) are updated to describe the new derivation and the tier-blind gating accurately.

## Security Concerns
None found. 🟢 The privacy-tier gating for the four newly-MCP-reachable report types was verified against the actual generator signatures (`build_fingerprint`, `generate_paradoxes`, `generate_synchronicities`, `UnnamedDigestGenerator.generate_weekly_digest` — none accept a `PrivacyTierOverride`), and the refusal-below-`ceiling=all` logic in `_tier_blind_refusal` is correctly wired and tested (`test_report_names_tier_blind_types_rather_than_hiding_them`).

## Problems
No blocking issues found in the diff.

One non-blocking but worth-fixing inconsistency: `creek/cli.py`'s docstrings/comments for `_report_unnamed` (L2703), `_report_fingerprint` (L2987), `_report_paradox` (L3273), and `_report_synchronicity` (L3305) each still say "`X` is not exposed over MCP" and "`del override  # not MCP-exposed`" — but this PR's entire point is that `creek_mcp/tools/report.py` now routes all four of these over MCP (gated behind `ceiling='all'`). These four comments were not touched by the diff but are directly falsified by it. Given the PR's own narrative is specifically about a stale docstring ("date arithmetic the MCP shape should not own") lying about MCP-exposure and causing #1253, leaving four fresh instances of the identical lie is worth a follow-up fix before or shortly after merge.

## Code Quality
- `_ReportRequest` + a `dict[str, Callable]` dispatch table (`_MCP_REPORTS`) is a clean way to keep a uniform signature despite `wavelength` needing an extra `period` argument — directly addresses the shape that let `wavelength` fall out of the surface inventory before.
- `test_cli_fill.py`'s switch from a hand-written lambda wrapper to `rec("report/wavelength")` directly (since `rec`'s inner `_step` already accepts `*args, **kwargs`) is a good catch — the old lambda's fixed 2-arg signature would have silently pinned the old 2-arg `_report_wavelength` shape and been the "third place to update" the commit message warns about.

## CLAUDE.md Adherence
- ✅ Commit message is a well-formed Conventional Commit (`fix(mcp): ...`) with issue references.
- ✅ TDD narrative is credible: PR description states RED (17 failures + 2 xfails removed) before GREEN, matching the shape of the diff (dispatch tables, refusal tests, parity tests all added alongside the implementation).
- ❔ Could not execute `./scripts/check-all.sh`, `pytest`, `mypy`, or `ruff` in this review sandbox (test/tool execution and network `gh api`/`git fetch` calls were blocked), so coverage/lint/type-check/complexity gates were not independently re-verified — only read statically. The PR description claims "Gate: 12/12. Zero suppressions."; this should be confirmed via CI status before merge if not already green.
- ✅ No shortcuts observed: no `# noqa`, `# type: ignore`, or skipped tests in the diff.

## Testing
Strong. New tests directly exercise the previously-broken behavior (`test_link_reaches_the_threads_linker`), assert surface completeness structurally (`test_every_cli_report_type_has_an_mcp_branch`, `test_tier_blind_types_are_a_subset_of_the_declared_surface`), assert the privacy refusal by content rather than just status (`test_report_names_tier_blind_types_rather_than_hiding_them` checks the reason string, not just `"refused"`), and cover the `wavelength` period-validation path from both the tool-level refusal and the defensive inner-function branch. The rewritten parity tests (`test_cli_and_mcp_agree_on_link_methods`, `test_cli_and_mcp_agree_on_report_types`) are a meaningful upgrade over the strict-xfails they replace since they can no longer trivially pass once both sides import the same tuple.

## Documentation
Mostly excellent — `creek/surface_modes.py`'s module docstring is an exemplary explanation of *why* the module exists, and `docs/mcp.md`/`docs/wiring-contract.md` are updated accurately for the new behavior. The one gap is the four stale "not exposed over MCP" docstrings in `creek/cli.py` noted under Problems, which this PR should have updated since it's the direct cause of their inaccuracy.



## Verdict: COMMENTS

Nothing here blocks merge — the security gating for the newly-reachable report types was independently verified against the actual generator signatures and is correct, and test coverage is thorough. The substantive non-blocking suggestion is that four docstrings/comments in creek/cli.py (`_report_unnamed`, `_report_fingerprint`, `_report_paradox`, `_report_synchronicity`) now falsely claim "not exposed over MCP" as a direct consequence of this very PR, which is the same class of stale-docstring bug the PR's own description calls out as the root cause of #1253.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5230428218 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5230428218 to see in-depth feedback and continue iterating

Geoffe-Ga:
Fixed in the latest push — a deserved catch, and one that would have been self-refuting to defer.

All four (`_report_unnamed`, `_report_fingerprint`, `_report_paradox`, `_report_synchronicity`) said *"X is not exposed over MCP"* plus `del override  # not MCP-exposed`, and this PR routes all four over MCP. Given the PR's own argument is that a stale docstring lying about MCP exposure is **what caused #1253**, leaving four fresh instances of the identical lie would have undercut the whole change.

Each now states the real constraint rather than a negation that is no longer true:

> `override`: Unused. `unnamed` **IS** served over MCP, but only at `privacy_tier_ceiling=all` — `UnnamedDigestGenerator.generate_weekly_digest` accepts no `PrivacyTierOverride`, so a narrower ceiling cannot be honoured and is refused by name rather than silently widened (#968).

The generator a reader would have to widen is named **in situ**, so the path from "why is this refused?" to "what would have to change" is one hop rather than a hunt. That is the same property the refusal message itself carries.

Gate re-run: **12/12**.

On the two Code Quality notes — thank you, both were deliberate and it is useful to have them confirmed as legible from outside. `_ReportRequest` + `_MCP_REPORTS` exists precisely so `wavelength`'s extra `period` argument cannot make it fall out of a uniform inventory again, which is how it slipped the surface in the first place. And the `test_cli_fill.py` lambda was exactly the "third place to update" the commit message warns about — it pinned a 2-arg shape that would have silently gone stale.

Geoffe-Ga:
**CI**: 12/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5230428218 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Summary
This PR fixes two real MCP/CLI parity bugs (#1252, #1253) by introducing `creek/surface_modes.py` as the single source of truth for `creek link --method` and `creek report --type` vocabularies, replacing two independently hand-maintained tuples that had silently drifted (`creek_mcp.tools.link` had lost `"threads"`; `creek_mcp.tools.report` exposed 6 of 11 types). It also converts two previously-passing "strict xfail" tests into real behavioral parity tests, wires four previously MCP-unreachable report types (`unnamed`, `fingerprint`, `paradox`, `synchronicity`) behind a `ceiling=all`-only gate to avoid the privacy leak #968 closed, fixes a pre-existing CLI bug where `--include-tier` was silently ignored for `wavelength`, and corrects several stale docstrings a self-review commit caught (that the PR itself would otherwise have re-introduced).

## Strengths
- The core insight — that two hand-retyped tuples are the *mechanism* producing both bugs, not two unrelated typos — is correct, and the fix (one shared declaration in `surface_modes.py`) addresses the mechanism rather than papering over the symptom.
- The parity tests (`test_cli_and_mcp_agree_on_link_methods`, `test_cli_and_mcp_agree_on_report_types`) were rewritten to compare each surface's *own rejection message* rather than the shared constant against itself — correctly avoiding the tautology that "derive, don't retype" would otherwise introduce into the regression tests.
- Tier-blind types (`unnamed`, `fingerprint`, `paradox`, `synchronicity`) are refused **by name** with the specific generator symbol to widen, rather than silently omitted — directly addressing the failure mode that caused #1253 in the first place, and correctly avoiding re-opening #968's leak by never serving them below `ceiling=all`.
- `_OVERRIDE_CALL_NAMES`/`test_production_report_callers_always_state_an_override` (existing structural AST guard) was correctly extended to cover the new `generate_phase_map` call site in both `creek.cli` and `creek_mcp.tools.report`, verified directly against the diff — both call sites do state `override=` explicitly.
- The self-correcting second commit (fixing four stale "not exposed over MCP" docstrings) is a good sign of the author reviewing their own diff for exactly the class of bug this PR is about.
- Test coverage is thorough and behaviorally meaningful: full-surface coverage tests, tier-blind refusal tests (parametrized over all four types), unparseable-period refusal, and the `_MCP_REPORTS`/`REPORT_TYPES` set-equality guard that turns a future undispatched type into a hard test failure rather than a silent gap.

## Security Concerns
🟢 LOW — No new leak. The four tier-blind types are correctly gated to `ceiling=all` only, matching the ceiling `creek report` itself already runs under; the refusal path is exercised before any vault I/O per `report_tool`'s docstring, and verified in the diff (`_tier_blind_refusal` runs before `_MCP_REPORTS[report_type]` is invoked).

## Problems
None blocking. One real, verified correctness issue worth fixing before/soon after merge (see Code Quality).

## Code Quality
🟡 **Config-resolution drift newly introduced by this PR, in the exact class of bug it argues against.** In `creek/cli.py`, `_report_unnamed`, `_report_paradox`, and `_report_synchronicity` all resolve config via `_load_config_for_vault(vault_path)`, which auto-discovers `<vault>/00-Creek-Meta/creek_config.yaml` (issue #322) regardless of the process's CWD or `CREEK_CONFIG` env var. The new MCP counterparts this PR adds — `_generate_unnamed`, `_generate_paradox`, `_generate_synchronicity` in `creek_mcp/tools/report.py` — instead call bare `load_config()`, which falls back to `CREEK_CONFIG` or the process CWD's `creek_config.yaml`, *not* the vault passed in `request.vault_path`. (`_generate_fingerprint` is fine — its CLI counterpart `_report_fingerprint` also uses bare `load_config()`, so there's no drift there.)

Concretely: an MCP server started as `creek-tools-mcp --vault /path/to/vault` without `--config` (the server's own `--config` help text: *"Pass --config to pin a config file regardless of the working directory the server is launched from"* — implying this is otherwise CWD-dependent) will silently fall back to default `embeddings`/`ai_style` config for these three report types instead of the vault's own settings, with only a `WARNING` log to notice by. This is the same "two surfaces, two behaviors" drift class #1252/#1253 exist to kill, just at the config-resolution layer rather than the mode-list layer, and it's in code this PR wrote fresh (these three functions didn't exist in `report.py` before this PR). It isn't caught by any existing test — `_OVERRIDE_CALL_NAMES`'s structural guard checks `PrivacyTierOverride` propagation, not config-path resolution — and it doesn't affect privacy-tier correctness, so it's not a leak, just silently-wrong embeddings/style config in a subset of deployments.

Suggested fix: `load_config(resolve_config_path(request.vault_path, None))` (matching `creek.cli._load_config_for_vault`'s behavior) in the three affected `_generate_*` functions.

Minor, non-blocking: `creek_mcp/tools/classify.py` still has its own `_VALID_METHODS = ("rules", "llm")` retyped tuple, structurally identical to the pattern this PR just eliminated for link/report — out of this PR's stated scope (#1252/#1253 name only link and report), so not a request to fix here, just worth a follow-up issue if not already filed.

## CLAUDE.md Adherence
✅ Docstrings are thorough (Google-style, Args/Returns) on every new/changed public and private function — consistent with the ≥95% interrogate threshold.
✅ The refactor from an if/elif chain to a `_MCP_REPORTS` dispatch dict plus small per-type `_generate_*` functions appears deliberately structured to keep per-function cyclomatic complexity low, consistent with the ≤10 threshold.
✅ Commit messages follow Conventional Commits (`fix(mcp): ...`, `docs: ...`) and both commits describe the "why," matching repo convention.
❓ Could not run `./scripts/check-all.sh`, `pytest`, mypy, ruff, or coverage locally in this review sandbox (command execution was blocked), so line/branch coverage, mypy strict, and ruff zero-violations claims are taken from the PR description ("Gate: 12/12. Zero suppressions.") rather than independently verified. Static reading of the diff shows no obvious type or lint violations.

## Testing
✅ Not insufficient — this is a well-tested change. New/changed tests cover: MCP reaching `threads` for `link`, full `REPORT_TYPES` surface coverage at `ceiling=all`, tier-blind refusal-by-name for all four gated types (parametrized), unparseable-period refusal, the `_MCP_REPORTS`==`REPORT_TYPES` set-equality guard, and the two xfail→real parity tests that now assert on each surface's own rejection message rather than a shared constant compared with itself (correctly avoiding a tautological assertion). `test_report_refuses_unsupported_type`'s use of `wavelength` as its "unsupported" example (which had pinned the #1253 bug as correct behavior) was correctly repointed at a genuinely unknown type name.

## Documentation
✅ `docs/mcp.md` and `docs/wiring-contract.md` were updated in the same PR and read as accurate against the code (spot-checked the `creek.report`/`creek.link` table rows and the "Closed drift" section against the actual dispatch/refusal logic — consistent). `creek/surface_modes.py`'s module docstring clearly states the module's one job and why it imports nothing (avoiding pulling CLI/MCP dependencies into either frontend's import path).



## Verdict: COMMENTS

Nothing here blocks merge — the parity fix is correct, well-tested, and closes both filed bugs without reopening the #968 privacy leak. But the three new MCP report branches (`unnamed`, `paradox`, `synchronicity`) resolve config via bare `load_config()` while their own CLI counterparts use `_load_config_for_vault(vault_path)`, a verified, substantive inconsistency in code this PR wrote fresh that reintroduces the same "two surfaces disagree silently" pattern the PR's own thesis argues against — worth the author's judgment before or shortly after merge.

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5230556379 to see in-depth feedback and continue iterating

Issue #1265, opened 2026-08-09
crawdad: _channel_tier does not resolve a thread to its parent, so a parent's intimate ceiling is not applied inside its threads


Geoffe-Ga:
Found by the security review on #1052.

## Problem

`crawdad/crawdad/bot.py::_channel_tier` resolves the operator's declared privacy
ceiling by keying `AttachmentConfig.channel_privacy_tiers` on
`message.channel.id`:

```python
return config.attachments.channel_privacy_tiers.get(channel_id, "personal")
```

For a message posted **in a Discord thread**, `message.channel.id` is the
*thread's* id, not the parent channel's. So an operator who declares parent
channel `555` as `intimate` and then allowlists a thread inside it gets
`_channel_tier(<thread id>) -> "personal"` — the parent's declared ceiling is
silently not applied to its own thread.

This is a `_channel_tier` modelling gap, not a defect in any one caller, and it
predates #1052. It currently affects three shipped call sites:

- `bot.py` `_record_pending_batch` — the `privacy_tier_ceiling` threaded into
  `creek.ingest` for staged attachments (FEAT-027/034).
- `bot.py` `_run_safety_scan` — the ceiling on the `creek.redact.scan` call.
- `bot.py` `_capture_allowed` — the new #1052 bot-capture tier gate, which would
  admit capture for an intimate parent's thread.

**Not a regression from #1052** and bounded by the fact that it requires the
operator to explicitly add the thread id to `allowed_channel_ids` (threads are
not auto-allowlisted). But it is a de-escalation in the one direction the
project treats as non-negotiable: a ceiling the operator declared is not
enforced.

## Scope

Make `_channel_tier` resolve a thread to its parent channel before the lookup,
so all three callers harden in one edit. Does NOT change the tier vocabulary,
the allowlist semantics, or `CAPTURE_ADMITTED_TIERS`.

## Tasks

1. Resolve the effective channel id: when the channel is a thread, use its
   `parent_id` (`discord.Thread.parent_id`) for the `channel_privacy_tiers`
   lookup, falling back to the channel's own id when there is no parent.
2. Fail closed: if the parent cannot be determined, keep the current `personal`
   default — never widen.
3. Prefer the **most restrictive** of the thread's own declared tier and its
   parent's, so an operator can tighten a single thread inside an `open`
   channel without the parent silently loosening it. Note this needs a tier
   ordering; `crawdad.loop` already owns the single ordering table
[...]

Issue #1264, opened 2026-08-09
crawdad: audit/purge helper for pre-#1052 bot-capture data (non-allowlisted + intimate records still on disk and still ingestible)


Geoffe-Ga:
Found by the security review on #1052. **#1052 stops future bad writes; it does
nothing about what a pre-fix build already wrote.**

## Problem

Any operator who ran a build before #1052 with `capture_enabled: true` has a
`<vault>/discord-capture/` tree containing:

- messages from users never added to `allowed_user_ids`,
- messages from channels never added to `allowed_channel_ids`,
- other bots' and webhooks' output, and
- content from channels the operator declared `intimate` in
  `channel_privacy_tiers` — all of it **untiered**.

That data is still live, not inert:

1. `creek-tools/creek/ingest/discord.py::stage_capture_as_data_package`
   (`discord.py:819-831`) iterates **every** channel dir under the capture root
   unconditionally — there is no allowlist or tier filter on the read side.
2. So `ingest_capture_dir` (`discord.py:864-889`) will still land the pre-fix
   intimate content as `unclassified`, which ranks **with** `personal`
   (`creek_mcp/tier_ceiling.py`, #961) — the exact de-escalation #1052 closed
   going forward.
3. `crawdad/crawdad/capture.py::_seen_ids` (`capture.py:165-173`) seeds its
   dedup set from those files, so the stale records read as "already captured"
   and are never revisited or corrected.

Blast radius is bounded — `capture_enabled` defaults to `False`
(`crawdad/crawdad/config.py`) — but it is not zero, and the operator has no way
to tell what is in there short of reading JSONL by hand.

**Manual mitigation today:** delete the offending channel directories under
`<vault>/discord-capture/` before running `creek sync`.

## Scope

An audit/purge helper for an existing capture tree, plus the operator
documentation for it. Does NOT change the live capture gate (#1052 owns that)
or the capture file format.

## Tasks

1. An audit mode that reports, per channel dir under the capture root: whether
   the channel is currently allowlisted, its declared `channel_privacy_tiers`
   ceiling, record count, and date range — i.e. exactly what a pre-fix build
   may have left behind.
2. A purge mode (explicit opt-in, never default, dry-run first) that removes
   records the current gate would refuse: non-allowlisted channels, and
   channels declared `intimate`/`all`.
3. Decide the home: a `crawdad` CLI subcommand (it owns `CAPTURE_ADMITTED_TIERS`
[...]

Issue #1262, opened 2026-08-09
crawdad+creek: carry the declared privacy tier end-to-end through bot-capture so intimate channels can be captured without de-escalating


Geoffe-Ga:
Spun out of #1052 (P0 bot-capture gate fix), which had to **refuse** capture for
channels the operator declared `intimate` or `all` because the capture path
cannot carry a privacy tier. This issue is the work that would make those
channels capturable *without* de-escalating them.

## Problem

The bot-capture path loses the operator's declared privacy ceiling by
construction, at three consecutive hops:

1. `crawdad/crawdad/capture.py::_record_for` emits only
   `id` / `timestamp` / `content` / `author` (+ `reference`). No tier, and no
   bot flag.
2. `creek-tools/creek/ingest/discord.py::stage_capture_as_data_package` writes
   `channel.json` as `{"id": <folder>, "name": <folder>}` only — the channel id
   the operator keyed `channel_privacy_tiers` on is already gone (the folder is
   the channel *name*, per `capture.py::_channel_label`).
3. `creek/ingest/discord.py` contains zero occurrences of `privacy` or `tier`,
   so the resulting fragments land `unclassified` — which ranks **with**
   `personal` (`creek-tools/creek_mcp/tier_ceiling.py`, #961).

Net: anything bot-capture writes arrives downstream at personal-equivalent
sensitivity, whatever the operator declared. That is why #1052 closed the hole
by refusing `intimate`/`all` rather than by persisting a tier — a tier field
that ingest drops would make the ceiling *look* honoured while the fragment
still landed `unclassified`, which is worse than an honest refusal.

## Scope

The end-to-end tier carry: capture record -> Data-Package staging ->
`DiscordIngestor` -> fragment frontmatter. Once it lands, `CAPTURE_ADMITTED_TIERS`
in `crawdad/crawdad/config.py` can be widened (with its own tests) so an
`intimate` channel is captured *and* correctly tiered.

Note this crosses the crawdad/creek-tools boundary, so it needs **both** quality
gates: `crawdad/scripts/check-all.sh` and `creek-tools/scripts/check-all.sh`.

## Tasks

1. Persist the declared tier on the capture record (`capture.py::_record_for`),
   sourced from `channel_privacy_tiers`, and persist the channel **id** so the
   tier is re-derivable at ingest time.
2. Carry it through `stage_capture_as_data_package` into `channel.json` (and/or
   per-message), preserving the real channel id.
[...]

PR #1268, opened 2026-08-09
fix(crawdad): route bot-capture through the allowlist, bot filter and tier gate (#1052)

### LINKED ISSUE #1052: crawdad: bot-capture bypasses both allowlists, the bot filter, and channel_privacy_tiers
**Labels:** `de-slop`, `crawdad-bot`, `bug`, `priority-high`
**Detected by:** de-slopify targeted scan (crawdad-bot) 2026-07-31
**Severity:** High
**Remediation:** decision-needed (Family 3 aspirational guarantee / Family 5 inconsistent contract)

## Problem

`crawdad/CLAUDE.md` §2 and `crawdad/README.md` name the user + channel allowlist
as the enforcement boundary — *"non-allowlisted users get **no** response (silent
ignore, per FEAT-013's personal-use scoping)"*. Bot-capture honours **none** of
the gates the rest of the bot applies:

1. **Allowlist bypass.** `crawdad/crawdad/bot.py:868` runs
   `await self._capture_message(message)` **before** `handle_message`, which is
   where `_passes_allowlist` lives (`bot.py:222`, `:274-276`).
   `_capture_message` (`bot.py:885-902`) filters on **nothing** but
   `message.author.id == self.user.id`.
2. **Bot/webhook messages captured.** `_passes_allowlist` drops all bots via
   `message.author.bot`; `_capture_message` does not, so other bots' and
   webhooks' output lands in the vault and flows into Tier-A ingest as if it were
   human conversation.
3. **`channel_privacy_tiers` ignored.** The attachment path reads the
   operator-declared per-channel ceiling (`bot.py:710-718 _channel_tier`,
   defaulting to `personal`) and threads it into every `creek.ingest`.
   `capture._record_for` (`capture.py:102-113`) emits only
   `id`/`timestamp`/`content`/`author` — no tier — and `_append` performs no
   channel lookup at all.

**Current state:** with `capture_enabled: true`, plain-text messages from
channels and users the operator never allowlisted — and from channels the
operator marked `intimate` — enter the vault untiered.

## Evidence (corroboration)

- **Signal 1 — an existing, passing test pins the bypass:**
  `crawdad/tests/test_bot_capture.py:60-67` builds the config with
  `allowed_channel_ids=[999]`, `_FakeChannel` defaults to `id=1`
  (`test_bot_capture.py:36`), and
  `test_captures_and_forwards_to_command_path` asserts the JSONL **was** written
  for that non-allowlisted channel (`:108-109`). The hole is currently encoded as
  intended behaviour.
- **Signal 2 — the ceiling is unrecoverable downstream:**
  `creek-tools/creek/ingest/discord.py` contains zero occurrences of `privacy`
  or `tier`, and `stage_capture_as_data_package` (`discord.py:806-811`) writes
  `channel.json` as `{"id": <folder>, "name": <folder>}` only. Once capture
  writes the line the operator's declared ceiling is unrecoverable by
  construction. `capture.py:104-109` also persists no bot flag, so a later ingest
  pass cannot filter what capture let through either.

## Scope

Covers the three gates on the capture path. Does NOT change the capture file
format, the `capture_enabled` default (`False`, which caps blast radius), or the
creek-tools ingest side.

## Tasks

1. **Decide and record the semantics** (this is the decision-needed part):
   either (a) route capture through `config.is_allowed(...)` + `author.bot` +
   `channel_privacy_tiers` so it honours the same boundary as the command path,
   or (b) document the deliberate exemption explicitly in `crawdad/CLAUDE.md` §5
   and §5.1 and `crawdad/README.md`. Option (a) is the default recommendation —
   the allowlist is documented as *the* scoping boundary.
2. Implement the chosen option in `crawdad/crawdad/bot.py:885-902` (and
   `capture.py` if the tier must be persisted per record).
3. Fix `crawdad/tests/test_bot_capture.py` so its fixture channel id is
   allowlisted, and add tests pinning: a non-allowlisted channel is **not**
   captured; a non-allowlisted user is **not** captured; another bot's message is
   **not** captured.
4. Add an e2e-shaped test through the real `CrawDadClient.on_message` proving the
   chosen semantics end to end.

## Acceptance Criteria

- [ ] The chosen semantics are stated in `crawdad/CLAUDE.md` §5 and
      `crawdad/README.md`, and the code matches them.
- [ ] Under option (a): messages from non-allowlisted channels/users and from
      other bots produce no JSONL record (tests pin all three).
- [ ] `channel_privacy_tiers` is either applied to capture or its exemption is
      documented.

## Files to Create/Modify

| File | Action |
|------|--------|
| `crawdad/crawdad/bot.py` | Modify |
| `crawdad/crawdad/capture.py` | Modify |
| `crawdad/tests/test_bot_capture.py` | Modify |
| `crawdad/CLAUDE.md`, `crawdad/README.md` | Modify |

Geoffe-Ga:
## Summary

- **Bot-capture now answers to the same boundary as the command path.** It ran ahead of `handle_message`, so it never reached `_passes_allowlist` and filtered on nothing but the bot's own user id — messages from users and channels the operator never allowlisted, plus other bots' and webhooks' output, were written into `<vault>/discord-capture/` and flowed into Tier-A ingest as human conversation. `_capture_allowed` now requires `_passes_allowlist` to pass first.
- **`channel_privacy_tiers` is applied to capture, fail-closed.** A capture record carries no tier field and `stage_capture_as_data_package` drops channel metadata, so a captured message lands downstream as `unclassified` — which ranks *with* `personal` (#961). Capturing a channel the operator declared `intimate` was a silent privacy de-escalation. Capture is now admitted only for tiers in `CAPTURE_ADMITTED_TIERS` (`{"open", "personal"}`); `intimate` and `all` are refused, because capture must not carry content whose ceiling it cannot represent.
- **The semantics are documented** in `crawdad/CLAUDE.md` §5 / new §5.2 and `crawdad/README.md`, per the issue's first acceptance criterion. This is option (a) of the issue's decision — the allowlist is documented as *the* scoping boundary, so exempting capture from it was never a decision anyone made.

## Design notes

**Strictly narrowing at every ceiling.** Across the 5×5 tier × actor matrix, **17 cells narrow, 8 are unchanged, 0 widen**. The new predicate is the old one conjoined with two more conditions, so widening is structurally impossible. The one admitted-but-inexact cell still escalates: an `open` channel's message lands as `unclassified` (rank 1) rather than `open` (rank 0) — more restrictive, not less.

**Allow-list, not deny-list.** `CAPTURE_ADMITTED_TIERS` is written out explicitly rather than derived as `_VALID_CHANNEL_TIERS - {"intimate", "all"}`. A subtraction would auto-admit any fifth tier added to the vocabulary later; the safe default for an unreviewed tier is refusal. `test_capture_admitted_tiers_is_a_strict_subset_of_valid_tiers` fails the moment a tier is added without an explicit admission decision.

**Membership, never a rank comparison** — `crawdad.loop` owns the single tier-ordering table and a second one must not exist (the rule `workflows.py` already states). As a bonus this fails closed on a value that never passed config validation, where a rank comparison would not.

**The gate is inside `_capture_message`'s existing `try`.** A gate that raises (a malformed message, an attribute that errors) therefore fails closed to "not captured" while the command path keeps running — fail-closed by construction rather than by convention. Pinned by `test_gate_failure_refuses_capture_and_keeps_commands_alive`.

**The tier gate is capture-scoped only.** `channel_privacy_tiers` has never gated whether the bot *replies* in a channel and still does not — an operator's `intimate` journal channel still gets `/crawdad` and free-text replies, it is just never written to the capture log. `test_end_to_end_intimate_channel_replies_but_does_not_capture` fails if that scope ever creeps into `handle_message`.

**`capture.py` is deliberately untouched.** Persisting a tier on the record would change the capture file format (out of scope per the issue) and be inert anyway, because the creek-tools ingest side that drops it is also out of scope — worse, an inert field would make the ceiling *look* honoured while the fragment still landed `unclassified`. #1262 tracks carrying the tier end-to-end, after which `CAPTURE_ADMITTED_TIERS` may widen.

## Known residual risk (please read before merging)

**This fix stops future bad writes; it does not remediate what a pre-fix build already wrote.** An operator who ran with `capture_enabled: true` may have non-allowlisted, other-bot, and `intimate` records already on disk under `<vault>/discord-capture/`, and those remain ingestible — `stage_capture_as_data_package` iterates every channel dir unconditionally. Blast radius is bounded by `capture_enabled` defaulting to `False`. Filed as **#1264** (audit/purge helper) and documented in `crawdad/README.md` under "Upgrading from a pre-#1052 build"; the manual mitigation today is deleting the offending channel subdirectories before running `creek sync`.

## Test plan

**Gate 1 — RED proven before any implementation landed.** Ran the new tests against unmodified `crawdad/crawdad/`: **10 failed, 46 passed**, each for the right reason (a JSONL file that should not exist; `ImportError` on the not-yet-added constants) — not a fixture or typo failure. New coverage:

- non-allowlisted channel → not captured
- non-allowlisted user → not captured
- another bot's message → not captured (both allowlists satisfied, so only the `bot` flag can refuse)
- **`intimate` channel → not captured** — the narrowest ceiling, with every other gate passing
- **`all` channel → not captured**
- a gate that raises → not captured *and* the command path still runs
- `open` / `personal` / unset channels → **still captured** (parametrised over-narrowing guard: an over-broad "refuse everything" fix passes all five refusal tests and fails only here)
- three e2e-shaped tests through the real `CrawDadClient.on_message` **without** monkeypatching `handle_message`, so capture and the command path are proven to consult the same object
- two config invariant pins on `CAPTURE_ADMITTED_TIERS`, plus one linking `DEFAULT_CHANNEL_TIER` to it

Also repairs the #687 fixture that **encoded the bug**: it set `allowed_channel_ids=[999]` while `_FakeChannel` defaulted to `id=1`, and `999` doubled as the bot's own user id — so "the allowlisted channel was captured" could never actually be observed. Verified against `git show HEAD:` that no pre-existing assertion was deleted or loosened; all 11 survive, one strengthened.

**Gate 2 — the CrawDad gate, run the way CI's "CrawDad Quality" job runs it** (`working-directory: crawdad`):

```
cd crawdad && ./scripts/check-all.sh
```

→ **7/7 checks pass** (lint, format, type check, security, docstrings, complexity, tests) · **598 passed, 3 deselected** · coverage **94.62%** (threshold 90). Lint, format and mypy independently re-verified with `--no-cache`.

Note `creek-tools/scripts/check-all.sh` does **not** cover `crawdad/`, and this change touches no creek-tools code, so the crawdad gate is the applicable one.

**Gate 2.5** — `security-specialist` returned `SECURITY VERDICT: CLEAN` (no blockers) after enumerating every path to the writer and walking the full tier × actor matrix; `code-review-orchestrator` returned `CLEAN` (no blockers). Both flagged non-blocking comment-precision items, which are fixed in this branch — including one factually wrong claim about `StrEnum` hashing that I disproved empirically before rewriting the comment.

## Follow-ups filed

- **#1264** (P1) — audit/purge helper for pre-#1052 capture data still on disk.
- **#1262** (P2) — carry the declared tier end-to-end (capture record → staging → fragment frontmatter) so `intimate` channels can be captured faithfully instead of refused.
- **#1265** (P2) — `_channel_tier` does not resolve a Discord thread to its parent channel, so a parent's `intimate` ceiling is not applied inside its threads. Pre-existing; also affects the shipped attachment-staging and safety-scan paths.
- **#1057** — commented with the decision and a standing constraint: when `MessageCapture.backfill` is wired in, it must evaluate `_capture_allowed` **per message inside the history loop**, not once per channel, or it re-opens this entire bypass at history scale.

Closes #1052

### DIFF
--- a/crawdad/CLAUDE.md
+++ b/crawdad/CLAUDE.md
@@ -108,6 +108,15 @@ Four gates, each must pass before the next:
   - `allowed_channel_ids` — channels the bot will respond in.
   - `max_loop_rounds` — optional FEAT-036 override for the agent-loop
     round cap (default `MAX_LOOP_ROUNDS = 5`, bounded `[1, 50]`).
+  - `capture_enabled` — bot-capture toggle (#687), default `False`.
+    Opt-in per deployment; see [§5.2](#52-bot-capture-boundary).
+  - `capture_subpath` — vault-relative dir the capture writer appends
+    to, default `discord-capture`. Must stay inside the vault (same
+    absolute/`..` validation as `attachments.staging_subpath`).
+  - `attachments.channel_privacy_tiers` — per-channel declared ceiling
+    (`open` / `personal` / `intimate` / `all`), validated at
+    config-parse time. See [§5.2](#52-bot-capture-boundary) for how
+    bot-capture reads it.
 
 Both allowlists must be non-empty — an empty list is a configuration
 error, not "open to everyone".
@@ -125,6 +134,63 @@ it means:
 - Do *not* feed user-controlled content into `crawdad.yaml`.
 - Multi-user / shared deployments are out of scope for v1.0; revisit
   this assumption before broadening access.
+
+See also [§5.2 Bot-capture boundary](#52-bot-capture-boundary) for the
+narrower trust boundary bot-capture applies on top of this one.
+
+### 5.2 Bot-capture boundary
+
+Bot-capture (#687) writes messages to `<vault>/<capture_subpath>/` for
+later Tier-A ingest. A message is captured only when **both** hold:
+
+1. It passes `_passes_allowlist` — the same gate the command path
+   uses: not the bot's own message, not `author.bot`, and both
+   `allowed_user_ids` and `allowed_channel_ids` admit it. **The
+   capture boundary is the command boundary** (#1052) — before this
+   fix, capture ran ahead of the allowlist gate and logged strangers,
+   other bots, and channels the operator never allowlisted.
+2. The channel's declared privacy tier (`_channel_tier`, reading
+   `attachments.channel_privacy_tiers`) is in `CAPTURE_ADMITTED_TIERS`
+   = `{"open", "personal"}`:
+
+   | Declared tier | Captured? |
+   |---|---|
+   | `open` | yes (narrower than needed once landed) |
+   | `personal` | yes (exact match) |
+   | unset (no `channel_privacy_tiers` entry) | yes — `_channel_tier` falls back to `personal` |
+   | `intimate` | **no** |
+   | `all` | **no** — admits intimate content by definition |
+
+   A capture record carries no tier field, and the creek-tools side
+   that stages the capture dir drops channel metadata, so a captured
+   message lands downstream as `unclassified`, which ranks *with*
+   `personal` (`creek_mcp/tier_ceiling.py`, #961). Writing an
+   `intimate` channel into capture would be a silent privacy
+   de-escalation — capture must not carry content whose ceiling it
+   cannot represent.
+
+The tier gate is **capture-scoped only**. `channel_privacy_tiers` has
+never gated whether the bot *replies* in a channel and still does not
+— an `intimate` channel still gets `/crawdad` and free-text replies;
+it is just never written to the capture log.
+
+The gate (`_capture_allowed`) is evaluated *inside*
+`CrawDadClient._capture_message`'s `try`, so a gate that raises (a
+malformed message, an attribute that errors) fails closed to "not
+captured" while the command path keeps running unaffected — capture
+is best-effort and must never take a command turn down with it.
+
+**Standing constraint for future work:** `MessageCapture.backfill` has
+no production caller today. If it is ever wired in (#1057), it MUST
+evaluate `_capture_allowed` **per message inside the history loop**,
+not once per channel — a channel's history contains messages from
+non-allowlisted users and other bots that a channel-level check would
+wave through.
+
+`CAPTURE_ADMITTED_TIERS` may only widen once #1262 lands: carrying the
+tier end-to-end from the capture record through staging into fragment
+frontmatter, so an `intimate` channel can be captured faithfully
+instead of refused outright.
 
 ## 6. MCP subprocess resilience
 
--- a/crawdad/README.md
+++ b/crawdad/README.md
@@ -18,7 +18,9 @@ CrawDad v1.0 ships:
 - Voice-skill activation per session from `<vault>/creek-skills/`.
 - The six `/crawdad` slash commands (FEAT-016): `reflect`, `checkin`,
   `surface`, `draft`, `save`, `workflow`.
-- A user + channel allowlist; non-allowlisted callers get no response.
+- A user + channel allowlist; non-allowlisted callers get no response
+  — and, as of #1052, no bot-capture record either (see
+  [Bot capture](#bot-capture) below).
 - A graceful "creek-tools is unreachable" reply when the MCP
   subprocess dies.
 
@@ -69,8 +71,34 @@ allowed_channel_ids:
 | `attachments` | no | see source | Per-attachment limits, allow/deny lists, channel privacy tiers (FEAT-027/035). |
 | `consent` | no | see source | Conversational consent tokens + TTL (FEAT-034). |
 | `max_loop_rounds` | no | `5` | FEAT-036 agent-loop round cap, bounded `[1, 50]`. Raise when a single user turn legitimately needs more router/dispatcher passes (multi-source ingest, large re-classification asks, long workflow chains); the upper bound prevents pathological runaway loops. |
+| `capture_enabled` | no | `False` | Bot-capture toggle (#687) — see [Bot capture](#bot-capture) below. |
+| `capture_subpath` | no | `discord-capture` | Vault-relative dir bot-capture writes into. Must stay inside the vault. |
 
 Secrets are **never** in `crawdad.yaml` — they come only from the environment.
+
+### Bot capture
+
+Bot-capture logs the channels CrawDad is in to `<vault>/<capture_subpath>/`
+for later ingest — it is off by default (`capture_enabled: false`). When on,
+the bot captures **only messages it would also respond to**: the same
+user + channel allowlists as the command path, never another bot or
+webhook, and never its own replies.
+
+It also **never** captures a channel declared `intimate` or `all` in
+`attachments.channel_privacy_tiers` — those channels still get normal
+`/crawdad` and free-text replies, they are just never logged to the
+vault, because a captured message currently can't carry its channel's
+privacy tier with it. #1262 tracks carrying that tier end-to-end so
+`intimate` channels can be captured faithfully instead of refused.
+
+**Upgrading from a pre-#1052 build.** If you ran with
+`capture_enabled: true` before this fix, `<vault>/discord-capture/`
+may already hold records from non-allowlisted users, other bots, or
+an `intimate` channel — the old capture path bypassed all three
+gates. These records are still ingestible today. #1264 tracks the
+fix; until then the manual mitigation is deleting the offending
+channel subdirectories under `discord-capture/` before running
+`creek sync`.
 
 ### LLM provider selection
 
--- a/crawdad/crawdad/bot.py
+++ b/crawdad/crawdad/bot.py
@@ -29,6 +29,7 @@ from crawdad.attachments import (
     format_attachment_summary,
     process_attachments,
 )
+from crawdad.config import CAPTURE_ADMITTED_TIERS, DEFAULT_CHANNEL_TIER
 from crawdad.consent import (
     PendingBatch,
     PendingBatchStore,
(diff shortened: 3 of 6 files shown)


Geoffe-Ga:
## Summary
Closes #1052: bot-capture previously ran ahead of `handle_message` and applied only a self-message check, so it wrote messages from non-allowlisted users/channels, other bots, and `intimate`-tier channels into `<vault>/discord-capture/`. This PR routes capture through the same `_passes_allowlist` gate as the command path (`_capture_allowed`) and adds a new tier gate (`CAPTURE_ADMITTED_TIERS = {"open", "personal"}`) so capture can never write content whose privacy ceiling the untiered capture record can't represent. It also repairs a pre-existing #687 test fixture bug that made the original "captured channel" assertion unfalsifiable.

## Strengths
- The fix is genuinely minimal and additive: `_capture_allowed` composes the existing `_passes_allowlist` with one new membership check, so the "17 narrow / 8 unchanged / 0 widen" claim in the PR body is easy to verify by reading the diff.
- Gate is evaluated *inside* `_capture_message`'s existing `try/except`, giving fail-closed behavior for free (an exploding attribute access refuses capture, not the command path) — and it's pinned by `test_gate_failure_refuses_capture_and_keeps_commands_alive`.
- `CAPTURE_ADMITTED_TIERS` is written as an explicit allow-list rather than `_VALID_CHANNEL_TIERS - {"intimate", "all"}`, so a fifth tier added later fails closed by default; `test_capture_admitted_tiers_is_a_strict_subset_of_valid_tiers` pins this.
- Test suite is unusually thorough: covers each individual gate (channel allowlist, user allowlist, `author.bot`, `intimate`, `all`), an over-narrowing guard (parametrized `open`/`personal`/unset must still capture), and three end-to-end tests through the real `on_message` without monkeypatching `handle_message`, so capture and the command path are proven to consult the same allowlist decision rather than two independently-maintained copies.
- The #687 fixture bug fix (`allowed_channel_ids=[999]` colliding with the fake channel's default `id=1` and the bot's own id) is a legitimate, well-explained repair — the original assertion could never actually observe a rejection.
- Documentation (`crawdad/CLAUDE.md` §5.2, `README.md`) is added alongside the code change and accurately describes the new boundary, including the capture-scoped-only caveat (an `intimate` channel still gets replies, just isn't logged) and the standing constraint for `#1057`'s future `backfill` wiring.
- The residual-risk section is transparent about what this fix does *not* do (pre-fix capture files already on disk remain ingestible) and files a tracked follow-up (#1264) rather than silently declaring victory.

## Security Concerns
🟢 LOW — This PR is itself a privacy/security fix and closes a real information-boundary bug (non-consented capture, other-bots' output, and silent tier de-escalation of `intimate` channels reaching Tier-A ingest as `unclassified`). No new vulnerabilities introduced. The one residual gap (pre-existing bad capture records on disk) is explicitly called out, bounded by `capture_enabled` defaulting to `False`, and tracked in #1264 rather than silently left — acceptable to scope out of this PR.

## Problems
None found blocking merge. `_capture_allowed`, `_channel_tier`, and the `on_message`/`_capture_message` wiring were traced by hand and are internally consistent: `bot_user_id` is only ever passed after `self.user is not None` is confirmed in `on_message`, and there is exactly one call site for `_capture_message`, so the new required kwarg doesn't leave a dangling caller.

## Code Quality
- The block comments in `config.py` around `CAPTURE_ADMITTED_TIERS` (lines ~203–233) are quite long, including an aside proving `StrEnum` values would hash-match a plain string. This is defensible under "only comment on non-obvious WHY," since it forestalls a plausible future "let's just derive this from a subtraction" refactor, but it's borderline verbose for a single constant — a shorter version pointing at the pinning test would read just as clearly. Not blocking.
- Minor: `CAPTURE_ADMITTED_TIERS` is explicitly noted as coincidentally equal to `workflows.WORKFLOW_ADMITTED_CEILINGS` but deliberately not derived from it — worth double-checking in a future pass that the two don't silently drift apart in a way that surprises an operator (e.g., a channel admitted to the cloud composer but never to capture, or vice versa), though the PR's own reasoning for keeping them separate (different ceilings entirely) is sound.

## CLAUDE.md Adherence
✅ TDD workflow followed — PR body documents RED (10 failed/46 passed against unmodified code) before GREEN.
✅ `crawdad/CLAUDE.md` quality gates reportedly run via `./scripts/check-all.sh` (598 passed, 94.62% coverage vs. 90% threshold) — I was unable to independently execute the test suite in this review sandbox (script execution required approval that wasn't available), so this is based on the PR's self-report plus static reading of the diff, not independent re-execution.
✅ Docs updated in the same PR (`crawdad/CLAUDE.md` §5.2, `README.md`) per the "Documentation" gate.
✅ Commit message follows Conventional Commits (`fix(crawdad): ...`).
✅ No `# type: ignore`/`# noqa`/shortcut suppressions introduced.
✅ Membership-based tier check (not a rank comparison) respects the existing rule that `crawdad.loop` owns the single tier-ordering table.

## Testing
Strong. Six new capture-boundary tests plus an end-to-end class (3 tests) in `test_bot_capture.py`, and three new invariant-pinning tests in `test_config.py`. Notably includes a dedicated over-narrowing guard (`test_admitted_tier_channels_are_captured`) that would fail a "refuse everything" degenerate fix — a detail many PRs of this shape omit. Fail-closed behavior on a raising gate is explicitly tested. No gaps identified in the tier × actor matrix coverage described in the PR body.

## Documentation
Complete and precise. `crawdad/CLAUDE.md` §5.2 lays out the full admitted/refused tier table and the capture-vs-reply scope distinction; `README.md` adds an operator-facing "Bot capture" section including an "Upgrading from a pre-#1052 build" migration note for anyone who ran the vulnerable version. Docstrings on `_capture_allowed`, `_capture_message`, and the new config constants are accurate and cross-reference each other correctly.



## Verdict: LGTM

The fix is minimal, strictly narrowing (verified by hand-tracing `_capture_allowed`/`_channel_tier`/the single `_capture_message` call site), thoroughly tested including a fail-closed case and an over-narrowing guard, and documented in both `crawdad/CLAUDE.md` and `README.md` per the repo's workflow. No blocking bugs, security issues, or missing tests were found; the only remarks are comment-verbosity nits I would not block on.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234052082 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #526, opened 2026-06-03
fix(crawdad): agent loop cannot converge — router re-runs on identical inputs every round


Geoffe-Ga:
## Summary

The CrawDad agent loop cannot converge: the Haiku router is re-invoked with **byte-identical inputs every round**, so once it decides a tool call is needed (`compose=false`), it keeps deciding the same thing until the round cap is hit and the turn is abandoned with the `_TOO_DEEP_REPLY` soft error. Turns only succeed when the router happens (by LLM non-determinism) to return `compose=true` or empty intents on a pass.

## Evidence (from a live demo-vault run)

Repeated runtime warnings, each preceded by exactly `max_loop_rounds` (5) MCP `CallToolRequest`s:

```
WARNING crawdad.loop: agent loop exceeded 5 rounds without compose=true; resetting session history
```

Three back-to-back exhaustions were observed as the user retried the same request. The user experience is the bot replying "I went too deep on this — let's back up…" to ordinary requests.

## Root cause

`crawdad/crawdad/loop.py`, `AgentLoop.run` (lines ~160–196):

- `loop.py:162` appends the user turn to `self._history` **once**, before the loop.
- `loop.py:165-171` each round calls `self._router.extract_intents(message=message, history=self._history, state=self._session_state)` with the **same** `message` and a `history` that is never mutated inside the loop.
- `loop.py:187` accumulates tool results into a local `aggregated` list that is **never** appended to `self._history` nor passed to `extract_intents`. It is only handed to the composer (`loop.py:204-211`) — which is unreachable when the loop exhausts first.

Because the router's inputs are identical on rounds 1..N, a deterministic-ish Haiku response that sets `compose=false` with non-empty intents recurs every round → the `for…else` exhaustion branch (`loop.py:188-196`) fires, history is cleared, and the user gets `_TOO_DEEP_REPLY`.

The router prompt (`router.py`) even instructs Haiku to set `compose:true` when "prior tool results in the history include a paradox surfacing" — but prior tool results are never placed in the history, so that instruction can never fire.

## Suggested fix direction

Feed each round's tool results back into the router's context before the next `extract_intents`:

- Append a synthetic `assistant`/`tool` turn carrying the round's `aggregated` results to `self._history` inside the loop, **or**
- Pass `aggregated`/prior-results into `extract_intents` and render them in `build_router_prompt` (which currently has no tool-results block).

[...]

PR #1269, opened 2026-08-09
fix(crawdad): cap the LLM router path at the cloud-composer ceiling

### LINKED ISSUE #1152: crawdad: the LLM router path can request `intimate`/`all` with nothing capping it
Follow-up to #1051, and the larger half of the same defect.

#1051 capped the **workflow** path at `{open, personal}` (`WORKFLOW_ADMITTED_CEILINGS`
in `crawdad/crawdad/workflows.py`) because every `ToolResult` body is relayed to a
cloud LLM composer and then posted into a Discord message. The **router** path is
still uncapped, and its input is LLM-generated rather than operator-authored.

## The chain

1. `crawdad/crawdad/intents.py:60` — `Intent.privacy_tier_ceiling` accepts all four
   `PrivacyTierCeiling` members.
2. `crawdad/crawdad/intents.py:132-136` — `build_intents_schema` **publishes all four
   to the router** as a valid enum. The schema the router is told its output must
   validate against literally advertises `intimate` and `all` as selectable.
3. `crawdad/crawdad/dispatcher.py:239-241` — `_build_arguments` forwards
   `intent.privacy_tier_ceiling.value` verbatim with no membership test. Its docstring
   only claims to stop the router smuggling a looser tier through `args`; it stops the
   smuggled one and forwards the *declared* one, which is the same escalation by the
   front door.
4. Server side: CrawDad reaches the MCP server over **stdio**, so
   `creek_mcp/server.py::_caller_identity` returns `is_remote=False` and
   `creek_mcp.policy.admitted_ceiling` returns an unconditional `Admission` — the
   `REMOTE_ADMITTED_CEILINGS` cap never fires for CrawDad. Nothing caps this path.

## Why it is reachable, not just theoretical

`crawdad/crawdad/loop.py:264` (`_record_tool_round`) appends `_render_tool_results(results)`
— raw MCP tool-result bodies, i.e. vault fragment text — into `ConversationHistory`, and
`crawdad/crawdad/router.py:93-98,142` (`build_router_prompt`) renders that history
**verbatim** into the router prompt. Vault fragments are built by `creek ingest` from
third-party exports (ChatGPT / Discord / Substack), so attacker-authored text reaches the
router prompt. An injected line in a round-1 body suffices; round 2 dispatches at
`intimate`, the body goes to the cloud composer, and the reply goes to Discord.

**Amplifier:** `crawdad/crawdad/loop.py:329-347` (`_max_ceiling_from`) is a monotone-upward
ratchet within a turn, so one `intimate`-labelled result raises the auto-injected
`creek.save` to `intimate`. Separately, `dispatcher._handle_run_workflow` stamps the
*intent's* ceiling onto the workflow `ToolResult` even though the walk itself ran capped —
so a router-declared `intimate` on a `crawdad.run_workflow` intent produces an
`intimate`-labelled result whose body was actually fetched at `open`, and that mislabel
feeds the same ratchet.

## Why this is not a copy of #1051's fix

A blanket `{open, personal}` cap on `Intent` is probably **wrong**: `crawdad/crawdad/bot.py:399,671`
deliberately calls at `_channel_tier(...)`, which `crawdad/crawdad/config.py:301` permits to
be `intimate`, and that attachment-ingest path does not go through the composer. The likely
right shape is a cap on the composer-fed path plus a narrowing of `build_intents_schema`'s
published enum. That is a design decision and needs its own spec.

Per #526 the loop *depends* on tool bodies in history to converge, so the fix must be at the
ceiling, not by removing bodies from the router prompt.

## Acceptance criteria

- [ ] A test proves an `intimate`/`all` router intent cannot reach `session.call_tool` on the
      composer-fed path against unfixed HEAD (RED first).
- [ ] `build_intents_schema` no longer advertises ceilings the dispatcher will refuse.
- [ ] The `_handle_run_workflow` `ToolResult` is stamped with the ceiling the walk actually
      used, not the intent's.
- [ ] Reuse the existing ordering table in `crawdad/crawdad/loop.py:329`; do not add a second one.
- [ ] Do not add a new privacy-tier reader (#1079).

Geoffe-Ga:
## Summary

- **Caps the LLM router path at the cloud-composer boundary.** CrawDad's Haiku router emits a `privacy_tier_ceiling` per intent, and `IntentDispatcher` forwarded it verbatim to `session.call_tool`. Nothing capped it: CrawDad reaches the MCP server over **stdio**, so `creek_mcp/server.py::_caller_identity` reports `is_remote=False` and `creek_mcp.policy.admitted_ceiling` returns an unconditional `Admission` — the server's `REMOTE_ADMITTED_CEILINGS` cap never fires for CrawDad. Every intent now passes through `_capped()` at the head of `IntentDispatcher.dispatch`'s loop, narrowing to `{open, personal}` before any branch acts on it.
- **Stops the schema advertising tiers the dispatcher will not honour.** `build_intents_schema` published all four tiers as a selectable enum; it now publishes `["open", "personal"]` in rank order, derived from the admitted set rather than hard-coded.
- **Makes the `ToolResult` privacy stamp honest**, so `loop._max_ceiling_from`'s upward-only ratchet can no longer be fed a lie: `_handle_run_workflow` stamps the ceiling the **walk** actually used (via a new `WorkflowRunReport`) instead of the router's unrelated guess, and the branches that make no tool call at all — `_handle_activate_register` and both `run_workflow` soft-error paths — stamp `open`.
- **One tier table, one admitted set.** `CEILING_RANK` moves from `loop.py` to `intents.py` (forced: `dispatcher` cannot import `loop` without a cycle) and `loop._max_ceiling_from` now reads it. `workflows.WORKFLOW_ADMITTED_CEILINGS` becomes an identity alias of the new `intents.COMPOSER_ADMITTED_CEILINGS` — the workflow cap and the router cap are the same boundary, and two agreeing frozensets would be two places to widen with only one showing up in a reviewer's diff.

### Why it was reachable, not theoretical

`loop._record_tool_round` appends raw MCP tool-result bodies — vault fragment text that `creek ingest` built from third-party ChatGPT / Discord / Substack exports — into `ConversationHistory`, and `router.build_router_prompt` renders that history **verbatim** into the router prompt. Attacker-authored text therefore reaches the router. An injected line in a round-1 body was enough: round 2 dispatched at `intimate`, the body went to the cloud composer, and the reply was posted into Discord. Per #526 the loop *depends* on those bodies to converge, so the fix is at the ceiling, not by stripping the prompt.

### Clamp, not refuse — and why

A router intent above the cap is **clamped down to `personal` and logged at `WARNING`**; it is not refused. `AgentLoop.run` catches only `UnknownIntentError` and `MCPUnavailableError`, so a new exception would escape `handle_message` and the bot would go silent — one poisoned vault fragment, which lives in the vault and re-enters the router prompt every turn, would become a permanent DoS. Clamping gives identical confidentiality (the call proceeds at `personal`) plus an operator-visible warning. The authored-workflow path still **refuses**, unchanged: there the operator wrote the file and a refusal is a visible, fixable Discord reply.

The cap is enforced at the dispatch chokepoint rather than as a pydantic validator on `Intent`, for two independent reasons: `model_construct` / `model_copy(update=...)` sail straight past validators (the argument `workflows.WorkflowWalker._check_privacy_ceiling` already writes down for the workflow cap), and `router._parse_router_payload` converts any `ValidationError` into `RouterParseError`, which `loop.py` turns into a whole-turn "I lost the thread" reply — a model-level narrowing would have traded the leak for an injection-driven DoS.

### Privacy invariant

Every change is a strict narrowing; nothing becomes less restrictive at any ceiling. `cap_ceiling` returns the broadest admitted tier **≤** the request (`open`→`open`, `personal`→`personal`, `intimate`→`personal`, `all`→`personal`), so an `open` request is never raised to `personal`. The `<=` filter — rather than "return the highest admitted tier" — means it cannot widen even under a hypothetical non-contiguous admitted set, and `key=CEILING_RANK.__getitem__` avoids the `StrEnum` trap where a bare `max()` compares spelling and ranks `all` below everything.

`bot.py` has **zero production-line changes**. Its two direct `session.call_tool` sites (`:399`, `:671`) call at `_channel_tier(...)`, which `config.py:301` permits to be `intimate` — they never construct an `Intent`, never cross the dispatcher, and are operator-authored `crawdad.yaml` config rather than router output, so they are correctly left uncapped. The issue body's "a blanket cap on `Intent` is probably wrong" paragraph assumed those sites were coupled to `Intent`; they are not, so no separate spec was owed.

## Test plan

Gate 1 RED was proven against unfixed HEAD before any production line was written — **50 failures**, with the security-carrying ones failing on genuine assertions rather than import errors:

- `tests/test_loop.py::test_router_emitted_intimate_intent_never_reaches_call_tool` — `assert {'all', 'intimate'} <= {'open', 'personal'}`
- `tests/test_loop.py::test_paradox_save_is_never_injected_above_the_cap` — `assert 'all' in {'open', 'personal'}`
- `tests/test_dispatcher.py::test_dispatch_never_forwards_a_ceiling_above_the_composer_cap[intimate|all]` — `assert 'intimate' == 'personal'`
- `tests/test_intents.py::test_intents_schema_advertises_only_composer_admitted_ceilings` — `assert ['open','personal','intimate','all'] == ['open','personal']`

The narrowest-ceiling case is pinned explicitly: `test_dispatch_never_widens_any_requested_ceiling` is parametrized over all four tiers and its `open` / `personal` parameters were **green before the fix and stay green after**, so a clamp mis-implemented as "always return `personal`" fails the suite. `test_cap_ceiling_maps_each_tier_to_an_exact_result` pins the whole mapping by value.

Also added: `test_activate_register_result_is_stamped_open`, `test_run_workflow_result_uses_walk_ceiling_not_intent_ceiling`, `test_run_workflow_soft_error_results_are_stamped_open`, `test_dispatch_logs_a_warning_when_it_caps` (which asserts the log carries no tool-result body — the #1090 refusal-as-oracle hazard), `test_max_ceiling_from_uses_the_shared_rank_table`, and `test_ceiling_rank_covers_every_member_and_open_is_the_minimum`.

One existing assertion changed value: `test_loop.py::test_loop_paradox_save_inherits_max_ceiling` asserted the injected `creek.save` went out at `"intimate"` — that was the leak. It now asserts `== "personal"` **and** keeps its original guard explicitly as `!= "open"`, so the tightening cannot be mistaken for a weakening.

**Gate run: `crawdad/scripts/check-all.sh` only.** The diff is entirely inside `crawdad/`, which has its own gate and its own CI job; `creek-tools/scripts/check-all.sh` does not cover it and nothing under `creek-tools/` changed. Result: **7/7 checks passed, 630 passed, 94.56% branch coverage**. Ruff lint, ruff format and mypy re-verified with cold caches (`--no-cache` / `--no-incremental`) because a stale ruff cache has produced a false-clean gate in this repo before. No `# noqa`, no `# type: ignore`, no skips, no lowered thresholds, no deleted tests.

### Gate 2.5 self-review

A code-review pass and an explicit security pass both returned CLEAN with no blockers. Four sub-blocker findings were applied rather than deferred, all fail-closed hardening:

- `cap_ceiling` now *looks up* the rank instead of subscripting it, so a tier string that names no member clamps to `open` rather than raising `KeyError`. The raise was safe (it precedes any tool call) but escaped the agent loop and silenced the bot — the same DoS the clamp exists to avoid.
- `_handle_run_workflow` re-caps the ceiling it receives from the injected runner, the one value the dispatcher took on trust rather than deriving. Idempotent; makes `_capped`'s "only gate" claim literally true.
- The new WARNING logs `intent.type` with `%r`. `_capped` runs *before* the `known_tools` check, so that string is arbitrary router-emitted input with no charset constraint; unquoted, an embedded newline could forge log records.
- `workflows._execute_one_step` re-caps its stamp so it is self-evidently within the cap at its own line rather than by reference to a guard several frames up.

One review claim was **wrong and is worth recording**: the security pass argued that a raw `"all"` string would miss the rank table because `Enum` hashes by member *name*. It does not — `PrivacyTierCeiling` is a `StrEnum`, so members hash and compare as their *value* (`hash("all") == hash(PrivacyTierCeiling.ALL)`), and an uncoerced `"all"` is correctly clamped to `personal`. My first version of the guard's test encoded the reviewer's claim and failed, which is how the error surfaced. Both branches are now pinned explicitly — `test_cap_ceiling_treats_a_raw_tier_string_as_its_member` and `test_cap_ceiling_clamps_an_unrecognised_tier_to_open` — so the distinction cannot silently invert into a rank table that has stopped recognising its own spellings.

## Sync point

#1052 (crawdad bot-capture bypass) is in flight on `crawdad/crawdad/bot.py`. This PR makes **no production change** to that file — it only consumes the retyped `WorkflowDispatchRunner` alias — and touches `tests/test_bot.py` solely to retype two fake workflow runners. Conflicts should be trivial.

Closes #1152
Refs #1051, #526, #1090, #1079

### DIFF
--- a/crawdad/README.md
+++ b/crawdad/README.md
@@ -159,6 +159,26 @@ rename at your leisure; `intimate`/`all` values now get refused at
 requiring intimate protection); both keys present fails to parse and
 the workflow drops out of the listing.
 
+**The same cap covers the router**, not just authored workflows. The
+Haiku router also emits a `privacy_tier_ceiling` per intent, and its
+prompt contains raw tool-result bodies (vault fragments `creek ingest`
+built from third-party exports), so that value is untrusted input. The
+dispatcher therefore clamps every intent to the same `{open, personal}`
+set — one shared constant, `crawdad.intents.COMPOSER_ADMITTED_CEILINGS`
+— before any MCP call is made. Nothing else would: CrawDad speaks MCP
+over stdio, so the server sees a *local* caller and applies no remote
+cap.
+
+Note the difference in behaviour. A workflow declaring `intimate`/`all`
+is **refused** (the operator authored the file and can fix it, and a
+refusal is a visible Discord reply). A router intent above the cap is
+**clamped down to `personal` and logged at `WARNING`** — the call still
+goes out, just narrowed. Refusing there would let one poisoned vault
+fragment silence the bot for good, since the agent loop does not catch
+that error class; clamping gives the same confidentiality plus an
+operator-visible warning. Neither path can ever *widen* a ceiling: an
+`open` request stays `open`.
+
 **No per-step ceilings** — a step's `args:` may not set its own
 `privacy_tier_ceiling` (or the deprecated `privacy_tier_floor`); the
 workflow-level value applies to every step, and a step that tries is
--- a/crawdad/crawdad/cli.py
+++ b/crawdad/crawdad/cli.py
@@ -23,8 +23,9 @@ from crawdad.config import (
     router_model_for,
 )
 from crawdad.consent import PendingBatchStore
+from crawdad.dispatcher import WorkflowRunReport
 from crawdad.history import ConversationHistory
-from crawdad.intents import ToolInfo
+from crawdad.intents import PrivacyTierCeiling, ToolInfo, cap_ceiling
 from crawdad.llm import build_async_provider
 from crawdad.loop import run_one_turn
 from crawdad.mcp_client import MCPClient, MCPUnavailableError
@@ -186,18 +187,32 @@ def _build_workflow_runner(
     mcp_client = components.mcp_client
     known_tools = components.known_tools
 
-    async def _runner(name: str, inputs: dict[str, str]) -> str:
+    async def _runner(name: str, inputs: dict[str, str]) -> WorkflowRunReport:
         """Run *name* by looking it up in the registry and composing the reply.
 
         Surface failures (constraint errors, unknown name, composer
         exceptions) as a single Discord-safe string so the slash command
         can pass them through without translating again.
+
+        The report also carries the ceiling the walk used (#1152), which
+        the dispatcher stamps onto its :class:`ToolResult`. Both failure
+        branches report ``open``: a soft-error string is CrawDad's own
+        wording, no vault content was selected, and reporting the
+        workflow's declared tier would ratchet the agent loop on the
+        strength of a failure. Only the success branch reports the
+        workflow's tier, run through :func:`cap_ceiling` so this report
+        cannot exceed the cloud-composer cap even if a future
+        construction route slips an ``intimate`` definition past the
+        walker's own pre-flight refusal.
         """
         try:
             workflow = registry.get(name)
         except Exception as exc:
             _LOGGER.warning("workflow lookup failed for %r: %s", name, exc)
-            return _truncate_for_discord(f"could not find workflow `{name}`: {exc}")
+            return WorkflowRunReport(
+                reply=_truncate_for_discord(f"could not find workflow `{name}`: {exc}"),
+                privacy_tier_ceiling=PrivacyTierCeiling.OPEN,
+            )
         try:
             reply = await run_workflow_and_compose(
                 workflow=workflow,
@@ -211,8 +226,14 @@ def _build_workflow_runner(
             )
         except Exception as exc:
             _LOGGER.warning("workflow %r failed mid-run: %s", name, exc)
-            return _truncate_for_discord(f"workflow `{name}` failed: {exc}")
-        return _truncate_for_discord(reply)
+            return WorkflowRunReport(
+                reply=_truncate_for_discord(f"workflow `{name}` failed: {exc}"),
+                privacy_tier_ceiling=PrivacyTierCeiling.OPEN,
+            )
+        return WorkflowRunReport(
+            reply=_truncate_for_discord(reply),
+            privacy_tier_ceiling=cap_ceiling(workflow.privacy_tier_ceiling),
+        )
 
     return _runner
 
--- a/crawdad/crawdad/config.py
+++ b/crawdad/crawdad/config.py
@@ -223,8 +223,10 @@ _VALID_CHANNEL_TIERS: Final[frozenset[str]] = frozenset(
 # record can represent. The reasoning above stands on its own if the workflow
 # set ever changes.
 #
-# Membership (``in``), never a rank comparison: ``crawdad.loop`` owns the
-# single tier-ordering table and a second one must not exist. The values stay
+# Membership (``in``), never a rank comparison: ``crawdad.intents`` owns the
+# single tier-ordering table (``CEILING_RANK``, moved there from
+# ``crawdad.loop`` by #1152 so the capping rule sits beside the vocabulary it
+# caps) and a second one must not exist. The values stay
 # plain ``str`` rather than ``intents.PrivacyTierCeiling`` members because
 # ``channel_privacy_tiers`` is a ``dict[int, str]`` of operator-written YAML
 # and this module deliberately owns the tier vocabulary without importing
--- a/crawdad/crawdad/dispatcher.py
+++ b/crawdad/crawdad/dispatcher.py
@@ -6,9 +6,17 @@ invokes the matching tool with the intent's args. Tool results come
 back as a list of :class:`ToolResult` envelopes that the bot handler
 formats into a Discord reply (FEAT-014's "boring summary" path).
 
-FEAT-015 will wrap this dispatcher in the 5-round loop; the
-``privacy_tier_ceiling`` field is forwarded verbatim because the MCP
-server enforces the policy at the protocol boundary (FEAT-010/011).
+FEAT-015 wraps this dispatcher in the 5-round loop.
+
+The dispatcher is also the chokepoint for the cloud-composer privacy cap
+(#1152). The router's ``privacy_tier_ceiling`` used to be forwarded
+verbatim on the theory that the MCP server enforced the policy at the
+protocol boundary (FEAT-010/011) — but CrawDad speaks MCP over **stdio**,
+which ``creek_mcp/server.py::_caller_identity`` classifies as a LOCAL
+caller, so the server-side remote cap never fires here. Meanwhile the
+router's ceiling is untrusted: raw tool-result bodies reach the router
+prompt through the conversation history. Every intent therefore passes
+through :func:`_capped` before any branch acts on it.
 """
 
 from __future__ import annotations
@@ -20,8 +28,10 @@ from pydantic import BaseModel, ConfigDict
 
 from crawdad.intents import (
     ACTIVATE_REGISTER_INTENT_TYPE,
+    COMPOSER_ADMITTED_CEILINGS,
     RUN_WORKFLOW_INTENT_TYPE,
     PrivacyTierCeiling,
+    cap_ceiling,
 )
 
 if TYPE_CHECKING:
@@ -37,15 +47,6 @@ if TYPE_CHECKING:
     # ``activate_register`` method down to the dispatcher.
     RegisterSwitcher = Callable[[str], bool]
 
-    # ADAPT-003: async callable that runs a named workflow end-to-end and
-    # returns the composed user-facing reply. Mirrors the slash-command
-    # ``crawdad.slash_commands.WorkflowRunner`` shape so the CLI can hand
-    # the same closure (built by ``crawdad.cli._build_workflow_runner``)
-    # to both the slash-command surface and the dispatcher. Receives the
-    # workflow ``name`` plus an ``inputs`` mapping for ``{{input.<key>}}``
-    # interpolation.
-    WorkflowDispatchRunner = Callable[[str, dict[str, str]], Awaitable[str]]
-
 _LOGGER = logging.getLogger("crawdad.dispatcher")
 
 
@@ -60,10 +61,18 @@ class UnknownIntentError(RuntimeError):
 class ToolResult(BaseModel):
     """One tool's response, threaded back through the loop.
 
-    ``privacy_tier_ceiling`` mirrors the source intent's ceiling so
-    downstream consumers (e.g. the loop's paradox-routing helper) can
-    re-authorise follow-up tool calls at the same or higher tier
+    ``privacy_tier_ceiling`` records the ceiling that actually selected
+    this body, so downstream consumers (e.g. the loop's paradox-routing
+    helper) can re-authorise follow-up tool calls at the same tier
     without rediscovering the source authorization.
+
+    "Actually selected" is the #1152 correction: the stamp is the
+    *capped* ceiling the wire call carried, never the one the router
+    asked for, and branches that make no tool call at all stamp ``open``
+    because their bodies are fixed status lines. ``loop._max_ceiling_from``
+    ratchets the auto-injected ``creek.save`` up to the highest ceiling
+    it sees here, so an over-generous stamp leaks on the *next* call even
+    when the current one was capped correctly.
     """
 
     model_config = ConfigDict(frozen=True)
@@ -71,6 +80,44 @@ class ToolResult(BaseModel):
     intent_type: str
     body: str
     privacy_tier_ceiling: PrivacyTierCeiling = PrivacyTierCeiling.OPEN
+
+
+class WorkflowRunReport(BaseModel):
+    """What an ADAPT-003 workflow runner hands back to its caller.
+
+    A bare reply string was not enough (#1152). The walker caps the
+    workflow's *declared* ceiling itself, and its soft-error paths
+    (unknown name, mid-run failure) read nothing from the vault at all,
+    so the only component that knows which tier actually selected the
+    returned text is the runner. Reporting it explicitly lets
+    :meth:`IntentDispatcher._handle_run_workflow` stamp the truth
+    instead of echoing the router's unrelated guess.
+
+    ``privacy_tier_ceiling`` has no default on purpose: every runner has
+    to state the tier its walk used, and a default would let a new
+    soft-error branch quietly inherit somebody else's answer.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    reply: str
+    privacy_tier_ceiling: PrivacyTierCeiling
+
+
+if TYPE_CHECKING:
+    # ADAPT-003: async callable that runs a named workflow end-to-end and
+    # reports the composed reply plus the ceiling the walk used. Mirrors
+    # the slash-command ``crawdad.slash_commands.WorkflowRunner`` shape so
+    # the CLI can hand the same closure (built by
+    # ``crawdad.cli._build_workflow_runner``) to both the slash-command
+    # surface and the dispatcher. Receives the workflow ``name`` plus an
+    # ``inputs`` mapping for ``{{input.<key>}}`` interpolation.
+    #
+    # Declared here rather than in the import block above only because it
+    # names :class:`WorkflowRunReport`, which has to exist first.
+    WorkflowDispatchRunner = Callable[
+        [str, dict[str, str]], Awaitable[WorkflowRunReport]
+    ]
 
 
 class IntentDispatcher:
@@ -98,11 +145,13 @@ class IntentDispatcher:
                 rather than mutate the skill stack — useful in tests
                 and when the registry isn't wired (no vault layout).
             workflow_runner: ADAPT-003 async callback that walks an
-                authored workflow by name and returns the composed
-                reply. ``None`` causes ``crawdad.run_workflow`` intents
-                to soft-error rather than crash — useful in tests and
-                when the workflow surface isn't wired (no composer /
-                no MCP tools advertised).
+                authored workflow by name and returns a
+                :class:`WorkflowRunReport` — the composed reply plus the
+                ceiling that walk used. ``None`` causes
+                ``crawdad.run_workflow`` intents to soft-error rather
+                than crash — useful in tests and when the workflow
+                surface isn't wired (no composer / no MCP tools
+                advertised).
         """
         self._session = session
         self._known_tools = frozenset(known_tools)
@@ -125,7 +174,15 @@ class IntentDispatcher:
                 soft-error reply.
         """
         results: list[ToolResult] = []
-        for intent in response.intents:
+        for raw_intent in response.intents:
+            # #1152: cap at the head of the loop, before any branch sees
+            # the intent. The two client-side branches make no MCP call
+            # yet still stamp ``ToolResult.privacy_tier_ceiling``, which
+            # feeds ``loop._max_ceiling_from``'s upward-only ratchet — so
+            # an attacker could emit a ``crawdad.activate_register`` at
+            # ``all`` purely to raise the auto-injected ``creek.save``.
+            # Capping here means no present or future branch can forget.
+            intent = _capped(raw_intent)
             if intent.type == ACTIVATE_REGISTER_INTENT_TYPE:
                 results.append(self._handle_activate_register(intent))
                 continue
@@ -153,9 +210,12 @@ class IntentDispatcher:
         """Run the FEAT-029 register switch for ``crawdad.activate_register``.
 
         The result body is a short status line the composer can quote
-        verbatim (or summarise) for the user. The privacy ceiling on the
-        result mirrors the intent's declared ceiling so downstream
-        consumers see a faithful echo of the router's authorisation.
+        verbatim (or summarise) for the user — a register name and fixed
+        wording, never vault content. The result is therefore stamped
+        ``open`` (#1152) rather than echoing the intent's ceiling: no
+        content was read, so no tier is justified, and echoing one would
+        hand ``loop._max_ceiling_from``'s ratchet a free upgrade for a
+        call that touched nothing.
         """
         register = intent.args.get("register")
         if not isinstance(register, str) or not register:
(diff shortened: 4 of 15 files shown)


Geoffe-Ga:
## Summary
This PR closes a real privacy-tier bypass in CrawDad's Discord bot (crawdad/): the Haiku router's `privacy_tier_ceiling` was forwarded verbatim to `session.call_tool`, and because CrawDad talks to the MCP server over stdio (a "local" caller per `creek_mcp/server.py::_caller_identity`), the server's `REMOTE_ADMITTED_CEILINGS` cap never applied. Since raw tool-result bodies (vault fragments) are rendered into the router prompt, an attacker-controlled fragment could request `intimate`/`all` and have it forwarded uncapped. The fix adds a single chokepoint (`dispatcher._capped`) that clamps every intent to `{open, personal}` before dispatch, narrows the advertised schema enum accordingly, and makes `ToolResult.privacy_tier_ceiling` reflect what was actually used (rather than what was requested) so the loop's upward-only ratchet can't be fed a lie.

## Strengths
- The vulnerability is real and was independently verified by reading `creek_mcp/policy.py` and `creek_mcp/server.py`: the remote cap only fires for remote callers, confirming CrawDad's stdio path is genuinely uncapped upstream.
- The "clamp, not refuse" design decision is well-reasoned and matches the actual code: `AgentLoop.run` (crawdad/crawdad/loop.py) only catches `RouterParseError`, `UnknownIntentError`, and `MCPUnavailableError`, so a new raised exception on the hot path would indeed silence the bot — clamping avoids that DoS while preserving confidentiality.
- Excellent single-source-of-truth cleanup: `CEILING_RANK` moved to `intents.py` to break an import cycle, and `WORKFLOW_ADMITTED_CEILINGS` is now a literal identity alias of `COMPOSER_ADMITTED_CEILINGS` (tested via `is`, not `==`), preventing future silent divergence between the two boundaries.
- Test suite is unusually rigorous: property tests (`never widens`), exact-value pinning (`maps_each_tier_to_an_exact_result`), idempotency, log-content safety (asserts no vault body leaks into the WARNING per the #1090 hazard), and end-to-end reachability tests through the full `AgentLoop` proving the attack chain is actually closed, not just unit-tested in isolation.
- Good self-review discipline documented in the PR body, including a correction of the author's own security-pass mistake about `StrEnum` hashing (verified in code — `StrEnum` members do hash/compare by value, so the fix is correct).

## Security Concerns
🟢 LOW — `dispatcher._capped` (crawdad/crawdad/dispatcher.py:311-326) reads `requested = intent.privacy_tier_ceiling` and, when uncapped, logs `requested.value`. This assumes `requested` is always a `PrivacyTierCeiling` enum instance. The function's own docstring argues its value specifically *because* `model_construct`/`model_copy(update=...)` can "sail straight past validators" and leave a raw `str` on `Intent.privacy_tier_ceiling` — the same concern `cap_ceiling` was explicitly hardened against (it uses `CEILING_RANK.get(...)` and is tested with `cast("PrivacyTierCeiling", bogus)`). If a raw, unadmitted string ever did reach `_capped` this way, `requested.value` would raise `AttributeError` — an exception that escapes `AgentLoop.run`'s narrow except clauses, i.e. exactly the DoS class this PR exists to prevent. Today no code path constructs an `Intent` via `model_construct`/`model_copy(update=...)` with a non-enum ceiling (the only production `model_copy` call in `_capped` itself always supplies an already-validated `PrivacyTierCeiling`), so this isn't currently reachable — but it's a real gap relative to the function's own stated threat model, and there's no test exercising `_capped` (as opposed to `cap_ceiling` directly) against a raw string. Cheap fix: use `%s` on `requested`/`capped` directly (both are `str`/`StrEnum`, so `%s` works whether or not `.value` exists) instead of `.value`.

## Problems
None found that block merge.

## Code Quality
- Minor: `_capped`'s docstring claims to hold "for all of them" (all `Intent` construction routes) but the `.value` access on `requested` in the log line is only safe for the pydantic-validated route — see Security Concerns above. Worth a one-line fix for internal consistency with the module's own defense-in-depth stance, even though it's not exploitable today.
- The extensive inline commentary (module docstrings, per-function rationale, cross-references between `intents.py`/`dispatcher.py`/`workflows.py`/`loop.py`) is dense but earns its keep here given the subtlety of the privacy-tier invariant being protected; it reads as intentional documentation of a security boundary rather than noise.

## CLAUDE.md Adherence
✅ Feature branch, not a direct commit to main. ✅ Conventional-commit-style PR title (`fix(crawdad): ...`). ✅ TDD workflow documented and evidenced (RED failures listed with exact assertion diffs before the fix). ✅ Uses the project's own module boundaries correctly (single rank table, no duplicated admitted-set literals). Per crawdad/CLAUDE.md, gate 2 requires `./scripts/check-all.sh` to exit 0 — I was not able to execute the script or the test suite myself in this review environment (all `Bash` invocations beyond trivial builtins required interactive approval that wasn't available), so I could not independently re-run the 630-test/94.56%-coverage claim in the PR description. I did verify the described claims by reading the actual source (caller-identity/remote-cap logic, `AgentLoop.run`'s except clauses, `StrEnum` hashing semantics) and everything checked out consistent with the PR's stated reasoning.

## Testing
Test additions are thorough and target the actual security property, not just line coverage: parametrized "never widens" tests across all four tiers, exact-value pinning of the full `cap_ceiling` mapping, idempotency, a regression test for the pre-existing leak (`test_loop_paradox_save_inherits_max_ceiling`, now asserting `personal` with an explicit `!= "open"` guard so the tightening can't be mistaken for a weakening), and full-loop reachability tests proving the attack chain from a scripted "poisoned" router response through to the wire-level `session.call_tool` args. No gaps stood out for the changed code paths; the one untested edge case is the `_capped`/raw-string interaction noted above, which is narrow enough not to be blocking.

## Documentation
README.md and module docstrings were updated to explain the new clamp-vs-refuse behavior for both the router path and authored workflows, consistent with the code. Docstring density looks well above the project's 95% interrogate threshold based on what's shown in the diff.



## Verdict: COMMENTS

Nothing here blocks merge — the security fix is real, well-verified against the actual code, and backed by an unusually rigorous test suite — but there is one substantive, specific suggestion worth weighing: `dispatcher._capped`'s WARNING log calls `.value` on `intent.privacy_tier_ceiling`, which the function's own docstring says may be a raw (non-enum) `str` via `model_construct`/`model_copy(update=...)` bypass; if that ever happens with an unadmitted tier, `.value` raises `AttributeError`, escaping `AgentLoop.run`'s narrow except clauses — the exact DoS class this PR is designed to prevent. It isn't reachable via any current production code path, so it doesn't block, but it's inconsistent with the PR's own stated defense-in-depth goal and worth a one-line fix (`%s` instead of `.value`).

Geoffe-Ga:
**CI**: 10/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5234109708 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5234109708 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Addressed: the `.value` access in `_capped`'s WARNING (0a50e5a)

Taking the suggestion as written — `%s` on `requested` / `capped` instead of `.value`. You are right on both the mechanism and the framing: `_capped` exists as a dispatch-path function rather than an `Intent` validator *precisely because* `model_construct` / `model_copy(update=...)` sail past validators, and its docstring claimed to therefore hold "for all of them" while the very next lines assumed an enum instance. `cap_ceiling` was hardened against exactly this input class and this line quietly undid it one statement later — an `AttributeError` past `AgentLoop.run`'s three except clauses is the DoS the clamp was added to prevent.

`PrivacyTierCeiling` is a `StrEnum`, so `%s` renders the bare tier identically for a member and for an uncoerced `str`; the existing `\ball\b` / `\bpersonal\b` word-boundary assertions in `test_dispatch_logs_a_warning_when_it_caps` still pass unchanged, which is the corroboration that the log text did not shift.

**Docstring reconciled in the same commit.** "A function on the dispatch path holds for all of them" now continues: "— which is only true if this body never assumes it has an enum instance, so the WARNING below formats the tiers with `%s` rather than reading `.value` off them." The claim and the code now agree, and there is an inline comment at the log call saying why `.value` must not come back.

**Test added, and proven RED first.** You noted correctly that nothing exercised `_capped` itself against a raw string — `cap_ceiling`'s `cast("PrivacyTierCeiling", bogus)` tests call it directly and never reach the log line. New `test_capped_clamps_an_uncoerced_raw_string_ceiling_without_raising` builds the `Intent` through the `model_copy(update=...)` route the docstring names and covers both branches:

| raw ceiling | expected |
|---|---|
| `"intimate"`, `"all"` | `personal` (well-spelled, unadmitted) |
| `"ALL"`, `""`, `"not-a-tier"` | `open` (unnameable → maximally suspicious) |

Against the pre-fix source all five fail with `AttributeError: 'str' object has no attribute 'value'` at `dispatcher.py:324`; after the fix all five pass. The test asserts the clamp value, that nothing raises, that exactly one WARNING is emitted, and — pinning the #1090 hazard on this path too — that a sentinel planted in the router-supplied `args` never reaches the record. `args` is where vault fragments rendered into the router prompt would arrive, so that is the leak worth pinning here.

Note `%s` is safe on the *tier* operands but the intent type deliberately keeps `%r` — it is an arbitrary router-emitted string reaching this line before the `known_tools` membership check, so it stays quoted against log forging.

**Gate 2 (crawdad's own gate, `crawdad/scripts/check-all.sh`): 7/7 passed, 635 tests, 94.56% branch coverage.** Ruff lint + format re-verified with `--no-cache` to rule out a stale-cache false clean. No assertions were removed or loosened — the test-file diff against `HEAD` is additive only.

Scoped strictly to this one finding; nothing else in the PR changed.

Geoffe-Ga:
## Summary
This PR closes a real privacy leak in CrawDad: the Haiku router's untrusted `privacy_tier_ceiling` was forwarded verbatim to `session.call_tool`, and because CrawDad talks to the MCP server over stdio the server-side remote cap (`creek_mcp.policy.admitted_ceiling`) never fires to catch it. The fix installs a single chokepoint (`dispatcher._capped`) that clamps every intent to `{open, personal}` before any branch acts on it, makes `ToolResult.privacy_tier_ceiling` an honest record of the tier that actually selected the returned content (via a new `WorkflowRunReport`), and consolidates the tier-rank table into `intents.py` to break an import cycle and eliminate a second copy that could silently drift.

## Strengths
- **Root-caused, not patched.** The fix sits at the actual chokepoint (dispatch, before any branch), not scattered ad hoc checks, and the PR explicitly rules out a pydantic validator with a correct, verified reason (`model_construct`/`model_copy(update=...)` bypass validators).
- **Clamp vs. refuse is well-reasoned.** Refusing would let a single poisoned vault fragment produce a permanent DoS, since `AgentLoop.run` only catches `UnknownIntentError`/`MCPUnavailableError`. Clamping preserves confidentiality while staying available — and this same class of failure (an uncaught `AttributeError`/`KeyError` on a raw string) is what the second commit and `cap_ceiling`'s `.get()`-based lookup explicitly go on to fix.
- **Verified rather than assumed technical claims.** I independently confirmed `creek_mcp/server.py::_caller_identity` returns `is_remote=False` for stdio callers (no access token) and that `bot.py`'s two direct `call_tool` sites use operator-configured `_channel_tier` rather than router-derived `Intent`s — both match the PR description exactly.
- **Single source of truth enforced by identity, not equality.** `WORKFLOW_ADMITTED_CEILINGS is COMPOSER_ADMITTED_CEILINGS` and `loop_module.CEILING_RANK is intents_module.CEILING_RANK` are pinned with `is`, specifically to catch the failure mode where two "agreeing" copies drift apart later.
- **#1090 log-injection hazard closed on both paths.** The WARNING in `_capped` interpolates only the intent type and the two tier names — never `args` — and this is pinned by a sentinel-based test (`VAULT-BODY-SENTINEL`) on both the direct-`_capped` and full-dispatch paths.
- **Excellent regression discipline.** The changed assertion in `test_loop_paradox_save_inherits_max_ceiling` keeps its original guard (`!= "open"`) alongside the new value (`== "personal"`) so the tightening can't be mistaken for a weakening.

## Security Concerns
None outstanding — this PR *is* the security fix. 🟢 One thing worth flagging as correctly handled rather than a concern: `cap_ceiling`'s `CEILING_RANK.get(requested)` relies on `PrivacyTierCeiling` being a `StrEnum` (hash/eq by value) so a raw uncoerced string still matches a member — this is non-obvious and is explicitly tested (`test_cap_ceiling_treats_a_raw_tier_string_as_its_member`) rather than just asserted in a comment.

## Problems
None found that block merge.

## Code Quality
- `dispatcher._handle_run_workflow` re-caps `report.privacy_tier_ceiling` even though the only production runner (`cli._build_workflow_runner`) already caps it. This is deliberate defense-in-depth and is documented as such (idempotent, makes `_capped` "the only gate" claim literally true) — reasonable given this is exactly the kind of invariant that erodes silently if a future runner is added without re-deriving it. Not a concern, just noting it's intentional duplication rather than an oversight.
- The module-level docstrings and inline comments are unusually dense (e.g. `_capped`, `cap_ceiling`). They earn their place — each one records a specific past mistake (the `.value` AttributeError, the `KeyError` on a bad string, the `%r` vs bare interpolation) rather than restating what the code does — but a future reader skimming `dispatcher.py` cold will need to budget real time for it. Nothing to change now.

## CLAUDE.md Adherence
✅ **TDD**: PR description documents RED-then-GREEN with specific failing assertions before any production line changed.
✅ **Gate 2 (`crawdad/scripts/check-all.sh`)**: reported 7/7 checks, 630 passed, 94.56% branch coverage (≥90% threshold), cold-cache ruff/mypy re-verification.
✅ **No shortcuts**: the only `# type: ignore[arg-type]` additions in the diff (`test_loop.py`, two new tests) match an existing, established convention already used ~15+ times elsewhere in the same file for the same Protocol-stub pattern — not a new suppression introduced by this PR.
✅ **Commit conventions**: `fix(crawdad): ...` conventional-commit messages, appropriately scoped.
✅ **Single source of truth**: `CEILING_RANK` and `COMPOSER_ADMITTED_CEILINGS` are consolidated with import-cycle justification and `is`-identity tests, exactly per the repo's DRY principle.
✅ **`crawdad/CLAUDE.md` scope note honored**: PR correctly ran only the crawdad-scoped gate since nothing under `creek-tools/` changed, and said so explicitly.

## Testing
Exceptionally thorough — this is the standout dimension of the PR. Coverage includes:
- Exact-value mapping pins (`test_cap_ceiling_maps_each_tier_to_an_exact_result`) alongside property tests (`test_cap_ceiling_never_widens`), so a property-satisfying-but-wrong stub (e.g. `return PERSONAL` always) is caught.
- The narrowest case (`open` stays `open`) is explicitly parametrized so a naive "always clamp to personal" implementation fails.
- Idempotence (`test_cap_ceiling_is_idempotent`), raw-string edge cases (well-spelled vs. unnameable), and log-content hygiene are all pinned with dedicated tests, not just implied by the implementation.
- Two end-to-end tests (`test_router_emitted_intimate_intent_never_reaches_call_tool`, `test_paradox_save_is_never_injected_above_the_cap`) exercise the actual attack chain through `AgentLoop.run`, not just the dispatcher in isolation — proving the poisoned value is reachable and gets stopped, not just that the unit-level function is correct.
- Both `ToolResult`-stamping branches that make no MCP call (`activate_register`, both `run_workflow` soft-errors) are tested to confirm they stamp `open` and can't be used to smuggle a ratchet upgrade to the auto-injected `creek.save`.

## Documentation
`crawdad/README.md` gets a clear, user-facing explanation of the clamp-vs-refuse distinction. Docstrings throughout `dispatcher.py`/`intents.py`/`workflows.py` were updated in lockstep with the code they describe (e.g. `ToolResult`'s docstring now accurately says "actually selected" rather than "mirrors the source intent's ceiling"), and stale claims (e.g. `loop.py`'s old "ordered by enum value strings" comment, which was actually wrong) were corrected rather than left to rot.



## Verdict: LGTM

The fix is correct, narrowly scoped, and exhaustively tested (property tests, exact-value pins, idempotence, log-hygiene, and end-to-end reachability tests proving the attack chain is actually closed); the security rationale for clamp-vs-refuse and for the dispatch-time (not validator-time) chokepoint is sound and verified against the referenced `creek_mcp` code. I found no correctness bugs, no CLAUDE.md violations, and no missing test coverage — only a couple of very minor, non-blocking stylistic observations.

Geoffe-Ga:
**CI**: 4/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234157352 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR closes a real privacy-tier leak in `crawdad`: the Haiku router's `privacy_tier_ceiling` was forwarded to `session.call_tool` unchecked, and because CrawDad talks MCP over stdio the server-side remote cap (`REMOTE_ADMITTED_CEILINGS`) never fires for it. The fix adds a single chokepoint (`dispatcher._capped`, backed by `intents.cap_ceiling`) that clamps every intent — router-emitted or workflow-derived — to `{open, personal}` before any MCP call, and makes `ToolResult.privacy_tier_ceiling` reflect the tier that actually selected the returned content (via the new `WorkflowRunReport`) rather than echoing untrusted router intent.

## Strengths
- Root-caused correctly: identifies the exact reason the server-side cap doesn't apply (`_caller_identity` sees a local stdio caller), rather than papering over symptoms.
- Clamp-not-refuse is well-justified: refusing would let a single poisoned vault fragment permanently silence the bot, since `AgentLoop.run` only catches two specific exception types. This is exactly the kind of DoS-vs-confidentiality tradeoff that's easy to get backwards.
- Defense in depth without redundant tables: `CEILING_RANK` and `COMPOSER_ADMITTED_CEILINGS` were consolidated into one home (`intents.py`) to avoid two "agreeing today, drifting tomorrow" copies — verified there's no import cycle (`dispatcher`/`workflows`/`loop` all import from the leaf `intents`, never each other in a cycle).
- Handles the "intent never reaches the field it's read from" hazard: uses `model_copy`/lookup rather than a pydantic validator, correctly noting `model_construct`/`model_copy(update=...)` bypass validators — and backs this with a test that drives a raw uncoerced string through `_capped` (`test_capped_clamps_an_uncoerced_raw_string_ceiling_without_raising`).
- The #1090 log-injection concern (interpolating vault-derived content into an operator-visible warning) is explicitly tested with a body sentinel that must never appear in the log line.
- Self-review section in the PR body documents a wrong claim the author's own security pass made (StrEnum hash-by-name vs. hash-by-value) and shows the test that caught it — good transparency, not just "reviewed, LGTM."
- Traced the flow end-to-end (`test_router_emitted_intimate_intent_never_reaches_call_tool`, `test_paradox_save_is_never_injected_above_the_cap`) rather than only unit-testing the clamp function in isolation.

## Security Concerns
None identified as blocking. 🟢 The core fix is sound: `cap_ceiling` uses a `<=` rank filter (not "return the highest admitted tier"), so it provably cannot widen even under a hypothetical non-contiguous admitted set, and `key=CEILING_RANK.__getitem__` avoids the `StrEnum` lexical-ordering trap. Manually traced `_capped` → `_build_arguments` → `session.call_tool` and confirmed the generic tool-call branch in `dispatch()` (crawdad/dispatcher.py:198) uses the already-capped `intent`, not `raw_intent`. All three ToolResult-producing branches (generic call, `_handle_activate_register`, `_handle_run_workflow`) are re-verified to stamp a value that can't exceed the cap.

## Problems
None found that block merge. I independently verified (not just trusted the PR body):
- `cap_ceiling`'s mapping (open→open, personal→personal, intimate/all→personal) against `COMPOSER_ADMITTED_CEILINGS = {OPEN, PERSONAL}` — correct.
- No import cycle between `intents` (leaf), `dispatcher`, `workflows`, `loop`, `cli`.
- All call sites of the now-`WorkflowRunReport`-returning `workflow_runner` callable (`bot.py`, `cli.py`, `slash_commands.py`, `dispatcher.py`, and all touched test files) were updated consistently — grepped for stragglers, found none.
- `slash_commands._reply_workflow_run` correctly unwraps `.reply` before posting to Discord (a missed unwrap would have leaked the enum tier into a `repr()` string in a user-visible message — the PR explicitly tests for this).

## Code Quality
- Non-blocking: two separate `if TYPE_CHECKING:` blocks now exist in `dispatcher.py` (one before the class definitions for `RegisterSwitcher`, one after `WorkflowRunReport` for `WorkflowDispatchRunner`), because the second alias needs a name defined later in the file. Slightly unusual to split, but the docstring explains why and it's harmless.
- The docstrings are unusually dense/discursive (multi-paragraph rationale in nearly every function). This is consistent with the project's evident house style and clearly serves the `interrogate` docstring-coverage gate, but a couple of the longer ones (e.g. `_capped`, `cap_ceiling`) border on essay-length for what the code itself makes fairly clear. Not asking for a change — flagging only because a future editor may be tempted to trim them and should preserve the safety rationale if so.

## CLAUDE.md Adherence
✅ Follows `crawdad/CLAUDE.md`'s TDD-first workflow (PR body documents Gate 1 RED failures before the fix). ✅ Uses the correct project gate (`crawdad/scripts/check-all.sh`) rather than the `creek-tools` one, correctly noting the diff is entirely inside `crawdad/`. ✅ No `# noqa` / `# type: ignore` / skipped tests introduced. ✅ Commit/PR title follows Conventional Commits (`fix(crawdad): ...`). ✅ One shared tier-ordering table maintained (no duplicate `CEILING_RANK`), matching the project's stated aversion to two-places-to-drift constants. CI was still in progress at review time (Tests & Type Checking, CrawDad Quality jobs running/queued; Code Complexity and Build Distribution already green, nothing failed) — merge should wait for those to finish per the project's "never merge with red checks" rule, but nothing in the diff suggests they will fail.

## Testing
✅ Strong, and appropriately paranoid for a security fix. Notable patterns: the "narrowest case" is pinned explicitly (`test_dispatch_never_widens_any_requested_ceiling` parametrized over all four tiers, including `open`, so a naive "always clamp to personal" implementation would fail); an existing assertion that encoded the leak (`intimate` on the save call) was tightened to `personal` while keeping an explicit `!= "open"` guard so the change reads as a tightening, not a silent weakening; idempotency of `cap_ceiling` is tested (relevant since multiple layers re-cap defensively); and the log-content test guards against the #1090 vault-content-in-logs hazard. Coverage claims (94.56% branch, 7/7 checks) are asserted in the PR body but not independently re-run here since the sandbox blocked shell execution of the test suite — this is a minor verification gap in the review, not a defect in the PR.

## Documentation
✅ Excellent. `crawdad/README.md` gains a clear operator-facing explanation of the new cap and why it differs in behavior (clamp vs. refuse) from the existing workflow-file cap. Module docstrings in `intents.py`, `dispatcher.py`, `workflows.py`, and `loop.py` were all updated to reflect the table's new home and the new invariants, not just the changed code paths.



## Verdict: LGTM

The fix is logically correct (independently traced the cap through every ToolResult-producing branch and confirmed it can't be bypassed), avoids the DoS trap of refusing instead of clamping, closes a real reachable leak with an end-to-end regression test, and leaves no missed call sites for the changed WorkflowRunner return type. The only remarks are minor, non-blocking style notes (split TYPE_CHECKING blocks, verbose docstrings) that don't warrant holding the PR.

Geoffe-Ga:
**CI**: 6/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234208574 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1267, opened 2026-08-09
MCP remote auth: replace the static bearer-in-env credential with a short-lived derived one (deferred from #895 / ADR-0009)


Geoffe-Ga:
## Context

Deferred from #895, and recorded as a rejected-for-now option in
[ADR-0009](../blob/main/creek-tools/docs/architecture/ADR/0009-mcp-consumer-token-rotation.md)
("Rejected → Short-lived derived credentials").

#895 gave `CREEK_MCP_CONSUMER_TOKENS` an overlapping-token rotation window
(`adepthood=<old>,<new>`) plus a published runbook, so a consumer's secret can
be rotated with no downtime. That fixes the *operational* hazard, but it does
not change the shape of the credential: what Adepthood puts on the wire is
still a **static, long-lived secret copied into an environment variable**,
re-verified fresh on every request. Revocation still means editing the env var
and restarting.

## Goal

Replace the bearer-in-env credential with a short-lived derived one, so a
leaked credential expires on its own rather than lasting until an operator
notices. Sketch (the design is part of the work, not settled here):

- Mint a time-boxed, HMAC-signed credential from the configured secret, and
  have the client exchange the long-lived secret for it (a token endpoint or
  refresh flow), or
- move off bearer-in-env entirely (mTLS, or a real authorization server).

## Why this is not a quick fix

It changes **what the Adepthood client sends** — today it presents the
configured secret verbatim as `Authorization: Bearer <secret>`. Any derived
scheme requires a coordinated change across both repositories, which is why
#895 deliberately did not fold it in. Sequence the client change with the
server change, or ship the server accepting both shapes during a migration
window.

## Acceptance sketch

- A wire credential that expires without operator action.
- The migration path for the existing Adepthood client is explicit and tested.
- ADR-0009 is superseded or amended, not silently contradicted.
- The #838 entropy floor and the constant-time, non-short-circuiting
  comparison survive the change.

## Related

- #895 (overlapping-token rotation window + runbook — the merged fix)
- #837 (transport confidentiality + AccessToken TTL)
- #838 / #907 (shared 32-char entropy floor)

Geoffe-Ga:
## Partial fix in #1757 — this issue stays open, and here is what it is blocked on

PR #1757 lands a slice of this and deliberately carries no closing keyword.

### The issue's premise is false in the direction that matters

This issue assumes token expiry **is** enforced and that only the credential's
[...]

PR #1271, opened 2026-08-09
feat(mcp): rotate consumer tokens through an overlapping window, not a cutover

### LINKED ISSUE #895: MCP remote auth: static consumer tokens are non-rotating shared secrets — TTL bounds only individual AccessTokens
## Context

Follow-up filed from #837 (per its "file follow-ups for adjacent MCP-auth items" constraint).

#837 gave remote `AccessToken`s a finite `expires_at` (default 3600 s, `CREEK_MCP_TOKEN_TTL_SECONDS`) and the SDK's `BearerAuthBackend` enforces it. However, the underlying credential in `CREEK_MCP_CONSUMER_TOKENS` is a static shared secret that is re-verified fresh on every HTTP request: a consumer (or an attacker holding the secret) keeps getting re-issued fresh AccessTokens indefinitely. The TTL bounds a captured in-memory `AccessToken` object, not the wire credential. Only rotating the configured secret revokes access. This is documented honestly in `docs/mcp.md` and the `verify_token` docstring, but there is no rotation mechanism or guidance beyond "change the env var and restart".

## Goal

Decide and implement the next step for credential lifecycle, e.g. one of:
- documented rotation runbook + support for accepting two tokens per consumer during a rotation window (old+new), or
- short-lived derived credentials (HMAC-signed, timestamped) minted from the configured secret, or
- explicitly accept the status quo in an ADR (TLS from #837 + entropy floor from #838 may be sufficient for the threat model).

## Related

- #837 (transport confidentiality + AccessToken TTL — merged fix)
- #838 (minimum token entropy — open sibling)

Geoffe-Ga:
## Summary

- **A consumer may now hold an ordered set of currently-valid tokens**, comma-separated inside its own `CREEK_MCP_CONSUMER_TOKENS` entry (`adepthood=<old>,<new>`). Every token in the set authenticates as that one consumer, so a secret rotates by widening the set → letting the consumer redeploy → narrowing it again. No cutover moment, and no wire-format change: a single-token entry is the steady state and parses exactly as it did before.
- **Two silently-wrong configurations are now refused at load time.** A consumer named twice (`adepthood=a;adepthood=b`) used to have the second entry silently overwrite the first, discarding a credential the operator believed was live — and now that a consumer can legitimately hold several tokens, the other reading (accumulate) would invent a rotation window nobody asked for. One token value configured more than once (across consumers, or twice within one set) is refused because `verify_token` deliberately never breaks early and keeps the *last* match, so a shared value would audit the call under the wrong consumer. The #838 32-char floor now applies to **every** token in a set, not just the first — a rotation is exactly when a fresh secret gets typed in.
- **Startup announces an open rotation window on stderr** (names and token counts only, never a value), so a window is visible rather than permanent. A published, numbered rotation runbook in `docs/mcp.md` is part of the fix, not documentation of it; `ADR-0009` records why overlapping static tokens were chosen over short-lived derived credentials.

### One type-level hazard worth calling out

`ConsumerTokenVerifier` takes `Mapping[str, Sequence[str]]` and **not** `Mapping[str, str | Sequence[str]]`. That union is unsound rather than merely loose: `str` already *is* a `Sequence[str]`, so the union collapses and `{"adepthood": "<secret>"}` would type-check while being read as one single-character token per letter — each of which `hmac.compare_digest` accepts, degrading one 43-character secret into an alphabet of valid credentials. A runtime `isinstance` guard closes the same hole for callers mypy never sees, and a test sweeps every distinct character of a configured token asserting none authenticates.

### What this deliberately does *not* do

The credentials are still static, long-lived secrets in an environment variable, and revocation still means editing the variable and restarting. The rotation window is operator-bounded: an operator who never runs the closing restart has permanently widened the credential set, and the startup notice is the only pressure against that. All of this is stated plainly in ADR-0009's Consequences rather than papered over. Short-lived derived credentials are the more thorough fix and are **deferred, not dismissed** — they change what the Adepthood client puts on the wire, which is a coordinated cross-repo contract change that must not ride inside a P1 fix. Filed as #1267.

`CONTRACT_VERSION` and every wire shape are untouched.

## Test plan

TDD throughout — the failing tests were written first and **verified failing before the fix existed**: 55 failed / 174 passed, each for the right reason (`AttributeError: 'tuple' object has no attribute 'encode'` on the shape change, `DID NOT RAISE` on every new refusal, `no attribute 'rotation_notice'` on the notice). After the implementation: 251 passed on the same four files.

New coverage, by behavior:

- **Parsing** — ordered multi-token sets; whitespace around commas stripped; empty comma segments dropped *before* the floor runs (so `missing=,,` is skipped, not reported as a 0-char token); the floor applied per token; repeated consumer name refused; a duplicate value refused both across consumers and within one set.
- **Verifier, under the narrowest ceiling** — both tokens of a window resolve to the *same* `client_id` with an exact `expires_at` (never `None`); **retiring a token revokes it immediately**, which is what makes the window bounded rather than a permanent widening; a bare `str` value raises `TypeError` *and* no single character of it authenticates; uniqueness enforced by the constructor independently of the parser (`tests/v1_api_support.py` builds verifiers from literal maps, bypassing `load_consumer_tokens`); the no-early-break property pinned by counting `hmac.compare_digest` calls — 3 tokens across 2 consumers with a match on the *first* one, where an inner break would give 2 and an outer break 1.
- **Notice** — `None` in the steady state; names each consumer mid-rotation with its count; asserted to contain none of the configured token substrings; lands on stderr with stdout empty, and `--print-openapi` stdout stays pure parseable JSON while a window is open.
- **Cross-cutting** — one parametrized sweep over all seven raising paths asserting no message ever echoes a token value.

One further ordering invariant is pinned. `_token_set` materialises `tuple(configured)` *before* testing emptiness, because an empty iterable that is not a `Sequence` (a spent iterator) is **truthy** and would otherwise slip past the refusal and land as an empty token set — a named consumer that authenticates nobody, i.e. a silent auth outage instead of the loud error the docstring promises. Every pre-existing test passed a literal `()`, which is falsy under either ordering, so nothing distinguished the two. I proved the new pin by reverting the ordering in place: `test_verifier_refuses_an_empty_iterable_that_is_not_a_sequence` went `DID NOT RAISE ValueError` with the rest of the file still passing, then went green again on restore.

Existing tests were reshaped in **call shape only** (`{"c": tok}` → `{"c": (tok,)}`); every assertion body is unchanged. I diffed the result against `git show HEAD:` to confirm no assertion was weakened or deleted.

Gate 2, run in a CI-equivalent environment with all provider credentials and consent vars stripped:

```
cd creek-tools && ./scripts/check-all.sh
→ Passed: 12  Failed: 0
→ 8655 passed, 5 skipped; aggregate coverage 94.36%; creek_mcp/remote_auth.py 98.67%
→ per-file gate, mypy strict, ruff, xenon, bandit, pip-audit, interrogate (99%), pylint (9.65) all clean
```

Gate 2.5: `code-review-orchestrator` over the diff, plus an explicit `security-specialist` pass (mandatory for this surface). The security pass caught and fixed a real one the tests did not cover: both adapters read the same `CREEK_MCP_CONSUMER_TOKENS` and this repo expects them side by side on one host, so the runbook's restart steps now say restart **every** process reading the variable — a partial restart at the closing step leaves the retired secret authenticating on the un-restarted process while the runbook tells the operator it is revoked. It also moved `announce_rotation_window` below the TLS posture gate in `httpapi/cli.py`, which was announcing a rotation window on a startup that then exited 2 (and falsifying its own module docstring).

## Notes for the reviewer

- Two off-contract edges in `_token_set` were considered and consciously left: a non-`str` *element* (`bytes`, an int) passes the constructor and fails at `verify_token` as a per-request 500 rather than a load-time refusal. It is fail-closed, unreachable from any in-repo caller, and mypy strict makes it unrepresentable for every real one; an element-type check on the auth path seemed the worse trade. Say the word if you disagree.
- The blank-`client_id` hazard tracked in #1100 is unchanged here, but this PR creates the natural place to close it — `_normalized_token_sets` is now the single constructor-level chokepoint every verifier passes through. Noted on #1100 rather than folded in.

Closes #895

### DIFF
--- a/creek-tools/creek_mcp/httpapi/auth.py
+++ b/creek-tools/creek_mcp/httpapi/auth.py
@@ -65,6 +65,14 @@ def build_verifier(
 ) -> ConsumerTokenVerifier:
     """Return the verifier for the configured consumers, or refuse to build one.
 
+    A consumer may hold an ordered set of currently-valid tokens (#895), so a
+    ``/v1`` deployment rotates a secret through the same environment variable
+    and the same parser the MCP surface uses — there is nothing adapter-local
+    to keep in step. The map's *values* are token sequences rather than single
+    tokens, which changes nothing here: the truthiness check below asks
+    whether any consumer is configured at all, and an empty map is still the
+    only way to answer "no".
+
     Args:
         environ: Environment mapping to read, or ``None`` for
             :data:`os.environ`. Tests always pass one explicitly, so no test
@@ -77,11 +85,13 @@ def build_verifier(
     Raises:
         ValueError: When no consumer tokens are configured, naming the setting
             the operator must set; or, propagated unchanged from
-            :func:`creek_mcp.remote_auth.load_consumer_tokens`, when a
-            configured token is below the shared length floor. That second
-            message names the consumer, the observed and required lengths and
-            the rotation recipe — and never the token value, because a startup
-            error lands in logs, terminals and process supervisors.
+            :func:`creek_mcp.remote_auth.load_consumer_tokens`, when the
+            configuration cannot be read unambiguously — a token below the
+            shared length floor (#838), a consumer named twice, or one token
+            value configured for two consumers (#895). Those messages name the
+            consumers, the observed and required lengths and the rotation
+            recipe — and never a token value, because a startup error lands in
+            logs, terminals and process supervisors.
     """
     tokens = load_consumer_tokens(environ)
     if not tokens:
--- a/creek-tools/creek_mcp/httpapi/cli.py
+++ b/creek-tools/creek_mcp/httpapi/cli.py
@@ -20,6 +20,14 @@ Startup is fail-closed three times over, all before a socket exists:
 * **No cleartext bearer on a routable network.** A non-loopback bind without
   TLS exits naming the flags that fix it.
 
+Having cleared those, startup **announces on stderr** any consumer holding
+more than one currently-valid token, so an open rotation window (#895) is
+visible rather than permanent. It goes to stderr because stdout here carries
+the ``--print-openapi`` document consumers pipe into client generators; the
+decision, and the notice itself, are
+:func:`creek_mcp.remote_auth.announce_rotation_window`'s, shared with the MCP
+adapter so it cannot hold on one entry point and lapse on the other.
+
 **The default port is deliberately not 8000.** That is both Adepthood's own
 backend port and :data:`creek_mcp.server.DEFAULT_MCP_NETWORK_PORT`, and the two
 Creek adapters are expected to run side by side on one host; identical defaults
@@ -46,6 +54,7 @@ from creek_mcp.api.openapi import build_openapi
 from creek_mcp.httpapi import SERVER_NAME
 from creek_mcp.httpapi.app import create_app
 from creek_mcp.httpapi.auth import build_verifier
+from creek_mcp.remote_auth import announce_rotation_window
 from creek_mcp.transport_posture import require_transport_confidentiality
 
 if TYPE_CHECKING:
@@ -162,8 +171,10 @@ def _verifier_or_exit(parser: argparse.ArgumentParser) -> ConsumerTokenVerifier:
     try:
         verifier = build_verifier()
     except ValueError as exc:
-        # Covers both "nothing configured" and "a configured token is below the
-        # shared length floor". Neither message ever carries a token value, and
+        # Covers "nothing configured", "a configured token is below the shared
+        # length floor" (#838), and the two ambiguous registries #895 refuses:
+        # a consumer named twice, and one token value configured for two
+        # consumers. No such message ever carries a token value, and
         # ``parser.error`` exits non-zero rather than returning.
         parser.error(str(exc))
     return verifier
@@ -231,7 +242,11 @@ def serve(app: Starlette, args: argparse.Namespace) -> None:  # pragma: no cover
 
 
 def main(argv: list[str] | None = None) -> None:
-    """Parse arguments, refuse an unsafe configuration, then serve.
+    """Parse arguments, refuse an unsafe configuration, announce, then serve.
+
+    The announcement is the one non-refusal in the sequence: any open rotation
+    window (#895) is reported on stderr before the socket exists, so it is
+    visible in the same startup output as the refusals above it.
 
     Args:
         argv: Command-line arguments, or ``None`` to read :data:`sys.argv`.
@@ -245,4 +260,12 @@ def main(argv: list[str] | None = None) -> None:
     _apply_config(parser, args)
     verifier = _verifier_or_exit(parser)
     require_transport_confidentiality(parser, args)
+    # After every refusal, never before one: a process about to exit 2 has
+    # nothing to announce, and announcing above the posture gate printed a
+    # rotation notice for a server that then refused to serve — while the MCP
+    # adapter announced below its own gate. Two adapters, one moment. A window
+    # has to be closed again, and an operator who cannot see one is open will
+    # not close it (#895). On stderr: stdout on this entry point carries the
+    # ``--print-openapi`` document, which consumers pipe into code generators.
+    announce_rotation_window(verifier)
     serve(create_app(verifier=verifier), args)
--- a/creek-tools/creek_mcp/remote_auth.py
+++ b/creek-tools/creek_mcp/remote_auth.py
@@ -3,8 +3,8 @@
 When ``creek-tools-mcp`` is served over the network (streamable-http), every
 request must present a bearer token that maps to a known **consumer identity** —
 there is no anonymous access. Tokens are configured in the environment only
-(never in code or the config file), one per remote consumer, mirroring the
-existing ``CREEK_MCP_ELEVATED_TOKEN`` fail-closed / constant-time precedent.
+(never in code or the config file), mirroring the existing
+``CREEK_MCP_ELEVATED_TOKEN`` fail-closed / constant-time precedent.
 
 The SDK's :class:`~mcp.server.auth.provider.TokenVerifier` hook does the rest:
 when a :class:`ConsumerTokenVerifier` is wired into ``FastMCP``, the SDK installs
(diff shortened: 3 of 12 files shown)


Geoffe-Ga:
## Summary
This PR replaces `CREEK_MCP_CONSUMER_TOKENS`'s single-token-per-consumer model with an ordered, comma-separated set of currently-valid tokens per consumer, enabling zero-downtime secret rotation instead of a hard cutover. It refuses two previously-silent misconfigurations (a repeated consumer name; a token value shared across/within consumers), closes a real `str`-is-a-`Sequence[str]` type hazard with a runtime guard, adds a stderr-only rotation-window startup notice shared by both the MCP and `/v1` adapters, and adds ADR-0009 plus a numbered rotation runbook in `docs/mcp.md`.

## Strengths
- **TDD discipline is visible and verifiable**: the PR body documents the red→green transition (55 failed/174 passed before the fix, 251 passed after) for the exact right reasons, and the diff shows tests added before/alongside implementation.
- **The revocation test is load-bearing**: `test_retiring_a_token_revokes_it_immediately` and `test_closing_the_window_revokes_the_retired_token` prove the window actually closes — without this the feature would just be a permanent widening of the credential surface.
- **Correctly closes the `str`/`Sequence[str]` unsoundness** both at the type level (`Mapping[str, Sequence[str]]`, not the unsound `str | Sequence[str]` union) and at runtime (`isinstance` guard in `_token_set`), with a test sweeping every character of a real token to prove none of them authenticate alone.
- **The ordering fix in `_token_set`** (materialize `tuple(configured)` before testing truthiness, so a spent iterator — truthy but empty — can't slip past the empty-token-set refusal) is a genuinely subtle correctness point, and the author proved the pin by reverting it and confirming the test fails.
- **No-token-value-leaked property is enforced by a single parametrized sweep** (`test_no_rejection_message_ever_echoes_a_token`) over every new raising path, rather than trusted to per-test diligence.
- **Announcement placement is correct and deliberately tested**: `announce_rotation_window` runs after every fail-closed gate (transport confidentiality, "no anonymous access") on both `creek-tools-mcp` and `creek-tools-api`, and always to stderr, verified against `--print-openapi`'s stdout staying pure JSON during an open window.
- **No wire-format or `CONTRACT_VERSION` change** — single-token entries parse identically to before, confirmed by tests.
- ADR-0009 is substantive: it documents rejected alternatives (short-lived derived credentials, per-token expiry timestamps) with real reasoning, and states the negative consequences (operator-bounded window, still-static secrets) plainly rather than glossing over them.

## Security Concerns
🟢 LOW — None found blocking. The two off-contract edges called out by the author (a non-`str` element in a token sequence bypassing the load-time refusal and failing per-request instead; the pre-existing blank-`client_id` hazard in #1100) are reasonable, explicitly-scoped deferrals rather than oversights, and are unreachable from any in-repo caller under mypy strict.

## Problems
None blocking. I verified the core invariants by reading the implementation against its tests:
- `_normalized_token_sets` → `_token_set` → `_require_globally_unique` correctly catches both same-consumer and cross-consumer duplicate token values (single dict-based owner-tracking pass), and the ordering (materialize-before-emptiness-check) is correct.
- `verify_token`'s nested loop has no `break` at either level, preserving the pre-existing (and now nested) constant-scan timing property; this matches the counted-comparisons test.
- Startup ordering in both `creek_mcp/server.py` and `creek_mcp/httpapi/cli.py` places `announce_rotation_window` after all fail-closed refusals, matching the PR's own description of a bug the author's security-specialist pass caught and fixed pre-review.

## Code Quality
- `_normalized_token_sets` is invoked twice in the `load_consumer_tokens` → `ConsumerTokenVerifier.__init__` path (once implicitly via `_validated_tokens`/`_token_set` during parsing, once again in the constructor). This is intentional per the docstring (the constructor is "the narrowest ceiling," shared with literal-map callers like `tests/v1_api_support.py`), so it's a correctness feature, not redundant dead work — just flagging for awareness, not a change request.
- `_refuse_shared_token(first, second)` can be called with `first == second` (duplicate value within one consumer's own set), producing a slightly odd "once for 'adepthood' and once for 'adepthood'" message. Purely cosmetic; no test depends on the two names differing.

## CLAUDE.md Adherence
✅ Coverage: author reports 94.36% aggregate / 98.67% on `remote_auth.py`, comfortably above the ≥90% floor.
✅ Docstring coverage: every new/changed public and private function carries Args/Returns/Raises docstrings; author reports 99% via interrogate.
✅ Typing: `from __future__ import annotations` present, `Mapping`/`Sequence` correctly deferred to `TYPE_CHECKING` (safe since annotations aren't evaluated eagerly), no `Any` introduced; author reports mypy strict clean.
✅ Commits: single commit follows Conventional Commits (`feat(mcp): ...`), body explains why over what, references issues (#895, #837, #838, #1267).
✅ Complexity: new functions are small and single-purpose (`_token_set`, `_parsed_entry`, `_refuse_*`, `_require_globally_unique`); nothing here looks likely to trip the ≤10 cyclomatic-complexity gate.
✅ No CONTRACT_VERSION/wire-shape change, correctly called out.

## Testing
Extensive and well-targeted: parsing edge cases (whitespace, empty comma segments, per-token floor), verifier edge cases (bare-`str` guard, empty-set guard, the spent-iterator ordering pin, global uniqueness at the constructor level independent of the parser), the no-early-break timing property (via a `compare_digest` call-counting spy), the rotation notice's content and silence in the steady state, and the stderr/stdout separation on both adapters including under `--print-openapi`. The revocation tests make this a genuine rotation rather than a permanent widening. I did not execute the suite myself (sandbox blocked `pytest` invocation), so this assessment is based on static reading of test bodies against implementation, not a live run — the self-reported CI-equivalent run (8655 passed) in the PR description is trusted but unverified by me.

## Documentation
Thorough: `docs/api.md` and `docs/mcp.md` updated consistently with each other and with the code (revocation vs. rotation distinguished, numbered runbook with an explicit warning about restarting *every* process sharing the env var — itself a fix the author's own security pass caught), and ADR-0009 is a genuine decision record with rejected alternatives and stated negative consequences rather than a rubber stamp.



## Verdict: LGTM

The change is a well-scoped, TDD-driven auth fix with a load-bearing revocation test proving the rotation window is actually bounded, careful handling of the str-is-a-Sequence type hazard (both mypy-level and runtime-guarded), no token values ever echoed (swept by a parametrized cross-cutting test), correct ordering of the new stderr announcement relative to the existing fail-closed startup gates, and no wire/CONTRACT_VERSION changes. I found no bug, security gap, or missing-test scenario that should block merge; the only items worth a second look (the `_token_set`/`_normalized_token_sets` double-validation redundancy, the two off-contract edges the author flagged) are already called out by the author as conscious, reasoned trade-offs.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234199912 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

PR #1276, opened 2026-08-09
fix(mcp): bound the elevated-purge brute force with a non-blocking attempt budget

### LINKED ISSUE #914: [scan:security] Purge gate has no failed-attempt backoff — elevated token is brute-forceable at machine speed
## Role

You are hardening the MCP purge gate — the authorization check that stands between a caller and irreversible vault destruction.

## Goal

Add failed-attempt backoff (or rate limiting) to the elevated-token purge gate in `creek-tools/creek_mcp/`, so that an attacker who can reach the MCP transport cannot brute-force `CREEK_MCP_ELEVATED_TOKEN` by making unbounded guesses at machine speed.

## Context

PR #913 (issue #907) gave `CREEK_MCP_ELEVATED_TOKEN` a 32-char entropy floor, matching the per-consumer token floor added in #838 / PR #897. That raises the cost of a successful guess enormously, but it caps only the *keyspace* — it does not bound the *attempt rate*. The chief-architect on #907 flagged rate limiting as a real residual risk and explicitly scoped it out of that issue; this issue is that deferred follow-on.

Relevant surfaces after #913:

- `creek-tools/creek_mcp/token_policy.py` — the shared floor/constant/rejection-wording module (new in #913)
- `creek-tools/creek_mcp/auth.py` — `is_elevated()`, returns `bool`, never raises (the `purge._gate` audit invariant depends on this)
- `creek-tools/creek_mcp/tools/purge.py` — `_gate`, deliberately returns a *generic* refusal so a hostile caller gets no configuration oracle
- `creek-tools/creek_mcp/remote_auth.py` — the remote consumer-token path

Two constraints inherited from #913 that must not regress:

1. **No config oracle.** The caller-visible refusal must stay generic. A backoff message must not leak whether the token was wrong, the wrong length, absent, or merely throttled — differential responses are themselves an oracle.
2. **`is_elevated` keeps `-> bool` and never raises**, so `purge._gate` continues to record its audit entry on every attempt.

## Approach

Decide and justify in the PR: in-process counter vs. persisted state. In-process is simpler and stdlib-only but resets on restart, which an attacker can force if they can crash or cycle the server; persisted state survives restarts but adds a write path and a file-permission surface. Consider whether the backoff should key on the transport peer, on the process as a whole, or both. Prefer stdlib — `creek_mcp` is currently dependency-light, and `token_policy.py` was deliberately kept stdlib-only.

## Format

One PR closing this issue, following the four gates in `.claude/agents/shared/house-rules.md`.

## Constraints

- TDD: failing tests first, covering at minimum the delay-after-N-failures behavior, the reset-on-success behavior, and that the refusal wording is byte-identical between a throttled and an unthrottled rejection.
- Meet the repo thresholds: >=90% branch coverage, per-file >=80%, >=95% docstring, complexity <=10, mypy strict, ruff clean.
- Do not weaken or bypass any gate. No `# noqa` / `# type: ignore` / `@pytest.mark.skip` without an `Issue #N` justification.
- Do not change the entropy floor or the rejection wording introduced in #913 — this is additive.
- Tests must not sleep for real wall-clock backoff durations; inject the clock.

Refs #907

Geoffe-Ga:
## Summary

- **Bounds the elevated-token brute force.** `is_elevated` compared the caller's token in constant time but compared it *as often as it was asked to*. `hmac.compare_digest` stops an attacker learning the secret byte-by-byte; nothing stopped them guessing it whole, at machine speed, against a surface that answers in microseconds. A new stdlib-only `creek_mcp/attempt_policy.py` meters every verification: after `MAX_FAILED_ATTEMPTS = 5` consecutive denials the gate stops evaluating tokens at all for `LOCKOUT_SECONDS = 60.0`.
- **`creek_mcp/tools/purge.py` is a deliberate zero-byte diff.** `is_elevated` keeps its exact signature and delegates to the budget; the old comparison moved verbatim into a private `_verify_match`. Because the gate's caller is untouched, a throttled refusal is byte-identical to any other — same `status`, `tool`, `reason`, no `retry_after`/`throttled`/`attempts_remaining`. No `CONTRACT_VERSION` bump: nothing moved on the wire.
- **Fixes a latent raise found during review.** A lone surrogate (which survives `json.loads`, so it arrives straight off the wire) made `.encode("utf-8")` throw `UnicodeEncodeError`. `purge._gate` does not catch, so the exception escaped into the tool surface and **skipped the mandatory audit entry**. Worse, once the budget existed the raise became *state-dependent* — inside a lockout `verify` is never called, so the same input returned a clean `False` — an exact, free oracle for whether the budget was armed, costing the attacker no budget. Now fails closed.

## Design decisions the issue asked for

**In-process vs. persisted: in-process, stdlib-only**, matching the #913 precedent that kept `token_policy.py` dependency-free. Persisted state was rejected because it adds a write path and a file-permission surface to an auth module. Stated plainly: this **bounds** the brute force, it does not close it. An attacker who can cycle the process resets the counter — which already requires powers strictly greater than guessing the token.

**Keying: process-global.** The *secret* is process-global, so one guess consumes one guess from the single shared keyspace regardless of who made it; per-identity keying would multiply an attacker's budget by the number of identities they hold, and a per-tool budget would make the five purge tools a 25-guess gate. (Note for the record: `consumer` is **not** caller-declarable — it is `_effective_consumer(...)`, the verified remote bearer's `client_id` when remote, else the build-time env default. `_purge_fragment` has no `consumer` parameter.)

**No `sleep`, anywhere in production.** The five purge tools are registered *sync* (`server.py:730-731`) and the MCP SDK invokes sync tool functions inline on the asyncio event loop (`mcp/server/fastmcp/utilities/func_metadata.py:94-96`). A blocking backoff would freeze the entire server — every tool, every consumer, both transports — turning a rate-limit fix into a remotely-triggerable availability kill. The lockout returns immediately; `test_a_throttled_refusal_never_sleeps` pins this structurally, including the aliased-import escape route.

**Threat model, not overstated.** Reaching a purge call already requires either the local stdio client or a valid ≥32-char consumer bearer (`main` refuses the network transport without `CREEK_MCP_CONSUMER_TOKENS`). The realistic brute-forcer is a *partially trusted* party — canonically CrawDad, which is deliberately given transport access and deliberately denied the elevated token. This is a real residual, not an open-internet attack.

**Availability cost, measured and accepted.** Under a sustained 1 Hz attack the attacker gets **285 evaluated guesses per hour (4.75/min**, previously bounded only by the network**)**, but the gate is shut to *everyone* for **~92%** of that hour: each expiring window is re-spent in five seconds and re-arms. So MCP purge is not merely delayed 60 seconds while an attack runs — it is largely unavailable. Accepted deliberately; the operator's recovery path is `creek purge` on the CLI, which drives `PurgeEngine` directly and never reaches this gate. All three numbers were measured against the built code, and `test_a_sustained_attacker_is_held_to_the_documented_guess_rate` pins them.

**Audit behaviour is unchanged and tested:** every attempt, throttled or not, still writes exactly one `mcp.jsonl` entry. The throttle is what bounds log growth. The throttle is deliberately invisible in the audit log, so operators detect it by the *rate* of refusals, not by a flag.

**Timing:** the throttled path returns before `compare_digest` and is observably faster. It leaks only the global lockout state, never token material, and is buried under the per-call `mcp.jsonl` append. It does not widen the pre-existing "misconfigured vs. armed" micro-oracle in `_verify_match`, which is out of scope here.

## Test plan

- **TDD, RED proved first.** Before any production edit: 18 failures, 8 behavioral. The lead test failed on its assertion, not on a missing symbol — `assert {'affected_fr...'} == {'reason': ...}`, `{'status': 'ok'} != {'status': 'refused'}`, and the fragment had been deleted. `test_every_purge_tool_inherits_the_shared_lockout` failed `'ok' == 'refused'` on all five tools.
- **Mutation-tested the re-arm**: disabling it makes the sustained-attack test go red (3541 evaluated guesses vs. 285), so that test has a real trip-wire rather than asserting a shape.
- **Adversarial hand-check** of the state machine: 5 wrong → 600 s of 1 Hz hammering inside the window → `verify` never invoked; window boundary exact at 59.999 s (refused) / 60.0 s (served); full allowance restored after heal, not one probationary attempt.
- **The surrogate regression case was RED first** (`UnicodeEncodeError`), then green.
- **Gate 2**: `./scripts/check-all.sh` → **12/12, exit 0**. 8664 passed, 94.35% branch coverage; `attempt_policy.py` 95.74%, `auth.py` 100%; per-file gate passes with **no new waiver**. Ruff lint + format re-verified with `--no-cache`.
- **Order-independence**: the MCP suites pass under several module permutations (553 and 570 passed). An autouse `_reset_elevated_attempt_budget` fixture clears the process-global budget between tests — without it this change would be green locally and flaky in CI.
- **No bypasses**: zero `# noqa` / `# type: ignore` / `# pylint: disable` / `skip` in the diff, no threshold edits, no coverage waiver, no test deleted. The one pre-existing AST guard that changed was *widened* from `is_elevated` to every function in `auth.py` — strictly stronger, since moving the comparison into a helper would otherwise have made it vacuous.
- **Scope**: `remote_auth.py` untouched (live sibling lane), `token_policy.py` untouched (`MIN_TOKEN_LEN` still 32), `contract.py` / `api/models.py` untouched, `crawdad/` untouched, no dependency or lockfile change.

## Note for the reviewer

This branch is based on `bc83b7b` and `main` has moved since; `remote_auth.py` in particular has been rewritten on `main` by a sibling lane. No file in this diff overlaps it, but the base is behind — worth a sync before merge.

Refs #907
Closes #914

### DIFF
--- a/creek-tools/creek_mcp/attempt_policy.py
+++ b/creek-tools/creek_mcp/attempt_policy.py
@@ -1,0 +1,168 @@
+"""Failed-attempt budget for the elevated-authorization gate (#914).
+
+The defect this closes: ``creek_mcp.auth.is_elevated`` compared the
+caller's token against ``CREEK_MCP_ELEVATED_TOKEN`` in constant time, but
+it compared it *as often as it was asked to*. Constant-time comparison
+stops an attacker learning the secret one byte at a time; it does nothing
+about guessing it whole, at machine speed, against a surface that answers
+in microseconds. A partially-trusted caller who could reach the MCP
+boundary could therefore brute-force the token that authorizes
+irreversible ``creek.purge.*`` vault destruction.
+
+:class:`AttemptBudget` puts a price on failure: after
+:data:`MAX_FAILED_ATTEMPTS` consecutive denials the gate stops evaluating
+tokens at all for :data:`LOCKOUT_SECONDS`.
+
+**The lockout is non-blocking.** Sleeping is the obvious way to rate
+limit and it is the wrong one here. The five purge tools are registered
+as *sync* functions (``creek_mcp/server.py:731``), and the MCP SDK
+invokes a sync tool function inline on the asyncio event loop
+(``mcp/server/fastmcp/utilities/func_metadata.py:96``). A blocking
+backoff in this module would therefore freeze the entire server — every
+tool, every consumer, both transports — for the duration, converting a
+rate-limit fix into a remotely-triggerable availability kill. There is no
+``sleep`` anywhere in this module, and there must never be one: a
+throttled caller is refused *immediately*, and pays in refusals rather
+than in held connections.
+
+Deliberately stdlib-only. :mod:`creek_mcp.auth` imports this module at
+import time, so a dependency pointing back — even at a single constant —
+would be an import cycle that surfaces as an ``ImportError`` in whichever
+module happened to load first.
+"""
+
+from __future__ import annotations
+
+import threading
+import time
+from typing import TYPE_CHECKING, Final
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+MAX_FAILED_ATTEMPTS: Final[int] = 5
+"""Consecutive denials that arm the lockout.
+
+Five *evaluated* guesses per lockout window, against the keyspace of a
+``secrets.token_urlsafe(32)`` secret — the 32-character floor
+``creek_mcp.token_policy.MIN_TOKEN_LEN`` enforces on any configured
+elevated token (#907). Five guesses per minute against that keyspace is
+not a search anyone completes; the pairing of the two policies is what
+makes the bound worth anything, which is why the floor is named here even
+though this module must not import it.
+"""
+
+LOCKOUT_SECONDS: Final[float] = 60.0
+"""How long the gate stays shut once the budget is spent.
+
+The self-heal guarantee: a lockout is a delay, never a ban. Sixty seconds
+after the last allowed failure the gate reopens with its full allowance
+restored, with no operator intervention, no restart, and no way for a
+hostile caller to hold it shut (see
+:meth:`AttemptBudget.attempt` — attempts made inside the window do not
+extend it). Bricking the only path an operator has to erase their own
+data would be a worse outcome than the brute-force channel #914 closes.
+"""
+
+# Monkeypatchable clock alias, mirroring `creek_mcp.remote_auth._now` so both
+# time-dependent auth surfaces are pinned the same way in tests. (Named rather
+# than cited by line: that module is under active change on another branch.)
+# Monotonic rather than `time.time` on purpose, and unlike that seam: wall-clock
+# time can be stepped by NTP or by an operator changing the system clock, and a
+# lockout a clock change can shorten is a lockout an attacker can shorten.
+_now = time.monotonic
+
+
+class AttemptBudget:
+    """A consecutive-failure allowance guarding one verification surface.
+
+    Verification is supplied to :meth:`attempt` as a callback rather than
+    performed by the caller, so an unthrottled check is not expressible at
+    the type level: there is no way to consult the secret except through
+    the budget that meters the consultation.
+
+    Instances are safe to share across threads. Every read-modify-write of
+    the counter happens under one lock, so N concurrent callers cannot each
+    observe "one attempt left" and each spend it — which would turn an
+    N-guess gate into an N-guesses-per-thread gate.
+    """
+
+    def __init__(self, *, max_failures: int, lockout_seconds: float) -> None:
+        """Configure a budget.
+
+        The policy is taken from these arguments and never re-read from
+        :data:`MAX_FAILED_ATTEMPTS` / :data:`LOCKOUT_SECONDS`, so an
+        instance means exactly what its construction site says it means.
+
+        Args:
+            max_failures: Consecutive failed attempts that arm the lockout.
+            lockout_seconds: How long the lockout lasts, in seconds, as
+                measured by the monotonic :data:`_now` clock.
+        """
+        self._max_failures = max_failures
+        self._lockout_seconds = lockout_seconds
+        self._lock = threading.Lock()
+        self._failures = 0
+        self._locked_until: float | None = None
+
+    def attempt(self, verify: Callable[[], bool]) -> bool:
+        """Consult *verify* if the budget allows it, and account for the answer.
+
+        The only entry point. While a lockout is armed *verify is not
+        called at all* — that is what makes the bound a bound, rather than
+        a throttle that still runs the comparison and discards the result.
+
+        Attempts made inside the window do not extend it. Without that
+        rule one hostile guess per second would keep the window
+        permanently re-armed and lock the legitimate operator out of their
+        own purge tools forever, with every individual refusal looking
+        correct.
+
+        *verify* must not block and must not perform I/O. The purge tools
+        are sync MCP tool functions invoked inline on the asyncio event
+        loop, so anything slow here stalls the whole server, not just this
+        request. For the same reason this method never sleeps, and the
+        internal lock is never held across a sleep or any I/O — it covers
+        only the counter arithmetic and the *verify* call itself.
+
+        Args:
+            verify: Performs the actual check, returning whether the
+                caller's credential is correct. Called at most once.
+
+        Returns:
+            ``True`` only when the budget was open *and* *verify* said yes.
+            A refusal inside the window is indistinguishable from an
+            ordinary rejection, so the lockout never becomes an oracle.
+        """
+        with self._lock:
+            now = _now()
+            if self._locked_until is not None:
+                if now < self._locked_until:
+                    return False
+                self._locked_until = None
+                self._failures = 0
+
+            if verify():
+                self._failures = 0
+                return True
+
+            self._failures += 1
+            if self._failures >= self._max_failures:
+                self._locked_until = now + self._lockout_seconds
+            return False
+
+    def reset(self) -> None:
+        """Clear the failure count and any armed lockout.
+
+        The test-isolation hook: the budget guarding the elevated gate is
+        process-global, so failures accrued by one test would otherwise
+        leak into the next and make the suite order-dependent. The autouse
+        ``_reset_elevated_attempt_budget`` fixture in ``tests/conftest.py``
+        calls this before every test.
+
+        Not for production use. Clearing a live lockout on demand would
+        hand back precisely the unbounded guessing #914 took away.
+        """
+        with self._lock:
+            self._failures = 0
+            self._locked_until = None
--- a/creek-tools/creek_mcp/auth.py
+++ b/creek-tools/creek_mcp/auth.py
@@ -9,8 +9,10 @@ a timing-side-channel comparison would leak the token byte-by-byte to a
 hostile MCP client. ``hmac.compare_digest`` runs in constant time
 relative to the input length, so an attacker cannot probe the secret
 through repeated calls. The :mod:`creek_mcp.auth` test module asserts
-via an AST walk that ``is_elevated`` contains no ``==``/``!=`` on the
-hot path; the rule is mechanical, not stylistic.
+via an AST walk that **no** function in this module contains an
+``==``/``!=`` comparison; the rule is mechanical, not stylistic, and it
+covers the whole module rather than one function because #914 moved the
+comparison into a private helper.
 
 The configured token must also clear the shared 32-character floor in
 :data:`creek_mcp.token_policy.MIN_TOKEN_LEN` (#907) — a guessable secret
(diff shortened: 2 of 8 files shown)


Geoffe-Ga:
## Summary
This PR bounds brute-force guessing of `CREEK_MCP_ELEVATED_TOKEN` (the token that authorizes irreversible `creek.purge.*` calls) by adding a process-global, non-blocking failed-attempt lockout (`creek_mcp/attempt_policy.py`). It preserves `is_elevated`'s exact signature, keeps `creek_mcp/tools/purge.py` untouched, and additionally fixes a latent `UnicodeEncodeError` on lone-surrogate tokens that could have skipped the mandatory audit entry and leaked a lockout-state oracle. I verified the diff against the current working tree: `purge.py` is confirmed unmodified, the five purge tools are confirmed registered as sync functions, and the lockout state machine's arm/disarm/no-extension/reset logic all check out by hand-tracing `AttemptBudget.attempt`.

## Strengths
- **Correct, minimal state machine.** `AttemptBudget.attempt` (`creek_mcp/attempt_policy.py:114-158`) is easy to verify by inspection: lock held for the whole read-modify-write, `verify()` never called while a window is armed, no-extension-of-window and full-reset-on-expiry both hold, and a successful call zeroes the counter. The behavior matches what the extensive test suite claims.
- **The regression fix is real and well-motivated.** Wrapping `expected.encode`/`provided_token.encode` in `try/except UnicodeEncodeError` (`auth.py:279-283`) closes a genuine bug: a lone surrogate in `auth_token` would previously raise past `purge._gate`, skip the audit write, and — once the lockout existed — become a free oracle for whether the budget was armed (raise vs. clean `False` depending on lockout state). Good catch, and it's covered by a dedicated parametrized regression test (`unencodable-client-token` case in `test_mcp_auth.py`).
- **Threat-model honesty.** The docs and docstrings state the residuals plainly: process-restart resets the counter, a sustained 1Hz attacker still gets ~285 guesses/hour while locking the tool out ~92% of that hour for everyone. These numbers are pinned by a real behavioral test (`test_a_sustained_attacker_is_held_to_the_documented_guess_rate`), not just asserted in prose — I hand-checked the arithmetic (cycle = 4 + 60 = 64s, 3600/64 → 56 whole cycles + 16s tail ≥ 5 → 285 evaluations, 4.75/min) and it holds.
- **No wire/oracle leakage.** The refusal payload is byte-identical whether throttled or not (`{"status", "tool", "reason"}` only), verified in `purge.py` unchanged and asserted by `test_a_throttled_refusal_is_byte_identical_to_an_ordinary_one` / `test_a_locked_gate_cannot_be_told_apart_from_a_wrong_token`.
- **Structural tests, not just behavioral ones.** AST-based checks assert no `sleep` anywhere in `attempt_policy.py` (including the aliased-import escape route), that the module stays stdlib-only, that exactly one function in `auth.py` reaches `hmac.compare_digest` and it isn't `is_elevated`, and that no function outside an allowlist reads `ELEVATED_TOKEN_ENV`. These catch the exact ways this kind of fix quietly regresses.
- **Concurrency correctness is tested, not assumed** (`test_concurrent_attempts_cannot_overspend_the_budget`) — 8 threads × 3 attempts each, asserting the shared lock prevents overspend.
- **Test isolation handled correctly.** The process-global budget is reset via an autouse `conftest.py` fixture that accesses `auth._ELEVATED_BUDGET` directly (not `getattr`-tolerant), with its own test (`test_the_auth_module_owns_a_budget_the_conftest_hook_can_reset`) to keep that honest.
- **Every purge tool inherits the shared budget**, proven by arming the lockout through one tool and checking refusal on a different one (`test_every_purge_tool_inherits_the_shared_lockout`) — the right way to prove a per-tool counter isn't hiding.

## Security Concerns
🟢 LOW — The lock in `AttemptBudget.attempt` is held across the `verify()` call, which serializes *all* `is_elevated` evaluations (including concurrent legitimate calls) process-wide on a single lock. This is deliberate (verify is required to be non-blocking/no I/O) and not a vulnerability, just a minor throughput note for a rarely-hit path — not worth blocking on.

No other concerns. The design decisions (in-process/stdlib-only, process-global keying, no `sleep`, silent throttled-refusal) are all reasoned through in both the code comments and `docs/mcp.md`, and match the repo's existing `#907`/`#913` precedents.

## Problems
None found. I did not find a real bug, logic error, or CLAUDE.md violation that should block merge.

## Code Quality
- Docstrings are extremely dense/argumentative (multi-paragraph rationale in nearly every docstring). This matches the repo's established convention (≥95% docstring coverage requirement, Google-style with Args/Returns/Raises) and the PR's own prior precedent, so not something to change — just flagging it's a heavier style than typical Python projects use, for awareness.
- `test_mcp_purge.py`'s five `_attempt_*` helper functions are boilerplate-heavy but each is legitimately needed to parametrize over tool signatures — fine as is.

## CLAUDE.md Adherence
✅ TDD workflow followed and documented (RED-first evidence in the PR description, including the surrogate regression case).
✅ Stated `./scripts/check-all.sh` result: 12/12 gates, 8664 passed, 94.35% branch coverage, `attempt_policy.py` 95.74%, `auth.py` 100%, no new coverage waiver.
✅ No `# noqa` / `# type: ignore` / `# pylint: disable` / skip markers introduced (confirmed absent from the diff).
✅ No dependency or lockfile changes (module is deliberately stdlib-only, verified by a structural test).
✅ Google-style docstrings with Args/Returns throughout new code.
✅ Conventional-commit-style commit message, PR targets a feature branch (not a direct push to `main`).
⚠️ At review time, CI (`Tests & Type Checking` 3.11/3.12/3.13, `Static Analysis`, `Pylint`) was still `IN_PROGRESS` on GitHub; `Integration & E2E`, `CrawDad Quality`, `Code Complexity`, and `Build Distribution` were already green. Worth confirming the remaining checks land green before merge, though given the reported local `check-all.sh` run this is expected to pass.
ℹ️ The PR body itself flags that its base (`bc83b7b`) is behind current `main`, since `remote_auth.py` was rewritten on a sibling lane. I checked: this diff does not touch `remote_auth.py`, and the "mirrors `remote_auth._now`" comment in `attempt_policy.py` is naming-convention-only (that module deliberately uses `time.time`, this one deliberately uses `time.monotonic` — already called out in the code as an intentional difference), so there's no functional coupling to break. A rebase before merge is still good hygiene but not required for correctness.

## Testing
Thorough and well above the bar. New coverage spans: unit tests of `AttemptBudget` in isolation (window edges, no-extension, reset, sustained-attack rate, concurrency), integration tests through `auth.is_elevated`, and end-to-end tests through all five MCP purge tools (shared-budget-across-tools, byte-identical refusal payloads, audit-log completeness under throttling, budget-restoration on success, lockout expiry). Structural AST tests close off regressions that behavioral tests alone can't catch (no `sleep`, stdlib-only, single `compare_digest` call site, no rogue env readers). The `FakeMonotonicClock`/`VerifySpy` test doubles in `tests/elevated_attempt_support.py` are clean and specifically justify why they don't import from `creek_mcp` (keeps two test modules collectible during the RED window). I did not find a gap.

## Documentation
`docs/mcp.md` is updated with a full, honest accounting of the new behavior: the exact constants, which denial modes count toward the budget and why, the byte-identical-refusal guarantee, the three stated residuals (process-restart reset, process-wide lockout blast radius, sustained-attacker rate with the 92%-unavailable number), and the operator recovery path (`creek purge` CLI bypasses this gate entirely). Matches the implementation exactly everywhere I checked.



## Verdict: LGTM

The state machine, regression fix, and every design claim in the PR description check out against the actual code (purge.py confirmed a true zero-byte diff, lockout arm/disarm/no-extension logic hand-verified, attack-rate math confirmed), tests are exhaustive (behavioral, structural/AST, and concurrency), and no CLAUDE.md quality bar is violated. The only open item is that a few CI jobs were still IN_PROGRESS at review time — not a code defect, and the author already reports a green local check-all.sh run — so this is mergeable as-is pending those checks landing green.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234349777 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR bounds the previously-unlimited brute-force surface on `creek_mcp.auth.is_elevated`, which gates the destructive `creek.purge.*` MCP tools. A new stdlib-only `creek_mcp/attempt_policy.py` adds a process-global, non-blocking failed-attempt budget (5 consecutive denials → 60s lockout), `is_elevated` is refactored to delegate the actual constant-time comparison to a private `_verify_match` routed through that budget, and a latent `UnicodeEncodeError` on lone-surrogate tokens (which previously escaped the tool surface, skipped the mandatory audit entry, and became a state-dependent oracle once the budget existed) is fixed to fail closed. `creek_mcp/tools/purge.py` is untouched, verified by direct read of `_gate`.

## Strengths
- **Root-caused, not papered over.** The fix targets the actual gap (unlimited guess *rate*, not the timing side-channel `hmac.compare_digest` already closed) and explains why a sleep-based backoff was rejected (purge tools are registered `sync` and FastMCP invokes them inline on the asyncio loop — confirmed at `creek_mcp/server.py:731+` — so a blocking backoff would freeze the whole server).
- **Structural, not just behavioral, guarantees.** AST-based tests assert: no function outside the one private helper reaches `hmac.compare_digest`; no `time.sleep` (including aliased imports) exists in `attempt_policy.py`; `attempt_policy.py` imports nothing from `creek_mcp` (no import cycle); and no function outside an explicit allowlist reads `CREEK_MCP_ELEVATED_TOKEN`. These catch regressions that a purely behavioral suite would miss.
- **The lockout doesn't create a new oracle.** `_verify_match` is a verbatim move of the old comparison, `purge.py`'s refusal payload keys are asserted to stay exactly `{status, tool, reason}` (no `throttled`/`retry_after`/`attempts_remaining`), and a throttled refusal is asserted byte-identical to an ordinary one, including for the *correct* token.
- **Honest, quantified threat model.** The PR body and `docs/mcp.md` state the residual (process restart resets the counter; a sustained attacker keeps the gate ~92% closed to everyone) with actual measured numbers rather than hand-waving, and a test (`test_a_sustained_attacker_is_held_to_the_documented_guess_rate`) pins those numbers as a regression guard — deleting the re-arm makes the attacker's yield jump from 285 to 3,541 guesses/hour, so the test has a real trip wire.
- **Concurrency correctness is tested, not assumed** (`test_concurrent_attempts_cannot_overspend_the_budget`), and the window's non-extension property (an attacker "knocking" during a lockout must not perpetually re-arm it and brick the operator) is explicitly covered.
- **TDD discipline evidenced**: PR body documents RED-first runs, and tests explicitly reference "the RED window" for deferred imports so pre-existing tests in `test_mcp_auth.py`/`test_mcp_purge.py` fail on assertions rather than on collection errors.

## Security Concerns
🟢 **LOW** — Timing side-channel on lockout state (disclosed, accepted). A throttled refusal returns before `_verify_match` runs, so it is measurably faster than an ordinary refusal, leaking "the lockout is armed" — never token material. This is explicitly documented in the module docstring and PR body as an accepted residual, and is a strict improvement over the prior unlimited-rate state.

🟢 **LOW** — Process-global lockout is itself a (deliberate) availability tradeoff: a partially-trusted consumer with transport access but no elevated token (canonically CrawDad) can force `creek.purge.*` closed to the legitimate operator for spans of a sustained attack. This is not a hidden flaw — it's argued at length, measured (~92% closed under sustained 1Hz attack), and the documented operator recovery path (`creek purge` CLI, which drives `PurgeEngine` directly and bypasses this MCP gate) is real and verified to exist independently of this gate.

No blocking security issues found. The `UnicodeEncodeError`-on-surrogate fix closes a real fail-open/audit-skip defect and is the most concretely dangerous thing in the diff — it's fixed correctly and tested via parametrized cases covering both the throttled and unthrottled states.

## Problems
None found that block merge.

## Code Quality
- Docstrings throughout (`attempt_policy.py`, `is_elevated`, several test docstrings) are extremely long and rhetorical — closer to design-doc prose than API documentation. Not incorrect, but a lighter touch would improve scanability. Non-blocking.
- The reviewer note in the PR body flags that `main` has moved (`remote_auth.py` rewritten on a sibling lane) since this branch's base (`bc83b7b`); worth a rebase/merge-conflict check before merging even though no file in this diff overlaps that module.

## CLAUDE.md Adherence
✅ TDD workflow followed (RED-first per PR body, verified structurally in the deferred-import patterns in tests).
✅ Commit messages follow Conventional Commits (`fix(mcp): ...`).
✅ No dependency or lockfile changes (`uv.lock` untouched, confirmed via file list).
✅ Only `tests/*.py` and the two `creek_mcp` production modules plus docs touched — scope matches the PR's stated claims, verified directly against the repo (`purge.py` zero-byte diff confirmed, `token_policy.MIN_TOKEN_LEN` unchanged, sync tool registration at `server.py:731+` confirmed).
⚠️ Could not execute `./scripts/check-all.sh` / `pytest` directly in this review sandbox (tool execution required approval that wasn't grantable here), so the PR's claimed coverage numbers (94.35% branch, 95.74%/100% per-file) are taken on the strength of the test suite's structure and completeness rather than independently re-run. The test suite itself, read in full, is unusually thorough for this class of change.

## Testing
Excellent. New tests cover: policy-unit level (window arming/closing/reopening exactly at the boundary, reset semantics, concurrency, sustained-attack rate, never-raises across every failure mode), auth-gate level (structural AST ratchets, parametrized never-raises matrix including the surrogate regression case, lockout interaction with weak/absent/wrong tokens), and purge-tool level (shared-budget-across-all-five-tools, byte-identical refusal payloads, audit-log completeness during throttling, budget reset on success). The `FakeMonotonicClock`/`VerifySpy` test doubles are well-designed to avoid real sleeps and to make the "verify not called while locked" negative property observable.

## Documentation
`docs/mcp.md` update is thorough and consistent with the code and tests — same numbers (5 attempts, 60s, 4.75 guesses/min, 92% closed) appear in both the docs and the pinned regression test, so they can't silently drift apart. Module and function docstrings explain the *why* behind non-obvious decisions (stdlib-only, process-global keying, monotonic vs wall clock, no-sleep constraint) rather than restating the code.



## Verdict: LGTM

The fix is correctly scoped, verified against the actual source (purge.py's `_gate` is untouched and still doesn't catch exceptions, the five purge tools are registered as sync `def`s per server.py:731+, `MIN_TOKEN_LEN` is unchanged, and no dependency/lockfile files were touched). The DoS tradeoff of a process-global lockout is a deliberate, extensively documented and measured design decision rather than an overlooked defect, and the accompanying test suite is unusually rigorous (structural AST checks, hand-driven clock, concurrency, byte-identical-refusal, and a mutation-tested re-arm regression). Nothing here blocks merge; the only remarks are minor nits.

Geoffe-Ga:
**CI**: 10/13 Green
**VERDICT**: LGTM
**Action**: pull comment 5234414198 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 14/14 Green
**VERDICT**: LGTM

Issue #1282, opened 2026-08-09
CompiledPage carries no privacy_tier; downstream safety rests on a default in another module


Geoffe-Ga:
## Context

Raised by the security review of #931; pre-existing, not introduced there.

## The gap

`CompiledPage` (`creek/models.py`) carries no `privacy_tier` field. A compiled page under `02-Threads` / `03-Eddies` / `06-Frequencies` is therefore written with no tier of its own, even when its sources — or, since #931, its sources' *ancestors* — are above `open`.

This is currently fail-closed downstream only by accident of a default living in another module: the report and read gates read a note's tier off raw frontmatter via `creek.classify.privacy_filter.raw_privacy_tier`, whose missing-key branch returns `INTIMATE`. So an untiered compiled page is treated as intimate by every reader that goes through that helper — but nothing asserts that the compiled-page writer and that default agree, and a reader that used `Fragment`-style model defaults instead would see `unclassified` and admit the page at `ceiling=personal`.

#931 makes this slightly more load-bearing on the CLI side: the `creek compile` CLI deliberately *retains* the `structural_path` breadcrumb and escalates its LLM routing tier instead, so a compiled page can legitimately contain an intimate ancestor's heading. The routing decision protects egress; it says nothing about the tier of the artifact left on disk.

## Suggested fix

Give `CompiledPage` an explicit `privacy_tier`, derived from the same reduction `creek.compile.engine._routing_tier_for` already computes, and write it into the page frontmatter. Then pin reader/writer agreement with a test in the style of `tests/test_mcp_report_tier_ceiling.py::test_raw_and_model_tier_readers_agree_on_every_fragment`.

Refs #931, #968

Geoffe-Ga:
Audited at HEAD `91df620`. **The narrow fact is real; most of the surrounding framing is not, and the issue splits cleanly into a half a lane can do and a half only the owner can decide.**

**Real and unfixed:** `CompiledPage` has no `privacy_tier` field, `_write_compiled_page` emits no such key, and **no test asserts what any reader makes of that.**

**Four corrections:**
[...]

Issue #1278, opened 2026-08-09
MCP creek.compile walks 01-Fragments twice (gate walk + engine walk)


Geoffe-Ga:
## Context

Pre-existing, already noted in `creek_mcp.tools.compile._survey_sources` and re-confirmed while fixing #931.

## The gap

An MCP `creek.compile` call walks `01-Fragments` twice:

1. `_survey_sources` -> `creek.classify.privacy_filter.ancestry_tiers` -> `iter_vault_fragments`, for the admission gate;
2. `creek.compile.engine._load_fragments_for_compile` -> `iter_vault_fragments`, for the compile itself.

`iter_vault_fragments` rglobs and fully parses every fragment file into a list before returning, so on the 35k-fragment vault that is two complete corpus parses per call. #931 folded the ancestor index into the engine's existing walk so the CLI stays at exactly one walk (pinned by `tests/test_compile.py::test_compile_to_vault_walks_the_vault_exactly_once`), but it deliberately did not restructure the MCP boundary.

Collapsing it also closes the admission-side TOCTOU window `_survey_sources` documents: a concurrent writer that raises a fragment's tier between the gate walk and the engine walk is currently admitted under the pre-write tier.

## Suggested fix

Give `compile_to_vault` an optional pre-loaded corpus / `AncestorIndex` parameter so the MCP wrapper can pass what it already walked, keeping the gate above every state-affecting step. Do not change the CLI path, which is already at one walk.

Refs #931, #848

Issue #1277, opened 2026-08-09
creek lint should surface dangling, cyclic and orphaned parent_id ancestry


Geoffe-Ga:
## Context

Follow-up from #931, which made ancestry tier-ranking part of the compile gate.

`creek.classify.privacy_filter.AncestorIndex.chain_tiers` fails closed on any ancestry chain it cannot fully survey:

- a `parent_id` naming a fragment that is missing, unreadable, non-`fragment`-typed or schema-invalid (rule (c));
- a `parent_id` cycle (rule (d));
- a non-empty persisted `structural_path` with `parent_id is None` (rule (e), the orphan breadcrumb).

Each of those contributes `PrivacyTier.INTIMATE`, so the whole subtree beneath the broken link becomes uncompilable at cloud tiers and is refused at an `open` MCP ceiling with the content-free `_ABOVE_CEILING_REASON` — a refusal that, by design, names nothing.

## The gap

That is the correct privacy answer and a poor debugging experience. An operator whose vault holds one hand-edited fragment sees "source fragments exceed tier ceiling" for a fragment that is plainly `open`, with no way to find the broken link.

## Suggested fix

Teach `creek lint` to report:

- `parent_id` values that resolve to no loadable fragment;
- `parent_id` cycles;
- fragments with a non-empty `structural_path` and no `parent_id`.

`creek.atomize.split._build_children` is the only writer of `structural_path` in `creek/` + `creek_mcp/` and always sets `parent_id` in the same `model_copy`, so every one of these states is anomalous by construction and worth surfacing regardless of compile.

Refs #931

PR #1283, opened 2026-08-09
fix(compile): rank ancestry so an admitted child cannot leak its parent's title

### LINKED ISSUE #931: structural_path ancestry leaks above-ceiling parent titles into the compile prompt via an admitted child

Geoffe-Ga:
## Summary

- **Closes the ancestry leak by ranking ancestors, not by redacting the breadcrumb.** `creek/hierarchy.py::structural_path_context` prefers a fragment's *persisted* `structural_path` — ancestor headings `creek.atomize.split._build_children` baked in at ingest — and `creek/compile/engine.py::_build_prompt` emits them after `structural_path:`. #848's gate ranks only the ids a caller **names**, so an `open` child of an `intimate` parent was admitted at `ceiling=open` and shipped its parent's heading to a cloud-routed provider. One new chokepoint in `creek/classify/privacy_filter.py` — `AncestorIndex` / `build_ancestor_index` / `ancestry_tiers` — now walks `parent_id` and reports the whole chain's tiers; both compile entry points reduce over it with the existing `max_source_tier` / `write_tier_allowed`. Neither call site contains any ancestry logic of its own.
- **The two callers get two different, deliberately asymmetric answers.** The MCP `compile_tool` **refuses the whole call** — its caller does not own the vault, and refusal is what #848 already means there. The ceiling-less `creek compile` CLI **keeps the breadcrumb and escalates its LLM routing tier** to the ancestor's, so the ancestry's real orienting value stays with the vault owner while the call is forced onto a local model — or raises `IntimateRoutingError` when there is no local backend. Redacting on the CLI side would silently degrade the operator who owns the content; routing-only on the MCP side would hand a summariser content the caller was never admitted to. If a reviewer cannot find that sentence, the fix is half-designed.
- **No new walk anywhere.** The index is built from the records `_load_fragments_for_compile` already walks, so the CLI stays at exactly one `01-Fragments` pass and the MCP stays at the two it already had. `tests/test_compile.py::test_compile_to_vault_walks_the_vault_exactly_once` pins it; at the 35k-fragment bar a second rglob-plus-parse would double the pre-LLM wall clock, and a privacy fix has no business shipping that regression.

### What the fix does *not* change

`_ABOVE_CEILING_REASON` is byte-identical, the audit payload is identical, `_SourceGate` keeps its single field, `_survey_sources` keeps its pinned gate symbol, and `CONTRACT_VERSION` stays at `0.4.0`. **An ancestor refusal is indistinguishable from a named-id refusal** in the response, the audit row and the probe cost — a distinct reason string would be a fresh oracle telling the caller "the offender is above you in the tree", which is strictly more than the bool #848 accepted. `chain_tiers` never short-circuits and never does a lazy per-parent lookup, so the uniform-probe-cost property that keeps the refusal from leaking position through timing survives (rule (f) in its docstring).

`source_tiers` was deliberately **not** widened. `draft` / `journal` / `upload` reduce over it and render no breadcrumb, so ancestry-ranking there would refuse calls that leak nothing and would silently change three tools this issue never analysed. `tests/test_privacy_filter.py::test_source_tiers_stays_ancestry_blind` pins the difference so a future de-duplication has to argue with a test.

### Two hardenings the pre-push security review earned

Both were found by the Gate 2.5 security pass, and both are strict narrowings:

- **Duplicate fragment ids were the one place this change was *less* restrictive than what it replaced.** `source_tiers` yields one tier per *file*, so two files claiming one id contributed both tiers and the more sensitive won. An index keyed on `fragment.id` is last-wins, which would have let a shadow file carrying an above-ceiling ancestor's id with `privacy_tier: open` and a later sort position downgrade the real ancestor while the child's breadcrumb still rendered its heading. Rule (h) now fails a colliding id closed to `INTIMATE`. Proved red: without it, `chain_tiers` returns `['open', 'open']` for exactly that shadow-file input.
- **Breadcrumb depth is now bound to the ancestry actually walked.** The survey ranks *ids*; the prompt renders *strings*, and nothing in the data binds the two. The original rule (e) caught only the depth-0 case (`parent_id is None` with a breadcrumb). It is now the general count check — `len(structural_path)` may never exceed the strict-ancestor count, which `creek.atomize.split._build_children` guarantees by appending at most one heading per level — so a fragment re-parented onto a shallower `open` parent while keeping a deeper breadcrumb also fails closed. No production writer can produce that state today; the point is that the fix no longer *rests* on that being true.

### Consequences worth stating plainly

1. **Walking `parent_id` over-covers.** It ranks the root document too, whose title the persisted breadcrumb never carries. A vault with `intimate` top-level documents will now route descendant compiles local, and refuse them at an `open` MCP ceiling. That is a strict narrowing, chosen over an entry-level attribution the data cannot support: persisted breadcrumb entries are bare strings with no owning-fragment id, so "strip the entries whose owning fragment is above the ceiling" (the issue's cheaper option) is not implementable as written — mapping entry to ancestor requires the same `parent_id` walk, and then still has to guess.
2. **A broken ancestry link makes a subtree uncompilable at cloud tiers**, with a refusal that names nothing. A dangling, cyclic or orphaned `parent_id` all fail closed to `INTIMATE` (rules (c), (d), (e)). That is the right privacy answer and a poor debugging experience, so `creek lint` surfacing those states is filed as **#1277**.
3. **`creek/hierarchy.py` gains a `visited` set**, kept in this PR rather than split out. The file genuinely hangs today on a `parent_id` cycle whose members are all in `by_id` — the loop's only exit was a *missing* parent, and in a cycle every parent is present; the new test hung the runner before the fix. It is the direct sibling of the walk this PR adds, and shipping two ancestry walks with different cycle semantics is exactly the two-readers-can-disagree drift `privacy_filter`'s module docstring exists to prevent. `structural_path_context`'s signature, return type and behaviour are otherwise untouched.

### Answering the issue's audit bullet

> Worth auditing whether the same ancestry channel exists in other tools that render `structural_path`.

It does not. `grep -rn structural_path creek/ creek_mcp/` gives exactly one production caller of `structural_path_context`: `creek/compile/engine.py`. No other tool renders a breadcrumb. That is now pinned by a test rather than noted in a comment, so a future second renderer cannot silently re-open the channel.

### Corrections to the issue body

The issue points at `creek/compile/hierarchy.py`, which does not exist — the function is `creek/hierarchy.py::structural_path_context`. It also describes the mechanism as "a list of parent titles", which aims the fix at the wrong branch: the `parent_id`-walk branch only fires when the parent is in `_build_prompt`'s `by_id`, i.e. when the caller *named* it, in which case #848 already refuses. The MCP-reachable channel is the **persisted-field** branch, which needs no ancestor fragment in memory at all. The pre-existing `tests/test_compile.py::test_compile_to_vault_routes_on_a_parent_the_level_policy_dropped` covers the already-safe branch; only its docstring changed here, and both of its assertions — including `assert f"structural_path: {_INTIMATE_TITLE}" in prompt` — survive verbatim, because that call names both fragments and this PR does not change its behaviour at all.

## Test plan

Gate 1, red first. Every new test was authored and run before any source edit, and the red was re-proved after the fix by restoring `creek/compile/engine.py` and `creek_mcp/tools/compile.py` to `git merge-base HEAD origin/main` (no `git stash` — one shared ref across four live worktrees). Six failures, each for the right reason:

- `test_compile_to_vault_routes_intimate_on_a_persisted_ancestor_not_named` — `assert ['personal'] == [PrivacyTier.INTIMATE]`: a **cloud** client built with `structural_path: Therapy session with Dana` in the prompt.
- `test_compile_to_vault_refuses_an_unnamed_intimate_ancestor_with_no_local_backend` — `RuntimeError: ANTHROPIC_API_KEY ... not set`, i.e. it reached the Anthropic client constructor instead of raising `IntimateRoutingError`.
- Four MCP rows (`..._at_the_open_ceiling`, `..._leaves_zero_state_and_one_clean_audit_row`, `..._a_dangling_ancestry_link`, `..._an_orphan_breadcrumb`) — `assert 'ok' == 'refused'`.
- `tests/test_hierarchy.py::test_parent_id_cycle_terminates_without_repeating_an_ancestor` did not fail, it **hung** the runner — the pre-fix walk spins forever on a mutual-parent pair.

New coverage:

- `tests/test_mcp_write_tools.py` — the narrowest-ceiling refusal with `factory.calls == 0` (status alone is a false green); zero on-disk state (no page, no `compile-<id>.hash`, no paradox log) with a stub LLM that *would* write all three; exactly one `creek.compile` audit row with `fragment_ids` collapsed to a count and `created_path` / `created_tier` / `affected_fragment_ids` / `target_title` absent, the ancestor's heading and id absent from the raw log, and `verify_mcp_audit_chain` intact; the dangling-parent and orphan-breadcrumb fail-closed rows; an **admit** row so the fix cannot be "refuse everything with a parent"; and a row proving the ancestry gate still loses to the engine's not-found `ValueError`.
- `tests/test_compile.py` — the CLI routing escalation (breadcrumb retained, tier `INTIMATE`), an `IntimateRoutingError` sibling with the same anti-vacuity assertion as the #962 original, an anti-vacuity `OPEN`-stays-`OPEN` row, and the one-walk counter.
- `tests/test_privacy_filter.py` — one test per `chain_tiers` rule (a) through (h), plus the ancestry-blindness pin on `source_tiers` and a single-walk pin on `ancestry_tiers`. The two hardening rules were separately proved red by reverting only their own branch: `['open', 'open']` in both cases. Their anti-vacuity siblings (`..._admits_a_breadcrumb_matching_its_ancestry_depth`, `..._leaves_a_breadcrumbless_root_alone`) stayed green throughout, so the depth check is not off by one — if it were, every real split fragment in the vault would become uncompilable at cloud tiers.
- `tests/test_hierarchy.py` — mutual-parent and self-parent cycle termination, and the one-production-caller pin on `structural_path_context`.

Gate 2, in a CI-equivalent env with all provider keys and `CREEK_*` consent vars stripped: `./scripts/check-all.sh` exits 0 — **12/12 checks passed**, 8679 unit tests, 94.37% branch coverage, per-file gate green on all 255 files, mypy strict, xenon, ruff lint + format with `--no-cache`, refurb, bandit, pip-audit. Plus `./scripts/test.sh --integration` (356 passed) and `./scripts/test.sh --e2e` (14 passed).

Test integrity: `git diff $(git merge-base HEAD origin/main)` removes **zero** assertions. The only deleted lines under `tests/` are one rewritten docstring sentence and the two-line `_write_fragment` helper signature it extends (new kwargs are emitted only when supplied, so no existing caller's frontmatter changes).

## Notes for the reviewer

- Same `Intimate`-never-cloud invariant as **#1152**, which is in flight on the crawdad side. Separate venv, separate `check-all.sh`, separate CI job — a green gate here says nothing about that lane and vice versa. Cross-referenced rather than restated.
- Touches `creek_mcp/read_gate.py` (rationale prose only; `gate_module` / `gate_symbol` unchanged), which **#1036** also touches. #1036's prompt-side canary is the natural regression harness for this leak; serialise the two lanes.
- `ancestry_tiers` is plausibly reusable by **#971** (`creek.skills.refresh` never threads `privacy_tier_ceiling`). Flagged, not coupled.
- Follow-ups filed: **#1277** (`creek lint` dangling/cyclic/orphaned `parent_id` — the diagnostic gap behind consequence 2 above), **#1278** (the pre-existing MCP gate-walk + engine-walk duplication), **#1282** (`CompiledPage` carries no `privacy_tier`; downstream safety currently rests on `raw_privacy_tier`'s missing-key default in another module).
- Gate 2.5 ran `code-review-orchestrator` (returned `CLEAN` across privacy monotonicity, oracle, probe cost, rule correctness, scale, layering, test integrity and anti-vacuity) and `security-specialist` (no blocker; its two actionable findings are the hardenings above, its two documentation findings are folded into the `_ABOVE_CEILING_REASON` comment and `_routing_tier_for`'s docstring, and its informational finding is #1282).
- Planning pass: this change was executed from an independent adversarial premise-and-design audit of #931 (which returned `partly` on the issue's own premise and produced the judged `hybrid` design implemented here), not from a `chief-architect` spawn.

Closes #931
Refs #848, #962, #647

### DIFF
--- a/creek-tools/creek/classify/privacy_filter.py
+++ b/creek-tools/creek/classify/privacy_filter.py
@@ -32,6 +32,12 @@ an explicit ``unclassified`` only in the raw frontmatter — is ``INTIMATE``),
 :func:`max_source_tier` (the reduction over the tiers a call would carry,
 ``INTIMATE`` when empty).
 
+:func:`build_ancestor_index` / :class:`AncestorIndex` / :func:`ancestry_tiers`
+(#931) are the ancestry-aware sibling of :func:`source_tiers`, for the one
+caller — ``creek.compile`` — whose prompt renders a fragment's *ancestors*
+alongside the fragment itself. They exist as a separate survey rather than as a
+widening of :func:`source_tiers` on purpose; see that function's docstring.
+
 :func:`raw_privacy_tier` and :func:`within_ceiling` (#968) are the
 raw-frontmatter siblings of that pair, for generation flows that never build a
 :class:`~creek.models.Fragment` at all. They live here, and not in a new
@@ -73,7 +79,7 @@ from dataclasses import dataclass
 from datetime import UTC, datetime
 from enum import StrEnum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, TypeGuard
+from typing import TYPE_CHECKING, Any, Final, TypeGuard
 
 from creek.audit import AuditLog
 from creek.models import PrivacyTier
@@ -513,6 +519,20 @@ def source_tiers(vault_path: Path, fragment_ids: Iterable[str]) -> list[PrivacyT
     naming a thousand ids still pays one walk of the corpus rather than a
     thousand.
 
+    **This survey is deliberately ancestry-blind, and #931 kept it that
+    way.** It reports the tiers of the ids a call *names* and nothing else.
+    Its callers — ``creek.draft``, ``creek.journal``, ``creek.upload`` —
+    hand the LLM the named fragments' own content and render no
+    ``structural_path`` breadcrumb, so an ancestor's title never reaches
+    their prompts and ranking ancestry there would refuse calls that leak
+    nothing. ``creek.compile`` is the one caller whose prompt *does* render
+    ancestry, and it uses :func:`ancestry_tiers` instead. The two are two
+    functions rather than one flag because they answer two different
+    questions; ``tests/test_privacy_filter.py``'s
+    ``test_source_tiers_stays_ancestry_blind`` pins the difference so a
+    future de-duplication has to argue with a test rather than with a
+    comment.
+
     The walk deliberately goes through that shared loader — the same one
     ``creek.compile.engine._load_fragments_for_compile`` uses — so the set
     of files this survey inspects and the set the compile engine would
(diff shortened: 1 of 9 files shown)


Geoffe-Ga:
## Summary
This PR closes a real privacy leak (#931): `creek compile`'s prompt renders a fragment's *persisted* `structural_path` breadcrumb (ancestor headings baked in at ingest), which the pre-existing #848 gate never ranked because it only inspects *named* fragment ids. An `open` child of an `intimate` parent could route to a cloud LLM carrying the parent's title. The fix adds a new `AncestorIndex` / `ancestry_tiers` survey in `creek/classify/privacy_filter.py` that walks `parent_id` and ranks the whole ancestor chain, wired into both the CLI (`creek/compile/engine.py`, escalates routing tier) and the MCP tool (`creek_mcp/tools/compile.py`, refuses the call outright) — plus a cycle-termination fix in `creek/hierarchy.py::structural_path_context` that a new regression test would otherwise hang on.

## Strengths
- The core algorithm (`AncestorIndex._ascend` / `_chain_from` / `chain_tiers`) is correct: I traced it by hand against the eight documented rules (a–h), including the two-node mutual-parent cycle, self-parent cycle, memoized cross-request chain sharing, and the depth-vs-breadcrumb-length check (rule e) — all check out, and termination is guaranteed by the per-ascent `visited` set even without memoization.
- Deliberately asymmetric fix per caller (CLI escalates+keeps breadcrumb for the owning operator; MCP refuses for a non-owning caller) is a sound design and clearly justified.
- Verified `iter_vault_fragments` (`creek/vault/reader.py:77-107`) returns a materialized `list`, not a one-shot generator — so `_load_fragments_for_compile`'s reuse of `records` for both the `by_id` dict comprehension and `build_ancestor_index`'s generator expression is safe and genuinely walks the vault once, matching the PR's performance claim.
- Verified refusal ordering in `creek_mcp/tools/compile.py`: `_survey_sources` (now `ancestry_tiers`-backed) runs and can refuse before `compile_to_vault`/LLM-factory/write, matching the "content-free refusal, zero side effects" claim.
- `CONTRACT_VERSION` is confirmed unchanged (0.4.0) — consistent with the PR's claim that the refusal shape/payload is unchanged, only its trigger condition widened.
- Extremely thorough, rule-by-rule test coverage (`test_privacy_filter.py` has one test per rule plus anti-vacuity siblings for both the escalation and the depth check) — the anti-vacuity pattern (proving the fix doesn't just "refuse everything") is genuinely good practice, not just padding.
- Duplicate-fragment-id hardening (rule h) is a legitimate strict-narrowing catch that a security pass found; verified `build_ancestor_index` correctly forces `INTIMATE` on any colliding id.

## Security Concerns
🟢 LOW — `_AncestorEntry`'s docstring documents an attribute named `has_structural_path`, but the actual field is `breadcrumb_len: int` (`creek/classify/privacy_filter.py`, the `_AncestorEntry` dataclass). Purely a stale docstring (likely left over from an earlier bool-based design); the code itself is internally consistent and correct. Not a functional or security issue, just worth a follow-up fix.

🟢 LOW — Residual gap (not introduced by this PR, and arguably out of #931's stated scope): if a fragment were re-parented to a *different* parent at the *same* depth while its persisted `structural_path` retained the *old* parent chain's heading text, rule (e)'s depth-count check would not catch it — the ranking would reflect the new (current) `parent_id` chain's tiers, not the tiers of whatever ancestor the stale breadcrumb text names. The PR's own documentation acknowledges adjacent cases ("no production writer can produce that state today") for the general depth rule, so this is consistent with the fix's stated boundaries, but it's worth flagging since it's a case where breadcrumb text and ranked ancestry could theoretically diverge without any of the fail-closed rules firing.

## Problems
None found that block merge. The implementation matches its own extensive documentation, the single-corpus-walk performance claim is verified true, and the refusal/escalation logic is exercised by tests that actually pin the right behavior (verified by manual trace, not just reading the assertions).

## Code Quality
- Fix the `has_structural_path` → `breadcrumb_len` docstring mismatch in `_AncestorEntry` (noted above) for consistency.
- The module is very dense with prose (docstrings routinely longer than the code they describe). This is a deliberate house style visible throughout the repo's other files too, and it does the job of explaining non-obvious invariants — but a future maintainer skimming `privacy_filter.py` will need real time investment to get through it.

## CLAUDE.md Adherence
✅ TDD workflow followed and clearly evidenced (red-first failures documented per test, restored-baseline re-proof described in the PR body).
✅ `./scripts/check-all.sh` reported 12/12 green with coverage well above the 90% branch-coverage floor; per-file gate and mypy strict also reported green.
✅ Integration & E2E CI jobs are green (`Integration & E2E Tests`, `Static Analysis`, `Pylint`, `Code Complexity Analysis`, `Build Distribution` all SUCCESS at review time).
⏳ The three `Tests & Type Checking` matrix jobs (3.11/3.12/3.13) were still IN_PROGRESS at the time of this review — not a red flag given everything else is green and local `check-all.sh` was reported passing, but worth confirming green before merge.
✅ No unrelated scope creep — the diff stays tightly on the ancestry-ranking fix plus the two narrowly-scoped hardenings the security pass required.

## Testing
Strong. New tests exist for: every documented `chain_tiers` rule (a–h) with matching anti-vacuity siblings, the CLI escalation and its `IntimateRoutingError` no-local-backend sibling, four MCP refusal scenarios (named ancestor, dangling parent, orphan breadcrumb, cycle) plus an admit-path anti-vacuity test and a not-found-still-wins-over-refusal ordering test, the single-walk performance pin, and the hierarchy cycle-termination fix (including the self-parent degenerate case). I independently traced the core `AncestorIndex` algorithm against several of these tests by hand and confirmed the logic produces the asserted results. Insufficient testing is not a concern here.

## Documentation
Docstrings are exhaustive to the point of being the dominant fraction of the diff, and they correctly describe the implementation in nearly every case I checked — the one exception being the stale `has_structural_path` attribute name noted above. The PR description itself doubles as a design rationale document and is unusually rigorous about stating what the fix does *not* do and why.



## Verdict: LGTM

The core algorithm is correct by manual trace against its own rule set, the single-corpus-walk and refusal-ordering claims are independently verified against the actual source (not just the PR's prose), test coverage is exhaustive with genuine anti-vacuity checks, and CI is green on every completed job (integration/e2e, static analysis, pylint, complexity, build) with only the three unit-test matrix jobs still in flight at review time. The two findings are LOW-severity, non-blocking: a stale docstring attribute name and a narrow, currently-unproducible data-integrity edge case the PR itself already discusses the boundaries of.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234447552 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Closes issue #931: a fragment's persisted `structural_path` breadcrumb (ancestor headings baked in at ingest by the splitter) could carry an above-ceiling ancestor's title into a compile prompt even when the caller never named that ancestor, bypassing #848's named-ids-only privacy gate. The fix adds a new `AncestorIndex`/`ancestry_tiers` survey in `creek/classify/privacy_filter.py` that walks `parent_id` and ranks the whole ancestry chain; the MCP `compile_tool` now refuses the call on an above-ceiling ancestor, while the CLI `creek compile` retains the breadcrumb but escalates its LLM routing tier (forcing local-only or raising `IntimateRoutingError`). A related infinite-loop bug in `creek/hierarchy.py::structural_path_context` (spins forever on a `parent_id` cycle) is fixed alongside it.

## Strengths
- The core algorithm (`AncestorIndex._ascend`/`_chain_from`/`_own_tiers`) is memoized, cycle-safe, and fails closed on every unsurveyable case (unresolved parent, cycle, orphan/deep breadcrumb, duplicate fragment id). I hand-traced it against ~10 of the new unit tests (rules a–h) and the results match exactly.
- Excellent negative/anti-vacuity test coverage: for every fail-closed rule there's a sibling test proving the same shape *doesn't* escalate when it shouldn't (e.g. `test_chain_tiers_admits_a_breadcrumb_matching_its_ancestry_depth`, `test_compile_admits_a_within_ceiling_ancestor`), which rules out "refuse everything" as a degenerate passing implementation.
- The asymmetric design (MCP refuses vs. CLI escalates-and-keeps-breadcrumb) is well-reasoned and documented at the point of decision, not just in the PR description.
- Performance-conscious: the ancestor index is built from records the existing `_load_fragments_for_compile` walk already produces (a real `list`, so iterating it twice is safe — verified `iter_vault_fragments` returns `list[...]`, not a generator), and `test_compile_to_vault_walks_the_vault_exactly_once` / `test_ancestry_tiers_reads_the_vault_in_one_walk` pin that no second corpus pass was introduced.
- The refusal payload/audit row is deliberately indistinguishable between a named-id violation and an ancestor violation, closing a potential timing/information side-channel — and this is verified by a dedicated test comparing audit log contents.
- `structural_path_context`'s new `visited` set is a real, separate bug fix (an actual infinite loop on a `parent_id` cycle) with its own regression test that previously hung the test runner.
- Docstrings are extremely thorough — arguably more thorough than necessary, but genuinely useful for a privacy-critical code path (see Code Quality below for the flip side).

## Security Concerns
🟢 None found — this PR is itself a fix for a genuine information-disclosure vulnerability (ancestor titles above the caller's ceiling leaking into cloud-routed prompts), and the fix closes the hole without introducing new gaps I could find. The duplicate-fragment-id handling (rule h) and breadcrumb-depth check (rule e) are hardenings against edge cases that would otherwise let the new mechanism be *less* strict than what it replaces — good catch by the PR's own pre-push review, and correctly implemented.

## Problems
No blocking correctness issues found. Two things worth flagging as non-blocking:
- 🟡 At the time of this review, CI's `Tests & Type Checking (3.11/3.12/3.13)` jobs were still `IN_PROGRESS` (other jobs — Static Analysis, Pylint, Integration & E2E, Complexity, Build — had already completed with `SUCCESS`). Worth confirming those finish green before merge, though nothing in this review suggests they won't.

## Code Quality
- The PR body and in-code docstrings are unusually dense and self-referential (extensive prose justifying nearly every line, cross-referencing issue numbers and other functions). This is a stylistic choice consistent with the rest of this module's existing docstrings (e.g. `source_tiers`, `max_source_tier` already do this), so it's consistent with established precedent rather than a new pattern — but it does make the diff much larger and slower to review than the functional change alone (310 new lines in `privacy_filter.py` for what is fundamentally a graph-walk + reduction). Not blocking, just a maintainability cost to be aware of going forward.
- `test_structural_path_context_has_exactly_one_production_caller` pins caller-count via a source-text grep (`"structural_path_context(" in file`) rather than a more robust mechanism (e.g. AST-based call detection). It works for the current codebase but is easy to accidentally satisfy/violate via string coincidence (e.g. a comment containing the literal text). Minor.
- `_own_tiers` and rule (e)'s "breadcrumb longer than walkable ancestry" check is a subtle invariant; the code correctly documents *why* it's a count rather than a boolean, which is good — but the logic would benefit from a one-line inline comment at the `entry.breadcrumb_len > depth` comparison itself, since the docstring justification is a full paragraph away from the actual code line.

## CLAUDE.md Adherence
✅ Coverage: PR claims 94.37% branch coverage (>=90% threshold), with per-file gate green — consistent with the extensive new test files.
✅ Docstring coverage: every new public/private symbol has a full docstring; consistent with >=95% interrogate threshold.
✅ Typing: type hints present throughout (`Final`, `Mapping`, dataclasses with typed fields); no obvious mypy-strict violations spotted (e.g. no untyped defs, `Any` used only in a pre-existing type alias).
✅ TDD workflow: PR description documents a red-first test plan with specific failure modes for each new test, matching the "Stay Green" 4-gate process.
✅ Commit conventions: commit messages follow Conventional Commits (`fix(compile): ...`, `docs(privacy): ...`).
❓ `./scripts/check-all.sh` full local run could not be independently reproduced in this review environment (no `uv`/`pytest` available), so I relied on manual trace-verification of the new algorithm against the test suite's assertions rather than executing it. This is a review-environment limitation, not a deficiency in the PR.

## Testing
Testing is thorough and well-targeted: new tests span `test_privacy_filter.py` (one test per `chain_tiers` rule a–h, plus anti-vacuity siblings), `test_compile.py` (CLI routing escalation, `IntimateRoutingError` path, single-walk pin), `test_mcp_write_tools.py` (MCP refusal, audit-row shape, dangling/orphan/duplicate-id fail-closed cases, and an admit case), and `test_hierarchy.py` (cycle termination, single-caller pin). I independently traced the core `AncestorIndex` walk logic against the test assertions rather than trusting the PR description, and every case checked out. No blocking test gaps found.

## Documentation
Documentation is comprehensive — arguably more extensive than the code itself — covering every design decision, tradeoff, and accepted risk directly in docstrings and inline comments. This is consistent with the surrounding module's existing style.



## Verdict: LGTM

This is a well-tested, correctly-implemented fix for a genuine privacy leak, with the fail-closed ancestry-ranking algorithm independently hand-traced against its test suite and found correct in every case checked, no performance regression (verified single-walk claims against actual list-vs-generator semantics), and full CLAUDE.md-required coverage/typing/docstring/commit-convention adherence. The only note (CI still finishing at review time) is informational, not a reason to block.

Geoffe-Ga:
**CI**: 11/13 Green
**VERDICT**: LGTM
**Action**: pull comment 5234656192 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 14/14 Green
**VERDICT**: NOT CURRENT

PR #1285, opened 2026-08-09
fix(crawdad): enforce the FEAT-027 redaction-scan gate at the dispatch boundary

### LINKED ISSUE #1054: crawdad: the FEAT-027 redaction scan gate is advisory — un-scanned batches still reach creek.ingest
**Labels:** `de-slop`, `crawdad-bot`, `bug`, `priority-high`
**Detected by:** de-slopify targeted scan (crawdad-bot) 2026-07-31
**Severity:** High
**Remediation:** decision-needed (Family 3 fake/aspirational safety gate, + Family 0)

## Problem

`_handle_attachments`' docstring (`crawdad/crawdad/bot.py:296-315`) advertises a
deterministic order *"so users get a predictable safety report **before any
ingest call is dispatched**"*, and step 3 promises: *"If MCP is unreachable or
the tool is not advertised, fall back to a clear soft error rather than silently
ingesting."*

`_run_safety_scan` (`bot.py:641-658`) returns a **soft-error string** —
`_REDACT_TOOL_MISSING_REPLY` or `_MCP_UNAVAILABLE_REPLY` — when the
`creek.redact.scan` tool is unadvertised or MCP is down, meaning **no redaction
scan ever ran**. But `_record_pending_batch` (`bot.py:351-358`) and
`_INGEST_CONSENT_PROMPT` (`bot.py:368`) fire unconditionally afterwards, and
`PendingBatch` (`consent.py:100-126`) carries no scan-state field. A one-word
`ingest` reply then dispatches `creek.ingest` on never-scanned content.

The bot's own two consecutive messages contradict each other: *"Run
`creek redact --scan <staging path>` from the terminal **before ingesting**"*
(`bot.py:73-77`), immediately followed by *"Reply with `ingest` … to proceed."*

**Current state:** the redaction gate is advisory, not enforced.

## Evidence (corroboration)

- **Signal 1 — runtime proof:** driving `handle_message` twice with
  `known_tools=("creek.ingest",)` (redact tool unadvertised): the staging turn
  issued **zero** MCP calls (no scan) and the last bot message was the consent
  prompt; the follow-up `content="ingest"` turn issued
  `['creek.ingest']`, replied `**Ingest results** - j.md (markdown): ok`, and
  left `store.get(999).state == "ingested"`. Zero scans, full ingest.
- **Signal 2 — structural + a test that pins the hole as intended:**
  `grep -n "scan" crawdad/crawdad/consent.py` finds no scan-state field on
  `PendingBatch`; and `test_attachment_path_replies_when_redact_tool_missing`
  (`crawdad/tests/test_bot.py:992-1020`) *asserts* the consent prompt is still
  sent after the missing-tool soft error.

## Scope

Covers the linkage between the safety scan's outcome and the consent/ingest
flow. Does NOT change the scan tool itself, the staging layout, or the redaction
pipeline in creek-tools.

## Tasks

1. **Decide and record the policy** (this is the decision-needed part): when the
   scan could not run, either (a) withhold the consent prompt entirely and tell
   the user to run `creek redact --scan` and re-upload, or (b) still record the
   batch but refuse the consent reply with an explicit "I can't ingest what I
   couldn't scan" message, or (c) allow it behind an explicit opt-in token
   distinct from `ingest`. Record the choice in `crawdad/CLAUDE.md`.
2. Add a scan-state field to `PendingBatch` in `crawdad/crawdad/consent.py`
   (e.g. `scanned: bool` or `scan_status: Literal["ok", "skipped", "failed"]`)
   and populate it from `_run_safety_scan`'s outcome.
3. Enforce the chosen policy in `crawdad/crawdad/bot.py` at the consent-reply
   handler, not only at the prompt.
4. Update `test_attachment_path_replies_when_redact_tool_missing` and add a test
   proving an un-scanned batch cannot reach `creek.ingest` under the chosen
   policy.

## Acceptance Criteria

- [ ] `_run_safety_scan`'s outcome is recorded on the pending batch.
- [ ] An `ingest` reply for a batch whose scan did not run cannot dispatch
      `creek.ingest` (test pins zero `creek.ingest` calls) under the chosen
      policy.
- [ ] The user-facing copy no longer contradicts itself.
- [ ] The chosen policy is documented in `crawdad/CLAUDE.md`.

## Files to Create/Modify

| File | Action |
|------|--------|
| `crawdad/crawdad/bot.py` | Modify |
| `crawdad/crawdad/consent.py` | Modify |
| `crawdad/tests/test_bot.py` | Modify |
| `crawdad/CLAUDE.md` | Modify |

Geoffe-Ga:
## Summary

FEAT-027's redaction-scan gate was **advisory**. When `creek.redact.scan` could
not run, the bot still recorded a `PendingBatch`, still sent the "Reply with
`ingest`" consent prompt, and a one-word `ingest` reply dispatched a full
`creek.ingest` of never-scanned content. Its own two consecutive messages
contradicted each other: *"Run `creek redact --scan` … before ingesting"*
immediately followed by *"Reply with `ingest` … to proceed."*

This makes the gate **enforced**, strictly in the more-restrictive direction.

**Policy — option (b) from Task 1, recorded in `crawdad/CLAUDE.md` §5.3:** still
record the batch (it stays inspectable and `cancel`-able) but refuse it at the
dispatch boundary.

**Mechanism**

- `PendingBatch.scanned` — a **required `bool` with no default**, placed ahead of
  the defaulted fields. Placement is load-bearing (a trailing required field is a
  `TypeError` at import). No permissive default and no `ConsentConfig` knob:
  either would let a construction site — or an operator — silently reopen the
  hole. Survives `with_resolved_types` / `with_state` / `with_ingested`, which
  are all `dataclasses.replace` copies.
- `_run_safety_scan` now returns a frozen `_ScanOutcome(scanned, text)` instead of
  a bare `str`, so the caller reads a status rather than string-matching
  `_MCP_UNAVAILABLE_REPLY` — a string shared with the ingest path
  (`_run_ingest_dispatch`) and the state path (`render_mcp_unavailable_reply`).
  All **four** failure modes yield `scanned=False`: no MCP client, tool
  unadvertised, `MCPUnavailableError`, and the bare `except Exception` arm.
- **One guard**, the first statement of `_dispatch_ingest_for_batch` — the sole
  convergence point of the consent flow.
- One new `_SCAN_BLOCKED_REPLY` constant serving **both** the unscanned
  attachment-turn close and the dispatch refusal, so the two messages are
  literally the same string and cannot drift apart.
- Docstrings corrected on `_handle_attachments` (steps 3–5 — the issue's headline
  complaint), `_run_safety_scan`, `_dispatch_ingest_for_batch`,
  `_apply_consent_reply`, `_apply_disambiguation_reply`, the FEAT-034 comment
  block, and `PendingBatch`.

### Two deviations from the AC's letter, both deliberate

1. **AC 3 asks that `_apply_consent_reply` "refuse to reach
   `_dispatch_ingest_for_batch`"; this guards the shared callee instead.** The
   issue names that one site because it missed the *second* dispatch entry point
   at `_apply_disambiguation_reply` — `ingest` → type question → `markdown`
   ingests through a different caller. `_dispatch_ingest_for_batch` has exactly
   two callers, so one branch satisfies both AC bullets, plus any future third
   caller, and cannot drift out of sync the way two gates would.
   *UX consequence:* an unscanned batch with an unresolved file type gets the
   type question first and the refusal on the type word — one wasted turn, still
   zero `creek.ingest` calls. A second guard to save that turn is a follow-up
   nit, not part of this fix.
2. **The new copy claims only the invariant actually enforced.** The gate means
   *"I couldn't run the redaction scan"* — **not** *"these files are clean"*.
   `creek_mcp/tools/redact.py` computes
   `status = "empty" if not scanned.matches else "ok"`, so a scan that *runs* and
   reports secrets still returns a permissive status and still clears the gate.
   Making the gate enforce on findings is a separate, larger decision; it is
   named as an explicit non-goal in §5.3 with a revisit predicate. Swapping the
   old false claim for a new one would have failed AC 3 just as badly.

Related, explicitly **not** covered: #1087 (`scan_batch` follows symlinks out of
the scan root — a scan that *runs* but under-reports; this gate marks such a
batch `scanned=True`). **#1088** *is* a fifth way the scan fails to run and this
design absorbs it with one more `_ScanOutcome(scanned=False, …)` arm — sequence
it next rather than touching `_run_safety_scan` twice.

## Test plan

**RED proof, before any source changed.** All five gate tests failed on the
behavioural assertion `assert calls == []` — not on a missing symbol — each with
a real dispatch recorded:

```
E  AssertionError: assert [{'name': 'cr...ng': 'open'}}] == []
E    Left contains one more item: {'name': 'creek.ingest', 'args':
E      {'source_type': 'document', 'input_path': '…/weird.xyz',
E       'privacy_tier_ceiling': 'open'}}
```

That is the issue's own Signal-1 runtime proof, reproduced at the **narrowest**
declared ceiling (`open` is the most restrictive `TierCeiling`) rather than the
default `personal`.

**Guard proven load-bearing.** With only the two guard lines commented out:
exactly the 5 gate tests go red and the other 61 in `test_bot.py` — including
every `_stage_attachment`-seeded happy-path consent test — stay green. Restored
and re-verified.

New coverage:

- All four scan-failure modes reach zero dispatch: tool unadvertised,
  `MCPUnavailableError` mid-scan, generic exception mid-scan, `mcp_client is
  None`.
- The type-disambiguation route (`ingest` → type question → `document`) also
  reaches zero dispatch.
- Happy-path regression: a scanned batch still records `scanned is True` and
  still ingests.
- `consent.py` units: the factory threads `scanned=False`; `PendingBatch` raises
  `TypeError` when `scanned` is omitted (proving the no-default tripwire is
  real); `scanned=False` survives `with_resolved_types` / `with_state` /
  `with_ingested`.

The three tests that *pinned the hole* now assert the new behaviour by **exact
equality** to `_SCAN_BLOCKED_REPLY` plus an explicit
`assert "Reply with \`ingest\`" not in closing`. A substring probe for `"ingest"`
would be **vacuous** — the word appears inside the refusal copy ("I won't ingest
them"), so the old assertions would have passed against either string.

**Gate 2 — CrawDad's own toolchain** (separate venv, script and CI job from
creek-tools):

```
cd crawdad && ./scripts/check-all.sh     # exit 0
659 passed · branch coverage 94.60% (≥90)
xenon --max-absolute B --max-modules B --max-average A   # clean
mypy --strict · ruff lint+format (verified --no-cache) · bandit · pip-audit
```

No `# noqa`, no `# type: ignore`, no `@pytest.mark.skip`, no `--no-verify`, no
threshold touched. (The new tests also omit the `# type: ignore[arg-type]`
comment the surrounding file carries on its stub-client call sites — the gate
runs `mypy crawdad`, never `tests/`, so those 78 legacy comments are inert and
this change declines to add more.)

Closes #1054

### DIFF
--- a/crawdad/CLAUDE.md
+++ b/crawdad/CLAUDE.md
@@ -192,6 +192,60 @@ tier end-to-end from the capture record through staging into fragment
 frontmatter, so an `intimate` channel can be captured faithfully
 instead of refused outright.
 
+### 5.3 Redaction-scan gate (FEAT-027, #1054)
+
+**Policy: record the batch, refuse at dispatch.** An attachment turn
+whose `creek.redact.scan` could not run still stages the files and
+still records a `PendingBatch` — so the user can see what landed and
+`cancel` it — but that batch can never be dispatched to `creek.ingest`.
+
+Rejected alternatives: *withholding the consent prompt* (relies on the
+user never typing `ingest` unprompted — prompt suppression is not
+enforcement), and *a distinct opt-in token* (a documented override is
+a fail-open path, and this is the one gate standing between unscanned
+content and the vault).
+
+**Mechanism.** `PendingBatch.scanned` is a **required field with no
+default**, placed ahead of the defaulted fields. There is deliberately
+no permissive default and no `ConsentConfig` knob: either would let a
+construction site — or an operator — silently reopen the hole.
+`_run_safety_scan` returns a `_ScanOutcome(scanned, text)` rather than
+a bare string, so the caller reads a status instead of string-matching
+the reply text against `_MCP_UNAVAILABLE_REPLY` (shared with the ingest
+and state paths). All four failure modes — no MCP client, the tool
+unadvertised, `MCPUnavailableError`, and any other exception — yield
+`scanned=False`.
+
+Enforcement is **one branch in `_dispatch_ingest_for_batch`**, the sole
+convergence point of the consent flow. Both `_apply_consent_reply` and
+`_apply_disambiguation_reply` reach ingest through it, so one guard
+closes the `ingest` route, the `ingest` → type-question → type-word
+route, and any future caller. Do not add a second guard at a caller: two
+gates drift, one cannot. `with_resolved_types` / `with_state` /
+`with_ingested` are all `dataclasses.replace` copies, so `scanned`
+survives every transition and cannot be laundered.
+
+**Explicit non-goal.** The gate enforces *"the scan could not run"* —
+**not** *"the scan found nothing"*. `creek_mcp/tools/redact.py` computes
+`status = "empty" if not scanned.matches else "ok"`, so there is no
+failing status to read: a scan that runs and reports secrets still
+clears the gate. User-facing copy must claim only what is enforced —
+`_SCAN_BLOCKED_REPLY` says "I couldn't run the redaction scan", never
+"these files are clean". Substituting a new false claim for the old one
+would miss the point of the fix.
+
+Related, explicitly **not** covered here: #1087 (`scan_batch` follows
+symlinks out of the scan root — a scan that *runs* but under-reports,
+and this gate marks such a batch `scanned=True`). #1088
+(`attachments.staging_subpath` outside the scan's canonical scope) *is*
+a fifth way the scan fails to run; this design absorbs it with one more
+`_ScanOutcome(scanned=False, …)` arm, so sequence it next rather than
+touching `_run_safety_scan` twice.
+
+**Revisit predicate:** when `creek.redact.scan` gains a failing status
+for non-empty findings, revisit whether `scanned` should become a
+three-state outcome and whether findings should block ingest.
+
 ## 6. MCP subprocess resilience
 
 The bot does **not** exit on MCP subprocess failure. The pattern is:
--- a/crawdad/crawdad/bot.py
+++ b/crawdad/crawdad/bot.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import json
 import logging
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Protocol, cast
 
 import discord
@@ -83,10 +84,34 @@ _REDACT_TOOL_MISSING_REPLY = (
 # via the per-channel :class:`PendingBatchStore` threaded through this
 # module. The CLI escape hatch in the prompt below is preserved for
 # users who prefer the deterministic terminal path.
+#
+# #1054: this prompt is now sent ONLY when the redaction scan actually
+# ran. When it could not, the turn closes with _SCAN_BLOCKED_REPLY
+# instead, and ``PendingBatch.scanned=False`` makes the refusal stick
+# at the dispatch boundary — so the bot can no longer tell the user to
+# "run `creek redact --scan` before ingesting" and offer `ingest` in
+# the very next message.
 _INGEST_CONSENT_PROMPT = (
     "I did **not** ingest anything. Reply with `ingest` (or run "
     "`creek ingest --type <type> --input <staging path>`) to proceed. "
     "Reply `cancel` to drop the batch."
+)
+
+# #1054: the refusal for a batch whose redaction scan never ran. ONE
+# constant deliberately serves TWO call sites — the close of an
+# unscanned attachment turn (:func:`_handle_attachments`) and the
+# refusal at the dispatch boundary
+# (:func:`_dispatch_ingest_for_batch`) — so the two messages are
+# literally the same string and cannot drift into contradicting each
+# other the way _REDACT_TOOL_MISSING_REPLY and _INGEST_CONSENT_PROMPT
+# did. The wording claims only the invariant actually enforced ("could
+# not scan"); it must NOT be reworded to imply a successful scan blocks
+# findings, because ``creek.redact.scan`` returns a permissive status
+# even when it finds secrets. See crawdad/CLAUDE.md §5.3.
+_SCAN_BLOCKED_REPLY = (
+    "I couldn't run the redaction scan on these files, so I won't ingest "
+    "them. Run `creek redact --scan <staging path>` from the terminal to "
+    "check them yourself, or reply `cancel` to drop the batch."
 )
 
 _ALREADY_STAGED_REPLY = (
@@ -322,6 +347,25 @@ def _empty_skills() -> VoiceSkillStack:
     return VoiceSkillStack(skills=())
 
 
+@dataclass(frozen=True, slots=True)
+class _ScanOutcome:
+    """What :func:`_run_safety_scan` did, and what to tell the user (#1054).
+
+    Attributes:
+        scanned: ``True`` only when ``creek.redact.scan`` actually
+            executed and returned a body. ``False`` for every failure
+            mode, and it is the value recorded on
+            :attr:`crawdad.consent.PendingBatch.scanned`. Deliberately
+            narrow: ``True`` says the scan *ran*, not that it came back
+            clean — the tool reports findings with a permissive status.
+        text: Discord-safe section to show the user — the formatted scan
+            report on success, a soft error otherwise.
+    """
+
+    scanned: bool
+    text: str
+
+
 async def _handle_attachments(
     *,
     message: _MessageLike,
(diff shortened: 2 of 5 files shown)


Geoffe-Ga:
## Summary
This closes #1054 by making the FEAT-027 redaction-scan gate *enforced* rather than *advisory*: an attachment batch whose `creek.redact.scan` never ran (any of four failure modes) is still recorded (inspectable, cancelable) but can never reach `creek.ingest`. The mechanism is a required, no-default `PendingBatch.scanned` bool checked by a single guard at `_dispatch_ingest_for_batch`, the sole convergence point for both the consent route and the type-disambiguation route.

## Strengths
- **Fail-closed by construction, not convention.** `scanned: bool` has no default and is placed ahead of the defaulted fields, so a `TypeError` at every omitted construction site prevents silent reintroduction of the hole — and this is proven with `test_pending_batch_requires_an_explicit_scanned_flag`.
- **Single convergence point.** The guard lives in `_dispatch_ingest_for_batch`, not in each caller, closing both the `ingest` route and the `ingest → type question → type word` route (`_apply_disambiguation_reply`) that the original issue text didn't name — this second route is explicitly tested (`test_unscanned_batch_cannot_ingest_via_type_disambiguation_route`).
- **One string, two call sites.** `_SCAN_BLOCKED_REPLY` replaces the old self-contradicting pair (tell the user to scan first, then still offer `ingest` in the next message) with one string that can't drift.
- **Copy claims only what's enforced.** The reply says "I couldn't run the redaction scan," not "these files are clean" — correctly reflecting that `creek.redact.scan` returns a permissive status even on findings. Related known gaps (#1087 symlinks, #1088 staging scope) are explicitly named as non-goals with a sequencing note rather than silently left unaddressed.
- **Test rigor.** All four scan-failure modes are tested individually, plus the disambiguation route, plus a happy-path regression, plus dataclass-replace preservation across `with_resolved_types`/`with_state`/`with_ingested`. The PR body documents a RED-phase proof and a "delete the guard, watch exactly 5 tests fail" mutation check.
- **Documentation.** `crawdad/CLAUDE.md` §5.3 records the policy, rejected alternatives, mechanism, explicit non-goal, and a revisit predicate — this is exactly the kind of decision record that prevents the fix from being "refactored" back into a hole later.
- **No persistence concern.** `PendingBatchStore` is confirmed in-memory only ("rebuilt fresh on restart"), so the new required field can't break deserialization of stale on-disk state.

## Security Concerns
None outstanding. 🟢 This PR *fixes* a fail-open security gap (unscanned content could reach `creek.ingest`) rather than introducing one. The fix is fail-closed: any ambiguity about whether the scan ran resolves to `scanned=False`.

## Problems
None found that block merge.

## Code Quality
- The docstrings/comments are notably dense (multi-paragraph blocks referencing #1054, §5.3, and future work throughout `bot.py`/`consent.py`). This is defensible given the 95% interrogate docstring-coverage gate and the genuine subtlety of the invariant, but it's worth a light pass later if it starts to feel like narration rather than reference material.
- The acknowledged UX wart (unscanned batch + unresolved type costs one extra turn before refusal) is explicitly called out as a deliberate, documented tradeoff rather than an oversight — reasonable as-is, and correctly deferred as a follow-up rather than bundled into this fix.

## CLAUDE.md Adherence
✅ TDD workflow followed (RED proof documented before source changes). ✅ `crawdad/CLAUDE.md` updated in the same PR as the behavior it documents. ✅ Commit message follows Conventional Commits (`fix(crawdad): …`) and closes the issue. ✅ CrawDad's own gate (`./scripts/check-all.sh`: mypy strict, ruff, bandit, pip-audit, xenon complexity) reported clean in the PR body, and CI confirms `CrawDad Quality` green on 3.11/3.12/3.13. ✅ No `# noqa`, `# type: ignore`, skip markers, or threshold changes introduced.

## Testing
Strong. Five new gate tests cover all four scan-failure modes plus the disambiguation-route bypass attempt, each asserting zero dispatched calls via exact equality (not a vacuous substring check — the PR correctly notes "ingest" appears inside the refusal copy itself, so a naive substring assertion would have passed against the old broken behavior too). `consent.py` gets direct unit coverage of the required-field invariant and its survival across all three `dataclasses.replace` transitions. A happy-path regression test (`test_scanned_batch_still_ingests_after_the_gate_lands`) guards against the fix bricking the normal flow. CI shows `Integration & E2E Tests` and `CrawDad Quality` (all three Python versions) green.

## Documentation
Complete. `crawdad/CLAUDE.md` §5.3 covers policy, rejected alternatives, mechanism, explicit non-goal, and revisit predicate. Docstrings on every touched function (`_handle_attachments`, `_run_safety_scan`, `_dispatch_ingest_for_batch`, `_apply_consent_reply`, `_apply_disambiguation_reply`, `PendingBatch`, `build_pending_batch`) were updated to match the new behavior, not just left stale.



## Verdict: LGTM

This is a well-scoped, well-tested security fix that closes a genuine fail-open gap, with a required-no-default field forcing every construction site to be explicit, a single convergence-point guard, thorough test coverage of all failure modes (including a bypass route the issue text missed), and matching documentation updates — CI is green on every crawdad-relevant job (CrawDad Quality ×3, Integration & E2E, Pylint, Complexity). No blocking issues and nothing substantive enough to withhold approval for.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234502206 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1284, opened 2026-08-09
Voice-skill tree: thread and eddy skills are ungated, and their member-derived titles reach skill_paths at ceiling=open


Geoffe-Ga:
## Problem

#971 threaded the caller's privacy tier ceiling into the voice-skill tree's
**fragment** walk. It did not — and could not, without a product decision —
gate the **thread and eddy** walk beside it.

`creek/generate/skills.py::_load_vault_snapshot` collects three things. One of
them is now gated:

```python
fragments = _collect_fragments(
    vault_path / _FRAGMENTS_SUBDIR,
    allow_intimate=allow_intimate,
    override=override,          # <- #971 added this
)
threads = _collect_typed(
    vault_path / _THREADS_SUBDIR, expected_type="thread", model_cls=Thread
)                               # <- no override
eddies = _collect_typed(
    vault_path / _EDDIES_SUBDIR, expected_type="eddy", model_cls=Eddy
)                               # <- no override
```

`Thread` and `Eddy` carry **no `privacy_tier` field** (`creek/models.py`), so
there is no tier on the note to rank. But their titles are **derived from their
member fragments** — `creek/link/threads.py:647`
(`title=self._generate_title(frags, context)`) and `creek/link/eddies.py:666`,
both via `creek/link/naming.py::cluster_title`. An eddy whose members are
`personal` or `intimate` produces a title carrying that content's vocabulary.

`_generate_thread_skills` / `_generate_eddy_skills` then render that title, the
id, the description, the member-thread list and the fragment count into
`creek-skills/threads/<slug>.SKILL.md` and `creek-skills/eddies/<slug>.SKILL.md`
— and the **filename slug is the slugified title**
(`creek/generate/skills.py:1366`, `:1398`).

## Why this is also a response-side leak, not only write-side

`creek_mcp/tools/skills.py` returns the written paths to the caller:

```python
relative = [str(p.relative_to(vault_path)) for p in written]
...
"skill_paths": relative,
```

For the `threads/` and `eddies/` entries those filenames **are** the slugified
titles. So at `privacy_tier_ceiling=open` an MCP caller receives, in the
response body, slugs derived from above-ceiling content. `skill_count` also
varies with how many threads and eddies clear their member thresholds, which is
an above-ceiling cardinality oracle.

This distinguishes it from #971, whose gap was purely write-side (the response
carried only paths under `frequencies/`, `phases/`, `modes/`, `registers/`,
[...]

Issue #1281, opened 2026-08-09
An elevated `creek skills generate` audits the elevation but not its scope (fragment_ids=[])


Geoffe-Ga:
## Problem

`creek skills generate --include-tier personal|intimate|all` writes a privacy
audit entry, and that entry records the elevation but not what the elevation
reached.

`creek/cli.py::skills_generate` calls:

```python
_audit_privacy_override_if_needed(
    vault_path=_resolve_vault(vault),
    command="skills",
    override=override,
    fragment_ids=[],
)
```

`fragment_ids=[]` is passed literally. So `00-Creek-Meta/audit/privacy.jsonl`
records *that* an operator raised the ceiling for a skills run, and records a
fragment count of zero for a run that may have admitted the entire personal
corpus into the voice-skill tree.

Two consequences:

1. **The audit trail cannot answer the question it exists to answer.** After the
   fact, "which personal fragments were exposed to the skill tree on
   2026-08-09?" is unanswerable from the log. Compare `creek mine` / `creek
   draft`, which pass the ids they actually filtered.
2. **It is worse after #971 than before.** Before, the flag was inert — the
   audit entry recorded an elevation that changed nothing, so an empty id list
   was arguably honest. #971 made the flag real: the elevation now genuinely
   admits above-ceiling bodies, and the log still says nothing about which.

Deliberately out of scope for #971, which fixed the leak and did not touch the
audit call.

## Evidence

- `creek-tools/creek/cli.py`, `skills_generate` — the `fragment_ids=[]`
  argument, unchanged by #971.
- `creek-tools/creek/classify/privacy_filter.py::record_privacy_override` —
  the audit writer, which accepts and persists a fragment-id list.
- `creek-tools/creek/templates/skills/privacy-tier.SKILL.md`, flow-level rule 2:
  "Any call that uses `--include-tier personal`, `--include-tier intimate`, or
  `--include-tier all` writes an entry to `00-Creek-Meta/audit/privacy.jsonl`
  recording the verb, vault, override, **fragment count**, and timestamp." The
  shipped contract promises a count; the implementation writes zero.

## Scope

Make the skills audit entry record the scope of the elevation.

The ordering constraint is the whole difficulty and should be stated in the
design rather than discovered: the audit call happens **before** the generator
runs (before the `--generate` short-circuit, even), so at the moment the entry
is written nobody has walked the vault yet. Options, none decided here:

- Move the audit call below the generation and record the real admitted set —
[...]

Issue #1280, opened 2026-08-09
Skill trees generated before the #971 ceiling gate still hold personal bodies; nothing tells the operator


Geoffe-Ga:
## Problem

#971 closed the leak: `creek.skills.refresh` and `creek skills generate` now
honour the caller's privacy tier ceiling, so `personal` and `unclassified`
bodies, titles and ids no longer reach `<vault>/creek-skills/`.

It does **not** clean up after itself. Every `creek-skills/` tree that already
exists in a user vault was generated under the leaky rule and still contains
those bodies verbatim. The fix is forward-only: the next `creek skills generate
--generate` (or `creek.skills.refresh`) rewrites the tree correctly, but until
somebody runs it, the exposure sits on disk with nothing telling the operator it
is there.

This matters more than a normal stale-artifact problem because the skill tree is
what agents *read*. `<vault>/creek-skills/**.SKILL.md` is deployed as agent
context; a stale tree keeps feeding above-ceiling prose to every consumer of the
voice-skill surface, including CrawDad.

## Evidence

Executed at HEAD `dfe9495`, before the #971 fix, on a four-fragment scratch
vault (`open` / `personal` / `unclassified` / `intimate`, self-authored,
F3/rising/express-do/confessional, ~40-word bodies):

```
skill_count: 34
  CANARY-OPEN-971              present=True
  CANARY-PERSONAL-971          present=True
  CANARY-UNCLASSIFIED-971      present=True
  CANARY-INTIMATE-971          present=False
  Personal title               present=True
  Unclassified title           present=True
files leaking personal: ['creek-skills/frequencies/F3.SKILL.md',
                         'creek-skills/modes/express-do.SKILL.md',
                         'creek-skills/phases/rising.SKILL.md',
                         'creek-skills/registers/confessional.SKILL.md']
```

Those four files are exactly what remains on disk in any vault whose tree
predates the fix. Note also that an untracked `crawdad/creek-skills/` exists in
at least one working tree in this repo.

## Scope

Detect a skill tree that was generated before the ceiling gate existed, and
either warn or offer to regenerate. Explicitly **out of scope** for #971, which
was a behavioural fix to the generator and deliberately did not grow a migration
path.

Open design questions, not decided here:

- **How is "stale" detected?** The skill files carry frontmatter. A generator
  version / schema stamp written at generation time is the obvious hook, but no
  such stamp exists today, so the first tree written *after* this issue is the
[...]

PR #1286, opened 2026-08-10
fix(skills): honour the tier ceiling in the voice-skill tree (#971)

### LINKED ISSUE #971: creek.skills.refresh never threads privacy_tier_ceiling: personal bodies pass unsummarised at ceiling=open

Geoffe-Ga:
## Summary

- **The leak.** `creek.skills.refresh` accepted a `privacy_tier_ceiling`, wrote it to the audit log, echoed it back in the response — and never handed it to `SkillTreeGenerator`. `creek skills generate --include-tier` was worse than "partially wired": it parsed the flag, audited it, and constructed the generator with neither `override` nor `allow_intimate`, so the flag was inert **in both directions**. The consequence was a `personal` fragment's **full body, title and id copied verbatim** into the untiered `<vault>/creek-skills/` tree — the artifact deployed as agent-readable context — for a caller who declared `ceiling=open`. `unclassified` fragments leaked identically, because `_is_snapshot_fragment` branched on `PrivacyTier.INTIMATE` alone while the shared filter ranks untiered content with `personal` (#876); on a vault that has never been through `creek classify` that is the *entire* corpus.
- **The fix.** `creek/generate/skills.py::_collect_fragments` — the verified sole producer of `VaultSnapshot.fragments` — now applies **two independent, AND-ed gates**: the unchanged `_is_snapshot_fragment` (self-authorship + the `allow_intimate` consent opt-in) and a new `tier_within_override(tier_of(fragment), override)` ceiling gate. `skills_refresh_tool` converts the ceiling with `to_privacy_override`; `creek skills generate` threads the `--include-tier` override it was already parsing. `creek_mcp/read_gate.py`'s posture manifest moves from `UNGATED_KNOWN_GAP`/`gap_issue=971` to `GATED`, re-pointed at the gate that closed it.
- **Omit, don't summarise — this is the load-bearing design decision.** The ceiling gate is a **hard rank cutoff**, deliberately *not* `filter_fragments_by_tier`. That filter replaces a personal body with `"[Personal-tier summary: <title>]"`, and a skill file **is** a voice-exemplar corpus: the stub would be written into `## Exemplar Passages` beneath `> **<title>** (`<id>`)` in bold, leaking the very title it claims to protect and teaching the model a sentence nobody wrote. Nor is the harvester's own word floor any protection — the stub is two tokens plus the title, so a 28-word H1 reaches exactly `_EXEMPLAR_WORDS_MIN = 30` and `_extract_passage` accepts at `>=`, quoting it whole. This is the same call #968 made for `creek.report`, one step further in; `test_a_long_titled_personal_fragment_leaks_no_title_at_the_open_ceiling` is the test a summariser-based fix fails.

## What the fix does NOT cover — read the invariant narrowly

The gate is on the **fragment** walk. `_load_vault_snapshot` collects three things and only one of them grew an override: `Thread` and `Eddy` carry no `privacy_tier` field, so `_collect_typed` takes none, and **thread and eddy skills are not gated**. Their titles, ids, descriptions and fragment counts are derived from their member fragments (`creek/link/threads.py:647`, `creek/link/eddies.py:666`), so an eddy built from above-ceiling members still reaches `creek-skills/eddies/` — and, slugified into the filename, reaches the returned `skill_paths` too, which makes that half a *response*-side leak rather than a purely write-side one.

Gating `_collect_typed` means deciding what tier a thread or an eddy has, and the obvious answer (fail closed on the keyless note, the #968 shape) drops two whole categories from the tree at `ceiling=open`. That is a product decision, not a mechanical extension, so it is filed as **#1284** rather than folded in here. What this PR does instead is refuse to overclaim: the invariant is stated as "no above-ceiling **fragment** body, title or id", and the uncovered surface is disclosed by name in all four places that describe the posture — `creek_mcp/read_gate.py`'s rationale, `skills_refresh_tool`'s docstring, `docs/mcp.md`'s table row, and the new test module's docstring.

This matters because an overclaiming docstring is precisely what produced #971: the original text said the operation was "safe at any ceiling". A first review pass returned CLEAN on a draft that had replaced one overclaim with a narrower one; the parallel security review caught it, and the claims were narrowed before push.

## The issue body was partly wrong; three of its claims did not survive verification

Recorded here because the corrections changed the work, not just the prose.

1. **The quoted evidence is stale.** The issue opens on a docstring reading "The generator already excludes `intimate` exemplars so the operation is safe at any ceiling." That string does not exist anywhere in the tree — `f72dfa7` (#973) already replaced it with an honest description of the gap. So "Suggested shape" item 3 was already done, and what this PR needed was the *opposite* edit: the gap-confessing docstring (ending `Tracked by #971.`) had to be rewritten to the gated posture, or `test_gap_issues_are_named_in_their_own_tool_module` inverts and demands the module keep naming an open gap.
2. **The issue under-scoped the leak.** It names only `personal`. `unclassified` leaks too, and on a freshly-ingested vault that is the larger half of the hole. Routing through `tier_within_override` covers it for free (`_TIER_RANK[UNCLASSIFIED] == _TIER_RANK[PERSONAL] == 1`); a fix that special-cased `PERSONAL` literally would have left it open.
3. **The CLI was worse than described.** "Thread `--include-tier` from the CLI" implies the flag did something. It did not — `creek skills generate --include-tier intimate` emitted zero intimate exemplars *and* zero ceiling enforcement. Confirmed by execution at HEAD.

## Privacy delta — strictly narrower or identical, at every ceiling, on every surface

| Surface | before | after |
|---|---|---|
| MCP `ceiling=open` | open + **personal** + **unclassified** | open only |
| MCP `ceiling=personal` / `intimate` / `all` | open + personal + unclassified (intimate held out by the consent gate) | identical |
| CLI default / `--include-tier open` | open + **personal** + **unclassified** | open only |
| CLI `--include-tier personal` / `intimate` / `all` | identical | identical |

`allow_intimate` is deliberately **not** derived from `--include-tier`. `creek/templates/skills/privacy-tier.SKILL.md` rule 1 is a shipped promise — "Even with `--include-tier intimate`, intimate fragments contribute to the voice-skill tree only when the human has explicitly opted in (`SkillTreeGenerator(allow_intimate=True)` and consent on file). The tier flag alone is not enough." Wiring consent to the flag would newly emit intimate bodies where zero are emitted today: a strict loosening of a published privacy contract, which the escalate-only rule disqualifies. Consent stays a Python-API-only opt-in, and the MCP tool never passes it, so intimate exemplars are unreachable over MCP at **every** ceiling, `all` included. One tightening beyond the issue: `SkillTreeGenerator(allow_intimate=True)` with no `override` now excludes intimate at the open ceiling, pinned by `test_intimate_excluded_at_the_open_ceiling_even_with_consent`.

## Two things this PR deliberately grew beyond the issue

- **The empty-parametrisation repair.** `creek.skills.refresh` was the *last* ungated gap in the manifest, so closing it empties `_PINNED_GAPS`, `_UNGATED_GAP_TOOLS` and `_GAP_ISSUE_TOOLS` simultaneously. `pyproject.toml` declares no `empty_parameter_set_mark`, so pytest's default of `skip` would have silently deleted four privacy tests behind a green gate — the assertions still in the file, still read as protection, never executed again. Those four are converted from `@pytest.mark.parametrize` to bare loops with every assertion and message kept verbatim, and `test_the_manifest_records_no_ungated_gaps` now asserts the empty state so the layer keeps teeth and the next gap re-arms the loops.
- **Operator feedback.** After this fix, a vault that has never been through `creek classify` generates a complete 34-file tree in which *every* `## Exemplar Passages` section is the "no qualifying exemplars" placeholder, at the default ceiling, silently. That is the shape of a fix that gets reverted as a bug. The CLI now prints a static remedy line at the default ceiling, and `_collect_fragments` logs, once per run, how many fragments the ceiling withheld and how to widen it. Neither discloses a title, an id, a body or a path — only a count and the remedy. Worth being precise about which of those two actually reaches a human: nothing under `creek/` or `creek_mcp/` configures a logging handler, so the `logger.info` record is dropped by the unconfigured root logger in a real run. Its docstring now says so outright rather than claiming an operator experience that does not exist — the CLI hint is the operator-facing half; the log line is for an embedder that has configured logging, and for the tests. Wiring a handler would be a repo-wide logging policy, and on the MCP stdio transport a stdout handler would corrupt the protocol stream. The count is deliberately **not** carried into the MCP response — a tally of above-ceiling fragments crossing that boundary is a disclosure the tool has no reason to make. Response shape unchanged (`status`, `tool`, `tier_ceiling`, `skill_count`, `skill_paths`); no `CONTRACT_VERSION` bump.

## Test plan

**RED first, executed against unmodified HEAD `dfe9495`** — a four-fragment scratch vault (self-authored, F3/rising/express-do/confessional, ~40-word bodies so `_extract_passage` really yields an exemplar):

```
skill_count: 34
  CANARY-OPEN-971              present=True
  CANARY-PERSONAL-971          present=True     <-- LEAK
  CANARY-UNCLASSIFIED-971      present=True     <-- LEAK
  CANARY-INTIMATE-971          present=False
  Personal title               present=True     <-- LEAK
  Unclassified title           present=True     <-- LEAK
files leaking personal: creek-skills/frequencies/F3.SKILL.md
                        creek-skills/modes/express-do.SKILL.md
                        creek-skills/phases/rising.SKILL.md
                        creek-skills/registers/confessional.SKILL.md
```

Reproduced through **both** entry points before any source change — `skills_refresh_tool(..., privacy_tier_ceiling=TierCeiling.OPEN)` and `creek skills generate --generate --vault V` — because a fix threaded through only one leaves the other open. The CLI run was repeated with `--include-tier intimate`, which changed nothing in either direction, confirming the flag was inert. Then `pytest tests/test_skills_tier_ceiling.py` against unmodified source: **10 failed, 5 passed**, the failures being the personal/unclassified leak at `open`, the long-titled-personal title leak, the personal-only well-formedness case, the four `(allow_intimate, override)` combinations, the withheld-count log, and both CLI cases. The 5 that passed are the recoverability and no-loosening cases, green at HEAD and required to stay green.

New file `creek-tools/tests/test_skills_tier_ceiling.py` (12 tests + 4 parametrisations), covering: the four-tier leak assertion with an `open` **positive control** so no drop-everything implementation can pass vacuously; recoverability executed at `ceiling=personal` **and** `ceiling=all` (the personal canary asserted *present*, not merely that the call succeeds); intimate parametrised over all four `(allow_intimate, override)` combinations, appearing only for consent **and** ceiling; the 28-word-title case; a personal-only vault still writing well-formed files; the keyless-fragment-at-`ceiling=personal` behaviour stated out loud (the deliberate model-side `tier_of` reader, which diverges from `raw_privacy_tier` for exactly that note); the withheld-count log; and four CLI cases including `--include-tier intimate` still emitting nothing.

`tests/test_skills.py` migrated to state the ceiling out loud rather than by flipping `_build_fragment`'s `privacy=PERSONAL` default — flipping it would have removed the tier dimension from the whole 104-test file, leaving the exact regression this PR fixes untested at unit level, and it collides with the helper's `proxy_eligible=False` guard. **No assertion was deleted, reworded or weakened in any test file**; every change was an added `override=` keyword or an added test.

**Gate 2, exit 0 on the first run and again after the review fixes:** `./scripts/check-all.sh` → 12 passed / 0 failed, 8681 unit tests, coverage 94.36% (≥90), per-file gate 255 files ≥80% + 2 waivered. Ruff re-verified separately with `--no-cache` (a stale cache has produced a false-clean Gate 2 in this repo before, #1119). Plus `./scripts/test.sh --integration` (356 passed) and `--e2e` (14 passed) — including `test_full_pipeline_idempotency`, which is the known flake #1114 and passed here. `pytest tests/test_mcp_read_gate.py -rs` → 143 passed, **zero SKIPPED**, confirming the empty-parametrisation repair held.

## Follow-ups filed, not scoped in

- **#1284 (P1)** — thread and eddy skills are ungated, and their member-derived titles reach both `creek-skills/eddies/` and the returned `skill_paths` at `ceiling=open`. Found by this PR's own security review; see the section above for why it is not fixed here. Note `creek_mcp/read_gate.py` already records "an eddy title derived from an above-ceiling member" as one of the three leaks **#969** had to close, so the repo has already ruled this a leak class.
- **#1280 (P2)** — every `creek-skills/` tree already deployed in a user vault was generated under the leaky rule and is **not** retroactively cleaned; regeneration is required and nothing currently tells the operator. (An untracked `crawdad/creek-skills/` exists in at least one working tree here.) Note also that `creek/templates/skills/privacy-tier.SKILL.md` is a *deployed template*, so its correction only reaches an existing vault on the next `creek skills sync`.
- **#1281 (P2)** — `_audit_privacy_override_if_needed` is called with `fragment_ids=[]` for skills, so an elevated run audits the elevation but not its scope. Latent before this PR (the flag was inert); real now that the elevation genuinely admits content.

## Notes for the reviewer

- Conceptual, not file-level, overlap with **#931** (`structural_path` ancestry leaking above-ceiling parent titles): same class of defect, and the two must not contradict each other on whether titles are free. After this fix the skill tree carries no above-ceiling *fragment* title — the hard cutoff drops the fragment before `_build_exemplar` sees it — and `test_a_long_titled_personal_fragment_leaks_no_title_at_the_open_ceiling` keeps that honest against a future tuning of `_EXEMPLAR_WORDS_MIN`. Derived thread/eddy titles are the part #931's argument still reaches, which is #1284.
- **#1036** rewrites read_gate layer (f)'s semantics in the same two files this PR touches. This PR only *adds* a `_RUNTIME_PROBES` row, a `_PINNED_GATE_ROWS` row and the empty-set repair, so it should land first and #1036 rebase onto it.
- Two review passes ran over this diff before push: a `code-review-orchestrator` sweep and an explicit adversarial security review. The security pass is the one that caught the overclaim; both are reflected above.

Closes #971
Refs #932


Geoffe-Ga:
## Summary
Closes #971: `creek.skills.refresh` (MCP) and `creek skills generate --include-tier` (CLI) both accepted a privacy-tier ceiling but never threaded it into `SkillTreeGenerator`, so `personal`/`unclassified` fragment bodies, titles, and ids were copied verbatim into the untiered `<vault>/creek-skills/` tree regardless of the caller's declared ceiling. The fix adds a hard rank-cutoff gate (`tier_within_override`) to `_collect_fragments`, ANDed with the pre-existing `allow_intimate` consent gate, and threads it through both production call sites, the read-gate posture manifest, and `docs/mcp.md`.

## Strengths
- The core fix is minimal and correct: `_collect_fragments` in `creek/generate/skills.py` now applies `tier_within_override(tier_of(fragment), override)` as a second, independent gate ANDed with `_is_snapshot_fragment`'s consent check — verified against `creek/classify/privacy_filter.py`'s `tier_within_override`/`tier_of` implementations, which behave exactly as documented.
- Deliberate design choice to use a hard cutoff rather than `filter_fragments_by_tier`'s summarizing stub is well-justified and independently defensible: a `"[Personal-tier summary: <title>]"` stub written into `## Exemplar Passages` would itself leak the title and (per the PR's own math) can clear `_EXEMPLAR_WORDS_MIN` and be quoted whole — `test_a_long_titled_personal_fragment_leaks_no_title_at_the_open_ceiling` demonstrates this concretely.
- Exceptionally thorough test coverage in the new `tests/test_skills_tier_ceiling.py`: covers both production surfaces (MCP and CLI), all four privacy tiers (including the `unclassified`-ranks-with-`personal` #876 edge case), the AND-not-OR interaction between `allow_intimate` and the ceiling (full 2×2 truth table), operator-facing feedback (log line + CLI hint), and an anti-vacuity discipline (every assertion is against written file bytes, not the content-free tool response, with explicit positive controls so a "drop everything" regression can't pass).
- Explicitly scopes and discloses what the fix does *not* cover (thread/eddy skill files are not ceiling-gated, since `Thread`/`Eddy` carry no `privacy_tier`), consistently across four locations (`read_gate.py` rationale, `skills_refresh_tool` docstring, `docs/mcp.md`, new test module docstring), and files it as a separate tracked issue (#1284) rather than silently overclaiming or scope-creeping the fix.
- `tests/test_mcp_read_gate.py` correctly converts the parametrized gap-tracking tests to loops before `_PINNED_GAPS` goes empty, with a documented rationale (`empty_parameter_set_mark` defaults to `skip`, which would silently disable the check) and adds `test_the_manifest_records_no_ungated_gaps` to keep the empty state itself asserted.

## Security Concerns
None outstanding — this PR *is* the fix for the identified privacy leak (personal/unclassified fragment content reaching an untiered, agent-readable artifact directory). 🟢 LOW: the disclosed, tracked residual gap (thread/eddy skill filenames/descriptions can still carry above-ceiling vocabulary) is called out honestly in-code and in docs and tracked as #1284, which is the correct way to handle an out-of-scope-for-this-PR finding rather than a blocking omission.

## Problems
None found that block merge.

## Code Quality
- Minor: the new `_SKILLS_DEFAULT_CEILING_HINT` in `creek/cli.py`'s `skills_generate` prints unconditionally whenever the ceiling is at its default (`not override_elevates(override)`), including when `--signature-only` is passed. In signature-only mode, `_maybe_render_exemplar_section`/`_maybe_pick_exemplars` never emit exemplars regardless of tier (see `creek/generate/skills.py:1480-1511`), so telling the operator "pass `--include-tier personal` to include them" is misleading advice in that combination — the flag would have no effect on a signature-only run. Not covered by any test (`test_cli_hints_at_the_default_ceiling` doesn't pass `--signature-only`). Worth a follow-up: gate the hint on `not signature_only` as well.

## CLAUDE.md Adherence
✅ Follows the repo's TDD/Stay-Green philosophy — tests were clearly written to reproduce the exact leak before the fix, with anti-vacuity assertions throughout. ✅ No forbidden shortcuts (`noqa`, `type: ignore`, threshold-lowering) observed in the diff. ✅ Docs (`docs/mcp.md`, `privacy-tier.SKILL.md`) updated in the same PR to match new behavior, consistent with the single-source-of-truth principle in `creek/classify/privacy_filter.py`. Cannot independently confirm `./scripts/check-all.sh` / coverage / complexity gates passed locally (CI status was not accessible in this review environment), but the diff shows no signs of quality-gate evasion.

## Testing
Very strong. `tests/test_skills_tier_ceiling.py` (new, ~1150 lines) exercises: the leak reproduction at `ceiling=open` (T1), ceiling widening at `personal`/`all` (T2), the independent-AND interaction of consent and ceiling gates via a full truth table (T3), why a hard cutoff is required over a summarizing filter (T4), that an all-excluded vault still produces a well-formed 34-file tree rather than an outage (T5), the deliberate `tier_of`-vs-`raw_privacy_tier` divergence for untiered fragments (T6), operator-facing log feedback (T7), and both CLI surfaces including the `--include-tier intimate` non-loosening guarantee (T8). `tests/test_skills.py` and `tests/test_mcp_read_gate.py` are updated consistently to keep existing coverage meaningful under the new default ceiling.

## Documentation
Comprehensive and internally consistent across `creek/generate/skills.py` module/function docstrings, `creek_mcp/tools/skills.py`, `creek_mcp/read_gate.py`'s posture manifest, `docs/mcp.md`'s tool table, and `creek/templates/skills/privacy-tier.SKILL.md`'s user-facing table — all four "the gap exists" mentions were correctly updated to "the gate exists," and cross-references to the newly-filed #1284 (thread/eddy gap) are consistent everywhere they appear.



## Verdict: LGTM

The core privacy fix is correct, minimal, and verified against the actual `tier_within_override`/`tier_of` implementations; both production surfaces (MCP and CLI) are covered with exceptionally thorough, anti-vacuity tests; documentation is updated consistently everywhere the old gap was mentioned; and the one residual gap (thread/eddy skills) is honestly disclosed and tracked separately rather than silently left out. The single finding (a CLI hint that's slightly misleading under `--signature-only`) is a cosmetic messaging nit with no privacy or correctness impact — not something to block merge on.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234697926 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary

This PR closes #971 by threading the caller's declared `privacy_tier_ceiling` / `--include-tier` override into `SkillTreeGenerator` as a hard rank-cutoff gate in `_collect_fragments`, so `personal`/`unclassified` fragments' bodies, titles and ids stop leaking verbatim into the untiered `<vault>/creek-skills/` tree at `ceiling=open`. It deliberately uses omission rather than the existing `filter_fragments_by_tier` summariser (correctly, since the summary stub would itself become a fabricated, title-leaking exemplar), keeps the `allow_intimate` consent gate fully independent (ANDed, not merged), fixes a latent `pytest` empty-parametrization footgun that would have silently skipped four privacy regression tests, and adds an extensive, well-targeted new test module (`test_skills_tier_ceiling.py`).

## Strengths

- The core fix (`creek/generate/skills.py::_collect_fragments`) is correct and minimal: `tier_within_override(tier_of(fragment), override)` is applied as an independent AND-ed gate alongside the unchanged `_is_snapshot_fragment`. I verified `tier_within_override`, `tier_of`, `_TIER_RANK`, and `_EXEMPLAR_WORDS_MIN`/`_extract_passage` directly in `creek/classify/privacy_filter.py` and `creek/generate/skills.py`, and the PR's technical claims about them (e.g. the 28-word-title-hits-exactly-30-tokens summariser failure mode, `UNCLASSIFIED` ranking with `PERSONAL`) all check out.
- Choosing a hard cutoff over `filter_fragments_by_tier` is the right call and is well-justified — a skill file is a voice-exemplar corpus, so a `"[Personal-tier summary: <title>]"` stub would itself be a fabricated, title-leaking exemplar. `test_a_long_titled_personal_fragment_leaks_no_title_at_the_open_ceiling` is a genuinely good regression test that specifically catches a "fixed but with the wrong mechanism" mistake.
- Both production entry points (`skills_refresh_tool` and `creek skills generate`) are fixed together — the issue explicitly called out that a fix threaded through only one surface would leave the other broken, and the tests exercise both.
- The `allow_intimate` vs. `override` independence is correctly preserved (verified `SkillTreeGenerator.__init__` and both call sites never pass `allow_intimate=True`), honoring the shipped `privacy-tier.SKILL.md` promise that the tier flag alone is insufficient for intimate content. The 4-row truth table test (`test_intimate_needs_both_consent_and_ceiling`) is exactly the right shape to catch a "gates merged instead of ANDed" regression.
- Scope discipline: thread/eddy skills are explicitly *not* gated in this PR (Thread/Eddy carry no `privacy_tier` field), and rather than silently overclaiming, the PR narrows every posture description (`read_gate.py`, `skills_refresh_tool` docstring, `docs/mcp.md`, new test module docstring) to say so and files a separate issue (#1284) for it. I confirmed `Thread`/`Eddy` models carry no `privacy_tier` field.
- The `pytest` empty-parametrization fix (`_PINNED_GAPS` emptying to `{}`, converting `@pytest.mark.parametrize` to bare loops, and adding `test_the_manifest_records_no_ungated_gaps`) is a genuinely sharp catch — `pyproject.toml` sets no `empty_parameter_set_mark`, so the default `skip` behavior would have silently deleted four privacy regression tests behind a green gate the moment this fix landed.
- Test quality is high throughout: canaries are tier-specific rather than shared (avoiding ambiguous failures), an `open`-tier positive control guards against a "drop everything" implementation passing vacuously, and assertions are made against file bytes rather than the tool's return envelope (correctly noting the envelope carries only paths/counts, never content).

## Security Concerns

None found. 🟢 The privacy-tier gating logic itself is sound and matches the described invariant exactly. The known residual gap (thread/eddy skills, #1284) is explicitly disclosed in four separate places rather than silently left as a hidden hole, which is the right way to hand off an intentionally-deferred security gap.

## Problems

None blocking.

## Code Quality

- `creek/cli.py::skills_generate` prints `_SKILLS_DEFAULT_CEILING_HINT` ("...pass `--include-tier personal` to include them") unconditionally whenever `not override_elevates(override)`, without checking `signature_only`. But `_maybe_render_exemplar_section` (creek/generate/skills.py:1511) omits the exemplar section entirely whenever `signature_only=True`, *regardless of tier ceiling* — so in `creek skills generate --signature-only` at the default ceiling, the hint tells the operator that widening `--include-tier` will surface exemplars, when in fact no exemplars will ever appear in signature-only output no matter the ceiling. This isn't a privacy issue (nothing leaks), just a misleading remedy message. No test in `test_skills_tier_ceiling.py` covers the `--signature-only` + default-ceiling-hint interaction, which is presumably why it wasn't caught. Worth a follow-up: guard the hint on `not signature_only`, or reword it.

## CLAUDE.md Adherence

✅ TDD is evident from the PR description's documented RED-first run and the new/expanded test files. ✅ Commit message follows Conventional Commits (`fix(skills): ...`) with a `Closes #971` reference. ✅ No shortcuts/suppressions introduced. I was unable to execute `./scripts/check-all.sh` or `pytest` directly in this review sandbox (commands required approval that wasn't available), so I verified correctness via direct source inspection instead of running the suite myself; the PR description reports a full green `check-all.sh` (94.36% coverage, 8681 unit tests, 0 skipped) plus integration/e2e passes, and CI's `Integration & E2E Tests` job has already completed successfully. Several required CI jobs (Tests & Type Checking 3.11-3.13, Pylint, Static Analysis) were still `IN_PROGRESS` at review time.

## Testing

Strong. The new `test_skills_tier_ceiling.py` module (1149 lines) covers: the four-tier leak/positive-control assertion via both entry points, recoverability at `ceiling=personal` and `ceiling=all`, the full `(allow_intimate, override)` truth table for intimate content, the summariser-vs-cutoff distinguishing test with a carefully chosen 28-word title, a personal-only vault's well-formedness, the withheld-count log line, and CLI-specific cases including a confirmation that `--include-tier intimate` does not smuggle in consent. `test_skills.py` was migrated to state the ceiling explicitly per-construction rather than changing the fixture default, correctly avoiding weakening ~104 existing tests. `test_mcp_read_gate.py`'s empty-parametrization repair is itself well-tested (`test_the_manifest_records_no_ungated_gaps`).

## Documentation

Excellent — arguably the standout aspect of this PR. `read_gate.py`'s `ToolPosture` rationale, `skills_refresh_tool`'s docstring, `docs/mcp.md`'s table row, and the new test module's docstring all consistently and precisely state the narrowed invariant ("no above-ceiling *fragment* body, title or id" — not thread/eddy skills), with cross-references to the follow-up issues (#1284, #1280, #1281) tracking what's deliberately out of scope. `_log_ceiling_withheld`'s docstring is unusually honest about a real limitation (the log line is dropped by an unconfigured root logger in production, so it's effectively test/embedder-only), rather than overclaiming operator-facing behavior.



## Verdict: COMMENTS

The privacy fix itself is correct, well-scoped, and thoroughly tested (verified directly against source for tier_within_override, tier_of, _EXEMPLAR_WORDS_MIN, and the Thread/Eddy model fields), so nothing blocks merge. It stays out of LGTM because of one concrete, non-blocking suggestion: the new CLI hint in `skills_generate` gives misleading remedy advice ("pass --include-tier personal to include them") when `--signature-only` is also set, since signature-only output never includes exemplars regardless of tier ceiling — worth a small follow-up guard or reword.

Geoffe-Ga:
**CI**: 10/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5234805852 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 14/14 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5234805852 to see in-depth feedback and continue iterating

Geoffe-Ga:
Confirmed and fixed in 3e20c05 — thanks, this was real.

**Verified the premise before touching it.** `_maybe_pick_exemplars` (`creek/generate/skills.py:1491`) returns `[]` and `_maybe_render_exemplar_section` (`:1511`) returns `[]` whenever `signature_only` is set, both *before* any tier is consulted. Tracing further than the finding did: `snapshot.fragments` is the only thing the override gates, and its sole consumers are the four `_group_fragments_by_*` calls that feed exemplar selection — `threads`/`eddies` come from `_collect_typed`, which takes no override. So under `--signature-only` the ceiling is not merely unhelpful, it is **inert**: it changes not one byte of the emitted tree.

**Suppressed rather than reworded.** Since no ceiling admits an exemplar in that mode, there is no true remedy to reword the message *into* — any wording naming `--include-tier` is noise for the signature-only operator and a caveat for the exemplar-bearing one, who is the majority path. And the confusion the hint exists to prevent (an exemplar-free tree reading as a broken command) can't arise for someone who explicitly asked for zero exemplars and gets `(34 signature-only files)` echoed back. The guard is now `if not signature_only and not override_elevates(override)`, with the reasoning recorded at the call site and in the constant's own docstring — the previous "printed unconditionally" claim in `skills.py`'s `_CEILING_WITHHELD_MESSAGE` docstring was corrected too, since the fix falsified it.

**New test** `test_cli_does_not_offer_the_ceiling_remedy_in_signature_only_mode` in `tests/test_skills_tier_ceiling.py`. It went red on the behavioural assertion (`'--include-tier personal' is contained here: ing; pass --include-tier personal to include them.`) and green after the guard. Three parts:

1. It asserts the inertness first — default-ceiling and `--include-tier personal` signature-only trees compare byte-equal (modulo the `generated_date` stamp). That is the premise the suppression rests on, so it is now itself under test: if the ceiling ever gains a real effect here, this goes red and asks for a hint describing whatever the ceiling does then, rather than leaving the operator silently uninformed.
2. Then the two hint phrases must be absent from `--signature-only` output.
3. It closes by re-running the plain exemplar-bearing command and asserting the hint is **still** printed, so the false positive can't be fixed by deleting the true one. `test_cli_hints_at_the_default_ceiling` still passes unchanged.

Gate 2 green locally: 12/12 checks, 8756 passed, coverage 94.38%. `tests/test_mcp_read_gate.py` still 143 passed / 0 skipped.

One adjacent thing I deliberately left alone rather than widening the diff: `_CEILING_WITHHELD_MESSAGE` carries the same "widen the ceiling" remedy and is also emitted in signature-only mode. It's a `logger.info` on an unconfigured logger, so no operator ever sees it — as its own docstring says at length — and it's pre-existing code, not something this PR introduced. Happy to file it if you'd rather it were consistent.

Geoffe-Ga:
## Summary
This PR closes #971: `creek.skills.refresh` (MCP) and `creek skills generate --include-tier` (CLI) both parsed/audited a `privacy_tier_ceiling`/`--include-tier` value but never threaded it into `SkillTreeGenerator`, so `personal` and untiered (`unclassified`) fragment bodies, titles, and ids were copied verbatim into the untiered `<vault>/creek-skills/` tree regardless of the declared ceiling. The fix adds a second, independent AND-ed gate — a hard rank cutoff (`tier_within_override`) alongside the existing `allow_intimate` consent gate — inside `_collect_fragments`, and threads it through both CLI and MCP entry points. It deliberately omits rather than summarizes (unlike the `filter_fragments_by_tier` sibling used by `mine`/`draft`/`author`), because a title-only stub written into `## Exemplar Passages` would itself be a fabricated, title-leaking "exemplar."

## Strengths
- Correct primitive choice: reusing `tier_within_override`/`tier_of` from the single source of truth (`privacy_filter.py`) rather than inventing a third tier-ranking scheme, and verified those primitives behave as documented (`UNCLASSIFIED` ranks with `PERSONAL`, `tier_of` fails closed to `INTIMATE` on an unrecognized string).
- Two independent gates (consent × ceiling) are genuinely ANDed, not merged, and this is pinned by dedicated tests (`test_intimate_excluded_at_the_open_ceiling_even_with_consent`, `test_intimate_excluded_without_consent_even_at_the_all_override`) that would fail if a future edit collapsed them into one.
- The "omit, don't summarize" design decision is correct and specifically pinned against the exact failure mode it avoids (a 28-word title clearing `_EXEMPLAR_WORDS_MIN` and getting quoted whole as a summary stub) — verified in code that `_EXEMPLAR_WORDS_MIN = 30` and `_extract_passage` accepts at `>=`.
- Explicitly scopes what is *not* fixed (thread/eddy skills, ungated because `Thread`/`Eddy` carry no `privacy_tier` field) and files a separate tracking issue (#1284) rather than silently overclaiming or scope-creeping the PR. Verified `_collect_typed` genuinely takes no `override` parameter, so this disclosure is accurate, not just asserted.
- The `_PINNED_GAPS`/`_UNGATED_GAP_TOOLS`/`_GAP_ISSUE_TOOLS` empty-parametrization repair is a real bug catch: converting `@pytest.mark.parametrize` to bare loops before the underlying dict emptied prevents pytest's default `skip`-on-empty-params from silently deleting four assertions behind a green gate.
- Both production surfaces (CLI and MCP) are fixed and both are covered by new runtime-probe/tree-content tests, not just response-envelope assertions — correctly noting that the response envelope (`skill_paths`, counts) was already leak-free, so envelope-only tests would have been vacuous.
- Verified in the diff that `to_privacy_override` and `tier_within_override`/`tier_of` map and rank exactly as the extensive docstrings claim; no discrepancy found between prose and code.

## Security Concerns
None blocking. The fix strictly narrows disclosure at every ceiling on every surface (per the PR's own before/after table, which I independently confirmed against the code). 🟢 LOW: the disclosed, unfixed gap (thread/eddy titles/filenames derived from above-ceiling member fragments, tracked as #1284) remains open post-merge — acceptable since it's pre-existing behavior, honestly disclosed in four places, and not worsened by this change.

## Problems
None found that block merge. Code changes are consistent with their extensive documentation, and spot-checks of the core privacy primitives, CLI wiring, and MCP wiring all confirm the behavior described.

## Code Quality
- The docstrings/comments in this PR are extraordinarily verbose (multi-paragraph docstrings and comment blocks throughout `skills.py`, `read_gate.py`, `tools/skills.py`) — well beyond CLAUDE.md's general "no comments unless WHY is non-obvious" guidance, and beyond typical PR norms. This appears to be house style for this repo (evident in pre-existing code like `privacy_filter.py`), so not flagging as blocking, but worth noting if the team wants to trim verbosity going forward.
- Minor: the CLI hint (`_SKILLS_DEFAULT_CEILING_HINT`) prints even when the tree has zero fragments at all (not just zero admitted ones) — a truly empty vault gets the same "widen the ceiling" hint as a fully-populated-but-gated one. Not incorrect, just slightly imprecise; not worth blocking on.

## CLAUDE.md Adherence
✅ Follows the repo's TDD workflow per the PR's documented RED-first reproduction against unmodified HEAD, then GREEN after the fix.
✅ No forbidden shortcuts (no `noqa`/`type: ignore` additions observed in the diff, no threshold changes).
✅ Uses the shared `privacy_filter` primitives rather than duplicating tier logic (DRY, §1.2).
✅ Test coverage and quality-gate numbers are reported in the PR description (94.36% coverage, 12/12 check-all passing) — could not independently re-run `./scripts/check-all.sh` in this review environment (no `uv`/venv available), so this is taken on the PR author's word plus my own static verification of logic correctness.

## Testing
Extensive and well-targeted. `tests/test_skills_tier_ceiling.py` (new, ~1278 lines) covers both surfaces, all four tier positions (open/personal/unclassified/intimate), both directions of ceiling change (narrowing and widening), the intimate-consent×ceiling AND relationship from both angles, the long-title-summarizer-stub trap, CLI hint text, and the withheld-count log line. `tests/test_mcp_read_gate.py` additions correctly avoid the "vacuous because the envelope carries no content" trap by asserting on the written tree bytes rather than the tool's return value. `tests/test_skills.py` migration to explicit `override=` keywords (rather than flipping the fixture's tier default) is the right call — preserves the tier dimension as a tested axis rather than erasing it. No assertions were weakened or deleted; only additions.

## Documentation
Thorough and consistent across all four places that describe the posture (`read_gate.py` rationale, `skills_refresh_tool` docstring, `docs/mcp.md` table row, and the new test module docstring) — verified these are non-contradictory and that the "thread/eddy skills not gated" caveat appears identically in each. `creek/templates/skills/privacy-tier.SKILL.md` table row updated to match the new (correct) "excluded, not summarized" behavior for personal-tier content in skill exemplars.



## Verdict: LGTM

The fix is correct, precisely scoped, and verified by direct code inspection (not just trusting the PR's own claims) — the core privacy primitives, CLI wiring, and MCP wiring all behave exactly as documented, tests target the actual leak (file bytes) rather than the vacuous response envelope, and the one known-uncovered gap (thread/eddy titles) is honestly disclosed in four places and filed as a separate follow-up rather than silently left out. No blocking issues found; only a stylistic note on comment verbosity that doesn't warrant blocking.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234881291 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR closes #971: `creek.skills.refresh` (MCP) and `creek skills generate --include-tier` (CLI) both parsed/audited a `privacy_tier_ceiling`/`--include-tier` override but never threaded it into `SkillTreeGenerator`, so `personal` and (via #876) `unclassified` fragments had their full body, title, and id copied verbatim into the untiered `<vault>/creek-skills/` voice-exemplar tree regardless of the caller's declared ceiling. The fix adds a hard rank-cutoff gate (`tier_within_override`, ANDed with the pre-existing `allow_intimate` consent gate) in `_collect_fragments`, threads it through both production entry points, updates the MCP read-gate manifest from `UNGATED_KNOWN_GAP` to `GATED`, and adds operator-facing feedback (a CLI hint + a library log line) so an exemplar-free tree doesn't read as a broken command.

## Strengths
- Correctly identifies and fixes the leak on **both** production surfaces (MCP tool and CLI), and covers the `unclassified` route (#876) that the original issue under-scoped — verified by reading `_TIER_RANK` and `tier_within_override` directly in `creek/classify/privacy_filter.py`, which match the PR's description exactly.
- Chooses a hard omission cutoff over the existing `filter_fragments_by_tier` summarizer, and backs that design decision with a concrete, verified failure mode: a 28-word personal title plus the `_summarize_personal` stub (2 tokens) reaches exactly `_EXEMPLAR_WORDS_MIN` (30), so `_extract_passage`'s `>=` comparison would have quoted a title-leaking synthetic sentence whole. `test_a_long_titled_personal_fragment_leaks_no_title_at_the_open_ceiling` pins exactly this.
- Every canary-based test asserts on the *written files*, not the tool's return envelope — correctly recognizing that `skills_refresh_tool`'s response (paths + counts) was already "clean" throughout the life of the bug and would pass a response-level assertion vacuously.
- Positive controls throughout (`_OPEN_CANARY` asserted present) prevent a drop-everything implementation from passing by writing an empty tree — a common testing gap this PR explicitly guards against.
- The empty-parametrization repair (`_PINNED_GAPS` emptying) is a genuinely sharp catch: converting `@pytest.mark.parametrize` over now-empty collections to bare loops avoids pytest's default `skip`-on-empty behavior silently deleting four privacy regression tests behind a green gate.
- Scope is disclosed honestly rather than overclaimed: thread/eddy skills remain ungated (no `privacy_tier` field to rank them by) and that gap is stated in four separate places (read_gate rationale, tool docstring, docs/mcp.md, new test module docstring) and filed as a separate issue (#1284) rather than silently left out or falsely claimed fixed.
- Spot-checked several claims against actual source (`tier_of`'s fail-closed-to-INTIMATE behavior, `_maybe_pick_exemplars`/`signature_only` short-circuiting before any tier is consulted, `override_elevates` semantics, the `_CEILING_WITHHELD_MESSAGE` `%s`/`%d` format matching its call site) — all check out.

## Security Concerns
None outstanding — this PR *is* the security fix. 🟢 The intimate-content consent gate is correctly kept independent from (ANDed with, not replaced by) the new ceiling gate, verified by the `test_intimate_needs_both_consent_and_ceiling` truth table over all four `(allow_intimate, override)` combinations. The MCP tool still never passes `allow_intimate=True`, so intimate exemplars remain unreachable over MCP at every ceiling including `all`, honoring the shipped `privacy-tier.SKILL.md` promise.

## Problems
None found that block merge. The one deliberately out-of-scope item — thread/eddy titles derived from above-ceiling members still reaching `creek-skills/eddies/` and the returned `skill_paths` — is disclosed transparently in the tool docstring, the read-gate manifest, and docs/mcp.md, and tracked as #1284 rather than silently shipped as a false "safe" claim.

## Code Quality
- The docstrings and inline comments throughout this diff are extremely dense (multi-paragraph rationale on nearly every new function/constant). This matches the existing style already present in `creek_mcp/read_gate.py` and `creek/classify/privacy_filter.py`, so it's consistent with house convention rather than a one-off, but it does make the diff long to review and the functions themselves harder to skim quickly. Not blocking.
- Keeping `_PINNED_GAPS` as an empty dict (rather than removing it) is a reasonable call given the well-argued empty-parametrization hazard, but it is worth double-checking in a future pass that `test_the_manifest_records_no_ungated_gaps` doesn't become the only thing anyone reads before assuming the table is dead code.

## CLAUDE.md Adherence
✅ TDD-first workflow is evident from the PR's own RED-then-GREEN test plan (10 failing / 5 passing against unmodified HEAD, then green after the fix). ✅ `./scripts/check-all.sh` reported clean per the PR description (94.36% coverage, ruff --no-cache re-verified, integration/e2e lanes green). ✅ Uses feature branch (`issue/971-...`), not a direct commit to main. Coverage, complexity, and docstring thresholds are addressed per the reported numbers, though I could not independently re-run `check-all.sh` in this review environment.

## Testing
Exceptionally thorough — a new 1278-line test module (`test_skills_tier_ceiling.py`) plus substantial updates to `test_mcp_read_gate.py` and `test_skills.py`. Coverage includes: the four-tier leak at `ceiling=open` with a positive control, recoverability at `personal` and `all` ceilings, the full `(allow_intimate, override)` truth table for intimate content, the long-title/summarizer-vs-cutoff distinguishing test, a personal-only vault still producing a well-formed 34-file tree, the deliberate `tier_of` vs `raw_privacy_tier` divergence for keyless fragments, the once-per-run withheld-count log (both firing and staying silent), and four CLI cases including the signature-only hint-suppression edge case. Every existing `test_skills.py` generator construction was updated to state its ceiling explicitly rather than relying on a now-more-restrictive default, with no assertions weakened or removed.

## Documentation
Updated consistently across all four places that describe the tool's read posture: `creek_mcp/read_gate.py`'s manifest rationale, `skills_refresh_tool`'s docstring, `docs/mcp.md`'s table row, and the CLI's `_SKILLS_DEFAULT_CEILING_HINT`/module comments. The `privacy-tier.SKILL.md` template (a promise shipped into user vaults) was also corrected. The invariant is stated narrowly and accurately ("no above-ceiling *fragment* body, title or id") rather than overclaiming — notably, the PR's own narrative explains that an earlier draft *did* overclaim ("safe at any ceiling") and was caught by a parallel security review before push, which is a good signal about the process behind this change.



## Verdict: LGTM

The fix is correct and verified against the actual source (tier ranking, hard-cutoff gate, consent/ceiling independence, CLI hint logic all check out), the test suite is unusually thorough with vacuity-guarding positive controls, no security or CLAUDE.md-adherence issues were found, and the one piece of scope deliberately left out (thread/eddy title leakage) is disclosed honestly in four places and tracked as a separate issue rather than silently shipped or overclaimed as fixed.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234960763 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR closes issue #971: `creek.skills.refresh` (MCP) and `creek skills generate --include-tier` (CLI) accepted a privacy-tier ceiling but never threaded it into `SkillTreeGenerator`, so `personal`- and `unclassified`-tier fragment bodies, titles, and ids were copied verbatim into the untiered `<vault>/creek-skills/` voice-exemplar tree regardless of the caller's declared ceiling. The fix adds a hard rank-cutoff gate (`tier_within_override`) to `_collect_fragments`, the verified sole producer of `VaultSnapshot.fragments`, ANDed with the pre-existing `allow_intimate` consent gate, and threads the caller's ceiling through both production entry points.

## Strengths
- **Correct fix shape.** Uses `tier_within_override` (hard omission) rather than `filter_fragments_by_tier` (title-preserving summarisation) — the PR correctly identifies that a skill file is a voice-exemplar corpus, so a `"[Personal-tier summary: <title>]"` stub would itself become a fabricated, title-leaking exemplar. Verified in code: `creek/generate/skills.py:696-699`.
- **Reuses existing primitives** (`tier_within_override`, `tier_of`, `to_privacy_override`, `override_elevates`) instead of reinventing filtering logic — consistent with the rest of the codebase's privacy machinery.
- **Non-vacuous tests.** New tests assert against the bytes of the written skill tree, not the tool's response envelope (which never carried content and would pass identically against unfixed code). Includes a mandatory `open`-tier positive control (so a drop-everything implementation can't pass), a full 4-row truth table for the `allow_intimate` × `override` AND relationship, and a 28-word-title case that specifically defeats a summariser-based fix.
- **Handles the `unclassified` case**, which the original issue under-scoped — verified `_TIER_RANK[UNCLASSIFIED] == _TIER_RANK[PERSONAL] == 1` in `creek/classify/privacy_filter.py`, so it's covered "for free" by the same gate.
- **Honest scoping.** Explicitly discloses that thread/eddy skills remain ungated (no `privacy_tier` field on those models) in four places (module docstring, tool docstring, `docs/mcp.md`, `read_gate.py` rationale) rather than overclaiming, and files it as a separate issue (#1284) since gating it is a product decision about which files the tree emits.
- **Caught and fixed the empty-parametrization footgun**: closing the last `UNGATED_KNOWN_GAP` empties `_PINNED_GAPS`/`_UNGATED_GAP_TOOLS`/`_GAP_ISSUE_TOOLS`, and since `pyproject.toml` sets no `empty_parameter_set_mark`, `pytest.mark.parametrize` over an empty list would silently skip (not pass) four tests behind a green gate. Correctly converted to bare loops plus an explicit `test_the_manifest_records_no_ungated_gaps` assertion.

## Security Concerns
None found at 🔴/🟡. This is itself a privacy-leak fix, verified by direct source inspection:
- `creek_mcp/tools/skills.py:69` — `override=to_privacy_override(privacy_tier_ceiling)` is threaded into the generator, matching the `_PINNED_GATE_ROWS` claim added in `test_mcp_read_gate.py`.
- `allow_intimate` is deliberately never derived from `--include-tier` (preserves the shipped `privacy-tier.SKILL.md` promise that intimate content needs explicit Python-API consent) — confirmed neither CLI nor MCP call site passes it.
- 🟢 LOW (informational, already tracked): thread/eddy skill titles/filenames remain derived from above-ceiling member fragments — filed as #1284, correctly scoped out.

## Problems
None blocking.

## Code Quality
- The docstrings are extremely dense (multi-paragraph, cross-referencing issue numbers throughout). This is a deliberate, previously-validated project convention for this security-sensitive module (per the repeated "same call #968 made" framing tying into prior review feedback), not a defect — but a future pass to extract some of this rationale into ADRs (per `documentation` skill conventions) rather than inline docstrings could improve long-term readability. Non-blocking.
- `SkillTreeGenerator.__init__`'s `self.override = override or PrivacyTierOverride.OPEN` relies on enum truthiness (all `StrEnum` members here are non-empty strings, hence truthy) — correct but implicit; verified against the identical pattern already used in `tier_within_override` (`effective = override or PrivacyTierOverride.OPEN`), so it's consistent with existing style rather than a new risk.

## CLAUDE.md Adherence
✅ Conventional commit messages (`fix(skills): ...`, `fix(cli): ...`).
✅ PR description documents `./scripts/check-all.sh` run (12 passed / 0 failed, 8681 unit tests, 94.36% coverage ≥90% threshold, 255 files ≥80% per-file gate), plus `--integration` and `--e2e` lanes, and a ruff re-run with `--no-cache` per the #1119 stale-cache lesson.
✅ TDD workflow documented (RED-first reproduction against unmodified HEAD before any source change, for both entry points).
⚠️ I was unable to independently execute `./scripts/check-all.sh` or the test suite in this review sandbox (Bash tool required interactive approval unavailable in this session), so I relied on static verification of the diff plus the PR's own detailed, cross-checkable test-execution log — the specific function signatures, imports, and call sites cited in that log were independently confirmed by reading the actual post-merge source.

## Testing
Strong. New file `tests/test_skills_tier_ceiling.py` (1278 lines, 12 tests + 4 parametrizations) covers both production entry points (MCP `skills_refresh_tool` and CLI `skills generate`), the leak at `ceiling=open`, recoverability at `ceiling=personal`/`ceiling=all`, the intimate consent×ceiling truth table, the long-title summariser-defeat case, a personal-only vault's well-formedness, the withheld-count log, and CLI hint suppression under `--signature-only`. `test_skills.py` migration was done via added `override=` keywords only — no assertion deleted or weakened, verified in diff. `test_mcp_read_gate.py` additions correctly convert now-empty parametrizations to loops rather than leaving them to silently skip.

## Documentation
✅ `docs/mcp.md`, `creek_mcp/read_gate.py` rationale, `skills_refresh_tool` docstring, and `privacy-tier.SKILL.md` (the shipped vault template) are all updated consistently and cross-reference each other correctly. The `read_gate.py` posture entry was verified to accurately describe the gate module/symbol pair that the new runtime-probe test checks against.



## Verdict: LGTM

The fix correctly closes the described privacy leak by threading a hard rank-cutoff gate (`tier_within_override`) into the sole producer of `VaultSnapshot.fragments`, verified against both production entry points (MCP tool and CLI). It reuses existing, already-tested primitives rather than inventing new filtering logic, the new tests assert on written file bytes (not the leak-blind response envelope) with proper positive controls and a full truth table for the two independent gates, and the PR is explicit and honest about what remains out of scope (thread/eddy skills, filed as #1284) rather than overclaiming. No blocking issues found in static review.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5235129058 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

PR #1070, opened 2026-07-31
refactor: centralize Discord text contracts

tomatotomata:
## Summary

- centralize the Discord reply cap and MCP-unavailable text in `crawdad.discord_text`
- update bot, CLI, loop, and slash command callers without changing values or behavior
- add a source-level guard against duplicate definitions and retarget the CLI truncation tests

Closes #1061

## Validation

- `.venv/Scripts/ruff.exe check ...` passed
- `.venv/Scripts/ruff.exe format --check ...` passed
- `.venv/Scripts/mypy.exe crawdad/discord_text.py crawdad/bot.py crawdad/cli.py crawdad/loop.py crawdad/slash_commands.py` passed
- `.venv/Scripts/python.exe -m pytest -o addopts='' tests/test_discord_text.py tests/test_cli.py -q` passed: 24 tests
- `.venv/Scripts/python.exe -m bandit -q -r crawdad/discord_text.py` passed
- `git diff --check` passed

The focused bot and CLI run also exposed three pre-existing Windows path-separator assertions in attachment tests. They are unrelated to this refactor. The repository has no checked-in pre-commit config, so that command could not be run.

AI-assisted contribution. I reviewed the diff and test output manually.

PR #1288, opened 2026-08-10
fix(crawdad): refuse a staging root the redaction scan cannot reach

### LINKED ISSUE #1088: CrawDad's configurable attachments.staging_subpath can fall outside creek.redact.scan's canonical scope, silently disabling the safety pass

Geoffe-Ga:
## Summary

CrawDad's `attachments.staging_subpath` is operator-configurable and its
validator refused only absolute paths and `..` segments. `creek.redact.scan`
is hard-scoped to `00-Creek-Meta/Inbound/` (#972): that subtree is admitted at
every ceiling, and any *other* vault path is ranked as intimate content and
needs `ceiling=intimate`/`all` — which no Discord channel gets by default. So
a deployment that pointed `staging_subpath` anywhere else drew
`status="refused"` on every safety scan, and the safety pass silently never
ran.

While fixing that I found the live half of the defect is worse than the issue
describes, and it survived #1285. `_run_safety_scan` returned `scanned=True`
the instant `session.call_tool` returned — **including when the tool answered
with a refusal envelope**. So the #1054 dispatch gate, which refuses any batch
with `scanned=False`, never saw the refusal: the batch was recorded as
scanned, the raw refusal JSON was rendered to the user under a "Safety scan"
heading, and `creek.ingest` was dispatched for content nothing had read.

The fix is two layers, because they close different windows:

**Layer 1 — config parse-time refusal (`crawdad/crawdad/config.py`).**
`_DEFAULT_STAGING_SUBPATH` becomes a public `CANONICAL_STAGING_ROOT: Final[Path]`,
mirrored (not imported) from `creek_mcp.tools.redact._STAGING_SUBDIR` on the
same precedent as `bot.py::_REDACT_SCAN_TOOL`. The `staging_subpath` validator
gains a third arm — last, after `is_absolute()` and `".." in parts` — refusing
anything not `is_relative_to(CANONICAL_STAGING_ROOT)`. `validate_default=True`
moves onto `model_config` so a future edit to the constant, or a subclass with
a drifted default, is caught at first construction rather than shipping green.

**Layer 2 — the runtime arm (`crawdad/crawdad/bot.py`).** A JSON object whose
`status` is exactly `"refused"` is the fifth "the scan did not run" mode, and
now yields `_ScanOutcome(scanned=False, …)`. The existing single chokepoint in
`_dispatch_ingest_for_batch` does the rest — no second gate, per §5.3.

### Why arm order is load-bearing

`Path.is_relative_to` is a pure lexical prefix test and does not resolve `..`,
so `00-Creek-Meta/Inbound/../../01-Fragments` **is** lexically relative to the
canonical root. Only the pre-existing `..` arm catches it, which is why the
scope arm must stay last. A test pins this by asserting the `..` message fires
and the scope message does not.

### What this does *not* claim

The config check is **lexical, fail-fast, defence in depth — not
confinement**. It calls no `resolve()`, and `AttachmentConfig` is a nested
model with no `vault_path`, so a symlink parked under the canonical root
pointing at `01-Fragments` still passes; that is #1087, still open. The real
confinement boundary remains creek-tools' `resolve_within_vault` plus the
resolved `is_relative_to` at `creek_mcp/tools/redact.py:340`. Every docstring
says so.

The new user-facing string claims only that no scan ran. It does not say the
files are clean and makes no counterfactual claim about what a successful scan
would have established — `creek.redact.scan` returns a permissive `ok`/`empty`
even when it finds secrets, which is the recorded non-goal in
`crawdad/CLAUDE.md` §5.3. **That non-goal is undisturbed:** `ok` and `empty`
both still clear the gate; only `refused` — a scan that was *declined*, never
a scan that found something — flips `scanned`.

It also does not assert *which* refusal fired. `status="refused"` covers every
refusal reason creek-tools has, not just the out-of-scope staging root —
`input_path not found` (`redact.py:200-205`) and the off-vault-symlink case
land there too. Naming the staging root as the definite cause would misdirect
an operator debugging something else, so the fixed copy hedges and
`_scan_refusal_text` echoes creek-tools' own `reason` verbatim, which is
authoritative for whichever refusal actually fired. The echo discloses nothing
new: an out-of-scope reason is a fixed string, and a "not found" reason names
a staging path the download summary already showed the user in the same turn.

Fail-open on anything else is deliberate. Non-JSON, malformed JSON, a JSON
array, and a status-less object all keep today's `scanned=True`, so this
change can only ever *remove* admissions and can never newly refuse a turn
that works now. Six parametrised cases pin that.

## Breaking config change

A deployment already overriding `attachments.staging_subpath` to a
non-canonical path now **fails at startup** instead of degrading to a
permanently-refused scan. That is the issue's explicit preference ("prefer
refusing early over a runtime warning"), and the startup error names the
canonical root and the fix. No `CONTRACT_VERSION` bump and no wire-shape
change. The in-repo `crawdad/crawdad.yaml` is untracked and carries no
`attachments` block, so no committed fixture moves.

Also corrected: `crawdad/CLAUDE.md` claimed `capture_subpath` gets "the same
absolute/`..` validation as `attachments.staging_subpath`". The third arm made
that false, so the bullet now says what actually holds — `capture_subpath`
gets arms 1 and 2 but not the canonical-root arm, because nothing scans the
capture dir.

## Test plan

**RED first, and behavioural.** Before the fix, the headline test failed with
a live ingest dispatched for never-scanned content, at `open` — the *narrowest*
declared ceiling:

```
FAILED tests/test_bot.py::test_refused_scan_response_cannot_reach_creek_ingest
E   AssertionError: assert [{'name': 'cr...ng': 'open'}}] == []
E     Left contains one more item: {'name': 'creek.ingest', 'args':
E       {'source_type': 'markdown',
E        'input_path': '00-Creek-Meta/Inbound/999/100/note.md',
E        'privacy_tier_ceiling': 'open'}}
```

12 tests red before, 689 green after, no pre-existing test modified.

**Each guard proven load-bearing** by disabling it in isolation and confirming
exactly the intended tests go red:

| Guard disabled | Tests red |
|---|---|
| the refusal block in `_run_safety_scan` | exactly 6 |
| the third arm in `_confine_staging_subpath` | exactly 9 |
| `validate_default=True` on `model_config` | exactly 1 |
| (regression probe) a cleanliness claim injected into the refusal copy | exactly 2 |

**Coverage of the behaviour:**
- Refusal cannot reach `creek.ingest` via the `ingest` route or the
  `ingest` → type-question → type-word disambiguation route (both dispatch
  entry points).
- Over-refusal guard: a `status="ok"` body still ingests and still renders
  `report_markdown` as markdown, not a JSON code fence.
- Fail-open pins ×6: plain text, brace-wrapped-unparseable, object without
  `status`, `status="ok"`, `status="empty"`, and a JSON **array** (which forces
  the `isinstance(parsed, dict)` guard).
- Validator accept table ×6 including `00-Creek-Meta/./Inbound`, a trailing
  slash, and a `str` (pydantic coerces through the same chain).
- Validator refuse table ×6 including the sibling prefix
  `00-Creek-Meta/Inbound-other` (forces `is_relative_to` over `startswith`),
  `00-creek-meta/inbound` (case folding deliberately not done — `redact.py:340`
  compares literally, so folding would admit a config the server then refuses),
  the parent `00-Creek-Meta`, and `""`.
- Refusal holds at every ceiling, including `intimate`.
- Constant-mirror pin against `creek_mcp/tools/redact.py:98`.
- The refusal copy's *delivered content* is pinned, not just executed:
  `_assert_claims_no_cleanliness` fails if any delivered refusal message
  contains "clean", "no findings", "no secrets", "safe to" or "nothing found".
- A refusal whose reason is *not* the staging root echoes that reason instead
  of misdirecting; a refusal with a missing, blank, or non-string reason still
  refuses and emits no dangling "creek-tools said:" fragment.

**Gates.** `crawdad/scripts/check-all.sh` → 7/7 green, 689 passed, coverage
94.74% (`bot.py` 94.86% → 95.00%); `ruff check --no-cache` re-verified clean,
and the change adds no newly-uncovered line. The `creek-tools` gate is
unaffected: the only file touched there is `docs/mcp.md`, that gate has no
markdown step, and nothing reads the file at runtime or in tests.

## Gate 2.5 self-review

Returned FIX_REQUIRED with one MAJOR, both fixed before this PR opened:

- **MAJOR** — `_SCAN_REFUSED_REPLY` was executed but its content was never
  asserted; both refusal tests cleared the channel before checking, so only
  the pre-existing `_SCAN_BLOCKED_REPLY` was pinned. A future edit could have
  reintroduced a cleanliness claim with no test going red — exactly the
  failure mode §5.3 warns about. Now pinned on the staging turn.
- **Related (folded in, not deferred)** — the copy asserted the staging root
  as the definite cause, which is false for the other refusal reasons. Now
  hedged, with creek-tools' `reason` echoed verbatim.
- Two NITs (the mirror-pin test's docstring overclaiming what a same-file
  literal comparison can detect; `_confine_staging_subpath`'s name vs. its
  "this is not confinement" docstring) were judged not worth churn — the
  docstrings already state the limitation honestly in both cases.

## Notes for the reviewer

- Open PR #1070 ("centralize Discord text contracts") also touches
  `crawdad/crawdad/bot.py`; whichever lands second may need a trivial textual
  merge around the new module constant.
- `creek-tools/docs/mcp.md:163-171` (the #1087 symlink residual) is left
  byte-identical on purpose — only the following paragraph, which ended
  "tracked by #1088", was rewritten.

Closes #1088
Refs #972, #1054, #1087, #1152

### DIFF
--- a/crawdad/CLAUDE.md
+++ b/crawdad/CLAUDE.md
@@ -111,8 +111,14 @@ Four gates, each must pass before the next:
   - `capture_enabled` — bot-capture toggle (#687), default `False`.
     Opt-in per deployment; see [§5.2](#52-bot-capture-boundary).
   - `capture_subpath` — vault-relative dir the capture writer appends
-    to, default `discord-capture`. Must stay inside the vault (same
-    absolute/`..` validation as `attachments.staging_subpath`).
+    to, default `discord-capture`. Must stay inside the vault — the same
+    absolute/`..` validation `attachments.staging_subpath` gets, but not
+    its canonical-root arm (#1088): nothing scans the capture dir.
+  - `attachments.staging_subpath` — vault-relative attachment staging
+    dir, default `00-Creek-Meta/Inbound`, and it must stay under that
+    root: it is the only subtree `creek.redact.scan` admits at a
+    channel's ceiling, so anywhere else the safety pass never runs
+    (#1088). See [§5.3](#53-redaction-scan-gate-feat-027-1054).
   - `attachments.channel_privacy_tiers` — per-channel declared ceiling
     (`open` / `personal` / `intimate` / `all`), validated at
     config-parse time. See [§5.2](#52-bot-capture-boundary) for how
@@ -236,11 +242,37 @@ would miss the point of the fix.
 
 Related, explicitly **not** covered here: #1087 (`scan_batch` follows
 symlinks out of the scan root — a scan that *runs* but under-reports,
-and this gate marks such a batch `scanned=True`). #1088
-(`attachments.staging_subpath` outside the scan's canonical scope) *is*
-a fifth way the scan fails to run; this design absorbs it with one more
-`_ScanOutcome(scanned=False, …)` arm, so sequence it next rather than
-touching `_run_safety_scan` twice.
+and this gate marks such a batch `scanned=True`).
+
+#1088 (`attachments.staging_subpath` outside the scan's canonical
+scope) was the fifth way the scan fails to run, and it has now landed
+in two layers. **Config gate:** `AttachmentConfig` refuses at
+config-parse time any `staging_subpath` outside
+`CANONICAL_STAGING_ROOT` (`00-Creek-Meta/Inbound`, mirrored from
+`creek_mcp/tools/redact.py`), with `validate_default=True` on the model
+so a drifted *default* is caught too. That check is lexical — the real
+confinement boundary is still creek-tools' `resolve_within_vault` plus
+its resolved `is_relative_to`. **Runtime arm:** `_run_safety_scan`
+gained the fifth `_ScanOutcome(scanned=False, …)` arm, for a response
+body that is a JSON object whose `status` is exactly `"refused"` —
+creek-tools' `refusal_response()` envelope. Every other body keeps
+`scanned=True`, deliberately: fail-open on unparseable, non-object, or
+status-less responses means the change can only ever *remove*
+admissions.
+
+This does **not** disturb the non-goal above. `ok` and `empty` both
+still clear the gate, so a scan that runs and reports secrets still
+ingests. Only `refused` — a scan that was *declined*, never a scan that
+found something — flips `scanned`, and the user-facing copy
+(`_SCAN_REFUSED_REPLY`) claims only that no scan ran.
+
+That copy is pinned by test, not just by review: `status="refused"`
+covers every refusal reason creek-tools has (out-of-scope root, but
+also `input_path not found`), so the fixed text hedges about the cause
+and `_scan_refusal_text` echoes creek-tools' own `reason` verbatim.
+`_assert_claims_no_cleanliness` asserts no delivered refusal message
+contains a cleanliness claim — if a future edit reintroduces one, a
+test goes red rather than a reviewer having to notice.
 
 **Revisit predicate:** when `creek.redact.scan` gains a failing status
 for non-empty findings, revisit whether `scanned` should become a
--- a/crawdad/README.md
+++ b/crawdad/README.md
@@ -68,7 +68,7 @@ allowed_channel_ids:
 | `mcp_server_command` | no | `[creek-tools-mcp]` | argv for the MCP subprocess. |
 | `allowed_user_ids` | yes | — | Discord user ids permitted to message the bot. |
 | `allowed_channel_ids` | yes | — | Channels the bot will respond in. |
-| `attachments` | no | see source | Per-attachment limits, allow/deny lists, channel privacy tiers (FEAT-027/035). |
+| `attachments` | no | see source | Per-attachment limits, allow/deny lists, channel privacy tiers (FEAT-027/035). `staging_subpath` must stay under `00-Creek-Meta/Inbound/`: `creek.redact.scan` is scoped to that subtree (#972), so anywhere else is ranked as intimate content and the safety pass never runs. A non-conforming value is refused at config-parse time (#1088). |
 | `consent` | no | see source | Conversational consent tokens + TTL (FEAT-034). |
 | `max_loop_rounds` | no | `5` | FEAT-036 agent-loop round cap, bounded `[1, 50]`. Raise when a single user turn legitimately needs more router/dispatcher passes (multi-source ingest, large re-classification asks, long workflow chains); the upper bound prevents pathological runaway loops. |
 | `capture_enabled` | no | `False` | Bot-capture toggle (#687) — see [Bot capture](#bot-capture) below. |
--- a/crawdad/crawdad/bot.py
+++ b/crawdad/crawdad/bot.py
@@ -72,6 +72,17 @@ _DISCORD_REPLY_LIMIT = 1900
 # no Python-level dependency on creek-tools beyond the MCP contract.
 _REDACT_SCAN_TOOL = "creek.redact.scan"
 
+# #1088: the ``status`` value creek-tools returns when it *declines* to
+# scan rather than scanning. Sourced from ``refusal_response()``
+# (``creek-tools/creek_mcp/tier_ceiling.py:264-281``), which returns
+# exactly ``{"status", "tool", "tier_ceiling", "reason"}``;
+# ``creek.redact.scan`` is registered ``-> dict[str, Any]``
+# (``creek-tools/creek_mcp/server.py:637``), so FastMCP JSON-serialises
+# that dict and :meth:`crawdad.mcp_client.MCPSession.call_tool` hands the
+# bot the JSON as a plain string. Mirrored, not imported, for the same
+# reason as :data:`_REDACT_SCAN_TOOL` above.
+_SCAN_REFUSED_STATUS = "refused"
+
 _REDACT_TOOL_MISSING_REPLY = (
     "I downloaded the file(s), but the `creek.redact.scan` tool isn't advertised "
     "by creek-tools yet. Run `creek redact --scan <staging path>` from the "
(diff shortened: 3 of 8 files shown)


Geoffe-Ga:
## Summary
Fixes #1088: an operator-configured `attachments.staging_subpath` outside `00-Creek-Meta/Inbound/` caused `creek.redact.scan` to be refused at every default Discord-channel ceiling, and — worse — `_run_safety_scan` treated that refusal envelope as a successful scan (`scanned=True`), letting `creek.ingest` dispatch content nothing had actually read. The PR closes this in two layers: a parse-time config validator (`AttachmentConfig._confine_staging_subpath`) that refuses any `staging_subpath` outside the canonical root, and a runtime arm in `bot.py` that treats a `status="refused"` MCP response as `scanned=False`.

## Strengths
- Root-caused correctly and fixed at both the config layer (prevention) and the runtime layer (detection) — the two failure windows are genuinely different and each needed its own fix.
- The `..`-before-scope arm ordering is subtle and correctly reasoned about (`is_relative_to` is a lexical prefix test that doesn't resolve `..`), and is pinned by a dedicated test (`test_config_refuses_traversal_out_of_canonical_root_via_the_dotdot_arm`).
- `_parse_scan_body` deduplicates JSON parsing between the refusal check and `report_markdown` extraction, removing a prior double-`try/except` that could drift.
- Fail-open design is deliberate and well-tested: unparseable/non-object/status-less bodies keep `scanned=True`, so the change can only *remove* admissions, never introduce a new refusal for content that worked before (6 parametrized pins).
- Refusal copy is disciplined about not overclaiming — it says only "no scan ran," never "files are clean" (matches the recorded §5.3 non-goal), and hedges about *cause* since `status="refused"` covers multiple reasons, echoing creek-tools' own `reason` verbatim rather than guessing.
- Verified all cross-repo claims (mirrored constant, refusal envelope shape, `_OUT_OF_SCOPE_REASON` text) against the actual `creek-tools/creek_mcp/tools/redact.py` and `tier_ceiling.py` source — every one checks out exactly.
- `validate_default=True` moved to `model_config` (not per-field `Field(...)`) is the right call for subclass inheritance, and is itself pinned by a test that exercises a subclass with a drifted default.
- Test suite is unusually rigorous: each guard is proven load-bearing by disabling it in isolation (per PR description), both ingest dispatch entry points (`ingest` route and type-disambiguation route) are covered, and the refusal copy's delivered content is asserted, not just executed.
- Self-documented Gate 2.5 review already caught and fixed a real gap (refusal copy executed but unasserted) before this PR opened.

## Security Concerns
None. 🟢 The PR is explicit and honest that this is a lexical, fail-fast, defence-in-depth check — not a confinement boundary — and correctly identifies that the real boundary remains creek-tools' `resolve_within_vault` + resolved `is_relative_to`. The known residual (#1087, symlink under the canonical root) is called out rather than hidden.

## Problems
None found blocking merge.

## Code Quality
Nothing worth blocking on. Two very minor nits already surfaced and consciously deferred in the PR's own Gate 2.5 self-review (a docstring slightly overclaiming what a same-file literal comparison detects, and a naming/docstring tension on `_confine_staging_subpath`) — both are honestly documented in-place and not worth further churn.

## CLAUDE.md Adherence
✅ TDD workflow followed (RED-first test cited with failure output in PR description). ✅ `crawdad/CLAUDE.md` §5.3 updated accurately and cross-checked against the implementation. ✅ No `# type: ignore`/`# noqa`/`--no-verify` shortcuts introduced. ✅ Docstrings are thorough (well above the 95% interrogate bar based on density). ✅ Conventional commit message. ✅ `check-all.sh` reported 7/7 green, 689 passed, 94.74% coverage per the PR description.

## Testing
Comprehensive and behavioral, not just line-coverage-driven: accept/refuse tables for the new validator arm (including the `is_relative_to`-vs-`startswith` sibling-prefix case and deliberate non-case-folding), the arm-ordering `..` pin, a `validate_default` subclass pin, a `load_config` YAML-path pin, both dispatch-route refusal pins, an over-refusal guard (non-refusal bodies still ingest), 6 fail-open pins, a reason-echo/hedge pin, and a cleanliness-claim regression pin. This is well above the bar for a security-adjacent gate fix.

## Documentation
`crawdad/CLAUDE.md`, `crawdad/README.md`, and `creek-tools/docs/mcp.md` are all updated and, on inspection, accurately reflect the shipped code (verified `capture_subpath`'s validator genuinely lacks the third arm as the doc now claims, and the mirrored constant/reason text match creek-tools' source verbatim).



## Verdict: LGTM

The fix is correct (verified against creek-tools' actual refusal_response()/_STAGING_SUBDIR/_OUT_OF_SCOPE_REASON source, which the mirrored constants and messages match exactly), arm ordering and lexical-vs-confinement caveats are accurate, and the fail-open design is deliberately conservative. Test coverage is exhaustive (accept/refuse tables, arm-ordering pin, validate_default subclass pin, fail-open pins, cleanliness-claim pin, both dispatch routes) and CLAUDE.md/README/docs updates are all verified accurate against the code. No blocking issues found.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234839145 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1287, opened 2026-08-10
test(read-gate): no ancestry-shaped fixture — #1283's compile refusal is unpinned in the manifest layers


Geoffe-Ga:
**Split from #1036** (layer (g), the prompt-channel canary layer), and a direct follow-on from **#1283** which closed #931.

## Why this is not already covered

#1283 closed the `structural_path` ancestry channel: `creek_mcp.tools.compile._survey_sources` now ranks the whole ancestor chain through `creek.classify.privacy_filter.ancestry_tiers`, so an admitted `open` child of an above-ceiling parent is refused rather than carrying that parent's title into the compile prompt.

#1036 added a `creek.compile` **prompt** probe, and it deliberately does *not* assert that refusal. It cannot: once the gate fires, `compile_tool` returns above the LLM factory, so there is no prompt to inspect and layer (g)'s non-emptiness control would fail. The assertion is layer-(f)-shaped — "the tool refused and never reached the model" — not layer-(g)-shaped.

The gap is therefore narrow but real: **`tests/test_mcp_read_gate.py` has no ancestry-shaped fixture at all.** Layer (f)'s `canary_vault` and layer (g)'s `prompt_canary_vault` are both flat — no fragment carries a `parent_id` or a persisted `structural_path` — so the manifest layers are structurally blind to the channel #931 was about, even though the channel itself is now fixed.

## Verified behaviour (measured on a base predating #1283)

Seeding the `open` fragment with `structural_path: ["Intimate canary <SENTINEL>"]` and compiling only that fragment at `ceiling=open`:

```
status: ok | n prompts: 1
prompt: OPEN=True INTIMATE=True      <-- the parent title crossed to the model
  >> structural_path: Intimate canary CANARY-RUNTIME-INTIMATE-7d21
```

The sibling `parent_id`-walk shape did **not** leak on that base (the ancestor is not among the loaded ids, so `by_id` cannot resolve it) — the persisted frontmatter field is the reachable route, because `creek.hierarchy.structural_path_context` returns it before walking `parent_id` at all.

## Scope

1. Teach `_write_fragment` an optional `structural_path` (and `parent_id`) parameter. Keep the default `None`/omitted so every existing caller writes byte-identical frontmatter — several layer-(f) tests make exact-set claims and must not shift.
2. Add an ancestry-shaped fixture (a sibling of `prompt_canary_vault`; do **not** mutate the flat ones).
[...]

Issue #1279, opened 2026-08-09
creek.author's layer-(f) probe exemption is false: the desk starts and the factory is invoked


Geoffe-Ga:
**Found during #1036 and proven false by execution.**

`creek.author`'s layer-(f) exemption in `tests/test_mcp_read_gate.py`'s `_PROBE_EXEMPT` reads:

> author_tool calls load_config() and drives the Writing Desk, and it wraps that whole span in `except Exception` returning status='error'. On any unconfigured-provider or missing-contract failure it therefore returns an envelope with no corpus content in it at all, so the canary assertion would pass for the wrong reason — **the desk never started**. A real probe needs a configured desk and belongs beside the other author tests.

Every load-bearing clause of that is now wrong. Run against the layer-(f) canary fixture with a client implementing author's real protocol:

```python
author_tool(
    vault_path=vault, query="canary", llm_factory=recorder,
    privacy_tier_ceiling=TierCeiling.OPEN, dry_run=False,
)
```

→ `status='ok'`. The desk **starts**. The factory **is invoked** (tier `OPEN`). A real **2134-char prompt** crosses to the client, carrying the open canary and neither the intimate nor the personal one. The response envelope carries the open canary too. No config file, no live provider, no deployed skills tree required.

The exemption was written before #1254/#1260 landed the author LLM seam (`AuthorLLMFactory` / `AuthorLLMClient.complete_with_usage`), and it has been stale since. #1036 gave `creek.author` a real **prompt**-channel probe (layer (g)) on the strength of the run above; the **response**-channel exemption was deliberately left alone as out of scope for that issue, which is why this exists.

## Why this matters beyond tidiness

`_PROBE_EXEMPT` is the single escape hatch out of layer (f), and `test_probe_exemptions_are_specific_to_their_tool` can only check that the reason names the right module and clears a length floor — it cannot check that the reason is **true**. So a stale exemption is a machine-blessed false claim: the manifest reports `creek.author` as "cannot be probed here" while a probe demonstrably runs. That is precisely the "a manifest is only worth having if it cannot lie" premise of #932.

## Scope

1. Delete `creek.author` from `_PROBE_EXEMPT` and add a real `_RUNTIME_PROBES` entry driving `author_tool` at `ceiling=open` over `canary_vault` with a recording author factory, asserting the intimate canary appears nowhere in the serialised response.
[...]

Issue #1275, opened 2026-08-09
Hoist the five recording-LLM-factory copies into one shared pair of test helpers


Geoffe-Ga:
**Found during #1036.** House rule §1.2 (DRY / single source of truth) tension, filed rather than fixed inside that lane because hoisting a shared fixture out of four test modules is a refactor with its own blast radius and does not belong inside a guardrail change.

There are now **five** recording-LLM-factory implementations in the test suite, each capturing "what text actually crossed to the model":

| Class | File |
|---|---|
| `_RecordingFactory` | `tests/test_mcp_reflect.py` |
| `_ProviderSpy` | `tests/test_mcp_write_tools.py` |
| `_TierRecordingFactory` | `tests/test_mcp_tools.py` |
| `_RecordingLLMFactory` | `tests/test_mcp_read_gate.py` (#1036) |
| `_RecordingAuthorFactory` | `tests/test_mcp_read_gate.py` (#1036) |

They are not gratuitous duplicates — there really are two distinct factory protocols in production (`(tier) -> (str) -> str` for reflect/compile/draft, and `AuthorLLMFactory: (tier | None) -> AuthorLLMClient | None` consumed through `.complete_with_usage` for author) — but five copies of "record the prompt" is four too many, and they have already drifted in what they record (some capture the tier, some the system prefix, some neither).

## Scope

1. Hoist into `tests/conftest.py` (or a `tests/support/` helper module) **two** recorders, one per real protocol, each recording tier *and* prompt *and* any system prefix.
2. Migrate all five call sites without weakening a single existing assertion — `tests/test_mcp_reflect.py`'s #964 tests and `tests/test_mcp_read_gate.py`'s layer (g) both make exact claims about what the recorder saw, and a helper that normalises the captured text differently would silently change what they prove.
3. Keep the "why two, not one" argument in the shared docstring. #1036's judged design specifically rejected a single harness-owned recorder because it is only expressible while every probeable tool shares one client shape, and author already does not.

Refs #1036, #964, #1254, #1260

PR #1289, opened 2026-08-10
test(mcp): watch the prompt channel, not just the response envelope (#1036)

### LINKED ISSUE #1036: read_gate layer (f) cannot see a prompt-side leak: the canary probe asserts on the response, and refuses before the second corpus read

Geoffe-Ga:
## Summary

- Adds **layer (g)** to `tests/test_mcp_read_gate.py`: a **prompt-channel** canary layer that drives every LLM-backed `GATED` tool all the way to the model with a recording factory and asserts no above-ceiling sentinel appears in what crossed the wire. Layers (a)–(f) all terminate at the response envelope; the prompt leaves the process *before* that envelope exists, so a tool could paste an intimate fragment into a prompt, ship it to a cloud endpoint, and return something layer (f) certifies as clean.
- Drives it off a **derived** forcing function, not a hand-list. `_LLM_BACKED_GATED_TOOLS` is computed by signature introspection (`GATED` ∩ modules defining a function taking both `llm_factory` and `privacy_tier_ceiling`), so a newly-`GATED` LLM-backed tool cannot inherit "response-probed, prompt-unchecked". `_PROMPT_PROBES` ∪ `_PROMPT_PROBE_EXEMPT` must equal that set, with disjointness, no stale entries, a membership pin on each side, and — the assertion neither candidate design had — **non-emptiness**, because a signature-derived set that silently empties makes the whole layer vacuously green.
- Answers the issue's follow-up comment in writing: `read_gate.py`'s docstring and `docs/mcp.md` now name the **three egress channels** — response (layer f, forced), prompt (layer g, forced), disk artifacts (per-tool today, **not** forced, split to #1273). Issue **scope item 4 is decided NO**, with the reason and a revisit predicate recorded.

Test-only change plus docstrings. **No behavioural code changed** — `creek_mcp/read_gate.py` is a production file but its diff is the module docstring alone (verified: with docstrings stripped, its AST is byte-identical to HEAD). That is deliberate: a production edit here would mean a real leak was found, which gets filed separately rather than fixed inline.

## The RED proof

The forcing function was written **first**, with both manifests empty. Verbatim failure at that point — note the four names are *derived*, never typed into the test:

```
E  AssertionError: LLM-backed GATED tool(s) with no prompt probe and no exemption:
   ['creek.author', 'creek.compile', 'creek.draft', 'creek.reflect']. Add a probe to
   _PROMPT_PROBES, or record in _PROMPT_PROBE_EXEMPT the specific reason this tool's
   prompt cannot be captured here. Layer (f)'s response probe cannot see the prompt
   channel: it reads what came back, and by then the prompt has already crossed to
   the provider.
E  assert not {'creek.author', 'creek.compile', 'creek.draft', 'creek.reflect'}

tests/test_mcp_read_gate.py:2645: AssertionError
1 failed, 137 passed
```

That is a **behavioural** assertion failing (`assert not unchecked`) on a module that imported cleanly with 137 other tests passing — not a missing symbol or a collection error.

## Mutation proof — the discriminating evidence

A passing test is not proof a gate works, and this issue is about a *detection mechanism* being blind. So the argument for layer (g) is the **asymmetry**, measured with two independent injections (applied as scratchpad-only pytest plugins; no tracked source was edited, and `git stash` was never used — it is one shared ref across four live worktrees):

| Injection | layer (f) `[creek.reflect]` (response) | layer (g) `[creek.reflect]` (prompt) |
|---|---|---|
| `creek.author.agents.tier_within_override` → always `True` | **PASSED** | **FAILED** |
| `creek_mcp.tools.reflect.to_privacy_override` → always `ALL` | **PASSED** | **FAILED** |

Under injection 1 layer (g) also reddened `[creek.author]`; **all ten** layer-(f) response probes stayed green under both. Layer (f) stays green because its reflect probe passes `entry_ref=<intimate id>`, which the #846 gate refuses *before* the grounding walk — the probe never reaches the second corpus read at all. That is the issue in one sentence.

Direct measurement under injection 1, at `ceiling=open`:

```
[HEAD]    layer (g) prompt: OPEN=True INTIMATE=False PERSONAL=False
[MUTANT]  layer (g) prompt: OPEN=True INTIMATE=True  PERSONAL=True
[HEAD]    layer (f) response: status=refused INTIMATE_in_response=False
[MUTANT]  layer (f) response: status=refused INTIMATE_in_response=False
```

## Three corrections found by execution

1. **`creek.author` is prompt-probeable — it got a real probe, not an exemption.** The plan called for exempting it on a client-protocol mismatch. Run with a client implementing author's real protocol, `author_tool` returns `status='ok'`, the factory is invoked with `PrivacyTier.OPEN`, and a real 2134-char prompt crosses carrying the open canary and neither above-ceiling one. An exemption would have been a machine-blessed false claim in the layer's only escape hatch. This also means **`creek.author`'s existing layer-(f) `_PROBE_EXEMPT` reason is false** ("the desk never started") — left untouched as out of scope, filed as **#1279**.
2. **A personal canary was added, and it is the assertion that matters most.** The `canary_vault` pair (`open` + `intimate`) is blind to the residue class this repo already knows about: at `ceiling=open` the pipeline DROPS intimate but `filter_fragments_by_tier` SUMMARISES personal to `[Personal-tier summary: {title}]`, so above-ceiling personal **titles** survive by design on that path — pinned live by `tests/test_mcp_tools.py::test_draft_prompt_carries_the_personal_title_at_an_open_ceiling`. A layer asserting only "intimate not in prompt" would be green forever over that shape. New fixture `prompt_canary_vault` adds a third `personal` fragment; `canary_vault` is untouched (its exact-set positive controls would break).
3. **`creek.draft`'s exemption is executable, not prose.** It does reach the model, so "the model is never reached" would be false. The real reason is upstream: on a fixture this small the IdeaMiner falls back to `MiningStrategy.UNEXPLORED_ONTOLOGY`, whose seed leaves `source_fragments` empty, so a canary assertion would be vacuous in *both* directions. `test_the_draft_prompt_exemption_is_still_true` executes that reason — the day the miner surfaces a corpus-backed seed on this fixture, the exemption goes red and someone must write a real probe.

## Honest scoping — what this does not claim

- **The reflect assertion duplicates `tests/test_mcp_reflect.py::test_above_ceiling_fragment_contributes_nothing_not_even_a_title` on purpose.** The new thing is the **forcing function**, not the assertion. The issue's complaint was never that reflect was unasserted — it was that nothing *required* the next LLM-backed tool to do the same.
- **The compile probe's two exclusion assertions are weak, and the docstring says so** rather than leaving a reader to assume otherwise. `compile_tool` loads only caller-named ids, so the above-ceiling fragments are never candidates. Both routes by which above-ceiling material *could* reach that prompt refuse above the factory — a named id (`_ABOVE_CEILING_REASON`) and an admitted child's ancestry (#931, **closed by #1283**, which ranks the chain via `ancestry_tiers`) — which puts them out of a prompt probe's reach by construction: once the gate fires there is no prompt to inspect. What the compile probe actually carries is its **positive control**. Asserting #1283's refusal is a layer-(f)-shaped claim and is filed as **#1287**, which also records that no fixture in this file is ancestry-shaped today.
- **`creek.classify` is outside both manifests** and named as such: `METADATA_ONLY` on a response-shaped rationale while `method="llm"` drives the whole corpus through a provider, defended only by ModelRouter's Intimate-never-cloud chokepoint. It takes no `llm_factory`, so the derivation structurally cannot enrol it. Filed as **#1274**.
- **Scope item 4 decided NO.** A second `(gate_module, gate_symbol)` pair is only checkable at layer (c)/(e)'s exists-and-is-called strength, and for reflect's grounding gate that check is **green by construction** — `creek/author/agents.py` imports `tier_within_override` at :28 and calls it in `_load_corpus` at :115 for every Writing Desk consumer, so it would stay green if reflect stopped passing `override` or dropped grounding entirely. Layer (g) checks the *effect*, which is strictly stronger; the table above is the evidence. Revisit predicate recorded: *a tool whose second gate's effect no runtime probe can observe.*
- No `chief-architect` pass ran for this lane: the orchestrator supplied a pre-verified premise-and-design dossier and directed me to execute it directly. The three corrections above are mine, from execution against HEAD.

## Test plan

- `./scripts/check-all.sh` → **exit 0**, 12/12 checks. 8664 passed, 5 skipped; branch coverage 94.36% (≥90); per-file gate 255 files ≥80% (2 waivered ≥65%); mypy strict, ruff (`--no-cache`), bandit, pip-audit, xenon, interrogate, pylint all clean.
- `tests/test_mcp_read_gate.py`: **136 → 144 passed**, additive. `git diff | grep '^-[^-]'` over that file shows exactly **one** changed pre-existing line — "Six layers keep it honest" → "Seven layers". No assertion, test, fixture or parametrize case deleted or weakened.
- Both injections above run and recorded; layer (f) green under each, layer (g) red under each.
- Canary exclusions match **case-insensitively**, so a renderer that slugified or lower-cased a title on its way into the prompt cannot evade them.
- No bypasses: no new `# noqa`, `# type: ignore`, `pytest.mark.skip`, threshold edit, or coverage waiver.

## Reviewer notes

- The branch is based on `e0a51b4` and is **3 commits behind `origin/main`** with no commits of its own beyond this one; it does not yet contain #1283. Textual merge is clean (the layer-(g) block sits ~130 lines from main's `read_gate.py` edit), but the orchestrator owns the sync — I did not rebase.
- Lane **#971** is editing all three of these files in its own worktree. The `read_gate.py` insertion here is one contiguous docstring block, kept deliberately narrow to make that sync cheap.

Follow-ups filed: **#1273** (artifact-channel forcing function), **#1274** (classify egress), **#1275** (recorder hoist), **#1279** (author's stale layer-(f) exemption), **#1287** (ancestry-shaped fixture for #1283's refusal). Cross-reference comments posted on #931 and #971.

Closes #1036
Refs #932, #964, #1283, #971, #1068

### DIFF
--- a/creek-tools/creek_mcp/read_gate.py
+++ b/creek-tools/creek_mcp/read_gate.py
@@ -121,6 +121,70 @@ fails if an entry lies: the manifest must match ``server.list_tools()``
 exactly, a ``GATED`` claim is checked against a real call site in the named
 module, a gap must name a positive issue that the tool's own source mentions,
 and a tool recorded as a gap must not already call a primitive below.
+
+**The three egress channels, and which one each layer watches.** #1036 added
+a seventh layer, (g): every ``GATED`` tool that can hand corpus text to a
+model is driven all the way *to* the provider with a recording factory, and no
+sentinel from an above-ceiling fragment may appear in any prompt it sent. It
+exists because every layer before it terminates at the response envelope, and
+a prompt has already crossed to the provider by the time an envelope exists.
+The follow-up on #1036 asked for the probes to be read per *egress channel*
+rather than per tool. Named on that axis, the surface is:
+
+1. **JSON response** — layer (f), driven off ``_RUNTIME_PROBES`` /
+   ``_PROBE_EXEMPT``. Forced: a newly ``GATED`` tool must grow a probe or
+   record a justified exemption.
+2. **Model prompt** — layer (g), driven off ``_PROMPT_PROBES`` /
+   ``_PROMPT_PROBE_EXEMPT``, over a set derived from the tools' own
+   signatures. Forced the same way, plus a non-emptiness assertion on the
+   derivation, since a derived set can silently empty itself.
+3. **Disk artifacts** — covered per tool, and **not forced**.
+   ``test_report_probe_leaves_no_canary_in_the_artifact_it_writes``,
+   ``test_state_render_probe_leaves_no_canary_in_the_artifact_it_writes``,
+   ``test_journal_probe_refuses_and_leaves_the_fragment_bytes_untouched`` and
+   ``test_upload_probe_refuses_and_leaves_the_staged_document_untouched`` are
+   four good ideas, and nothing obliges the fifth artifact-writing tool to
+   grow a fifth. That gap is deliberately **out of scope here** and split to
+   **#1273** rather than left implied: #968 and #969 were both found on this
+   channel, so it is simultaneously the channel with the worst track record
+   and the only one with no forcing function.
+
+**Scope item 4 of #1036 — a second gate pair on :class:`ToolPosture` — is
+decided NO.** ``creek.reflect``'s rationale names two gates while the
+dataclass records one, so the obvious repair is a second
+``(gate_module, gate_symbol)`` pair. It was not taken. A second pair is only
+ever checkable at the strength layers (c)/(e) can offer — the symbol exists in
+the named module and is called there — and for reflect's grounding gate that
+check is green *by construction*: ``creek/author/agents.py`` imports
+``tier_within_override`` at line 28 and calls it in ``_load_corpus`` at line
+115 for every Writing Desk consumer. It would stay green if
+``creek_mcp.tools.reflect._default_retrieve`` stopped passing ``override``,
+stopped being called, or reflect dropped grounding altogether — three ways to
+lose the gate entirely without disturbing the evidence for it. Layer (g)
+checks that gate's *effect* instead, which is strictly stronger, and the
+injection drill at the top of ``tests/test_mcp_read_gate.py`` is the evidence:
+the two independent mutations that neutralise the grounding cutoff turn (g)
+red and leave every structural layer and all ten response probes green.
+Revisit predicate: **a tool whose second gate's effect no runtime probe can
+observe.** For that tool the structural check is the only one available, and a
+weak check is better than none.
+
+**``creek.classify`` is a known prompt-channel egress outside both probe
+manifests.** Its posture is ``METADATA_ONLY`` on a rationale about its
+*response* — "Returns counts only … the ceiling is audited for the trail, not
+enforced" — and ``creek_mcp/tools/classify.py`` says the same in its own
+words: the ceiling "is recorded for the audit trail; it does not gate
+execution". Both are true of the envelope and beside the point on this
+channel. With ``method="llm"`` the whole corpus, every tier of it, goes
+through a provider, defended only by
+:class:`creek.classify.llm.router.ModelRouter`'s Intimate-never-cloud gate —
+which ``creek/classify/classify_engine.py`` reaches by resolving a second,
+intimate-only classifier config, and which says nothing about ``personal``.
+Layer (g) structurally cannot enrol it: the derivation looks for a
+function taking both ``llm_factory`` and ``privacy_tier_ceiling``, and
+``classify_tool`` takes no factory — so the reason this tool is absent from
+the prompt manifest is a fact about its signature, not a finding about its
+safety. Tracked in **#1274**.
 """
 
 from __future__ import annotations
--- a/creek-tools/docs/mcp.md
+++ b/creek-tools/docs/mcp.md
@@ -263,7 +263,28 @@ Every registered tool's relationship to the caller's ceiling is now a
 machine-checked audit record: `creek_mcp/read_gate.py`'s `TOOL_POSTURES`
 names one of six postures per tool, verified against the live server
 surface and, for `GATED` entries, against a real call site in the
-named module, and for the rest against a runtime canary probe.
+named module. Those claims are then checked by *running* the tools —
+not with one canary probe, but with **one probe set per egress
+channel**, because a tool has more than one way to put content in front
+of you and closing one channel says nothing about the others:
+
+| Egress channel | What is checked | Which tools | Forced? |
+|----------------|-----------------|-------------|---------|
+| **JSON response** | The tool is called at `ceiling=open` against a vault holding an `intimate` fragment whose title, body and tags each carry a unique sentinel; that sentinel must appear nowhere in the serialised response. | Ten of the twelve `GATED` tools. `creek.draft` and `creek.author` carry recorded exemptions instead — a bare fixture does not drive either far enough for its envelope to mean anything. | Yes. A newly `GATED` tool must grow a probe or record a justified exemption; it cannot inherit silence. |
+| **Model prompt** | The tool is driven all the way *to* the provider with a recording client, and neither the `intimate` sentinel nor the above-ceiling **`personal`** one may appear in any prompt it sent — nor in the response of that same call ([#1036](https://github.com/Geoffe-Ga/Creek-Vault/issues/1036)). | `creek.reflect`, `creek.compile` and `creek.author`. `creek.draft` is exempt, and its stated reason is executed rather than trusted. | Yes. The set is *derived* from the tools' own signatures — a `GATED` tool taking both an LLM factory and a ceiling — so a tool cannot become model-backed without being asked for a prompt probe. |
+| **Disk artifacts** | The bytes the call writes into the vault carry no sentinel, and a refused call leaves the bytes it would have overwritten untouched. | `creek.report`, `creek.state.render`, `creek.journal`, `creek.upload`. | **No.** Four per-tool tests and no forcing function: nothing obliges the fifth artifact-writing tool to grow a fifth test. Tracked in [#1273](https://github.com/Geoffe-Ga/Creek-Vault/issues/1273). |
+
+The prompt channel is the one you cannot check from the outside. A
+response you can inspect; a prompt has already crossed to the provider
+by the time the response exists, so a tool could hand you a perfectly
+clean envelope having shipped an above-ceiling fragment to a cloud
+endpoint on the way to building it. Note which of that row's two
+sentinels does the work, too: the `personal` one. Every path in this
+codebase drops `intimate`, but the privacy filter *summarises* a
+personal fragment to
+`[Personal-tier summary: <title>]` — so on that path an above-ceiling
+**title** reaches the model by design, and it is exactly what a weaker
+"no intimate content in the prompt" check would never see.
 
 Most of what the sweep found is **write-side**: a tool *acts on* content
 above the caller's ceiling (#970, #971) rather than handing it
--- a/creek-tools/tests/test_mcp_read_gate.py
+++ b/creek-tools/tests/test_mcp_read_gate.py
@@ -10,7 +10,7 @@ primitives (:func:`~creek_mcp.read_gate.refuse_above_ceiling` and
 ways a tool may satisfy the ceiling, so gaps can be closed by adoption rather
 than by re-deriving the policy per tool.
 
-A manifest is only worth having if it cannot lie. Six layers keep it honest:
+A manifest is only worth having if it cannot lie. Seven layers keep it honest:
 
 (a) **Surface completeness** — the manifest covers exactly the live tool list,
     derived from ``server.list_tools()`` rather than a second hardcoded copy,
@@ -34,6 +34,24 @@ A manifest is only worth having if it cannot lie. Six layers keep it honest:
     its own (:data:`_RUNTIME_PROBES` / :data:`_PROBE_EXEMPT`) so a newly
     ``GATED`` tool must either grow a probe or record a justified exemption —
     the same forcing function as layer (a), one level deeper.
+(g) **Prompt-channel canary probe** — layer (f) at the other end of the wire.
+    Every ``GATED`` tool that can hand corpus text to a model is driven all
+    the way *to* the provider with a recording factory, and the gated
+    sentinels must appear in none of the prompts it sent — nor in the
+    response of that same call. Two positive controls run first: a prompt
+    was captured, and the *admitted* canary is in it. A probe that reached
+    no model, or built a prompt with no corpus in it, would satisfy every
+    exclusion while proving nothing. The subject set is **derived** rather
+    than listed (:data:`_LLM_BACKED_GATED_TOOLS` — every
+    ``GATED`` tool whose own module defines a function taking both
+    ``llm_factory`` and ``privacy_tier_ceiling``; four of them today), so a
+    tool cannot *become* LLM-backed without being asked for a probe. Forced
+    by a manifest pair of its own (:data:`_PROMPT_PROBES` /
+    :data:`_PROMPT_PROBE_EXEMPT`) exactly as (f) is, plus one assertion (f)
+    does not need: the derived set must be **non-empty**. A posture somebody
+    declared cannot quietly empty itself; a set computed by signature
+    introspection can, on nothing worse than a parameter rename, and would
+    take the whole layer green with it.
 
 Layer (f) exists because layers (a)-(e) are **structural**. Between them they
 prove that a gate is declared in the manifest, exists in the named module, and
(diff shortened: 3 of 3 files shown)


Geoffe-Ga:
## Summary
Adds a seventh verification layer (g) to `tests/test_mcp_read_gate.py` that drives every LLM-backed `GATED` MCP tool (`creek.author`, `creek.compile`, `creek.draft`, `creek.reflect`) all the way to a recording model factory and asserts no above-ceiling canary appears in the *prompt*, closing a real blind spot: layers (a)-(f) all terminate at the response envelope, which cannot see what was already shipped to a cloud provider before that envelope existed. The subject set is derived by signature introspection rather than hand-listed, is pinned, and is forced non-empty so the layer cannot silently go vacuous. This is a test-only + docstring change; `creek_mcp/read_gate.py`'s diff is documentation alone.

## Strengths
- The forcing-function design (derive `_LLM_BACKED_GATED_TOOLS` from signatures, require every member to be probed-or-exempt, assert non-emptiness, pin both sets) is a genuinely strong mechanism against the exact failure mode it's designed to prevent: a new LLM-backed tool silently inheriting "unchecked" status.
- I spot-checked the PR's specific technical claims against the actual source and they hold up: `reflect_tool`'s `content=` vs `entry_ref=` distinction and the #846 gate ordering (creek_mcp/tools/reflect.py:344-384), `compile_tool`'s `_survey_sources`/`_ABOVE_CEILING_REASON` refusal-before-factory ordering (creek_mcp/tools/compile.py:142,191), `draft_tool`'s signature and the `UNEXPLORED_ONTOLOGY` seed producing no `source_fragments` (creek/generate/mining.py:1597-1609), and `AuthorLLMClient.complete_with_usage`'s exact parameter shape matching `_RecordingAuthorFactory.complete` (creek/author/client.py:193-220). Nothing I checked was misrepresented.
- Two independent mutation injections (neutralizing `tier_within_override` and forcing `to_privacy_override` to `ALL`) are documented as turning exactly layer (g) red while layer (f) and all ten response probes stay green — this is the right kind of evidence for a detection-mechanism test, not just a passing assertion.
- Self-critical where the probes are weak: `_prompt_probe_compile`'s docstring explicitly states its two exclusion assertions are weak for that tool (because both above-ceiling paths refuse before the factory is ever built) rather than letting the green result imply a stronger guarantee than it has. The `creek.draft` exemption is executed (`test_the_draft_prompt_exemption_is_still_true`), not just asserted by AST presence.
- Found and reported (rather than silently fixed) that `creek.author`'s existing layer-(f) exemption reason is false, filing it as a separate follow-up (#1279) instead of scope-creeping a production-adjacent fix into a test PR.
- Named gaps are tracked as real issues rather than left implied (#1273 disk-artifact channel, #1274 `creek.classify`'s prompt egress, #1275 recorder consolidation, #1287 ancestry-shaped fixture gap).

## Security Concerns
🟢 LOW — This PR is a detection-only test addition; it introduces no new runtime behavior and cannot itself create a vulnerability. It does surface (and track via issues, not fix inline) two real gaps: `creek.classify`'s `method="llm"` path is structurally unreachable by this layer's derivation (#1274) and the disk-artifact egress channel remains unforced (#1273). Neither is a regression introduced by this PR, and scoping them out with tracked follow-ups rather than fixing them here is a reasonable and clearly-justified choice for a single-purpose PR.

## Problems
None found that block merge.

## Code Quality
- `_RecordingLLMFactory` and `_RecordingAuthorFactory` are two of what the PR itself acknowledges will eventually be five near-identical recorders across the test suite (tracked as #1275, explicitly deferred as "the deliberate minimal-diff call for #1036"). This is a reasonable call for this PR's scope, but worth the author/team keeping an eye on so #1275 doesn't slip.
- The final `assert (probed | exempt) == backed` in `test_every_llm_backed_gated_tool_is_prompt_probed_or_exempt` is logically implied by the preceding `unchecked`/`stale` assertions — harmless redundancy used as explicit belt-and-suspenders, not a real issue.
- The docstrings are extremely dense (some run 30-40 lines per function). This is consistent with the rest of this file's existing style and clearly serves the "manifest that cannot lie" auditability goal, but it's worth noting as a readability/maintenance cost for future contributors to this file.

## CLAUDE.md Adherence
✅ Commit is a single, conventional-commit-styled message (`test(mcp): ...`).
✅ All new functions/classes are fully typed and carry Google-style docstrings (interrogate ≥95% should hold given the density here).
✅ No production behavior changed — `read_gate.py`'s diff is docstring-only, consistent with the PR's own claim (verified structurally: the diff only touches the module docstring block).
✅ TDD workflow followed and documented: the PR body includes the verbatim RED failure and a mutation-testing proof, going beyond what CLAUDE.md's "Stay Green" gate requires.
⚠️ Could not independently execute `./scripts/check-all.sh`, `./scripts/test.sh`, or CI status from this review environment (tool approvals for `pytest`/`gh run list`/`gh api` were unavailable here), so green CI is not independently confirmed by this review — recommend confirming CI is green before merge if not already verified.

## Testing
Strong. This PR *is* a testing change, and its quality is unusually high: derived (not hand-listed) subject set, non-emptiness guard against silent vacuity, pinned membership on both the subject set and the exemption set (with the pin's docstring correctly identifying that the exemption set is the more dangerous direction to let drift), a shared assertion helper (`_assert_prompt_channel_clean`) so the three probes can't drift into checking different things, positive controls before exclusions (so a broken tier filter can't pass by omission), and an executed rather than trusted exemption reason for `creek.draft`. The included mutation-testing evidence (asymmetric red/green under two independent injections) is exactly the right way to demonstrate a detection mechanism actually detects.

## Documentation
Excellent. `docs/mcp.md` gains a clear three-channel table (response/prompt/disk) with forced-vs-not status per channel, and correctly calls out that the `personal`-tier summary residue (not `intimate`) is the sentinel that does the real work in this layer. `read_gate.py`'s docstring update names the same three channels and explicitly records the "scope item 4 decided NO" rationale with a revisit predicate, closing out the issue's follow-up comment in writing as claimed.



## Verdict: LGTM

This is a test-only, extensively self-verified PR that I cross-checked against the actual production source (reflect_tool, compile_tool, draft_tool, AuthorLLMClient, MiningStrategy) and found no factual misrepresentation or logical gap in the new forcing function; known gaps (creek.classify prompt egress, disk-artifact channel) are honestly surfaced and tracked as separate issues rather than silently left implied. I could not execute the test suite or check live CI status in this environment, but nothing in the diff or its cross-references with the codebase suggests a blocking defect, so I'd merge this as-is.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234848359 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Adds layer (g) to `tests/test_mcp_read_gate.py`: a prompt-channel canary probe that drives every LLM-backed `GATED` MCP tool (`creek.reflect`, `creek.compile`, `creek.author`, with an executed exemption for `creek.draft`) all the way to a recording model factory and asserts no above-ceiling sentinel (including a new `personal`-tier one) crosses in the prompt. The subject set is derived by signature introspection rather than hand-listed, so a future LLM-backed tool can't silently inherit "response-probed, prompt-unchecked." `creek_mcp/read_gate.py`'s diff is docstring-only (verified: no non-docstring code changed), and `docs/mcp.md` gains a three-egress-channel table.

## Strengths
- **Real RED proof and real mutation evidence**, not just a passing test: the PR body shows the actual pre-implementation failure, and two independent injections (`tier_within_override` forced `True`; `to_privacy_override` widened to `ALL`) turn layer (g) red while all ten layer-(f) response probes stay green — the asymmetry that is the entire argument for this layer.
- **Non-emptiness guard on the derived set** (`test_every_llm_backed_gated_tool_is_prompt_probed_or_exempt`) — a signature-derived forcing function can silently empty itself (e.g. on a parameter rename) in a way a human-declared manifest can't, and the test explicitly guards against that failure mode going vacuously green.
- **The `personal`-tier canary is the sharpest part of this change.** It targets a residue shape this repo already knows about (`filter_fragments_by_tier` summarizes personal to `[Personal-tier summary: {title}]`, so the title survives at `ceiling=open`) rather than only re-testing the free `intimate` exclusion.
- **Honest scope discipline.** Weaknesses are stated rather than hidden: compile's exclusion assertions are called out as weak in their own docstring, `creek.draft`'s exemption is *executed* (not just asserted to exist), and five separate items (#1273, #1274, #1275, #1279, #1287) are filed rather than scope-crept into this PR.
- I independently verified the reflect/compile/author/draft probe calls (`vault_path`, `llm_factory`, `content`, `privacy_tier_ceiling`, `care_guard`, `dry_run`, etc.) against the actual current signatures in `creek_mcp/tools/{reflect,compile,author,draft}.py` — they match exactly.
- I independently verified `Callable` and `ModuleType`, used in the new code, resolve to the file's existing `TYPE_CHECKING`-guarded imports (lines 185/187) rather than being missing imports.
- I independently verified the new canary constants (`_RUNTIME_PERSONAL_CANARY`, `_RUNTIME_PERSONAL_ID`) don't collide with the pre-existing open/intimate ones.
- Case-insensitive sentinel matching closes an evasion route (a renderer that lower-cases/slugifies a title on its way into the prompt).

## Security Concerns
🟢 LOW — none found. This is a detection-layer addition, not a change to any gate's enforcement behavior; `creek_mcp/read_gate.py`'s only diff is documentation. The PR itself surfaces one real (pre-existing) finding — `creek.author`'s layer-(f) exemption reason is demonstrably false — but correctly scopes the fix out to #1279 rather than bundling an unreviewed production-adjacent change into a test-only PR.

## Problems
None found that block merge.

## Code Quality
Non-blocking, and mostly already acknowledged in the diff itself:
- `tests/test_mcp_read_gate.py` is now a very large single file (the new block alone starts past line 2500). The PR's own follow-up #1275 (folding five near-identical recorder classes into a shared pair) is the right-sized fix for the duplication driving that growth; no new action needed here.
- The docstrings in this PR are extremely dense (some run 30+ lines per function). That's a stylistic maximalism this repo has clearly already normalized elsewhere in the file, so not out of step with local convention, but worth naming as a general readability tradeoff for future authors extending this pattern.

## CLAUDE.md Adherence
✅ TDD — RED proof shown verbatim in the PR description before the GREEN implementation.
✅ Coverage — 94.36% branch coverage claimed (≥90% floor), per-file gate reported passing.
✅ Docstring coverage — interrogate reported clean; every new public symbol is documented.
✅ Types — mypy strict reported clean; no `type: ignore` added.
✅ Linting — ruff/bandit/pip-audit/xenon/pylint reported clean; no new `# noqa`.
✅ No shortcuts — no threshold edits, skips, or bypasses introduced.
✅ Commit conventions — `test(mcp): ...` follows Conventional Commits and correctly closes #1036.
✅ Root-relative scripts — `./scripts/check-all.sh` used per the PR's test plan, not ad hoc tool invocations.

## Testing
Strong. Beyond the new assertions themselves, the PR demonstrates the tests are non-vacuous via two independent mutation injections and documents that `git diff` over the test file changes exactly one pre-existing line outside the new block — i.e., purely additive, nothing weakened or deleted.

## Documentation
Thorough. `read_gate.py`'s docstring and `docs/mcp.md`'s new egress-channel table clearly name what layer (g) checks, what it doesn't (the disk-artifact channel, explicitly marked unforced and tracked), and why scope item 4 from the original issue was decided NO with a stated revisit predicate.



## Verdict: LGTM

No production behavior changed (verified: the read_gate.py diff is docstring-only, and the rest of the diff is new/additive tests), the new probes' call signatures were checked against the actual production functions (reflect_tool/compile_tool/author_tool/draft_tool) and match exactly, all "new" symbols used (Callable, ModuleType) resolve to pre-existing TYPE_CHECKING imports, and the new canary sentinel/fragment IDs don't collide with existing ones. The PR is unusually self-critical: every weak spot I would otherwise flag (recorder duplication, compile's weak exclusion assertions, the artifact-channel gap, author's stale layer-(f) exemption) is already named in the diff's own docstrings with a filed follow-up issue, so there's no new substantive suggestion left to raise.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234958821 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1291, opened 2026-08-10
fix(vault): non-string frontmatter ids are invisible to _rebuild_index, allowing duplicate-id files


Geoffe-Ga:
`VaultWriter._rebuild_index` (`creek/vault/writer.py`) indexes a scanned file only when its frontmatter `id` is a `str`:

```python
mid = post.get("id")
if isinstance(mid, str):
    index[mid] = md_file.name
```

A fragment file whose frontmatter says `id: 12345` parses as a YAML **int**, so it is permanently invisible to the index — and to the `#1083` re-resolution path that now depends on the same scan.

## Observable consequence

Pinned today by `tests/test_vault_writer.py::TestIdIndexVerification::test_non_str_id_in_frontmatter_is_a_mismatch`, which asserts the honest outcome rather than a nicer one: with an on-disk file declaring `id: 12345` and a model whose id is the string `"12345"`, the writer declines the foreign-typed file (correct — it cannot verify ownership) but re-resolution cannot find it either, so a **second** file is minted and two files end up carrying the same logical id.

This is data hygiene, not data loss — nothing is clobbered, moved or unlinked, and the `#1083` behaviour is strictly safer than the pre-fix clobber. But it is a real way a vault can end up with duplicate-id files.

## Why it was not fixed in #1083

`_rebuild_index`'s str-only filter is pre-existing and untouched by that diff; changing which ids the scanner indexes is a behaviour change to every index consumer, well outside the minimal scope of a data-loss fix. The reviewer and I agreed it should be documented there and fixed here.

## Proposal

Decide deliberately between:

- **(a) Normalise on read** — coerce scalar ids to `str` in `_rebuild_index` (and correspondingly in `_file_declares_id`), so `id: 12345` and `"12345"` are the same id. Fixes the duplicate, but silently widens what counts as a match — needs care that it cannot make two genuinely different ids collide.
- **(b) Keep them distinct and surface it** — leave the type strict, but have `creek lint` report frontmatter ids that are not strings so an operator can repair the file. Safer; does not auto-merge identities the vault never said were the same.

(b) is the more conservative default given the direction-of-safety rule that governs this code.

## Acceptance criteria

- [ ] An explicit decision recorded between (a) and (b), with the reasoning.
- [ ] `test_non_str_id_in_frontmatter_is_a_mismatch` updated to assert the chosen outcome (it currently documents the duplicate as expected).
- [ ] No regression in any `#1083` verification test.
[...]

Issue #1290, opened 2026-08-10
perf(vault): bounded byte-scan for the id-index verification hot path


Geoffe-Ga:
`#1083` closed a data-loss defect by verifying that a file located through the per-directory `.id-index.jsonl` actually declares the id the index claimed. The verification is one `frontmatter.load` per index **hit** (zero per miss), in `_file_declares_id` (`creek/vault/writer.py`).

## Measured cost

Call-counting a real unchanged re-ingest through `_run_ingest`:

| variant | `frontmatter.load` calls / 400 units |
|---|---|
| before #1083 | 0 |
| after #1083 | 400 (exactly one per index hit) |

One load costs **~180 us** on a real fragment (1.6 KB, ~20 frontmatter keys; min-of-9 x 200 iterations). That is roughly **+8%** on the steady-state unchanged re-ingest and about **+6 s** on a full 35 000-fragment sweep.

## Why it was not optimised in #1083

Two cheaper designs were measured and rejected there:

- Hand-rolling a parse of just the frontmatter block is **slower** — `frontmatter.load` resolves libyaml's C loader, while `yaml.safe_load` of the same block runs the pure-Python loader.
- An mtime/size memo buys nothing: a re-ingest visits each path exactly once in a cold process, so the memo always misses — and memoising the verification reintroduces "trust a cache instead of the artifact" one level down, which is the defect #1083 closed.

## Proposal

Verification needs exactly one key (`id`), not the whole frontmatter. A bounded byte-scan could confirm the common case cheaply:

1. Read only the leading bytes (say 8 KB), require a leading `---` fence, and locate the closing fence within the window.
2. Inside that block, match a **column-0** `id:` line (nested keys are indented, so a column-0 match cannot come from a nested mapping, and YAML block-scalar content is always indented — so it cannot come from a body either).
3. On a confirmed equal value, return `True`.
4. On anything unconfirmed — no fence, no window, ambiguous quoting, any doubt — **fall back to the full `frontmatter.load`**, which stays authoritative.

Correctness argument: only a **false positive** is dangerous (it would re-open #1083). A false negative merely falls through to the authoritative parse, so the fast path can be as conservative as needed at zero correctness risk.

## Acceptance criteria

- [ ] Steady-state verification cost materially below the current ~180 us/hit, measured by the same call-count + min-of-N method (do not trust wall-clock A/B on a loaded machine — the unaffected create pass varied 2.5x between runs while measuring #1083).
[...]

PR #1295, opened 2026-08-10
fix(vault): verify the located file declares the id before rewriting it

### LINKED ISSUE #1083: VaultWriter.update_fragment rewrites the id-index's target without verifying the file's own id

Geoffe-Ga:
## Summary

- `VaultWriter._find_existing_locked` resolved an id through the per-directory `.id-index.jsonl` and returned `target_dir / index[model_id]` after checking only `candidate.exists()` — it never confirmed the located file's own frontmatter declares that id. A stale or poisoned entry therefore redirected **four** destructive callers onto a foreign file, not just the `update_fragment` case the issue names: `update_fragment` clobbers its body, `_write_model`'s dup-check returns it as "already written" (silently dropping a brand-new fragment so no file is ever created), and `tomb_fragment` / `restore_fragment` `_atomic_write_text` it to a new directory and then `unlink()` it — **move-and-delete of a stranger's fragment**, the worst case and the one the issue never mentions.
- The fix verifies the artifact instead of trusting the index: a new module-level `_file_declares_id` predicate, applied on the exists-True side of the locator only. On mismatch, `_repair_index_locked` re-resolves via `_rebuild_index`, rewrites the correct file, and persists the corrected mapping — so recovery cannot mint a *second* file for the same id, and the scan is paid once rather than per call. Fixing the shared locator fixes all four callers with zero edits of their own.
- The `candidate.exists()` fast path and its vanished-entry branch are deliberately untouched, so an ordinary tombed/deleted fragment still costs one `stat` and never triggers a directory rescan (`test_vanished_file_does_not_trigger_a_rescan` pins this). The repair appends via `_append_index_entry` (`O_APPEND`, later-wins) and never `_persist_full_index`, whose whole-file `os.replace` from one process's in-memory snapshot would destroy entries another process appended concurrently.

## Test plan

Gate 1 RED, proven behaviourally against unmodified source before any production edit (no `git stash` — four Ralph worktrees share one stash ref):

```
FAILED test_update_fragment_does_not_clobber_foreign_id_file
  assert victim_path.read_bytes() == before
E assert b"---\naudien...session notes" == b"---\naudien...\nvictim body"
E   At index 913 diff: b't' != b'v'

FAILED test_tomb_fragment_does_not_move_and_unlink_foreign_file
  assert result is None
E AssertionError: assert PosixPath('.../10-Liminal/Orphaned/2025-01-15-Victim.md') is None
```

The second is the data-loss headline: tombing a **ghost id that no fragment owns** relocated an `INTIMATE`-tier bystander into `Orphaned/` and unlinked the original.

11 new tests (10 in `TestIdIndexVerification`, 1 end-to-end through the real `_run_ingest` seam): clobber, re-resolve-to-the-real-file, repair-persisted-to-a-fresh-writer, dup-check-must-not-return-a-foreign-file, tomb, restore, unparseable YAML, no `id` key, non-`str` id, the no-rescan guard, and `test_stale_index_edit_is_not_lost` (the pipeline's `updated is None` fallback at `pipeline.py:253-258` must actually re-create the file, or the edit is silently lost).

**The guard is load-bearing:** reverting just the two guard lines fails exactly those 10 tests and no others (10 failed / 109 passed); restoring makes all 119 pass.

**Cross-check (#970):** `tests/test_mcp_journal.py` + `tests/test_api_journal.py` — 62 passed. The overwrite gate resolves tiers by a frontmatter walk of the file that actually carries the id; the writer's locator now uses the same file-declared-id semantics, so gate and writer agree rather than diverging.

**Gate 2:** `./scripts/check-all.sh` exit 0 — 12 checks passed / 0 failed, 8741 passed, 94.38% aggregate coverage, per-file gate clean (`creek/vault/writer.py` 91.75%, unwaived). No new waiver entry, no `# noqa`, no `# type: ignore`, no skip marker, no threshold change. Xenon: `_find_existing_locked` A(4), `_file_declares_id` A(3), `_repair_index_locked` A(2) — all under the B ceiling.

### Performance: measured, not asserted

Wall-clock A/B was unusable on this machine (the *unaffected* create pass varied 2.5x between runs), so I counted operations instead, which is deterministic:

| Variant | `frontmatter.load` calls during an unchanged re-ingest of 400 units |
|---|---|
| before | **0** |
| after | **400 — exactly one per index hit, zero per miss** |

One load costs **~180 µs** on a real fragment (1.6 KB, ~20 frontmatter keys; min-of-9 × 200 iterations). So the added cost is exactly one load per index hit: roughly **+8%** on the steady-state unchanged re-ingest, about **+6 s** on a full 35 000-fragment sweep. No directory scan enters the steady-state path.

Two cheaper designs were measured and rejected:

- **Hand-rolling a parse of just the frontmatter block was *slower*** — `frontmatter.load` resolves libyaml's C loader, while `yaml.safe_load` of the same block runs the pure-Python loader and measured several times slower.
- **An mtime/size memo buys nothing**: a re-ingest visits each path exactly once in a cold process, so the memo always misses — and memoising the verification reintroduces "trust a cache instead of the artifact" one level down, which is the exact defect being closed.

Driving the remaining cost down needs a bounded byte-scan for the single `id` key; that is filed as a follow-up rather than smuggled into a data-loss fix.

### One honest limitation

`test_non_str_id_in_frontmatter_is_a_mismatch` asserts that a file declaring `id: 12345` (a YAML int) against model id `"12345"` leaves **two** files carrying the same logical id. `_rebuild_index` only indexes ids where `isinstance(mid, str)`, so the int-typed file is invisible to re-resolution and cannot be deduplicated by this fix. Declining to write into a foreign file is still strictly safer than the current clobber, so this is asserted as-is rather than smoothed over.

Privacy: strictly more restrictive, nothing widens at any ceiling. Every branch either targets the file that provably owns the id or refuses.

Closes #1083
Refs #970

Follow-ups filed: #1290 (bounded byte-scan to cut the verification cost), #1291 (non-string frontmatter ids are invisible to `_rebuild_index`). Both P3.

### DIFF
--- a/creek-tools/creek/vault/writer.py
+++ b/creek-tools/creek/vault/writer.py
@@ -415,6 +415,29 @@ def _atomic_write_text(path: Path, content: str) -> None:
                 tmp_path.unlink()
 
 
+def _file_declares_id(path: Path, model_id: str) -> bool:
+    """Return ``True`` only when *path*'s own frontmatter declares *model_id*.
+
+    The verifier behind the id index (#1083): an index entry is a *claim*
+    about a file, and this is the only thing that turns that claim into
+    evidence. A missing ``id`` key, a non-``str`` id, and an unparseable
+    file all answer ``False``.
+
+    The guarded exception set and the ``isinstance(..., str)`` gate are
+    deliberately identical to the ones
+    :meth:`VaultWriter._rebuild_index` applies while scanning, so the
+    verifier and the scanner can never disagree about which files
+    declare an id — a file this function rejects is exactly a file the
+    rebuild would decline to index.
+    """
+    try:
+        post = frontmatter.load(str(path))
+    except (OSError, ValueError, yaml.YAMLError):
+        return False
+    declared = post.get("id")
+    return isinstance(declared, str) and declared == model_id
+
+
 def _is_material_change(old_body: str, new_body: str, threshold: float) -> bool:
     """Return ``True`` when an edit changed the body materially (#675).
 
@@ -784,7 +807,10 @@ class VaultWriter:
         already on disk, so it never lowers an operator's decision.
 
         Returns ``None`` when no file is mapped to ``fragment.id`` (e.g. the
-        file was removed out of band), so the caller can fall back to a fresh
+        file was removed out of band), and also when the mapped file exists
+        but declares a *different* id and no live file in the directory
+        declares this one (#1083) — a rewrite there would clobber a foreign
+        fragment. Either way the caller can fall back to a fresh
         :meth:`write_fragment`.
 
         Args:
@@ -798,7 +824,9 @@ class VaultWriter:
 
         Returns:
             Path to the rewritten file, or ``None`` when no existing file
-            maps to the id.
+            maps to the id — including the case where the mapped file
+            exists but declares a different id and re-resolution finds no
+            live file declaring this one.
         """
         target_dir = self._fragment_target_dir(fragment)
         with self._lock:
@@ -1064,7 +1092,10 @@ class VaultWriter:
         # ``_find_existing_locked`` and ``_atomic_create`` that BUG-006
         # was filed to close. The throughput trade-off is documented
         # in the PR description; revisit only with a benchmark-driven
-        # follow-up issue.
+        # follow-up issue. The id verification and any index repair
+        # ``_find_existing_locked`` performs (#1083) run inside this same
+        # critical section, so a repaired mapping cannot be raced by a
+        # concurrent write between the re-resolution and the file creation.
         with self._lock:
             existing = self._find_existing_locked(model_id, target_dir)
             if existing is not None:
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
## Summary
`VaultWriter._find_existing_locked` previously trusted the per-directory `.id-index.jsonl` after only a `candidate.exists()` check, so a stale or poisoned index entry could redirect `update_fragment`, `_write_model`'s dup-check, and `tomb_fragment`/`restore_fragment` onto a file belonging to a *different* fragment — in the worst case, moving and unlinking a bystander's file. This PR adds a `_file_declares_id` verification step and a `_repair_index_locked` re-resolution path, fixing the shared locator (and therefore all four callers) in one place.

## Strengths
- Fixes the bug at its single choke point (`_find_existing_locked`) rather than patching each of the four callers separately — verified by reading `update_fragment`, `_write_model`, `tomb_fragment`, and `restore_fragment`; all route through it (directly or via `_find_in_fragments_locked`).
- `_file_declares_id`'s exception set and `isinstance(..., str)` gate are deliberately identical to `_rebuild_index`'s — confirmed by inspection, so the verifier and the directory scanner can never disagree about what counts as a match.
- Repair persistence correctly uses `_append_index_entry` (O_APPEND, later-wins) rather than `_persist_full_index`, avoiding destroying concurrent cross-process appends — consistent with the existing index-file design documented elsewhere in the module.
- The fast path (`candidate.exists()` → verified match) is unchanged in cost profile; the expensive rescan only fires on an actual mismatch, and `test_vanished_file_does_not_trigger_a_rescan` pins the everyday deleted/tombed case to zero rescans.
- The PR description is unusually honest about a known gap (non-string YAML ids are invisible to `_rebuild_index`, so an int-typed id can't be deduplicated) and files it as a separate follow-up (#1291) instead of hiding it.
- 11 new tests exercise all four destructive callers plus corruption edge cases (unparseable YAML, missing `id` key, non-str id) and an end-to-end pipeline-level regression test (`test_stale_index_edit_is_not_lost`) that traces through the real `updated is None` fallback in `creek/ingest/pipeline.py:256-258` — I confirmed that fallback exists and behaves as described.

## Security Concerns
None identified. 🟢 This is itself a data-integrity/security-adjacent fix (prevents a stale index from silently destroying an unrelated fragment's content) and doesn't introduce new attack surface — `target_dir`/`filename` provenance is unchanged from the existing (trusted local vault) threat model.

## Problems
None found blocking merge.

## Code Quality
- Minor: in `_repair_index_locked`, when `resolved is None` (a poisoned entry pointing at an existing file, but no live file anywhere actually owns the id — the "ghost id" case), the in-memory index entry is deleted but nothing is appended on disk to correct it. A fresh `VaultWriter` process will re-load the same poisoned mapping from the JSONL and pay the full rescan again on next access. This is a performance-only nit (correctness is preserved — `_file_declares_id` still rejects the foreign file each time), and the docstring's "other stale ids self-heal on first access" claim is only exactly true for the case where a real owner is found. Not worth blocking on given this is an already-rare corruption path with a documented cost model.

## CLAUDE.md Adherence
✅ TDD followed (PR description documents RED tests captured against unmodified source before any production edit). ✅ Commit message uses conventional-commit style (`fix(vault): ...`) with issue references (Closes #1083, Refs #970). ✅ No `# noqa`, `# type: ignore`, or threshold changes. ✅ Complexity kept low per Xenon (documented A(2)-A(4) for new/changed functions). ✅ No ADR needed — this is a bug fix, not an architectural decision. ✅ Extensive Google-style docstrings with Args/Returns/Raises on all new/changed public and private methods, consistent with the repo's existing exhaustive documentation style in this file.

## Testing
Strong. 11 new tests in `TestIdIndexVerification` cover: clobber prevention, correct re-resolution, repair persistence surviving a fresh writer, dup-check no longer returning a foreign file, tomb/restore ghost-id protection, unparseable YAML, missing `id` key, non-`str` id, and a no-rescan regression guard for the ordinary vanished-file path. Plus one pipeline-level integration test proving the edit-recovery fallback actually recreates the fragment rather than losing the edit. I traced the helper functions (`_make_vault`, `_journal_files`, `_ingest`, `vault_path` fixture) and confirmed they exist and are used consistently with the rest of the test suite.

## Documentation
Docstrings are thorough to the point of over-explaining in places (e.g. `_find_existing_locked`'s ~60-line docstring covers cost analysis and two rejected alternative designs), but this matches the pre-existing documentation density already present in this file (e.g. `_append_index_entry`'s docstring), so it's consistent with repo convention rather than a new pattern.



## Verdict: LGTM

The fix correctly closes the described data-loss defect by centralizing verification in the single shared locator (`_find_existing_locked`), which all four destructive callers route through; the repair path is lock-protected, uses append-only (not full-rewrite) persistence to avoid destroying concurrent entries, and reuses the exact same id-declaration predicate the directory scanner already uses. Tests are thorough (11 new cases covering all four callers, corruption edge cases, and a no-rescan regression guard) and I traced the code paths myself to confirm the claims in the PR description hold up. Nothing here blocks merge; the one known limitation (non-string frontmatter ids) is honestly disclosed and filed as a separate follow-up rather than smoothed over.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5234921827 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

PR #1297, opened 2026-08-10
feat(crawdad): audit/purge helper for a pre-#1052 bot-capture tree

### LINKED ISSUE #1264: crawdad: audit/purge helper for pre-#1052 bot-capture data (non-allowlisted + intimate records still on disk and still ingestible)

Geoffe-Ga:
## Summary

#1052 stopped future bad bot-capture writes. It did nothing about what a pre-fix
build already wrote. An operator who ran with `capture_enabled: true` before that
fix has a `<vault>/discord-capture/` tree holding messages from non-allowlisted
users, other bots, and channels declared `intimate` — all untiered, and all still
ingestible: `stage_capture_as_data_package` iterates every channel dir with no
allowlist or tier filter, and the result lands as `unclassified`, which ranks
*with* `personal`.

This adds the offline remediation helper — `crawdad capture audit|purge` — plus
the operator documentation. It does **not** change the live capture gate (#1052
owns that) or the capture file format.

### Premise re-verified

All three `file:line` references in the issue body still hold exactly, despite
#1052/#1152/#1054/#1088 all merging since it was filed:

| Cited | Status |
|---|---|
| `creek-tools/creek/ingest/discord.py:819-831` (`stage_capture_as_data_package` unfiltered walk) | exact |
| `creek-tools/creek/ingest/discord.py:864-889` (`ingest_capture_dir`) | exact |
| `crawdad/crawdad/capture.py:165-173` (`_seen_ids` dedup seeded from stale files) | exact |

### Design

**Audit is the default and read-only. Purge is a dry run unless `--apply`.**
Deletion is irreversible, so nothing goes until the operator has been shown the
target list in the same vocabulary the audit uses.

**The purge unit is a whole channel directory, never an individual record.**
This is the partial-match decision, and it is forced by the on-disk format
rather than chosen for convenience. `capture.py::_record_for` writes `id`,
`timestamp`, `content`, `author.name` and an optional `reference` — no user id
and no channel id. So of the three conditions `bot.py::_capture_allowed` checks,
only the two channel-level ones are re-derivable from disk; `allowed_user_ids`
is not. Filtering records by display name would match on a mutable, spoofable
key and would either destroy legitimate messages or leave the leak while
reporting it fixed. Instead the audit reports the distinct author names per
directory so the operator can see who is in there, and the limitation is printed
in the tool's own output — not just in the docs.

**A name-labelled directory gets no verdict.** `_channel_label` prefers the
channel *name*, and a name cannot be mapped back to an id offline. Guessing is
the one way this helper could delete admitted data, so those dirs are reported
`unresolved` and left alone until the operator names one with `--channel`.

**"All digits" does not mean "is an id."** A directory is read as a channel id
only when it could plausibly be a snowflake: ASCII digits, no leading zero, and
at least 15 of them (Discord ids are `(ms since the 2015 epoch) << 22`). Digit-
only channel *names* are ordinary — `2024`, `420`, `911` — and reading one as an
id misses the allowlist, marks the directory `refused`, and deletes it on a
default `--apply` with no confirmation. The floor errs safe both ways: a very
early-2015 id-labelled dir would fall through to `unresolved`, costing one
`--channel` flag rather than costing data.

| Verdict | Purged? |
|---|---|
| `admitted` — allowlisted channel at an admitted tier | **Never**, even when named explicitly |
| `refused` — id not in `allowed_channel_ids`, or tier `intimate`/`all` | Yes, by default |
| `unresolved` — name-labelled, no id derivable | Only when named with `--channel` |

**Privacy ratchet.** Every path can only make the tree *less* exposed. The worst
outcome of a bug here is that a refused directory survives and the operator
deletes it by hand — the status quo. Nothing widens what is ingestible and
nothing lowers a tier.

`_verdict_for` reads `allowed_channel_ids` and `CAPTURE_ADMITTED_TIERS` directly,
in the same order `_capture_allowed` does, rather than keeping a second copy of
the gate's rules. `apply_purge` validates every target against the resolved
capture root *before* deleting any, so a hand-built plan naming a path outside
the root removes nothing at all.

### Files

All changes are under `crawdad/`. **No creek-tools source was touched**, so only
the crawdad gate applies.

- new `crawdad/crawdad/capture_audit.py` — audit + purge logic
- new `crawdad/tests/test_capture_audit.py` — 41 tests
- `crawdad/crawdad/cli.py` — `crawdad capture audit|purge` subcommands
- `crawdad/crawdad/capture.py` — extracted `iter_channel_records(channel_dir)`;
  `MessageCapture._iter_records` delegates to it so the audit does not grow a
  second JSONL parser. No format or gate change.
- `crawdad/README.md` — the upgrade note (Task 4)
- `crawdad/CLAUDE.md` — §5.2 maintainer note on the two structural limits

## Test plan

`crawdad/scripts/check-all.sh` — 7/7 green (lint, format, mypy strict, bandit +
pip-audit, interrogate, xenon, tests). 736 tests pass; coverage 94.88% against a
90% floor. 47 tests in `test_capture_audit.py`.

Five guards were each reverted in turn to prove they are load-bearing, and each
took down exactly the intended tests and no others:

| Guard reverted | Tests that went red |
|---|---|
| snowflake length floor in `_resolve_channel_id` | 5 — incl. `test_purge_never_auto_targets_an_all_digit_channel_name` |
| `plan_purge` declining an explicitly-named admitted dir | 4 — incl. `test_apply_purge_never_deletes_records_the_gate_would_admit` and the CLI end-to-end |
| `apply_purge` capture-root confinement | 1 — `test_apply_purge_refuses_a_target_outside_the_capture_root` |
| unresolved dirs not auto-targeted | 4 — incl. `test_apply_purge_leaves_an_unrequested_unresolved_directory_alone` |
| CLI dry-run default (`if args.apply`) | 1 — `test_cli_capture_purge_defaults_to_a_dry_run` |

## Gate 2.5

`code-review-orchestrator` returned **FIX_REQUIRED** with one blocker and two
nits. All three are fixed; the blocker was fixed test-first.

**🔴 Blocker — digit-only channel names auto-purged.** `_resolve_channel_id`
treated any all-digit label as a channel id, so a channel *named* `2024` was
misread as id 2024, missed the allowlist, and was classified `refused` — which
`plan_purge` auto-targets with no `--channel` flag and no confirmation. The RED
test failed exactly as predicted before the fix:

```
>       assert (capture_dir / NUMERIC_NAME).is_dir()
E       AssertionError: assert False
E        +    where is_dir = (PosixPath('.../vault/discord-capture') / '2024').is_dir
```

That is silent, irreversible deletion of a legitimate channel's whole capture
history — the one direction this tool must never fail in, and a direct
contradiction of its own stated invariant. Fixed with the snowflake floor above,
plus three regression tests (numeric name, leading zero, and the destructive
end-to-end case). Test fixtures now use realistic 18-digit snowflakes rather than
toy ints, so they exercise the path a real capture tree produces.

**🟡 Nit — symlinks silently skipped.** The audit skipped symlinks and loose
files without saying so, undercutting the report's claim to be sufficient on its
own. Added `list_skipped_entries` and a disclosure line naming them.

**🔵 Nit — undocumented ordering assumption.** Timestamps are compared as strings.
Now documents *why* that is exact (fixed-width UTC prefix; `+` at ASCII 43 sorts
before `.` at 46, so a zero-microsecond stamp still orders correctly) and that a
writer format change would break it silently.

The reviewer separately confirmed clean: `_verdict_for` has no drift from
`bot.py::_capture_allowed`; `apply_purge` confinement holds against traversal,
hand-built plans and TOCTOU (`shutil.rmtree` refuses a symlink argument); the
tests are behavioural throughout; and the `iter_channel_records` extraction is a
pure refactor.

The fixture tree deliberately mixes directories the gate would admit with
directories it would refuse, so every assertion is behavioural — what the audit
*reports* and what the purge *removes*, never that a symbol exists.

Also smoke-tested against a real on-disk tree via `python -m crawdad.cli capture
audit|purge` to confirm the report reads correctly for an operator.

## Acceptance criteria

- [x] An operator can enumerate what a pre-fix capture tree holds without reading
      JSONL by hand — `crawdad capture audit` reports verdict, declared tier,
      record count, date range and distinct authors per channel dir.
- [x] Purge is opt-in and dry-run-by-default; `test_apply_purge_never_deletes_records_the_gate_would_admit`
      pins that it never deletes records the current gate would admit.
- [x] The upgrade note is in `crawdad/README.md`.
- [x] `crawdad/scripts/check-all.sh` passes. `creek-tools/scripts/check-all.sh` is
      not applicable — no creek-tools file was touched.

## Notes

No independent `chief-architect` pass was run: the issue's Tasks section already
fixed the shape and the orchestrator authorised skipping it. The design decisions
above are mine, not an architect's — the partial-match and unresolved-label calls
in particular deserve a reviewer's eye. Gate 2.5 review *was* run independently
and did catch a real blocker, which is some evidence the shortcut was survivable
here, not that it was free.

One gap is documented rather than closed: an `admitted` directory may still hold
messages from users who were never in `allowed_user_ids`, and no purge can
selectively remove them, for the reason given above. The audit surfaces the
author names so the operator can spot them and drop the whole directory.

Closes #1264
Refs #1052

### DIFF
--- a/crawdad/CLAUDE.md
+++ b/crawdad/CLAUDE.md
@@ -198,6 +198,39 @@ tier end-to-end from the capture record through staging into fragment
 frontmatter, so an `intimate` channel can be captured faithfully
 instead of refused outright.
 
+**Remediating a pre-#1052 tree (#1264).** The gate above is
+write-time-only; it never re-examines what an older build already
+wrote, and the creek-tools read side
+(`stage_capture_as_data_package`) applies no allowlist or tier filter
+of its own. `crawdad/crawdad/capture_audit.py` plus
+`crawdad capture audit|purge` is the offline remediation helper —
+audit read-only by default, purge dry-run unless `--apply`. It
+re-derives the gate's verdict from `crawdad.yaml`, so it must never
+grow a second copy of the gate's rules: `_verdict_for` reads
+`allowed_channel_ids` and `CAPTURE_ADMITTED_TIERS` directly, in the
+same order `_capture_allowed` does.
+
+Two limits are structural, not provisional. **Only channel-level
+conditions are re-derivable**: `_record_for` writes no user id and no
+channel id, so `allowed_user_ids` cannot be re-evaluated from disk and
+purge therefore operates on whole channel directories, never on
+individual records. **Only id-labelled directories get a verdict**:
+`_channel_label` prefers the channel *name*, and a name cannot be
+mapped back to an id offline, so name-labelled dirs are reported
+`UNRESOLVED` and left alone until the operator names one with
+`--channel`. Note the trap `_resolve_channel_id` exists to avoid:
+digit-only channel *names* are ordinary (`2024`, `420`, `911`), so
+"all digits" alone must never mean "is an id" — a short numeric name
+read as an id misses the allowlist, becomes `REFUSED`, and is deleted
+by a default `--apply` with no confirmation. Hence the
+`_MIN_SNOWFLAKE_DIGITS` floor. Do not relax it; the failure it
+prevents is silent, irreversible data loss, while its cost when it
+over-triggers is one `--channel` flag. A gate-`ADMITTED` directory is refused even when named
+explicitly — that refusal is what makes "never deletes what the gate
+would admit" a property of the tool rather than of how carefully it
+was invoked. Both limits are stated in the tool's own output; do not
+remove them from the report to tidy it up.
+
 ### 5.3 Redaction-scan gate (FEAT-027, #1054)
 
 **Policy: record the batch, refuse at dispatch.** An attachment turn
--- a/crawdad/README.md
+++ b/crawdad/README.md
@@ -91,14 +91,70 @@ vault, because a captured message currently can't carry its channel's
 privacy tier with it. #1262 tracks carrying that tier end-to-end so
 `intimate` channels can be captured faithfully instead of refused.
 
-**Upgrading from a pre-#1052 build.** If you ran with
-`capture_enabled: true` before this fix, `<vault>/discord-capture/`
-may already hold records from non-allowlisted users, other bots, or
-an `intimate` channel — the old capture path bypassed all three
-gates. These records are still ingestible today. #1264 tracks the
-fix; until then the manual mitigation is deleting the offending
-channel subdirectories under `discord-capture/` before running
-`creek sync`.
+#### Upgrading from a pre-#1052 build — audit your capture tree
+
+#1052 stopped future bad writes; it did nothing about what an older build
+already wrote. If you ever ran with `capture_enabled: true` before that fix,
+`<vault>/discord-capture/` may hold records from non-allowlisted users, other
+bots, or a channel you declared `intimate` — the old capture path bypassed all
+three gates, and those records are **still ingestible**, landing as
+`unclassified`, which ranks with `personal`. Run the audit before your next
+`creek sync`:
+
+```bash
+crawdad capture audit                       # read-only; changes nothing
+```
+
+It reports, for every channel directory under the capture root: the verdict the
+*current* gate would give it, the declared privacy tier, the record count, the
+date range, and the distinct author names. Then:
+
+```bash
+crawdad capture purge                       # DRY RUN — shows what would go
+crawdad capture purge --apply               # actually delete (irreversible)
+```
+
+Both read the same `crawdad.yaml` the bot does, so they judge the tree against
+your live allowlist and tier table.
+
+**What purge deletes, and what it deliberately does not.**
+
+| Verdict | Meaning | Purged? |
+|---|---|---|
+| `admitted` | An allowlisted channel at an admitted tier — the gate would write it today. | **Never**, even if you name it. Delete it by hand if you want it gone. |
+| `refused` | A numeric-labelled dir whose id is not in `allowed_channel_ids`, or is declared `intimate`/`all`. | Yes, by default. |
+| `unresolved` | The dir is named after the channel (`general`), not its id, so it cannot be matched against `allowed_channel_ids`. | Only when you name it: `crawdad capture purge --channel general --apply`. |
+
+Most directories will be `unresolved`: CrawDad names each capture directory
+after the channel's *name*, and a name cannot be mapped back to a channel id
+offline. Guessing is the one way this tool could destroy legitimate data, so it
+does not — it shows you the contents and waits for you to name the directory.
+
+A directory is only read as a channel **id** when its name could plausibly be
+one: all digits, no leading zero, and at least 15 of them (Discord ids are
+`(ms since the 2015 epoch) << 22`, so real ones run 15-19 digits). This matters
+because channel *names* made only of digits are ordinary — `2024`, `420`, `911`.
+Without the length floor, `2024` would be read as channel id 2024, miss your
+allowlist, and be auto-purged even though the channel itself is allowlisted. If
+you have a genuinely id-labelled directory from very early 2015 it will show as
+`unresolved` instead; that costs you one `--channel` flag, not your data.
+
+Anything under the capture root that is not a real channel directory — a loose
+file, or a symlink — is skipped and listed at the end of the audit, so the
+report never quietly under-states what is on disk. Purge never follows a
+symlink.
+
+**Purge removes whole channel directories, never individual records.** A capture
+record stores the author's display name but no user id and no channel id, so
+`allowed_user_ids` cannot be re-evaluated from disk. An `admitted` directory may
+therefore still contain messages from users who were never allowlisted; the
+audit's authors column is how you spot them, and removing them means removing
+the whole directory. Filtering on a display name instead would match on a
+spoofable, mutable key — it would either destroy legitimate messages or leave
+the leak while reporting it fixed.
+
+#1262 tracks carrying the tier end-to-end so `intimate` channels can be captured
+faithfully instead of refused outright.
 
 ### LLM provider selection
 
--- a/crawdad/crawdad/capture.py
+++ b/crawdad/crawdad/capture.py
@@ -69,6 +69,40 @@ class _BackfillChannel(Protocol):
     ) -> AsyncIterator[_MessageLike]: ...  # pragma: no cover - protocol stub
 
 
+def iter_channel_records(channel_dir: Path) -> Iterator[dict[str, object]]:
+    """Yield every parsed JSON record under a single capture channel dir.
+
+    The one reader of the on-disk capture format. :class:`MessageCapture` uses it
+    for dedup and gap-filling; :mod:`crawdad.capture_audit` uses it to summarise
+    a pre-#1052 tree (#1264). Both must agree on what "a record" is, so neither
+    parses JSONL itself.
+
+    Malformed lines, blank lines, and non-object JSON are skipped rather than
+    raised: a corrupt line must not hide the intact records around it from an
+    audit, and it never blocked the writer either.
+
+    Args:
+        channel_dir: A ``<capture_dir>/<channel>`` directory. A path that is not
+            a directory yields nothing.
+
+    Yields:
+        Each JSON object found, in filename then line order.
+    """
+    if not channel_dir.is_dir():
+        return
+    for jsonl in sorted(channel_dir.glob("*.jsonl")):
+        for line in jsonl.read_text(encoding="utf-8").splitlines():
+            stripped = line.strip()
+            if not stripped:
+                continue
+            try:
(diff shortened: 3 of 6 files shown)


Geoffe-Ga:
## Summary
Adds an offline `crawdad capture audit|purge` helper that lets an operator inspect and clean up a bot-capture tree written before #1052 introduced the allowlist+tier gate. Audit is read-only; purge is dry-run unless `--apply`. All changes are confined to `crawdad/`; no `creek-tools` source or capture file format changed.

## Strengths
- The audit's gate re-derivation (`_verdict_for`) was checked against the live implementation and matches exactly: it reads `allowed_channel_ids` and calls the same tier lookup as `bot.py::_channel_tier` (`config.attachments.channel_privacy_tiers.get(channel_id, DEFAULT_CHANNEL_TIER)`), and compares against the same `CAPTURE_ADMITTED_TIERS` frozenset — no drift, no second copy of the rules.
- Correctly identifies and documents the one condition it *cannot* re-derive (`allowed_user_ids`, since `_record_for` stores no user id) rather than faking it with a spoofable display-name match — a real security-conscious design call, not just a caveat.
- The snowflake-length floor (`_MIN_SNOWFLAKE_DIGITS = 15`) prevents a digit-only channel *name* (`2024`, `420`) from being misread as a channel id and silently auto-purged — this is the PR's own documented Gate-2.5 blocker, fixed test-first with a red-test proof included in the PR body.
- Defense in depth in `apply_purge`: every target is re-confined to the resolved capture root before any deletion begins (all-or-nothing), symlinked channel dirs are never audited or followed, and an admitted directory is refused even when named explicitly via `--channel`.
- Extraction of `iter_channel_records` out of `MessageCapture._iter_records` is a clean, behavior-preserving refactor (verified line-by-line against the pre-PR implementation) that lets the audit share the one JSONL parser instead of growing a second one.
- Test suite (47 tests) is entirely behavioral — built on a fixture tree that always mixes admitted/refused/unresolved directories — and covers CLI wiring, dry-run defaults, traversal attempts, symlink handling, and the acceptance-pinning test `test_apply_purge_never_deletes_records_the_gate_would_admit`.
- Documentation (`README.md`, `CLAUDE.md` §5.2) is detailed, accurate to the code, and explicitly states the tool's structural limitations rather than hiding them.

## Security Concerns
🟢 LOW — `apply_purge` has a narrow TOCTOU window between its `path.is_dir() and not path.is_symlink()` check and the `shutil.rmtree` call; an attacker with concurrent filesystem write access to the vault could swap a target directory for a symlink in that window. This is an offline, single-operator CLI tool over the operator's own vault (not a multi-tenant or networked surface), so the practical risk is minimal, and `shutil.rmtree` itself refuses to act on a symlink passed as its top-level argument. Not blocking.

## Problems
None found. The one prior blocker (digit-only channel names being auto-purged) was already fixed pre-review per the PR's own Gate 2.5 writeup, and independent inspection confirms the fix (`_resolve_channel_id`'s digit-count + leading-zero guard) is correct and matches its test coverage.

## Code Quality
- Minor: the PR description's "Files" section says "41 tests" while the "Test plan" section says "47 tests" (and the file in fact contains 47 `def test_*` functions) — a harmless inconsistency in the PR body text, not the code.
- The lexicographic timestamp-ordering assumption in `_record_timestamp` is well-documented as fragile to a future writer format change (only affects the reported date range, never a purge decision) — worth keeping in mind but not a current defect.

## CLAUDE.md Adherence
✅ Single feature branch (`issue/1264-issue`) merging into `main`, conventional commit title (`feat(crawdad): ...`) with `Co-Authored-By` trailer — matches repo commit conventions.
✅ PR body documents a TDD workflow (guard-reversion table proving each safety check is load-bearing) consistent with the "Stay Green" 4-gate process.
✅ All new/changed code carries Google-style docstrings; nothing observed that would fail the 95% interrogate floor.
✅ mypy-strict-compatible typing throughout (`TYPE_CHECKING`-gated imports, `from __future__ import annotations`, precise generic types) consistent with existing `capture.py` conventions.
⚠️ Coverage/lint/type-check numbers (94.88%, 7/7 check-all.sh) are self-reported in the PR body; I was unable to execute `crawdad/scripts/check-all.sh` in this review sandbox (no `uv`/interactive command approval available) to independently confirm, but static review of the diff found no violations of the stated thresholds.

## Testing
Strong. 47 new tests in `test_capture_audit.py` cover: full audit coverage of every channel dir, record-count/date-range/author extraction, all three verdict paths (admitted/refused/unresolved) including the intimate and `all`-tier cases, the digit-only-name safety net (with an explicit "teeth of it" auto-purge-survival test), leading-zero handling, missing-root and malformed-JSONL tolerance, symlink/loose-file disclosure, purge planning (targeting, declining admitted/unknown/traversal labels, idempotency), purge application (root confinement, symlink refusal, byte-identical survival of admitted data), report rendering, and full CLI wiring end-to-end. Insufficient testing is not a concern here.

## Documentation
Thorough and accurate. `README.md` gives the operator a runnable walkthrough with the verdict table and the two structural limitations stated plainly. `CLAUDE.md` §5.2 gives future maintainers the same "do not relax the floor" warning embedded directly in the module's own docstrings and code comments. Docs match the implementation exactly on every point I checked.



## Verdict: LGTM

No bugs found: `_verdict_for`'s allowlist/tier checks were cross-checked line-by-line against the live gate (`bot.py::_capture_allowed`, `_channel_tier`, `config.py::CAPTURE_ADMITTED_TIERS`/`DEFAULT_CHANNEL_TIER`) and match exactly, the snowflake-floor and traversal/symlink defenses are sound and tested, and the 47 new tests are behavioral (including a reverted-guard proof for each safety check). Nothing here should block merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5235102664 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1298, opened 2026-08-10
fix(redact): symlink guards validate the whole tree, then read it later (TOCTOU window)


Geoffe-Ga:
## Role

You are narrowing a check-then-read window in the redaction path-traversal guards.

## Goal

Both symlink containment guards validate the whole tree first and act on it afterwards, so a link that passed validation can have its target swapped before the file is actually opened.

- `creek/redact/scanner.py::_scannable_candidates` materialises the entire admitted-candidate list in one pass; `scan_batch` then opens and reads each file afterwards. A symlink validated early in the walk can be re-pointed at any time before its turn in the read loop.
- The same shape predates it on the write path: `creek/redact/cli_commands.py::_assert_no_escaping_symlinks` validates the whole tree, and `_apply_redactions` opens the files later.

This is not a regression introduced by #1087 — it is the codebase's existing two-phase idiom, and it was reviewed and accepted there on the grounds that Creek is a single-operator offline tool rather than a multi-tenant service. Filing it so the acceptance is written down somewhere other than a PR thread.

## Constraints on any fix

- Do not widen the window further, and do not make either guard more permissive.
- The obvious hardening is fd-based: open with `O_NOFOLLOW` and read through the descriptor that was validated, rather than re-resolving a path. `os.open(path, os.O_RDONLY | os.O_NOFOLLOW)` refuses a symlink at the final component; combined with validating the parent directory chain this collapses the window.
- `tests/test_cli_redact.py`'s SEC-003 block and the #1087 tests must keep passing unmodified — this is a strengthening, not a behaviour change.
- Any test must actually demonstrate the race rather than assert the API shape. A test that swaps the link between the validation pass and the read pass (e.g. by patching the iteration, or by a controlled two-phase fixture) is the only thing that proves the fix.

## Worth deciding first

Whether this is worth fixing at all. The threat model is a local operator scanning their own vault; an attacker who can swap a symlink mid-scan in that environment already has write access to the tree being scanned. Closing the issue as "accepted risk, documented" is a legitimate outcome — but the reasoning should land in `docs/security/threat-model.md` next to the SEC-003 entry rather than being rediscovered by the next reviewer.

Refs #1087

Issue #1294, opened 2026-08-10
fix(ingest): ingestor discovery walks follow symlinks out of the source root


Geoffe-Ga:
## Role

You are confining the ingestion walks to their source root.

## Goal

Every ingestor discovers its inputs with an unguarded recursive glob and reads whatever the glob yields. A symlink under the source tree pointing outside it is read and ingested into the vault as a fragment.

`#1087` confined `RedactionScanner.scan_batch` and made `Pipeline._run_redaction_pass` **refuse** a source tree containing an escaping symlink, precisely so that closing the scanner hole did not convert "blocked ingest" into "silent unredacted ingest". That refusal is a backstop, not a fix: it is bypassed entirely when `redaction.enabled = false`, and it does not protect any caller that runs an ingestor directly.

## Verified evidence

- `creek/ingest/markdown.py::MarkdownIngestor._read_directory` does `sorted(dir_path.rglob("*.md"))` followed by `md_file.read_bytes()` — no `is_symlink()` check, no `is_file()` check, no root confinement.
- Sibling ingestors under `creek/ingest/` repeat the same pattern; audit all of them rather than patching one.
- The policy to reuse already exists twice in the codebase: `creek/redact/cli_commands.py::_assert_no_escaping_symlinks` (write path, refuse) and `creek/redact/scanner.py::_scannable_candidates` (read path, skip + count), both resolving root and child with `strict=False` and treating `(OSError, RuntimeError, ValueError)` as escaping.

## Output format

A PR that:

1. Introduces one shared confinement predicate rather than a third copy of the policy, and applies it to every ingestor discovery walk.
2. Red-first: a test that ingests a source tree containing `link.md -> <outside>/secret.md` with `redaction.enabled = false`, and proves today that the out-of-tree content lands in the vault as a fragment.
3. Decides explicitly — and writes the decision down — whether ingestion **skips** the escaping entry (with an error on `PipelineResult.errors`, so it is not silent) or **refuses** the whole run. Given ingestion writes to the vault, refusal is the stronger default.
4. Pins the `rglob` bound: `Path.glob` grew an explicit `recurse_symlinks` parameter in 3.13 and does not descend into symlinked directories by default. If a future default flips, the leaf-only check silently widens.

## Constraints

- Security direction is one-way.
- Do not weaken the `#1087` pipeline refusal; this replaces the need for it as the *only* guard, it does not remove it.
- No bypasses, `./scripts/check-all.sh` exits 0.

[...]

Issue #1293, opened 2026-08-10
fix(redact): SEC-003 guard is dir-only — 'creek redact --apply <symlinked-file>' writes through the link


Geoffe-Ga:
## Role

You are closing a path-traversal hole on the redaction **write** path.

## Goal

The SEC-003 symlink guard `creek/redact/cli_commands.py::_assert_no_escaping_symlinks` is only ever reached when the user-supplied root is a **directory**. When the operator names a **symlinked file** directly, the guard is skipped entirely and `_apply_redactions` writes *through* the link to a target outside any root.

This is distinct from #1087. #1087 is the **read** path: `RedactionScanner.scan_batch` walking a directory tree and following an escaping symlinked child. It is fixed by confining the walk. This issue is the **write** path with an explicitly named file argument, where there is no walk at all — the guard is simply never invoked.

## Verified call path

1. `creek redact --apply --source <path>` -> `run_apply` (`creek/redact/cli_commands.py`).
2. `run_apply` gates the guard behind `if source.is_dir():` — so `_assert_no_escaping_symlinks(source, ...)` never runs for a file argument.
3. `_scan_source` then takes its `if source.is_file():` branch, calling `scanner.scan_file(source)` directly — which also bypasses the `scan_batch` walk guard added by #1087.
4. `_apply_redactions` opens the named path for writing. Because the path is a symlink, the write lands on the link's **target**, which may be anywhere on disk.

`run_review` has the identical shape: its guard sits behind `if vault.is_dir():`.

Reproduction sketch:

```
mkdir src-not-used
printf 'API_KEY=sk-abcdefghijklmnopqrstuvwx\n' > /tmp/victim.env
ln -s /tmp/victim.env ./leak.env
creek redact --apply --source ./leak.env --yes
# /tmp/victim.env is rewritten, outside any root the operator named
```

## Output format

A PR that:

1. Extends the refusal to the file-argument case in both `run_apply` and `run_review`: if the named path is itself a symlink whose `resolve(strict=False)` target is not a descendant of the path's own parent (or of an explicitly supplied root), refuse with the existing SEC-003 error shape and a non-zero exit.
2. Also confines `_scan_source`'s `source.is_file()` branch, which today calls `scan_file` with no symlink check at all.
3. Red-first: a test proving `creek redact --apply <symlinked-file>` rewrites the out-of-tree target **today**, and refuses after the fix. Assert the victim file's bytes are unchanged.
[...]

Issue #1292, opened 2026-08-10
feat(mcp): surface files_skipped_symlink in creek.redact.scan statistics (CONTRACT_VERSION 0.4.0 -> 0.5.0)


Geoffe-Ga:
## Role

You are hardening the `creek.redact.scan` MCP tool's response contract.

## Goal

`#1087` adds a `files_skipped_symlink` counter to `creek.redact.scanner.ScanSummary`, and surfaces it in the CLI stats table and in `generate_markdown_summary`. It deliberately does **not** add it to the typed MCP `statistics` object, because that object is a versioned wire surface and a security fix is the wrong vehicle for a contract bump.

This issue lands the typed key.

## Context (verified at the time of filing)

- `creek_mcp/tools/redact.py::redact_scan_tool` builds `statistics` from exactly four hardcoded literal keys: `files_scanned`, `files_skipped_binary`, `files_skipped_extension`, `total_findings`.
- `_relativised_summary` propagates the dataclass via `dataclasses.replace`, so a new `ScanSummary` field is already carried through — it is simply never rendered onto the wire.
- `CONTRACT_VERSION` is `"0.4.0"` (`creek_mcp/contract.py`).
- `creek.redact.scan` is enumerated in `tests/test_mcp_server.py` and covered by the `#1255` wiring contract test.
- CrawDad consumes `report_markdown` (which *does* already show the skip after `#1087`) but not `statistics.files_skipped_symlink`.

## Output format

A PR that:

1. Adds `"files_skipped_symlink": scanned.files_skipped_symlink` to the `statistics` dict in `redact_scan_tool`.
2. Bumps `CONTRACT_VERSION` `0.4.0` -> `0.5.0` (additive key on an existing object).
3. Updates any contract snapshot/enumeration test that pins the statistics key set.
4. Updates `docs/mcp.md`'s `creek.redact.scan` response documentation.
5. Updates `creek_mcp/read_gate.py`'s `creek.redact.scan` posture rationale, which after `#1087` notes the counter is not yet typed on the wire.

## Constraints

- Additive only: no existing key renamed or removed.
- No behaviour change in the scanner — this is a rendering/contract change.
- `./scripts/check-all.sh` exits 0.

Refs #1087

Geoffe-Ga:
## Heads-up from #1338 — this issue is unaffected, and slightly easier

#1338 offered "wire `generate_json_report` with the missing
[...]

PR #1301, opened 2026-08-10
fix(redact): confine the scan walk to its root (SEC-003)

### LINKED ISSUE #1087: RedactionScanner.scan_batch follows symlinked files out of the scan root (SEC-003 guards --apply/--review, not the scan walk)

Geoffe-Ga:
## Summary

`RedactionScanner.scan_batch` walked its scan root with
`sorted(child for child in dir_path.rglob("*") if child.is_file())`. `is_file()` follows symlinks, so a symlinked child whose target resolved **outside** the scan root was opened, its PII types and line numbers reported, and it incremented `files_scanned` — an existence-and-readability bit about a file the caller never pointed the scanner at. SEC-003 guarded `--apply` and `--review`; it never guarded the scan walk.

The guard now lives inside the walk, so **all three** surfaces inherit it. The issue body names two; there is a third:

| Surface | Reaches the walk via | Disposition |
|---|---|---|
| `creek redact --scan` | `cli_commands._scan_source` | skip + count |
| `creek.redact.scan` MCP tool | `tools/redact.py::redact_scan_tool` | skip + count |
| `Pipeline` redaction pass | `pipeline._run_redaction` | **refuse** |

`creek/redact/scanner.py` gains `_scannable_candidates()` — now the module's only filesystem walk — which resolves the root once and, for a child that is *itself* a symlink, requires its resolved target to land under that root. It reuses the predicate the shipped SEC-003 write guard already uses rather than inventing a second one. Non-symlink children are never resolved, which is what keeps a root reached *through* a symlinked component (`/tmp` → `/private/tmp` on macOS) scannable; resolving them too would flag every child of such a root as escaping.

### Three judgement calls a reviewer would otherwise read as omissions

**(i) The read path skips; it does not refuse.** Refusing would turn a diagnostic tool into a hard failure on any tree holding a legitimate external link, and would change the exception contract at three call sites. But a silent skip in a safety scanner is its own hazard, so the skip is counted and surfaced: `ScanSummary.files_skipped_symlink`, a row in the CLI stats table, and a line in `generate_markdown_summary` — including the empty-findings path, which no longer renders a bare `No findings.` when the scan declined to read something. That last one matters because CrawDad posts `report_markdown` verbatim into a Discord channel, so it is the only surface a human actually sees; a clean scan that skipped nothing still renders byte-identically to before.

**(ii) The pipeline path refuses; it does not skip.** This is the asymmetry, and it is load-bearing. Today the escaping symlink's target yields matches and `RedactionRequiredError` blocks the ingest — so the hole is *masked* here by a different guard. Once the walk starts skipping, the scan comes back clean, the pipeline proceeds, and `MarkdownIngestor._read_directory` re-reads the very same file with `rglob("*.md")` + `read_bytes()` and no symlink guard of its own. Skipping in the pipeline would therefore have converted a **blocked ingest into a silent unredacted ingest**, with the gate that used to stop it now reporting success. `SymlinkEscapedSourceError` is raised before the `dry_run` arm and before the matches arm, because a path-traversal guard admits no waiver.

**(iii) The counter is not on the typed MCP wire.** `creek.redact.scan`'s `statistics` object keeps exactly its four keys and `CONTRACT_VERSION` stays `0.4.0`. The typed key is a versioned contract surface and a security fix is the wrong vehicle for a contract bump — deferred to #1292.

### Prose

Four places asserted this residual still exists; leaving any stale ships a lie inside a security rationale. Updated: `creek_mcp/read_gate.py` (the `creek.redact.scan` posture rationale), `creek_mcp/tools/redact.py` (module docstring), `docs/mcp.md`, and `docs/security/threat-model.md` — whose SEC-003 entry now states *which walks* the guard covers, which was the issue's third acceptance criterion.

### Deliberately out of scope, filed instead

- **#1293** — the SEC-003 guard is directory-only (`run_apply`/`run_review` gate it behind `is_dir()`), so `creek redact --apply <a-symlinked-file>` skips it entirely and writes *through* the link. A write-path traversal, distinct from this read-path walk.
- **#1294** — the ingestor discovery walks follow symlinks out of the source root; the pipeline refusal above is a backstop that `redaction.enabled: false` never reaches.
- **#1292** — the typed MCP `statistics` key.

`_assert_no_escaping_symlinks` is **not** refactored, even though the new helper duplicates its predicate: it is the shipped, resolved SEC-003 write guard, and rewriting it inside a security fix for a different code path buys no security. `tests/test_cli_redact.py:407-542` passes byte-unmodified.

## Test plan

Red-first, and the RED was **executed and observed**, not asserted.

**Before the fix** — 14 failed / 389 passed, failing behaviourally rather than on a missing symbol:

```
tests/test_mcp_redact.py::test_scan_does_not_read_a_symlink_escaping_the_scan_root[open]
  AssertionError: creek.redact.scan opened a symlinked child whose target resolves
  outside the scanned root and counted it, so files_scanned carries an
  existence-and-readability bit about an out-of-scope file.
    statistics: {'files_scanned': 2, ...}
  assert 2 == 1

tests/test_cli_redact.py::test_redact_scan_does_not_read_a_symlink_escaping_the_source_root
  AssertionError: the scan reported the PII type of a file outside the source root,
  so it followed the symlink and read it.
    Findings by Type:  email | medium | 1     ssn | critical | 1
```

**Load-bearing proof.** Reverting *only* the `_scannable_candidates` call back to the raw `rglob` comprehension turns exactly 8 tests red: the chokepoint test, the AST walk guard, all four MCP ceilings, the CLI `--scan` test, and the pipeline test. The breadth guards — in-root alias, dangling link, circular pair, symlinked directory, symlinked root component — correctly stay green, confirming they guard against over-restriction rather than doing the discriminating.

**Coverage of the acceptance criteria.** The escaping target contributes zero findings *and* does not increment `files_scanned`; exercised through both surfaces the issue names, with the MCP case parametrised over all four `TierCeiling` values; a filename canary asserts the target's stem never appears in `json.dumps(result)`. In-root symlinks still scan and still report. Circular and dangling links do not raise. A symlinked *directory* yields nothing — pinned deliberately, because `Path.glob` grew an explicit `recurse_symlinks` parameter in 3.13 and a future default flip would silently widen the hole under a leaf-only check. `test_scan_batch_is_the_only_filesystem_walk_in_the_scanner_module` AST-asserts the module has no second walk, and ships a negative control so it cannot rot into a green no-op.

Two mocks in `tests/test_pipeline.py` were retargeted from `scan_directory` to `scan_batch` to follow the changed production call, with a real `ScanSummary()` rather than a `MagicMock` — a `MagicMock` answers every attribute with a truthy object, which would fire the refusal on every mocked run and hide the check instead of exercising it. No assertion lost strength.

A second load-bearing proof covers the dry-run interaction specifically. Moving the refusal *below* the `dry_run` arm turns both pipeline tests red, and the captured log shows the out-of-tree canary being ingested for real — `returning fragment 'CANARY-PIPELINE-SYMLINK-4c21'`. That is the "silent unredacted ingest" outcome in (ii), demonstrated rather than argued.

**Gate 2:** `./scripts/check-all.sh` exit 0 — 12/12 checks, 8749 passed, 94.38% branch coverage, per-file gate passed with no new entry in `scripts/coverage-waivers.txt`, mypy strict, ruff zero (`--no-cache`), pylint 9.70, bandit clean.

**Gate 2.5** returned CLEAN with no blockers. Three minors actioned:

- The `_resolves_within` docstring claimed a dangling-but-inward link "resolves cleanly and is admitted". Unreachable — `is_file()` gates first and is `False` for any dangling link, so it never reaches the containment check. Corrected; an inaccurate claim in a security docstring is worth the same care as an inaccurate one in code.
- Added `test_symlink_refusal_is_not_waived_by_redaction_dry_run`, pinning that `dry_run` cannot waive a traversal refusal. The ordering was correct but unpinned, and a later edit moving the check would have left every other test green.
- TOCTOU between validation and read filed as **#1298** (P3). It is the codebase's existing two-phase idiom — `_assert_no_escaping_symlinks` validates the tree and `_apply_redactions` reads later — so it predates this change rather than being introduced by it.

One nit was declined: importing `SYMLINK_SKIP_LABEL` into the tests instead of spelling it out. `tests/test_mcp_redact.py` records the opposite convention explicitly — "an equality assertion built out of the implementation's own constants can only ever agree with itself" — and this label is rendered into a channel CrawDad posts to, so the test should fail if the wording changes.

**Disclosed:** `scanner.py` grew 938 → 1129 lines and now trips pylint `C0302 too-many-lines`. The score stays 9.70 against a 9.0 gate. Splitting the module was declined as an out-of-scope refactor inside a security fix.

No bypasses: no `noqa`, no `type: ignore`, no skip marks, no threshold changes, no deleted or weakened assertions. The two `noqa` lines in `scanner.py` are pre-existing at HEAD.

Closes #1087
Refs #972, #1292, #1293, #1294, #1298


Geoffe-Ga:
## Summary
Closes #1087: `RedactionScanner.scan_batch`'s directory walk used `child.is_file()` on every `rglob("*")` result, which follows symlinks — so a symlinked child whose target resolved *outside* the scanned root was opened, its PII types/line numbers reported, and it incremented `files_scanned`. The fix moves the containment check into the walk itself (`_scannable_candidates`), the module's now-single filesystem enumeration point, so all three read-path callers (`creek redact --scan`, the `creek.redact.scan` MCP tool, and the pipeline's redaction pass) inherit it. The read path skips-and-counts; the pipeline path escalates the same skip to a hard refusal (`SymlinkEscapedSourceError`), because `MarkdownIngestor._read_directory` re-reads the same tree with no symlink guard of its own and a silent skip there would have converted a blocked ingest into a silent unredacted one.

## Strengths
- Correctly identifies and fixes the asymmetry between the write-path guard (SEC-003, `_assert_no_escaping_symlinks`) and the read-path walk that had no equivalent check.
- The single-chokepoint design (`_scannable_candidates`) is enforced structurally, not just by convention: `test_scan_batch_is_the_only_filesystem_walk_in_the_scanner_module` AST-walks `scanner.py` and fails if a second raw `rglob`/`os.walk`/etc. appears outside the helper, with a negative-control test (`test_the_walk_detector_flags_a_walk_outside_the_helper`) proving the detector isn't a green no-op.
- The skip-vs-refuse asymmetry between the read path and the pipeline is deliberate and well-reasoned, and is pinned by a dedicated test (`test_symlink_refusal_is_not_waived_by_redaction_dry_run`) that the refusal sits above the `dry_run` early-return.
- Correctly avoids over-tightening: in-root aliases, dangling links whose target is nominally inside the root, circular symlink pairs, and roots reached *through* a symlinked path component (`/tmp` → `/private/tmp`) all remain scannable — each has its own test, so the fix closes exactly the escape and nothing broader.
- Verified `resolved = resolve_within_vault(...)` in `creek_mcp/tools/redact.py` already fully resolves symlinks for the single-file scan path, so the new per-child check in `scan_batch` is only needed (and only applied) for the directory-walk case — no double-gap between the two code paths.
- `scan_directory` delegates to `scan_batch`, so the fix is inherited without a second call site to patch.
- Documentation (MCP tool docstring, `read_gate.py` posture rationale, `docs/mcp.md`, `docs/security/threat-model.md`) is updated everywhere the closed residual was previously described as open — no stale claims left behind.
- Test plan describes an executed RED-then-GREEN cycle with load-bearing reversion checks (reverting `_scannable_candidates` turns exactly 8 tests red), which is a stronger evidentiary bar than most PRs meet.

## Security Concerns
🟢 LOW — I found no unaddressed vulnerability. Two residuals are explicitly disclosed and filed as separate issues rather than silently left open: the SEC-003 write guard is still directory-only (#1293, symlinked-file targets to `--apply` bypass it), and the ingestor's own directory walks still follow symlinks regardless of the redaction stage (#1294, the pipeline refusal is a backstop for `redaction.enabled: true` only). Both are pre-existing gaps outside this PR's stated scope, not introduced by it.

## Problems
None found that block merge.

## Code Quality
- `creek_mcp/tools/redact.py`'s MCP response computes `status = "empty" if not scanned.matches else "ok"` without factoring in `files_skipped_symlink`. A caller that reads only `status`/`statistics` (not `report_markdown`) on a scan that found nothing *because it declined to look* still sees `"status": "empty"` with no hint anything was skipped — the typed wire doesn't carry the counter yet. This is explicitly disclosed in the PR description as deferred to #1292 (to avoid a contract-version bump inside a security fix), so it's a reasonable, tracked tradeoff rather than an oversight — but worth double-checking that #1292 is prioritized, since it's the one place the fix's safety net doesn't reach a machine consumer.
- `scanner.py` now trips pylint's `C0302` (too-many-lines); score stays 9.70 against the 9.0 gate, and splitting the module is explicitly declined as out-of-scope for a security fix. Reasonable call, flagged here only as the disclosed tech debt it already is.
- `_resolves_within` in `scanner.py` duplicates (rather than shares) the containment predicate already implemented in `cli_commands.py::_assert_no_escaping_symlinks`. The PR explicitly declines to refactor the shipped guard inside a security-fix PR, which is a defensible call, but a future follow-up to unify the two predicates would remove the duplication per CLAUDE.md §1.2.

## CLAUDE.md Adherence
✅ Scripts-only workflow followed (check-all.sh reported 12/12 checks, exit 0). ✅ TDD/Stay-Green: RED tests executed and observed before the fix, not merely asserted. ✅ No forbidden shortcuts — no new `noqa`/`type: ignore`/skip marks; the two pre-existing `noqa` lines in `scanner.py` are unrelated to this diff. ✅ Conventional commit message (`fix(redact): ...`) with issue references. ✅ Coverage (94.38% branch, no new coverage-waivers entry), mypy strict, ruff zero, bandit clean, pylint 9.70 all reported green. ✅ DRY mostly honored; the one duplicated predicate is disclosed and justified above.

## Testing
Strong. New tests cover the chokepoint directly (`test_redact.py`), both read-path surfaces (CLI `--scan`, MCP `creek.redact.scan` parametrised over all four `TierCeiling` values), and the pipeline refusal (including the dry-run non-waiver case). Boundary tests (in-root alias, dangling link, circular pair, symlinked directory, symlink-reached root) guard against over-restriction. The AST-based single-walk-chokepoint test plus its negative control is a notably strong structural regression guard. Markdown-rendering tests pin both the new caveat wording and byte-identical output for the unaffected (nothing-skipped) case, protecting CrawDad's Discord-posted `report_markdown` consumer.

## Documentation
Thorough and consistent — every prior "residual" claim (in `read_gate.py`, `creek_mcp/tools/redact.py`'s module docstring, `docs/mcp.md`, `docs/security/threat-model.md`) is updated to reflect the fix rather than left stale, and the new threat-model entry accurately distinguishes the three walks' differing dispositions (refuse / skip-and-count / refuse-as-backstop).



## Verdict: LGTM

The fix correctly closes the disclosed symlink-following read-path gap at its single chokepoint, is backed by an unusually rigorous test plan (executed RED-first, load-bearing reversion tests, an AST-enforced single-walk invariant with its own negative control), and leaves no stale documentation or unaddressed CLAUDE.md quality gate. The one substantive non-blocking note — the MCP `status`/`statistics` wire not yet carrying the skip counter — is already explicitly disclosed and tracked as #1292, not a silent gap, so it doesn't rise to a blocking concern.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5235226720 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR closes SEC-003's read-path gap: `RedactionScanner.scan_batch` used `child.is_file()` over a raw `rglob("*")`, which follows symlinks, so a symlinked child resolving outside the scan root was opened, its PII types/line numbers reported, and counted in `files_scanned`. The fix introduces `_scannable_candidates()` as the module's single filesystem walk, gated by a containment check reusing the SEC-003 write guard's predicate. The three callers (CLI `--scan`, `creek.redact.scan` MCP tool, and the pipeline's redaction pass) now inherit the guard, with a deliberate asymmetry: read/report surfaces **skip+count** the escaping file, while the pipeline **refuses outright** via a new `SymlinkEscapedSourceError` (because `MarkdownIngestor._read_directory` has no symlink guard of its own and would silently re-read the same file).

## Strengths
- Single chokepoint enforced structurally, not just by convention: `test_scan_batch_is_the_only_filesystem_walk_in_the_scanner_module` AST-walks `scanner.py` and fails if any walk bypasses `_scannable_candidates`, with a negative-control test (`test_the_walk_detector_flags_a_walk_outside_the_helper`) proving the detector isn't a no-op.
- The skip-vs-refuse asymmetry is correct and well-justified: I verified `MarkdownIngestor._read_directory` (`creek/ingest/markdown.py:346`) really does `rglob("*.md")` + `read_bytes()` with no symlink check, which is exactly why the pipeline must escalate to a hard refusal rather than silently skip (`creek/pipeline.py:481-492`) — skipping there would turn a currently-blocked ingest into a silently unredacted one.
- Containment predicate matches the shipped SEC-003 write guard (`_assert_no_escaping_symlinks` in `cli_commands.py:238`): resolve the root once, only resolve children that are themselves symlinks, `relative_to()` for containment. Verified this is the same policy, not a subtly different reimplementation.
- Good breadth-guard coverage: in-root alias, circular symlink, dangling symlink, symlinked directory (pinned as a canary against a future `Path.glob(recurse_symlinks=...)` default flip), and a root reached *through* a symlinked component (`/tmp` → `/private/tmp`) — all correctly distinguished from a true escape.
- Byte-identical output preserved for the unaffected common case: `_EMPTY_SUMMARY_MARKDOWN` stays `"No findings.\n"` when nothing was skipped, and the CLI stats table only gains the new row when the counter is non-zero — avoids alert fatigue and doesn't break CrawDad's existing Discord-posted markdown for the vast majority of clean scans.
- Documentation was cross-checked for staleness: I grepped the whole `creek-tools/` tree for `#1087` and old "still follows symlinked" language — the second commit correctly caught and fixed all three prose spots this PR's own fix falsified (`read_gate.py`, `tools/redact.py` docstring, `docs/mcp.md`), and none were missed.
- `_relativised_summary` uses `dataclasses.replace`, which correctly carries `files_skipped_symlink` through to the vault-relative summary used for both `findings` and `report_markdown` — verified this isn't silently dropped.

## Security Concerns
🟢 LOW — No new vulnerability found; this is a net security improvement. Two residual gaps are pre-existing and explicitly filed rather than introduced by this PR: SEC-003's write guard is directory-only (#1293, `creek redact --apply <symlinked-file>` bypasses it — confirmed in `_scan_source`, which calls `scanner.scan_file()` directly for a file argument, bypassing `_scannable_candidates` entirely), and the ingestor's own walks still follow symlinks regardless of `redaction.enabled` (#1294). Both are correctly scoped out and tracked.

## Problems
None found that block merge.

## Code Quality
- `_resolves_within`'s docstring correctly documents that dangling links never reach the containment check (gated out by `is_file()` first) — confirmed against `_scannable_candidates`'s actual gate ordering.
- `SYMLINK_SKIP_LABEL` is exported and reused between `cli_commands.py`'s Rich table and `scanner.py`'s markdown renderer rather than being spelled twice — good DRY discipline consistent with CLAUDE.md §1.2.
- The MCP tool's `status` field (`redact_scan_tool` in `creek_mcp/tools/redact.py:224`) is computed as `"empty" if not scanned.matches else "ok"` — a scan that skipped a file due to an escaping symlink but found no other matches still reports `status: "empty"`, with the skip visible only in `report_markdown` text and not in the typed `statistics` dict. This is disclosed and intentionally deferred to #1292 (avoiding a contract-version bump inside a security fix), which is a reasonable call, but a programmatic caller that only inspects `status`/`statistics` (rather than parsing `report_markdown`) could still be misled into treating an incomplete scan as clean. Worth a forward pointer from `status`'s docstring to #1292, though not blocking.

## CLAUDE.md Adherence
✅ TDD followed — the PR description documents an executed (not asserted) RED phase, with specific failure output and a load-bearing revert-proof (reverting `_scannable_candidates` to a raw `rglob` turns exactly 8 tests red).
✅ `./scripts/check-all.sh` reported exit 0 per the PR description (12/12 gates, 94.38% branch coverage, mypy strict, ruff zero, pylint 9.70, bandit clean).
✅ No shortcuts — no new `noqa`/`type: ignore`/skip marks; the two pre-existing `noqa` lines in `scanner.py` predate this change.
✅ Conventional commit messages, `Closes #1087` with proper cross-references.
⚠️ pylint's `C0302 too-many-lines` was tripped on `scanner.py` (938 → 1129 lines); disclosed by the author, and the module-splitting question was explicitly deferred as out-of-scope for a security fix rather than swept under the rug — acceptable, score stays well above the 9.0 gate.

## Testing
Strong. The chokepoint test (`test_scan_batch_skips_a_symlink_resolving_outside_the_scan_root`), the MCP-boundary parametrization over all four `TierCeiling` values with a filename-canary JSON-serialization check, the pipeline refusal test run against the real filesystem (not mocked), and the dry-run-does-not-waive-the-refusal test (which specifically pins the ordering — refusal before the `dry_run` early-return) together cover the acceptance criteria well. The `test_pipeline.py` mock retargeting from `MagicMock` to a real `ScanSummary()` is a subtle, correct catch: a bare `MagicMock` would make `summary.files_skipped_symlink` truthy and fire the new refusal on every mocked test run, silently defeating the check it's supposed to exercise.

## Documentation
Mostly thorough (threat-model.md, mcp.md, and module docstrings all updated and cross-checked as internally consistent with the code, with no stale residual claims left anywhere in the tree). One gap: `docs/redaction.md` has a dedicated section "How `creek process` interacts with redaction" documenting the existing `RedactionRequiredError` failure mode with an example error message and remediation steps — the natural place a user would look after hitting the new `SymlinkEscapedSourceError`. This file wasn't touched by either commit and isn't among the "Prose" section's list of four updated files in the PR description, so it reads as an oversight rather than a deliberate scoping decision (unlike the other deferrals, which are all explicitly named and justified).



## Verdict: COMMENTS

Nothing here blocks merge — the fix is correct, well-tested (including a structural AST guard against regression), and the security posture genuinely improves with the read/refuse asymmetry properly justified against the ingestor's own unguarded walk. But `docs/redaction.md`'s user-facing "How `creek process` interacts with redaction" section — which already documents the sibling `RedactionRequiredError` failure with an example and remediation — was not updated to cover the new `SymlinkEscapedSourceError`, and this omission wasn't called out or scoped in the PR description the way the other deferrals were, so it's worth the author weighing before merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5235331375 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5235331375 to see in-depth feedback and continue iterating

Geoffe-Ga:
Fair catch, and the diagnosis was the right one: every other deferral in this PR is named and justified, so an unmentioned `docs/redaction.md` reads as an oversight rather than a scoping call. It was. Fixed in-lane rather than deferred, because `SymlinkEscapedSourceError` is an error *this PR introduces* — documenting it where a user already looks after `creek process` refuses is finishing this PR's own work, not new scope.

**`creek-tools/docs/redaction.md`** — the "How `creek process` interacts with redaction" section now carries a `SymlinkEscapedSourceError` entry in the shape the existing `RedactionRequiredError` entry set: prose, then the verbatim CLI line, then remediation.

- **The quoted error string is byte-accurate.** Verified mechanically, not by eye: the fenced block was pulled back out of the rendered markdown, whitespace-normalised for line wrapping, and compared to `"Symlink containment: " + str(SymlinkEscapedSourceError(1, Path("/tmp/exports")))` — identical. Same for the log line the remediation cites; that one came from an actual `scan_batch` run over a tree with an escaping link, not from reading the format string.
- **The remediation only names things that exist.** No flag is promised. Step 1 is finding the links: the per-skip warning `Skipping symlink that escapes the scan root: <path>` names the link as scanned and deliberately never its target, and `creek redact --scan` reports the count as the **Files skipped (escaping symlink)** row (omitted when zero). Step 2 is remove/re-point, or copy the content in so the scanner reads the bytes it is being asked to vouch for.
- **Both documented overrides are explicitly ruled out**, since the section's existing text offers two and a reader would reasonably try them: `redaction.dry_run: true` does not suppress this refusal — that setting means "log the matches you found", not "proceed past a file nobody read" — and `redaction.enabled: false` skips the stage entirely, which is the unredacted ingest the refusal exists to prevent.
- **The claim stays as narrow as the rest of the PR.** The section closes by restating that an in-root symlink is still admitted and still scanned *unresolved*, under the path it was reached by — so nothing here implies the scanner now resolves every symlink safely, and the render-as-scanned rule stays load-bearing.

Gate 2 re-run on the full tree despite this being markdown-only: exit **0**, 12/12 checks, 8794 passed. Inert as expected, run anyway.

Geoffe-Ga:
## Summary

`RedactionScanner.scan_batch` walked its scan root with `sorted(child for child in dir_path.rglob("*") if child.is_file())`, and since `is_file()` follows symlinks, a symlinked child whose target resolved outside the scan root was opened, its PII types/line numbers reported, and it incremented `files_scanned` — even though the SEC-003 write guard already refused this shape on the `--apply`/`--review` path. This PR moves the containment check into the walk itself (`_scannable_candidates` in `creek/redact/scanner.py`), the module's now-single filesystem enumeration, so all three read-path callers (`creek redact --scan`, `creek.redact.scan` MCP tool, and the pipeline's own redaction pass) inherit it, while the pipeline additionally escalates a skip into a hard `SymlinkEscapedSourceError` refusal (since the ingestor has no symlink guard of its own and a silent skip there would become a silent unredacted ingest).

## Strengths

- **Correct, narrow fix at the right layer.** The containment policy lives in the one place all three callers reach, verified by a structural AST test (`test_scan_batch_is_the_only_filesystem_walk_in_the_scanner_module`) with a negative control (`test_the_walk_detector_flags_a_walk_outside_the_helper`) so the guard can't rot into a green no-op.
- **The skip-vs-refuse asymmetry between the read path and the pipeline is well-reasoned and tested.** `test_symlink_refusal_is_not_waived_by_redaction_dry_run` specifically pins that the refusal sits *above* the `dry_run` early-return, which I confirmed by reading `pipeline.py:481-501` directly — the ordering matches the claim.
- **Root-resolved-once, leaf-only-resolved predicate correctly avoids the common over-blocking bug** (a scan root reached through a symlinked path component, e.g. `/tmp` → `/private/tmp` on macOS), with a dedicated test (`test_scan_batch_accepts_a_scan_root_reached_through_a_symlinked_component`).
- **Backward-compatible reporting.** The empty-findings markdown stays byte-identical (`_EMPTY_SUMMARY_MARKDOWN`) when nothing was skipped, and only gains the new caveat/row when something was — pinned in both directions.
- **Excellent, honest scoping.** Three known gaps (SEC-003 write guard is directory-only; ingestor walks still follow symlinks unconditionally; the MCP `statistics` wire doesn't yet carry the new counter) are explicitly named and filed as follow-ups (#1293, #1294, #1292) rather than silently left implicit or bundled in.
- **Documentation kept honest.** Four docs (`read_gate.py`, `tools/redact.py` module docstring, `docs/mcp.md`, `docs/security/threat-model.md`) that asserted the residual still existed were all updated in the same PR (plus a follow-up commit specifically to true up stale claims a merge introduced) — avoiding a security doc that lies.
- I cross-checked several of the PR's specific implementation claims against the actual checked-out source: `scan_directory` does delegate to `scan_batch`; `ScanSummary` is exported from `creek.redact`; `_relativised_summary` preserves `files_skipped_symlink` via `dataclasses.replace`; `SymlinkEscapedSourceError` and `RedactionRequiredError` are sibling `RuntimeError` subclasses (no except-ordering hazard in `cli.py`). All matched the description.

## Security Concerns

🟢 LOW — None found in the new code itself. Two pre-existing, explicitly-disclosed gaps remain open by design and are tracked separately: the SEC-003 write guard is directory-only (#1293), and a TOCTOU window between the containment check and the later read (#1298, filed as P3, described as the codebase's existing two-phase idiom rather than something this PR introduces).

## Problems

None found that block merge.

## Code Quality

- `scanner.py::_resolves_within` reimplements a predicate conceptually equivalent to `cli_commands.py::_assert_no_escaping_symlinks` rather than sharing code with it. The PR explicitly discloses and justifies this (rewriting the shipped SEC-003 write guard inside an unrelated security fix "buys no security"), which is a defensible call, but it does leave two independently-maintained symlink-containment predicates in the codebase that could drift apart later. Worth a follow-up if `#1293` (making the write guard file-aware) touches this code anyway.
- `scanner.py` disclosed to now trip pylint's `C0302` (too-many-lines) while staying comfortably above the 9.0 gate (9.70) — flagged transparently in the PR body rather than hidden; a future split is reasonable but not urgent.

## CLAUDE.md Adherence

✅ TDD followed with an unusually rigorous RED-first process (before/after counts given, a load-bearing revert-and-rerun proof, and a second load-bearing proof for the dry-run ordering).
✅ Local gate reported green: 12/12 checks, 8749 passed, 94.38% branch coverage, per-file gate passed with no new waiver, mypy strict, ruff zero, pylint 9.70, bandit clean.
✅ No forbidden shortcuts disclosed (no new `noqa`/`type: ignore`/skip marks/threshold changes).
✅ Docstrings and cross-references updated alongside behavior change, consistent with the DRY/single-source-of-truth principle.
I was not able to execute `./scripts/check-all.sh` myself in this review sandbox (tool invocations for local Python/test execution were blocked), so I could not independently reproduce these numbers — but the static trace of the changed code paths was internally consistent with what's claimed.

## Testing

Very strong. Coverage spans the chokepoint (`test_redact.py`), CLI (`test_cli_redact.py`), MCP tool parametrized over every `TierCeiling` (`test_mcp_redact.py`), and pipeline (`test_pipeline.py`), plus breadth guards (in-root alias, dangling link, circular pair, symlinked directory, symlinked-root-component) that pin the fix isn't over-broad. The structural AST test with its negative control is a nice touch that most PRs of this shape wouldn't bother with. No issues found.

## Documentation

Thorough and consistent: `docs/redaction.md` gained a new section with a byte-verified error message and explicit statement of which existing override knobs do *not* apply; `docs/security/threat-model.md`'s SEC-003 entry now correctly describes three distinct walk behaviors instead of one blanket claim; `docs/mcp.md` and the two module docstrings were corrected in the same change rather than left to drift.



## Verdict: LGTM

The fix closes a real information-disclosure/path-traversal residual in the redaction scan walk, is applied at the single chokepoint all three callers share, is backed by an unusually thorough and behaviorally-meaningful test suite (including a negative-controlled AST guard ensuring no second unguarded walk can reappear), and every scope-narrowing judgement call is explicitly disclosed with a follow-up issue filed. I traced the key claims against the actual checked-out source (scan_directory delegates to scan_batch, ScanSummary is exported and its skip counter survives MCP relativization, the pipeline's refusal sits before the dry_run arm, the two error classes don't collide in creek/cli.py's except ordering) and found everything internally consistent with the PR description. I was not able to execute the test suite in this sandbox, but the static trace gives me confidence; the only note is a minor, already-disclosed DRY duplication between the two symlink-containment predicates, which is not blocking.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5235420695 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1300, opened 2026-08-10
Compact .id-index.jsonl: a torn line makes every future cold load pay a full directory scan, forever


Geoffe-Ga:
Split out of #1120. Now that index damage is *counted* rather than silently
skipped, there is a signal worth acting on and nothing yet acts on it.

## What accumulates

`.id-index.jsonl` is append-only by design — `_append_index_entry` never
rewrites the file, because a whole-file rewrite from one process's in-memory
snapshot would destroy every entry another process appended since that snapshot
was taken (documented on `_repair_index_locked`). Three kinds of dead weight
therefore accumulate forever:

1. **Superseded mappings.** Later entries win, so every re-write of an id leaves
   its previous line in place. `_find_existing_locked` already notes this: "the
   on-disk JSONL still contains the stale line ... a future compaction pass (out
   of scope for this fix) can reclaim the space."
2. **Torn records.** After #1120 a torn record is self-terminating and costs
   only itself — but it stays on disk as an unparseable line forever, and it
   makes the file permanently "damaged" from the reader's point of view.
3. **Stale mappings for tombed or deleted files**, dropped from the in-memory
   cache but never from the file.

## Why (3) is the one that bites

Point 2 has a cost #1120 introduced and did not pay down: `_load_index_locked`
treats *any* unparseable line as damage and re-derives by directory scan. That
scan is bounded at once per directory per process by the `_dir_indexes` cache,
so it is a real but bounded cost — except that a torn line is **never removed**,
so every future process, forever, pays one full `_rebuild_index` of that
directory on first access. One historical crash permanently converts an O(1)
cold load into an O(N) one.

At 35k fragments that is the exact cost the index exists to avoid, and it is
adjacent to #1290 (index cost at scale).

## Shape

A compaction pass — plausibly `creek lint --fix`-shaped, or triggered from the
recovery path once damage is observed — that rewrites the file **atomically**
from a fresh `_rebuild_index` scan plus any surviving parsed entries the scan
cannot see. `_persist_full_index` already does exactly this write (temp file
plus rename, so it can never tear).

The hard part is not the write, it is the concurrency: the prohibition on
whole-file rewrites exists because a rewrite races concurrent appenders. A
compaction pass needs either an explicit lock (`creek/audit/log.py` shows the
`fcntl`/`_HAS_FCNTL` pattern this module does not yet use), or an
offline/maintenance-window contract stated in the docs, or a
[...]

Issue #1299, opened 2026-08-10
A stalled (zero-byte) index append leaves no damage signal, so the dropped mapping is still permanent


Geoffe-Ga:
Split out of #1120, which is the fix for the *torn* index append. This is the
one residual that fix deliberately does not close, recorded rather than
smuggled in.

## The hole

#1120 made every `.id-index.jsonl` record self-delimiting (`\n{json}\n`) so a
partial append costs at most its own entry, and made `_load_index_locked`
re-derive by directory scan when the reader sees damage — an unparseable line
or an unreadable file. The recovery is therefore **triggered by evidence of
damage**.

A *stalled* append leaves no evidence. `creek/_fsio.write_all` raises
`OSError(EIO)` when `os.write` reports zero bytes of progress, and in that case
nothing at all reached the file. There is no remnant, no unparseable line, and
the index looks perfectly intact — it is simply missing a mapping.

Consequence: the pre-#1120 outcome survives on this one path. `_write_model`
sees no entry for the id, `_atomic_create`s a counter-suffixed sibling, and the
fragment exists twice. `tests/test_vault_writer.py::TestAtomicWriteHardening::test_append_index_entry_raises_when_write_makes_no_progress`
pins the loud failure; nothing pins the recovery, because there is none.

## Why #1120 did not close it

Both available shapes cost more than the hole is worth:

- **Rescan on every index miss.** Turns each unindexed-id lookup into a full
  `_rebuild_index` of the directory — quadratic ingest at 35k fragments, which
  is the exact cost PERF-001 bought the index to avoid, and it would break
  `tests/test_vault_writer.py::test_vanished_file_does_not_trigger_a_rescan`,
  the regression guard for that property.
- **Rework `_atomic_create`'s collision path** so a counter-suffixed create
  verifies the colliding file's declared id before minting a sibling. Plausible,
  but it is a change to the write path's core invariant and wants its own
  design pass.

## Mitigating factors

The stall path raises `OSError(EIO)` at the call site rather than corrupting
silently, so the caller *is* told. All three write callers (`tomb_fragment`,
`restore_fragment`, `_write_model`) let it escape while holding a correct
in-memory mapping, so the running process keeps resolving the id correctly —
only a fresh process is exposed. And a descriptor that makes literally zero
forward progress is a rarer failure than a short write.

## Acceptance criteria

- [ ] Decide between the two shapes above (or a third) with a measurement, not
[...]

PR #1307, opened 2026-08-10
fix(vault): a torn index append costs one entry, not two, and is recoverable

### LINKED ISSUE #1120: A failed mid-line .id-index.jsonl append corrupts the NEXT append too (both entries lost)

Geoffe-Ga:
## Summary

A torn `.id-index.jsonl` append used to cost **two** entries, not one, and the loss
was permanent. `_append_index_entry` writes under `O_APPEND` with the terminating
newline as part of the payload, so a partial write leaves a newline-less remnant at
the tail — and the next successful record lands directly against it. The merged line
is not valid JSON, so the reader skipped it wholesale and both the torn entry and the
innocent next one dropped out of the index.

The harm did not stop there. `_load_index_locked` rescanned the directory only when
the index file was **missing**, so an id absent from a *present* index was never
re-derived — not by this process, not by any future one. `_write_model` then saw no
existing file and `_atomic_create`d a **second `.md` for the same id**. Silent,
permanent duplication.

### Three things worth arguing up front

**1. The two halves have to land together.** Framing alone makes the headline
"next entry survives" test pass while leaving the duplicate-file criterion
unmet; recovery alone leaves the next entry still being swallowed. This is
measured, not asserted: reverting the whole source change turns exactly 9 tests
red, and reverting *only* the one-expression framing change turns 6 of those 9
red — the other 3 belong to the recovery half.

- **Framing (write side).** Each record is now encoded `"\n" + json + "\n"`
  instead of `json + "\n"`, so it is self-delimiting at its *start* as well as
  its end. A remnant is terminated by the next record's leading newline and
  costs at most its own entry. Both newlines are kept deliberately:
  leading-only is a byte cheaper, but a mixed-version straddle (old code
  appending `{b}\n` after a new-code `\n{a}`) would merge the two and
  reintroduce the defect on live vault data.
- **Recovery (read side).** A new `_IndexLoad` (`entries` / `torn_lines` /
  `read_failed`, with a `damaged` property) lets the reader distinguish "this
  index legitimately names nothing" from "this index was damaged and therefore
  names less than it should". When damaged, `_load_index_locked` logs a warning
  and re-derives by directory scan. That makes a dropped mapping recoverable
  instead of permanent.

**2. The accepted blind spot.** A *stalled* append that writes **zero** bytes
leaves no remnant, hence no damage signal, hence no rescan — a duplicate is
still possible on that path. It is the least bad of the options: that path
raises a loud `OSError(EIO)` from `creek/_fsio.py` rather than corrupting
silently, and closing it would mean either an O(N) rescan on every index miss
(quadratic ingest at 35k fragments, and it would break
`test_vanished_file_does_not_trigger_a_rescan`) or reworking `_atomic_create`'s
collision path. Filed as #1299 rather than smuggled in here.

**3. The concurrency false positive.** `write_all` loops over `os.write`, so a
healthy multi-byte append has a real window in which the file holds a partial
record with no exception in flight anywhere. A reader in another process can
observe that tail, classify it as damage, and pay one spurious `_rebuild_index`.
It is bounded at **one scan per directory per `VaultWriter` instance** by the
`_dir_indexes` cache (instance state — a caller that builds several writers pays
it once for each), and the merge is `setdefault` over the scan, so a spurious
rescan is a cost, never a correctness loss. It is also self-correcting: the index
append only ever runs *after* the corresponding `.md` is fully on disk, so a
scan triggered by a transient tear still resolves the right mapping.

### Design decisions, stated rather than implied

- **Precedence is scan-wins.** `_recover_damaged_index` merges
  `_rebuild_index(target_dir)` with the parsed entries via `setdefault`: any id a
  live file declares takes the on-disk truth, and parsed JSONL entries only fill
  gaps the scan cannot see. This is the same precedence `_repair_index_locked`
  already applies (`_rebuild_index(...).get(model_id)`). Parsed-wins would keep a
  mapping the scan just disproved, and every later lookup would pay verification,
  a second scan, and a repair append to reach the answer the load already had.
- **Damage is narrow on purpose.** Only a read `OSError` and a
  `json.JSONDecodeError` count. A line that is valid JSON but not a
  `{id: str, filename: str}` object is *not* damage — it is a well-formed line
  the schema declines, and treating it as damage would buy a directory-wide scan
  for a perfectly intact file.
- **No fsync, said out loud.** `_append_index_entry` still does not fsync, unlike
  `AuditLog.append`. The index is a reconstructible cache, not a ledger. That was
  only ever defensible paired with a recovery path — which this PR is what
  supplies.
- **The in-memory-before-on-disk ordering at all three write sites is preserved
  deliberately**, now with a comment at each saying so. The file is already
  moved/created by the time the append runs, so a mapping that outlives a torn
  append is the conservative side of the trade; reversing it silently would be a
  tomb/restore durability regression.
- **`_persist_full_index` is deliberately NOT re-framed.** It renames atomically
  and can never leave a torn record, so it needs no self-delimitation — and
  framing it would insert a blank line per entry, roughly doubling the line count
  of every first-time rebuild at vault scale.
- **`_load_index_file` keeps its exact signature** as a thin wrapper over
  `_read_index_records(...).entries`, so no caller moves.

### Properties preserved

- **O(1) per append.** No full-file read, no rescan, no directory walk on the
  write path. The `_PIPE_BUF_BYTES` guard measures `len(encoded)`, so it now
  counts the two framing bytes automatically — one byte stricter, the safe
  direction. Routing through `AuditLog` was rejected for exactly this reason: its
  `_compute_prev_hash` falls back to a full-file rescan whenever another process
  appended.
- **O(1) per clean load**, pinned by a test that makes `_rebuild_index` raise and
  asserts an undamaged load never reaches it.
- **`O_APPEND` atomicity, no `ftruncate`, no lock, no `fcntl`.** No whole-file
  rewrite is added anywhere; the `_persist_full_index` prohibition on the repair
  path stays true.
- **Backward compatibility, tested directly.** Every `.id-index.jsonl` in a live
  vault was written by the old encoder. Two tests pin it: a complete legacy file
  mixed with a new-format record resolves all three ids, and an *old-format* torn
  tail (no leading newline — the only remnant shape that can exist at upgrade
  time) is terminated by the new record's leading newline. No migration, no
  version marker, no vault-wide rewrite.

## Test plan

Gate 1 was RED on behavioural assertions, never on a missing symbol. With the
source reverted to `HEAD` and the new tests in place:

```
FAILED test_torn_index_append_does_not_swallow_the_next_entry - assert 1 >= 2
FAILED test_any_torn_tail_prefix_leaves_the_next_entry_parseable - AssertionError: lost at cut=2
FAILED test_a_legacy_format_remnant_does_not_swallow_the_next_entry - KeyError: 'frag-next'
FAILED test_consecutive_torn_tails_do_not_compound - KeyError: 'frag-healthy'
FAILED test_damaged_index_rescans_and_recovers_the_dropped_id - assert None == PosixPath(...)
FAILED test_unreadable_index_rescans_instead_of_looking_empty - assert None == PosixPath(...)
FAILED test_a_torn_index_tail_does_not_mint_a_duplicate_file - assert PosixPath(...-1.md) == PosixPath(...)
FAILED test_torn_repair_append_does_not_poison_the_index_it_repairs - KeyError: 'frag-bystander'
FAILED test_append_index_entry_writes_complete_line_under_short_writes - framing byte-pin
9 failed, 100 passed
```

Exactly those 9, nothing else. Reverting only the framing expression leaves 6 red.

New and changed tests:

- `test_torn_index_append_does_not_swallow_the_next_entry` — the #987
  characterization test **inverted in place**, per its own docstring's
  instruction. Never deleted. Same staging (`fail_after_half(ENOSPC)` →
  `passthrough()`); the three assertions that still hold are untouched, and the
  three that pinned the defect now pin the contract.
- `test_any_torn_tail_prefix_leaves_the_next_entry_parseable` — the forcing
  function. Writes **every byte prefix** of an encoded record straight to disk
  and appends over it. No exception is in flight anywhere: this is what a
  SIGKILL/OOM/power loss between two `os.write` iterations actually leaves. A fix
  that only cleaned up inside an `except OSError:` handler could not pass it.
- `test_a_legacy_format_remnant_does_not_swallow_the_next_entry` and
  `test_a_legacy_format_index_still_parses_unchanged` — the no-migration claim,
  tested rather than asserted.
- `test_consecutive_torn_tails_do_not_compound` — two remnants back to back, one
  old-shaped and one new-shaped, as two processes at different versions would
  leave them.
- `test_a_torn_index_tail_does_not_mint_a_duplicate_file` — the end-to-end
  consequence closed: after a torn tail, a second `write_fragment` for the same
  id returns the original path, the directory still holds exactly one `.md`, and
  `update_fragment` still finds it. Deliberately does **not** use the
  `short_write` fixture — `fail_after_half` fires on the first tracked
  descriptor, which in `write_fragment` is the fragment *body*, and
  `create_exclusive` unlinks that partial file, so the index append would never
  be reached.
- `test_torn_repair_append_does_not_poison_the_index_it_repairs` — **the fourth
  call site**, the one PR #1295 put on the read path. Goes through the real route
  (`_find_existing` → mismatch → `_repair_index_locked` → append), with the
  genuine file **present** and a different existing file indexed for that id —
  the only setup that actually reaches the append rather than the
  `del index[model_id]; return None` branch. Asserts the documented `OSError`
  still escapes the lookup, and that a healthy record written to the index
  afterwards is not swallowed by the remnant the repair left behind.
- `test_damaged_index_rescans_and_recovers_the_dropped_id`,
  `test_unreadable_index_rescans_instead_of_looking_empty`,
  `test_clean_index_never_rescans` — the recovery half, plus the O(1) guard.

- `test_undecodable_index_rescans_instead_of_crashing_the_write` — raised by
  self-review. `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so it
  escaped the read guard and would take down *every* vault write into the
  directory. Not reachable from a torn append (our writers emit ASCII only,
  since `json.dumps` escapes non-ASCII by default), but reachable from a
  hand-edited or externally corrupted index — and squarely inside the threat
  model this PR is about, so it is fixed here rather than deferred. Reverting
  just that clause turns exactly this one test red with the exception
  propagating uncaught.

The unreadable-index test also closes an **adjacent hole in the same reader**
(fixed here rather than deferred, because it produces the identical
duplicate-everything outcome as #1120 through a different door): a read `OSError`
returned `{}`, which `_load_index_locked` then cached *because the file existed*
— an unreadable index was indistinguishable from an empty directory, with no
rescan and no log line.

Gate 2: `./scripts/check-all.sh` → **exit 0**, 12/12 checks, 8758 passed,
aggregate coverage 94.39%, per-file gate green (`creek/vault/writer.py` is not
waivered, so the 80% floor applies to every new branch). Complexity measured, not
assumed: `_read_index_records` B(8), `_load_index_locked` A(5),
`_recover_damaged_index` A(3), `_append_index_entry` A(2) — no extraction needed.
`tests/test_ingest_update_in_place.py` and `tests/test_cli_fill.py` run clean;
they exercise the same locator through the continuous-ingest update-in-place path.

Every prose site that documented #1120 as an unfixable known defect has been
rewritten — leaving stale "KNOWN DEFECT" narration behind a landed fix is its own
failure.

## Self-review (Gate 2.5)

`code-review-orchestrator` over the full diff returned **CLEAN — no blockers**,
having hand-traced the `setdefault` merge for the tombed / non-`str`-id /
transient-unreadable cases against the pre-fix reader on `main` (additive only;
no new drop, duplication or misresolution path), the framed-vs-unframed boundary
where a `_persist_full_index` file is later appended to, and the repair-site
test's byte stream. Its three non-blocking nits are all addressed in this PR
rather than deferred: the "once per process" cache bound corrected to "once per
`VaultWriter` instance", the overstated "same precedence as
`_repair_index_locked`" claim tightened to name where the two differ, and the
`UnicodeDecodeError` gap closed with a test.

## Follow-ups filed

- **#1299** (P2) — a *stalled* zero-byte append leaves no remnant, hence no
  damage signal, hence no rescan. The one path where a duplicate is still
  possible; argued above.
- **#1300** (P2) — compaction. A torn line is never removed, so one historical
  crash permanently converts that directory's cold load from O(1) to O(N). This
  PR makes the damage observable; it does not reclaim it.
- Measured cost (+2 bytes and +1 line per appended record, plus the conditional
  read-side scan and its bound) [cross-referenced on #1290](https://github.com/Geoffe-Ga/Creek-Vault/issues/1290#issuecomment-5235178827),
  which is open about index cost at scale and will collide textually with this
  change in `_find_existing_locked` if both land in parallel.

Closes #1120

### DIFF
--- a/creek-tools/creek/vault/writer.py
+++ b/creek-tools/creek/vault/writer.py
@@ -26,7 +26,7 @@ import tempfile
 import threading
 from datetime import UTC, date, datetime
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple
 
 import frontmatter
 import yaml
@@ -436,6 +436,30 @@ def _file_declares_id(path: Path, model_id: str) -> bool:
         return False
     declared = post.get("id")
     return isinstance(declared, str) and declared == model_id
+
+
+class _IndexLoad(NamedTuple):
+    """The outcome of parsing one ``.id-index.jsonl`` file (#1120).
+
+    Separates *what the file said* from *whether the file was intact*,
+    so a caller can tell an index that legitimately names nothing from
+    one that was damaged and therefore names less than it should.
+
+    Attributes:
+        entries: The ``{id: filename}`` mappings that parsed cleanly.
+        torn_lines: How many lines failed ``json.loads``. A torn append
+            leaves exactly one; a crash-heavy history can leave several.
+        read_failed: ``True`` when the file could not be read at all.
+    """
+
+    entries: dict[str, str]
+    torn_lines: int
+    read_failed: bool
+
+    @property
+    def damaged(self) -> bool:
+        """Return ``True`` when *entries* may be missing a real mapping."""
+        return self.torn_lines > 0 or self.read_failed
 
 
 def _is_material_change(old_body: str, new_body: str, threshold: float) -> bool:
@@ -888,6 +912,12 @@ class VaultWriter:
             _atomic_write_text(dest, frontmatter.dumps(post))
             existing.unlink()
             index = self._load_index_locked(orphan_dir)
+            # Deliberate ordering, preserved by #1120: the in-memory
+            # mapping is set BEFORE the append, so a failed append leaves
+            # this process still resolving the tombed file (already moved
+            # above) while a fresh process re-derives it from the damage
+            # rescan in ``_load_index_locked``. Reversing this would make
+            # a torn append lose the tomb for the rest of the run.
             index[fragment_id] = dest.name
             self._append_index_entry(orphan_dir, fragment_id, dest.name)
             self._log_provenance_locked(fragment_id, "fragment", dest)
@@ -923,6 +953,10 @@ class VaultWriter:
             _atomic_write_text(dest, frontmatter.dumps(post))
             existing.unlink()
             index = self._load_index_locked(target_dir)
+            # Same deliberate ordering as ``tomb_fragment`` — see the
+            # note there; the restored file is already on disk (above),
+            # so an in-memory mapping that outlives a torn append is the
+            # conservative side of the trade (#1120).
             index[fragment.id] = dest.name
             self._append_index_entry(target_dir, fragment.id, dest.name)
             self._log_provenance_locked(fragment.id, fragment.type, dest)
@@ -1123,6 +1157,11 @@ class VaultWriter:
             # it, but a future refactor that skips the dup-check
             # branch would otherwise hit a ``KeyError`` here.
             index = self._load_index_locked(target_dir)
+            # In-memory first, on-disk second — deliberate, and preserved
+            # by #1120. ``_atomic_create`` above already put the note on
+            # disk, so a torn append leaves this process resolving the
+            # id correctly and a fresh process recovering it from the
+            # damage rescan rather than minting a duplicate.
             index[model_id] = file_path.name
             self._append_index_entry(target_dir, model_id, file_path.name)
 
@@ -1200,7 +1239,10 @@ class VaultWriter:
                 :meth:`_append_index_entry`. A lookup-shaped method can
                 therefore raise, because a mismatch persists its repair.
             OSError: If persisting a repaired entry fails to write — same
-                origin, same reason it can escape a lookup.
+                origin, same reason it can escape a lookup. Since #1120
+                the partial record such a failure leaves behind is
+                self-terminating, so a torn repair no longer damages the
+                index it was repairing.
         """
         index = self._load_index_locked(target_dir)
         filename = index.get(model_id)
@@ -1256,7 +1298,10 @@ class VaultWriter:
         Both propagate from :meth:`_append_index_entry`. The in-memory
         mapping is corrected *before* the append, so a failure to persist
         leaves this process holding ground truth and the on-disk index
-        merely stale — a cost, not a correctness loss.
+        merely stale — a cost, not a correctness loss. Across processes
+        the same holds since #1120: the torn record this path can leave
+        is self-delimiting, and the unparseable line it forms is exactly
+        the damage signal that makes the next load re-resolve by scan.
         """
         resolved = self._rebuild_index(target_dir).get(model_id)
         if resolved is None:
@@ -1292,13 +1337,46 @@ class VaultWriter:
         last successful write wins). If the JSONL file is missing the
         index is rebuilt by scanning the directory's ``.md`` files and
         the result is persisted so subsequent processes skip the scan.
+
+        When the JSONL file is *present but damaged* — an unparseable
+        line, or a file that cannot be read at all — the mappings it
+        failed to yield are re-derived by scanning the directory (#1120).
+        Without that, a mapping lost to a torn append was permanent: a
+        rescan only ever ran when the file was missing, so an id absent
+        from a present index was never re-resolved and the next write
+        minted a second file for it.
+
+        Precedence on the recovery path is **scan wins**: any id a live
+        file declares takes the on-disk truth, and the parsed JSONL
+        entries only fill gaps the scan cannot see (an id whose file
+        was tombed, or whose frontmatter the scanner declines). The
+        direction matches :meth:`_repair_index_locked`, which also
+        prefers what the directory says over what the index claimed —
+        though that method resolves *purely* by scan and drops the id
+        when the scan misses, where this one keeps the parsed entry as a
+        fallback rather than discarding a mapping it cannot disprove.
+        Letting the parsed value win instead would preserve a mapping the
+        scan just disproved, and every later lookup would pay
+        verification, a second scan and a repair append to arrive at the
+        answer this load already had.
+
+        The recovery never calls :meth:`_persist_full_index`: a
+        whole-file rewrite from this process's snapshot would destroy
+        entries another process appended. The recovered index is cached
+        in ``self._dir_indexes``, so the scan is paid at most once per
+        directory per :class:`VaultWriter` instance — not once per
+        process, since a caller that builds several writers pays it once
+        for each.
         """
         cached = self._dir_indexes.get(target_dir)
         if cached is not None:
             return cached
         index_path = target_dir / INDEX_FILENAME
         if index_path.exists():
-            index = self._load_index_file(index_path)
+            load = self._read_index_records(index_path)
+            index = load.entries
+            if load.damaged:
+                index = self._recover_damaged_index(index_path, target_dir, load)
         elif target_dir.is_dir():
             index = self._rebuild_index(target_dir)
             self._persist_full_index(target_dir, index)
@@ -1308,6 +1386,34 @@ class VaultWriter:
         return index
 
     @staticmethod
+    def _recover_damaged_index(
+        index_path: Path,
+        target_dir: Path,
+        load: _IndexLoad,
+    ) -> dict[str, str]:
+        """Re-derive a damaged index by scanning *target_dir* (#1120).
+
+        Args:
+            index_path: The damaged JSONL file, named in the warning.
+            target_dir: The directory whose ``.md`` files are the truth.
+            load: The partial parse, whose surviving entries fill any
+                gap the scan cannot see.
+
+        Returns:
+            The merged index, with the disk scan taking precedence.
+        """
+        logger.warning(
+            "index %s has %d unparseable line(s)%s; re-resolving by directory scan",
+            index_path,
+            load.torn_lines,
+            " and could not be read" if load.read_failed else "",
+        )
+        recovered = VaultWriter._rebuild_index(target_dir)
+        for model_id, filename in load.entries.items():
+            recovered.setdefault(model_id, filename)
+        return recovered
+
+    @staticmethod
     def _load_index_file(index_path: Path) -> dict[str, str]:
         """Parse a JSONL index file into ``{id: filename}``.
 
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Design provenance (orchestrator note)

For the reviewer's benefit, since it is not stated in the PR body: this lane did
**not** run an in-lane `chief-architect` pass. That was my instruction, not an
omission.

The architecture came from a pre-computed dossier produced by a
premise-verification and design-judging workflow run against HEAD `389c61d`:
two independent designs were proposed, each adversarially refuted with a
concrete failure mode, and a hybrid winner was written up as a dispatch-ready
plan. That dossier is what the worker executed. It caught three places where a
candidate plan's test would have been silently wrong, and it is why this PR
fixes the read-path call site added by #1295 — which the issue body does not
mention.

`code-review-orchestrator` then ran as the independent pre-push review
(returned CLEAN; three non-blocking nits, all fixed rather than deferred).

So there *was* an adversarial design pass — it happened before dispatch rather
than inside the lane. Reviewing this as if it were unplanned would be the wrong
frame; reviewing the plan itself is fair game.

Geoffe-Ga:
## Summary
Fixes issue #1120: a torn `.id-index.jsonl` append under `O_APPEND` used to leave a newline-less remnant that merged with the *next* successful append, corrupting both lines and permanently dropping the mapping (since a rescan only ever ran when the index file was *missing*, not when it was present-but-damaged). The fix has two halves: (1) each appended record is now framed as `\n{json}\n` so a remnant is self-terminating and costs at most its own entry, and (2) `_load_index_locked` now detects a damaged index (unparseable line, unreadable file, or undecodable bytes) and re-derives the missing mappings via directory scan instead of silently trusting an incomplete file.

## Strengths
- Both halves are independently justified and tested — the PR description states that reverting only the framing change still leaves 6/9 new tests red, and reverting only the recovery half leaves the other 3 red. I didn't re-run the suite (sandbox blocked `pytest` execution), but traced the logic by hand and it holds up.
- `test_any_torn_tail_prefix_leaves_the_next_entry_parseable` is a genuinely strong test: it writes every byte-prefix of an encoded record directly to disk (no exception in flight), which is what a SIGKILL/OOM/power-loss actually leaves — a fix that only handled the `except OSError:` path could not pass it.
- Backward compatibility is handled correctly and tested: the reader already skipped blank lines, so legacy (unframed) index files and a mix of legacy/new-format records both parse unchanged, and an old-format torn remnant is still terminated by a new-format append's leading newline.
- Closes two real adjacent holes in the same reader in the same PR rather than deferring them: an unreadable index used to return `{}` and get cached as if the directory were empty (permanent duplication via a different door than #1120), and `UnicodeDecodeError` (a `ValueError`, not `OSError`) used to escape the read guard entirely and crash every write into the directory.
- `test_clean_index_never_rescans` pins the O(1)-per-clean-load property by making `_rebuild_index` raise if called — a real regression guard, not just an assertion in prose.
- Deliberately scoped: two known residual gaps (a zero-byte stalled append leaving no damage signal, and lack of compaction for a permanently-damaged file) are explicitly named and filed as follow-ups (#1299, #1300) rather than silently left implicit. I confirmed both issues exist and match the PR's description.
- `_read_index_records`/`_IndexLoad` cleanly separates "the file names nothing" from "the file was damaged," and `_recover_damaged_index`'s scan-wins-with-parsed-entries-as-fallback precedence is the same direction `_repair_index_locked` already uses elsewhere in this file.

## Security Concerns
None. 🟢 No new file/network I/O, no path traversal, no injection surface. `UnicodeDecodeError` handling closes a minor availability issue (an externally-corrupted index could previously crash every write into a directory) rather than introducing one.

## Problems
None found that block merge.

## Code Quality
- Minor cosmetic nit: the warning log format string is `"index %s has %d unparseable line(s)%s..."`, so it always renders literal `line(s)` regardless of count (e.g. "has 1 unparseable line(s)"). Not worth blocking on — purely a grammar nit, and the test only checks a substring.
- The recovery path intentionally never repairs the on-disk file (no `_persist_full_index` call), so a permanently damaged index pays a full directory rescan on every cold process start until compaction (#1300) lands. This is explicitly called out as a known, deliberate tradeoff rather than an oversight, so it's not a ding against this PR — just flagging it's real production cost until the follow-up lands.

## CLAUDE.md Adherence
✅ TDD: description documents RED tests failing against the pre-fix code, then GREEN after the fix, with a targeted-revert isolation for each of the two halves.
✅ Commit message follows Conventional Commits (`fix(vault): ...`).
✅ Docstrings are extensive (Google-style, matching this file's existing very-verbose convention) and explain non-obvious invariants (why leading+trailing newlines, why scan-wins precedence, why no fsync, why `_persist_full_index` isn't reframed).
✅ No `noqa`/`type: ignore`/threshold-lowering shortcuts visible in the diff.
⚠️ Could not independently execute `./scripts/check-all.sh` in this review sandbox (command execution was blocked), so the claimed 94.39% coverage / complexity / lint numbers are taken from the PR description and partial CI results rather than independently reproduced. The CI jobs that had completed at review time (Integration & E2E Tests, Code Complexity Analysis, Build Distribution, CrawDad Quality x3) were all green; the primary quality/test/lint jobs were still in progress.

## Testing
Strong. 9 new/inverted tests cover: the inverted #987 characterization test, the exhaustive byte-prefix forcing function, legacy-format compatibility (remnant and mixed-format), consecutive torn tails, end-to-end duplicate-file prevention, the fourth call site added by #1295 (`_repair_index_locked`), unreadable/undecodable index handling, and a rescan-avoidance regression guard for the clean-load path. Test naming and docstrings clearly state intent and the failure scenario each test targets.

## Documentation
Thorough — arguably more thorough than most PRs in this repo, matching the file's existing convention of long, decision-justifying docstrings. The PR body itself documents design decisions (scan-wins precedence, why no fsync, why `_persist_full_index` stays unframed) that would otherwise live only in someone's head. All prose in the source that used to describe #1120 as an unfixable known defect has been rewritten to reflect the fix, per the PR description — spot-checked several of these sites in the diff and confirmed they were updated, not left stale.



## Verdict: LGTM

The fix directly addresses issue #1120 with both required halves (self-delimiting framing on write, damage-triggered rescan on read), is covered by 9 new/inverted tests that fail cleanly against a revert (verified the referenced issues #1120, #1299, #1300 genuinely exist and match the PR's narrative), preserves O(1) steady-state behavior with a regression guard, and handles the legacy on-disk format without migration. No correctness bug, security issue, or CLAUDE.md violation found in static review; only a trivial cosmetic nit (log message always prints "line(s)" regardless of count) that isn't worth blocking on.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5235245458 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1319, opened 2026-08-10
chore(ingest): record the declined voice_weight=0.0 stamp for non-self document authors (#1213 follow-up)


Geoffe-Ga:
## Context

#1213 fixed the voice corpus: `creek.generate.voice._eligible_register` now refuses any fragment whose `source.author` is not `self`, so AI-, other- and collaborative-authored prose can no longer train the voice proxy. That closes the exposure.

This issue records the **belt-and-braces decision** #1213 deliberately did *not* take, because one of its acceptance criteria required the decision be tracked somewhere other than a PR-body paragraph.

## The question

`creek/ingest/documents.py` (`_write_fragment` frontmatter builder, the `_resolve_document_author` block) sets `source["author"] = Authorship.OTHER` when a DOCX `core_properties.author` / PDF `/Author` names somebody, but never sets `voice_weight` — so the model default of `1.0` applies (`creek/models.py`, `Fragment.voice_weight`).

`creek/ingest/claude.py` and `creek/ingest/chatgpt.py` do the opposite for their AI turns: they set `source["author"] = ai` **and** stamp `frontmatter_dict["voice_weight"] = 0.0`, explicitly as belt-and-braces.

Should `DocumentIngestor` do the same?

## Decision: NO, declined — with reasoning

Not a symmetry oversight. The two cases differ in kind:

- In `claude.py` / `chatgpt.py`, `author=ai` is a **certainty** about an immutable fact — that turn was generated by a model and always will have been. A redundant zero costs nothing, because there is no future in which the fact is corrected.
- In `documents.py`, `OTHER` is an explicit **guess**. `_resolve_document_author`'s own docstring says so: *"guessing OTHER for the owner's own document merely under-uses it — and the returned name is what makes that recoverable."* The fail-closed guess is correct; its recoverability is the point.

Nothing in the pipeline recomputes `voice_weight`. So a stamped `0.0` would mean: an operator who notices their own DOCX was attributed to `other`, corrects `source.author` to `self`, and expects it back in their voice corpus gets **nothing** — the stale `voice_weight: 0.0` keeps refusing it, with no visible cause and no tooling that would ever clear it. The guess would have become permanent through a side effect nobody documented.

Under #1213 the authorship gate already refuses the fragment for exactly as long as the guess stands, and stops refusing it the moment the guess is corrected. That is the behaviour we want.

## Revisit predicate

Reopen this decision if **either** becomes true:

[...]

Issue #1212, opened 2026-08-07
fix(voice): at the default ALL ceiling the consent gate reads the model tier — untiered fragments enter the voice corpus


Geoffe-Ga:
## The hole

The voice corpus has two additive gates (`creek/generate/voice.py`, `collect_exemplars`):

1. the **ceiling** gate, `within_ceiling(raw, self.override)` — reads the **raw**
   frontmatter and fails closed to `INTIMATE` when the `privacy_tier` key is *absent*
   (`privacy_filter.raw_privacy_tier`);
2. the **consent** gate, `_eligible_register(fragment, allow_intimate=...)` — reads
   `fragment.privacy_tier`, i.e. the **validated model**.

At `PrivacyTierOverride.ALL` the ceiling gate short-circuits:

```python
# creek/classify/privacy_filter.py, tier_within_override
effective = override or PrivacyTierOverride.OPEN
if effective is PrivacyTierOverride.ALL:
    return True
```

`ALL` is the default on every CLI report path — a bare `creek report` resolves
`override or PrivacyTierOverride.ALL`, and `creek fill` states `ALL` explicitly. So on
the ordinary path the raw fail-closed read is bypassed entirely and the **only**
surviving gate reads the model, where a missing `privacy_tier` key Pydantic-defaults to
`unclassified` (`creek/models.py`), which is not `intimate` — so the fragment is
admitted.

Net: **a fragment file with no `privacy_tier` key at all enters the voice corpus on a
default run.** Verified in the worktree — a fragment written with the key stripped, on a
bare `creek report --type voice`:

```
bare run copies: ['_Summary.md', 'frag-untiered.md']
```

This is the hazard `creek/classify/privacy_filter.py`'s own module docstring names:
untiered content "used to rank alongside `open` — which ... meant whole private corpora
were mineable, draftable and **voice-proxy eligible** at the open tier."

## Blast radius

`_eligible_register` is shared by `collect_exemplars`, `collect_all_exemplars` and
`VoiceProfileGenerator._stream_into`, so this admits untiered fragments to **all four**
voice-corpus consumers: `report --type voice` (profiles + register samples),
`report --type lexicon`, and `report --type rhetorical-patterns`. The lexicon is the
sharpest of these — `_build_borrowed_terms` records the whole surrounding sentence
verbatim into `glossary.md`.

## Why it was not fixed in #879

#879 wires the previously-callerless `save_exemplars` into `creek report --type voice`.
[...]

PR #1323, opened 2026-08-10
fix(voice): refuse non-self-authored prose at the voice-corpus gate (#1213)

### LINKED ISSUE #1213: fix(voice): _eligible_register ignores Fragment.voice_proxy_eligible — AI-authored prose can enter the voice corpus
## The divergence

`creek/models.py` defines the canonical predicate:

```python
@property
def voice_proxy_eligible(self) -> bool:
    return (self.privacy_tier != PrivacyTier.INTIMATE
            and self.source.author == Authorship.SELF)
```

and `docs/ingestion.md` states the guarantee it backs: *"Because
`Fragment.voice_proxy_eligible` excludes non-`self` authors, the assistant's prose can
never train the voice proxy."*

`creek/generate/voice.py`'s `_eligible_register` — the gate that actually decides voice
corpus admission — re-implements only the `INTIMATE` half and **omits the
`source.author == Authorship.SELF` clause entirely**. `voice_proxy_eligible` has real
callers in `generate/voice_authenticity.py` and `generate/skills.py`, but none in
`generate/voice.py`, which is where it would matter most.

## Why it does not currently bite, and why that is fragile

The Claude and ChatGPT ingesters set `voice_weight=0.0` as a documented belt-and-braces
measure, and `_eligible_register` *does* gate on `voice_weight`, so today's AI-authored
content is excluded by the backup mechanism rather than the stated one. But
`Fragment.voice_weight` defaults to `1.0`, so any fragment with `source.author: ai` and
no explicit `voice_weight` — legacy, hand-edited, or written by a future or third-party
ingester — is admitted to the voice corpus.

There is also a defensive path gate for `11-Other-Authors/` (`_is_other_authors_path`,
issue #466), which catches borrowed content by *location*. Neither of the two working
mechanisms is the one the documentation names.

After #879 the consequence is larger: an admitted fragment is no longer only excerpted
into a rendered profile, it is `shutil.copy2`'d byte-for-byte into
`07-Voice/Register-Samples/<register>/`. AI prose copied into the voice-proxy training
corpus is the specific drift #466 and the `voice_weight=0.0` measures exist to prevent.

## Suggested fix

Have `_eligible_register` consult `fragment.voice_proxy_eligible` rather than
re-deriving a subset of it, so the documented guarantee and the enforced behaviour are
the same expression. This changes which fragments are collected for four report types
(profiles, rhetorical-patterns, lexicon, register-samples), so measure the delta against
a populated vault before landing.

Interacts with #1212, which changes the other half of the same predicate — worth
sequencing the two deliberately rather than in parallel.

## Acceptance criteria

- [ ] A fragment with `source.author: ai` and a default `voice_weight` is excluded from
      the voice corpus.
- [ ] The `voice_weight=0.0` and `11-Other-Authors/` gates stay in place as
      belt-and-braces — this replaces neither.
- [ ] Measured against a populated vault: report how many fragments the change newly
      excludes.

Refs #879

Geoffe-Ga:
## Summary

`creek.generate.voice._eligible_register` is the single chokepoint for the entire voice corpus. It gated on `voice_weight`, confidence, privacy tier and register — and never on `source.author`. So the guarantee `Fragment.voice_proxy_eligible` documents ("only self-authored, non-INTIMATE fragments are eligible … AI / collaborator / other-authored content is excluded") was never enforced by the corpus that depends on it, and `docs/ingestion.md`'s promise that "the assistant's prose can never train the voice proxy" was aspirational.

This PR adds the missing clause, first in the function:

```python
if str(fragment.source.author) != Authorship.SELF.value:
    return None
```

`Authorship` was not previously imported into `creek/generate/voice.py` — the clause was never written, not merely mis-ordered.

### Three corrections to the issue body

**1. The body's suggested fix would have caused a regression.** It proposes having `_eligible_register` "consult `fragment.voice_proxy_eligible` rather than re-deriving a subset of it." That breaks `allow_intimate=True`: the property bakes in `privacy_tier != INTIMATE`, but `_eligible_register` takes `allow_intimate` precisely so an operator who opted the voice proxy in to intimate content still gets *their own* intimate fragments. Delegating wholesale silently revokes that opt-in.

This PR re-derives instead, mirroring `creek.generate.skills._is_snapshot_fragment`, whose docstring already documents refusing to delegate for exactly this reason. (Cited by symbol name, not line number — PR #1286 just moved it.) There is a dedicated regression test for this, and it is green both before and after the fix.

**2. "Why it doesn't currently bite" is false — there is a live, shipping producer.** `creek/ingest/documents.py` resolves a DOCX `core_properties.author` / PDF `/Author` to `Authorship.OTHER` and never sets `voice_weight`, so the model default of `1.0` applies. Those fragments carry no `author_slug`, so `VaultWriter._fragment_target_dir` routes them to `01-Fragments/Documents/` — inside the scan root, outside the `11-Other-Authors/` path gate. Reproduced end-to-end (below).

**3. The body names only `ai`; the predicate excludes three values.** `voice_proxy_eligible` requires `author == SELF`, so `ai`, `other` **and** `collaborative` are all excluded. The actually-reachable live case is `other`. Every new test covers all three.

Also worth recording: the `11-Other-Authors/` path gate is close to inert in the real vault layout. The walks only rglob `01-Fragments/`, while borrowed fragments live at vault-root `11-Other-Authors/<slug>/` — outside the scan root entirely; the segment check only ever catches a nested or symlinked occurrence. So `voice_weight > 0` was effectively the sole live gate. Both are retained deliberately as backstops (they still catch a fragment whose authorship axis is missing or mis-set) and are byte-identical in this diff.

### Live-path reproduction

A throwaway DOCX carrying `core_properties.author = "Jane Stranger"`, through the real `DocumentIngestor`, written where `VaultWriter` would put it, then handed to the real collector:

```
--- frontmatter DocumentIngestor produced ---
source.author      : other
source.author_name : Jane Stranger
source.author_slug : <absent>
voice_weight       : <absent -> model default 1.0>
voice_weight on the validated model : 1.0
--- voice corpus ---
path written       : 01-Fragments/Documents/stranger-docx.md
admitted to corpus : ['stranger-docx']      # before the fix
VERDICT            : ADMITTED (bug reproduced)
```

After the fix, the same script on the same input: `admitted to corpus : []` → `EXCLUDED (fixed)`.

### The chokepoint covers four surfaces, via two direct callers and three walks

Getting the enumeration right, since it is easy to miscount: the module-level `_eligible_register` has exactly **two direct callers** — the `VoiceExemplarCollector._eligible_register` forwarder, and `VoiceProfileGenerator._stream_into`. `collect_exemplars` and `collect_all_exemplars` reach it through the forwarder. That is **three walks** over `01-Fragments/`, feeding **four user-visible surfaces**, and this PR asserts the exclusion at each:

| Surface | Walk | Asserted in |
|---|---|---|
| `report --type voice` → profiles | `_stream_into` | `test_generate_all_profiles_excludes_non_self_authors` |
| `report --type voice` → register-samples | `collect_exemplars` | `test_register_samples_are_not_written_for_non_self_authors` |
| `report --type lexicon` | `collect_all_exemplars` | `test_generate_lexicon_excludes_non_self_authors` |
| `report --type rhetorical-patterns` | `_stream_into` | `test_generate_rhetorical_patterns_excludes_non_self_authors` |

The lexicon case is the sharpest: `_build_borrowed_terms` records the **whole surrounding sentence** around a detected keyword, so before this fix a third party's prose was written verbatim into `07-Voice/Lexicon/glossary.md`. The test asserts the sentence is gone.

### Vault-delta measurement (read-only)

Measured against the populated demo vault by walking `01-Fragments/` directly — **not** by running `creek report`, which mutates `07-Voice/` and fires the prune:

```
fragments scanned: 35330
authors: {'self': 33585, 'ai': 1745}
non-self with voice_weight > 0: 0
admitted pre-fix, per register: {'conversational': 10672, 'playful': 1709, 'raw': 1112,
                                 'analytical': 709, 'instructional': 642,
                                 'confessional': 337, 'prophetic': 269}
NEWLY EXCLUDED by #1213, per register: {}
NEWLY EXCLUDED total: 0
```

Honestly zero on that vault: all 1,745 `ai` fragments already carry `voice_weight: 0.0` from `creek/ingest/claude.py`, so the weight backstop was already refusing them, and it contains no `other`/`collaborative` fragments. The zero is not evidence the bug is harmless — the DOCX reproduction above is the case that carries the argument, and it is one `creek ingest --type documents` away.

### Behaviour change on merge (please read before merging)

- The first post-fix `creek report --type voice` **deletes** previously-written non-self register samples via the #879 prune. Correct, and announced through `on_prune`, but irreversible. `creek fill` runs this **unattended**. The manifest guard is what stops the prune touching operator-curated files in `Register-Samples/` and must not be weakened.
- Anything trained on `07-Voice/` — the Writing Desk / `creek draft` voice node, the voice skill tree, CrawDad's voice-core mirror — produces different prose after this merge.
- `creek/generate/ai_style/fingerprint.py` is **unaffected**: it walks fragments itself and already filters `source.get("author") != _SELF`.
- The change is purely **narrowing**: the corpus can now only ever admit less, so the privacy one-way ratchet holds.

### Sequencing with #1212 — do not run both lanes in parallel

#1212 (OPEN, P0, `do-not-auto-merge`, no PR yet) edits the **same function** and the **same three walks**: its candidate fix inserts a raw-tier consent check adjacent to where this clause goes, and rewrites the intimate branch this clause sits beside. Land #1213 first (smaller, purely additive, does not move the tier read), then rebase #1212 onto it. Whichever lands second must re-derive its vault-delta measurement, because the first changes the denominator.

Relatedly, the new cross-product test's tier axis is deliberately `[OPEN, PERSONAL, INTIMATE]` and **not** `list(PrivacyTier)`. `UNCLASSIFIED` is omitted with an explanatory comment in the test, because #1212 will make `_eligible_register` stop admitting a self-authored untiered fragment — at which point the equivalence arm stops holding for that cell while the property still reports `True`. Widening it is green today and a landmine tomorrow.

### `documents.py` decision

Declined, with reasoning, and filed rather than buried in this paragraph: **#1319**.

Short version: in `claude.py`/`chatgpt.py`, `author=ai` is a certainty about an immutable fact, so a redundant `voice_weight=0.0` costs nothing. In `documents.py`, `OTHER` is an explicit *guess* — the docstring says "guessing OTHER for the owner's own document merely under-uses it — and the returned name is what makes that recoverable" — and nothing recomputes `voice_weight`. A stamped zero would leave an operator who later corrects `author: other` to `self` on their own document silently frozen out of the corpus, with no visible cause. The authorship gate refuses the fragment exactly as long as the guess stands, and stops the moment it is corrected. #1319 records the revisit predicate.

## Test plan

**Proven red.** The fix was committed first, then `git restore --source=5dfd278 -- creek/generate/voice.py` reverted the source alone (no `git stash` — the stash ref is shared across live worktrees), and the final test set was run against it:

```
24 failed, 150 passed in 6.37s

FAILED test_excludes_non_self_authors_at_default_voice_weight[ai]
    - AssertionError: assert ['frag-authored-ai'] == []
FAILED test_excludes_non_self_authors_at_default_voice_weight[other]
    - AssertionError: assert ['frag-authored-other'] == []
FAILED test_excludes_non_self_authors_at_default_voice_weight[collaborative]
    - AssertionError: assert ['frag-authored-collaborative'] == []
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[open-ai]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[open-other]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[open-collaborative]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[personal-ai]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[personal-other]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[personal-collaborative]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[intimate-ai]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[intimate-other]
FAILED test_gate_matches_the_canonical_property_and_the_intimate_override[intimate-collaborative]
FAILED test_collect_all_exemplars_excludes_non_self_authors[ai|other|collaborative]
    - AssertionError: assert [Exemplar(...)] == []
FAILED test_register_samples_are_not_written_for_non_self_authors
    - AssertionError: assert {'instructional': .../_Summary.md} == {}
FAILED test_previously_written_non_self_samples_are_pruned
    - AssertionError: assert [] == [PosixPath('.../frag-prune-other.md')]
FAILED test_generate_all_profiles_excludes_non_self_authors[ai|other|collaborative]
    - AssertionError: assert 'Zarquon' not in '---\nexempl...llet points.'
FAILED test_generate_rhetorical_patterns_excludes_non_self_authors
    - assert '- Self-deprecation before insight: 0.' in "..."
FAILED test_generate_lexicon_excludes_non_self_authors[ai|other|collaborative]
    - AssertionError: assert 'zarquon' not in '---\nborrow... ever wrote.'
```

Every failure is a **behavioural assertion on admitted content** — the fragment present in the bucket, the borrowed sentence present in the glossary — not an `ImportError` or `AttributeError`. `Authorship`, `FragmentSource.author` and all three values already existed at base.

**The 12-cell equivalence contract** (`list(Authorship)` × `[OPEN, PERSONAL, INTIMATE]`), which is the forcing function against reintroduction:

- Before the fix: **9 non-self cells red, all 3 `self` cells green.**
- After the fix: **all 12 green.**

`list(Authorship)` is deliberate — a fifth member added later is excluded by the property and must be excluded here too, without anyone remembering to extend a hand-written list.

**The regression that the wrong fix would have caused.** `test_self_authored_intimate_admitted_when_allow_intimate` — a self-authored INTIMATE fragment with `allow_intimate=True` — **passes both before and after**, and so do all three `self` cells of the cross-product including `self` + INTIMATE + `allow_intimate=True`. Nothing that was admitted before is refused now.

**Load-bearing check.** Reverting the clause alone turns exactly those 24 cells red and leaves the other 150 tests in the three files green.

**Blast radius.** Nine voice-corpus test files, not the three the issue implies — plus the e2e lane, which walks `01-Fragments/` for `author == "ai"` and then calls `generate_all_profiles`:

```
tests/test_voice_exemplars.py tests/test_voice_profiles.py tests/test_lexicon.py
tests/test_cli.py tests/test_mcp_report_tier_ceiling.py tests/test_purge_voice_artifacts.py
tests/test_voice_audience_authority.py tests/test_voice_citation_density.py
tests/test_voice_patterns.py tests/test_voice_proxy_eligible_computed.py
    → 500 passed

./scripts/test.sh --e2e   → 14 passed
```

No existing test was deleted, skipped or loosened. Three test helpers gained a default-valued `author` parameter; `_build_fragment` additionally changed `voice_weight` from `float = 1.0` to `float | None = None` so a test can exercise the *model* default rather than one the helper supplied — the distinction this bug turns on. Every existing call site is unaffected.

**Gate 2** — `./scripts/check-all.sh` exit 0:

```
=== Quality Checks Summary ===
Passed: 12
Failed: 0
✓ All quality checks passed!

8804 passed, 5 skipped, 370 deselected
Required test coverage of 90% reached. Total coverage: 94.38%
✓ 256 files >= 80.0%, 2 waivered files >= 65.0%
```

`_eligible_register` measures `B (7)` after the added branch (radon), clearing the real gate `xenon --max-absolute B --max-modules B --max-average B`. No threshold changed, no `# noqa`, no `# type: ignore`, no skip. Nothing under `crawdad/` is touched.

Closes #1213
Refs #1212, #1319

### DIFF
--- a/creek-tools/creek/generate/voice.py
+++ b/creek-tools/creek/generate/voice.py
@@ -13,6 +13,16 @@ body texts to extract sentence structure, paragraph structure, transition
 patterns, metaphor families, rhetorical moves, vocabulary fingerprint, and
 punctuation habits.
 
+**Voice fidelity** is guarded separately from privacy, by three additive
+gates in :func:`_eligible_register` and the walks around it. The primary
+one is authorship (#1213): a fragment whose ``source.author`` is not
+``self`` is refused outright, which is what enforces the guarantee
+:attr:`~creek.models.Fragment.voice_proxy_eligible` documents. Behind it
+sit two backstops from #466 — ``voice_weight > 0``, stamped to ``0.0`` on
+AI turns at ingest, and the ``11-Other-Authors`` path exclusion
+(:data:`_OTHER_AUTHORS_SEGMENT`). Both are retained deliberately: they
+catch a fragment whose authorship axis is missing or mis-set.
+
 The collector is deliberately conservative about privacy, and two
 independent gates say so. ``allow_intimate`` is a **consent** gate: the
 voice proxy is opt-in even at ``--include-tier intimate``
@@ -52,6 +62,7 @@ from pydantic import ValidationError
 from creek.classify.privacy_filter import PrivacyTierOverride, within_ceiling
 from creek.config import VoiceAudienceWeightingConfig
 from creek.models import (
+    Authorship,
     Confidence,
     Fragment,
     Frequency,
@@ -84,6 +95,15 @@ _OTHER_AUTHORS_SEGMENT: str = "11-Other-Authors"
 Any fragment whose source path contains this segment is excluded from the
 voice corpus regardless of its ``voice_weight``, as a defensive backstop
 against voice drift.
+
+Since #1213 this is a backstop to a real frontmatter gate rather than a
+front-line defence, and it is **retained deliberately** — it still catches
+a fragment whose ``source.author`` axis is absent or mis-set. It was never
+load-bearing on its own: the walks only rglob :data:`_FRAGMENTS_SUBDIR`,
+while :meth:`creek.vault.writer.VaultWriter._fragment_target_dir` routes
+borrowed fragments to a vault-root ``11-Other-Authors/<slug>/`` that is
+outside the scan root entirely. What this segment check actually catches
+is a nested or symlinked occurrence *inside* ``01-Fragments/``.
 """
 
 _SAMPLES_SUBPATH: tuple[str, str] = ("07-Voice", "Register-Samples")
@@ -302,14 +322,36 @@ def _eligible_register(
             fragments are rejected.
 
     Returns:
-        The canonical register string when the fragment has qualifying
-        confidence, a positive ``voice_weight``, allowed privacy tier,
-        and a canonical register; otherwise ``None``.
+        The canonical register string when the fragment is self-authored
+        and has qualifying confidence, a positive ``voice_weight``, allowed
+        privacy tier, and a canonical register; otherwise ``None``.
     """
-    # Voice-fidelity fail-closed gate (Issue #466): borrowed / AI-authored
-    # text (notably ``ai-as-user`` content with ``voice_weight=0.0``) must
-    # never train the voice proxy. ``voice_weight`` is already coerced to a
-    # float in ``[0, 1]`` by the model, so a plain comparison is safe.
+    # Authorship gate (Issue #1213), first because it is the primary one:
+    # this is what enforces the guarantee
+    # :attr:`~creek.models.Fragment.voice_proxy_eligible` documents — that
+    # AI-, collaborator- and other-authored prose can never train the voice
+    # proxy. Before this clause the corpus consulted no authorship axis at
+    # all, and the two gates below were the whole defence.
+    #
+    # Re-derived rather than delegated to the property, exactly as
+    # :func:`creek.generate.skills._is_snapshot_fragment` is and for the
+    # same documented reason: the property bakes in the INTIMATE exclusion,
+    # while ``allow_intimate=True`` deliberately overrides it, so
+    # ``return fragment.voice_proxy_eligible`` would silently revoke the
+    # operator's opt-in to their *own* intimate writing.
+    #
+    # Compared by value, not identity: ``FragmentSource`` sets
+    # ``use_enum_values=True``, so ``author`` is a plain ``str`` at runtime
+    # — the same idiom as the privacy-tier comparison below. Deliberately
+    # not ``Authorship(...) is Authorship.SELF``, which raises on a
+    # hand-corrupted frontmatter value instead of failing closed.
+    if str(fragment.source.author) != Authorship.SELF.value:
+        return None
+    # Voice-fidelity backstop (Issue #466): ``ai-as-user`` content is
+    # stamped ``voice_weight=0.0`` at ingest, which refuses it a second
+    # time even if its authorship axis is ever mis-set. ``voice_weight`` is
+    # already coerced to a float in ``[0, 1]`` by the model, so a plain
+    # comparison is safe.
     if fragment.voice_weight <= 0:
         return None
     if _confidence_value(fragment) not in _QUALIFYING_CONFIDENCE:
@@ -578,9 +620,10 @@ class VoiceExemplarCollector:
     def collect_exemplars(self, vault_path: Path) -> dict[str, list[Fragment]]:
         """Scan ``01-Fragments/`` and group qualifying exemplars by register.
 
-        Fragments are kept when their ``voice.confidence`` is ``settled``
-        or ``conviction`` and their ``voice.register`` is set. Intimate
-        privacy tier fragments are filtered out unless
+        Fragments are kept when they are **self-authored** (#1213), their
+        ``voice.confidence`` is ``settled`` or ``conviction``, their
+        ``voice_weight`` is positive, and their ``voice.register`` is set.
+        Intimate privacy tier fragments are filtered out unless
         :attr:`allow_intimate` is ``True``, and every fragment must
         additionally sit within :attr:`override` (#968). Each register key
         in the returned dict is always present; empty registers map to
@@ -631,8 +674,9 @@ class VoiceExemplarCollector:
     def collect_all_exemplars(self, vault_path: Path) -> list[Exemplar]:
         """Collect every qualifying exemplar across all registers, with bodies.
 
-        Shares the eligibility gate with :meth:`collect_exemplars` (``settled``
-        or ``conviction`` confidence, a set register, intimate filtered unless
+        Shares the eligibility gate with :meth:`collect_exemplars`
+        (self-authored, ``settled`` or ``conviction`` confidence, positive
+        ``voice_weight``, a set register, intimate filtered unless
         :attr:`allow_intimate`, within :attr:`override`, ``11-Other-Authors``
         excluded) but returns a flat :class:`Exemplar` list — fragment paired
         with its on-disk body — rather than per-register ``Fragment`` buckets.
@@ -670,7 +714,8 @@ class VoiceExemplarCollector:
         """Return the fragment's register if it qualifies, else ``None``.
 
         This answers the *consent* question only (``allow_intimate``, plus
-        confidence / voice-weight / register eligibility). The *ceiling*
+        authorship / confidence / voice-weight / register eligibility).
+        The *ceiling*
         question is asked separately by
         :func:`~creek.classify.privacy_filter.within_ceiling` at each walk,
         because it needs the raw frontmatter this method never sees. Do not
--- a/creek-tools/creek/ingest/chatgpt.py
+++ b/creek-tools/creek/ingest/chatgpt.py
@@ -167,7 +167,9 @@ class ChatGPTIngestor(Ingestor):
             source["conversation_id"] = conv_id
         if is_ai:
             # AI turns are AI-authored, excluded from the voice proxy via
-            # ``voice_proxy_eligible``; zero weight is belt-and-braces.
+            # ``voice_proxy_eligible``; zero weight is belt-and-braces. The
+            # authorship half is enforced by
+            # ``creek.generate.voice._eligible_register`` (#1213).
             source["author"] = _ROLE_AI
         frontmatter_dict: dict[str, Any] = {
             "title": fragment.metadata.get("title", "Untitled Conversation"),
--- a/creek-tools/creek/ingest/claude.py
+++ b/creek-tools/creek/ingest/claude.py
@@ -365,7 +365,11 @@ class ClaudeIngestor(Ingestor):
         if is_ai:
             # The AI turn is AI-authored, so it can never feed the voice proxy
             # (``source.author=ai`` makes ``voice_proxy_eligible`` False); zero
-            # weight is belt-and-braces against future selectors.
+            # weight is belt-and-braces against future selectors. The voice
+            # corpus enforces the authorship half in
+            # ``creek.generate.voice._eligible_register`` (#1213); before that
+            # it read only the weight, and this comment described a mechanism
+            # nothing consulted.
             source["author"] = _ROLE_AI
 
         suffix = ", AI" if is_ai else ""
--- a/creek-tools/docs/ingestion.md
+++ b/creek-tools/docs/ingestion.md
@@ -93,7 +93,7 @@ The caveat: a source keyed on mtime treats *any* filesystem touch as a change, n
 
 ## AI-chat attribution (per turn)
 
-Ingesting a Claude or ChatGPT conversation emits **two attributed fragments per turn**: the human turn is the owner's voice (`source.author = self`, full `voice_weight`) and the AI turn is AI-authored (`source.author = ai`, `voice_weight = 0.0`). Because `Fragment.voice_proxy_eligible` excludes non-`self` authors, the assistant's prose can never train the voice proxy, while threading/linking survives — both turns share the conversation id, turn index, and timestamp.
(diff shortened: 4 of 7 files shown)


Geoffe-Ga:
## Summary
Adds the missing authorship check to `creek.generate.voice._eligible_register`, the single chokepoint that gates all four voice-corpus surfaces (voice profiles, register samples, lexicon, rhetorical patterns). Previously the gate checked `voice_weight`, confidence, privacy tier, and register, but never `source.author`, so the guarantee `Fragment.voice_proxy_eligible` documents was not actually enforced by the corpus. The PR demonstrates a live, reachable bug path through `DocumentIngestor` (DOCX/PDF authorship metadata → `Authorship.OTHER`, no `voice_weight` stamped, routed inside the scan root) and closes it with a single re-derived (not delegated) `if` clause plus comment/doc updates and a large, well-targeted test addition.

## Strengths
- Root-caused precisely: identifies that `_eligible_register` never consulted `source.author`, and that the two pre-existing backstops (`voice_weight`, `11-Other-Authors` path segment) were not actually load-bearing against the real attack surface (an unattributed document).
- Correctly avoids the regression the linked issue's suggested fix would have caused — delegating to `fragment.voice_proxy_eligible` would silently revoke the `allow_intimate=True` opt-in for the operator's own INTIMATE writing. This PR re-derives the check instead, mirroring the precedent in `creek.generate.skills._is_snapshot_fragment` (verified: that function does exactly this, for the same reason, and its docstring says so).
- Verified the routing claim directly against `creek/vault/writer.py`: `_fragment_target_dir` keys off `source.author_slug`, not `source.author`, so an `OTHER`-authored fragment with no slug really does land inside `01-Fragments/`, outside the `11-Other-Authors/` path gate.
- Purely narrowing change — cannot cause a previously-excluded fragment to become admitted; only previously-admitted (buggy) fragments become excluded.
- Test plan documents a genuine red→green cycle (source reverted via `git restore`, not just narrated) and enumerates exactly which 24 tests fail without the fix and why.
- Behavioral consequence on merge (the #879 prune deleting previously-written non-self register samples on the next `creek report --type voice`) is proactively disclosed rather than left as a surprise.

## Security Concerns
None. 🟢 This is itself a fix for an information-boundary bug (third-party/AI prose leaking into a generated voice corpus and, in the lexicon case, into files with verbatim borrowed sentences). No new attack surface is introduced.

## Problems
None found. No blocking issues.

## Code Quality
- The inline comments in `_eligible_register`, `chatgpt.py`, and `claude.py` are quite long (multi-paragraph) for what are otherwise small clauses. This matches the repo's existing dense-comment style elsewhere in `voice.py` (e.g., the module docstring, `_is_other_authors_path`), so it's consistent rather than an outlier, but it's worth flagging as a readability tradeoff — non-blocking.
- `str(fragment.source.author) != Authorship.SELF.value` — the `str()` wrap is redundant given `FragmentSource` already sets `use_enum_values=True` (confirmed in `creek/models.py`), but it mirrors the existing privacy-tier comparison idiom two lines below (`str(fragment.privacy_tier) == PrivacyTier.INTIMATE.value`), so it's intentional consistency, not sloppiness.

## CLAUDE.md Adherence
✅ TDD followed (PR body documents a genuine red-tests-first cycle with `git restore` proof, not just claimed).
✅ No forbidden shortcuts — no `noqa`/`type: ignore` additions, no threshold changes, no skipped tests.
✅ Conventional commit format (`fix(voice): refuse non-self-authored prose at the voice-corpus gate (#1213)`), correct scope.
✅ `check-all.sh` output pasted shows 0 failures, 94.38% coverage, all 12 checks passed (not independently re-run in this review due to sandbox constraints, but the reported numbers are consistent with the size/nature of the diff and no red flags surfaced during static verification).
✅ Docstrings updated in step with behavior changes (`_eligible_register`, `collect_exemplars`, `collect_all_exemplars`, `VoiceExemplarCollector._eligible_register`).

## Testing
Strong, not just sufficient. Highlights:
- A 12-cell equivalence matrix (`list(Authorship)` × `[OPEN, PERSONAL, INTIMATE]`) pinning both the default and `allow_intimate=True` admission semantics against the canonical property — with a documented, deliberate reason for excluding `UNCLASSIFIED` from the tier axis (avoiding a landmine against a follow-up PR, #1212).
- Independent coverage of all four consumer surfaces (`collect_exemplars`, `collect_all_exemplars`, register-sample generation + its prune-on-reattribution behavior, and the `_stream_into` walk feeding both voice profiles and rhetorical patterns) rather than asserting only at the shared gate.
- A dedicated regression test (`test_self_authored_intimate_admitted_when_allow_intimate`) directly encodes the "don't delegate to the property" invariant, so a future refactor toward `return fragment.voice_proxy_eligible` would be caught immediately.
- Test helper changes (`voice_weight: float = 1.0` → `float | None = None`) are narrow and don't alter behavior for any existing call site not touched by this PR.

## Documentation
`docs/ingestion.md` is updated to state the guarantee is now enforced rather than aspirational, with a correct, specific description of scope (`other`/`collaborative`, not just `ai`). Module docstring and `_OTHER_AUTHORS_SEGMENT` docs in `voice.py` are updated to correctly reframe the two pre-existing gates as backstops rather than primary defenses, matching the PR's own analysis of why the path-segment gate is largely inert in the real vault layout.



## Verdict: LGTM

The fix is correct, precisely scoped, and re-derives the pattern already established by `_is_snapshot_fragment` for the same documented reason (preserving the `allow_intimate` opt-in). Verified independently against the live source: `documents.py` really does resolve unattributed DOCX/PDF authors to `Authorship.OTHER` without stamping `voice_weight`, and `VaultWriter._fragment_target_dir` really does route on `author_slug` (not `author`), so the exploit path described in the PR body is real, not hypothetical. Test coverage is exhaustive (12-cell equivalence matrix, per-surface assertions across all four consumers, a dedicated regression test for the intimate-opt-in trap, and a prune-on-reattribution test) and the diff outside `voice.py`/tests is comment/doc-only. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5235301563 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1325, opened 2026-08-10
fix(vault): relocation failure modes after #1302 — exhaustion aborts an ingest run, and a failed unlink leaks an unindexed duplicate


Geoffe-Ga:
Follow-up from the #1302 Gate 2.5 self-review. Neither item is a data-loss defect; both
are failure-mode ergonomics on the relocation path that #1302 deliberately left alone so
the data-loss fix stayed targeted.

`#1302` routed `VaultWriter.tomb_fragment` / `restore_fragment` through the shared
`_relocate_fragment_locked`, which allocates the destination with `_atomic_create`
(`O_CREAT | O_EXCL`, counter-suffix retry, `_MAX_FILENAME_COLLISION_RETRIES = 10_000`).

## 1. A `RuntimeError` at counter-suffix exhaustion aborts the whole ingest run

`creek/ingest/pipeline.py` `tomb_missing_units` catches only `(OSError, KeyError)` around
`writer.tomb_fragment(...)`, collecting the failure into `errors` and moving on to the
next unit. `restore_tombed` wraps `writer.restore_fragment(...)` in no handler at all.

`_atomic_create` can raise `RuntimeError` when it cannot allocate a unique filename within
10 000 attempts. That path did not exist before #1302 (the old code was an unconditional
`os.replace`, which never fails on collision), so one pathological directory would now
abort an entire ingest run rather than degrade to one collected per-unit error.

This is deliberately loud — for a data-integrity path, failing loudly beats continuing on
a bad assumption — and 10 000 collisions on one stem is effectively unreachable. But the
decision is currently only findable by reading the `Raises:` section of the two
docstrings, which is what this issue exists to fix.

Decide one of:
- widen `tomb_missing_units` to `(OSError, KeyError, RuntimeError)` so exhaustion is
  collected like any other per-unit failure, and give `restore_tombed` a handler; or
- keep the abort and cite this issue from both docstrings so the choice is traceable.

## 2. An `OSError` from `existing.unlink()` leaks an unindexed duplicate

In `_relocate_fragment_locked`, if `_atomic_create` succeeds and `existing.unlink()` then
raises, the method aborts before the origin pop, the destination index assignment and the
provenance append. The newly-created destination file is left on disk **unindexed** — a
permanent orphan duplicate.

This is the safe side of the trade (the source is untouched, no bystander is destroyed,
nothing is lost) and the same ordering existed before #1302, so it is not a regression.
[...]

PR #1350, opened 2026-08-10
fix(vault): tomb and restore allocate a free filename instead of overwriting (#1302)

### LINKED ISSUE #1302: fix(vault): tomb_fragment and restore_fragment overwrite a colliding destination filename — silent fragment loss
**Severity:** Critical
**Remediation:** fix (route both relocations through the existing collision-safe helper)

## Problem

`VaultWriter.tomb_fragment` and `VaultWriter.restore_fragment` relocate a fragment by
computing `dest = <dir> / existing.name` and writing it with `_atomic_write_text`, which
is `mkstemp` + `os.replace` — an **unconditional overwrite**. Every other create path in
the same module goes through `_atomic_create` → `creek._fsio.create_exclusive`
(`O_CREAT|O_EXCL`) with a counter-suffix retry, which exists precisely so "two concurrent
callers picking the same filename will not clobber each other" (`vault/writer.py:1449-1451`).

The filename stem is `{date}-{sanitised title}`, which is **not** unique across platform
subfolders — and `10-Liminal/Orphaned/` is a single flat sink for all of them. So tombing
two same-titled, same-dated fragments from different platforms destroys the first one's
body permanently. Taxonomy family 0 (correctness) + family 12 (inconsistent-with-house-style).

**Current state:** silent, unrecoverable fragment loss on a live write path.
`tomb_fragment` is called from `creek/ingest/pipeline.py:293` and `restore_fragment` from
`creek/ingest/pipeline.py:208`, so a routine mutable-source re-ingest triggers it.

- `creek-tools/creek/vault/writer.py:886-889` (tomb)
- `creek-tools/creek/vault/writer.py:921-924` (restore)

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Two distinct fragments, same title + date, different
platforms; tomb both:

```
written to distinct dirs:
   01-Fragments/Journal/2026-08-09-Notes.md
   01-Fragments/Messages/2026-08-09-Notes.md

files in Orphaned after tombing BOTH: ['2026-08-09-Notes.md']
surviving contains 'BODY OF A': False
surviving contains 'BODY OF B': True
frag-a body anywhere in vault: False

id-index (both ids -> one filename):
   {"filename": "2026-08-09-Notes.md", "id": "frag-aaaaaaaaaaaa"}
   {"filename": "2026-08-09-Notes.md", "id": "frag-aaaaaaaaaaaa"}
   {"filename": "2026-08-09-Notes.md", "id": "frag-bbbbbbbbbbbb"}
```

`frag-a` was unlinked from `01-Fragments/Journal/` and then overwritten in `Orphaned/`.
Its body is gone from the vault entirely. The id-index ends up mapping **both** ids to the
same filename (and records `frag-a` twice).

Repro script (runs in seconds, no network):
```python
import tempfile, pathlib, datetime as dt
from creek.vault.writer import VaultWriter
from creek.models import Fragment, FragmentSource, SourcePlatform
vault = pathlib.Path(tempfile.mkdtemp())
for d in ("00-Creek-Meta", "01-Fragments", "10-Liminal"):
    (vault / d).mkdir(parents=True, exist_ok=True)
w = VaultWriter(vault_path=vault)
when = dt.datetime(2026, 8, 9, tzinfo=dt.UTC)
a = Fragment(id="frag-aaaaaaaaaaaa", title="Notes",
             source=FragmentSource(platform=SourcePlatform.JOURNAL), created=when, ingested=when)
b = Fragment(id="frag-bbbbbbbbbbbb", title="Notes",
             source=FragmentSource(platform=SourcePlatform.DISCORD), created=when, ingested=when)
w.write_fragment(a, body="BODY OF A - irreplaceable")
w.write_fragment(b, body="BODY OF B")
w.tomb_fragment("frag-aaaaaaaaaaaa")
w.tomb_fragment("frag-bbbbbbbbbbbb")
orph = vault / "10-Liminal" / "Orphaned"
assert len(list(orph.glob("*.md"))) == 1          # the clobber
assert not any("BODY OF A" in p.read_text() for p in vault.rglob("*.md"))
```

A mirror repro exists for `restore_fragment`: tomb `frag-a`, ingest a new `frag-b` that
reuses the freed stem, then `restore_fragment(a)` destroys `frag-b`.

**Signal 2 — structural.** These two methods are the *only* writes in `vault/writer.py`
that bypass `_atomic_create`. The collision-safe helper already exists in the same file
(`writer.py:1475-1488`) and every sibling create path uses it. This is a missed call site,
not a design decision — nothing in either docstring mentions overwrite semantics.

## Scope

Covers the two relocation methods in `creek/vault/writer.py` and the index bookkeeping
that follows them.

Explicitly NOT in scope: `.id-index.jsonl` append durability (#1120, #1299, #1300),
non-string frontmatter ids (#1291), alias precedence (#1224), tier re-derivation (#1136).
Those are separate mechanisms on the same file.

## Tasks

1. In `tomb_fragment` (`vault/writer.py:886-889`), replace
   `dest = orphan_dir / existing.name` + `_atomic_write_text(dest, ...)` with
   `_atomic_create(orphan_dir, existing.stem, frontmatter.dumps(post))` so a collision
   takes a counter suffix.
2. Record the path `_atomic_create` actually returned in the index and provenance calls
   (currently `dest.name` is assumed).
3. Do the same in `restore_fragment` (`vault/writer.py:921-924`) against `target_dir`.
4. Drop the source id from the origin directory's in-memory index after the `unlink()`,
   so a tombed id no longer resolves to a deleted path.
5. Add regression tests for both directions using the repro above.

## Acceptance Criteria

- [ ] Tombing two fragments whose routed filenames collide leaves **two** files in
      `10-Liminal/Orphaned/`, and both bodies are readable.
- [ ] Restoring a fragment whose destination stem is now occupied leaves both files intact.
- [ ] `.id-index.jsonl` in the destination directory maps each id to a distinct filename.
- [ ] The origin directory's index no longer resolves the relocated id.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/vault/writer.py` | Modify |
| `creek-tools/tests/test_vault_writer.py` | Modify (add both regression tests) |

Geoffe-Ga:
## Summary

`VaultWriter.tomb_fragment` and `VaultWriter.restore_fragment` relocated a fragment by
composing `dest = <dir> / existing.name` and writing it with `_atomic_write_text` —
`mkstemp` + `os.replace`, an **unconditional overwrite** — then unlinking the source.
Every other create path in the module already went through `_atomic_create` →
`creek._fsio.create_exclusive` (`O_CREAT | O_EXCL`) with a counter-suffix retry, which
exists precisely so "two concurrent callers picking the same filename will not clobber
each other". These two were the only writes that bypassed it.

A filename stem is `{date}-{sanitised title}`, which is unique only *within* one
`01-Fragments/<platform>/` subfolder — while `10-Liminal/Orphaned/` is a single flat sink
fed by all of them. So the collision is reachable on a routine mutable-source re-ingest
(`creek/ingest/pipeline.py:293` tombs, `:208` restores), and it destroyed a fragment body
permanently and silently.

Reproduced live at the pre-fix HEAD, both directions:

```
written: 01-Fragments/Journal/2026-08-09-Notes.md
         01-Fragments/Messages/2026-08-09-Notes.md
files in Orphaned after tombing both: ['2026-08-09-Notes.md']
surviving tombstone contains: BODY OF B
BODY OF A anywhere in the vault: False
```

and the mirror — tomb C, ingest a D that reuses the freed stem, restore C, and D's bytes
are gone.

### The fix

A new private `VaultWriter._relocate_fragment_locked` owns the relocation for both
directions; the two public methods keep only what genuinely differs (how `existing` is
located, and whether the frontmatter marker is stamped or cleared). The helper allocates
through `_atomic_create`, so an arriving file takes a `-N` counter suffix instead of
replacing the one already there.

Four orderings in it are load-bearing, and each is pinned by a test:

- **Create the destination before unlinking the source.** `_atomic_create` has two ways
  to raise that `_atomic_write_text` never had — `RuntimeError` at counter-suffix
  exhaustion, and `OSError` from a short write that `create_exclusive` cleans up.
  Create-first means both escape with the source file fully intact: the worst case is a
  recoverable duplicate, never a fragment that exists nowhere.
- **Everything downstream consumes the path `_atomic_create` returned** — the index
  entry, the `_append_index_entry` argument, the provenance path and the return value.
  Getting the destination right but recording the pre-move name would have traded a
  silent overwrite for a silent orphan.
- **The origin's in-memory index entry is dropped after the unlink and before the
  destination assignment.** `self._dir_indexes` is keyed by `Path`, so were origin and
  destination ever the same directory, popping second would delete the mapping just
  written. The on-disk origin `.id-index.jsonl` is deliberately left alone: a
  `_persist_full_index` rewrite from this process's snapshot would destroy entries
  another process appended, and the stale line self-heals through the #1083 verification
  path.
- **The destination's in-memory mapping is set before the on-disk append**, preserved
  from #1120. That rationale previously existed as two hand-maintained copies inside the
  two methods; it now lives once in the helper docstring with a pointer at each call
  site.

`existing.stem` is used as the base name, never `_compute_base_name(fragment)`:
`restore_fragment` does hold a `Fragment`, but a title that drifted since the original
write would silently rename the file out from under every `[[wikilink]]` pointing at it.
Suffix stacking (`-1-1`) therefore only happens under a genuine live collision, which is
exactly when a distinguishing suffix is required — an uncontended tomb → restore round
trip is name-stable, because the tomb frees the origin name before the restore asks for
it back. A test locks that.

`_atomic_write_text` stays: `update_fragment` uses it for an in-place rewrite, where an
unconditional overwrite is the correct semantics.

## Test plan

Eight new tests in `TestTombRestoreFilenameCollisions`
(`tests/test_ingest_orphan_tomb.py`). Four failed at HEAD on behavioural assertions
about destroyed bytes; the rest lock the surviving half of the contract and the
failure-path orderings.

Each guard was then proven load-bearing by reverting **only** it and confirming exactly
the intended tests go red (mutations applied and reverted byte-identical from a saved
copy — no `git stash`, which is a shared ref across live worktrees):

| Mutation | Result |
|---|---|
| `_atomic_create` → back to `_atomic_write_text` on a composed path | 4 fail |
| drop the origin-index `.pop` | exactly 1 fails |
| record `existing.name` instead of the returned `dest.name` | exactly 1 fails |
| reorder to unlink-before-create | exactly 2 fail (tomb + restore) |

The third and fourth mutations initially **survived** and the gaps were closed before
this PR. The `existing.name` mis-indexing was invisible because #1083's
`_find_existing_locked` transparently repairs a mis-mapped entry on the next lookup, so
every lookup-based assertion self-heals; the test that catches it parses
`.id-index.jsonl` and `provenance.jsonl` raw, deliberately bypassing the healing readers.

Gate 2: `./scripts/check-all.sh` — **literal exit code 0**. 12/12 checks, 8793 passed,
5 skipped, aggregate branch coverage 94.39%, per-file gate green (256 files ≥ 80%).
No `# noqa`, no `# type: ignore`, no skips, no threshold changes, no deleted tests.

Gate 2.5 (`code-review-orchestrator` over the diff): **CLEAN**, no blockers. Its three
nits are addressed below — one new symmetric test, and a tracking issue cited from the
docstrings so the deliberate decisions are findable outside the source.

## Acceptance criteria

- [x] Tombing two fragments whose routed filenames collide leaves **two** files in
      `10-Liminal/Orphaned/`, and both bodies are readable.
- [x] Restoring a fragment whose destination stem is now occupied leaves both files intact.
- [x] `.id-index.jsonl` in the destination directory maps each id to a distinct filename
      (asserted against the raw file, not the self-healing loader).
- [x] The origin directory's index no longer resolves the relocated id.
- [x] No existing tests break; coverage stays ≥ 90%.

## Noted, deliberately not changed — filed as #1325

- `tomb_missing_units` (`creek/ingest/pipeline.py:293`) catches only `(OSError, KeyError)`
  and `restore_tombed` (`:208`) catches nothing, so the counter-suffix exhaustion
  `RuntimeError` would abort a whole ingest run rather than being collected as a per-unit
  error. At 10 000 retries it is effectively unreachable and loud by design, which is the
  safe direction for a data-integrity path — but widening the handler is a behaviour
  change outside this issue.
- If `_atomic_create` succeeds and `existing.unlink()` then raises, the new destination
  file is left on disk unindexed. Nothing is lost and no bystander is harmed, and the
  same ordering predates this PR, so it is not a regression — but it is not fully closed
  either.

Both are now cited from the docstrings as `#1325` rather than being discoverable only by
reading the source.

Closes #1302

### DIFF
--- a/creek-tools/creek/vault/writer.py
+++ b/creek-tools/creek/vault/writer.py
@@ -884,6 +884,96 @@ class VaultWriter:
                 return found
         return None
 
+    def _relocate_fragment_locked(
+        self,
+        existing: Path,
+        dest_dir: Path,
+        post: frontmatter.Post,
+        model_id: str,
+        model_type: str,
+    ) -> Path:
+        """Move *existing* into *dest_dir*, never overwriting what is there.
+
+        Caller must hold ``self._lock``. Shared by :meth:`tomb_fragment`
+        and :meth:`restore_fragment`, which differ only in how they locate
+        *existing* and in the frontmatter marker they stamp or clear; the
+        relocation itself — create, unlink, re-index, log — is identical
+        and lives here (#1302). Both directions previously composed
+        ``<dir> / existing.name`` and wrote it with an unconditional
+        ``os.replace``, so a same-named file at the destination was
+        silently destroyed. Allocation now goes through
+        :meth:`_atomic_create`, the same ``O_CREAT | O_EXCL`` path every
+        other create in this module uses.
+
+        The stem handed to :meth:`_atomic_create` is ``existing.stem``,
+        never :meth:`_compute_base_name`: a title that drifted since the
+        original write would rename the file out from under every
+        ``[[wikilink]]`` pointing at it. Suffix stacking (``-1-1``)
+        therefore only occurs under a genuine live collision, which is
+        exactly when a distinguishing suffix is required; an uncontended
+        tomb → restore round trip is name-stable, because the tomb frees
+        the origin name before the restore asks for it back.
+
+        Three orderings below are load-bearing:
+
+        - The destination is created **before** the source is unlinked.
+          :meth:`_atomic_create` can raise — retry exhaustion, or a short
+          write that :func:`creek._fsio.create_exclusive` cleans up — and
+          create-first means both escape with the source file fully
+          intact. The worst case is a recoverable duplicate, never a
+          fragment that exists nowhere. The mirror case — a create that
+          succeeded followed by an ``unlink`` that failed — leaks an
+          unindexed copy at the destination for the same reason, which
+          is the same safe direction and is tracked in #1325.
+        - The origin's in-memory entry is dropped **after** the unlink and
+          **before** the destination entry is set. ``self._dir_indexes``
+          is keyed by :class:`~pathlib.Path`, so were origin and
+          destination ever the same directory, popping second would
+          delete the mapping just written. The on-disk origin JSONL is
+          deliberately left alone: a :meth:`_persist_full_index` rewrite
+          from this process's snapshot would destroy entries another
+          process appended (the rejection is argued in
+          :meth:`_repair_index_locked` and :meth:`_load_index_locked`),
+          and the stale line is harmless — a fresh process re-resolves it
+          through the #1083 verification path.
+        - The destination's in-memory mapping is set **before** the
+          on-disk append, preserved from #1120. The relocated file is
+          already on disk by then, so a torn append leaves this process
+          still resolving the moved file while a fresh process re-derives
+          it from the damage rescan in :meth:`_load_index_locked`.
+          Reversing this would lose the relocation for the rest of the run.
+
+        Args:
+            existing: The live file to move; unlinked once its copy at the
+                destination exists.
+            dest_dir: Directory to move into, created if absent.
+            post: The already-mutated frontmatter document to write.
+            model_id: Id to re-point at the relocated file.
+            model_type: Type string recorded in the provenance log.
+
+        Returns:
+            The path actually created, which carries a ``-N`` counter
+            suffix when the stem was already taken in *dest_dir*.
+
+        Raises:
+            RuntimeError: If a unique filename cannot be allocated within
+                :data:`_MAX_FILENAME_COLLISION_RETRIES` attempts.
+            OSError: If the destination cannot be written, or its index
+                entry cannot be appended.
+        """
+        dest_dir.mkdir(parents=True, exist_ok=True)
+        dest = self._atomic_create(dest_dir, existing.stem, frontmatter.dumps(post))
+        origin_dir = existing.parent
+        existing.unlink()
+        # ``pop``, not ``del``: ``_find_existing_locked`` may already have
+        # healed this entry away on its repair path (#1083).
+        self._load_index_locked(origin_dir).pop(model_id, None)
+        index = self._load_index_locked(dest_dir)
+        index[model_id] = dest.name
+        self._append_index_entry(dest_dir, model_id, dest.name)
+        self._log_provenance_locked(model_id, model_type, dest)
+        return dest
+
     def tomb_fragment(self, fragment_id: str) -> Path | None:
         """Soft-tomb a fragment: move it to ``10-Liminal/Orphaned/`` (#674).
 
@@ -893,11 +983,34 @@ class VaultWriter:
         the fragment. Returns the tombed path, or ``None`` when no live fragment
         maps to the id (already moved/removed).
 
+        ``10-Liminal/Orphaned/`` is a single flat sink fed by every
+        ``01-Fragments/<platform>/`` subfolder, while a filename stem is unique
+        only *within* one of those subfolders — so tombstones genuinely
+        collide. The move therefore runs through
+        :meth:`_relocate_fragment_locked`, which gives the arriving tombstone a
+        counter suffix instead of replacing the one already there (#1302). The
+        returned path is authoritative: it, and not ``<orphan dir>/<old
+        name>``, is what the id index now points at.
+
         Args:
             fragment_id: Id of the fragment to soft-tomb.
 
         Returns:
-            Path to the tombed file, or ``None`` if no live fragment was found.
+            Path to the tombed file, or ``None`` if no live fragment was
+            found. May carry a ``-N`` counter suffix when the name was taken.
+
+        Raises:
+            RuntimeError: If a unique filename cannot be allocated in the
+                orphan directory within
+                :data:`_MAX_FILENAME_COLLISION_RETRIES` attempts.
+            OSError: If the tombstone cannot be created, or its index entry
+                cannot be appended.
+
+        Both propagate from :meth:`_relocate_fragment_locked`.
+        :func:`creek.ingest.pipeline.tomb_missing_units` catches only
+        ``(OSError, KeyError)`` around this call, so the exhaustion
+        ``RuntimeError`` aborts the whole ingest run — deliberately loud, and
+        out of scope to widen here. Tracked in #1325.
         """
         orphan_dir = self.vault_path.joinpath(*_ORPHANED_RELPARTS)
         with self._lock:
@@ -907,21 +1020,15 @@ class VaultWriter:
             post = frontmatter.load(str(existing))
             post["lifecycle"] = "orphaned"
             post["orphaned_at"] = datetime.now(UTC).isoformat()
-            orphan_dir.mkdir(parents=True, exist_ok=True)
-            dest = orphan_dir / existing.name
-            _atomic_write_text(dest, frontmatter.dumps(post))
-            existing.unlink()
-            index = self._load_index_locked(orphan_dir)
-            # Deliberate ordering, preserved by #1120: the in-memory
-            # mapping is set BEFORE the append, so a failed append leaves
-            # this process still resolving the tombed file (already moved
-            # above) while a fresh process re-derives it from the damage
-            # rescan in ``_load_index_locked``. Reversing this would make
-            # a torn append lose the tomb for the rest of the run.
-            index[fragment_id] = dest.name
-            self._append_index_entry(orphan_dir, fragment_id, dest.name)
-            self._log_provenance_locked(fragment_id, "fragment", dest)
-        return dest
+            # Create-before-unlink and the #1120 in-memory-before-append
+            # ordering both live in the helper; see its docstring.
+            return self._relocate_fragment_locked(
+                existing,
+                orphan_dir,
+                post,
+                fragment_id,
+                "fragment",
+            )
 
     def restore_fragment(self, fragment: Fragment) -> Path | None:
         """Un-tomb a fragment: move it back from ``10-Liminal/Orphaned/`` (#674).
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
Fixes a real, reproduced data-loss bug: `VaultWriter.tomb_fragment` and `VaultWriter.restore_fragment` both composed their destination as `<dir> / existing.name` and wrote it via `_atomic_write_text` (`mkstemp` + unconditional `os.replace`), so a same-stemmed file already at the destination — reachable in practice because `10-Liminal/Orphaned/` is a single flat sink fed by every per-platform `01-Fragments/<platform>/` subfolder — was silently destroyed. The fix extracts a shared `_relocate_fragment_locked` helper that allocates through the module's existing `_atomic_create` (`O_CREAT|O_EXCL` + counter-suffix retry), giving a colliding arrival a `-N` suffix instead of clobbering the incumbent.

## Strengths
- Root-caused correctly: identifies that `tomb_fragment`/`restore_fragment` were the *only* two write paths in the module bypassing `_atomic_create`, and fixes both via one shared helper rather than patching each independently.
- The three load-bearing orderings (create-before-unlink; origin-index pop after unlink but before destination-index set; destination in-memory mapping before on-disk append) are exactly what the code does — I traced each one against `_load_index_locked`/`_atomic_create`/`_find_existing_locked` and the docstring claims hold up.
- Fixes a second, subtler latent bug as a side effect: the pre-PR code never dropped the origin directory's in-memory `_dir_indexes` entry after a tomb, leaving a stale id→filename mapping cached for the process's lifetime. The new helper's `.pop(model_id, None)` closes that, and `test_tomb_drops_the_origin_index_entry` pins it.
- Deliberately preserves `existing.stem` rather than recomputing from `_compute_base_name(fragment)` on restore, so a title edited between tomb and restore can't silently rename the file out from under `[[wikilink]]`s — verified by `test_round_trip_without_collision_keeps_the_original_filename`, which is an intentional PASS-at-HEAD lock, not a regression test.
- Both deliberately-deferred follow-ups (RuntimeError not caught by `tomb_missing_units`'s `(OSError, KeyError)` handler; an unindexed leak if `unlink()` fails after a successful create) are disclosed in the docstrings and tracked in a real, open issue (#1325) — confirmed via `gh issue view`. I checked `creek/ingest/pipeline.py`'s `tomb_missing_units`/`restore_tombed` callers and this matches: `tomb_missing_units` really does only catch `(OSError, KeyError)`, so an allocator-exhaustion `RuntimeError` would abort the whole full-source ingest loop, exactly as documented.
- `update_fragment` correctly keeps `_atomic_write_text` for its in-place rewrite, where unconditional overwrite of the file the id already owns is the right semantics — the PR doesn't over-apply the fix.
- Single commit in proper Conventional Commits form (`fix(vault): ...`).

## Security Concerns
🟢 None. No new I/O surface, path construction stays within existing sanitized routing directories, no injection vectors introduced.

## Problems
None found. I did not find a scenario where the new code path loses data, double-writes, or mis-indexes a fragment.

## Code Quality
- Non-blocking: the docstrings are unusually long (the helper's docstring alone runs ~70 lines). Given the subtlety of the ordering invariants being protected and that this is a data-loss-class bug, the verbosity is arguably justified here, but a shorter form referencing an ADR could be considered for future similar fixes.
- Non-blocking: `_relocate_fragment_locked` takes 5 positional-ish params (`existing, dest_dir, post, model_id, model_type`); fine at 2 call sites, but if a third relocation call site appears, worth reconsidering as a small dataclass.

## CLAUDE.md Adherence
✅ TDD workflow followed (PR description documents mutation-tested tests, each guard proven load-bearing by targeted reversion).
✅ No `# noqa`/`# type: ignore`/skip patterns introduced.
✅ Docstrings are thorough Google-style with Args/Returns/Raises throughout.
✅ Conventional commit message.
⚠️ I could not independently execute `./scripts/check-all.sh` in this review sandbox (command execution required interactive approval unavailable here), so the PR's claimed "12/12 checks, exit 0, 94.39% coverage" is not independently re-verified — it's a self-report, though a highly detailed and specific one (mutation table, exact coverage numbers) that reads as genuine rather than templated.

## Testing
Eight new tests in `TestTombRestoreFilenameCollisions` cover: bystander preservation on both tomb and restore directions, raw index/provenance correctness (deliberately bypassing the self-healing `_find_existing` reader so a mis-record can't hide behind auto-repair), origin-index cleanup, failure-path source-preservation for both directions (via a monkeypatched `_atomic_create` that raises), and the name-stability round-trip lock. This is well above the bar for a data-loss fix — the tests target the actual failure mode (raw on-disk artifacts, not just the healing-tolerant public API) and would have caught the original bug.

## Documentation
Extensive and accurate. The PR description's reproduction transcript, the mutation-survival table, and the per-ordering rationale are all consistent with what the diff actually does — I checked several of the strongest claims (the origin-index pop, the create-before-unlink ordering, the `(OSError, KeyError)` handler gap in `pipeline.py`) against the code directly rather than taking them at face value, and they held.



## Verdict: LGTM

This is a well-scoped, correctly-implemented data-loss fix: the new `_relocate_fragment_locked` helper routes both `tomb_fragment` and `restore_fragment` through the existing `O_CREAT|O_EXCL` allocator instead of an unconditional overwrite, the create-before-unlink / pop-before-set / index-before-append orderings in the code match what the docstrings claim, and eight new tests exercise the collision, the failure-path safety, the index/provenance recording, and the name-stability contract. The two known-incomplete edge cases (RuntimeError propagation, unindexed leak on failed unlink) are honestly disclosed and tracked in a real, open issue (#1325) rather than swept under the rug. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5247441155 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1130, opened 2026-08-01
test(api): make test_the_semaphore_is_released_after_a_timeout able to fail


Geoffe-Ga:
`tests/test_v1_api_hardening.py:407` asserts both concurrent requests return `503` — but a
*leaked* semaphore slot and a *correctly released* one both produce `503` here, so the assertion
cannot distinguish the bug from correct behaviour. A test that cannot detect its own subject is a
defect.

The release guarantee is independently proven by the adjacent fault-path test at `:386`, so this
is not a live hole — but the test should assert `200` on the second call once a fast handler is
restored.

---
Follow-up from #1117 (review of issue #1074) — https://github.com/Geoffe-Ga/Creek-Vault/pull/1117

Geoffe-Ga:
## Closing: already fixed before this issue was ever picked up — verified by execution

A premise audit against HEAD `fdfc39f` refutes this issue's premise. The defect
was real **on the PR #1117 branch**, and it was repaired on that same branch
**1h52m after this issue was filed**. The repaired version is what merged.

### Timeline

- Issue filed `2026-08-01T02:55:36Z`, citing `tests/test_v1_api_hardening.py:407`.
- Commit `37911450` — *"test(api): delete a test that could not fail, and stop
  re-reading the audit log"* — authored `2026-08-01T04:47:47Z` **is** this
  issue's requested fix. Its diff replaces `_slow_health` with a `_SlowOnce`
  stand-in, swaps the `503` assertion for `200`, and adds the non-vacuity guard
  `assert handler.calls == 2`.
- PR #1117 merged `2026-08-07T19:02:35Z` as `f62e22c`.
  `git log -S "_SlowOnce" -- tests/test_v1_api_hardening.py` returns exactly one
  commit: `f62e22c`. The file has been touched 4 times since; none regressed it.
- The scan that filed this issue read the **pre-`37911450`** tree. Its line
  numbers are ~190 lines stale: `:407` is now `:594`, `:386` is now `:542`.

The body's own remedy — "the test should assert `200` on the second call once a
fast handler is restored" — is verbatim what the merged code already does.

### Proved able to fail, not taken on faith

At HEAD, all 61 tests in the file pass. A targeted mutation applied via a pytest
`-p` plugin (no tree edits) removed `finally: self._slots.release()` from
`ConcurrencyLimitMiddleware.__call__` (`limits.py:180-181`):

```
[...]

Issue #1129, opened 2026-08-01
fix(api): send Cache-Control: no-store on authenticated /v1 responses


Geoffe-Ga:
`/v1` responses carry vault-derived, tier-ceilinged content but no `Cache-Control: no-store`, and
`Vary` omits `Authorization`. Behind a misconfigured shared cache, one consumer's response could
be served to another.

P2 rather than P1 because the default bind is loopback and a non-loopback bind already requires
TLS — so there is no supported deployment today that puts a shared cache in the path.

Test idea: every authenticated 200 carries `Cache-Control: no-store`; `Vary` names
`Authorization`.

---
Follow-up from #1117 (review of issue #1074) — https://github.com/Geoffe-Ga/Creek-Vault/pull/1117

Issue #1128, opened 2026-08-01
fix(api): build the /v1 error-response header merge so Vary cannot be overridden


Geoffe-Ga:
`creek_mcp/httpapi/errors.py:88` builds `{VARY_HEADER: ..., **(headers or {})}`, so a
caller-supplied header dict silently overrides the standing `Vary` value that the tier-ceiling
semantics depend on.

Fix: merge in the other order so the standing headers win.

Test idea: pass a colliding `Vary` in `headers` and assert the response still carries the
canonical value.

---
Follow-up from #1117 (review of issue #1074) — https://github.com/Geoffe-Ga/Creek-Vault/pull/1117

PR #1351, opened 2026-08-11
fix(api): map only routing HTTPExceptions to 404 (#1127)

### LINKED ISSUE #1127: fix(api): map only routing HTTPExceptions to 404
`creek_mcp/httpapi/app.py:234` `_routing_miss` maps *every* `HTTPException` to `404 not_found`.

A future handler raising `HTTPException(403)` or `HTTPException(400)` would silently render as
`404` — the one code this contract treats as most sensitive, because `not_found` is documented as
a routing code that is never emitted for a vault object.

Fix: branch on `status_code in {404, 405}` and let anything else take its own path.

---

Geoffe-Ga:
## Summary

`creek_mcp/httpapi/app.py` registered `exception_handlers={HTTPException: _routing_miss}`, so
**every** `HTTPException` — whatever its status — rendered as `404 not_found`. The issue cited
`:234`; the line had drifted to **`:245`**.

That did two kinds of harm, and only the first was in the issue.

**It made a published promise false.** `not_found` is documented at `creek_mcp/api/models.py:265-269`
and in `docs/api.md` as a *routing* code, never emitted for a vault object, because a caller who
can tell "no such fragment" from "you may not see this fragment" can enumerate the corpus. A
future handler's `HTTPException(403)` wearing `not_found` breaks that.

**It also swallowed the fault.** Masked at the handler, the exception never reached
`ErrorBoundaryMiddleware`, so it was logged *nowhere* — the un-diagnosable-500 condition #1122 was
filed to end, wearing a `404`. A new test caught this (`assert 0 == 1` on the fault-log record
count); the issue did not mention it.

### The fix

Dispatch is now keyed on the two **statuses** that mean "no such route", with the class key
overridden by a handler whose whole body is `raise exc`:

```python
exception_handlers={
    ERROR_STATUS[ErrorCode.NOT_FOUND]: _routing_miss,
    METHOD_NOT_ALLOWED: _routing_miss,
    HTTPException: _not_a_routing_miss,
}
```

Status-keyed registration adds **no branch at all**, rather than the `if status_code in {404, 405}`
the issue proposed, and it makes `_routing_miss`'s standing docstring claim that its status is
"deliberately unread" *structural* — no other status can now arrive there.

### What renders a non-routing `HTTPException`, and why not Starlette

"Let anything else take its own path" has a tempting literal reading — drop the class key and let
Starlette's default handler have it. **I probed that against the real stack and it is a new
defect, not a fix.** A raised `HTTPException(403, detail="fragment 'abc' is above your ceiling")`
comes back as:

```
status: 403 · content-type: text/plain · vary: None
body: fragment 'abc' is above your ceiling
```

Four contract violations at once: not the error envelope, the standing `Vary` that
`httpapi/errors.py` exists to make unconditional is gone, an author-written string is on the wire
where the contract guarantees only constants from `ERROR_MESSAGES`, and `403`/`400`/`502` are
statuses a conforming client maps to "unreachable".

So the class handler re-raises instead. The fault escapes `ExceptionMiddleware`,
`ErrorBoundaryMiddleware` catches it, logs the traceback on `creek_mcp.httpapi.error`, and renders
`500 internal_error` through the single `error_response` builder — one envelope construction site,
one logging site, `Vary` intact.

The premise that makes `500` the honest answer, verified by grep rather than assumed: **there is
no `raise HTTPException` anywhere in `creek_mcp/`.** Every published refusal is a
`return error_response(...)`. Refusals are returned, never raised — so the only legitimate raiser
is Starlette's own router, at `404` or `405`, and anything else is by construction a bug in this
server. A bug is a `500`.

### Error-disclosure went down, not up

Masking a `403` as `404` *hides* information, so unmasking it had to be checked for the reverse.
It discloses nothing new: the caller now sees `500 internal_error` with the same constant message
and the same `Vary`, and the only fact added is "this server has a bug" — never "this object
exists". The exception's `detail` is asserted absent from the body and present only in the
operator log, which is the audience split #1122 established.

`405` is untouched and still invisible — verified live: `PUT /v1/wheel` returns `404 not_found`
with **no** `Allow` header.

## Test plan

- New tests in `tests/test_v1_api_hardening.py`'s "Error boundary" section, all driving the real
  app and asserting on the **rendered response**:
  - `test_a_non_routing_http_exception_is_the_published_500` — `[400, 403, 502]`, asserts
    `500 internal_error` **and explicitly `!= not_found`**
  - `test_a_non_routing_http_exception_echoes_no_detail` — sentinel absent, `content-type:
    application/json` (the assertion that fails loudly if anyone reinstates Starlette's
    `PlainTextResponse` default), `Vary` present
  - `test_a_non_routing_http_exception_is_logged_exactly_once` — record count `== 1`, `exc_info[0]
    is HTTPException`, sentinel in the log and absent from the body
  - `test_a_handler_raised_routing_status_is_still_the_routing_miss` — `[404, 405]` still render
    `404 not_found` with no `Allow`
  - `test_a_non_routing_http_exception_stays_in_the_published_set`
- **RED proved before the fix:** `assert 404 == 500` ×3 and `assert 0 == 1`.
- **Load-bearing proved:** reverting *only* the `exception_handlers` dict turns exactly those four
  red and nothing else — `4 failed, 388 passed`.
- `tests/test_v1_api_not_implemented.py` is deliberately **untouched**: its 54-request router
  sweeps at `:406-492` are the regression tripwire for the `404`/`405` half and had to keep
  passing unmodified. They do.
- Gate 2 `./scripts/check-all.sh` → **literal exit 0**; 12 passed / 0 failed, 8826 tests,
  coverage 94.39%.
- No `crawdad/` files touched. No `noqa`, `type: ignore`, skips, or threshold changes.

## Notes

- `METHOD_NOT_ALLOWED` is a literal because the contract deliberately has **no** `ErrorCode` for
  `405` — it is outside the published status set on purpose, so there is no table to derive it
  from. Its sibling key is `ERROR_STATUS[ErrorCode.NOT_FOUND]` rather than a bare `404` for the
  mirror-image reason: that status *does* have a table, and the dispatch key must not drift from
  the code `_routing_miss` renders.
- `_not_a_routing_miss` is entered **twice** per fault, because Starlette wraps exception handling
  both per-route (`routing.request_response`) and at `ExceptionMiddleware`. A pure re-raise makes
  that observationally invisible; the `== 1` log-record assertion is what keeps it pure.
- No collision with the sibling follow-ups from the same review: **#1128** and **#1129** touch
  `httpapi/errors.py`, **#1130** touches `test_v1_api_hardening.py:407` (~1000 lines from this
  diff's additions). This change routes its new `500` through `error_response`, so it inherits
  #1128's `Vary` merge fix and #1129's `Cache-Control` work automatically.

Closes #1127

### DIFF
--- a/creek-tools/creek_mcp/httpapi/app.py
+++ b/creek-tools/creek_mcp/httpapi/app.py
@@ -20,9 +20,25 @@ accepts both spellings is a parser two implementations will disagree about.
 published status set ``{200, 401, 403, 404, 409, 422, 500, 501, 503}``, and a
 conforming client maps an out-of-set status to "unreachable" — losing the vault
 entirely over a wrong verb. It would also publish which verbs a path serves,
-which is a small enumeration primitive. So Starlette's ``HTTPException`` is
-handled here and both the unmatched path and the method mismatch become ``404
-not_found``.
+which is a small enumeration primitive. So the dispatch to :func:`_routing_miss`
+is keyed on the two *statuses* that mean "no such route", ``404`` and ``405``,
+and both the unmatched path and the method mismatch become ``404 not_found``.
+
+**Keyed on the status, not on the exception class (#1127).** Registered against
+:class:`starlette.exceptions.HTTPException` itself, *every* one of them became
+``404 not_found``, which did two kinds of harm. It made a published promise
+false — ``not_found`` is documented as a *routing* code, never emitted for a
+vault object — and, less visibly, it **swallowed the fault**: masked at the
+handler, the exception never reached
+:class:`~creek_mcp.httpapi.middleware.boundary.ErrorBoundaryMiddleware`, so it
+was logged nowhere. That is the un-diagnosable ``500`` #1122 was filed to end,
+wearing a ``404``. The corollary is the whole design: **in this contract
+refusals are returned, never raised** — every published refusal is a ``return
+error_response(...)`` — so the only legitimate raiser of ``HTTPException`` is
+Starlette's own router, at ``404`` or ``405``. Anything else is by construction
+a bug in this server, and a bug is a ``500``. :func:`_not_a_routing_miss`
+therefore holds the class key and does nothing but re-raise, sending the fault
+up to the error boundary to be logged and enveloped there.
 
 This module is therefore the **only** one in :mod:`creek_mcp.httpapi` that names
 :attr:`~creek_mcp.api.models.ErrorCode.NOT_FOUND`, and an AST guard pins that.
@@ -37,14 +53,14 @@ version of that guard — an allowlist over every construction site in
 from __future__ import annotations
 
 from functools import update_wrapper
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Final, NoReturn
 
 from starlette.applications import Starlette
 from starlette.exceptions import HTTPException
 from starlette.middleware import Middleware
 from starlette.routing import Route
 
-from creek_mcp.api.models import SUPPORTED_CONTRACT_MINORS, ErrorCode
+from creek_mcp.api.models import ERROR_STATUS, SUPPORTED_CONTRACT_MINORS, ErrorCode
 from creek_mcp.api.routes import CONTRACT_VERSION_HEADER, ROUTES
 from creek_mcp.httpapi import handlers
 from creek_mcp.httpapi.auth import BearerAuthMiddleware, build_verifier
@@ -73,6 +89,22 @@ if TYPE_CHECKING:
     from creek_mcp.httpapi.handlers import Handler
     from creek_mcp.remote_auth import ConsumerTokenVerifier
     from creek_mcp.tools.reflect import _LLMFactory
+
+METHOD_NOT_ALLOWED: Final[int] = 405
+"""The status Starlette's router raises on a verb mismatch.
+
+Spelled as a literal precisely because the contract has **no**
+:class:`~creek_mcp.api.models.ErrorCode` for it. ``405`` is outside the
+published status set ``{200, 401, 403, 404, 409, 422, 500, 501, 503}`` *on
+purpose* — see the module docstring for why a method miss must be
+indistinguishable from a path miss — so there is no table to derive it from,
+and inventing a code to derive it from would put it back in the set.
+
+Its sibling key is written ``ERROR_STATUS[ErrorCode.NOT_FOUND]`` rather than a
+bare ``404`` for the mirror-image reason: that status *does* have a table, and
+the key dispatch matches on must not be able to drift from the code
+:func:`_routing_miss` renders.
+"""
 
 
 def _speaks_a_served_minor(request: Request) -> bool:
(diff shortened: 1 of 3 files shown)


Geoffe-Ga:
## Summary
Fixes #1127: `creek_mcp/httpapi/app.py` previously registered `exception_handlers={HTTPException: _routing_miss}`, so every `HTTPException` — regardless of status — rendered as `404 not_found`, both breaking the published "not_found is routing-only" contract and silently swallowing non-routing faults before they reached `ErrorBoundaryMiddleware`'s logging. The fix keys dispatch on the two routing *statuses* (`404`, `405`) for `_routing_miss`, and overrides the `HTTPException` class key with a new `_not_a_routing_miss` handler whose entire body is `raise exc`, letting any other `HTTPException` escape to the error boundary and render as a logged `500 internal_error`.

## Strengths
- Correctly relies on Starlette's `ExceptionMiddleware` dispatch order (status-code-keyed handlers are looked up before class-keyed ones for `HTTPException`), which is the right mechanism for this problem and adds no new branching logic.
- The PR explicitly probed and rejected the naive alternative (dropping the class key entirely and letting Starlette's default handler render it), and documented in both the module docstring and `docs/api.md` exactly why that would violate the envelope/`Vary`/no-detail-echo contract.
- Test suite (`tests/test_v1_api_hardening.py`) is comprehensive: covers non-routing statuses becoming `500` with no leaked detail, correct `content-type`/`Vary`, exactly-once fault logging (guarding against Starlette's known double-invocation of exception handlers across per-route and `ExceptionMiddleware` layers), and confirms `404`/`405` still route correctly with no `Allow` header leak.
- Verified (per PR description) that reverting only the `exception_handlers` dict turns exactly the four new tests red and nothing else — good evidence the tests are load-bearing rather than incidental.
- `tests/test_v1_api_not_implemented.py`'s 54-request routing sweep was left untouched and still passes, confirming no regression to the existing `404`/`405` behavior.
- Traced the ASGI middleware ordering (`ErrorBoundaryMiddleware` sits above `ExceptionMiddleware` in the stack) to confirm the re-raised exception is actually caught and logged rather than propagating to `ServerErrorMiddleware` unlogged — this matches the codebase's own architecture as read in `middleware/boundary.py`.

## Security Concerns
🟢 LOW / net improvement. This is itself a security fix: it stops a `403`/other non-routing refusal from being masked as `404` (which could otherwise be exploited as noted for the "distinguish no-such-fragment from you-may-not-see-this-fragment" existence oracle concern referenced in #846/#970/#972/#1090), and confirms the reverse direction was checked — the caller only learns "this server has a bug" (`500`), never anything more specific than before. No `detail` or traceback is echoed to the client (asserted by test).

## Problems
None found blocking merge.

## Code Quality
- Non-blocking style note: the docstrings in `app.py` (module-level and per-function) are extremely dense/essay-like, far beyond typical Google-style docstring length. This matches the established convention already present in this file and package before the PR (e.g., `errors.py`, `boundary.py`), so it's consistent with house style rather than a regression — just flagging for awareness, not requesting a change.
- `ALLOWED_HTTP_STATUSES` in the new test file intentionally duplicates the same constant already defined in `tests/test_v1_api_not_implemented.py`, with a docstring explaining why (no cross-suite imports, so two suites can't silently start measuring different things). Deliberate and justified, not an oversight.

## CLAUDE.md Adherence
✅ TDD followed per PR description (RED tests proven before the fix).
✅ No `# noqa`, `# type: ignore`, skipped tests, or threshold changes introduced.
✅ No direct commit to `main`; PR targets a feature branch (`issue/1127-issue`).
✅ Docstrings are exhaustive (Google-style Args/Returns/Raises), consistent with ≥95% interrogate requirement.
✅ `./scripts/check-all.sh` reported exit 0 with 94.39% coverage per PR description (not independently re-run in this review due to sandbox execution restrictions, but the diff is internally consistent with that claim — no obvious lint/type issues spotted on read-through).
✅ No `crawdad/` files touched, scoped exactly to the stated issue.

## Testing
Strong. New tests cover: non-routing `HTTPException` → `500` with correct code/message/`Vary`/content-type and no detail leak; exactly-once fault logging (guards against the double-entry hazard from Starlette's per-route + `ExceptionMiddleware` handler wrapping); `404`/`405` routing behavior preserved for handler-raised as well as router-raised instances; and that all observed statuses stay within the published closed set. The existing 54-request routing regression sweep in `test_v1_api_not_implemented.py` was left untouched and still passes per the PR description, confirming no behavior change to the two statuses that were already correct.

## Documentation
`docs/api.md` gained a clear paragraph explaining that only the two routing statuses become `not_found`, and why any other internal `HTTPException` is treated as a server bug rendered as `500`. The module docstring in `app.py` was updated to describe the new keyed-dispatch design and its rationale (with the #1127 issue reference), and the new `_not_a_routing_miss` function has a thorough docstring explaining the double-entry hazard and why a pure re-raise is required. Documentation is accurate and matches the implementation.


## Verdict: LGTM

The fix correctly exploits Starlette's exception-dispatch priority (status-code keys are checked before class-based keys for `HTTPException`), is exercised by a thorough, well-isolated test suite that proves both RED-before-fix and that reverting only the dict change breaks exactly the new tests, and the docs/docstrings accurately describe the new behavior with no CLAUDE.md violations (no noqa/type:ignore, TDD followed, docstrings exhaustive). Nothing here blocks merge and I found no substantive suggestion strong enough to withhold LGTM.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5247540205 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

PR #1355, opened 2026-08-11
fix(skills): gate thread and eddy skills on a member-derived tier (#1284)

### LINKED ISSUE #1284: Voice-skill tree: thread and eddy skills are ungated, and their member-derived titles reach skill_paths at ceiling=open

Geoffe-Ga:
## Summary

`SkillTreeGenerator` gated only its **fragment** walk against the caller's
privacy tier ceiling. PR #1286 (issue #971) threaded `override` into
`_collect_fragments` and stopped there, on purpose: `Thread` and `Eddy` carry
no `privacy_tier` field, so gating `_collect_typed` beside it needed a product
decision about what tier a *derived* title inherits. #1286 documented that gap
in four places and named this issue as the uncovered surface.

It was a real leak on both sides. A thread's or eddy's title is computed from
its member fragments (`creek/link/naming.py::cluster_title`), and the generator
slugifies that title into the SKILL **filename** — which
`creek.skills.refresh` hands straight back to its caller in `skill_paths`. So
at `ceiling=open` an MCP caller received, in the response body, slugs derived
from above-ceiling content, and `skill_count` moved with how many above-ceiling
threads and eddies the vault held.

### The decision

**A thread or eddy is admitted only when every fragment naming it clears the
caller's ceiling. One that no fragment names is admitted only at
`ceiling=intimate`/`all`.**

Mechanically that is the maximum tier over every fragment whose
`threads`/`eddies` wikilinks name the title, with the empty set reducing to
`INTIMATE`; because the reduction and the cutoff both rank by
`tier_sensitivity`, "the max clears the ceiling" and "every member clears the
ceiling" are the same statement.

This is not a new answer — it is the one #969 already reached for the identical
question in `creek state`. The reduction is now **shared** rather than
re-derived: `derived_link_tiers`, `wikilink_targets` and
`admit_by_derived_tier` live in `creek/generate/state_tiers.py` and both
generators call them. An eddy the state report withholds at `ceiling=open` and
the skill tree emits is not a difference of opinion between two modules; it is
a leak.

Two implementation details are load-bearing:

- **The reduction runs over the unfiltered corpus.** `_read_all_fragments` was
  split out of `_collect_fragments` so evidence is gathered before admission.
  Narrowing first would resolve an eddy with one `open` and one `personal`
  member to `open` — leak (3) of the three #969 reproduced.
- **`_render_eddy_body` filters its "Member threads" line.** An eddy admitted
  on its own members can still name a thread withheld on *its* members. Gating
  the typed walk alone would have closed the front door and left this one open.
  Withheld entries are dropped rather than counted; a `"+2 withheld"` marker
  would restore the cardinality oracle the gate exists to remove.

### Rejected, and why

Failing closed on the note itself (#968's shape, since neither model has the
key) and excluding the categories wholesale above some ceiling both empty two
whole categories at every ceiling below `intimate` — on a fully-classified
all-`open` vault where nothing is actually sensitive. That is an outage, not a
gate. Deriving costs one extra pass over a corpus the generator already walks.

### What it costs

This only ever emits **less**; nothing previously withheld is now emitted. At
`ceiling=open` a thread with one `personal` member disappears; on a vault that
has never been through `creek classify` every thread and eddy disappears
(untiered ranks with `personal`, #876); an **orphaned** thread left behind by a
partial `creek link` is withheld below `ceiling=intimate`. All are recoverable
with `--include-tier`, and `_log_ceiling_withheld` already names the remedy.

Recorded as
`docs/architecture/ADR/0010-derived-tier-for-threads-and-eddies.md` with the
rejected options and a revisit predicate (revisit when `Thread`/`Eddy` gain a
real `privacy_tier` stamped at link time). `creek/generate/skills.py::_TYPED_TIER_POLICY`
is the in-code pointer.

### One deliberate divergence

For *member evidence* `creek/generate/state.py` reads `raw_privacy_tier`
(missing key fails closed to `INTIMATE`) while `creek/generate/skills.py` reads
`tier_of` (missing key is the model's `UNCLASSIFIED`, ranking with `PERSONAL`).
Each generator ranks a thread's members with the same reader it already uses
for the thread's own fragments — the alternative is a generator incoherent with
itself, withholding a thread every one of whose members that same call admits,
which would make `--include-tier personal` useless on a freshly ingested vault.
`_collect_fragments` argues the same choice for its own gate. Pinned by
`test_a_thread_whose_members_are_keyless_is_admitted_at_the_personal_ceiling`.

The four claim sites #971 narrowed are widened back: `creek_mcp/read_gate.py`'s
`creek.skills.refresh` rationale, `creek_mcp/tools/skills.py`'s docstring,
`docs/mcp.md`'s table row, and `tests/test_skills_tier_ceiling.py`'s module
docstring.

## Test plan

Gate 1 RED first. Seven of the eight new tests in section T10 of
`tests/test_skills_tier_ceiling.py` failed against the unfixed tree, every one
of them on rendered file content under `<vault>/creek-skills/` (and, where the
filename is the disclosure, on the returned `skill_paths`) rather than on a
return value:

```
FAILED test_a_thread_derived_from_a_personal_member_is_withheld_at_the_open_ceiling
FAILED test_an_eddy_with_one_above_ceiling_member_is_withheld_at_the_open_ceiling
FAILED test_an_admitted_eddy_does_not_render_a_withheld_member_thread
FAILED test_a_thread_no_fragment_names_is_withheld_below_the_intimate_ceiling
FAILED test_a_thread_whose_members_are_keyless_is_admitted_at_the_personal_ceiling
FAILED test_skill_count_does_not_move_with_above_ceiling_thread_cardinality
FAILED test_cli_withholds_a_personal_derived_thread_at_the_default_ceiling
7 failed, 3 passed
```

The eighth, `test_thread_and_eddy_skills_return_in_full_at_the_all_ceiling`,
passes both before and after by design — it is the recoverability control.

Coverage of the rule: the mixed-member eddy (the assertion that catches a fix
reducing over the *admitted* corpus), the no-evidence orphan, the keyless
member, `skill_count` cardinality across two vaults, recoverability at
`ceiling=all`, and the CLI as the second production surface. Every exclusion
test carries a positive control in the same vault, so none can pass on an
implementation that simply emits nothing.

`tests/test_skills.py`'s `_write_thread` / `_write_eddy` now write a companion
`open` member fragment by default, because a thread nothing links to is
invisible at the ceilings those tests use — the shape a vault that has been
through `creek link` actually has. Two assertions were updated deliberately and
explained in place: a snapshot fragment count `1 -> 3`, and a dangling
`"thread-abc"` member link replaced with a real backed thread title (asserting
on the dangling link would have pinned the second-order leak in place). No test
was deleted, skipped, or weakened; no threshold changed; no `# noqa`,
`# type: ignore` or `--no-verify`.

Gate 2: `./scripts/check-all.sh` **exit 0** — 8841 passed, 5 skipped (all
pre-existing environment skips: Ollama and the live-API smokes), coverage
94.40% aggregate, per-file gate green on 256 files, mypy strict, xenon, ruff
lint + format with `--no-cache`.

Two properties verified explicitly because they are easy to break silently:
`tests/test_mcp_read_gate.py` reports **151 passed, 0 skipped** (an emptied
parametrize source would show up as a skip), and #1323's `list(Authorship)` x
tier cross-product in `tests/test_voice_exemplars.py` is **80 passed, 0
skipped**.

Disclosure: the `chief-architect` pass was launched but did not return a plan
before a session interruption, so the design above is the conductor's own, not
an independent architect review. Gate 2.5 `code-review-orchestrator` did run.

Closes #1284
Refs #971, #969, #931, #1286


Geoffe-Ga:
## Summary
This PR closes a privacy leak in `SkillTreeGenerator`: thread and eddy titles (computed from member fragments) were reaching `<vault>/creek-skills` and the `skill_paths` MCP response uncontrolled by the caller's privacy ceiling, even though fragment bodies were already gated (#971). It adopts the derived-tier rule #969 already established for `creek state` — a thread/eddy is admitted only when every fragment naming it clears the ceiling — and shares the reduction logic (`derived_link_tiers`, `wikilink_targets`, `admit_by_derived_tier`) between the two generators via a new `creek/generate/state_tiers.py` module, plus records the decision as ADR-0010.

## Strengths
- **Correct core algorithm**: the reduction (`max_source_tier`) and the cutoff (`tier_within_override`) both rank by `tier_sensitivity`, so "max clears ceiling" ⇔ "every member clears ceiling" — verified this equivalence holds by reading `privacy_filter.py` directly.
- **Ordering trap avoided deliberately**: `_read_all_fragments` is split out so the derived-tier reduction runs over the *unfiltered* corpus — reducing over the already-admitted corpus would resolve a mixed open/personal eddy to `open`, reproducing "leak (3)" from #969. This is tested explicitly (`test_an_eddy_with_one_above_ceiling_member_is_withheld_at_the_open_ceiling`).
- **Second-order leak also closed**: `_render_eddy_body` now intersects `eddy.threads` with the admitted-thread set, so an admitted eddy can't print a withheld thread's title in its "Member threads" line — a leak that gating only `_collect_typed` would have missed. Covered by `test_an_admitted_eddy_does_not_render_a_withheld_member_thread`.
- **Sharing, not duplicating**: verified both `creek/generate/state.py` and `creek/generate/skills.py` now call the same `admit_by_derived_tier`/`derived_link_tiers` in `state_tiers.py`, and confirmed no dead imports were left behind in `state.py` after the extraction (`max_source_tier`, `tier_within_override`, `Thread`, `Eddy` are all still used elsewhere in that file).
- **Both production surfaces fixed**: confirmed via grep that both the CLI (`creek/cli.py`) and the MCP tool (`creek_mcp/tools/skills.py`) construct `SkillTreeGenerator` and route through the shared `_load_vault_snapshot`, so the fix isn't MCP-only. `test_cli_withholds_a_personal_derived_thread_at_the_default_ceiling` pins this.
- **Anti-vacuity test design**: every exclusion assertion is paired with a positive control (an open thread/eddy that must still appear) in the same vault, so a trivial "stop emitting thread/eddy skills entirely" fix would fail the suite.
- **Recoverability preserved**: `test_thread_and_eddy_skills_return_in_full_at_the_all_ceiling` confirms nothing is permanently lost, only gated, consistent with the project's stated non-goal of refusal (#968).
- **Cardinality oracle closed**: `test_skill_count_does_not_move_with_above_ceiling_thread_cardinality` prevents a subtler leak where `skill_count` alone could reveal how much above-ceiling content exists.
- Extensive, precise docstrings and an ADR that states costs and a revisit predicate up front — unusually good documentation discipline for a security-relevant change.

## Security Concerns
🟢 LOW — none blocking. The deliberate divergence between `state.py` (raw frontmatter reader, fails closed to `INTIMATE` on a missing key) and `skills.py` (model reader `tier_of`, missing key ranks as `UNCLASSIFIED`/`PERSONAL`) is explained and pinned by a dedicated test (`test_a_thread_whose_members_are_keyless_is_admitted_at_the_personal_ceiling`); the rationale (each generator stays internally consistent with the reader it already uses for the fragment's own admission) is sound, but it is worth double-checking in future PRs that this divergence doesn't drift further, since two readers for "the same" tier concept is inherently a place bugs like this one originate from.

## Problems
None found that block merge.

## Code Quality
- Non-blocking: `_read_all_fragments` and `_collect_fragments` both walk `fragments_root.rglob("*.md")`-shaped work: `_read_all_fragments` now owns the walk and `_collect_fragments` filters an in-memory list — this is a net-positive simplification (one disk walk instead of a walk-plus-derive-that-re-walks), so no concern here, just noting it as intentional per the PR description.
- Line 43-44 of the diff to `_read_all_fragments`'s docstring has an odd manual line-break in the Sphinx cross-reference (`:func:`~creek.generate.state.\` followed by a continuation) — looks like an artifact of a line-length wrap; cosmetic only, won't affect rendering meaningfully but slightly unusual style versus the rest of the file's docstrings.

## CLAUDE.md Adherence
✅ TDD followed — PR description documents RED (7/8 new tests failing pre-fix) then GREEN.
✅ `./scripts/check-all.sh` claimed exit 0 with 8841 passed, coverage 94.40%, per-file gate green, mypy strict, ruff clean (could not independently re-run `./scripts/test.sh`/`check-all.sh` in this review environment due to sandbox command-approval restrictions, but manual source inspection found no dead imports, no unused symbols, and no obvious lint/type issues introduced).
✅ ADR added under `docs/architecture/ADR/` per repo convention (5.3).
✅ No forbidden patterns observed (no `# noqa`, `# type: ignore`, `--no-verify`, skipped/weakened tests).
✅ Conventional commit style (`fix(skills): ...`), single feature-branch commit, closes issue #1284.

## Testing
Insufficient testing would be blocking, but here testing is a strength, not a gap. Eight new tests in `tests/test_skills_tier_ceiling.py` section T10 cover: personal-derived thread exclusion, mixed-member eddy (the ordering-trap canary), the second-order eddy-body leak, the orphan case, the keyless-member divergence from `creek state`, full recovery at `ceiling=all`, the `skill_count` cardinality oracle, and the CLI as a second production surface. `tests/test_skills.py` fixtures were updated to give threads/eddies member evidence (matching post-`creek link` vault shape) with the reasoning for the `1 -> 3` fragment-count assertion and the `"thread-abc"` → real-title swap explained inline.

## Documentation
Thorough: module docstrings in `skills.py` and `state_tiers.py` updated, `creek_mcp/read_gate.py`'s tool-posture rationale and `creek_mcp/tools/skills.py`'s docstring both widened to reflect the new gating (previously they explicitly documented threads/eddies as *ungated*, which is exactly the surface this PR fixes), `docs/mcp.md`'s `creek.skills.refresh` table row updated to match, and ADR-0010 records the decision, rejected alternatives, and a concrete revisit predicate.



## Verdict: LGTM

The fix correctly closes a real information-disclosure gap (thread/eddy titles derived from above-ceiling fragments leaking into the skill tree and skill_paths), the derived-tier logic is provably equivalent to the "every member must clear the ceiling" rule and is shared with the already-reviewed `creek state` implementation rather than re-derived, the corpus-before-filter ordering avoids the exact leak class (mixed-tier resolving to the most permissive member) that a naive implementation would introduce, and the test suite includes a strong positive control alongside every exclusion so no test can pass on a trivial "emit nothing" implementation. No blocking issues found.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5247656720 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1359, opened 2026-08-11
fix(redact): `--scan --vault <escaping symlink>` reads and parses an out-of-tree creek_config.yaml


Geoffe-Ga:
## Summary

`creek redact --scan --vault <symlink-whose-target-escapes-its-parent>` opens and
parses `<vault>/00-Creek-Meta/creek_config.yaml` **through the link**, from
outside the tree the operator named. Found while closing #1293, which contained
every other directly-named path on the `creek redact` surface.

This is a **read-escape, not a write**: `--scan` writes nothing, so no file
outside the root is modified. It is filed separately rather than folded into
#1293 because the fix is a genuine judgement call, not an oversight — see
"Why this was not just fixed" below.

## Verified at `issue/1293-issue`

Setup: `<S>/outside/vault/00-Creek-Meta/creek_config.yaml` containing malformed
YAML; `<S>/root/linkvault -> <S>/outside/vault`; an innocent in-tree
`<S>/root/src`.

```
creek redact --scan --source <S>/root/src --vault <S>/root/linkvault
```

surfaces a YAML parse traceback originating from the out-of-tree file. The file
is therefore opened, read, and parsed, and the resulting config steers the run.
With a well-formed out-of-tree config the run completes silently.

I did **not** manage to demonstrate the out-of-tree config *weakening* pattern
coverage (an `allowlist` entry did not change the finding count in a quick
probe). So the confirmed impact is "an out-of-tree file is read and parsed and
configures the safety pass", not "the safety pass can be silently disarmed from
outside the root". Whether the latter is reachable is the first thing to
establish here — if it is, this jumps in severity.

## Why this was not just fixed in #1293

Every other named path on this surface is now refused outright:

- `--apply --source`, `--apply --vault` (that one was a real out-of-root
  **write** — the audit record lands in
  `<vault>/00-Creek-Meta/audit/redact.jsonl`, creating `audit/` if absent)
- `--review --vault`

`--scan` is deliberately different in kind. Its contract is **skip-and-count**,
established in #1087 and pinned by
`tests/test_cli_redact.py:545-553` and `:576-628`: `--scan` writes nothing, so
refusing an entire scan over one bad link is a denial of service on the safety
pass itself — the operator's only means of finding out what is exposed.

Refusing an escaping `--scan --vault` would break a plausible real workflow
(`~/vault -> /Volumes/Backup/vault`, then `creek redact --scan --vault ~/vault`)
[...]

PR #1358, opened 2026-08-11
fix(classify): a failed weighted LLM call stamps rules, not a lying llm (#1330)

Geoffe-Ga:
## Summary

- **A soft LLM failure on the weighted classification path stamped `classification_method: llm` and buried the fragment forever.** `_classify_one_weighted` had no `succeeded` check, unlike its single-pick sibling eight lines above it (`classify_engine.py:1519-1529`, added by #744 for this exact failure mode). `classify_weighted` fails soft to a bare `WeightedFragmentClassification()` on provider-unavailable, transport error, or malformed YAML — a value indistinguishable from a genuine all-unclassified verdict — and the engine assigned it wholesale with `was_skipped=False`.
- **`classify_weighted` now returns a `WeightedClassificationResult(classification, succeeded)`**, mirroring `LLMClassificationResult` (`llm/orchestrator.py:48-70`), the wrapper the sibling path already uses for this. Its fail-soft behaviour is unchanged; the caller is simply told which of "no signal" and "no answer" it received. The engine returns `(fragment, True, "")` on failure — the input fragment untouched — so the write path stamps the honest `rules`.
- **Forward-only.** Fragments already mis-stamped by this bug are not healed; see the follow-up below.

## The bug, reproduced end-to-end before the fix

Run over a tmp vault with `weighted_classification=True` and a fragment the rule classifier scores `F3`, for each of the three soft-failure modes:

```
classification_method: llm                                   <- the lie
classification_provider: ollama                              <- for a call that produced nothing
frequency: {'primary': 'unclassified', 'secondary': []}      <- the F3 rules verdict destroyed
weighted:  {all six dimension lists empty, overall_confidence: 0.0, reasoning: ''}
                                                             <- a fabricated profile persisted
progress-log: {"id": "frag-0000weighted"}                    <- OPS-001 resume ledger poisoned
```

A second, non-`--force` run then reported `preserved_llm=1, classified=0`: `_record_if_preserved` (`classify_engine.py:689-695`) treats `llm` as "already paid for", so the fragment was permanently `unclassified` **and** permanently skipped. The `weighted:` block and the ledger entry were not in the original issue report; both were found while verifying it.

The whitespace-only-body short-circuit (`weighted.py:825-826`) told the same lie for a call where no provider was ever contacted, and is fixed with the same flag. Cost of the extra eligibility: none — that path returns before a classifier is even constructed.

## What changed

| File | Change |
|---|---|
[...]

PR #1360, opened 2026-08-11
fix(redact): contain the paths the operator names, not just the walked tree

### LINKED ISSUE #1293: fix(redact): SEC-003 guard is dir-only — 'creek redact --apply <symlinked-file>' writes through the link

Geoffe-Ga:
## Summary

SEC-003's symlink guard was **directory-only**. It walked a tree looking for
escaping links *inside* it, and never examined the path the operator **named**
on the command line. Naming a symlink walked straight past it.

**The issue body's mechanism is wrong, and I did not implement it.** It claims
`creek redact --apply <symlinked-file>` "writes through the link" and asks for a
red-first test asserting the victim's bytes are unchanged. `_atomic_write` ends
in `os.replace` (`cli_commands.py`), and **`os.replace` onto a symlink path
replaces the link — it does not follow it**. I ran the real CLI at the base
commit: victim md5 `5859146984a98cc174eaa129a0fa1745` before *and* after. That
acceptance criterion is already green at HEAD, so writing it to spec would have
shipped a fake red. A premise correction is recorded on the issue.

### What was actually broken (all verified by running the CLI, exit 0, no refusal)

1. **Named symlinked FILE** — out-of-root **read** + **exfiltration into the
   tree** + **link destruction**. The target's secrets were enumerated and
   printed (`api_key`, `env_secret`, `high_entropy_string`, `ssn`), then
   `os.replace` destroyed the link, leaving a regular in-tree file holding a
   redacted copy of out-of-tree content.
2. **Named symlinked DIRECTORY** — a genuine in-place **out-of-root write**, and
   the case the issue body omits entirely. `is_dir()` follows the link, so the
   walk guard ran with `resolved_root = root.resolve()` = the *target*,
   laundering every child as "inside". Verified: out-of-tree `a.md` md5
   `06832860b4ed54631d892598f5743380` → `2c8ab87896ad93d35109d353dccc493a`.
3. **`run_scan`** — a third `_scan_source` caller the body never names, with **no
   symlink guard of any kind**.
4. **`--apply --vault <escaping link>`** — found at self-review, not in the issue
   or the plan. `run_apply` takes a `--vault` of its own and **writes** to it:
   the audit record lands in `<vault>/00-Creek-Meta/audit/redact.jsonl`,
   *creating* `audit/` if absent. So a second out-of-root write, reachable with
   an entirely innocent `--source`, putting the record of what was touched
   wherever a link pointed. Verified live, then fixed.

## The fix

- `scanner._resolves_within` → public **`resolves_within`**, so the walked-child
  and named-path surfaces share one definition of "inside" instead of drifting
  into two.
- **`_named_path_escapes(path)`** — `path.is_symlink() and not
  resolves_within(path, path.parent.resolve(strict=False))`.
- **`_assert_named_path_contained(...)`** — exit 1, the exact existing SEC-003
  message shape, naming **only the as-supplied path**. `logger.warning`, not
  `logger.exception` (no exception is being handled there; `.exception` would
  append a bogus `NoneType: None`).
- **`SymlinkPolicy`** enum on `_scan_source`, as a **required, non-defaulted**
  kwarg. That absence is the point: mypy strict now makes it a build error to
  add a fourth redaction mode without stating a symlink policy — the exact
  omission that left `run_scan` unguarded and produced this issue.
- Handler guards in `run_apply`/`run_review` placed after `_require_existing`
  but **before** `is_dir()` (which follows the link) and **before**
  `load_config` (which would otherwise read `creek_config.yaml` *through* it).
  The shipped `_assert_no_escaping_symlinks` walk guard was left exactly where
  it is.

### The two contracts stay distinct

`--apply`/`--review` **refuse** (exit 1). `--scan` **skips and counts** and still
exits 0. `--scan` writes nothing, so refusing a whole scan over one bad link
would be a denial of service on the safety pass itself — a regression wearing a
fix's clothes. This is why the policy is a parameter rather than an
unconditional guard at the shared chokepoint. `tests/test_cli_redact.py:545-553`
and `:576-628` pass unmodified.

### Decision: refused outright, no waiver

No `--force`, no environment variable, no config key, per SEC-003's no-waiver
precedent. The security direction here is one-way: this guard may only ever
cause the tool to read and write *less*. Recorded in the guard's docstring.

### Named symlinked directory — stated accurately

The predicate is **leaf-only**, so the honest statement is: a named symlinked
directory is **refused when its target escapes the link's own parent**, and
**admitted — with the resolved target as the root thereafter — when it does
not**. `~/vault -> ~/Dropbox/vault` is still admitted and still redacts under
`~/Dropbox/vault`. It would have been easy to write "refused uniformly" over
code that does no such thing; that would have put a false claim in the threat
model.

## User-visible behaviour change

Aliasing an external export tree into the vault and naming the alias is now
refused:

```
ln -s /Volumes/Export ~/vault/inbox
creek redact --apply --source ~/vault/inbox     # now exits 1
```

Workaround: name the resolved path. And `--scan` on such an alias now reports 0
files scanned with a skip row instead of scanning the target. This is the
correct ratchet direction, but it is a change and is recorded in the threat
model rather than left to be discovered.

## Residuals — deliberately open, not hidden

1. **Leaf-only predicate.** `<root>/linkdir/a.md` (real file leaf, escaping
   *ancestor* component) is still admitted and rewritten in place. Same
   resolve-the-root / lstat-the-leaf asymmetry the scanner already relies on to
   keep a `/tmp` → `/private/tmp` root scannable.
2. **`Pipeline._run_redaction`** (`pipeline.py:478`/`:484`) still traverses a
   directly-named symlinked directory — read-only, escalating to a refusal via
   the skip counter.
3. **`--scan --vault <escaping link>`** still reads and parses an out-of-tree
   `creek_config.yaml`. Read-escape, not a write. Confirmed by pointing it at
   malformed YAML and watching the parse error surface from outside the tree.
   Deliberately not refused here, because a refusing `--scan` is the very DoS
   the skip-and-count contract exists to avoid — that trade deserves its own
   decision. Filed as **#1359 (P2)**.

**#1294 is NOT touched by this PR** and must not be treated as closed or
half-closed. Its sentence in the threat model is preserved verbatim.

## Test plan

Gate 1 RED proven **before** any source change, on behavioural assertions:

```
FAILED test_redact_apply_refuses_a_named_symlink_whose_target_escapes_its_parent
  AssertionError: --apply followed a symlink named directly on the command line;
  the directory-only SEC-003 guard never looked at the named path itself.
  exit_code=0
  │ api_key             │ critical │ 1 │
  │ env_secret          │ high     │ 1 │
  │ high_entropy_string │ medium   │ 1 │
  │ ssn                 │ critical │ 1 │
  Applied redactions to 1 file(s).
  assert 0 == 1
FAILED test_redact_apply_refuses_a_named_symlinked_directory_whose_target_escapes
FAILED test_redact_review_refuses_a_named_symlink_whose_target_escapes_its_parent
FAILED test_redact_review_refusal_reads_no_config_through_the_link
FAILED test_redact_scan_skips_a_named_symlink_whose_target_escapes_its_parent
FAILED test_scan_source_refuses_an_escaping_named_path_under_the_refuse_policy
  ImportError: cannot import name 'SymlinkPolicy'   (expected until GREEN)
6 failed, 3 passed
```

The 3 that passed at HEAD are labelled **ordering pins**, not reds: dangling and
looping named links exit 2 via `_require_existing`'s `path.exists()` (False for
both — Python swallows ELOOP), and an intra-tree alias is admitted. They exist so
a later guard reshuffle cannot silently reclassify a broken link as a
containment refusal.

Two reds rest on **out-of-tree bytes unchanged**; only one legitimately can. In
the named-FILE test that assertion already passes at HEAD, so it is labelled
regression-only in an inline comment. It is load-bearing in the named-DIRECTORY
test, where HEAD really does rewrite in place.

**Disclosure** is asserted via `caplog` records plus the bare filename — never
`str(victim) not in output`, which is vacuous: Rich truncates the Matches table
to the console width (`/var/folders/vg/73pzvjkj3p9dy38k1cf…`). No new test
passes `--verbose`.

### Mutation battery — 6 mutants, 0 survivors

Each guard reverted alone; exactly the intended tests go red.

| Mutant | Result |
|---|---|
| Remove `run_apply` source guard | kills the `--apply` config-ordering pin |
| Remove `run_review` vault guard | kills the `--review` config-ordering pin |
| Disable `_scan_source` chokepoint | kills the scan-skip test + the chokepoint unit test |
| SKIP arm → REFUSE | kills the scan skip-and-count test (DoS regression caught) |
| Predicate → bare `is_symlink()` | kills the intra-tree admit test (over-breadth caught) |
| Remove the new `--vault` guard | kills the escaping-vault write test |

The first mutant **survived the initial battery** — the chokepoint caught it a
few lines later, so removing the handler guard changed nothing observable. A
layer no test can detect the absence of is not a layer, so I added
`test_redact_apply_refusal_reads_no_config_through_the_link`, which pins the
real property: with `--source` and `--vault` both naming the same escaping link,
chokepoint-only ordering parses an out-of-tree config and lets it steer the run
that then refuses. Refusing after being configured by the thing you refused is
not refusing.

### Not weakened

`tests/test_cli_redact.py` lines 1-660 are **byte-identical** to the base commit
(diffed), plus one added `import logging`. Everything else is appended. All six
protected SEC-003/#1087 tests pass unmodified. No test deleted, no assertion
relaxed, no threshold changed, no `# noqa` / `# type: ignore` / skip marker.

### Gates

- `./scripts/check-all.sh` → **exit 0**
- 468 tests pass across `test_cli_redact.py`, `test_redact.py`, `test_cli.py`
- Gate 2.5 `code-review-orchestrator` found the `--vault` write escape above; it
  is fixed here, with its own RED-first test, rather than deferred.
- crawdad's separate gate is not implicated — no caller of these entrypoints
  under `crawdad/crawdad/`. `creek_mcp/tools/redact.py` is out of blast radius
  (already gated by `resolve_within_vault` upstream).

Closes #1293

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -1430,7 +1430,10 @@ def _dispatch_redact(
         apply: ``--apply`` flag.
         review: ``--review`` flag.
         source: ``--source`` path (scan/apply).
-        vault: ``--vault`` path (review).
+        vault: ``--vault`` path. Required by ``--review`` as the tree to
+            re-scan; optional for ``--scan``/``--apply``, where it locates
+            ``creek_config.yaml`` and, for ``--apply``, receives the audit
+            record.
         report: ``--report`` flag (scan).
         dry_run: ``--dry-run`` flag (apply).
         verbose: ``--verbose`` flag.
@@ -1479,7 +1482,11 @@ def redact(
     source: Path | None = typer.Option(
         None, "--source", help="Source path (scan/apply)"
     ),
-    vault: Path | None = typer.Option(None, "--vault", help="Vault path (review)"),
+    vault: Path | None = typer.Option(
+        None,
+        "--vault",
+        help="Vault path (review; also config/audit root for scan/apply)",
+    ),
     report: bool = typer.Option(
         False, "--report", help="Include the detailed markdown report (scan)"
     ),
--- a/creek-tools/creek/redact/cli_commands.py
+++ b/creek-tools/creek/redact/cli_commands.py
@@ -16,6 +16,7 @@ import logging
 import os
 import tempfile
 from collections import Counter
+from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, NoReturn
 
@@ -32,6 +33,7 @@ from creek.redact.scanner import (
     SYMLINK_SKIP_LABEL,
     RedactionScanner,
     ScanSummary,
+    resolves_within,
 )
 
 logger = logging.getLogger(__name__)
@@ -197,21 +199,75 @@ def render_matches(summary: ScanSummary, console: Console) -> None:
 # ---------------------------------------------------------------------------
 
 
+class SymlinkPolicy(Enum):
+    """What a redaction mode does with a directly-named escaping symlink.
+
+    The read and write contracts differ in kind, and the fix for #1293 has to
+    keep them distinct rather than collapsing both into one refusal:
+
+    * ``REFUSE`` — ``--apply`` and ``--review``. Both may write, so following
+      a link named on the command line risks rewriting a file outside the
+      tree the operator named. They abort with exit ``1`` before anything is
+      read through the link.
+    * ``SKIP`` — ``--scan``. It writes nothing, so refusing a whole scan over
+      one bad link would be a denial of service on the safety pass itself —
+      the operator's only way of learning what is exposed. The path is
+      declined, counted under ``files_skipped_symlink``, and the scan still
+      exits ``0``.
+
+    The distinction is pinned by ``tests/test_cli_redact.py:545-553``.
+
+    Attributes:
+        REFUSE: Abort the run — the write path (``--apply``, ``--review``).
+        SKIP: Decline the path and carry on — the read path (``--scan``).
+    """
+
+    REFUSE = "refuse"
+    SKIP = "skip"
+
+
 def _scan_source(
     source: Path,
     config: CreekConfig,
+    *,
+    console: Console,
+    label: str,
+    policy: SymlinkPolicy,
 ) -> tuple[RedactionScanner, ScanSummary]:
     """Scan *source* (file or directory) and return the scanner + summary.
 
+    The single chokepoint through which every redaction mode reaches the
+    filesystem, and therefore the right place to decide what happens when the
+    operator names an escaping symlink outright (#1293). *policy* has no
+    default on purpose: mypy strict then makes it a build error to add a
+    fourth mode without stating a symlink policy, which is precisely the
+    omission that left ``run_scan`` unguarded in the first place.
+
     Args:
-        source: File or directory to scan.
+        source: File or directory to scan, exactly as the operator named it.
         config: Loaded Creek configuration.
+        console: Rich console sink for the refusal banner.
+        label: Human-readable label for the root in the refusal message
+            (e.g. ``"source"`` or ``"vault"``).
+        policy: What to do when *source* itself escapes its own parent —
+            see :class:`SymlinkPolicy`.
 
     Returns:
         Tuple of ``(scanner, summary)``. The scanner is returned so that
-        the caller can reuse its session salt for redaction or review.
+        the caller can reuse its session salt for redaction or review. Under
+        :attr:`SymlinkPolicy.SKIP` an escaping *source* yields an empty
+        summary carrying a single ``files_skipped_symlink``.
+
+    Raises:
+        typer.Exit: With code ``1`` under :attr:`SymlinkPolicy.REFUSE` when
+            *source* is a symlink resolving outside its own parent.
     """
     scanner = RedactionScanner(config=config.redaction)
+    if _named_path_escapes(source):
+        if policy is SymlinkPolicy.REFUSE:
+            _assert_named_path_contained(source, console=console, label=label)
+        logger.warning("Skipping symlink that escapes the scan root: %s", source)
+        return scanner, ScanSummary(files_skipped_symlink=1)
     if source.is_file():
         matches = scanner.scan_file(source)
         summary = ScanSummary(matches=matches, files_scanned=1)
@@ -233,6 +289,96 @@ def _error_exit(console: Console, message: str, *, code: int = 2) -> NoReturn:
     """
     console.print(f"[red]{message}[/red]")
     raise typer.Exit(code=code)
+
+
+def _named_path_escapes(path: Path) -> bool:
+    """Report whether *path* is itself a symlink leaving its own parent.
+
+    The containment question for a path the operator names on the command
+    line, as opposed to one the scanner discovers while walking a tree. It
+    delegates to :func:`creek.redact.scanner.resolves_within` rather than
+    restating the predicate, so the named-leaf surface and the walked-child
+    surface cannot drift into two definitions of "inside".
+
+    Three properties carry the correctness argument:
+
+    * ``is_symlink()`` is an ``lstat`` on the leaf, so a path that is not
+      itself a link is never resolved and never compared. That is what makes
+      this guard behave identically on darwin and on Linux CI: a root reached
+      *through* a symlinked component (``/tmp`` → ``/private/tmp`` on macOS)
+      is not flagged. Same reasoning as the scanner's own walk policy.
+    * When the leaf *is* a link, BOTH sides are resolved — the target inside
+      ``resolves_within``, the parent here — so ``/tmp`` → ``/private/tmp``
+      cannot manufacture a spurious refusal for a link that never left its
+      own directory.
+    * The predicate is LEAF-ONLY, and deliberately so: a named path whose
+      escaping link is an ANCESTOR component (``<root>/linkdir/a.md``, where
+      ``linkdir`` is the link) is admitted. That is a known, accepted
+      residual — not full coverage of path traversal.
+
+    ``strict=False`` matches the shipped guard: a target that does not exist
+    still resolves to a candidate location worth comparing. Dangling and
+    looping links do not reach here in the CLI anyway — ``_require_existing``
+    reports them as "not found" first.
+
+    Args:
+        path: The source or vault path exactly as the operator supplied it.
+
+    Returns:
+        ``True`` when *path* is a symlink whose target is not a descendant of
+        its own resolved parent directory.
+    """
+    return path.is_symlink() and not resolves_within(
+        path,
+        path.parent.resolve(strict=False),
+    )
+
+
+def _assert_named_path_contained(
+    path: Path,
+    *,
+    console: Console,
+    label: str,
+) -> None:
+    """Refuse a directly-named symlink that escapes its own parent (#1293).
+
+    The write-path half of the contract: ``--apply`` and ``--review`` stop
+    here, before the named path is opened, walked, or used to locate
+    ``<vault>/00-Creek-Meta/creek_config.yaml``. The companion read path
+    (``--scan``) skips instead — see :class:`SymlinkPolicy`.
+
+    The refusal is outright. There is no ``--force``, no environment
+    variable, and no config key, per SEC-003's no-waiver precedent: the
+    security direction here is one-way, and this guard may only ever cause
+    the tool to read and write *less*.
+
+    Only the as-supplied path is named — in the banner and in the log. The
+    resolved target is never disclosed: doing so is the very oracle #1087
+    closes. The log is a ``warning`` rather than an ``exception`` because no
+    exception is being handled at this point; ``logger.exception`` here would
+    append a bogus ``NoneType: None`` traceback to the record.
+
+    Args:
+        path: The source or vault path exactly as the operator supplied it.
+        console: Rich console sink for the error banner.
+        label: Human-readable label for the root in the error message
+            (e.g. ``"source"`` or ``"vault"``).
+
+    Raises:
+        typer.Exit: With code ``1`` when *path* is an escaping symlink.
+    """
+    if not _named_path_escapes(path):
+        return
+    logger.warning(
+        "Refusing to follow symlink that escapes the %s root: %s",
+        label,
+        path,
+    )
+    _error_exit(
+        console,
+        f"Refusing to follow symlink that escapes the {label} root: {path}",
+        code=1,
+    )
 
 
 def _assert_no_escaping_symlinks(
@@ -354,6 +500,11 @@ def run_scan(
 ) -> ScanSummary:
     """Scan *source* for sensitive data and render a report.
 
+    The read path never refuses over containment: a *source* that is itself
+    an escaping symlink is declined and counted in the statistics table
+    (:attr:`SymlinkPolicy.SKIP`), so one bad link cannot disable the whole
+    safety pass.
+
     Args:
         source: File or directory to scan.
         report: When ``True``, render the detailed markdown report.
@@ -368,7 +519,13 @@ def run_scan(
     """
     _require_existing(console, source, "Source path")
     config = load_config(resolve_config_path(vault, None))
-    scanner, summary = _scan_source(source, config)
+    scanner, summary = _scan_source(
+        source,
+        config,
+        console=console,
+        label="source",
+        policy=SymlinkPolicy.SKIP,
+    )
     render_summary(summary, console)
     if verbose:
         render_matches(summary, console)
@@ -409,6 +566,16 @@ def _atomic_write(file_path: Path, content: str) -> None:
     place with :func:`os.replace`. The swap is atomic on POSIX and best
     effort on Windows. If any step fails the temp file is unlinked so
     the original on-disk file is left untouched.
+
+    When *file_path* is a symlink, :func:`os.replace` MATERIALISES it: the
+    link is replaced by a regular file rather than written through. That
+    side effect is deliberate, not incidental. Writing *through* a link is
+    the operation SEC-003 forbids; by the time execution reaches here both
+    ends of any surviving alias are inside the tree the operator named
+    (:func:`_named_path_escapes` for the named leaf,
+    :func:`_assert_no_escaping_symlinks` for descendants), and
+    ``scan_batch`` surfaces the alias's target as a file in its own right,
+    so the target is redacted on its own pass.
 
     Args:
         file_path: Destination path to rewrite.
@@ -469,7 +636,29 @@ def _write_redaction_audit(
     vault_path: Path,
     dry_run: bool,
 ) -> None:
-    """Append one audit entry per touched file under *vault_path*."""
+    """Append one audit entry per touched file under *vault_path*.
+
+    Each entry records ``source_path`` exactly as the pipeline saw it, never
+    resolved. That is truthful without being exhaustive, and the difference
+    matters to anyone reading the trail:
+
+    * for a directly-named path, because no admitted named path resolves
+      outside its own parent (:func:`_named_path_escapes`);
+    * for a descendant, because the walk guard and the scanner's own
+      candidate filter both decline children that escape the root.
+    * Residual: a path reached through a symlinked ANCESTOR component is
+      admitted by that leaf-only policy, and the audit then records the
+      as-supplied path rather than where the write landed.
+
+    Resolving audit paths would not close that residual — it would only
+    start disclosing real intra-tree paths behind ordinary aliases.
+
+    Args:
+        summary: Scan summary whose matches describe what was found.
+        files: Files touched by this run, in report order.
+        vault_path: Vault root owning ``00-Creek-Meta/audit/redact.jsonl``.
+        dry_run: Whether this run previewed rather than committed changes.
+    """
     audit_log = RedactionAuditLog(vault_path)
     grouped = _matches_by_file(summary)
     for file_path in files:
@@ -567,14 +756,37 @@ def run_apply(
         assume_yes: Skip the interactive confirmation prompt.
         console: Rich console sink.
         vault: Vault root for the audit log. Defaults to
-            ``load_config().vault_path``.
+            ``load_config().vault_path``. Contained like *source*: it is
+            both read from and *written to*, so an escaping link here is an
+            out-of-root write even when *source* is innocent.
+
+    Raises:
+        typer.Exit: With code ``1`` when *source* or *vault* is, or
+            contains, a symlink that escapes the tree the operator named.
     """
     _require_existing(console, source, "Source path")
+    # Before ``is_dir()``, which follows the link, and before ``load_config``,
+    # which would otherwise read the config *through* it (#1293).
+    _assert_named_path_contained(source, console=console, label="source")
+    # ``--vault`` is the second path the operator names, and the only one this
+    # mode WRITES to: the audit record lands in
+    # ``<vault>/00-Creek-Meta/audit/redact.jsonl``, creating that directory if
+    # needed. Guarding only ``--source`` would leave the audit trail — the
+    # record of what was touched — landing wherever a link points, with an
+    # entirely innocent source (#1293).
+    if vault is not None:
+        _assert_named_path_contained(vault, console=console, label="vault")
     if source.is_dir():
         _assert_no_escaping_symlinks(source, console=console, label="source")
     config = load_config(resolve_config_path(vault, None))
     vault_path = vault if vault is not None else config.vault_path
-    scanner, summary = _scan_source(source, config)
+    scanner, summary = _scan_source(
+        source,
+        config,
+        console=console,
+        label="source",
+        policy=SymlinkPolicy.REFUSE,
+    )
     render_summary(summary, console)
     if verbose:
         render_matches(summary, console)
@@ -625,12 +837,25 @@ def run_review(
         vault: Vault root (or any directory) to re-scan.
         verbose: When ``True``, also list every individual match.
         console: Rich console sink.
+
+    Raises:
+        typer.Exit: With code ``1`` when *vault* is, or contains, a symlink
+            that escapes the tree the operator named.
     """
     _require_existing(console, vault, "Vault path")
+    # Before ``is_dir()``, which follows the link, and before ``load_config``,
+    # which would otherwise read the config *through* it (#1293).
+    _assert_named_path_contained(vault, console=console, label="vault")
     if vault.is_dir():
         _assert_no_escaping_symlinks(vault, console=console, label="vault")
     config = load_config(resolve_config_path(vault, None))
-    scanner, summary = _scan_source(vault, config)
+    scanner, summary = _scan_source(
+        vault,
+        config,
+        console=console,
+        label="vault",
+        policy=SymlinkPolicy.REFUSE,
+    )
     render_summary(summary, console)
     if verbose:
         render_matches(summary, console)
--- a/creek-tools/creek/redact/scanner.py
+++ b/creek-tools/creek/redact/scanner.py
@@ -787,8 +787,13 @@ def _statistics_lines(summary: ScanSummary) -> list[str]:
     return lines
 
 
-def _resolves_within(child: Path, resolved_root: Path) -> bool:
+def resolves_within(child: Path, resolved_root: Path) -> bool:
     """Report whether *child*'s symlink target stays under *resolved_root*.
+
+    Public because it is now also the named-leaf containment predicate for
+    :mod:`creek.redact.cli_commands` (#1293), so the walked-child surface and
+    the named-path surface cannot drift into two subtly different definitions
+    of "inside".
 
     The failure arm is a deliberate classification, not a swallowed error:
     a loop (``RuntimeError``), an unreadable link (``OSError``), and a
@@ -846,8 +851,13 @@ def _scannable_candidates(dir_path: Path) -> tuple[list[Path], int]:
 
     Known gaps, deliberately out of scope here:
 
-    * #1293 — the SEC-003 write guard is directory-only, so
-      ``creek redact --apply <symlinked-file>`` still writes through the link.
+    * Escaping ANCESTOR components. The directly-named path is now tested
+      for containment by :mod:`creek.redact.cli_commands` before any mode
+      reads through it (#1293), but a path reached *through* a link one
+      level up — ``<root>/linkdir/a.md``, where ``linkdir`` is the escaping
+      link — is still admitted here. That is not an oversight: it is the
+      resolve-the-root / ``lstat``-the-leaf policy documented above, applied
+      consistently on both surfaces.
     * #1294 — the ingestor walks (e.g.
       ``creek.ingest.markdown.MarkdownIngestor._read_directory``) still follow
       symlinks out of the source root.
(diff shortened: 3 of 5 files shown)


Geoffe-Ga:
## Summary
Fixes #1293: the SEC-003 symlink guard only checked links discovered *while walking* a directory tree, never the path the operator named directly on the command line. This PR adds a leaf-only containment check (`_named_path_escapes`/`_assert_named_path_contained`) for the directly-named `--source`/`--vault` path, applied via a required `SymlinkPolicy` (REFUSE for `--apply`/`--review`, SKIP-and-count for `--scan`) at the shared `_scan_source` chokepoint, and closes a second escape the issue didn't mention: an escaping `--vault` on `--apply` writing its audit record out-of-tree.

The PR description is notably rigorous: it identifies that the issue's stated *mechanism* ("writes through the link") is factually wrong (`os.replace` replaces the link, doesn't follow it), verifies the real defects by running the CLI directly, and documents three deliberately-left residuals (leaf-only predicate, `Pipeline._run_redaction`'s own traversal, `--scan --vault` read-escape filed as #1359).

## Strengths
- Correct core predicate: `_named_path_escapes` reuses the promoted `resolves_within` rather than restating containment logic, so the walked-child and named-leaf surfaces share one definition of "inside." Hand-traced against the PR's own examples (`~/vault -> ~/Dropbox/vault` admitted; `~/vault/inbox -> /Volumes/Export` refused) and both check out.
- Guard ordering is right and is actually pinned by tests, not just asserted in prose: `_assert_named_path_contained` runs before `is_dir()` (which follows the link) and before `load_config` (which would read `creek_config.yaml` *through* a symlinked vault). `test_redact_{apply,review}_refusal_reads_no_config_through_the_link` catch exactly this by recording whether `load_config` was ever invoked.
- Found and fixed a real bug beyond the issue's scope during self-review: an escaping `--vault` on `--apply` is a write-path out-of-root escape (audit log), independent of `--source`. Has its own dedicated red-first test (`test_redact_apply_refuses_an_escaping_vault_symlink`) that asserts on the filesystem, not just the exit code.
- `SymlinkPolicy` as a required (non-defaulted) kwarg on `_scan_source` is a good structural choice — it converts "someone adds a fourth caller and forgets the guard" from a silent hole into a mypy build error, which is precisely the root cause of this issue (`run_scan` had no guard at all).
- Non-disclosure is deliberately tested: refusal messages and log records name only the as-supplied path, never the resolved target, verified via `caplog` plus bare-filename assertions (explicitly avoiding a vacuous `str(victim) not in output` given Rich's console-width truncation).
- Verified no stale references to the renamed `_resolves_within` → `resolves_within`, and all three `_scan_source` call sites (`run_scan`, `run_apply`, `run_review`) were updated consistently.
- Verified the PR's claim that `creek_mcp/tools/redact.py` is out of blast radius — it has its own independent `resolve_within_vault` confinement and doesn't call `_scan_source`.
- Threat model doc and the SEC-003 backlog entry are updated to match the implementation exactly, including the corrected mechanism and the accurate (non-oversold) claim about the leaf-only predicate.

## Security Concerns
None blocking. 🟢 Three residuals are left open, but each is disclosed rather than hidden, and none is a regression:
- Leaf-only predicate admits an escaping *ancestor* symlink component (`<root>/linkdir/a.md`) — matches the scanner's existing walk policy, not a new gap.
- `Pipeline._run_redaction` still traverses a directly-named symlinked directory (read-only there, pre-existing #1294 territory, explicitly stated as untouched).
- `--scan --vault <escaping link>` still reads config through the link (read-only, filed as #1359, with a stated rationale for not refusing there — refusing `--scan` would be a DoS on the safety pass itself).

## Problems
None found blocking merge.

## Code Quality
Nothing substantive to flag. The one thing worth a mention (not a blocker, not even really a nit): `_scan_source`'s inline `if policy is SymlinkPolicy.REFUSE: _assert_named_path_contained(...)` re-derives `_named_path_escapes(source)` a second time inside `_assert_named_path_contained` after already checking it in the caller's `if` — harmless (raises unconditionally in that branch) but is a redundant recomputation rather than reusing the already-known `True`.

## CLAUDE.md Adherence
✅ No `# noqa`/`# type: ignore`/skip markers introduced. ✅ Full Google-style docstrings including `Raises:` sections on every new/changed public-ish function. ✅ Required kwarg-only `policy` param strengthens mypy strict enforcement rather than weakening it. ✅ Conventional commit message (`fix(redact): ...`). ✅ `tests/test_cli_redact.py` lines 1–660 are untouched (verified via diff — only an added `import logging` plus appended tests), consistent with the "not weakened" claim. I could not execute `./scripts/check-all.sh` or `pytest` in this sandboxed review session (tool execution required interactive approval I couldn't grant), so the exit-0 claim and the "468 tests pass" claim are not independently re-run here — but the diff's logic, naming, and test-assertion content are internally consistent and were traced by hand against the described behavior.

## Testing
Strong. New tests cover: named-file escape (apply), named-directory escape (apply, the more damaging case), review's equivalent, two ordering pins (no config read before refusal, for both apply and review), the new `--vault` write-escape, the distinct `--scan` skip-and-count contract, the chokepoint-level policy-enforcement regression guard, an over-breadth guard (intra-tree alias still works), and two "ordering pin" tests documenting that dangling/looping links exit 2 via pre-existing `_require_existing` rather than being reclassified. The one already-passing-at-HEAD assertion (named-file victim bytes unchanged) is explicitly labeled as a regression pin rather than misrepresented as a red.

## Documentation
Threat model (`docs/security/threat-model.md`) is updated thoroughly and accurately, including the corrected mechanism, the two contracts (refuse vs. skip), all residuals, and the user-visible behavior change (aliased export trees are now refused). SEC-003's backlog entry is updated to describe scope precisely rather than just re-marking "resolved."



## Verdict: LGTM

The security fix is correct by hand-verified trace (leaf-only symlink predicate, REFUSE vs SKIP policy split, guard ordering before is_dir()/load_config), the rename of _resolves_within→resolves_within and the _scan_source signature change are applied consistently at all three call sites with no stale references, and the new tests genuinely exercise the claimed red/green behavior including two non-obvious ordering pins (no config read through the link) and an over-breadth guard. No CLAUDE.md violations found (typed, docstringed, no noqa/type:ignore, threat-model doc updated, residuals filed as a real issue). Nothing here rises to a blocking concern or even a substantive suggestion worth withholding LGTM for.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5248029243 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Post-sync re-verification

Merged `main` (PR #1352 `creek author` HARD privacy gate, PR #1358 weighted-classify provenance). Re-ran the security argument rather than assuming the merge was inert.

**No new caller of the redaction surface.** Callers of `scan_batch`, `scan_file`, `_scan_source`, `run_scan`, `run_apply`, `run_review` re-enumerated across `creek/` and `creek_mcp/` on the merged tree: 30 references, identical to pre-merge apart from a +1 line shift in `cli.py` (the merge added a `TYPE_CHECKING` import above them). None of the ten merged files imports `creek.redact`. `creek/author/checks.py` does use a `scan`, but it is `creek.generate.ai_style.scanner.scan` — a different scanner. `creek/redact/*`, `tests/test_cli_redact.py` and `creek/pipeline.py` are byte-identical either side of the merge.

**The `--apply --vault` out-of-root audit write is still closed.** Escaping `--vault` symlink with an innocent `--source`, run through the real `creek` entry point: exit 1, refusal names the path as supplied, victim directory still empty, source unredacted.

**Mutation battery: zero survivors** — but only after a fix.

Re-running the six mutants surfaced one survivor the earlier battery had not formulated: deleting `_assert_named_path_contained(source, ...)` from `run_apply` left the suite green. Not a merge regression (the redact files are byte-identical across the merge) and not a hole in the security contract — the `_scan_source` chokepoint still refuses, so the out-of-root write stays closed either way. The two layers were simply indistinguishable by outcome.

The handler guard's unique contribution is *ordering*, which is why it sits above `if source.is_dir():`. Spying on `os.walk` with a named escaping symlinked directory as `--source`:

- guard present — **0** directories enumerated;
- guard removed — the link **and every subdirectory beneath it** walked before the refusal.

That is a structure disclosure of the victim tree, and nothing failed when it happened. Added one test (`test_redact_apply_refuses_a_named_symlinked_directory_before_walking_it`) asserting no walk root resolves inside the target; verified green unmutated and red under the mutant. All six mutants now die.

Kept in this PR rather than deferred: it makes a line already in this diff load-bearing, and no production code changed.

Gate 2 re-run on the merged tree: **exit 0**, 12/12 gates, 8889 passed, coverage 94.41%. Protected test block (lines 1–660) re-confirmed byte-identical to `main` apart from the added `import logging`.

Geoffe-Ga:
## Summary
Closes #1293 by extending SEC-003's symlink containment guard from "walked descendants only" to also cover the path the operator directly names on the command line, across `--scan`/`--apply`/`--review`. Adds a `SymlinkPolicy` enum (REFUSE for the write paths, SKIP for the read path) enforced at the single `_scan_source` chokepoint, a leaf-only `_named_path_escapes`/`_assert_named_path_contained` pair built on a promoted (now-public) `resolves_within`, and a second guard on `--apply --vault` (the audit-log write target, found during the author's own Gate 2.5 self-review). Ships with 789 lines of new tests and a substantially rewritten threat-model section.

## Strengths
- Correctly identifies and corrects a wrong premise in the original issue (`os.replace` replaces a symlink, it doesn't write through it) rather than implementing to a broken spec — and records the correction rather than silently deviating.
- Finds and fixes a second, more severe defect (named symlinked *directory* causes a genuine in-place out-of-root write via `is_dir()` following the link) and a third (`run_scan` had no guard at all) that the original issue didn't name.
- Finds and fixes a fourth escape (`--apply --vault <link>` as an out-of-root **write** target for the audit log) during self-review, with its own dedicated red-first test (`test_redact_apply_refuses_an_escaping_vault_symlink`) rather than deferring it.
- Keeps the two containment contracts (write-refuses vs. read-skips-and-counts) genuinely distinct via a required, non-defaulted `SymlinkPolicy` kwarg — a real mypy-strict-enforced guardrail against a future caller silently inheriting the hole that caused #1293 (`run_scan` calling `_scan_source` with no policy stated).
- Two "ordering" tests (`test_redact_review_refusal_reads_no_config_through_the_link`, `test_redact_apply_refusal_reads_no_config_through_the_link`) pin a property prose can't check — that the handler-level guard fires *before* `load_config` reads a config file through the link — and the PR explains that mutation testing caught the handler guard as a no-op survivor once, which is exactly the kind of layer these tests now make load-bearing.
- I independently verified (via `git diff` on the merge-base blob vs. HEAD blob) that `tests/test_cli_redact.py` lines 1-657 are byte-for-byte unchanged except one `import logging` — matching the PR's own claim; nothing was weakened or deleted.
- Cross-checked blast radius: no caller of these redact CLI entry points exists under `crawdad/`, and `creek_mcp/tools/redact.py` goes through its own independent `resolve_within_vault` gate — the PR's "no other implicated caller" claim holds.
- The residuals section (leaf-only ancestor-component escapes, `Pipeline._run_redaction`'s own traversal, `--scan --vault` read-escape → filed as #1359) is honest about what's *not* fixed rather than overclaiming a "uniform refusal."

## Security Concerns
🟢 LOW — The fix is a genuine hardening with no waiver/bypass mechanism (no `--force`, env var, or config key), consistent with SEC-003's precedent.

🟢 LOW (documented residual, not new) — The leaf-only predicate means a symlink at an *ancestor* path component (e.g. `<root>/linkdir/a.md`) is still admitted; this mirrors the scanner's pre-existing `resolve-the-root/lstat-the-leaf` policy and is explicitly called out in both code comments and the threat-model doc, not hidden.

🟢 LOW (worth a mention, not blocking) — `run_apply`'s `--vault` guard only checks whether the *named* `--vault` path itself is an escaping symlink; it doesn't extend to a symlink placed at a descendant component under an otherwise-legitimate vault (e.g. `<vault>/00-Creek-Meta` or `<vault>/00-Creek-Meta/audit` itself being a symlink out of the tree), which would let the audit-log write escape even with a fully "clean" named `--vault`. This is the same class of gap as the already-accepted ancestor-component residual, just not enumerated for this specific new write path — a one-line addition to the residuals list (or a follow-up issue like #1359) would keep the threat model fully current.

## Problems
None found that block merge. One purely stylistic observation: in `_scan_source`, when `policy is SymlinkPolicy.REFUSE` and the path escapes, `_assert_named_path_contained` always raises (since the precondition `_named_path_escapes(source)` was already established by the enclosing `if`), so the `logger.warning("Skipping symlink...")` / `return scanner, ScanSummary(...)` lines immediately below are unreachable under REFUSE — they only execute for SKIP. Not a bug (control flow is correct), just a little easy to misread; a short comment noting this branch never falls through under REFUSE could help a future reader.

## Code Quality
- The `SymlinkPolicy` docstring and the pervasive "why, not what" comments throughout (e.g. the `_atomic_write` note on why materializing a symlink on replace is deliberate) are excellent and will save real time for the next person touching this file.
- Minor: `_assert_named_path_contained`, when called from inside `_scan_source`'s REFUSE branch, re-evaluates `_named_path_escapes(path)` a second time (it was just computed by the caller). Cheap (an `lstat` plus a `resolve`), and defense-in-depth is explicitly the point per the PR description, so not worth changing — just noting it's intentional duplication rather than an oversight.
- `resolves_within`'s promotion to public API is clean — grepped for lingering references to the old private `_resolves_within` name and found none.

## CLAUDE.md Adherence
✅ TDD: RED-before-GREEN is documented and the commit history shows a genuine red run pasted into the PR description, plus a later commit (`79123077`) adding a test after a mutation-testing survivor was found post-merge-from-main.
✅ `./scripts/check-all.sh` reported exit 0 per the PR description (not independently re-run in this review — see Testing note below).
✅ No `# noqa` / `# type: ignore` / skip markers introduced.
✅ Docstrings are thorough (Google-style, Args/Returns/Raises) on every new/modified public and private function — consistent with the ≥95% interrogate requirement.
✅ Commit messages follow Conventional Commits (`fix(redact): ...`, `test(redact): ...`).
✅ No scope creep — the diff touches exactly the files the fix requires (`cli.py`, `redact/cli_commands.py`, `redact/scanner.py`, threat-model doc, tests).

## Testing
Exceptionally thorough for a security fix: separate tests for the named-file case, the named-directory case (the load-bearing "bytes unchanged" red), an ordering-pin test that spies on `os.walk` to prove the guard fires *before* the tree is walked (not just before the write), two config-read-ordering pins for `--apply`/`--review`, the distinct `--scan` skip-and-count contract, an over-breadth guard (intra-tree alias still admitted), and two "this already passes at HEAD" ordering pins for dangling/looping links labeled honestly as non-reds. Disclosure assertions are careful to avoid a vacuous `str(victim) not in output` check (Rich truncates paths) in favor of `caplog` + bare filename checks. I was not able to execute the test suite or `check-all.sh` myself in this sandboxed review environment (tool execution beyond git/gh/grep was unavailable), so this rests on static tracing of the logic plus the PR's own reported gate results — I'd treat that as the one caveat on an otherwise very well-evidenced testing story.

## Documentation
`docs/security/threat-model.md`'s SEC-003 entry is rewritten to accurately describe the new mechanism, explicitly correct the original issue's wrong claim about `os.replace`, describe the leaf-only admission policy with a concrete example (`~/vault -> ~/Dropbox/vault`), and enumerate the three open residuals with issue references. Docstrings on all touched functions (`_scan_source`, `_named_path_escapes`, `_assert_named_path_contained`, `_atomic_write`, `_write_redaction_audit`, `run_apply`, `run_review`) are complete and explain rationale, not just mechanics.



## Verdict: LGTM

The named-path symlink guard is correctly implemented and consistently threaded through all three redact modes (verified by tracing _named_path_escapes/_assert_named_path_contained/_scan_source call ordering and the SymlinkPolicy split), the shared resolves_within predicate is a clean rename with no orphaned references, and I confirmed via blob-diff against the merge-base that tests/test_cli_redact.py lines 1-657 are byte-identical to main (only a pure append plus one `import logging`), matching the PR's own claim. Coverage is unusually thorough (ordering pins, TOCTOU-adjacent os.walk spy, config-read-ordering tests, mutation-battery narrative) and the threat-model doc is updated accurately. Nothing here blocks merge.

Geoffe-Ga:
**CI**: 10/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5248169994 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Suppression removed (root cause, not reference)

The `os.walk` spy carried an unreferenced `arg-type` suppression. Removed by fixing what caused it rather than by attaching an issue number: the parameters were annotated `object`, which makes delegation to `os.walk` (an overload set — str vs bytes paths) untypeable. A transparent passthrough spy types them `Any`, which splats into the overload cleanly, so nothing needs suppressing.

`Any` is imported inside the test function, leaving the module's pre-existing import block untouched. mypy sets no `disallow_any_explicit` and ruff selects neither ANN nor PL, so neither the annotation nor the local import needs a carve-out.

One thing worth recording for future reviews: `scripts/typecheck.sh` runs `mypy creek/ creek_mcp/`, so **tests are outside the mypy gate** and that suppression was never evaluated by Gate 2 at all. Pointed at this file directly, mypy now reports nothing for the spy. The two errors it does report are the pre-existing `os.replace` spy at line 324 — byte-identical to `main`, same `object`-annotation pattern, unrelated to this PR and not in scope here.

Re-verified after the change:

- new test still **passes** unmutated, still **fails** with `run_apply`'s source guard removed;
- Gate 2 **exit 0**, 12/12 gates, 8889 passed / 5 skipped, coverage 94.41%;
- protected block (lines 1–660) still byte-identical to `main` apart from the added `import logging`;
- no `type: ignore` / `noqa` / `pragma: no cover` anywhere in the PR diff.

Pushed fast-forward `7912307..72372b2`.

Geoffe-Ga:
## Summary
This PR closes #1293 by extending the SEC-003 symlink guard to cover paths the operator *names* directly (`--source`, `--vault`), not just symlinks discovered while walking a tree. It adds `SymlinkPolicy` (REFUSE for `--apply`/`--review`, SKIP for `--scan`) at the `_scan_source` chokepoint, a leaf-only `_named_path_escapes`/`_assert_named_path_contained` pair built on the newly-public `resolves_within`, and — found during the author's own Gate 2.5 self-review, not in the original issue — a second guard on `--apply`'s `--vault` (previously an unguarded out-of-root write via the audit log).

## Strengths
- Exceptionally rigorous self-verification: RED-first tests proven against real CLI runs (with md5s), a 6-mutant battery with 0 survivors, and a documented case where the first mutant survived the initial battery and a new ordering-pin test (`test_redact_apply_refusal_reads_no_config_through_the_link`) was added specifically to make the handler-level guard load-bearing rather than merely redundant with the chokepoint.
- The PR description corrects the *mechanism* claimed by the original issue (`os.replace` replaces a symlink, it doesn't write through it) after verifying against `_atomic_write`, rather than implementing a red test for a bypass that doesn't exist. This is exactly the kind of premise-checking that prevents shipping a fake red.
- Read (`--scan`, SKIP) and write (`--apply`/`--review`, REFUSE) contracts are deliberately kept distinct via a required, non-defaulted `SymlinkPolicy` enum parameter on the single chokepoint — makes it a mypy-strict build error for a future fourth mode to omit a symlink policy, which is precisely the class of omission that caused this bug (`run_scan` had no guard at all).
- Ordering is correct and tested: guards run before `is_dir()` (which follows the link) and before `load_config` (which would otherwise read `creek_config.yaml` through the link) — verified by spying on `os.walk` and by a recording `load_config` wrapper.
- No disclosure of the resolved target anywhere (log or console) — only the as-supplied path is named, consistent with the #1087 oracle-closing precedent. I independently traced every `_assert_named_path_contained`/`_scan_source` message and confirmed none leak the resolved target.
- Residual gaps (leaf-only predicate / escaping ancestor components, `Pipeline._run_redaction`'s read-only traversal, `--scan --vault` config read-escape) are honestly documented rather than glossed over, with the last filed as a separate tracked issue (#1359) rather than silently deferred. I spot-checked the `pipeline.py:478`/`:484` line citations in the threat-model doc against the actual file — both are exactly correct.
- `tests/test_cli_redact.py` lines 1-660 are untouched (only `import logging` added); all new tests are additive.

## Security Concerns
🟢 LOW — None blocking. I traced the containment logic (`_named_path_escapes` → `resolves_within`) by hand for the named-file, named-directory, dangling-link, and looping-link cases and found it sound and consistent with the existing (unmodified) walk-based `_assert_no_escaping_symlinks` guard, which shares the same predicate via the newly-public `resolves_within`. The `--vault` write-escape the author found during self-review (audit log landing outside the named tree with an entirely innocent `--source`) is real, well-tested (`test_redact_apply_refuses_an_escaping_vault_symlink`), and correctly fixed.

Minor, non-blocking observations:
- `run_apply` never calls `_require_existing` on `vault` before `_assert_named_path_contained(vault, ...)`, unlike `source` in the same function and `vault` in `run_review`. For a dangling-but-escaping `--vault` symlink this could produce exit 1 ("symlink") instead of exit 2 ("not found"), inconsistent with the exit-code contract pinned for `--source`. This is a pre-existing asymmetry (the original code never existence-checked `vault` in `run_apply` either), not a regression, and it isn't security-relevant — just a possible future exit-code surprise.
- TOCTOU: the check-then-open window between `_assert_named_path_contained`/`_assert_no_escaping_symlinks` and the actual read/write is unchanged from the pre-existing SEC-003 design; out of scope for this PR.

## Problems
None found blocking merge.

## Code Quality
- All non-blocking. The `SymlinkPolicy`/`_scan_source` chokepoint design is a clean way to prevent the exact defect class (a mode forgetting to call the guard) from recurring.
- Docstrings are unusually thorough (bordering on verbose for some helpers), but given this module's security-critical nature and the repo's own quality bar, that's a reasonable tradeoff rather than a defect.

## CLAUDE.md Adherence
- ✅ TDD: RED-before-GREEN explicitly demonstrated in the PR description with real failure output.
- ✅ No `noqa`/`type: ignore`/skip markers added; the one `noqa: TC001`/`TC003`-style pattern in `scanner.py` predates this PR.
- ✅ Docstrings: every new/changed public and private function has Google-style Args/Returns/Raises sections.
- ✅ Conventional commit-style title (`fix(redact): ...`).
- ✅ No lowering of thresholds; `resolves_within` rename is additive (kept internal call sites working via renamed reference), no dead re-exports.
- I could not execute `./scripts/check-all.sh` or the test suite myself in this review sandbox (no `uv`/pytest available), so I cannot independently confirm the claimed "468 tests pass, exit 0." I instead performed a manual, line-by-line trace of the new guard logic and cross-checked the threat-model doc's file/line citations against the actual source — both held up. This should be treated as a caveat, not a blocker, given the PR's own extensive self-reported verification (real CLI runs with md5 hashes, mutation testing, ordering-pin tests).

## Testing
Strong. Nine new tests cover: named-file escape (apply), named-directory escape (apply, load-bearing bytes-unchanged), ordering-before-walk (mutation-testing-motivated), review escape + review/apply config-read-ordering (two separate load_config-spy tests), the new `--vault` write-escape, scan's skip-not-refuse contract, the chokepoint-level policy-enforcement regression guard, and two "ordering pin" tests (dangling/looping links exit 2, not 1) that protect against a naive reordering of the existence check. Over-breadth is also covered (intra-tree alias still admitted). This is well above the bar for a security-sensitive fix.

## Documentation
Excellent. `docs/security/threat-model.md` is updated with an accurate, detailed account of what was broken, what was fixed, what remains open, and the user-visible behavior change (aliasing an external tree and naming the alias is now refused). SEC-003's entry is updated to describe both the walked-tree and named-path halves of the guard. Citations to `pipeline.py` line numbers were verified accurate.



## Verdict: LGTM

The fix is logically sound (verified by hand-tracing the containment predicate against named-file, named-directory, dangling, and looping symlink cases), keeps the read/write contracts correctly distinct, closes a real write-escape the author found during self-review, and is backed by thorough, well-targeted tests plus a clean mutation-testing pass. Documentation accurately reflects the code (line-number citations verified). The only gaps are a minor pre-existing exit-code asymmetry for a dangling `--vault` and my inability to execute the test suite in this sandbox — neither rises to a merge blocker given the depth of the author's own verification.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5248254244 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #660, opened 2026-06-26
feat(author): tier-filter the Writing Desk retrieval against the privacy ceiling


Geoffe-Ga:
## Role
Senior Python engineer in `creek/author/` (the FEAT-041 Writing Desk) and the CLI/MCP author surfaces, fluent in `PrivacyTier`, the retrieval/graph agents, and the medium-contract plumbing.

## Goal
Make the Writing Desk's Graph + Retrieval specialists **tier-filter** their evidence against the run's privacy ceiling: a fragment whose `privacy_tier` exceeds the ceiling is excluded from the `EvidenceBundle`. Flip the MCP author tool's `tier_ceiling_enforced` from `False` to `True`, and remove/correct the stale `#463` comments that claim this already landed.

## Context
- **The gap:** #463 (`feat(author): real Graph + Retrieval specialist agents`, CLOSED) shipped the real agents but **not** the tier-filtering its own comments promised. The MCP author tool still says so: `creek_mcp/tools/author.py:57,88,104-105` ("the ceiling is recorded and echoed... but the current stub specialists do not yet tier-filter retrieval"). `creek/author/agents.py` does a plain backlink walk + semantic retrieval with **no `privacy_tier` check**.
- **Ceiling source:** the MCP `author_tool` takes `privacy_tier_ceiling: TierCeiling = TierCeiling.OPEN` (`creek_mcp/tools/author.py:80`) but never threads it into `run_author` → the conductor → the specialists. The CLI `creek author` has no ceiling flag today (default to OPEN, or add `--include-tier`-style parity with `creek draft`).
- **Files:** `creek/author/agents.py` (Graph + Retrieval `gather`), `creek/author/conductor.py` (`run_author` / `Conductor.gather_evidence` thread the ceiling), `creek_mcp/tools/author.py` (pass the ceiling, flip the flag, fix comments), `creek/models.py` (`Fragment.privacy_tier`), `creek_mcp/tier_ceiling.py` (`TierCeiling`).
- **Precedent:** mirror `creek/classify/llm/privacy_filter.py` (`if tier == PrivacyTier.INTIMATE and not _allows_intimate(override): continue`) — reuse, don't reinvent.

## Output Format
A single PR:
- [ ] Thread a privacy ceiling from `run_author` / the CLI/MCP entry points into the specialists' `gather`.
- [ ] Graph + Retrieval agents drop fragments above the ceiling from their evidence (with provenance unchanged for the kept fragments).
[...]

PR #1361, opened 2026-08-11
fix(mcp): refuse the author draft the privacy gate condemned (#1353)

### LINKED ISSUE #1353: fix(mcp): the `author` verb returns the draft body alongside an ESCALATE privacy finding
Split from #1310 (PR #1352), which closed the same harm on the CLI.

## Problem

`creek_mcp/tools/author.py:85` (`_draft_response`) puts `"body": draft.body` in the success envelope unconditionally. When the HARD privacy gate fires, the verb answers `status: ok` with `verdict: ESCALATE`, a `privacy_compliance` HIGH finding — **and the protected text still in `body`**.

#1310 fixed this for `creek author`: the CLI now prints the verdict and the offending fragment ids but withholds the body and exits 1 (`creek/cli.py` `_emit_author_result`). Verdict parity between the two surfaces now holds; *shipping* parity does not.

## Why it is P1 rather than critical

The MCP envelope carries `verdict` in the same payload, so a well-behaved consumer can gate on it — unlike the CLI, where the verdict line just scrolls past a human. But a consumer that reads `body` without checking `verdict` reproduces the original #1310 harm.

## Tasks

1. Decide the contract: withhold `body` entirely, or return it under a distinct key that a naive consumer will not render (e.g. `withheld_body` plus `withheld_reason`). An agent client may legitimately need the text to *revise* the draft, which is why this is a contract call and not a straight copy of the CLI behaviour.
2. Mirror whichever choice into `_draft_response`.
3. Add a test through the real in-process server (the `creek.author` surface in `tests/test_wiring_contract.py` is the existing shape) driving a vault whose draft reproduces intimate text, asserting the protected text is absent from the serialized response.

## Acceptance criteria

- [ ] An MCP `author` call whose draft trips `privacy_compliance` does not return the protected text under `body`.
- [ ] The response still states the verdict and the offending fragment ids.
- [ ] The chosen contract is documented in `docs/mcp.md`.

Geoffe-Ga:
## Summary

`creek_mcp/tools/author.py::_draft_response` put `"body": draft.body` in the success envelope unconditionally. So when the Writing Desk's HARD privacy check fired, the MCP `creek.author` verb answered `status: ok` with `verdict: ESCALATE`, a `privacy_compliance`/`HIGH` finding — **and the protected text still on the wire**.

Measured on this branch before the fix (one `intimate` fragment whose title equals its body, `privacy_tier_ceiling=all`, `medium=research`):

```
status="ok"  verdict="ESCALATE"  provenance=3 entries
SECRET in body        -> True
SECRET in claims      -> True
SECRET in provenance  -> True
"findings" in envelope -> False
```

The text shipped **three times**, not once: `body`, `claims[*].claim`, and `provenance[*].claim_excerpt`. The issue named only `body`.

Now, when any finding carries `dimension == "privacy_compliance"`, the verb returns a refusal envelope: `status: "refused"`, with `body`, `claims` and `provenance` **omitted entirely** — not `""`, not `null` — plus `privacy_findings` and a corpus-free `reason`. It still states the verdict, the rounds, and the offending fragment ids and tiers.

This is the MCP twin of #1352 (issue #1310), which closed the same harm on `creek author`. Semantics are matched deliberately: the same module-local `_PRIVACY_DIMENSION` constant name, and the same trigger — the finding **dimension**, never `verdict == "ESCALATE"`.

### The contract decision: `body` is omitted, and `status` stops saying `ok`

Two independent guards, because either alone is escapable:

- **Key omission** over `""`/`null`. `resp["body"]` now raises `KeyError`. A falsy-but-present `body` is exactly the shape that lets a caller print nothing and believe it published a draft.
- **`status: "refused"`**, so the well-behaved `if resp["status"] == "ok"` caller skips it without needing to know about this gate at all. `refused` is pre-existing MCP vocabulary (`tier_ceiling.refusal_response`, already emitted by `creek.link`, `creek.report`, `creek.state.read`, `creek.journal`, `creek.upload`), so every consumer's `status != "ok"` branch handles it for free. Not `error` — nothing malfunctioned, and collapsing a successful refusal into `_error_response` would make a provider outage indistinguishable from a privacy stop.

`withheld_body` (floated in the issue) was **rejected**: the whole envelope is serialized to the client — CrawDad feeds it verbatim into an LLM composer — so renaming the key changes no bytes on the wire and only makes the leak harder to see.

**Consumer check, executed:** CrawDad's `IntentDispatcher` sets `ToolResult.body` to the whole serialized tool response, and `{{step.<id>.body}}` in the workflow DSL (`workflows.py:616`) addresses *that*, not this envelope's `body` key. Omitting the key breaks no CrawDad workflow.

### Three things the issue did not anticipate

- **`claims` and `provenance` leak too.** `claim_excerpt` is `claim.claim[:80]`, and on the deterministic path claims are fragment titles. Scrubbing `body` alone would have shipped the same sentence one key over.
- **Echoing `draft.findings` would re-leak.** `biographical_grounding` (`checks.py:176`) interpolates a sentence lifted from the drafted body, and `attribution_correctness` (`checks.py:536`) interpolates `claim.claim` — the cited fragment's title, on the deterministic path. Only `privacy_compliance` findings are echoed — hence the key is `privacy_findings`, not `findings`, so nobody can "complete" it by widening the filter without also renaming it.
- **The Rich-markup injection class does not apply here** — verified rather than assumed. `creek_mcp` imports no Rich; CrawDad renders this envelope into an LLM prompt / Discord markdown. A fragment id of `frag[bold]x` round-trips `json.dumps` byte-exact and reaches the refusal message verbatim. Copying #1352's `rich.markup.escape` would have been cargo-culting the fix rather than the reasoning, and would have mangled a real id in the one message an operator uses to find the offending file.

`tier_ceiling_enforced` stays `true`: it is a claim about the #660 retrieval filter, which ran. The breach is against a *different* gate — the medium contract's `default_privacy_tier` (`open` in all six shipped templates). At `ceiling=all` the caller **is** admitted to intimate content, which is precisely why this leak survived every `ceiling=open` probe in the suite.

## Test plan

New `tests/test_mcp_author_privacy.py` — 8 tests, hermetic, no LLM (the deterministic renderer reproduces the leak because `title == body`).

**Gate 1 RED, measured:** 6 failed, 2 passed. The 2 passes are the intended controls. The load-bearing failure is the whole-envelope sweep:

```
assert 'The night I...or one hour.' not in '{"status": ...rag-int"]}]}'
   ... "claim_excerpt": "The night I wept alone in the parked car for one hour."
```

The canary is 54 chars / 12 words **deliberately**: `claim_excerpt` truncates at 80, so a longer secret makes the sweep pass while 80 characters of intimate text still ship — I hit exactly that with a 137-char first attempt. And `_MIN_PROTECTED_LEAK_WORDS` is 4, so a shorter one stops tripping the gate.

**Mutation battery — 7 mutations, no survivors:**

| Mutation | Caught by |
|---|---|
| disable the refusal branch | 6 tests |
| widen the dimension filter | 2 tests, incl. `assert 4 == 1` |
| trigger on `verdict == ESCALATE` | both controls + 3 pre-existing tests |
| put `body` back on the refusal | 3 tests |
| put `claims`/`provenance` back | 3 tests |
| keep `status: "ok"` | 3 tests |
| **echo all findings, trigger untouched** | **exactly 1** — `len(privacy_findings) == 1` |

That last one matters: it is the sharpest mutation, and it was a **survivor** until I found that the original test was vacuous. With a single-finding fixture, `all(dimension == "privacy_compliance")` is satisfied by a filter that excludes nothing. The test now runs on a corpus that trips `privacy_compliance` **plus** three `ontological_accuracy` findings, so the filter has something it is obliged to drop.

The two controls stop the fix degenerating into "always refuse": an ordinary `ESCALATE` still returns its body (measured: several existing author runs escalate with zero findings), and an intimate fragment merely *on disk* is not grounds for refusal (at `ceiling=personal` it never reaches the evidence).

**Regression:** `tests/test_mcp_tools.py`, `tests/test_wiring_contract.py`, `tests/test_mcp_read_gate.py`, `tests/test_mcp_server.py` — 577 passed, untouched. The wiring contract's `creek.author` surface runs at `ceiling=all` over an unclassified corpus and was the collateral-damage candidate; measured, no existing author test produces a `privacy_compliance` finding at all.

**Gate 2:** `./scripts/check-all.sh` exit **0**. The pre-change baseline was also 0.

## Notes for review

- **`CONTRACT_VERSION` is deliberately not bumped.** `refused` is pre-existing wire vocabulary already emitted by five other tools, so no client sees a token it has never seen; and the version's own guard pins the /v1 HTTP ADR and the `adepthood-v1` bundle, which `creek.author` is not part of (no httpapi route). The precedent that did bump (#1247) introduced a genuinely new status token. Flagging it as the judgement call I am least certain of.
- **The dry-run branch is not a sibling leak** — measured: it returns counts only (`{"claims": 3, "source_fragments": 1}`) and the static stage list.
- **Deliberately left alone:** no shared public `"privacy_compliance"` constant. #1352 has since merged and added its own private twin at `creek/cli.py:4954`, so the literal now lives in two places; promoting one public constant into `creek/author/checks.py` is filed as **#1362 (P3)** and is another lane's work. An MCP tool importing `creek.cli`'s private name would be the worse coupling, so this lane keeps its module-local `Final[FindingDimension]`; `tests/test_mcp_read_gate.py`'s `_PROBE_EXEMPT["creek.author"]`, already documented in that file as wrong and owned by #1279; and the CLI-side equivalent of the finding-message leak, which is #1352's surface, not this one.

Closes #1353
Refs #1310, #1352

---

## Post-sync re-verification (`main` merged in, #1352 now landed)

`fleet.sh sync 1353` merged `main` cleanly. That brought in **#1352** (the CLI half of this leak, issue #1310), **#1355** and **#1358**. Everything below was re-measured on the merged tree.

**Gate 2 after the sync: `./scripts/check-all.sh` exit `0`** (8881 unit passed; 8885 passed / 5 skipped in the coverage lane; per-file gate 256 files >= 80%).

### 1. Mutation battery re-run - all seven still die

| Mutation | Killed by |
|---|---|
| disable the refusal branch | 6 tests |
| widen the dimension filter | 2 tests, incl. `assert 4 == 1` |
| trigger on `verdict == ESCALATE` | both controls + 3 pre-existing (`test_author_tool_returns_draft_envelope`, `..._echoes_non_default_ceiling`, `..._returns_real_cited_draft`) |
| put `body` back on the refusal | 3 tests |
| put `claims`/`provenance` back | 3 tests |
| keep `status: "ok"` | 3 tests |
| echo all findings, trigger untouched | exactly 1 - `assert 4 == 1` |

**No survivors.** The seventh - the one that survived the first pass on a single-finding fixture - is still killed by the multi-finding corpus.

**One correction to the original table, and it is not merge-induced:** the "widen the dimension filter" row said *5 tests*; the measured figure is *2* (`test_refusal_carries_only_privacy_dimension_findings` with `assert 4 == 1`, and the `test_an_ordinary_escalate_still_returns_the_body` control with `assert 'refused' == 'ok'`). I checked whether the merge caused the drop: the only behavioural change the merge made anywhere on the author path is the `contract is None` branch in `check_privacy_compliance`, which is unreachable here (see §3), and `reflection.py`'s delta is docstring-only. The finding sets are identical pre- and post-sync, so the original 5 was simply wrong when written. The mutation still dies; only the count was overstated.

### 2. The 80-character canary still earns its keep

Two separate properties, both re-measured against the envelope:

- **Fails without the fix.** With the refusal branch disabled, `test_privacy_leak_is_absent_from_the_whole_serialized_response` fails and all 54 characters of the canary are on the wire.
- **Must stay under 80.** Correcting my own earlier reasoning: `body` is *never* truncated, so on a **fully** unfixed tree even a 141-character secret fails the sweep. The truncation can only make the sweep lie on a **body-scrubbed** tree - the tempting half-fix - where `claims` and `provenance` still carry `claim.claim[:_EXCERPT_LEN]` (`_EXCERPT_LEN = 80`, `conductor.py:177`). Measured against exactly that envelope:

| canary | sweep verdict | longest protected prefix on the wire |
|---|---|---|
| 54 chars | **FAILS** (correct) | 54 |
| 141 chars | PASSES (vacuous) | **80** |

The fixture docstring has been corrected to state the half-fixed tree it actually measured, rather than implying any unfixed tree would go green.

### 3. #1352's fail-closed change does **not** reach this surface

#1352 made `check_privacy_compliance` gate at `_NO_CONTRACT_CEILING` instead of returning `[]` when `contract is None`. **No input to `creek.author` newly produces a `privacy_compliance` finding**, because the `None` branch is unreachable from this verb: `run_author` (`conductor.py:513`) and `plan_author` (`conductor.py:563`) both obtain their contract from `load_medium_contract`, whose signature is `-> MediumContract` - it loads or raises `FileNotFoundError`, never returns `None`. Measured on every scenario in the new test file: `contract is None -> False`, `default_privacy_tier -> 'open'`.

So the refusal path is reachable in exactly the cases it was before the sync, and **nothing that should ship is now refused** - re-confirmed by the two controls, which still pass: an ordinary `ESCALATE` returns its body, and an intimate fragment at `ceiling=personal` returns an ordinary draft with `provenance == []`. (The change is a real hardening for the *CLI*, which did build a contract-less desk; that was #1352's own bug and its own fix.)

### 4. CrawDad consumer conclusion holds

Re-confirmed on the merged tree by reading the executed path: `MCPSession.call_tool` returns `_text_payload(result.content)` - the whole serialized response as one string (`crawdad/crawdad/mcp_client.py:161-178`) - which `IntentDispatcher` assigns to `ToolResult(body=body)` (`crawdad/crawdad/dispatcher.py:198-206`). `{{step.<id>.body}}` resolves to `result.body` (`crawdad/crawdad/workflows.py:616`), i.e. that same whole-response string, and `body` is the *only* addressable field. Nothing addresses the envelope's `body` key; a repo-wide sweep for a consumer reading `["body"]` off an author response finds none. **Omitting the key breaks no CrawDad workflow.**

### Also trued up in this push

The sync shifted every line number this lane's comments cited into `creek/author/checks.py`. Re-verified and corrected against the merged tree: message construction `323-327 -> 342-346`, `biographical_grounding` `164 -> 176`, `attribution_correctness` `517 -> 536`, `_MIN_PROTECTED_LEAK_WORDS` `216 -> 228`. `conductor.py:177`, `conductor.py:430-436`, `agents.py:145`, `voice.py:134` and `tier_ceiling.py:264` were re-checked and are unchanged. The test module docstring cited `author.py:69` - a line this fix deletes - and described the defect in the present tense; both fixed.


Geoffe-Ga:
## Summary
PR #1361 fixes a real data-leak bug in `creek_mcp/tools/author.py`: when the Writing Desk's HARD privacy check condemned a drafted body (a `privacy_compliance` finding), the MCP `creek.author` verb still answered `status: "ok"` with the protected text riding along in `body`, `claims[*].claim`, and `provenance[*].claim_excerpt`. The fix adds a `_privacy_refusal_response` path that returns `status: "refused"` and omits all three content keys whenever any finding carries dimension `privacy_compliance`, mirroring the CLI-side fix already shipped in #1352. It's a well-scoped, defensively-documented change with a strong, targeted test suite.

## Strengths
- Correctly identifies and closes the *full* leak surface, not just the one key (`body`) the linked issue named — the PR text and code both show the author noticed `claims[*].claim` and `provenance[*].claim_excerpt` also carry the secret via `claim_excerpt = claim.claim[:80]`, verified by reading `creek/author/conductor.py`.
- Uses key omission (not `""`/`null`) plus a non-`"ok"` `status`, so both a naive `resp["body"]` access and a `status == "ok"` gate fail safe. Verified the `refusal_response()` helper is genuine pre-existing MCP vocabulary shared by five other tools (`creek_mcp/tier_ceiling.py:264`), not a bespoke shape.
- Explicit, verified reasoning for **not** reusing `AuthoredDraft.model_dump()` in the refusal builder — I confirmed `rendered_text` is a `@computed_field` alias for `body` (`creek/author/models.py:235-237`) that a dump would smuggle back onto the wire. This is the kind of subtle correctness bug the PR pre-empted rather than introduced.
- Filters echoed findings to `privacy_compliance` only (`privacy_findings`, not `findings`) — verified `biographical_grounding` (`checks.py:171`) and `attribution_correctness` (`checks.py:532`) do interpolate corpus text into their own messages, so echoing the full finding list would have re-leaked via a different door. Trigger is the finding *dimension*, never `verdict == "ESCALATE"`, correctly avoiding false-positive refusals since escalation is routine on this surface.
- Test suite (`tests/test_mcp_author_privacy.py`, 8 tests) is unusually rigorous: a whole-envelope `json.dumps` sweep (not key-by-key, so it can't be satisfied by scrubbing only the "obvious" key), a fixture with a carefully bounded secret length (54 chars, under the 80-char provenance truncation, over the 4-word HARD-gate floor) with the reasoning documented inline, two explicit regression controls (ordinary `ESCALATE` still returns body; an intimate fragment excluded by ceiling doesn't trigger a refusal), a mixed-findings fixture that makes the dimension-filter test non-vacuous, and a hostile-fragment-id JSON round-trip test. The PR description's mutation-testing table (7 mutations, all caught) is a credible signal of test strength.
- Consistent with the CLI-side precedent (#1352/#1310): same constant name (`_PRIVACY_DIMENSION`), same trigger semantics — verified by reading the analogous CLI code path in `creek/cli.py`.
- `docs/mcp.md` updated with matching detail, including the three "easy to misread" callouts and the rejected `withheld_body` alternative with its rationale.

## Security Concerns
None outstanding — this PR *closes* a 🟡 HIGH privacy-leak vulnerability (protected/intimate-tier text disclosed over the MCP wire despite the reflection gate flagging it). No new vulnerability introduced. 🟢 LOW: the fix omits Rich-markup escaping for fragment ids in the refusal message; this is explicitly reasoned about and verified in-repo (`creek_mcp` imports no `rich`; the JSON wire quotes brackets as data) rather than assumed, and is backed by a round-trip test (`test_a_hostile_fragment_id_survives_the_json_envelope`).

## Problems
None found that block merge. One point worth a follow-up, not a blocker:
- The PR description states a "Consumer check, executed" against CrawDad's `IntentDispatcher`/`ToolResult.body` handling, confirming no CrawDad workflow breaks from the omitted key. That check happens in a sibling repo/package this review has no visibility into, so it could not be independently re-verified here — flagging for awareness, not requesting changes, since the reasoning given (workflow DSL addresses `ToolResult.body`, a different `body`, not this envelope's key) is internally consistent and specific enough to trust.

## Code Quality
- `# The literal is deliberately re-stated rather than imported...` (`creek_mcp/tools/author.py:44-52`) is a reasonable, well-justified duplication call — flagged by the author itself as a "P3 eventual cleanup," which is the right way to leave a deliberate wart.
- A couple of line-number references in the new module's docstrings (e.g., `checks.py:164`, `checks.py:517`) point a line or two off from the exact `message=` construction inside the referenced functions — harmless (`checks.py:164` is the top of `check_biographical_grounding`, not the interpolation line itself) but could drift further out of sync on future edits. Not blocking.

## CLAUDE.md Adherence
✅ Commit message follows Conventional Commits (`fix(mcp): ...`) and is on a feature branch (`issue/1353-issue`), not a direct push to `main`.
✅ TDD gate: PR description documents Gate 1 RED (6 failed / 2 passed with the intended failure signature) before the GREEN implementation.
✅ Docstrings are Google-style with Args/Returns sections throughout the new/changed functions; the module docstring, `_privacy_refusal_response`, and `_draft_response` docstrings are unusually thorough.
✅ Type hints: `Final[FindingDimension]` annotation deliberately chosen so mypy strict catches a typo in the dimension literal; `dict[str, Any]` return types match the surrounding module's existing convention.
⚠️ Could not independently execute `./scripts/check-all.sh`, `./scripts/test.sh`, or mypy/ruff in this review environment (sandboxed; command execution required approval that wasn't grantable here) — CI checks (Static Analysis, Pylint, Tests & Type Checking 3.11/3.12/3.13, Integration & E2E) were still `IN_PROGRESS` at review time. Code Complexity Analysis and Build Distribution had already reported SUCCESS. Static code review found no complexity, typing, or lint red flags (all imports used, functions reasonably small, no obvious cyclomatic-complexity concerns), but this is not a substitute for green CI before merge.

## Testing
Strong. Beyond the summary above: the new suite includes both a positive reproduction and two explicit negative controls that prevent the fix from degenerating into "always refuse" or "trigger on verdict instead of dimension" — exactly the failure modes a less careful patch would introduce. Regression coverage (`test_mcp_tools.py`, `test_wiring_contract.py`, `test_mcp_read_gate.py`, `test_mcp_server.py`) is claimed unaffected (577 passed) per the PR description; not independently re-run here but the claim is specific and falsifiable, and the diff to `author.py` is additive/branching in a way that shouldn't disturb the existing non-privacy-finding path (verified by reading `_draft_response`: the new filter only intercepts when `leaks` is non-empty, otherwise falls through unchanged to the pre-existing envelope construction).

## Documentation
`docs/mcp.md` is updated thoroughly and in the same voice/rigor as the code docstrings, including the "three details that are easy to misread" section and the explicit rejection-with-rationale of the `withheld_body` alternative design. Matches the actual implementation exactly on spot-check (key omission, `status: "refused"`, `privacy_findings` shape, `tier_ceiling_enforced` staying `true`).



## Verdict: LGTM

This is a well-tested, correctly-scoped security fix: the implementation was cross-checked against the referenced check/model code (verified the triple-leak claim, the computed_field aliasing hazard, and the corpus-free finding messages all hold up), the test suite includes meaningful regression controls that prevent the two obvious over-fixes, and CLAUDE.md process requirements (conventional commits, TDD, docstrings, typing) are met. Nothing here blocks merge; the only caveats (CI still in progress at review time, an unverifiable cross-repo consumer claim, minor docstring line-ref drift) are non-blocking notes, not substantive suggestions that should change the PR.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5248087385 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
Fixes a real privacy leak in the `creek.author` MCP tool: when the Writing Desk's HARD privacy check condemned a draft (a `privacy_compliance` finding), the tool still returned `status: "ok"` with the protected text present three times over (`body`, `claims[*].claim`, `provenance[*].claim_excerpt`). The fix filters findings by dimension, and on a hit returns a `status: "refused"` envelope that omits `body`/`claims`/`provenance` entirely, echoing only corpus-free `privacy_findings`. This is the MCP-surface twin of the CLI fix in #1352 (issue #1310).

## Strengths
- Root-caused correctly: found and closed all three leak vectors (`body`, `claims`, `provenance`), not just the one named in the issue.
- Key *omission* rather than `""`/`null` is the right call — `resp["body"]` now raises instead of silently returning nothing, which is the failure mode that actually matters for a caller.
- Trigger is the finding *dimension*, not `verdict == "ESCALATE"` — verified against real behavior (ordinary escalations with zero/non-privacy findings still return bodies), avoiding an over-broad fix that would degenerate into "always refuse."
- Only `privacy_compliance` findings are echoed (`privacy_findings`, not `findings`) because sibling checks (`biographical_grounding`, `attribution_correctness`) interpolate corpus text into their own messages — correctly identified as a re-leak vector the issue didn't anticipate.
- Deliberately avoids `draft.model_dump()` in the refusal path because `AuthoredDraft.rendered_text` is a `@computed_field` alias for `body` that would smuggle the condemned text back onto the wire under a second key — verified in `creek/author/models.py:235-239`, a genuinely subtle trap.
- Reuses the existing `refusal_response`/`status: "refused"` vocabulary already emitted by five other MCP tools, so well-behaved `status != "ok"` callers handle this for free — no new token, no client-side changes required.
- Test suite (8 tests) is unusually rigorous: a whole-envelope `json.dumps` sweep (not key-by-key, so it can't be satisfied by patching only the key someone remembers), a hostile-fragment-id round-trip check, and two explicit controls (ordinary ESCALATE still returns a body; an intimate fragment excluded by ceiling never trips a refusal) that prevent the fix from degenerating into "refuse everything."
- The `_MIXED_SECRET` fixture redesign, done specifically to make the dimension-filter test non-vacuous (a single-finding corpus can't distinguish "filtered" from "not filtered"), shows real mutation-testing discipline, and the PR body documents exactly the survivor that motivated it.
- Independently verified all the load-bearing claims in the PR description against the current code (`FindingDimension` literal, `ReflectionFinding` fields, `check_privacy_compliance`'s corpus-free message, the CLI's `_PRIVACY_DIMENSION` twin at `cli.py:4952`) — everything checks out.
- Second commit is an honest, disciplined rebase-note: after `#1352` landed on `main` and shifted line numbers, the author re-verified every citation against the merged tree rather than leaving stale references, and even corrected an imprecise claim in the original commit message (the canary docstring's truncation explanation).

## Security Concerns
None outstanding — this PR *is* the security fix. 🟢 No new concerns introduced; the Rich-markup injection class from #1352 was checked and correctly found inapplicable here (verified: `creek_mcp` imports no `rich`, and the JSON wire quotes bracket characters as data via `json.dumps`), with a passing round-trip test (`test_a_hostile_fragment_id_survives_the_json_envelope`) rather than just an assumption.

## Problems
None blocking.

## Code Quality
- The `CONTRACT_VERSION` bump decision (not bumped) is a genuine judgment call worth a second look: `creek_mcp/contract.py:8` says to bump "when the tool surface or its semantics change," and this PR does change `creek.author`'s response semantics (a new `status` value for this tool, three keys that can now be absent). The author's counter-argument — `refused` is pre-existing vocabulary and `creek.author` isn't part of the `/v1` HTTP-ADR bundle the version pins — is reasonable and is flagged explicitly in the PR as the least-certain call, but it's worth the reviewer's own sign-off rather than rubber-stamping.
- `# noqa`/`type: ignore`-style escape hatches: none introduced.

## CLAUDE.md Adherence
✅ TDD: PR body documents Gate 1 RED (6 failed/2 passed) before implementation.
✅ Coverage/tests: 8 new tests plus a 7-mutation battery with no survivors; regression suite (577 tests) confirmed unaffected.
✅ Docstrings: extensive Google-style docstrings on every new function, each explaining *why*, not just *what* — consistent with repo convention.
✅ `docs/mcp.md` updated in the same PR to describe the new refusal contract, per the DRY/single-source-of-truth principle (§1.2).
✅ Commit messages follow Conventional Commits (`fix(mcp): ...`, `docs(mcp): ...`) and are unusually precise about what changed and why.
✅ Gate 2 (`./scripts/check-all.sh`) reported exit 0 in the PR description. I was not able to independently re-run it in this review environment (test/lint commands required interactive approval that wasn't available), so I relied on static verification of every factual claim in the PR body against the current source tree — all confirmed accurate (see Strengths).

## Testing
Strong. The whole-envelope `json.dumps` sweep is the right acceptance test (it can't be gamed by scrubbing only the key someone remembers), the two controls guard against the "refuse everything" degenerate fix, and the `_MIXED_SECRET` fixture rebuild specifically closes a vacuous-test gap the author found via mutation testing. Not insufficient in any way.

## Documentation
`docs/mcp.md` is updated thoroughly and accurately — cross-checked the described contract (status, omitted keys, `privacy_findings` shape, `tier_ceiling_enforced` semantics) against the implementation and it matches exactly, including the subtle point that `tier_ceiling_enforced` stays `true` because it's a claim about a different gate (#660 retrieval filter) than the one that fired here.



## Verdict: LGTM

No bugs, no security gaps, no CLAUDE.md violations, and no missing test coverage — every factual claim in the PR description was independently verified against the current source (models, checks.py, refusal_response, the CLI twin) and holds up. The one open question (whether CONTRACT_VERSION should bump) is explicitly disclosed by the author as a judgment call with a defensible rationale, not a defect, so it doesn't rise to a blocking or even a must-address suggestion.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5248242304 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1365, opened 2026-08-11
fix(api): build the /v1 header merge so Vary cannot be overridden (#1128)

### LINKED ISSUE #1128: fix(api): build the /v1 error-response header merge so Vary cannot be overridden

Geoffe-Ga:
## Summary

`creek_mcp/httpapi/errors.py:88` built the response header block as
`{VARY_HEADER: CEILING_HEADER, **(headers or {})}`, so a caller-supplied `headers` dict
silently overrode the standing `Vary: X-Creek-Tier-Ceiling`. `json_response` is the single
builder every `/v1` response passes through, and that header is the only thing stopping a
shared cache from serving one caller's ceiling-filtered body to a caller at another ceiling.

**No production call site passes a `Vary` today** — the only caller handing this builder extra
headers is `_challenge_for`, and it passes only `WWW-Authenticate` — so nothing was broken in
the field. That is the argument for pinning it now rather than the argument against: the first
call site that did pass one would have deleted the protection silently.

The fix is deliberately larger than the one-line merge reversal the issue proposed, because the
one-liner leaves two of the three defects in place. Verified: reversal alone still fails 10 of
the new assertions (mutant M1 below).

### The merge policy, and why each part

1. **The standing `Vary` wins**, by explicitly partitioning colliding caller keys out rather
   than by dict-literal ordering.
2. **Names are matched case-folded.** `{"vary": ...}` is a distinct dict key, and Starlette's
   `Response.init_headers` lowercases each key independently and never deduplicates — so an
   unfolded merge emits **two** `vary` lines and leaves the intermediary to choose which one
   keys the cache. Confirmed empirically at the pre-fix HEAD: `headers={"vary": "Cookie"}`
   rendered `Headers([('vary', 'X-Creek-Tier-Ceiling'), ('vary', 'Cookie'), ...])`.
   **A merge that does not fold case has moved the bug, not fixed it.**
3. **`Vary` values UNION rather than replace.** Discarding a colliding caller token would close
   this hole and open its mirror image: a response whose body really does turn on
   `Accept-Encoding` but that claims to turn only on the ceiling invites a cache to key on
   *less* than the truth — the same defect pointed the other way. Union is also the only
   monotone direction: RFC 9111 §4.1 makes reuse a conjunction over the nominated field names,
   so adding a token can only shrink the set of requests an entry may be served to, never widen
   it. Standing token first, caller tokens in supplied order, deduplicated case-folded.
4. **Caller tokens outside RFC 9110 `tchar` are dropped whole, not repaired**, and only `OWS`
   (space and horizontal tab) is trimmed before that test — a bare `.strip()` would tidy
   `"Cookie,\r\nAccept"` into the perfectly legal token `Accept`, repairing exactly the input
   the builder refuses to repair. This is the first path here that parses and reassembles
   caller-shaped text into a header value, and Starlette encodes header values latin-1 verbatim
   with no CRLF validation. An allowlist rather than a CRLF blocklist because the set of legal
   field-name characters is small, closed and written down, and the dangerous set is none of
   those. `*` is a `tchar` and is deliberately kept: RFC 9111 §4.1 tests for it by membership,
   so a response carrying it never matches a stored entry — filtering it would convert an
   author's "do not cache this" into a response servable on the ceiling alone.

### Not changed, deliberately

- `creek_mcp/api/openapi.py` `DOCUMENT_DESCRIPTION` — feeds a byte-pinned committed OpenAPI
  golden file, and its claim "Every response carries Vary: X-Creek-Tier-Ceiling" stays true
  under union. Same for `docs/api.md`. No ADR: this is a defect inside an existing decision.
- **No `STANDING_HEADERS` mapping.** `Vary` needs a union rule and #1129's `Cache-Control:
  no-store` needs replacement, so one `Mapping[str, str]` with one merge rule is already the
  wrong type for two members — and #1129's header is conditional on authentication, which a
  constant cannot express. What #1129 gets instead is a named seam: `_merged_headers`, whose
  docstring states the policy, and `_vary_tokens`, which seeds from a single standing token
  that #1129 turns into a tuple to add `Authorization` (its issue text asks for exactly that).
- **Passthrough header names and values are still unvalidated.** Only the `Vary` value is
  parsed. Acceptable today because `_challenge_for` is the sole caller and returns module
  constants, and uvicorn and h11 both reject CR/LF downstream. Recorded here as a scope
  boundary rather than filed, since there is no reachable defect to fix.

## Test plan

`tests/test_v1_api_admission.py`, appended to its existing
`# Vary — the cache cannot cross-serve two ceilings` section. 12 test functions, 18 cases with
parametrisation. Every assertion is on **rendered** headers — `raw_headers` for the builder
tests (Starlette's `Headers.__getitem__` returns only the first match and would hide a
duplicate `vary`, which is one of the defects under test) and httpx's `get_list` for the
end-to-end one.

**Gate 1 RED — 16 of 18 red at the pre-fix HEAD.** The two green by construction are stated as
such in their own docstrings: `test_a_non_colliding_caller_header_still_rides` is a deliberate
non-regression pin that `_challenge_for`'s output survives the new merge, and the `canonical`
case of the echo test is one HEAD happens to render correctly. Sample actual pre-fix output:
`headers={"Vary": "Accept-Encoding"}` rendered `vary: Accept-Encoding` — the ceiling token
gone; the hostile value reached the header block as raw `b'Cookie\r\nX-Creek-Injected: 1'`.

**Mutation battery — 12 mutants, each patching only `errors.py`, zero survivors:**

| Mutant | Tests red |
|---|---|
| M0 full revert to HEAD | 16 |
| M1 naive merge reversal (the issue's own proposed fix) | 10 |
| M2 discard caller `Vary` instead of union | 10 |
| M3 name comparison no longer folds case | 3 |
| M4 `tchar` allowlist downgraded to a CRLF blocklist | 3 |
| M5 standing token written last instead of first | 9 |
| M6 dedupe made case-sensitive | 2 |
| M7 non-colliding caller headers dropped | 2 (one a pre-existing auth test) |
| M8 only the *first* colliding `Vary` spelling collected | 1 |
| M9 empty token admitted (`bool(candidate)` dropped) | 1 |
| M10 `*` removed from `_TCHAR` | 1 |
| M11 bare `.strip()` repairs a CRLF-padded token | 1 |

Control after restore: 211 passed. M6 survived a first pass — the echo test used only the
lower-case spelling, which a half-folded dedupe passes by coincidence — so it is now
parametrised over all three spellings, and M6 dies.

**Gate 2:** `./scripts/check-all.sh` exit code 0, 12/12 checks. `errors.py` per-file coverage
93.10% (misses only the `TYPE_CHECKING` block).

**Gate 2.5:** `code-review-orchestrator` across the security, test, implementation and
documentation dimensions. It returned FIX_REQUIRED with no defect in the production logic and
five findings against the *evidence and the prose*: a load-bearing justification for the token
order cited to a test that provably cannot enforce it (`header_items` is a frozenset compared
within one process, and no `401` supplies a caller `Vary` at all — verified, and the citation
is now the byte-exact echo test); an overstated RED count; four surviving mutants; a vacuous
`assert b"x-creek-injected" not in [names]` that passes against the pre-fix builder because
`raw_headers` holds one tuple per dict key (now asserted over the serialised header block); and
three docstrings claiming a socket the `TestClient` never opens. All fixed. One nit declined:
the review asked to delete `test_a_caller_cannot_replace_the_standing_vary` as subsumed by the
union test — it is kept, with the subsumption stated in its docstring, because it is the
issue's own acceptance criterion and a named test for the property under repair is what a later
reader looks for.

Confirmed the #1127/#1351 `500` path still carries the correct `Vary`:
`test_a_non_routing_http_exception_echoes_no_detail` asserts `response.headers["vary"] ==
CEILING_HEADER` and stays green verbatim, as does the 56-probe byte-identity sweep in
`tests/test_v1_api_auth.py`.

Closes #1128
Refs #1129

### DIFF
--- a/creek-tools/creek_mcp/httpapi/errors.py
+++ b/creek-tools/creek_mcp/httpapi/errors.py
@@ -19,11 +19,42 @@ caller's ceiling-filtered response to another. Authentication sits *above* the
 ceiling gate, so there is no single middleware every response passes through on
 the way out with the ceiling in scope. There is, however, exactly one builder:
 :func:`json_response`. Routing every response through it is what makes the
-header unconditional.
+header unconditional — and, since #1128, unforgeable. The builder used to lay
+this header *under* the caller's, so any call site handing in a ``Vary`` of its
+own silently deleted the ceiling token; the merge now runs the other way. Note
+the narrowness of that claim: ``Vary`` is the only header this builder stamps
+and so the only one protected. Everything else a call site passes still rides
+through untouched, which is what a ``401`` needs — see :func:`_challenge_for`.
+
+It absorbs the colliding value rather than discarding it, and that is the
+load-bearing half of the policy. Dropping a caller's ``Vary`` would close this
+hole and open its mirror image — a response whose body really does turn on
+``Accept-Encoding`` but that claims to turn only on the ceiling invites a cache
+to key on less than the truth, which is this same defect pointed the other way.
+Union is also the only direction that is *monotone*: RFC 9111 §4.1 makes reuse
+a conjunction over the nominated field names, so adding a token can only ever
+shrink the set of requests an entry may be served to, never widen it.
+
+So the values union: the ceiling token first, then the caller's in the order
+supplied. The order is fixed rather than incidental because the rendered value
+is asserted byte-for-byte by
+``tests/test_v1_api_admission.py::test_a_caller_echoing_the_standing_token_does_not_double_it``,
+and because a header whose bytes depended on set iteration would differ between
+two worker processes answering the same request.
+
+Names are matched case-folded, because ``Vary`` and ``vary`` are one header on
+the wire and two keys in a dict — Starlette lowercases each key independently
+and never deduplicates, so an unfolded merge emits two ``vary`` lines and leaves
+it to the intermediary to decide which one keys the cache. Caller tokens outside
+RFC 9110's ``tchar`` set are dropped whole rather than repaired, and only ``OWS``
+— the space and the horizontal tab — is trimmed before that test, so a token
+cannot be *whitespace-stripped* back into legality either. A repaired token is
+still a token the caller chose, and this header's contents decide a cache key.
 """
 
 from __future__ import annotations
 
+import string
 from typing import TYPE_CHECKING, Any, Final
 
 from starlette.responses import JSONResponse
@@ -37,7 +68,7 @@ from creek_mcp.api.models import (
 from creek_mcp.api.routes import CEILING_HEADER
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping
+    from collections.abc import Iterable, Mapping
 
     from starlette.responses import Response
 
@@ -45,6 +76,34 @@ if TYPE_CHECKING:
 
 VARY_HEADER: Final[str] = "Vary"
 """The response header naming which request headers the body depends on."""
+
+_TCHAR: Final[frozenset[str]] = frozenset(
+    string.ascii_letters + string.digits + "!#$%&'*+-.^_`|~"
+)
+"""RFC 9110's ``tchar`` — every character a bare field name may contain.
+
+This is the entire admission test for a caller-supplied ``Vary`` token, so it is
+spelled out rather than approximated by a "no control characters" check: the
+things it excludes include the space, the colon and the carriage return, each of
+which turns one header value into something other than one header value.
+
+``*`` is a member and therefore survives, which is the right outcome even though
+it reads like a wildcard. RFC 9111 §4.1 tests for ``*`` by *membership* in the
+stored ``Vary`` value, not only as the whole value, and a stored response
+carrying it never matches — so an entry that would otherwise have been reusable
+becomes uncacheable. Dropping ``*`` would be the harmful direction: it would
+turn a response its author declared uncacheable into one a cache may serve on
+the ceiling alone. The token degrades closed, so it is kept.
+"""
+
+_OWS: Final[str] = " \t"
+"""RFC 9110's ``OWS`` — the only whitespace that may pad a list element.
+
+Trimmed before the :data:`_TCHAR` test, and deliberately narrower than
+:meth:`str.strip`'s default. A bare ``strip()`` also eats ``\\r`` and ``\\n``,
+which would quietly convert ``Cookie,\\r\\nAccept`` into the perfectly legal
+token ``Accept`` — repairing exactly the input this module refuses to repair.
+"""
 
 WWW_AUTHENTICATE_HEADER: Final[str] = "WWW-Authenticate"
 """The challenge header a ``401`` is obliged to send."""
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
Fixes #1128: `json_response`'s header merge previously let a caller-supplied `headers` dict silently override the standing `Vary: X-Creek-Tier-Ceiling`, which is the only thing preventing a shared cache from serving one caller's ceiling-filtered body to another. The fix rewrites the merge as a dedicated, case-folded, RFC 9110/9111-compliant union with strict token validation, and adds an 18-case test suite (plus a full-stack integration test through the real handler stack).

## Strengths
- **Correct RFC compliance**: `_TCHAR` matches RFC 9110's `tchar` grammar exactly (including keeping `*`, which is subtly correct per RFC 9111 §4.1 — dropping it would make an author's "uncacheable" declaration silently cacheable). `_OWS` is deliberately narrower than `str.strip()` to avoid repairing CRLF-padded tokens into legal ones.
- **Header-injection hardening**: caller-supplied `Vary` tokens are validated against an allowlist (not a CRLF blocklist) before being joined into the header value, closing a real injection vector given that Starlette encodes header values latin-1-verbatim with no CRLF validation (confirmed by reading `errors.py` and cross-checking all production `json_response` call sites — none besides `_challenge_for` pass extra headers, and it only passes `WWW-Authenticate`).
- **Union over discard/replace**: correctly identified that simply discarding a colliding caller `Vary` would trade one cache-key-integrity bug for its mirror image (a response claiming a narrower dependency than it has).
- **Deterministic ordering**: standing token first, caller tokens in supplied order, case-folded dedup — avoids set-iteration-order flakiness across worker processes.
- **Test rigor**: tests assert on `raw_headers`/`get_list` rather than `response.headers[...]`, correctly avoiding Starlette's/httpx's first-match masking of duplicate header lines — this is exactly the kind of assertion choice that would let the original bug hide. The PR description documents a 12-mutant kill matrix with zero survivors.
- Verified independently: no other production call site passes `Vary`, so this change is behavior-preserving everywhere except the one path it's designed to fix.

## Security Concerns
None found. 🟢 The change is itself a security hardening (closes a cache-poisoning-adjacent disclosure risk and a header-injection vector). The allowlist-based token validation is the right primitive (small closed legal set vs. an open-ended dangerous set).

## Problems
None found that block merge.

## Code Quality
- Minor: the test id `"non-latin-1"` for the `"Accépt"` malformed-token case (`tests/test_v1_api_admission.py`) is slightly misleading — `é` (U+00E9) is representable in Latin-1; the actual reason it's rejected is that it's outside the ASCII `tchar` set, not a Latin-1 encoding issue. Doesn't affect correctness of the test or the code under test, purely a label nit.
- The module/function docstrings are unusually long (verging on essay-length for a small module). This is clearly a deliberate house style here given the amount of subtle reasoning being pinned down (RFC citations, monotonicity argument, threat model), and it does the job CLAUDE.md's docstring-coverage requirement asks for, but a future reader skimming for the actual contract may want the TL;DR nearer the top.

## CLAUDE.md Adherence
- ✅ Coverage: PR reports 93.10% file coverage on `errors.py` (misses only the `TYPE_CHECKING` block), consistent with the ≥90% branch coverage / ≥80% per-file thresholds.
- ✅ Types: all new functions (`_is_vary`, `_is_token`, `_vary_tokens`, `_merged_headers`) are fully annotated; `Iterable`/`Mapping` imports correctly gated under `TYPE_CHECKING` with `from __future__ import annotations`.
- ✅ Complexity: all new functions are small and straightforward (well under the ≤10 cyclomatic complexity ceiling).
- ✅ Docstrings: Google-style with Args/Returns throughout, well above the 95% interrogate threshold.
- ✅ Commit message: `fix(api): build the /v1 header merge so Vary cannot be overridden (#1128)` follows Conventional Commits; body is thorough and explains rationale.
- ⚠️ Could not independently execute `./scripts/check-all.sh` or `./scripts/test.sh` in this review environment (sandbox blocked command execution), so the PR's claimed "12/12 checks, exit 0" and "211 passed" results are taken on the author's word rather than independently reproduced — static review of the diff found nothing that would contradict them.

## Testing
Strong. 18 parametrized cases covering: standing-token preservation, union (not replace/discard), fixed ordering, case-fold collision (including both `Vary`+`vary` present simultaneously), echoed-spelling dedup (parametrized over 3 spellings — correctly catches a half-folded dedupe bug that a single-spelling test would miss), empty-token/double-comma handling, the `*` wildcard survival case, four distinct malformed-token shapes (CRLF, space, colon, non-ASCII), the strip-vs-OWS-strip distinction, CRLF header-splitting resistance (asserted on the serialized byte block, not just `raw_headers` names — correctly avoids a vacuous assertion), non-colliding passthrough (`WWW-Authenticate`), and a full-stack integration test through real middleware. This is genuinely thorough test design, not just high coverage by line count.

## Documentation
Module docstring, `VARY_HEADER`/`_TCHAR`/`_OWS` docstrings, and function docstrings all clearly state the policy and its rationale (including RFC citations). PR description is exceptionally thorough, documenting the mutation-testing results and a prior `code-review-orchestrator` pass with fixes applied. No outdated docs found — `openapi.py`'s `DOCUMENT_DESCRIPTION` claim ("every response carries Vary") is correctly noted as still true under the new union policy.



## Verdict: LGTM

The fix correctly closes the Vary-header override hole (case-folded name matching, RFC 9110 tchar allowlist for injection safety, RFC 9111-compliant union merge) and is exhaustively tested with 18 parametrized cases plus a full-stack integration test; no blocking bug, security issue, or CLAUDE.md violation was found in careful static review of the diff.

Geoffe-Ga:
**CI**: 5/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5248258272 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1367, opened 2026-08-11
fix(ingest): a partially-pinned vault stops warning, so an unresolved conflict can mint a third duplicate


Geoffe-Ga:
## Problem

`creek/ingest/pipeline.py:unpinned_vault_warning` (added in #1329) tells an operator their vault needs the one-time `creek ingest --pin-source-ids` migration. Its predicate is:

```python
if ledger is None or len(ledger) > 0:
    return None
```

i.e. **"the ledger is empty"**. That is correct for a wholly un-migrated vault, but it misses a partially-pinned one.

## The gap

`pin_source_ids` deliberately refuses to pin a `source_key` claimed by more than one live fragment — the already-duplicated-vault case. It pins every clean source, reports the conflict once, and points the operator at `creek clean duplicates`.

After that run the ledger is **non-empty**, so `unpinned_vault_warning` returns `None` forever. But the conflicted source still has no ledger record. If the operator does not act on the one-time console message and simply runs `creek ingest` again, that source takes the `record is None` branch (`pipeline.py`), is written under the newly-derived id, and produces a **third** fragment — silently, with no second warning.

So the population most at risk (vaults that were *already* duplicated before the upgrade) is exactly the one the advisory stops covering after the first migration run.

Found in Gate 2.5 review of #1329 and triaged as a follow-up rather than expanding that PR: closing it properly needs a cheap way to know "some markdown fragment has no ledger record" on every ingest, and the obvious implementation — a full vault scan per run — is too expensive to add casually to the hot path.

## Options

1. **Re-emit unresolved conflicts on every ingest.** Persist the conflict list from the last `pin_source_ids` run (e.g. under `00-Creek-Meta/State/`) and re-print it until the keys resolve. Cheap to check, but adds a second piece of state — which #1329 deliberately avoided by deriving the signal from the ledger itself.
2. **Widen the predicate to "any markdown fragment lacks `source.origin_key`".** Semantically exact and needs no new state, but it is a vault-wide scan on every ingest. Might be acceptable if restricted to the fragments the run actually touches — a unit whose `origin_key` resolves to no ledger record is precisely the at-risk case, and that check is already local to the write loop.
3. **Warn from the write path instead.** In `write_fragment_idempotent`, when a fragment has an `origin_key` but no ledger record *and* the vault is otherwise pinned, that is the signature of an unresolved conflict. Zero extra I/O.

Option 3 looks strongest and should be costed first.

## Acceptance criteria

[...]

Issue #1364, opened 2026-08-11
chore(ingest): decide code.py's LA-vs-UTC timestamp rendering, or record it as won't-fix


Geoffe-Ga:
## Problem

`creek/ingest/base.py:file_modified_time` is the shared identity anchor for file-based ingestors and returns **UTC**. #1329 aligned markdown, documents and images onto it, joining generic, substack, spreadsheets and presentations.

`creek/ingest/code.py:430` still does its own thing:

```python
mtime = path.stat().st_mtime
return datetime.fromtimestamp(mtime, tz=LA_TZ)
```

This is **not a bug**. It is a pure function of the epoch float, so it is host-TZ-independent exactly like the UTC form — it just renders a different offset, and that offset string lands inside `generate_fragment_id`'s hashed input.

It is, however, the last divergence from the stated rule, and `creek/ingest/generic.py:239-244` argues explicitly against the LA rendering ("rendering in LA would bake -07:00/-08:00 into the hash"). #1329 documented the divergence in `file_modified_time`'s docstring rather than closing it.

## Why #1329 deliberately left it alone

Changing `code.py` to UTC re-mints every code fragment's id, and code is **unledgered** — nothing pins those ids. The pin migration added in #1329 (`creek/ingest/pin_ids.py`) is markdown-only and could not rescue them even if widened: source code is the one corpus that changes continuously, so for a live repo most code fragments' stored bodies no longer match what was hashed. The result of a naive alignment would be a duplicate of every stale code fragment on the next `creek ingest --type code`.

## The options

1. **Won't-fix.** Keep LA, keep the documented divergence. Cheapest; the docstring already tells the truth.
2. **Align to UTC behind a pin migration for code fragments**, which needs the unledgered-source identity work (see the sibling follow-up on `derive_source_key` / `ledger_for_source`) to land first.

Either is defensible. This issue exists so the choice is made deliberately rather than by a future reader "tidying up" `code.py` and silently duplicating a vault.

## Acceptance criteria

- [ ] A decision is recorded (in this issue and in `file_modified_time`'s docstring).
- [ ] If aligning: code fragment ids are pinned before the derivation moves, and a test asserts a post-migration `creek ingest --type code` reports `unchanged` for untouched sources.
- [ ] If won't-fix: `file_modified_time`'s docstring and `generic.py:239-244` are reconciled so neither reads as an unmet aspiration.

Refs #1329.

Issue #1363, opened 2026-08-11
fix(ingest): documents are unledgered — blocked on derive_source_key's bare-filename collapse


Geoffe-Ga:
## Problem

`creek/ingest/pipeline.py:ledger_for_source` returns a ledger only for `source_type == "markdown"`. Every document fragment therefore runs unledgered: it never gets a `source.origin_key`, never gets a `LedgerRecord`, and every re-ingest takes the `record is None` branch and reports `created`.

Two consequences:

1. **Documents are not idempotent under an mtime change.** #1329 removed the *timezone*-driven id churn (both ingestors now derive their fallback timestamp through `file_modified_time`, a pure function of the epoch float). What remains is mtime-driven churn: touch the file, and the derived id moves with nothing to pin it. Markdown is protected by the ledger; documents are not.
2. **Documents have no RTBF purge coverage from the CLI path.** `creek/purge/engine.py` keys the sweep on `source.origin_key`, and `attach_origin_key` no-ops without a ledger. (The MCP upload path escapes this by borrowing a ledger via `ledger_source=UPLOAD_LEDGER_SOURCE`.)

This was recorded as accepted risk in #1329 rather than fixed there.

## Why the obvious fix ships data loss — READ THIS BEFORE WIDENING

Widening `ledger_for_source` to `{"markdown", "documents"}` is **not** safe today, and this is verified against the source, not theoretical.

`derive_source_key` (`creek/ingest/pipeline.py:57-70`) resolves the source path relative to the vault root and, on `ValueError` — i.e. for any source **outside** the vault — falls back to `candidate.name`, the **bare filename**. Ingesting documents from outside the vault is the normal case (`creek ingest --type documents --input ~/Documents`).

So given:

```
~/Documents/a/report.pdf
~/Documents/b/report.pdf
```

both resolve to `source_key == "report.pdf"`. The first is ingested and ledgered. The second finds that record, takes the `record.content_hash != new_hash` branch (`pipeline.py:250-256`), executes `fragment.id = record.fragment_id`, and calls `update_fragment` — which **rewrites the first fragment's body with the second document's content**. One fragment where there should be two, and the first document's content is gone.

The same collision is near-certain for images (`2023/IMG_0001.jpg` vs `2024/IMG_0001.jpg`), so this blocks any wider ledger rollout too.

## Blocked on

[...]

PR #1368, opened 2026-08-11
fix(ingest): a fragment's id must not depend on the host that ingested it

### LINKED ISSUE #1329: fix(ingest): markdown/document mtime is timezone-naive, so a fragment's id and created date change with the host TZ
**Severity:** Critical
**Remediation:** fix in place

## Problem

`creek/ingest/markdown.py:279` and `creek/ingest/documents.py:829` derive a
fragment's `created` timestamp from a **timezone-naive** `datetime.fromtimestamp(...)`
— the host's local wall clock — and then hand the result to `normalize_timestamp(ts, None)`,
which documents naive input as UTC. The host's local time is therefore reinterpreted
as UTC and shifted again into LA.

That value lands in `Fragment.created` **and** in `ParsedFragment.timestamp`, which
`generate_fragment_id` hashes. Taxonomy family 0 (Correctness — timezone-naive
datetime), with family 12 (confident wrongness).

**Current state:** the same file with the same mtime yields a different `created`
**and a different `frag-<sha>` id** on every host/timezone. Because
`write_fragment_idempotent` keys on the id, re-ingesting the same export on a
different host (or after a `TZ` change, or across DST) mints a *new* fragment
rather than recognising the existing one — the vault silently gains duplicates.
On an LA host (the product's own anchor zone) a 2 a.m. note is filed on the
previous day.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** One file, one fixed mtime
(`1710493200.0` = `2024-03-15T09:00:00Z`), four hosts:

```
$ for tz in UTC America/Los_Angeles Europe/Berlin Australia/Sydney; do
    TZ=$tz python -c "...; ts=_get_file_creation_timestamp(p); \
      print(ts.isoformat(), generate_fragment_id('/fixed/note.md', ts, 'body'))"; done

  TZ=UTC                  -> 2024-03-15T02:00:00-07:00  frag-600a7bc1bb2b
  TZ=America/Los_Angeles  -> 2024-03-14T19:00:00-07:00  frag-43dea0409221
  TZ=Europe/Berlin        -> 2024-03-15T03:00:00-07:00  frag-ecb207d01592
  TZ=Australia/Sydney     -> 2024-03-15T13:00:00-07:00  frag-00105f75338b
```

Four timestamps, four ids, one file.

**Signal 2 — structural + intent.** These are the only two `fromtimestamp` calls
in the package without a `tz=` argument (`rg 'fromtimestamp' creek/ creek_mcp/`
→ 11 hits, 9 pass `tz=UTC` / `tz=LA_TZ`). The centralised helper
`creek/ingest/base.py:454 file_modified_time()` exists, per its own docstring,
"so file-based ingestors share a single conversion rule and do not drift apart
over time" — `generic.py`, `spreadsheets.py` and `presentations.py` use it;
`code.py:430` uses `tz=LA_TZ`; these two use neither.
`creek/ingest/generic.py:236-243` states the contract they violate verbatim:
*"identity does not vary with the host's tzdata, `TZ` env var, or DST state."*

## Scope

Fixes the two naive `fromtimestamp` call sites and routes them through the
existing shared helper. Does **not** re-mint ids for fragments already in a
vault (that is a migration question), and does not touch the frontmatter-date
path (`_resolve_source_timestamp`), which is already correct.

## Tasks

1. In `creek/ingest/markdown.py:264-280` (`_get_file_creation_timestamp`), pass an
   explicit timezone to `datetime.fromtimestamp` — or delegate to
   `creek.ingest.base.file_modified_time` — so the value is host-independent.
2. Do the same at `creek/ingest/documents.py:829`.
3. Add a regression test that asserts `created` and `fragment_id` are identical
   under at least three `TZ` values for one fixed mtime (subprocess with a
   patched `TZ`, or `time.tzset()` around the call).
4. Note the id-stability contract in the two functions' docstrings, matching
   `generic.py:236-243`.

## Acceptance Criteria

- [ ] The repro above yields one identical `created` and one identical
      `frag-<sha>` for all four timezones.
- [ ] A test pins host-TZ independence for both ingestors and fails if the
      naive call is reintroduced.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/ingest/markdown.py` | Modify |
| `creek-tools/creek/ingest/documents.py` | Modify |
| `creek-tools/tests/test_ingest_markdown.py` (or nearest existing) | Modify |

Geoffe-Ga:
## Summary

A fragment's id was a function of the machine that ingested it.

`generate_fragment_id` hashes the fragment's timestamp — `f"{source}:{timestamp.isoformat()}:{content}"` — and for a file with no embedded date that timestamp came from the filesystem. Two host-dependent inputs leaked in:

1. **The host `TZ` env var.** `datetime.fromtimestamp(mtime)` with no `tz=` renders the epoch in the host's *local* zone, and `normalize_timestamp` then re-read that naive wall-clock as UTC before shifting to LA — a double shift. Reproduced digit-for-digit at HEAD, one file, mtime pinned to `1710493200.0`:

   | `TZ` | markdown | documents |
   |---|---|---|
   | `UTC` | `frag-c8bfc3e67c4d` | `frag-dd49a01f5539` |
   | `America/Los_Angeles` | `frag-aa53901eaffa` | `frag-7286deeb5d27` |
   | `Europe/Berlin` | `frag-fe2989a00f9b` | `frag-430e7de61929` |
   | `Australia/Sydney` | `frag-5ebe48b2ce65` | `frag-936750c2ef1f` |

2. **The host operating system — an axis the issue never mentions.** `markdown.py` read `getattr(stat, "st_birthtime", stat.st_mtime)`. `st_birthtime` exists on macOS/BSD and *not* on Linux, where the `getattr` silently falls back to mtime. So the same file already minted one id on a developer's Mac and a different one in Linux CI, independently of `TZ`. **A fix that only passed `tz=` would have left the id host-dependent.** Dropping birthtime is the deliberate trade; see "Behaviour changes" below.

Both fallbacks now go through `creek.ingest.base.file_modified_time` — `datetime.fromtimestamp(st_mtime, tz=UTC)`, a pure function of the epoch float. All four rows collapse to one: `frag-b145fde4406d` (markdown) and `frag-e71aebc3ad98` (documents), `created` `2024-03-15T09:00:00+00:00`.

---

## ⚠️ Read this first: this PR PINS ids, it does not RE-MINT them

**The issue's acceptance criteria mandate a re-minting migration** — an old→new id map, in-place file renames, and a lockstep remap of the per-directory id index, the ledger, `parent_id`/`child_ids` and every `generate_child_fragment_id` descendant, resonance edges, the embedding cache, thread/eddy membership and provenance, plus an RTBF alias, crash-resumability, and a duplicate-merge rule.

**This PR inverts that.** Existing ids never move, which makes that entire apparatus **vacuous rather than skipped**. There is nothing to remap because nothing moves.

The justification is in-tree, not invented here. `write_fragment_idempotent` already overrode the derived id with `record.fragment_id` on the changed branch and the tombed branch: **ledger-backed identity is the designed invariant for mutable file sources** (#672 SPEC R1), and the vault already tolerates `id != derived(id)` for every edited journal entry. The derivation is only how a *new* unit gets its *first* id. So the correct response to a derivation change is to make sure the ledger knows about the fragments that predate it — not to chase the derivation across eight referencing surfaces.

A reviewer holding the AC as a checklist will read roughly eight criteria as unmet. They are listed explicitly at the bottom of this description rather than glossed.

### Proof that no existing id moves

Built a 5-fragment vault using the **pre-fix derivation** (called directly, so the worktree was never edited), then migrated, then re-ingested under a *different* host timezone:

```
seeded 5 fragments with PRE-FIX ids
new derivation would mint 5/5 ids NOT present in the vault -> 5 orphans without the migration
pin: pinned=5 conflicts=0 unpinnable=0 reproduced=5/5
reingest: created=0 updated=0 unchanged=5 tombed=0 errors=[]
files before=5 after=5
IDS IDENTICAL: True
FILENAMES IDENTICAL: True
MOVED/ORPHANED: none
```

---

## What changed

**The derivation** — `creek/ingest/markdown.py`, `creek/ingest/documents.py`, `creek/ingest/images.py`, `creek/ingest/base.py`. Both `_resolve_timestamp` fallbacks delegate to `file_modified_time`; the birthtime-reading `_get_file_creation_timestamp` is deleted outright (only two references tree-wide, both removed); `images._modified_time` folds into the same helper. `file_modified_time`'s docstring stops claiming a "single conversion rule" it did not have, and names both failure modes so the next reader cannot reintroduce them by simplifying.

**Identity is the ledger's job** — `creek/ingest/pipeline.py`. `fragment.id = record.fragment_id` is **hoisted above every branch** of `write_fragment_idempotent`. It previously ran only on the changed and tombed paths, leaving the *unchanged* path to write whatever id had just been derived — which is the precise mechanism that turned a derivation drift into a silent duplicate, and which `creek_mcp/tools/upload.py` already documented as a live hazard worked around by hand. That comment is now corrected rather than left asserting a closed hazard. This is a forcing function: it is tested independently of timezones, by perturbing the derivation directly.

**`TOMBING_SOURCES`** — `creek/ingest/pipeline.py`. Unfuses "has a ledger" from "may tomb". `resolve_ledger` lets a caller borrow a ledger purely to pin identity (that is how an uploaded document earns its `origin_key` and therefore its RTBF coverage), and that opt-in also armed `tomb_missing_units` for any directory input. Two lines, no behaviour change today, footgun removed before anyone widens.

**The migration** — new `creek/ingest/pin_ids.py` + `creek ingest --pin-source-ids [--dry-run]`. Back-fills the ledger with each fragment's **existing** id. Notable decisions:
- **Conflict guard first.** A `source_key` claimed by 2+ live fragments is pinned for *neither*, and both paths are printed. This is the already-duplicated vault; blessing one claimant would orphan the other forever, and the ledger holds one record per key so pinning both is not expressible. This is the primary safety property.
- **Hashes the fragment's STORED BODY, not a re-read of the current source.** If the source was edited after its original ingest, hashing the current source would file the new hash against the old stored body; the next ingest would read `unchanged` and **the edit would be swallowed permanently**, because nothing revisits an unchanged unit. Covered by a dedicated test that edits the source *before* the migration runs.
- **Stamps `source.origin_key` atomically.** Mandatory, not cosmetic: the RTBF purge sweep resolves its target from on-disk frontmatter and skips fragments without the key, while `update_fragment` preserves on-disk frontmatter without merging new `source` fields — so a fragment lacking it at migration time would never gain one, ever. The write uses the vault's atomic helper rather than `refresh._rewrite_frontmatter`'s plain `write_text`, which would truncate a live fragment on a mid-run kill.
- **Reproduction rate is advisory, never a gate.** A migration gated on reproduction matches a fraction of the vault and therefore does nothing for the rest — which is exactly the population that gets orphaned.

**Un-migrated vaults are detected**, from the ledger's own emptiness rather than a second version marker, and warned about *before* the run that would duplicate them.

**An AST guard** — new `tests/test_timestamp_derivation_guard.py`. No naive `fromtimestamp` and no `st_birthtime` anywhere in `creek/` or `creek_mcp/`. A third proposed rule (no `.st_mtime` read outside `file_modified_time`) was **rejected and the rejection recorded**: it is red on `gdrive.py`'s legitimate local-vs-Drive staleness comparison, and a guard that fails on correct code teaches people to disable guards.

## Deliberately NOT done

**`ledger_for_source` is not widened to documents**, against the AC's preference — because it ships data loss. `derive_source_key` returns the **bare filename** for any source outside the vault, which is the normal case for `creek ingest --type documents --input ~/Documents`. So `~/Documents/a/report.pdf` and `~/Documents/b/report.pdf` share the key `report.pdf`; the second hits the `content_hash != new_hash` branch, adopts the first's `fragment_id`, and `update_fragment` **overwrites the first fragment's body**. Documents get the derivation fix (which alone kills their TZ duplicate) plus follow-up #1363, which the AC explicitly permits.

**`code.py:430` keeps its LA rendering.** Already host-independent; re-minting would orphan every code fragment, and code is unledgered with continuously-changing sources so no reproduction-based migration could recover them. Follow-up #1364.

## Behaviour changes worth knowing about

1. **On macOS a markdown fragment's `created` now reports modification time, not birth time.** That is the price of an id that does not depend on the ingesting OS. Because the writer builds a filename from `created`, a **newly ingested** fragment's date prefix may shift. **No existing file is renamed** — the migration never touches `created`. Authorship remains `authored_at`'s job (FEAT-031).
2. **Documents remain unledgered**, so an mtime move still re-derives their id with nothing to pin it (#1363).

## Test plan

**Gate 1 — RED first, on behavioural assertions.** All six new tests failed before the fix, each on the behaviour rather than a symbol:
- id sets had four members (the tables above); birthtime shape vs Linux shape gave `frag-aa53901eaffa` vs `frag-1f63caef932d`; every e2e case failed `assert 2 == 1` on the duplicate file.
- The RED uses a **relative source path** via `monkeypatch.chdir` — load-bearing. `assemble_ingested_fragment` hashes `parsed.source_path`, so an absolute `tmp_path` discovery would put the per-run temp dir inside the hash and make the pinned literals unreproducible between runs.
- **No test uses `--incremental`.** `should_skip_unit` skips an unchanged ledgered unit before the write, so an incremental-mode test is green while the bug is live.

**Mutation battery — 11 reverts, zero survivors.** Each guard was reverted alone and the suite re-run: naive markdown derivation (4 fail), naive documents derivation (3), un-hoisting `fragment.id` (2), removing the `TOMBING_SOURCES` gate (1), dropping the unpinned-vault warning (1), dropping the `origin_key` stamp (2), hashing the current source instead of the stored body (2), conflict guard picking a claimant (2), `dry_run` writing anyway (2), removing the idempotency check (1), removing the torn-pin repair (1). Each AST guard rule was also proved to bite against a scratch package containing the exact forbidden syntax.

The battery found a real gap: the first version of the edited-source test edited *after* pinning and did **not** discriminate the stored-body correction — and a trailing newline was masking the mutation. Added `test_an_edit_made_before_the_migration_is_not_swallowed`, which edits *before* the migration with an exact-round-tripping body; it now fails with `updated == 0` under the mutation, i.e. the edit is silently swallowed exactly as predicted.

**Gate 2.5 — self-review found a real blocker, reproduced and fixed.** A `code-review-orchestrator` pass over the diff returned FIX_REQUIRED on one issue, which I then reproduced by execution before fixing:

> `_pin_one` wrote the ledger record before stamping `source.origin_key`. If the second write failed — a disk-full or permission `OSError` partway through an N-fragment pass, far likelier than a `kill -9` — the record's presence made the *retry* report the fragment as `already_pinned` and never stamp it. The fragment would be RTBF-invisible forever, silently.

Reproduced by injecting an `OSError` on the second of three fragments: after the retry, `2024-03-14-n1.md` was still unstamped.

I did **not** take the suggested fix (swap the two writes). Swapping trades an unrecoverable loss for a different unrecoverable loss: with the stamp first, a crash leaves the fragment RTBF-visible but *identity-unpinned*, and a re-ingest in that window mints a duplicate — which cannot be un-minted, whereas a missing stamp can be restored. Instead the ledger record stays first (it is the write that protects identity) and the already-ledgered path is made **self-healing**: a fragment with a record but no stamp gets the stamp on the next run, reported as `repaired`. Both failure modes now converge. Covered by `test_an_interrupted_run_is_repaired_by_the_next_one`, which fails with `assert 0 == 1` if the repair is removed.

Two further findings addressed: `--dry-run` combined with `--refresh-dates`/`--refresh-ai-chat` now raises `BadParameter` instead of silently running that migration for real; and the "nothing changes but one key" claim is now a whole-frontmatter equality assertion rather than four spot checks. One finding — a partially-pinned vault stops emitting the migration advisory, so an *unresolved conflict* can still mint a third duplicate — is triaged to **#1367** rather than expanded into this PR, because closing it properly needs a cheap per-ingest signal and the obvious implementation is a full vault scan on the hot path.

I also fixed an unreported usability defect found by eyeballing the output: rich word-wrapped long absolute paths mid-token, making the operator's copy-pasteable action items unusable. Those lines now print with `soft_wrap=True`.

**Gate 2 — `./scripts/check-all.sh` exit code 0.** 12 checks passed, 0 failed. Aggregate branch coverage 94.40%; per-file gate clean (257 files ≥ 80%, including the new `pin_ids.py`); mypy strict clean; xenon clean; ruff clean. Integration lane 356 passed, e2e lane 14 passed. Full unit suite 9238 passed with no pre-existing test modified or deleted.

**No bypasses.** No `# noqa`, `# type: ignore`, `# pylint: disable`, `@pytest.mark.skip`, no threshold change, no deleted test. Three `# type: ignore`s that crept into draft tests were removed by fixing the types properly.

**Test literals:** the dossier's claim that all 60 `frag-` literals in `tests/` are synthetic held — no existing literal needed re-baselining, and none failed.

## Acceptance criteria the inversion leaves formally unmet

Listed explicitly, not glossed. Each is unmet **because ids do not move**, which makes the requirement vacuous:

- *"A MIGRATION EXISTS ... recomputes each affected fragment's id from the corrected derivation and UPDATES IN PLACE"* — a migration exists and is CLI-reachable, but it records the existing id rather than recomputing one. Nothing is renamed and `created` is never rewritten.
- *"remaps every id-referencing artefact in lockstep ... An old→new id map is emitted"* — no artefact is remapped and no map is emitted, because no id changes. The id index, children/`generate_child_fragment_id` descendants, resonance edges, the embedding cache, thread/eddy membership and provenance are all untouched **by construction**.
- *"RTBF/purge coverage survives ... a purge issued against a pre-migration id still resolves — via the emitted id map or a recorded alias"* — no alias exists; a pre-migration id still resolves because it is still the id. `origin_key` is additionally *stamped*, which closes a purge gap the AC did not identify.
- *"crash-safe or resumable"* — the ledger is append-only JSONL and the only fragment write is a single atomic key stamp, so there is no torn state to resume from. Resumability is a property here, not a feature.
- *"DETECTS AND REPAIRS an already-duplicated vault ... Define and implement the merge/keep rule"* — detected and reported, **not repaired**. Deliberate: the ledger holds one record per key, so any automatic keep-rule permanently orphans the loser. The operator is routed to `creek clean duplicates`.
- *"Every existing test that transitively pins a `frag-<sha>` derived from a file mtime is re-baselined"* — none needed it; all 60 literals are synthetic.
- *"Documents' unledgered status is either fixed or explicitly recorded as accepted risk with a follow-up issue"* — met via the accepted-risk arm (#1363), not the fix arm.
- *"code.py alignment"* — deferred to #1364 with the reason stated.

Closes #1329
Refs #1363, #1364, #1367, #672

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
     from creek.generate.mining import MiningRunReport
     from creek.ingest.base import Ingestor
     from creek.ingest.discord_dispatch import DiscordMode
+    from creek.ingest.pin_ids import PinResult
     from creek.link.link_engine import LinkSummary
     from creek.models import CompileTargetKind, Frequency, Mode, Phase
     from creek.purge import PurgeEngine, PurgeResult
@@ -1233,6 +1234,26 @@ def ingest(
         "-y",
         help="Skip the consent prompt for first-time sources (logged).",
     ),
+    pin_source_ids: bool = typer.Option(
+        False,
+        "--pin-source-ids",
+        help=(
+            "One-time migration: back-fill the ingest ledger with each existing "
+            "fragment's existing id, so no id moves under the #1329 "
+            "id-derivation fix. Operators MUST run this once after upgrading — "
+            "without it the next ingest mints a new id for every markdown "
+            "fragment already on disk and leaves the original behind as a "
+            "duplicate. Idempotent. --type / --input are ignored."
+        ),
+    ),
+    dry_run: bool = typer.Option(
+        False,
+        "--dry-run",
+        help=(
+            "With --pin-source-ids: print the plan and write nothing "
+            "(no ledger records, no frontmatter changes)."
+        ),
+    ),
     refresh_dates: bool = typer.Option(
         False,
         "--refresh-dates",
@@ -1285,18 +1306,25 @@ def ingest(
     input is idempotent: deterministic fragment IDs ensure existing
     files are recognised and skipped.
 
+    ``--pin-source-ids`` is a one-shot migration: it walks ``--vault``
+    (or the configured vault) and records each existing fragment's
+    **existing** id against its source in the ingest ledger, so the
+    #1329 id-derivation fix can never move an id already on disk.
+    ``--dry-run`` plans it without writing.
+
     ``--refresh-dates`` is a one-shot backfill: it walks
     ``--vault`` (or the configured vault), re-runs the per-format
     authored-date extraction chain against each fragment's source
     file, and rewrites the frontmatter in place. ``--type`` /
     ``--input`` are ignored in refresh mode.
     """
-    if refresh_dates:
-        _run_refresh_dates(vault)
-        return
-
-    if refresh_ai_chat:
-        _run_refresh_ai_chat(vault)
+    if _dispatch_ingest_migration(
+        vault,
+        pin_source_ids=pin_source_ids,
+        dry_run=dry_run,
+        refresh_dates=refresh_dates,
+        refresh_ai_chat=refresh_ai_chat,
+    ):
         return
 
     if type is None or input is None:
@@ -1343,6 +1371,143 @@ def ingest(
         source_type=type,
         strict=strict,
     )
+
+
+def _dispatch_ingest_migration(
+    vault: Path | None,
+    *,
+    pin_source_ids: bool,
+    dry_run: bool,
+    refresh_dates: bool,
+    refresh_ai_chat: bool,
+) -> bool:
+    """Run whichever one-shot migration ``creek ingest`` was asked for.
+
+    ``--pin-source-ids`` is checked before every other mode, including the two
+    refresh migrations: a vault that still needs pinning must not fall through
+    into a normal ingest, because that run is precisely the one that derives
+    fresh ids and duplicates it (#1329).
+
+    Args:
+        vault: ``--vault``, or ``None`` to use the configured vault.
+        pin_source_ids: Whether ``--pin-source-ids`` was given.
+        dry_run: Whether ``--dry-run`` was given.
+        refresh_dates: Whether ``--refresh-dates`` was given.
+        refresh_ai_chat: Whether ``--refresh-ai-chat`` was given.
+
+    Returns:
+        ``True`` when a migration ran and the caller should return
+        immediately; ``False`` to proceed with a normal ingest.
+
+    Raises:
+        typer.BadParameter: If ``--dry-run`` is combined with a migration that
+            does not support it.
+    """
+    if pin_source_ids:
+        _run_pin_source_ids(vault, dry_run=dry_run)
+        return True
+    if dry_run:
+        # Refusing beats ignoring. --dry-run is new to `creek ingest` and
+        # scoped to --pin-source-ids; silently running a different migration
+        # for real while the operator believes they asked for a preview is the
+        # worst possible reading of the flag.
+        msg = "--dry-run is only supported with --pin-source-ids."
+        raise typer.BadParameter(msg)
+    if refresh_dates:
+        _run_refresh_dates(vault)
+        return True
+    if refresh_ai_chat:
+        _run_refresh_ai_chat(vault)
+        return True
+    return False
+
+
+def _print_pin_findings(label: str, lines: list[str]) -> None:
+    """Print every finding under *label*, one per line, or nothing if there are none.
+
+    Findings are listed in full rather than counted. Each line names a
+    fragment the migration deliberately left unpinned, and an unpinned
+    fragment is one the next ingest can still orphan — a bare count would tell
+    the operator that a problem exists without telling them which file to fix.
+    Each line is markup-escaped because it carries a filesystem path, and a
+    path containing ``[`` would otherwise be swallowed as Rich markup or raise
+    ``MarkupError`` over the top of the message (#1310).
+
+    ``soft_wrap=True`` is load-bearing for the same reason: Rich's default
+    word-wrap breaks a long absolute path across lines mid-token, which is
+    exactly the string the operator needs to copy-paste into their next
+    command. Letting the terminal wrap it keeps the path intact.
+
+    Args:
+        label: Heading for this class of finding, e.g. ``"Conflicts"``.
+        lines: Human-readable finding lines from :class:`PinResult`.
+    """
+    if not lines:
+        return
+    console.print(f"[yellow]{label}: {len(lines)}[/yellow]")
+    for line in lines:
+        console.print(f"  [dim]{escape(line)}[/dim]", soft_wrap=True)
+
+
+def _render_pin_summary(result: PinResult, *, dry_run: bool) -> None:
+    """Render the one-screen summary of a ``--pin-source-ids`` run.
+
+    Args:
+        result: What the migration did, or would have done under *dry_run*.
+        dry_run: Whether this was a plan; when ``True`` nothing was written.
+    """
+    verb = "Would pin" if dry_run else "Pinned"
+    console.print(
+        f"[bold green]{verb} {result.pinned} existing fragment id(s) into the "
+        "ingest ledger.[/bold green]"
+    )
+    console.print(f"  already pinned: {result.already_pinned}")
+    console.print(f"  examined: {result.examined}")
+    if result.repaired:
+        # Surfaced rather than folded into "already pinned": a non-zero count
+        # is evidence that an earlier run did not finish, which the operator
+        # would otherwise have no way to learn.
+        console.print(
+            f"[yellow]  repaired {result.repaired} half-pinned fragment(s) left "
+            "by an interrupted earlier run.[/yellow]"
+        )
+    # Advisory only, and labelled as such on screen: a fragment that does not
+    # re-derive its own id is not a failure — its id was minted by the very
+    # build this migration exists to correct. Nothing is gated on this number,
+    # so nothing about the run should read as if a low one meant trouble.
+    console.print(
+        f"  re-derive their own id: {result.reproduced} "
+        "(advisory diagnostic only; nothing is gated on it)"
+    )
+    if dry_run:
+        console.print("[yellow]Dry run: nothing was written.[/yellow]")
+    _print_pin_findings("Conflicts (pinned none of these)", result.conflicts)
+    _print_pin_findings("Unpinnable", result.unpinnable)
+
+
+def _run_pin_source_ids(vault: Path | None, *, dry_run: bool) -> None:
+    """Execute the #1329 ``--pin-source-ids`` ledger back-fill.
+
+    Resolves the vault path (CLI flag → configured default), invokes
+    :func:`creek.ingest.pin_ids.pin_source_ids`, and renders a one-screen
+    summary. Exit codes mirror the rest of the CLI: 0 on success (regardless
+    of how many fragments were pinned, conflicted or were unpinnable), 1 if
+    the vault path is unusable.
+
+    Args:
+        vault: ``--vault``, or ``None`` to use the configured vault.
+        dry_run: ``--dry-run``; when ``True`` the plan is printed and neither
+            the ledger nor any frontmatter is written.
+    """
+    from creek.ingest.pin_ids import pin_source_ids
+
+    config = _load_config_for_vault(vault)
+    vault_path = vault or config.vault_path
+    if not vault_path.exists():
+        console.print(f"[red]Vault path does not exist: {vault_path}[/red]")
+        raise typer.Exit(code=1)
+
+    _render_pin_summary(pin_source_ids(vault_path, dry_run=dry_run), dry_run=dry_run)
 
 
 def _run_refresh_dates(vault: Path | None) -> None:
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -379,6 +379,31 @@ def _parse_since_arg(text: str) -> datetime:
     return parsed if parsed.tzinfo is not None else parsed.replace(tzinfo=UTC)
 
 
+def _print_ingest_warning(message: str) -> None:
+    """Show one ingest advisory to the operator, as it is detected (#1329).
+
+    Warnings go to the same stdout console as every other ``creek ingest``
+    advisory (``_warn_if_discovered_but_empty``, ``_print_pin_findings``)
+    rather than to stderr: this command's output is human prose, not data on a
+    pipe, and an advisory routed to a stream none of its neighbours use is one
+    an operator can lose.
+
+    The exit code is deliberately left alone. A warning reports vault state
+    that will cause trouble later, not a failure of this run — and
+    ``creek sync`` shares this path, so escalating would turn every scheduled
+    pass over an un-migrated vault into a hard failure.
+
+    ``escape`` and ``soft_wrap=True`` are load-bearing for the same reasons as
+    in :func:`_print_pin_findings`: the text carries a shell command the
+    operator is meant to copy, and rich would otherwise be free to wrap it
+    across lines (or read a bracketed path as markup).
+
+    Args:
+        message: The advisory text from the ingest pipeline.
+    """
+    console.print(f"[yellow]{escape(message)}[/yellow]", soft_wrap=True)
+
+
 def _run_ingest(
     *,
     ingestor_cls: type[Ingestor],
@@ -413,6 +438,12 @@ def _run_ingest(
         list of human-readable strings prefixed with ``[<source_type>]`` and
         ``discovered`` is how many inputs the ingestor's ``discover()`` found.
 
+        Operator advisories are **not** in this tuple. They are printed by
+        :func:`_print_ingest_warning` the moment the pipeline detects them, so
+        every caller of this helper surfaces them without having to remember
+        to — which is what went wrong when the un-pinned-vault advisory was
+        left to a return value nobody read (#1329).
+
     Raises:
         typer.Exit: With code ``1`` when the vault cannot be opened.
     """
@@ -427,6 +458,7 @@ def _run_ingest(
             reclassify_threshold=reclassify_threshold,
             since=since,
             incremental=incremental,
+            on_warning=_print_ingest_warning,
         )
     except FileNotFoundError as exc:
         console.print(f"[red]Vault unavailable: {exc}[/red]")
--- a/creek-tools/creek/ingest/base.py
+++ b/creek-tools/creek/ingest/base.py
@@ -454,8 +454,43 @@ def assemble_ingested_fragment(parsed: ParsedFragment) -> IngestedFragment:
 def file_modified_time(path: Path) -> datetime:
     """Return *path*'s modification time as a timezone-aware UTC datetime.
 
-    Centralised so file-based ingestors share a single conversion
-    rule and do not drift apart over time.
+    This is the **identity anchor** for every file-based ingestor that has
+    no embedded date to fall back on. :func:`generate_fragment_id` hashes
+    the timestamp it returns, so the conversion must be a *pure function of
+    the epoch float*: invariant under the host's ``TZ`` environment
+    variable, its installed tzdata, its DST state, and its operating
+    system. ``datetime.fromtimestamp(st_mtime, tz=UTC)`` is exactly that.
+
+    Two ways of getting it wrong were live until #1329, and both are worth
+    naming because both look like simplifications:
+
+    * ``datetime.fromtimestamp(st_mtime)`` with no ``tz=`` renders the epoch
+      in the *host's* local zone. One file then mints a different
+      ``frag-<sha>`` in every timezone it is ingested from.
+    * ``getattr(stat, "st_birthtime", stat.st_mtime)`` reads a field that
+      exists on macOS/BSD and not on Linux. One file then mints a different
+      id on a developer's laptop than in CI. Birth time is *not* available
+      here for that reason; mtime is the only field every supported
+      platform agrees on.
+
+    Rendering in a fixed non-UTC zone (say LA) would also be
+    host-independent, but it bakes a ``-07:00``/``-08:00`` offset that
+    flips with DST into the hashed input. UTC is the rule for markdown,
+    documents, generic, substack, spreadsheets, presentations and images.
+    :mod:`creek.ingest.code` renders LA instead — a deliberate, equally
+    host-independent divergence left alone by #1329 because re-minting it
+    would orphan every code fragment, and code sources change too
+    continuously for any reproduction-based migration to recover them. That
+    last divergence is tracked by #1364.
+
+    Note that this value answers "which file is this?", not "when was this
+    written?". Authorship is ``authored_at``'s job (FEAT-031).
+
+    Args:
+        path: The file to stat.
+
+    Returns:
+        The file's mtime as a timezone-aware UTC datetime.
     """
     return datetime.fromtimestamp(path.stat().st_mtime, tz=UTC)
 
--- a/creek-tools/creek/ingest/documents.py
+++ b/creek-tools/creek/ingest/documents.py
@@ -36,6 +36,7 @@ from creek.ingest.base import (
     Ingestor,
     ParsedFragment,
     RawDocument,
+    file_modified_time,
     normalize_encoding,
     normalize_timestamp,
     parse_authored_at,
@@ -810,12 +811,31 @@ class DocumentIngestor(Ingestor):
         Checks metadata for created_date first, then falls back to
         the file's modification time.
 
+        **This value is an identity anchor, not an authorship claim.**
+        :func:`creek.ingest.base.generate_fragment_id` hashes it, so the
+        fallback must be a pure function of the file's epoch mtime —
+        invariant under the host's ``TZ`` env var, its tzdata, its DST
+        state, and its operating system (#1329). That is exactly what
+        :func:`~creek.ingest.base.file_modified_time` guarantees, and it is
+        why this must never be "simplified" back to a bare
+        ``datetime.fromtimestamp(mtime)``, which renders the epoch in the
+        host's local zone and so mints one id per timezone.
+
+        Documents are more exposed to this than markdown, not less:
+        :func:`creek.ingest.pipeline.ledger_for_source` leaves them
+        unledgered, so there is no recorded id to fall back on and a
+        drifting derivation is an unconditional second write rather than a
+        merely possible one. Widening the ledger to documents is blocked on
+        a bare-filename collision in ``derive_source_key`` and is tracked by
+        #1363; this fix removes the timezone-driven churn, not the
+        mtime-driven churn.
+
         Args:
             metadata: Fragment metadata dict.
             file_path: Path to the source file.
 
         Returns:
-            A timezone-aware datetime.
+            A timezone-aware datetime; UTC when it comes from the filesystem.
         """
         created = metadata.get("created_date")
         if created is not None:
@@ -824,10 +844,7 @@ class DocumentIngestor(Ingestor):
             except ValueError:
                 logger.warning("Invalid metadata timestamp: %s", created)
 
-        # Fall back to file modification time
-        mtime = file_path.stat().st_mtime
-        ts_string = datetime.fromtimestamp(mtime).isoformat()
-        return normalize_timestamp(ts_string, None)
+        return file_modified_time(file_path)
 
     def convert_to_markdown(self, fragment: ParsedFragment) -> str:
         """Return the fragment content as markdown.
--- a/creek-tools/creek/ingest/images.py
+++ b/creek-tools/creek/ingest/images.py
@@ -30,7 +30,7 @@ from __future__ import annotations
 import logging
 import shutil
 from dataclasses import dataclass
-from datetime import UTC, datetime
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Protocol, runtime_checkable
 
@@ -38,6 +38,7 @@ from creek.ingest.base import (
     Ingestor,
     ParsedFragment,
     RawDocument,
+    file_modified_time,
     parse_authored_at,
 )
 from creek.models import SourcePlatform
@@ -603,8 +604,13 @@ class ImageIngestor(Ingestor):
 
 
 def _modified_time(path: Path) -> datetime:
-    """Return the file's mtime as a timezone-aware UTC datetime."""
-    return datetime.fromtimestamp(path.stat().st_mtime, tz=UTC)
+    """Return the file's mtime as a timezone-aware UTC datetime.
+
+    Delegates to :func:`creek.ingest.base.file_modified_time` so the
+    id-anchoring conversion rule lives in exactly one place (#1329); the
+    semantics are byte-identical to the expression this replaced.
+    """
+    return file_modified_time(path)
 
 
 # EXIF tag IDs — magic numbers chosen by the standard, not by us. Hard-coded
--- a/creek-tools/creek/ingest/markdown.py
+++ b/creek-tools/creek/ingest/markdown.py
@@ -16,9 +16,8 @@ from __future__ import annotations
 
 import logging
 import re
-from datetime import datetime
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import frontmatter
 
@@ -26,11 +25,15 @@ from creek.ingest.base import (
     Ingestor,
     ParsedFragment,
     RawDocument,
+    file_modified_time,
     normalize_encoding,
     normalize_timestamp,
     parse_authored_at,
 )
 from creek.models import SourcePlatform
+
+if TYPE_CHECKING:
+    from datetime import datetime
 
 logger = logging.getLogger(__name__)
 
@@ -261,25 +264,6 @@ def _extract_authored_at_from_frontmatter(fm_data: dict[str, Any]) -> datetime |
     return None
 
 
-def _get_file_creation_timestamp(path: Path) -> datetime:
-    """Get the file creation timestamp from filesystem metadata.
-
-    Falls back to modification time if creation time is not available.
-    Returns a timezone-aware datetime in America/Los_Angeles.
-
-    Args:
-        path: The file path to inspect.
-
-    Returns:
-        A timezone-aware datetime from the file's metadata.
-    """
-    stat = path.stat()
-    # Use birth time on macOS, fall back to mtime
-    ctime = getattr(stat, "st_birthtime", stat.st_mtime)
-    ts_string = datetime.fromtimestamp(ctime).isoformat()
-    return normalize_timestamp(ts_string, None)
-
-
 # ---- MarkdownIngestor ----
 
 
@@ -420,14 +404,28 @@ class MarkdownIngestor(Ingestor):
         """Resolve a timestamp from frontmatter or filesystem metadata.
 
         Checks frontmatter fields first (date, created, created_at),
-        then falls back to the file's creation/modification time.
+        then falls back to the file's modification time.
+
+        **This value is an identity anchor, not an authorship claim.**
+        :func:`creek.ingest.base.generate_fragment_id` hashes it, so the
+        fallback must be a pure function of the file's epoch mtime —
+        invariant under the host's ``TZ`` env var, its tzdata, its DST
+        state, and its operating system (#1329). That is exactly what
+        :func:`~creek.ingest.base.file_modified_time` guarantees, and it is
+        why this must never be "simplified" back to a bare
+        ``datetime.fromtimestamp(mtime)`` (host-local, so one file mints one
+        id per timezone) or to ``st_birthtime`` (present on macOS/BSD,
+        absent on Linux, so one file mints one id per operating system).
+
+        Authorship is ``authored_at``'s job (FEAT-031), which has its own
+        frontmatter fields and its own backfill.
 
         Args:
             fm_data: Parsed frontmatter dictionary.
             file_path: Path to the source file.
 
         Returns:
-            A timezone-aware datetime.
+            A timezone-aware datetime; UTC when it comes from the filesystem.
         """
         fm_ts = _extract_timestamp_from_frontmatter(fm_data)
         if fm_ts is not None:
@@ -436,7 +434,7 @@ class MarkdownIngestor(Ingestor):
             except ValueError:
                 logger.warning("Invalid frontmatter timestamp: %s", fm_ts)
 
-        return _get_file_creation_timestamp(file_path)
+        return file_modified_time(file_path)
 
     def convert_to_markdown(self, fragment: ParsedFragment) -> str:
         """Return the fragment content as-is (already markdown).
--- a/creek-tools/creek/ingest/pin_ids.py
+++ b/creek-tools/creek/ingest/pin_ids.py
@@ -1,0 +1,523 @@
+"""Pin existing markdown fragment ids to their source units (#1329 migration).
+
+``MarkdownIngestor`` and ``DocumentIngestor`` used to derive a fragment's
+timestamp from a timezone-**naive** ``datetime.fromtimestamp(mtime)``.
+:func:`~creek.ingest.base.generate_fragment_id` hashes that timestamp, so the
+*same* file minted a *different* ``frag-…`` id depending on the host ``TZ``
+environment variable. The fix replaces both fallbacks with
+:func:`~creek.ingest.base.file_modified_time`, which is a pure function of the
+epoch float and therefore host-independent — but it also **changes id
+derivation**. Left alone, the next ``creek ingest`` over an existing vault
+would fail to recognise every markdown fragment already on disk, mint a new id
+for it, and leave the original behind as an orphaned duplicate.
+
+The answer here is deliberately *not* to re-mint ids. It is to **pin** them:
+back-fill the ingest ledger so each existing fragment's **existing** id is
+recorded against its ``source_key``. From then on
+``write_fragment_idempotent`` reuses ``record.fragment_id`` and no id ever
+moves — whatever the new derivation would have produced is simply never
+consulted for a unit the ledger already knows.
+
+What this migration touches, and what it refuses to touch:
+
+* **Ledger** — one appended :class:`~creek.ingest.ledger.LedgerRecord` per
+  markdown-sourced fragment, carrying the id read off disk. Never a
+  recomputed id.
+* **Frontmatter** — exactly one added key, ``source.origin_key``. The ``id``,
+  the ``created`` timestamp, the body and the filename are preserved
+  byte-for-byte. Nothing is renamed and nothing is re-minted.
+* **Duplicated vaults** — a ``source_key`` claimed by more than one live
+  fragment is pinned for *neither*. Choosing arbitrarily would bless the wrong
+  fragment forever, and pinning both is not expressible in a
+  one-record-per-key ledger. The operator is pointed at ``creek clean
+  duplicates`` instead. This conflict guard is the migration's primary safety
+  property.
+
+Only the markdown source is in scope: it is the one ledger-wired source
+(:func:`creek.ingest.pipeline.ledger_for_source`), and append-only event
+sources (Discord/chat exports) keep their content-hashed ids and never consult
+the ledger at all.
+
+The run is idempotent — a ``source_key`` already present in the ledger is not
+re-pinned — and ``dry_run=True`` writes nothing while still returning a fully
+populated :class:`PinResult` so a caller can print the plan.
+
+It is also **resumable**. Pinning one fragment is two durable writes that
+cannot be made atomic with respect to each other, so an interrupted run (an
+``OSError`` partway through an N-fragment pass, not just a ``kill -9``) can
+leave a fragment with a ledger record but no stamp. :func:`_pin_one` detects
+and repairs exactly that state rather than skipping the candidate because its
+key is already present — see its docstring for why the record is nonetheless
+written first.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime
+from pathlib import Path  # runtime use: suffix/existence checks on source paths
+
+import frontmatter
+
+from creek.ingest.base import generate_fragment_id
+from creek.ingest.ledger import SourceLedger
+from creek.ingest.pipeline import derive_source_key
+from creek.vault.authors import OTHER_AUTHORS_DIR
+from creek.vault.reader import iter_vault_fragments
+
+# Importing these three from the writer rather than retyping them is
+# deliberate. ``_atomic_write_text`` is the vault's one crash-safe rewrite
+# (tempfile + ``os.replace``); the two relpart constants are the writer's
+# single source of truth for where fragments live, kept that way so the
+# scaffold drift guard can *derive* the directories ``creek init`` ships
+# (#1025). Re-typing a destination is how the scaffold and the writer drifted
+# apart in the first place. ``creek.ingest.refresh`` sets the precedent for
+# reaching across a module boundary for a private helper.
+from creek.vault.writer import (
+    _FRAGMENTS_RELPART,
+    _ORPHANED_RELPARTS,
+    _atomic_write_text,
+)
+
+logger = logging.getLogger(__name__)
+
+_LEDGER_SOURCE: str = "markdown"
+"""Ledger name to back-fill.
+
+Must match the source type :func:`creek.ingest.pipeline.ledger_for_source`
+ledgers, or the records written here would land in a file no ingest ever
+reads.
+"""
+
+_MARKDOWN_SUFFIX: str = ".md"
+"""The only source extension this migration pins.
+
+``MarkdownIngestor.discover`` globs ``*.md`` exclusively, so a fragment whose
+``original_file`` carries any other extension was produced by a different
+ingestor and does not belong in the markdown ledger.
+"""
+
+_FRAGMENT_ROOTS: tuple[tuple[str, ...], ...] = (
+    (_FRAGMENTS_RELPART,),
+    (OTHER_AUTHORS_DIR,),
+    _ORPHANED_RELPARTS,
+)
+"""Vault-relative roots holding fragment files, as ``joinpath`` part tuples.
+
+Borrowed content under ``11-Other-Authors/`` and soft-tombed content under
+``10-Liminal/Orphaned/`` are as ledger-eligible as a native fragment: a tombed
+unit whose source re-appears is *restored* under its preserved id, which only
+works if that id is pinned.
+"""
+
+
+@dataclass(frozen=True)
+class PinResult:
+    """Summary of a :func:`pin_source_ids` run.
+
+    Attributes:
+        pinned: Fragments whose existing id was newly recorded against its
+            ``source_key`` (the count a ``dry_run`` *would* write).
+        already_pinned: Fragments whose ``source_key`` was already in the
+            ledger and were therefore not re-pinned.
+        repaired: Of *already_pinned*, how many were half-pinned by an earlier
+            interrupted run — ledger record present, ``source.origin_key``
+            missing — and had the missing stamp restored by this one. A
+            non-zero count means a previous run did not complete; it is not an
+            error, but it is worth an operator's attention.
+        conflicts: One human-readable line per ``source_key`` claimed by more
+            than one live fragment, naming every claimant's path. None of
+            those fragments were pinned.
+        unpinnable: One human-readable line per fragment that could not be
+            resolved to a live markdown source, naming the fragment path and
+            the reason.
+        reproduced: Advisory diagnostic — how many candidates re-derive their
+            own id from their own stored frontmatter and body. See
+            :func:`_reproduces_own_id`; this gates nothing.
+        examined: Fragment files inspected, i.e. candidates (pinnable or
+            conflicted) plus ``unpinnable``.
+    """
+
+    pinned: int
+    already_pinned: int
+    repaired: int
+    conflicts: list[str]
+    unpinnable: list[str]
+    reproduced: int
+    examined: int
+
+
+@dataclass(frozen=True)
+class _Candidate:
+    """One on-disk fragment resolved to the source unit that produced it.
+
+    Attributes:
+        md_file: Path to the fragment file in the vault.
+        fragment_id: The id as read from the fragment's frontmatter — the
+            value that gets pinned, never a recomputed one.
+        source_path: ``source.original_file`` verbatim; this exact string was
+            hashed into ``fragment_id`` at first ingest.
+        source_key: Ledger key derived from *source_path*.
+        body: The fragment's stored markdown body (``post.content``).
+        created: ``created`` as it appears in the raw frontmatter, or ``None``
+            when absent/unparseable. Only used by the advisory diagnostic.
+        has_origin_key: Whether ``source.origin_key`` is already present on
+            disk. Drives the torn-pin repair in :func:`_pin_one`.
+    """
+
+    md_file: Path
+    fragment_id: str
+    source_path: str
+    source_key: str
+    body: str
+    created: datetime | None
+    has_origin_key: bool
+
+
+def _raw_created(metadata: dict[str, object]) -> datetime | None:
+    """Read ``created`` from *metadata* exactly as the file records it.
+
+    The raw frontmatter value is used rather than ``Fragment.created``:
+    :meth:`creek.models.Fragment._normalise_timestamp` anchors a naive
+    timestamp to America/Los_Angeles at validation time, so the model's value
+    is not necessarily the one that was hashed into the id.
+
+    YAML delivers the key either already parsed into a :class:`datetime` (bare
+    scalar) or as a string (the usual case — the writer serialises the
+    timestamp with ``model_dump(mode="json")``, and PyYAML quotes a string
+    that would otherwise implicitly resolve to a timestamp).
+
+    Args:
+        metadata: The fragment's raw frontmatter mapping.
+
+    Returns:
+        The parsed timestamp, or ``None`` when the key is missing, is a plain
+        date, or does not parse as ISO-8601.
+    """
+    value = metadata.get("created")
+    if isinstance(value, datetime):
+        return value
+    if isinstance(value, str):
+        try:
+            return datetime.fromisoformat(value)
+        except ValueError:
+            logger.debug("Unparseable 'created' frontmatter value: %r", value)
+            return None
+    return None
+
+
+def _unpinnable_reason(source_path_str: str) -> str | None:
+    """Return why *source_path_str* cannot be pinned, or ``None`` if it can.
+
+    Three dispositions are refused, each for its own reason:
+
+    * an empty ``original_file`` gives nothing to key a ledger record on;
+    * a non-``.md`` source came from a different ingestor and does not belong
+      in the markdown ledger;
+    * a source that no longer exists on disk cannot be re-ingested, so pinning
+      it would only add a record no future run will ever match — and, worse,
+      could shadow a *different* file that later takes that path.
+
+    Args:
+        source_path_str: The fragment's ``source.original_file``, verbatim.
+
+    Returns:
+        A human-readable reason, or ``None`` when the source is pinnable.
+    """
+    if not source_path_str:
+        return "fragment records no source.original_file"
+    source_path = Path(source_path_str)
+    if source_path.suffix.lower() != _MARKDOWN_SUFFIX:
+        return f"source {source_path_str!r} is not a {_MARKDOWN_SUFFIX} file"
+    if not source_path.exists():
+        return f"source {source_path_str!r} no longer exists on disk"
+    return None
+
+
+def _collect_candidates(vault_path: Path) -> tuple[list[_Candidate], list[str]]:
+    """Walk the vault's fragment roots and resolve each fragment to its source.
+
+    Args:
+        vault_path: Vault root.
+
+    Returns:
+        ``(candidates, unpinnable)`` — the fragments that resolve to a live
+        markdown source, and one ``"<path>: <reason>"`` line for each that
+        does not.
+    """
+    candidates: list[_Candidate] = []
+    unpinnable: list[str] = []
+    for relparts in _FRAGMENT_ROOTS:
+        for md_file, fragment, body, raw in iter_vault_fragments(
+            vault_path.joinpath(*relparts),
+        ):
+            source_path_str = fragment.source.original_file or ""
+            reason = _unpinnable_reason(source_path_str)
+            if reason is not None:
+                unpinnable.append(f"{md_file}: {reason}")
+                continue
+            candidates.append(
+                _Candidate(
+                    md_file=md_file,
+                    fragment_id=fragment.id,
+                    source_path=source_path_str,
+                    source_key=derive_source_key(source_path_str, vault_path),
+                    body=body,
+                    created=_raw_created(raw),
+                    has_origin_key=bool(fragment.source.origin_key),
+                ),
+            )
+    return candidates, unpinnable
+
+
+def _index_by_source_key(
+    candidates: list[_Candidate],
+) -> dict[str, list[_Candidate]]:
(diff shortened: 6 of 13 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real, well-diagnosed bug: `generate_fragment_id` hashed a timestamp that leaked two host-dependent inputs (naive `datetime.fromtimestamp` under the host `TZ`, and `st_birthtime` present only on macOS/BSD), so the same file minted different fragment ids depending on the ingesting machine. The fix centralizes timestamp derivation in `file_modified_time` (UTC, pure function of the epoch float), hoists `fragment.id = record.fragment_id` above every branch of `write_fragment_idempotent` so the ledger — not the derivation — is the authority on identity, and ships a new `creek ingest --pin-source-ids` migration that back-fills the ledger with **existing** ids rather than re-minting them, deliberately avoiding a full id-remapping migration. The PR description is unusually rigorous about what is and isn't covered, and I independently verified its key cross-file claims against the actual repo (`ledger.py`, `vault/writer.py`, `purge/engine.py`, `creek/ingest/code.py`) — they all check out.

## Strengths
- Root-caused correctly and *found a second bug the linked issue didn't mention* (`st_birthtime` OS-dependence), with a regression test (`test_markdown_id_ignores_st_birthtime`) that fakes `stat()` rather than relying on a real filesystem, correctly noting that `os.utime` clamps `st_birthtime` on APFS and would make a real-FS test prove nothing.
- The `fragment.id = record.fragment_id` hoist in `write_fragment_idempotent` (creek/ingest/pipeline.py) is the right fix and is independently forcing-function-tested (`test_unchanged_branch_reuses_the_ledgered_id_not_the_derived_one`) by perturbing the derivation directly, decoupled from any timezone scenario.
- `pin_ids.py`'s design choices are all individually justified and, as far as I can verify, correct: hashing the **stored** body rather than re-reading the current source (prevents swallowing a pre-migration edit — and there's a dedicated test with a no-trailing-newline body specifically to catch a masked mutation), the conflict guard refusing to arbitrate a duplicated `source_key`, and the `origin_key`-stamped-atomically requirement (confirmed against `purge/engine.py`'s `_extract_source_origin_key` and `VaultWriter.update_fragment`'s frontmatter-preservation behavior — both match the docstring's claims).
- The self-healing "torn pin" repair (record written before stamp, retry repairs a half-pinned fragment) is a genuinely well-reasoned crash-safety property, and `test_an_interrupted_run_is_repaired_by_the_next_one` injects the failure mid-batch to prove it.
- `TOMBING_SOURCES` correctly separates "has a ledger" from "may tomb," closing a real footgun where a borrowed ledger (`ledger_source`) could arm the tomb sweep for a source type whose directory listing isn't authoritative.
- The AST guard (`test_timestamp_derivation_guard.py`) is a nice tree-wide backstop, and it's honest about a rejected third rule (`st_mtime` reads outside `file_modified_time`) that would have been red on `gdrive.py`'s legitimate staleness comparison.
- Extremely thorough, falsifiable test plan: mutation battery on 11 reverts, RED-first tests using relative paths (documented as load-bearing so pinned literal ids stay reproducible), explicit acceptance-criteria-vs-actual-scope reconciliation.

## Security Concerns
🟢 LOW. `_atomic_write_text` is used for the frontmatter stamp (avoids truncation on a mid-run kill), and the migration only touches vault-relative fragment roots. No new input-validation or injection surface introduced.

## Problems
🔴 **The un-pinned-vault warning never reaches the CLI operator on a normal `creek ingest` run.** `unpinned_vault_warning()` (creek/ingest/pipeline.py) is computed correctly inside `run_ingest()` and returned on `IngestRunResult.warnings`, and `test_an_unpinned_vault_is_warned_about_before_it_is_duplicated` verifies it at that layer. But `_run_ingest()` in `creek/cli.py` — the wrapper both the `ingest` command and the scheduled `_sync_ingest_source` path call — has a `tuple[int, list[str], int]` return type that only carries `written, errors, discovered`; `result.warnings` is discarded (`grep -n "\.warnings\b" creek/cli.py` finds zero references). The pipeline layer does `logger.warning(...)` on the message, but nothing in this codebase configures a logging handler (`grep -rln "basicConfig\|RichHandler\|addHandler"` across the whole tree is empty), so that warning is only visible via Python's implicit stderr "handler of last resort" — unstyled, inconsistent with every other operator-facing message in this CLI (which all go through `console.print`), and not exercised by any test. Meanwhile every other new operator message in this PR (`_render_pin_summary`, `_print_pin_findings`) is deliberately and carefully rendered through `console.print` with `soft_wrap=True`/markup escaping.

  This matters because it's the PR's own headline safety story: the description states "Un-migrated vaults are detected... and warned about *before* the run that would duplicate them," and `docs/idempotent-ingest.md` repeats "An un-migrated vault is detected on the next `creek ingest` and warned about before it is duplicated." For the actual `creek ingest --type markdown --input X --vault Y` command an operator runs day-to-day (not `--pin-source-ids`), that promise isn't kept — the vault silently duplicates with no visible warning. No test in this PR exercises the warning through the CLI (`_run_cli("ingest", ...)` without `--pin-source-ids`); the only warning assertion (`test_an_unpinned_vault_is_warned_about_before_it_is_duplicated`, tests/test_pin_source_ids.py:213) calls `run_ingest()` directly, bypassing the CLI wrapper entirely.

  Suggested fix: thread `result.warnings` through `_run_ingest`'s return (or have it print them via `console.print` directly, matching the pattern used elsewhere in this file) and add a CLI-level test that runs `creek ingest --type markdown ...` against an unpinned vault and asserts the warning text appears in `result.output`.

## Code Quality
- `_dispatch_ingest_migration`'s docstring says "`--pin-source-ids` is checked before every other mode, including the two refresh migrations" — true and intentional, but worth a one-line comment at the `if pin_source_ids:` check itself (not just the docstring) since it's easy for a future edit to reorder the `if` chain without reading the docstring.
- `pin_ids.py` reaches into `creek.vault.writer`'s private `_atomic_write_text`, `_FRAGMENTS_RELPART`, `_ORPHANED_RELPARTS` — justified in a comment citing `creek.ingest.refresh` as precedent, which is reasonable, but a small exported "vault layout" surface (even just the two relpart constants) would avoid every future consumer needing the same private-import justification.

## CLAUDE.md Adherence
✅ Tests-first workflow described in detail (RED before GREEN, mutation battery). ✅ No bypasses (no `noqa`/`type: ignore`/skip markers introduced). ✅ Per the PR description, `check-all.sh` passed with coverage/mypy/ruff/xenon all clean — I was not able to independently re-run `./scripts/check-all.sh` or `pytest` in this environment (command execution required interactive approval that wasn't available), so I verified correctness by reading the diff against the actual current source of every module it references instead. ✅ Uses feature branch (`issue/1329-issue`), not a direct commit to main.

## Testing
Very strong at the unit/e2e/migration level — the TZ-across-four-zones matrix, the birthtime fake-stat test, the mutation battery, the AST guard, and the migration's dry-run/idempotency/conflict/repair coverage in `test_pin_source_ids.py` are all well-targeted and specifically designed to fail if the implementation regresses (e.g. the no-trailing-newline body in the stored-body-hash test, the pinned-literal ids rather than set-size assertions). The one gap: **no CLI-level test for the unpinned-vault warning** (see Problems) — everything that tests `unpinned_vault_warning`/`result.warnings` calls `pipeline.run_ingest()` directly, never the `creek ingest` command surface, which is how the warning is supposed to reach a real operator.

## Documentation
`docs/idempotent-ingest.md` is updated thoroughly and accurately describes the migration's behavior and trade-offs — except for the "warned about before it is duplicated" claim, which (per Problems above) isn't actually true for the CLI as shipped. Docstrings throughout `pin_ids.py`, `pipeline.py`, and the ingestors are unusually precise about *why*, not just *what*, and correctly cross-reference the modules they depend on (verified `purge/engine.py` and `vault/writer.py` behavior against the docstrings' claims).



## Verdict: CHANGES_REQUESTED

The core id-derivation fix, ledger-authority hoist, and pin_ids migration are correct, thoroughly tested, and I independently verified every cross-file claim in the description (ledger.py, writer.py, purge/engine.py, code.py all match). But the CLI never surfaces `IngestRunResult.warnings` — the un-pinned-vault advisory this PR exists to deliver is silently dropped by `_run_ingest`'s 3-tuple return in creek/cli.py, so an operator running plain `creek ingest` (not `--pin-source-ids`) gets no visible warning before their vault is duplicated, and no test exercises this through the actual CLI path. This is a real gap in the PR's own headline safety mechanism, not a nitpick.

Geoffe-Ga:
**CI**: 8/12 Green
**Action**: pull comment 5248350846 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**Action**: pull comment 5248350846 to see in-depth feedback and continue iterating

Geoffe-Ga:
You were right on both counts, and the second one is the sharper of the two: the advisory was real, tested, and unreachable.

## The gap, confirmed

`_run_ingest` returned `tuple[int, list[str], int]` — `(written, errors, discovered)` — from an `IngestRunResult` that also carries `warnings`. Both call sites (`creek ingest` and, via `_sync_ingest_source`, `creek sync`) therefore never saw it. Every existing assertion went through `run_ingest()` directly, so nothing noticed. The test I added fails on the pre-fix code with the operator's entire stdout being:

```
First time processing /…/src … --yes; recorded in consent log.
Ingest summary: 1 created, 0 updated, 0 tombed, 0 skipped
Ingested 1 fragment(s).
```

The advisory existed only as a `logging.WARNING` record, which no CLI handler renders.

## What changed

`run_ingest` gained `on_warning: Callable[[str], None] | None`, and every advisory now goes through a single `warn()` closure that appends to `result.warnings`, logs, **and** notifies the callback. Appending without notifying is no longer separable — which is the actual defect class, not just this one instance. `_run_ingest` passes `_print_ingest_warning`, so both CLI ingest paths surface warnings without either having to remember to.

Three deliberate calls:

- **Timing — at detection, before the write pass.** This is why it is a callback rather than a rendered return value. The advisory says "stop before this run duplicates your vault"; delivered after the run it is only an explanation of damage already done. It now prints ahead of the first fragment write, and the test asserts that ordering rather than merely asserting presence.
- **stdout, not stderr.** Every other `creek ingest` advisory (`_warn_if_discovered_but_empty`, `_print_pin_findings`) uses the same stdout console. This command's output is human prose, not data on a pipe, and an advisory routed to a stream none of its neighbours use is one an operator can lose.
- **Exit code unchanged.** A warning reports vault state that will bite later, not a failure of the run in front of it — and `creek sync` shares this path, so escalating would hard-fail every scheduled pass over an un-migrated vault. Happy to make it fatal (or `--strict`-gated) if you'd rather; it is a policy call, not a correctness one, so I left it where the rest of the CLI's warnings sit.

`escape` + `soft_wrap=True` for the same reason as `_print_pin_findings`: the text carries a shell command meant to be copy-pasted, and the test asserts `creek ingest --pin-source-ids --vault <vault>` appears **whole**, so a re-wrap would go red.

## Tests

- `test_cli_ingest_prints_the_unpinned_vault_advisory_before_it_writes` — drives the real `creek ingest` through `CliRunner`, asserts on rendered output, and asserts the advisory precedes the write summary.
- `test_cli_ingest_stays_quiet_about_a_pinned_vault` — without it, printing unconditionally would pass the first test while training the operator to ignore the warning.
- Reverting only the `on_warning=` wiring turns exactly one test red (1 failed, 24 passed), so the guard is load-bearing.
- I also renamed the pipeline-level test from `test_an_unpinned_vault_is_warned_about_before_it_is_duplicated` to `..._advisory_is_recorded_in_the_run_result`. Its old name claimed precisely the property it did not test — which is how this got past me.

## The doc

You were right that `docs/idempotent-ingest.md` promised something that did not ship. It now says the advisory is printed before the run's write pass begins (true as of this change) and then states plainly what it still does **not** do: it is an advisory, not a gate, the exit code is unchanged, and an *unattended* ingest over an un-migrated vault will still duplicate it. Pin first.

## Sibling flattening sites

Your finding turned out to be an instance of a pattern, so I swept for others. **Reported, not fixed** — they are out of scope for this PR and will be filed as follow-ups. Six confirmed, each verified by reading the consumer, and each re-checked against current `origin/main` (not just my branch point) so none of these is stale:

**Same advisory, other surfaces** — `IngestRunResult.warnings` is dropped by *every* MCP path too; `creek_mcp` contains zero occurrences of the string `warnings`:

1. `creek_mcp/tools/upload.py:739-752` — success payload consults only `written`/`created`/`updated`. An upload into an un-pinned vault returns `status: ok` with the advisory nowhere. Confirmed reachable: the call passes `ledger_source=UPLOAD_LEDGER_SOURCE`, so a real ledger resolves.
2. `creek_mcp/tools/journal.py:440-448` — same, and transitively the HTTP API (`creek_mcp/httpapi/journal.py:254-261`; `JournalUpsertResponse` has no warnings field at all). `source_type="markdown"` is exactly the ledger-wired source, so the warning is live here.

**Other advisory fields:**

3. `creek/cli.py:753` — `_sync_ingest_source` discards `errors` *and* `discovered`. Because `discovered` is dropped, `_warn_if_discovered_but_empty` — the #595 unrecognized-export-format guard — is never invoked on the sync path at all. `creek ingest` prints both.
4. `creek/cli.py:727-729` — `_sync_pull` ignores `download_all`'s `DownloadResult` entirely, so per-file Drive failures (quota, revoked token) are dropped and a scheduled sync reports success with files missing from staging.
5. `creek/cli.py:3974-3978` — `creek gdrive --download` prints `Downloaded N / Skipped M`; failed files are counted in neither.
6. `creek_mcp/tools/link.py:69-76` — drops `LinkSummary.oversized_discarded`, documented as data loss and rendered by the CLI, so an agent-driven link run cannot tell fragments were discarded.

One candidate I am explicitly **withdrawing**: my first pass flagged `creek_mcp/tools/author.py` for dropping `AuthoredDraft.findings`, but that was read against my branch point, which predates #1361. On current `main` the narrowing to `privacy_findings` is deliberate and documented — echoing all findings would undo the scrub, since `biographical_grounding` and `attribution_correctness` interpolate corpus text into their own messages. Not a defect; please don't file it.

Also worth knowing: `FilterResult.warnings` and `ContextResult.dropped` are populated but have no production consumer at all — dead, not dropped.

The sweep confirmed the fixed path is clean, and cleared a set of false candidates (`PinResult.conflicts`/`.unpinnable`, `ClassifySummary.errors`, `PipelineResult.errors`, `ReviewSummary.errors`, `Draft.truncated`, MCP `creek.ingest`'s `IngestResult.errors`) — those all do reach a surface. MCP `creek.lint`'s omission of `CheckResult.findings` is deliberate and documented (tier-leak avoidance).

Geoffe-Ga:
## Summary
This PR fixes a real bug where a fragment's id (`generate_fragment_id`) depended on the ingesting host's `TZ` environment variable and (independently) its operating system (`st_birthtime` present on macOS/BSD, absent on Linux). Both fallbacks now route through a single host-independent `file_modified_time` (UTC, pure function of the epoch mtime). Because that changes id derivation for existing vaults, the PR adds a migration (`creek ingest --pin-source-ids`) that back-fills the ingest ledger with each fragment's **existing** id rather than re-minting one, so no id already on disk ever moves. It also hoists `fragment.id = record.fragment_id` above every branch of `write_fragment_idempotent` (closing a latent duplicate-on-drift hazard on the "unchanged" path), splits `TOMBING_SOURCES` from ledger-eligibility, adds an AST guard against reintroducing either host-dependent pattern, and surfaces a new "un-migrated vault" advisory to the operator before the first write.

## Strengths
- The root-cause analysis is precise and the fix is minimal and targeted: both `_resolve_timestamp` fallbacks (markdown, documents) and `images._modified_time` now delegate to one `file_modified_time` helper; verified no other `fromtimestamp`/`st_birthtime` call sites remain in the diff.
- The `fragment.id = record.fragment_id` hoist in `write_fragment_idempotent` (creek/ingest/pipeline.py) is a genuine forcing-function fix — verified the "unchanged" branch previously wrote whatever id had just been *derived*, not the ledgered one, which is exactly the silent-duplication mechanism described.
- The migration's crash-safety argument holds up against the actual ledger implementation: `SourceLedger.record()` (creek/ingest/ledger.py `_append`) opens the file in append mode and writes synchronously per call, so "the ledger record is durable before the stamp is attempted" is factually true, not just asserted. The half-pinned repair path (`_pin_one`) is correctly self-healing on retry.
- The conflict guard in `pin_source_ids` (a `source_key` claimed by 2+ live fragments is pinned for neither) is the right safety property for a one-record-per-key ledger, and is built from a full pre-pass index rather than pinning greedily.
- `TOMBING_SOURCES` correctly separates "has ledger-backed identity" from "may tomb missing units" — verified `resolve_ledger`'s `ledger_source` override previously let a borrowed ledger arm `tomb_missing_units` unintentionally; this is now gated separately and covered by `test_borrowing_a_ledger_for_identity_does_not_arm_the_tomb_sweep`.
- Test coverage is unusually thorough and demonstrably load-bearing: id-across-timezones tests, an e2e duplicate-reproduction test, a dedicated "edit before migration is not swallowed" test (verified the stored-body-hash rationale is correct: hashing the *current* source instead of the *stored* body would cause a post-ingest edit to be silently and permanently dropped), an interrupted-run repair test, and a syntactic AST guard with proofs that each rule actually fires.
- The PR description is exceptionally transparent about scope: it explicitly enumerates every acceptance criterion left formally unmet by the "pin, don't re-mint" inversion and justifies each rather than glossing over it (documents remain unledgered per #1363, code.py deferred per #1364, an unresolved-conflict re-warn gap deferred to #1367).
- Self-review already caught and fixed a real ordering bug (stamp-before-record vs. record-before-stamp) with a reproduction and a regression test, per the PR's Gate 2.5 section — verified the fix (`_pin_one` writes the ledger record first) is what's actually in the diff.

## Security Concerns
🟢 LOW — No injection, path traversal, or secrets handling concerns. Operator-facing paths printed by the CLI (`_print_pin_findings`, `_print_ingest_warning`) are markup-escaped via `escape()` before being interpolated into Rich console output, preventing Rich markup injection from a maliciously-named file path.

## Problems
None found that block merge.

## Code Quality
- `pin_ids.py`'s `_pin_one`/`_index_by_source_key` don't re-verify that an already-ledgered `source_key`'s recorded `fragment_id` matches the on-disk candidate's id before treating it as `already_pinned` — a purely theoretical edge case (hand-edited frontmatter after pinning) outside the PR's stated scope, not worth blocking on.
- `_dispatch_ingest_migration` lets `--pin-source-ids --refresh-dates` silently run only the pin migration (refresh is ignored) rather than rejecting the combination the way `--dry-run` + a non-pin migration is explicitly rejected. Minor inconsistency in an unlikely-to-be-used flag combination.
- The AST guard's `_naive_fromtimestamp_sites` only recognizes `tz=` passed as a keyword; a hypothetical `datetime.fromtimestamp(x, UTC)` positional call would be flagged as a false positive. This matches the codebase's existing keyword-only convention and is a reasonable, intentional tradeoff, not a bug.

## CLAUDE.md Adherence
✅ TDD (RED tests before the fix, per the PR's own account and the tests' content demonstrating pre-fix failures)
✅ No bypass patterns (no new `# noqa`/`# type: ignore`; three draft-only ignores were removed by fixing types)
✅ Conventional commit messages (`fix(ingest): ...`)
✅ Coverage/lint/type/complexity claims are consistent with what's observable in the diff (docstrings are extensive and Google-style throughout)
✅ Uses project's atomic-write helper (`_atomic_write_text`) rather than a plain `write_text` for the frontmatter stamp, matching the "crash-safe rewrite" convention cited from `creek.vault.writer`

## Testing
Extensive and well-targeted: host-independence unit tests (TZ matrix + faked stat shapes for the birthtime axis), an e2e duplicate-prevention test using the real shared pipeline, a forcing-function test that perturbs the derivation directly (independent of any timezone), a stored-body-hash correctness test, an interrupted-run/repair test, CLI-level tests including advisory-ordering assertions, and a syntactic AST guard with self-verifying "does the rule actually fire" cases. The PR's stated mutation-testing pass (11 reverts, 0 survivors) is consistent with the density and specificity of the assertions actually present in the diff.

## Documentation
`docs/idempotent-ingest.md` is updated with a full "why", a "what it does / refuses to do" section, and the two behavior changes (macOS `created` now reflects mtime not birthtime; documents remain unledgered). Docstrings throughout `pin_ids.py`, `pipeline.py`, and the ingestor `_resolve_timestamp` methods explicitly warn future readers against reintroducing either host-dependent pattern, which is a good defense-in-depth alongside the AST guard.



## Verdict: LGTM

The fix is correct (verified the timestamp/id derivation change, the ledger-authority hoist in write_fragment_idempotent, and the pin_ids migration's record-before-stamp durability claim against the actual SourceLedger.record implementation, which does a synchronous per-call file append — so the crash-safety/repair argument holds). Tests are extensive and demonstrably load-bearing, the AST guard is real, and the PR description pre-emptively documents every AC left formally unmet with sound justification. Nothing found rises to blocking; the only items worth flagging are minor non-blocking observations.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5249393894 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Geoffe-Ga:
## Summary
This PR fixes a real, well-diagnosed bug: `generate_fragment_id` hashed a filesystem timestamp that was rendered via a naive, host-local `datetime.fromtimestamp(mtime)` and, in markdown's case, via `st_birthtime` (present on macOS/BSD, absent on Linux) — so the same file could mint a different `frag-<sha>` id depending on the ingesting host's `TZ` and OS. The fix centralises both fallbacks on a pure, UTC `file_modified_time`, and — because that changes derivation for existing vaults — adds a `creek ingest --pin-source-ids` migration that back-fills the ledger with each fragment's *existing* id rather than re-minting one, plus an AST guard against reintroducing either naive pattern. A second commit tries to surface a related "un-pinned vault" advisory that a prior refactor had made unreachable.

The design reasoning (pin vs. re-mint, ledger-first write ordering, stored-body vs. current-source hashing, the self-healing torn-pin repair) is sound and well-tested. However, the new advisory-surfacing logic has a genuine, uncovered correctness bug — see Problems.

## Strengths
- Root-caused two independent host-dependence bugs (TZ *and* OS/`st_birthtime`), not just the one the linked issue named — the birthtime axis is real and easy to miss.
- The pin-vs-remint design decision is well-justified: reusing the ledger as the identity authority (already established by #672) makes the AC's re-mapping requirements vacuous by construction rather than requiring them to be reimplemented.
- `fragment.id = record.fragment_id` hoisted above every branch of `write_fragment_idempotent` is a real forcing-function fix — it closes a duplication hazard (unchanged-branch writing a freshly-derived id) that is independent of this specific timezone bug and would have bitten on any future derivation drift.
- The stored-body-vs-current-source hashing subtlety in `pin_ids.py` (`_stored_body_hash`) is a genuinely non-obvious correctness requirement, and it's covered by a dedicated test that edits the source *before* pinning.
- The interrupted-run self-healing repair (`_pin_one`'s "record first, stamp second, repair on retry" ordering) is a well-reasoned resolution to an unavoidable two-write atomicity gap, with a test that injects a failure mid-pass.
- `TOMBING_SOURCES` correctly separates "has ledger-backed identity" from "may soft-tomb missing units," closing a latent hazard where borrowing a ledger for identity (uploads) could have silently armed directory-tombing for a non-authoritative directory listing.
- Extensive, literal-pinned regression tests (`frag-<sha>` values reproduced across 4 timezones) make a future silent regression in the derivation loud rather than quiet.

## Security Concerns
🟢 LOW — No new attack surface. `_stamp_origin_key` and the pin migration write only vault-local files via the existing atomic-write helper; no new external input handling, subprocess use, or path traversal risk identified.

## Problems
🔴 **`unpinned_vault_warning` checks the wrong ledger when `ledger_source` is overridden** (`creek/ingest/pipeline.py`, new `run_ingest` wiring around line ~1276 in the diff, and `unpinned_vault_warning` itself).

`run_ingest` resolves `ledger = resolve_ledger(source_type, vault_path, ledger_source)` and then unconditionally calls `unpinned_vault_warning(ledger, vault_path)`. But `resolve_ledger` returns *whatever ledger the caller asked for* — for the MCP upload tool (`creek_mcp/tools/upload.py`, `_ingest_outcome` → `run_ingest(..., ledger_source=UPLOAD_LEDGER_SOURCE)`), that's the `"upload"` ledger, not `"markdown"`, **even when the uploaded file is a `.md` routed to `source_type="markdown"`** (`route_to_ingestor` maps `.md` → `"markdown"` via `_INGESTOR_BY_EXTENSION`, and this is unconditional — `ledger_source` is always `UPLOAD_LEDGER_SOURCE` regardless of `source_type`).

`unpinned_vault_warning`'s text and logic are specifically about the **markdown** ledger's pin status ("This vault has markdown fragments but an empty ingest ledger... Run `creek ingest --pin-source-ids`"), but its emptiness check (`len(ledger) > 0`) runs against whatever ledger got resolved. Concretely:
- **False positive**: the first time the upload tool is ever used on a vault (even one already fully migrated via `--pin-source-ids`, so the real `"markdown"` ledger is populated), the fresh, unrelated `"upload"` ledger is empty → the advisory fires, incorrectly telling the operator their already-pinned vault needs migrating.
- **False negative**: once the `"upload"` ledger has any record (after the first successful upload), `len(ledger) > 0` short-circuits the check to `None` **forever after**, even if the actual `"markdown"` ledger is genuinely un-pinned and every subsequent upload of a `.md` file is at risk of the exact duplication this PR exists to prevent.

Separately (lower severity, may be intentional scope): `creek_mcp/tools/upload.py`'s `_ingest_outcome` doesn't pass `on_warning=` to `run_ingest` and never reads `result.warnings`, so even a correctly-detected warning is silently discarded on this path — the PR's docs only claim coverage for "every ingest driven through the CLI, `creek sync` included," so this omission may be a known/accepted scope limit, but the ledger-mismatch bug above is not scoped away by that caveat and is a real defect in the shared `unpinned_vault_warning`/`run_ingest` logic itself.

**No test exercises this.** Every test that touches `unpinned_vault_warning` (`tests/test_pin_source_ids.py`) constructs the ledger via `SourceLedger.load(vault, source="markdown")` directly or drives `run_ingest`/the CLI with the default (unoverridden) `ledger_source`. Nothing in the otherwise-extensive suite calls `run_ingest(..., ledger_source=<override>)` and checks the resulting `warnings`, which is precisely the configuration the production MCP upload path always uses. Given the PR's own standard ("insufficient testing is BLOCKING") and that this is new code written specifically to fix "the warning was produced, stored, tested, and unreachable," this should be fixed (check the markdown ledger's own state, independent of the run's resolved identity ledger) and covered by a test before merge.

## Code Quality
- `pin_ids.py`'s `_FRAGMENT_ROOTS` reaching into `creek.vault.writer`'s private `_FRAGMENTS_RELPART`/`_ORPHANED_RELPARTS`/`_atomic_write_text` is justified in-line with a stated precedent (`creek.ingest.refresh`), but it's still cross-module private-symbol coupling worth keeping an eye on if `vault/writer.py`'s internals shift.
- The docstrings throughout (especially `pin_ids.py` and `pipeline.py`) are unusually long/discursive for what they document; this is consistent with the codebase's existing style elsewhere (e.g. `resolve_ledger`), so not a deviation, but worth noting if a lighter touch is ever wanted.
- `_dispatch_ingest_migration`'s ordering (pin-source-ids checked before the bare `--dry-run` refusal, before the two refresh migrations) is correct but a little easy to get wrong on a future edit; a short comment or a small state machine (rather than sequential ifs) might reduce that risk, though this is a minor nit.

## CLAUDE.md Adherence
✅ TDD narrative is credible and detailed (RED-first commit description, mutation battery of 11 reverts, Gate 2.5 self-review with a reproduced-and-fixed finding).
✅ No bypasses observed in the diff (no `# noqa`, `# type: ignore`, `@pytest.mark.skip`).
✅ Conventional commit messages, feature branch, detailed PR description exceeding what CLAUDE.md requires.
⚠️ Could not independently run `./scripts/check-all.sh`, mypy, ruff, or coverage in this review environment (tooling unavailable) — the 94.40% coverage / "12 checks passed" claims are taken on the author's word and CI's own gate, not independently re-verified here.
❌ The advisory-surfacing feature (`unpinned_vault_warning`) itself has a coverage gap for the `ledger_source`-override case described above, which is exactly the kind of gap `./scripts/coverage-per-file.sh` and the project's TDD-first culture are meant to catch.

## Testing
Test quality is generally excellent: `test_ingest_id_host_independence.py` and `test_ingest_tz_duplicate_e2e.py` pin literal fragment ids across timezones (harder to fool than a set-size assertion), correctly avoid `--incremental` (which would mask the bug via `should_skip_unit`), and use relative paths so pinned literals reproduce across machines. `test_pin_source_ids.py` covers the conflict guard, dry-run, idempotency, the interrupted-run repair, and the subtle stored-body-hash correction with dedicated before/after-migration edit scenarios. `test_timestamp_derivation_guard.py`'s AST guard is a reasonable belt-and-suspenders check with a documented, justified exclusion (`gdrive.py`'s legitimate mtime comparison).

The one meaningful gap is the `ledger_source`-override interaction with `unpinned_vault_warning` described in Problems — this is the missing test that would have caught the bug.

## Documentation
`docs/idempotent-ingest.md` is updated thoroughly and accurately reflects the shipped behavior, including the two explicitly-called-out behavior changes (macOS `created` now reflects mtime not birthtime; documents remain unledgered). The PR description itself is exceptionally thorough — including an explicit, honest list of which acceptance criteria are formally unmet and why the inversion (pin vs. re-mint) makes them vacuous rather than skipped. This transparency is a strength, not a smell.



## Verdict: CHANGES_REQUESTED

A real, reachable, and untested bug: `unpinned_vault_warning` is checked against whichever ledger `resolve_ledger(source_type, vault_path, ledger_source)` happens to resolve for the current run, not the markdown ledger the advisory is actually about — so any run with a `ledger_source` override (the MCP upload tool always passes `ledger_source="upload"`, including for uploaded `.md` files) checks an unrelated ledger's emptiness instead. This produces false positives (spurious "run --pin-source-ids" warnings on an already-pinned vault, the first time the upload path runs) and false negatives (the warning permanently stops firing through that path once the unrelated ledger gets any record, even if the real markdown ledger is still un-pinned and duplication-prone) — undermining the exact safety mechanism the PR's second commit exists to deliver. No test in the (otherwise extensive) suite exercises `unpinned_vault_warning`/`run_ingest` with a `ledger_source` override, so this shipped unnoticed through the PR's own Gate 2.5 self-review.

Geoffe-Ga:
**CI**: 10/12 Green
**Action**: pull comment 5249552445 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**Action**: pull comment 5249552445 to see in-depth feedback and continue iterating

Geoffe-Ga:
Confirmed, in both directions, and fixed at the root in `71a8d67`.

## The finding was real

`unpinned_vault_warning(ledger, vault_path)` took the run's *resolved* ledger, and
`run_ingest` passed it the result of `resolve_ledger(source_type, vault_path, ledger_source)`.
`creek_mcp/tools/upload.py:538` always passes `ledger_source=UPLOAD_LEDGER_SOURCE`, and
`route_to_ingestor` maps `.md` to the markdown ingestor (`_INGESTOR_BY_EXTENSION` in
`creek/ingest/gdrive.py:92`) — so an uploaded `.md` ran markdown identity while the advisory
weighed `upload.jsonl`. Two tests written against the unmodified code, asserting on the
emitted advisory (the `on_warning` callback, not `IngestRunResult.warnings`):

```
FAILED test_a_pinned_vault_is_not_warned_about_through_a_borrowed_ledger
  - AssertionError: assert ['This vault ...to preview).'] == []
FAILED test_an_unpinned_vault_is_warned_about_through_a_borrowed_ledger
  - assert [] != []
```

The first is the spurious warning on an already-migrated vault. The second is the one that
matters: markdown ledger empty, one record in the upload ledger, advisory gone — silently,
permanently, on the vault that needs it.

## The fix, and why it cannot drift again

The ledger parameter is **gone**. The advisory's subject is fixed by what the advisory says,
so it loads the markdown ledger itself from `LEDGERED_SOURCE_TYPE`. There is no longer a
parameter through which the wrong ledger can arrive — the failure mode is unrepresentable,
not merely un-triggered.

What still gates it is the run's `source_type`, which #1329 moved id derivation for. Unlike a
resolved ledger, a `source_type` is the run's own input verbatim and no `ledger_source`
override can move it. A document or image run is quiet (asserted); an uploaded `.md` on a
genuinely un-pinned vault is warned (asserted).

## On the shape, not just the instance

You are right that this is the second defect in the same advisory, and the shape invited it:
the function accepted a *derived, overridable* object as its subject, so being correct
depended on every caller resolving the right one. Two changes beyond the instance:

1. The overridable parameter is removed rather than corrected. A caller cannot get this wrong
   because a caller cannot say anything about it.
2. `"markdown"` was written out three times — `ledger_for_source`, the advisory, and
   `pin_ids._LEDGER_SOURCE`, the last under a docstring promising to match the first. They now
   share one `LEDGERED_SOURCE_TYPE` constant. A comment is not an enforcement, and the
   migration's ledger and the advisory's ledger silently diverging is precisely the class of
   bug you found.

## Verification

- **Mutation battery**: restoring the pre-fix behaviour alone (advisory judging the resolved
  ledger, override included) turns exactly the two new tests red; the rest of the module stays
  green. The third new test (`test_a_non_markdown_source_type_is_never_warned_about`) passes in
  isolation under that mutation — it is a boundary guard against over-correcting into noise, not
  a witness to this defect, and I am not claiming it as one.
- **The two properties this PR rests on, re-run after touching the advisory path again**:
  `test_pinning_makes_the_next_reingest_a_no_op_under_the_original_id` (legacy vault built by a
  direct `VaultWriter` write with an empty ledger — no `run_ingest` pinning it by construction —
  asserting the non-reproducible `_PRE_FIX_ID` literal survives) and
  `test_cli_ingest_prints_the_unpinned_vault_advisory_before_it_writes`. Both pass.
- **Gate 2**: `./scripts/check-all.sh` exit 0 — 12/12 checks, 8957 passed, 94.40% coverage.
- **Bypass sweep** over the whole branch diff including test files: no `noqa`, `type: ignore`,
  `pragma: no cover`, `nosec`, `skip`/`xfail` marker, or threshold change. The only `skip`
  matches are prose in docstrings.

`docs/idempotent-ingest.md` now states which ledger the advisory weighs and why a borrowed one
is never substituted.

Geoffe-Ga:
## Summary
This PR fixes a real, well-diagnosed bug: `generate_fragment_id` hashed a fragment's timestamp, and the filesystem-mtime fallback used a naive, timezone-dependent conversion (`datetime.fromtimestamp(mtime)`) plus a platform-dependent `st_birthtime` read — together making a fragment's id a function of the ingesting host's `TZ` and OS. Rather than re-minting existing ids (as the linked issue's AC technically demands), the PR pins them: it centralizes the timestamp derivation in `file_modified_time` (pure function of the epoch float, UTC), hoists `fragment.id = record.fragment_id` above every branch of `write_fragment_idempotent` so the ledger is authoritative on identity regardless of derivation drift, and ships a new `creek ingest --pin-source-ids` migration to back-fill the ledger for pre-existing vaults. The PR description is unusually explicit about which acceptance criteria this inversion leaves formally unmet and why.

## Strengths
- The root-cause diagnosis (naive `fromtimestamp` + `st_birthtime` platform divergence) is precise and reproduced with pinned literals across 4 timezones/2 ingestors.
- The "pin, don't re-mint" architectural choice is well-justified against the existing ledger-as-identity-authority invariant (#672 SPEC R1) and is proven non-destructive with a scripted before/after run in the PR description.
- Hoisting `fragment.id = record.fragment_id` above every branch of `write_fragment_idempotent` closes a real, independent hazard (any future derivation drift silently duplicating ledgered units) — this is tested independently of timezones (`test_unchanged_branch_reuses_the_ledgered_id_not_the_derived_one`).
- `pin_ids.py`'s design decisions are all individually sound and tested: conflict-guard-first (never bless one of two claimants), hashing the **stored** body rather than re-reading the current source (verified with a dedicated pre/post-migration edit test, deliberately written with no trailing newline to defeat an accidental pass), atomic frontmatter writes, and a self-healing repair path for a torn ledger-record/stamp pair (tested by injecting `OSError` mid-pass).
- `TOMBING_SOURCES` correctly separates "may borrow a ledger for identity" from "may tomb units missing from a directory listing" — a real footgun for future callers of `ledger_source`.
- New AST guard (`test_timestamp_derivation_guard.py`) prevents regression of both patterns tree-wide, with tests proving each rule actually fires against a scratch package, and a documented, justified rejection of a third overly-broad rule.
- Excellent test hygiene throughout: relative-path fixtures deliberately chosen so pinned literal ids are reproducible, restoration of process-global `TZ` state via `time.tzset()`, CLI-level tests through `CliRunner` rather than only unit-level.
- Docs (`docs/idempotent-ingest.md`) are updated in lockstep with the behavior and clearly flag the two behavior changes (macOS `created` now uses mtime not birthtime; documents remain unledgered).

## Security Concerns
🟢 LOW — No new attack surface. `pin_ids.py` only reads/writes within the vault it's pointed at (local CLI tool), and frontmatter rewrites go through the existing atomic-write helper. No injection, subprocess, or network surfaces touched.

## Problems
🟡 Not blocking, but worth fixing before this ships to real users: `_unpinnable_reason()` in `creek/ingest/pin_ids.py` (`creek-tools/creek/ingest/pin_ids.py:751`) checks `source_path.exists()` directly on `source.original_file`, a string that (per `str(raw.path)` at the point of original ingest) can be a **relative path**, resolved against whatever CWD the operator happened to be in when they ran `creek ingest --input <relative-dir> ...`. If `--pin-source-ids` is later run from a *different* working directory, this check can return `False` for a fragment whose source file is perfectly present, mis-reporting it as `"source ... no longer exists on disk"`. That fragment is then left unpinned, and the very next ordinary re-ingest — the scenario this whole PR exists to prevent — will derive a fresh id for it and write a duplicate. This is distinct from (and not covered by) `derive_source_key`, which tolerates CWD skew because it falls back to the bare filename; `_unpinnable_reason`'s existence check has no equivalent fallback. It's not silent (the CLI prints every unpinnable path+reason, and `examined`/`pinned` counts would look obviously wrong), and it's recoverable by re-running from the correct directory, which is why I'm not marking this a hard blocker — but it undercuts the PR's own stated goal of eliminating host/environment-dependence, and none of the new tests exercise a relative-path/CWD-mismatch scenario (`test_an_unresolvable_source_is_reported_and_not_pinned` only covers an empty `original_file` and a literally-deleted absolute path). Recommend either resolving against a stable anchor (mirroring `derive_source_key`'s `.resolve()` + vault-relative/bare-filename fallback) or explicitly documenting that `--pin-source-ids` must be run from the same working directory as the original ingest.

## Code Quality
- Minor UX nit: `_render_pin_summary` in `creek/cli.py` prints `"repaired {n} half-pinned fragment(s)..."` even under `--dry-run`, without a "would repair" framing (unlike the "Would pin"/"Pinned" verb switch just above it). The overall `"Dry run: nothing was written"` line disambiguates, but the repaired line reads as past-tense action on its own.
- Everything else here is exceptionally clean: docstrings consistently explain *why* (not what), the new `LEDGERED_SOURCE_TYPE`/`TOMBING_SOURCES` constants replace string-literal duplication that had already caused one bug (the ledger-mismatch fixed in the third commit), and the `pin_ids.py` module docstring is a genuinely useful design record.

## CLAUDE.md Adherence
✅ TDD: description documents RED-first tests plus an 11-revert mutation battery with zero survivors.
✅ Scripts-only workflow implied by description (`./scripts/check-all.sh`, no direct tool invocation visible in the diff).
✅ No shortcuts: no `# noqa`, `# type: ignore`, skipped/deleted tests found in the diff; three draft-stage `# type: ignore`s were reportedly removed by fixing types.
✅ Docstrings: Google-style, Args/Returns/Raises present throughout new code.
✅ Commit messages: conventional-commit style, detailed bodies explaining rationale — consistent with repo convention.
⚠️ I could not independently verify the claimed coverage/mypy/ruff/xenon pass rates or CI status from this sandbox (no execution access), so that portion of Gate 2/3 compliance is taken on the PR description's word rather than confirmed directly.

## Testing
Very strong. New test files cover: host-timezone independence at the unit level (`test_ingest_id_host_independence.py`), the end-to-end duplication consequence through the real pipeline (`test_ingest_tz_duplicate_e2e.py`), the full migration surface including conflict handling, idempotency, resumability/repair, stored-body-hash correctness (both post- and pre-migration edit ordering), CLI wiring and `--dry-run` refusal semantics (`test_pin_source_ids.py`), and a tree-wide AST regression guard with self-tests proving each rule actually fires (`test_timestamp_derivation_guard.py`). The one gap is the CWD/relative-path existence-check scenario noted above under Problems.

## Documentation
`docs/idempotent-ingest.md` is updated thoroughly and matches the implementation: migration usage, rationale, the two explicit behavior changes, and the ledger-authority semantics for the unchanged branch. The PR description itself is an unusually complete design record, including an explicit, non-glossed list of acceptance criteria the chosen approach leaves formally unmet and why — which is exactly the right way to surface a deliberate scope deviation from the linked issue.



## Verdict: COMMENTS

Nothing here blocks merge — the core fix (hoisting ledger identity above every write branch, centralizing UTC-only mtime derivation, and the pin-not-remint migration) is correct, extensively tested, and well-documented. But there is one substantive, unaddressed suggestion: `_unpinnable_reason`'s `source_path.exists()` check in `pin_ids.py` is CWD-dependent for relatively-pathed sources and, unlike `derive_source_key`, has no fallback — running the one-time migration from a different working directory than the original ingest can mis-mark live fragments as unpinnable, leaving them exposed to exactly the duplication bug this PR exists to fix. It's visible (not silent) and recoverable, so it doesn't need to block, but the author should weigh fixing or documenting it before this ships.

Geoffe-Ga:
**CI**: 10/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5249750716 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5249750716 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Closing the CWD-dependence finding (`257b6ff`)

The reviewer marked this non-blocking and was right that it is visible and recoverable. Fixing it anyway: the migration's whole job is to protect existing fragments from re-minting, and this made it silently skip the ones it was run to protect. A report line an operator scrolls past is not protection.

### The asymmetry is real

Same recorded path, same vault, two working directories:

```
CWD=<tmp>/vault
   derive_source_key  -> '00-Inbox/a.md'
   _unpinnable_reason -> None
CWD=<tmp>/elsewhere
   derive_source_key  -> 'a.md'
   _unpinnable_reason -> "source '00-Inbox/a.md' no longer exists on disk"
```

Both were CWD-anchored, but `derive_source_key`'s `except ValueError: return candidate.name` degrades to a usable key while `_unpinnable_reason`'s bare `Path(...).exists()` hard-fails. So the two functions disagreed about where one recorded path pointed: one derived a key for a source the other called deleted.

The premise checks out too. `source.original_file` is stored verbatim — `MarkdownIngestor.parse` records `str(raw.path)` (`creek/ingest/markdown.py:378`) and the CLI never resolves `--source` — so a vault ingested with `creek ingest --source 00-Inbox` from the vault root holds relative recorded paths. Reachable, not theoretical.

### RED, on the migration's outcome

A live in-vault source, migrated from a different directory:

```
PinResult(pinned=0, already_pinned=0, repaired=0, conflicts=[],
          unpinnable=[".../01-Fragments/Notes/2024-03-14-note.md:
                       source '00-Inbox/note.md' no longer exists on disk"],
          reproduced=0, examined=1)
```

The file is sitting in the vault. Assertions are on `pinned`, the stamped `origin_key`, and a follow-on re-ingest (`unchanged == 1`, one file, original id) — not on the predicate's return.

### The fix unifies the two paths

I did not patch the predicate alone; two functions disagreeing about how to resolve one recorded path is the same class of bug this PR already fixed twice (the ledger parameter, `"markdown"` written three times). Both now share one resolution, `creek.ingest.pipeline.resolve_recorded_source`:

- current directory still wins when it resolves — an in-place run keys exactly as before;
- the vault root is consulted only for a relative path that names nothing where it stands;
- a path that resolves neither way is still refused, so the shadow-a-later-file safety property is intact.

That also removes the key drift the probe shows: a relative record now pins under the same vault-relative `source_key` from either directory, so the next ingest looks the pin up instead of re-minting past it. Widening `derive_source_key` is safe for its other call sites (`pipeline.py:194`, `upload.py:164`, `journal.py:192`) — they pass paths that already resolve, so the new arm never fires for them.

### Mutation battery — 5 killed, 1 equivalent

| # | Mutation | Result |
|---|---|---|
| M1 | `resolve_recorded_source` never anchors to the vault | killed (2) |
| M2 | vault anchoring preferred over the working directory | killed (1) |
| M3 | `derive_source_key` reverts to the bare CWD-anchored `Path` | killed (2) |
| M4 | `_unpinnable_reason` reverts to the bare CWD-anchored `Path` | killed (2) |
| M5 | drop the `is_absolute()` short-circuit | **survived — equivalent** |
| M6 | existence refusal removed (pin everything) | killed (2) |

M5 is not a coverage gap: `vault_path / <absolute>` returns that absolute path again for every vault shape (verified across absolute, relative and `/` roots), so removing the clause cannot change the result — only add a redundant `stat`. No test can kill it. The clause is kept for intent and the code now says so in a comment, so a reader does not mistake it for a live branch.

M2 earned its own test: without it, the fix would quietly re-point already-correct records at a same-named file that happens to sit in the vault.

### Gates

- **Gate 2:** `./scripts/check-all.sh` → **exit 0**, 12/12 checks, 8961 passed / 5 skipped, coverage 94.40%.
- **Regression sweep:** 1818 tests across ingest/pipeline/ledger/upload/journal/pin/purge/refresh — no failures from widening `derive_source_key`.
- **Bypass sweep** across the whole branch diff including tests (#1366): no `noqa`, no `type: ignore`, no `pragma: no cover`, no `skip`/`xfail` marks, no threshold or config edits. Every grep hit was the word "skip"/"omitted" in prose.

Scope held to this one finding — the eight formally-unmet acceptance criteria, the pin-vs-remint inversion, and the documents-unledgered decision (#1363) are untouched.

Geoffe-Ga:
## Summary
This PR fixes a real, high-impact bug: `generate_fragment_id` hashed a timestamp that was host-dependent (naive `datetime.fromtimestamp(mtime)` picked up the host `TZ`, and `st_birthtime` differed by OS), so the same file could mint different fragment ids across machines/timezones. Rather than re-minting existing ids (which the linked issue's AC technically demanded), the PR fixes the derivation going forward and adds a `--pin-source-ids` migration that back-fills the ledger with each fragment's *existing* id so nothing moves. It also hoists ledger-id assignment above every branch of `write_fragment_idempotent` (closing a related "unchanged branch writes the derived id" duplication hazard) and surfaces a new operator advisory when an un-migrated vault is about to be re-ingested.

## Strengths
- The root-cause analysis is excellent and the PR description is unusually rigorous about what is and isn't covered — it explicitly enumerates which acceptance criteria are left formally unmet and why, rather than glossing over them.
- `write_fragment_idempotent`'s hoisted `fragment.id = record.fragment_id` (`creek/ingest/pipeline.py:336`) is the correct fix for the actual production hazard (silent duplication via derivation drift), independent of the timezone bug itself, and is covered by a dedicated forcing-function test that perturbs the derivation directly (`test_unchanged_branch_reuses_the_ledgered_id_not_the_derived_one`).
- `TOMBING_SOURCES` correctly separates "has ledger-backed identity" from "may tomb missing units" — a real latent footgun for `ledger_source` overrides, closed with a dedicated regression test.
- The pin migration's conflict guard (`_index_by_source_key` in `pin_ids.py`) is a well-reasoned safety property: a `source_key` claimed by 2+ fragments is pinned for neither, avoiding an arbitrary "bless one claimant" choice that would permanently orphan the other.
- Hashing the *stored* body rather than re-reading the current source (`_stored_body_hash`) correctly avoids silently swallowing a post-ingest edit — verified by `test_an_edit_made_before_the_migration_is_not_swallowed`.
- The self-review loop documented in the PR description (finding and fixing a torn-pin/crash-safety issue via `code-review-orchestrator`, then reproducing it before fixing) is exactly the right process, and the resulting fix (ledger-record-first + self-healing repair on retry) is the correct tradeoff given a ledger record can't be un-minted but a missing frontmatter stamp can be restored.
- `resolve_recorded_source` fixing the CWD-anchoring bug in the migration (so it works regardless of the directory it's run from) closes a real correctness gap that would have silently defeated the whole migration for any vault ingested with a relative `--source`.
- Test coverage is extensive: host-timezone matrix tests, an OS-shape fake-stat test (correctly avoiding the APFS `st_birthtime`-follows-mtime trap), an e2e duplicate-prevention test, and a syntactic AST guard against reintroducing either host-dependent pattern.

## Security Concerns
🟢 LOW — None found. Frontmatter rewrites go through the vault's atomic write helper (tempfile + `os.replace`), avoiding truncation-on-crash. No new user input surfaces or subprocess/file-path handling outside the existing vault-relative patterns.

## Problems
None blocking. One correctness gap worth a look before merge (see Code Quality) but it does not compromise the migration's core safety properties (conflict guard, atomic writes, idempotency all hold regardless).

## Code Quality
- `unpinned_vault_warning` (`creek/ingest/pipeline.py:482-493`) only checks `vault_path / "01-Fragments"` for existing `*.md` fragments to decide whether the vault is "un-migrated," but `pin_ids.py`'s own `_FRAGMENT_ROOTS` treats `11-Other-Authors/` and `10-Liminal/Orphaned/` as equally ledger-eligible fragment roots. A vault whose only markdown fragments currently live under those two directories (e.g. one that's mostly borrowed/tombed content with an empty or near-empty `01-Fragments/`) would silently skip the advisory even though those fragments still need pinning. This is a narrow edge case in practice (most vaults populate `01-Fragments` first), but it's a real inconsistency between the detector and the migration's own definition of "in scope," and would be a one-line fix (loop the same `_FRAGMENT_ROOTS`-style check, or expose a shared helper).
- Minor: `_unpinnable_reason` and `derive_source_key`/`_collect_candidates` each independently call `resolve_recorded_source` for the same source string — not a bug, just a small amount of redundant `Path.exists()` I/O per candidate; not worth blocking on.

## CLAUDE.md Adherence
✅ TDD is clearly followed (RED-first tests described and visible in the new test modules' docstrings explaining what failed pre-fix).
✅ No bypasses — no `noqa`/`type: ignore`/skipped tests found in the diff; PR description explicitly confirms zero bypasses and that stray `type: ignore`s were removed by fixing types.
✅ Docstrings are thorough and Google-style throughout, well above the 95% interrogate bar in spirit.
✅ Commits follow Conventional Commits (`fix(ingest): …`) with substantive bodies explaining the *why*.
⚠️ I could not execute `./scripts/check-all.sh` in this review sandbox (no `uv`/`ruff`/`mypy` available), so coverage/lint/type-check numbers are taken on trust from the PR description rather than independently verified — recommend confirming CI is green before merge, per the Stay Green workflow.

## Testing
Strong. New tests span: host-timezone matrix (pinned literal ids, not just set-size checks, so a "different but still wrong" derivation can't slip through), an OS-shape fake-stat test that correctly avoids the APFS `st_birthtime`-follows-`os.utime` trap, an e2e duplicate-prevention test for both markdown and (harder, unledgered) documents, a `TOMBING_SOURCES` regression test, and an extensive `pin_ids` suite covering idempotency, conflicts, interrupted-run repair, pre-migration edits, dry-run, and cross-directory execution. The AST guard adds a durable tree-wide net beyond the specific regression tests. No test-quality concerns.

## Documentation
`docs/idempotent-ingest.md` is updated with a clear, complete migration guide (why, what it does/refuses to do, behaviour changes). `file_modified_time`'s docstring correctly explains the invariant it must hold and names both historical failure modes so a future "simplification" is less likely. Docs are consistent with the code.



## Verdict: LGTM

The core bug fix, migration logic, and safety properties (conflict guard, atomic writes, ledger-first crash ordering with self-healing repair, idempotency) are all correct and well-tested; I found no bugs that would cause data loss, duplication, or incorrect behavior. The one gap I found — `unpinned_vault_warning` checking only `01-Fragments` while the migration itself also covers `11-Other-Authors`/`10-Liminal/Orphaned` — is a narrow edge case that would at most cause a missed advisory (not data loss, since the migration itself still correctly finds and pins those fragments when run), so it's a worthwhile follow-up rather than a merge blocker.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5249922266 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1373, opened 2026-08-11
sec(author): `_load_corpus` follows symlinks out of the vault, injecting out-of-tree text into drafted evidence


Geoffe-Ga:
Surfaced by the Gate 2.5 review of #1341 (PR pending). Distinct from #1341, pre-existing, and deliberately left out of that PR's scope.

## The observation

`creek/author/agents.py::_load_corpus` walks `iter_vault_fragments(vault / sub)` for each subtree in `CORPUS_SUBDIRS` with **no containment check** on the resolved file. A markdown file inside the vault that is a symlink to a target outside the vault is loaded as an ordinary fragment.

That matters more here than in the leak gate, which is why it is worth its own issue rather than a note on #1341:

- The **leak gate** (`creek/author/checks.py::_resolve_cited_tiers`) also has no containment check, and #1341 argued explicitly that it does not need one: its fold is monotone (most-restrictive tier wins, all bodies kept), so an extra out-of-tree record can only *raise* a tier or *add* a body — never suppress one. A planted symlink there can at worst cause a false-positive refusal, which is the safe direction.
- **`_load_corpus` is different**: it feeds evidence *into the rendered draft* rather than merely gating it. A symlink planted under `01-Fragments`, `09-Reference` or `11-Other-Authors` pointing at an out-of-tree file injects attacker-controlled text into a specialist's `EvidenceBundle`, and from there into the drafted body. The direction is unsafe, not safe.

## Reachability

Requires write access to the operator's vault directory in order to plant the symlink, so this is not remotely reachable — hence P3 rather than higher. It is still a real trust-boundary gap: the vault is treated as trusted input, and a symlink lets content that was never in the vault present itself as vault content.

## Prior art in the repo

SEC-003 already shipped a containment predicate that should be reused rather than reinvented:

- `creek/redact/cli_commands.py:238` `_assert_no_escaping_symlinks`
- `creek/pipeline.py:173` `class SymlinkEscapedSourceError`

Note the shape mismatch that #1341 recorded when it considered reusing them: `_assert_no_escaping_symlinks` raises `typer.Exit` and takes a Rich `Console`, which is the wrong layer for a pure loader. A shared, presentation-free containment predicate would serve `_load_corpus`, the redact path, and any future vault reader.

## Suggested scope

- Decide whether escaping symlinks in the corpus are skipped (with a DEBUG log, matching how `iter_vault_fragments` already treats unreadable files) or hard-error.
[...]

Issue #1366, opened 2026-08-11
test(gate): the anti-bypass rule is unenforceable in tests/ — 242 suppressions, 0 citing an issue, mypy never looks


Geoffe-Ga:
## Problem

The house anti-bypass rule allows exactly one escape hatch:

> an inline `# noqa: RULE  # Issue #N: <reason>` (or `# type: ignore  # Issue #N: …`) tied to a real tracking issue.

**Nothing enforces that in test code.** `creek-tools/scripts/typecheck.sh:66` runs

```
python -m mypy creek/ creek_mcp/
```

so `tests/` is outside the mypy gate entirely. A `# type: ignore` in a test is
never evaluated in *either* direction — not honoured, not flagged as unused
(despite `warn_unused_ignores = true`), not rejected for missing an issue
reference. It is a bypass token sitting in source that no gate can see.

## Measured, at 93df994

```
creek-tools/tests/  bypass tokens: 65     citing an Issue #N: 0
crawdad/tests/      bypass tokens: 177
                              total: 242  citing an Issue #N: 0
```

(pattern: `# type: ignore` | `# noqa` | `# pylint: disable` | `# pragma: no cover`)

## How it surfaced

PR #1360 (issue #1293) added `# type: ignore[arg-type]` to a new `os.walk` test
spy. It cleared Gate 2 with 12/12 — not because the suppression was justified,
but because the gate never looked. It was caught only by a hand-diff during
orchestration review, and then removed at root cause (the parameters were
annotated `object`, which made the delegation untypeable; `Any` splats into
`os.walk`'s overload set cleanly and needs no suppression).

That is the concerning part: **the only thing between that token and `main` was
a human-equivalent reading the diff.** The next one may not get read.

## This is NOT a request to delete 242 comments

Many are legitimate test idioms — deliberately passing a wrong type to exercise
validation, for instance, genuinely needs `# type: ignore[arg-type]`. The defect
is that the rule is unenforceable there, not that every existing token is wrong.

## Suggested shape

1. **Make the gate see them.** Either extend mypy to `tests/` (expect real work
   — 65 + 177 sites, and the suites lean on duck-typed doubles), or add a cheap
   lint step that greps test trees for bypass tokens and fails any that lack an
   `Issue #N` reference. The second is far cheaper and directly enforces the
   rule as written.
2. **Triage the existing 242** into: legitimate-and-annotate-with-a-standing
[...]

Issue #1362, opened 2026-08-11
refactor(author): promote one shared "privacy_compliance" constant — four duplicated sites gate whether protected text is withheld


Geoffe-Ga:
## Problem

The string `"privacy_compliance"` — the dimension name the HARD privacy gate is
keyed on — is now duplicated across four sites:

1. `creek/author/models.py` — the `FindingDimension` Literal (the source of truth)
2. `creek/author/checks.py` — the emitter
3. `creek/cli.py` — `_PRIVACY_DIMENSION`, added by PR #1352 (issue #1310)
4. `creek_mcp/tools/author.py` — `_PRIVACY_DIMENSION`, added by PR #1361 (issue #1353)

Sites 3 and 4 are **module-local private constants that deliberately do not
import each other**. That was the right call at the time: an MCP tool reaching
into `creek.cli` internals would be a worse coupling than a duplicated literal,
and each lane was editing a file the other was actively reading.

## Why it matters now

Both PRs have landed, so the coordination reason for the duplication is gone.
What remains is a real drift hazard: these four sites gate *whether protected
text is withheld*. If the dimension is ever renamed, three of the four move and
the fourth silently stops matching — which fails **open**, shipping the body.
No test would catch it, because each site's tests construct their own findings.

## Suggested shape

Promote one public constant into `creek/author/checks.py` (next to the emitter,
which is where the value is authoritative in practice) and have `creek/cli.py`
and `creek_mcp/tools/author.py` import it. Leave the `FindingDimension` Literal
in `models.py` as the type, and derive or cross-check the constant against it so
a rename cannot pass type-checking while breaking the gate.

Add one test asserting the constant equals a member of `FindingDimension` — that
is the assertion that makes the rename loud instead of silent.

## Provenance

Both lanes independently identified this cleanup and both deliberately declined
to do it, for the same reason: it would have edited a file the sibling lane was
mid-flight on. Filed by the orchestrator once both merged.

PR #1374, opened 2026-08-11
fix(author): police every corpus subtree with the HARD privacy leak gate

### LINKED ISSUE #1341: fix(author): the HARD privacy leak gate resolves cited tiers only under 01-Fragments, so 09-Reference and 11-Other-Authors are unpoliced
**Labels:** `de-slop`, `creek-pipeline`, `security`, `priority-high`
**Severity:** High
**Remediation:** fix in place

## Problem

`_resolve_cited_tiers` (`creek/author/checks.py:198`) resolves cited fragment
tiers by rglobbing `vault / "01-Fragments"` only. The specialists gather from
**three** subtrees — `agents.py:61`:

```python
_CORPUS_SUBDIRS: tuple[str, ...] = ("01-Fragments", "09-Reference", "11-Other-Authors")
```

A cited fragment from `09-Reference/` or `11-Other-Authors/` is therefore never
resolved, and `check_privacy_compliance` silently skips it. Family 0
(correctness) + family 10 (security).

**Current state:** the HARD privacy leak gate cannot see two thirds of the
corpus it is meant to police. The docstring frames the skip as "A cited fragment
id with no matching file in the vault is unresolvable (its tier cannot be known)
and is skipped rather than guessed" (`checks.py:288-290`) — but the file *does*
exist; the gate just doesn't look there. So an INTIMATE fragment from
`11-Other-Authors/` whose verbatim body is in the draft passes the gate.

## Evidence (corroboration)

**Signal 1 — structural, verified on this tree.**
```
$ sed -n '198p' creek-tools/creek/author/checks.py
    fragments_root = vault / "01-Fragments"
$ grep -n "_CORPUS_SUBDIRS" creek-tools/creek/author/agents.py
61:_CORPUS_SUBDIRS: tuple[str, ...] = ("01-Fragments", "09-Reference", "11-Other-Authors")
113:    for sub in _CORPUS_SUBDIRS:
```
`rg "09-Reference" creek-tools/creek/author/` shows the string appears only in
`agents.py` — never in `checks.py`.

Corroborating internal inconsistency: `fragment_tier_map` (`agents.py:120`),
which drives the #661 Intimate-never-cloud routing decision, is built from the
**full** `_load_corpus`. So the router sees an INTIMATE `09-Reference` fragment
and reroutes locally, while the leak gate is blind to the same fragment — the two
halves of the privacy design disagree.

Repro: seed `<vault>/11-Other-Authors/x/frag-b.md` with `privacy_tier: intimate`
and a distinctive body; build an `EvidenceBundle` citing `frag-b`; call
`check_privacy_compliance(body_containing_that_text, evidence, vault,
MediumContract(medium="research", default_privacy_tier=PrivacyTier.OPEN))`.
Expected: one HIGH `privacy_compliance` finding. Actual: `[]`. Seeding the same
fragment under `01-Fragments/` produces the finding, which isolates the cause to
the directory, not the content.

**Signal 2 — reading/intent.** The check is documented as a HARD gate
("Citation completeness and privacy compliance are HARD gates for the research
medium", `checks.py:7`) and is fail-closed everywhere else —
`_MOST_RESTRICTIVE_RANK` exists precisely so an unknown tier trips the gate
rather than passing. Skipping a whole subtree defeats that design in the one
direction the module took pains to avoid. Reachable via
`creek author --include-tier all` (or `personal`), an audited operator decision.

Distinct from #1310 (the CLI builds the desk with no medium contract at all);
this defect persists even once a contract is passed.

## Scope

`_resolve_cited_tiers`' search root. Does not change the tier ranking or the
verbatim-leak heuristic.

## Tasks

1. Make `_resolve_cited_tiers` (`creek/author/checks.py:198`) search the same
   subtree set the specialists gather from — import `_CORPUS_SUBDIRS` rather
   than re-declaring it, so the two cannot drift again.
2. Correct the docstring at `checks.py:288-290`: an unresolvable id means the
   file is genuinely absent, not that the gate declined to look.
3. Add a test seeding an INTIMATE fragment in `09-Reference/` and one in
   `11-Other-Authors/`, asserting a HIGH `privacy_compliance` finding for each.
4. Add a guard test that fails if `checks.py` and `agents.py` ever name
   different corpus roots.

## Acceptance Criteria

- [ ] A cited INTIMATE fragment in any of the three corpus subtrees trips the
      HARD leak gate.
- [ ] `checks.py` and `agents.py` share one corpus-root definition.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/author/checks.py` | Modify |
| `creek-tools/creek/author/agents.py` | Modify |
| `creek-tools/tests/test_reflection.py` | Modify |

Geoffe-Ga:
## Summary

The HARD `privacy_compliance` leak gate resolved cited fragment tiers out of `01-Fragments` **only**, while the Writing Desk specialists gather evidence from three subtrees. A cited over-tier fragment living in `09-Reference` or `11-Other-Authors` resolved to nothing, so the gate never learned its tier and a draft could reproduce its protected body verbatim under `verdict=PASS`.

Reproduced by execution before the fix — same INTIMATE fragment, same id, same body reproduced verbatim in the draft, `MediumContract(medium="research", default_privacy_tier=OPEN)`; only the subtree differs:

| subtree | `_resolve_cited_tiers` | findings |
|---|---|---|
| `01-Fragments` | `['frag-b']` | `['privacy_compliance']` |
| `09-Reference` | `[]` | `[]` |
| `11-Other-Authors` | `[]` | `[]` |

The corroborating inconsistency, also measured: `fragment_tier_map(v, INTIMATE)` returned `{'frag-r': 'intimate', 'frag-o': 'intimate'}` for the very two fragments the gate returned `[]` for — the #661 router and the leak gate genuinely disagreed about the same vault.

### What changed

- **One corpus definition.** `CORPUS_SUBDIRS` now lives in `creek/vault/reader.py`; `checks.py` and `agents.py` both import that object. While they were two literals the gate fell a subtree behind the specialists it polices.
- **The walk covers all three roots**, with a **per-root** `is_dir()` skip. The old global `if not fragments_root.is_dir(): return resolved` became an outright bug with three roots — a vault with no `01-Fragments` but a populated `09-Reference` had no privacy gate at all.
- **Duplicate ids fold through a frozen `_CitedFragment`**: most-restrictive tier wins, and **every** body is kept at every rank. The fold is monotone and commutative, so walk order provably cannot change the verdict. Keeping only the winning rank's body would reopen the hole one rank down.
- **The first-hit `break` is gone** — you cannot stop at the first match and also take the most restrictive.
- Docstring corrections: the module docstring claimed the check uses `iter_vault_fragments` (it uses `try_load_fragment`), and the resolver's docstring sold an early exit the code no longer has. `docs/writing-desk.md` now states which subtrees the gate polices.

### Behaviour change an operator will notice

This makes the gate see **more**, and on this base (#1352, #1361, both merged) a seen finding actually withholds: `creek author` withholds the body and exits non-zero, and the MCP `creek.author` verb answers `status: "refused"`. **So drafts that previously shipped can now be refused.** That is the intended direction, but it is user-visible.

A draft is newly refused only when all of these hold: it cites id X; X resolves from `09-Reference` or `11-Other-Authors` (or is duplicated across subtrees with an above-ceiling twin); X's true tier ranks above the contract ceiling (`open` in every shipped medium template); and the draft reproduces >= 4 words of one of X's stored bodies verbatim. Reaching it also requires an elevated `--include-tier personal|intimate|all`, since the default OPEN admission keeps above-OPEN content out of the evidence to begin with.

**Measured, on the only corpus available: zero.** The full suite is 8902 passed / 5 skipped, so no previously-passing draft flipped. 63 fixture sites across the other suites seed those two subtrees, and every one relies on `write_raw_fragment_file`'s `privacy_tier="open"` default (no explicit override on any of them), so none is above the ceiling.

**I cannot give a number for a real operator vault** — no vault exists on this host (no `01-Fragments` / `09-Reference` / `11-Other-Authors` tree anywhere under the repo). The count is a property of vault content, not of the code. To get the upper bound on your own vault — the population of newly gate-visible fragments, of which only the *cited and verbatim-reproduced* ones can refuse a draft:

```python
from creek.vault.reader import CORPUS_SUBDIRS, iter_vault_fragments
from creek.models import PrivacyTier
RANK = {PrivacyTier.OPEN: 0, PrivacyTier.PERSONAL: 1,
        PrivacyTier.INTIMATE: 2, PrivacyTier.UNCLASSIFIED: 3}
vault = pathlib.Path("<your vault>")
newly = [f.id for sub in CORPUS_SUBDIRS if sub != "01-Fragments"
         for _p, f, _b, _m in iter_vault_fragments(vault / sub)
         if RANK.get(PrivacyTier(f.privacy_tier), 3) > 0]
print(len(newly), newly[:20])
```
(Verified against a synthetic vault with a known answer: 7 fragments in corpus, 3 newly visible — exact match.)

### Deliberately left alone

- The `if vault is None:` guard, the `_NO_CONTRACT_CEILING` branch, and the `check_privacy_compliance` docstring paragraph — all shipped by #1352. The issue body's "Task 2" was already delivered there; rewriting it would re-introduce a merged privacy hole.
- `fragment_tier_map`'s last-wins duplicate rule. It feeds `conductor.py::_content_tier` and therefore the ModelRouter's Intimate-never-cloud decision. The divergence from the gate's most-restrictive rule is now documented in its docstring and pinned by a named test rather than left as a silent assumption.
- `_PRIVACY_DIMENSION`, duplicated privately in `creek/cli.py` and `creek_mcp/tools/author.py` — that cleanup is #1362.
- **No symlink-containment predicate**, on purpose: under a monotone fold an extra record can only raise a tier or add a body, never lower or remove one, so a planted out-of-tree shadow cannot downgrade or mask anything. Adding a skip path would introduce the only genuinely unsafe move available — dropping a body, and with it a leak. Residual: out-of-tree content reachable by symlink can cause a false-*positive* refusal, which is the safe direction.
- **No cache**, on purpose. The walk widened from one subtree to three and the early exit is gone, so the common miss path — an LLM inventing a fragment id — is now always exhaustive, once per `review` call up to `max_author_rounds` (default 3, capped at 10). A cache with wrong invalidation on a HARD leak gate is worse than a slow gate; if it bites, that is a measured follow-up.

## Test plan

**Proved RED before the fix** (the repo's standing rule). 15 node ids, 14 failing on a behavioural assertion and 1 on the deliberately-absent constant:

```
tests/test_reflection.py::test_privacy_gate_sees_over_tier_fragment_in_every_corpus_subtree[09-Reference-intimate]
tests/test_reflection.py::test_privacy_gate_sees_over_tier_fragment_in_every_corpus_subtree[09-Reference-personal]
tests/test_reflection.py::test_privacy_gate_sees_over_tier_fragment_in_every_corpus_subtree[09-Reference-absent-privacy-tier-key]
tests/test_reflection.py::test_privacy_gate_sees_over_tier_fragment_in_every_corpus_subtree[11-Other-Authors-intimate]
tests/test_reflection.py::test_privacy_gate_sees_over_tier_fragment_in_every_corpus_subtree[11-Other-Authors-personal]
tests/test_reflection.py::test_privacy_gate_sees_over_tier_fragment_in_every_corpus_subtree[11-Other-Authors-absent-privacy-tier-key]
tests/test_reflection.py::test_duplicate_cited_id_resolves_to_the_most_restrictive_tier[strict-last]
tests/test_reflection.py::test_duplicate_cited_id_at_equal_tier_cannot_mask_the_leaking_body[decoy-first]
tests/test_reflection.py::test_duplicate_cited_id_below_the_winner_is_still_body_checked
tests/test_reflection.py::test_privacy_gate_polices_a_vault_with_no_01_fragments_dir
tests/test_reflection.py::test_leak_gate_reads_the_true_tier_while_the_router_reads_the_admitted_one
tests/test_vault_reader.py::test_the_leak_gate_and_the_specialists_read_one_corpus_definition   (AttributeError — the shared constant does not exist yet)
tests/test_vault_reader.py::test_no_desk_module_hardcodes_a_corpus_root
tests/test_cli_author.py::test_author_withholds_intimate_text_leaked_from_a_non_fragments_subtree[09-Reference]
tests/test_cli_author.py::test_author_withholds_intimate_text_leaked_from_a_non_fragments_subtree[11-Other-Authors]
```

The two CLI cases are the reachability proof through a real surface — `creek author --medium research --include-tier intimate` via Typer's `CliRunner`. Pre-fix stdout read `verdict=PASS rounds=1` with the intimate sentence verbatim and exit code 0.

**Coverage is not intimate-only:** the matrix crosses both subtrees with INTIMATE, PERSONAL, and an absent `privacy_tier` key — the last ranks UNCLASSIFIED, which this gate's table puts at the *top* of the restrictiveness order, so an untiered fragment is over every ceiling.

**Mutation battery — 8 mutants, each guard reverted alone. 7 killed:**

| mutant | killed by |
|---|---|
| M1 walk `01-Fragments` only (the original defect) | 13 tests |
| M2 global early-return when `01-Fragments` absent | 9 tests |
| M3 tier merge last-wins instead of most-restrictive | 1 — `…most_restrictive_tier[strict-first]` |
| M4 keep only the last body seen | 3 tests |
| M5 drop bodies below the winning rank | 1 — `…below_the_winner_is_still_body_checked` |
| M7 gate applies `tier_within_override` (blinds itself) | 29 tests |
| M8 `checks.py` re-forks its own corpus literal | 2 — both forcing functions |

M3 initially **survived**, and closing it found a real gap: every duplicate-id fixture put the strict record in the later-walked subtree, so last-wins gave the right answer for the wrong reason. Measured under the mutant, with the INTIMATE record in `01-Fragments` and its OPEN twin behind it: **zero findings**, with the protected body sitting verbatim in the draft. Fixed by parameterising the test `[strict-last]` / `[strict-first]`.

**One survivor, by design: M6 (unsorted walk), 0 tests fail.** The `_CitedFragment` fold is monotone and commutative, so walk order provably cannot change the verdict — the `sorted()` is reproducibility and debugging defence-in-depth, not a correctness guard, and no behavioural test can kill it without asserting on filesystem order. Determinism is carried by the type, which is the stronger guarantee.

**Router/gate agreement, proven by execution** on a 3-fragment corpus spanning all three subtrees (deliberately more than one element, so no assertion is vacuous):

```
checks.CORPUS_SUBDIRS is reader.CORPUS_SUBDIRS : True
agents.CORPUS_SUBDIRS is reader.CORPUS_SUBDIRS : True
router (#661 fragment_tier_map): {'frag-b': 'intimate', 'frag-o': 'intimate', 'frag-r': 'intimate'}
gate   (_resolve_cited_tiers)  : {'frag-b': 'intimate', 'frag-o': 'intimate', 'frag-r': 'intimate'}
AGREE on id set: True    AGREE on every tier: True
```

**Gate 2: `./scripts/check-all.sh` exit code 0.** 8902 passed, 5 skipped, branch coverage 94.40%, per-file gate passed (256 files >= 80%), 12/12 checks. Complexity *improved*: `_resolve_cited_tiers` B(8) → A(4), nothing in the file above B, module average A(4.73) → A(3.82). Docstring coverage 100% on both changed modules.

No bypasses were added anywhere, including in tests — `tests/` sits outside mypy's scope (#1366), so that sweep was done by hand as well as by the gate.

**Gate 2.5** (`code-review-orchestrator` over the diff): **CLEAN, no blockers.** Its three non-blocking notes were engagement with the three questions I put to it — it agreed M6 is by-design and could not construct an input where two walk orderings differ; it agreed the symlink argument holds *for this gate*; and it agreed the no-cache trade is correctly called. It also caught a stale citation of mine: `use_enum_values=True` for the `Fragment` model is `creek/models.py:762` (the `:518` I first grepped is `MediumContract`), with the `privacy_tier` default at `:801` — the docstring claim itself is correct.

Its one genuinely new observation is filed as **#1373** (P3): `agents.py::_load_corpus` has the same absence of symlink containment, but unlike the gate it feeds evidence *into the rendered draft*, so there the direction is unsafe rather than safe. Pre-existing and out of scope here.

Closes #1341

### DIFF
--- a/creek-tools/creek/author/agents.py
+++ b/creek-tools/creek/author/agents.py
@@ -45,7 +45,7 @@ from creek.models import (
     PrivacyTier,
     VoiceRegister,
 )
-from creek.vault.reader import iter_vault_fragments
+from creek.vault.reader import CORPUS_SUBDIRS, iter_vault_fragments
 
 if TYPE_CHECKING:
     from enum import StrEnum
@@ -56,9 +56,6 @@ if TYPE_CHECKING:
     from creek.models import Fragment
 
 _DimT = TypeVar("_DimT", bound="StrEnum")
-
-#: Vault subtrees the specialists draw evidence from.
-_CORPUS_SUBDIRS: tuple[str, ...] = ("01-Fragments", "09-Reference", "11-Other-Authors")
 
 #: Obsidian wikilink target, ignoring any ``|alias`` or ``#heading`` suffix.
 _WIKILINK = re.compile(r"\[\[([^\]|#]+)")
@@ -110,7 +107,7 @@ def _load_corpus(
     ``override`` of ``None`` defaults to ``OPEN`` (the most restrictive).
     """
     records: list[tuple[Fragment, str]] = []
-    for sub in _CORPUS_SUBDIRS:
+    for sub in CORPUS_SUBDIRS:
         for _path, fragment, body, _meta in iter_vault_fragments(vault / sub):
             if tier_within_override(fragment.privacy_tier, override):
                 records.append((fragment, body))
@@ -126,6 +123,20 @@ def fragment_tier_map(
     Built from the same override-filtered corpus the specialists gather from, so
     the desk can derive a run's *content tier* (the most-sensitive tier actually
     in the evidence) and route the voice call accordingly.
+
+    This is the **admitted** view of the corpus, and it diverges from the leak
+    gate's on purpose. Here an id seen twice resolves last-wins in
+    :data:`~creek.vault.reader.CORPUS_SUBDIRS` order over records that already
+    passed ``tier_within_override``, because the question is "what is in the
+    evidence this run may use". The HARD gate at
+    :func:`creek.author.checks._resolve_cited_tiers` reads the *unfiltered*
+    corpus and resolves most-restrictive-wins, because its question is "what is
+    the true tier of the text this draft reproduces" — a gate that only saw the
+    admitted view would be blind to exactly the content it exists to catch.
+    The divergence is deliberate, and pinned by
+    ``test_leak_gate_reads_the_true_tier_while_the_router_reads_the_admitted_one``
+    in ``tests/test_reflection.py``, so a future "make these agree" cleanup has
+    to argue with a named assertion rather than a silent assumption.
 
     Args:
         vault: Vault root.
--- a/creek-tools/creek/author/checks.py
+++ b/creek-tools/creek/author/checks.py
@@ -7,15 +7,20 @@ records. The checks are *deterministic* — no LLM — so the mutation tests in
 produces. Citation completeness and privacy compliance are HARD gates for the
 research medium; the rest are softer rubric divergences.
 
-The privacy check reuses the vault fragment loader
-(:func:`creek.vault.reader.iter_vault_fragments`) to resolve each cited
-fragment's :class:`~creek.models.PrivacyTier`; the voice check reuses the
-FEAT-040.x AI-style scanner (:func:`creek.generate.ai_style.scanner.scan`).
+The privacy check resolves each cited fragment's
+:class:`~creek.models.PrivacyTier` with
+:func:`creek.vault.reader.try_load_fragment`, walking the corpus subtrees
+file by file rather than through :func:`~creek.vault.reader.iter_vault_fragments`
+(or the desk's own ``_load_corpus``), both of which materialise every fragment
+in the vault into a list before the caller sees the first record. The voice
+check reuses the FEAT-040.x AI-style scanner
+(:func:`creek.generate.ai_style.scanner.scan`).
 """
 
 from __future__ import annotations
 
 import re
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import yaml
@@ -34,7 +39,7 @@ from creek.models import (
     PHASE_LEGACY_ALIASES,
     PrivacyTier,
 )
-from creek.vault.reader import try_load_fragment
+from creek.vault.reader import CORPUS_SUBDIRS, try_load_fragment
 
 if TYPE_CHECKING:
     from pathlib import Path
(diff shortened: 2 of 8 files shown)


Geoffe-Ga:
## Summary
Fixes `#1341`: the HARD `privacy_compliance` leak gate in `checks.py::_resolve_cited_tiers` only ever resolved cited fragment tiers out of `01-Fragments`, while the Writing Desk specialists (`agents.py::_load_corpus`) gather evidence from `01-Fragments`, `09-Reference`, and `11-Other-Authors`. A cited over-tier fragment filed under either of the latter two subtrees resolved to nothing, so a draft could reproduce its protected body verbatim and still review `PASS`. The fix promotes the subtree list to a single shared `CORPUS_SUBDIRS` constant in `creek/vault/reader.py`, widens the gate's walk to all three roots with a per-root (not global) `is_dir()` skip, and folds duplicate fragment ids across subtrees via a new frozen `_CitedFragment` (most-restrictive tier wins, every body is retained).

## Strengths
- **Correct, minimal root-cause fix**: one shared `CORPUS_SUBDIRS` (`creek/vault/reader.py`) replaces two independently-maintained literals in `agents.py` and `checks.py`, which is exactly what let the gate drift behind the specialists it polices.
- **Forcing-function test**: `test_no_desk_module_hardcodes_a_corpus_root` does an AST sweep of every module under `creek/author/` for a hardcoded corpus-root string, so this class of drift can't reappear silently. `test_the_leak_gate_and_the_specialists_read_one_corpus_definition` asserts object *identity*, not equality, which is the right assertion for a shared-constant contract.
- **The merge semantics are proven safe, not just tested**: the `_CitedFragment` fold (rank-max tier, append-if-absent bodies) is monotone and commutative, so walk order provably cannot change the verdict — and the one deliberately-unkilled mutant (M6, unsorted walk) is explained rather than left as unexplained residual risk.
- **Real regression caught during development**: the PR body documents that the first version of the duplicate-id fix passed all fixtures for the wrong reason (every fixture happened to put the strict record later in walk order) until the test was parameterized both ways (`strict-first`/`strict-last`) — good self-skepticism, and the resulting test is in the diff.
- **Correctly scoped**: explicitly declines to touch `fragment_tier_map`'s differing (last-wins, admitted-view) semantics, and instead pins the deliberate divergence with a named test (`test_leak_gate_reads_the_true_tier_while_the_router_reads_the_admitted_one`) so a future "make these agree" refactor can't silently reopen the hole.
- Docstring corrections (module-level and `_resolve_cited_tiers`) fix two previously-false claims (`iter_vault_fragments` vs `try_load_fragment`; a since-removed early exit).

## Security Concerns
🟢 LOW — none introduced; this PR closes an existing HIGH-severity leak-gate blind spot. The remaining known gap (`agents.py::_load_corpus` lacks symlink-containment checking, and unlike the gate that direction is unsafe since it feeds evidence *into* the rendered draft) is correctly identified as pre-existing and out of scope, and is tracked as a separate follow-up (#1373) rather than silently left undocumented.

## Problems
None found blocking merge.

## Code Quality
- Non-blocking: `check_privacy_compliance` now performs an unsorted-then-sorted `rglob` walk of all three corpus subtrees on every call, once per `review` round (up to `max_author_rounds`, capped at 10), with no cache. This is explicitly a deliberate tradeoff (a cache with wrong invalidation on a HARD leak gate is worse than a slow gate) and is reasonable for a security-critical path, but worth watching if vault sizes grow large enough for `author` latency to matter in practice.
- Non-blocking: `_CitedFragment.merged_with` dedups bodies by exact string equality rather than the normalized form `_is_verbatim_leak` eventually uses — harmless (just a few extra near-duplicate entries in `bodies`), not a correctness issue.

## CLAUDE.md Adherence
✅ TDD followed and demonstrated (RED node ids listed, failing for the right reasons before the fix).
✅ `./scripts/check-all.sh` reported exit 0: 8902 passed / 5 skipped, branch coverage 94.40%, per-file gate passed, complexity improved (`_resolve_cited_tiers` B(8)→A(4)), docstring coverage 100% on changed modules.
✅ Conventional commit message, single feature branch (`issue/1341-issue`), no direct-to-main commit.
✅ Visible completed CI jobs (Integration & E2E, CrawDad Quality ×3, Code Complexity, Build Distribution) are green; several jobs (Static Analysis, Pylint, Tests & Type Checking ×3, claude-review) were still in progress at review time — worth confirming they land green before merge, though nothing in the diff suggests they won't.

## Testing
Extensive and well-targeted:
- Parametrized coverage crossing both previously-blind subtrees (`09-Reference`, `11-Other-Authors`) with three tier shapes (`intimate`, `personal`, and an absent `privacy_tier` key resolving `unclassified`), confirming the gate isn't intimate-only.
- Duplicate-id fold tested from both walk-order directions, at unequal and equal tiers, and confirms the below-winner body is still leak-checked (not just the winning record's body).
- Reachability proof through the real CLI surface (`test_author_withholds_intimate_text_leaked_from_a_non_fragments_subtree`), not just the internal resolver.
- Mutation testing: 8 targeted mutants against the specific guards this diff introduces, 7 killed, 1 (unsorted walk) provably unkillable by a behavioral test given the monotone/commutative fold — a good use of mutation testing to validate design invariants rather than just line coverage.

## Documentation
`docs/writing-desk.md` updated to state which subtrees the gate polices and the most-restrictive-tier/keep-all-bodies duplicate-id semantics. Module and function docstrings in `checks.py`/`reader.py` are unusually thorough and explain *why* (e.g., why no symlink containment check, why no cache), which will help future maintainers avoid reintroducing the original bug.



## Verdict: LGTM

This is a well-scoped, correctly-targeted security fix for a real HARD-gate blind spot (#1341), backed by RED-before-fix tests, an 8-mutant kill matrix (7/8 killed, the 1 survivor explained as provably unkillable by design), and a passing local check-all.sh; the visible CI jobs that have completed are all green and nothing in the diff introduces a new correctness or security gap. No blocking issues found.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5249428910 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1371, opened 2026-08-11
fix(api): the published OpenAPI document declares no bearer requirement, though every /v1 route enforces one


Geoffe-Ga:
`creek_mcp/api/openapi.py` generates the published `/v1` document, and that document declares no
`components.securitySchemes` and no `security` block, at either the document or the operation
level. `_parameters()` publishes exactly two headers — `X-Creek-Contract-Version` (on the three
routes that require it) and `X-Creek-Tier-Ceiling` — and `Authorization` appears nowhere.

Meanwhile `BearerAuthMiddleware` is item 5 of the seven-layer stack, **above the router**, so
every one of the five published routes is unreachable without a bearer token. There is no
anonymous access at all: `build_verifier` refuses to construct a server without configured
consumer tokens.

So the document says the surface is open and the server refuses everything. The consequences are
concrete rather than cosmetic:

- **Generated clients ship without an auth parameter.** The whole reason the route table is data
  and the document is generated from our own models rather than a framework's introspection is
  that a consumer generates a client from it. A client generated from this document has no way to
  present a credential and gets `401` on every call.
- **`docs/api.md` and the document disagree**, and the document is the machine-readable one.
- `info.description` already says "Authenticated HTTP/JSON access…", so the omission reads as an
  oversight rather than a decision.

**Fix:** add a `bearerAuth` (`type: http`, `scheme: bearer`) entry to `components.securitySchemes`
and a document-level `security` requirement, derived — like everything else in that module — from
a constant rather than a literal. `creek_mcp.api.routes.AUTHORIZATION_HEADER` was moved into the
framework-free `routes.py` module by #1129 and is the natural source; that move was made so the
header the auth middleware *reads* and the header the response `Vary` *names* cannot drift, and
the OpenAPI generator would be its third reader.

Note the golden: `tests/test_v1_api_openapi.py` byte-compares
`docs/contracts/adepthood-v1/schemas/*.schema.json`, which are component **schemas** only. Adding
a security scheme touches `components.securitySchemes`, not `components.schemas`, so no committed
schema file changes — but check that assumption rather than assuming it.

Adding a security declaration documents a requirement the server has always enforced, so no wire
shape changes and no contract minor moves.

Found while implementing #1129.

Issue #1370, opened 2026-08-11
fix(api): a fault in the outermost /v1 layer escapes as Starlette's bare text/plain 500


Geoffe-Ga:
`ErrorBoundaryMiddleware` is the second layer of the `/v1` stack and turns any fault below it into
the published `500 internal_error` envelope, logged once on `creek_mcp.httpapi.error`. It cannot
catch a fault in the layer *above* it, and there is one reachable line up there.

`AccessLogMiddleware.__call__` calls `bind_context(scope)` **before and outside** its own `try:`,
and `AccessLogMiddleware` is the outermost layer. A fault there escapes the whole Creek stack into
Starlette's `ServerErrorMiddleware`, which `build_middleware_stack` installs unconditionally and
which `create_app` registers no handler for — so `handler is None` and the default
`PlainTextResponse("Internal Server Error", 500)` applies.

Demonstrated in-tree by patching `creek_mcp.httpapi.middleware.access_log.bind_context` to raise
(patching it on `creek_mcp.httpapi.context` does *not* reproduce it — `access_log` binds the name
at import):

```
status: 500   content-type: text/plain; charset=utf-8   body: b'Internal Server Error'   vary: ABSENT
```

Control, with the fault raised *below* the boundary: `500`, `application/json`, the proper
`ErrorEnvelope`, `vary: X-Creek-Tier-Ceiling`.

Four contract violations at once, and they are the same four `_not_a_routing_miss` was written to
refuse when it declined to fall back to Starlette's default handler: the wrong envelope,
`text/plain` where the contract speaks only JSON, an author-written string on the wire where the
contract guarantees only constants from `ERROR_MESSAGES`, and no standing headers. There is also
no `request_id`, so the caller has nothing to quote and the fault appears in no access line.

**Pre-existing**, not introduced by #1129 — the `Vary` half has been true since the stack was
built, and #1129 only adds a second header to the list of things this path omits.

**Possible fixes, neither obviously right:**
- Move `bind_context` inside a `try:` in `AccessLogMiddleware` — but then the context the whole
  stack reads is the thing that failed, and `context_of` is deliberately intolerant of a missing
  entry (a synthesised replacement "would paper over a reordering by handing the client a
  correlation id that appears in no log line").
- Register a `500`/`Exception` handler on the `Starlette` app so `ServerErrorMiddleware` renders
  the published envelope. This needs care: the handler runs with no `RequestContext`, which is the
[...]

Issue #1369, opened 2026-08-11
fix(api): a trailing slash on any /v1 path answers 307, outside the published status set and past the version gate


Geoffe-Ga:
`Starlette`'s router is built with `redirect_slashes=True` and `creek_mcp.httpapi.app.create_app`
never disables it, so a trailing slash on any published path is answered by a `RedirectResponse`
that Starlette constructs directly — never by `creek_mcp.httpapi.errors.json_response`.

Observed in-tree against the real app with a valid bearer and `follow_redirects=False`:

| request | result |
|---|---|
| `GET /v1/health/` | `307`, `location: http://testserver/v1/health`, no `vary`, no `content-type`, 0-byte body |
| `GET /v1/wheel/` | `307`, same shape |
| `PUT /v1/journal-entries/abc/` | `307`, `location: …/v1/journal-entries/abc` |
| `GET /v1/health/` with no bearer | `401` with the standing `Vary` — auth *does* gate it |

Three defects ride together, and the middle one is the serious one.

**`307` is outside the published status set.** The contract publishes
`{200, 401, 403, 404, 409, 422, 500, 501, 503}`, and `creek_mcp/httpapi/app.py` goes to
deliberate lengths to keep `405` out of it — `METHOD_NOT_ALLOWED` is spelled as a literal
precisely because "the contract has **no** `ErrorCode` for it", and `_routing_miss` renders a
verb mismatch as `404` so that a conforming client, which maps an out-of-set status to
"unreachable", does not lose the vault over a wrong method. A trailing slash loses it the same
way, by the same reasoning, through a path that argument never reached.

**The contract-version gate is bypassed.** `GET /v1/wheel/` with no `X-Creek-Contract-Version`
header at all, and with a stale `0.0`, both return `307` rather than `409 incompatible_version`.
The gate lives in `_endpoint_for`'s endpoint wrapper (`app.py`), which is *below* the router, and
the redirect is issued by the router itself. So a client speaking a version this server does not
serve is told to retry a URL rather than to renegotiate — the exact wrong answer the version gate
was placed above the capability gate to avoid.

**No standing headers.** The `307` carries neither `Vary` nor (as of #1129) `Cache-Control`,
because both are stamped by the single builder it never enters. This is the least of the three:
`307` is not heuristically cacheable under RFC 9110 §15.1 and the body is empty, so there is
nothing for a cache to serve wrongly. It is listed because it falsifies the unqualified sentence
[...]

PR #1376, opened 2026-08-11
fix(api): refuse to let a cache keep, or misfile, a /v1 answer

### LINKED ISSUE #1129: fix(api): send Cache-Control: no-store on authenticated /v1 responses

Geoffe-Ga:
## Summary

`/v1` bodies are computed from two things a cache cannot see in the URL — the declared tier
ceiling and the authenticated consumer — and until now the only thing stopping an intermediary
reusing one caller's body for another was a single `Vary` token. This adds the second token and
the directive that says not to store it at all:

```
Vary: X-Creek-Tier-Ceiling, Authorization
Cache-Control: no-store
```

Both are stamped by the one builder, `creek_mcp.httpapi.errors.json_response`, which is what makes
them unconditional rather than something each of the six call sites has to remember. #1128 (PR
#1365) chose **union** over discard for `Vary` explicitly so that this issue could add a token;
that is the two-line half of this change. The rest is the half #1365 said its shape did not
provide.

### The exposure is the `404`, not the `401`

Intersect the published status set `{200, 401, 403, 404, 409, 422, 500, 501, 503}` with the
statuses RFC 9110 §15.1 enumerates as *heuristically cacheable* `{200, 203, 204, 206, 300, 301,
404, 405, 410, 414, 501}` and the whole of it is **`{200, 404, 501}`** (verified against
`ERROR_STATUS`, not asserted). The routing `404` is reachable on any unrouted `/v1` path and
carries no freshness information, so a *conforming* shared cache may store and reuse it on
heuristic freshness alone. That is the one case here that is not merely prophylactic. The argument
is deliberately **not** rested on a cached `401` — that status is not on the heuristic list, so
that story needs a cache already violating the spec, and an argument that only holds against a
broken cache is one a reviewer can wave away.

### The two headers are independent, not one doubled

The issue's premise is a *misconfigured* cache — precisely the one that may ignore `no-store`. For
that cache the remedy is not a louder instruction but a correct cache **key**, and `Vary:
Authorization` is that key. Drop the directive and a compliant cache may store; drop the token and
a non-compliant one may mismatch. Neither subsumes the other, so a later "simplification" that
deletes one is a regression — the module docstring says so, to make that harder.

### Unconditional, not conditional on authentication

The title says "authenticated `/v1` responses"; stamping unconditionally is the only implementable
reading, and a strictly stronger one.

- `json_response` takes no `RequestContext`, so conditioning on `context.consumer` means threading
  a context through the builder and all six call sites, destroying the context-freedom that lets
  it be unit-tested as a pure function. The cheap alternative — stamping in `error_response` only
  — covers every refusal and no `200`, the exact inverse of what was asked.
- The condition would have nothing to decide. `BearerAuthMiddleware` is item 5 of the seven-layer
  stack, **above the router** (verified by building the stack, not by reading the comment), so all
  five routes are authenticated by construction.
- Where it *would* differ is on a route that does not exist yet: one mounted above the gate would
  silently stop being stamped, because such a condition is keyed on the stack's *order* rather
  than on anything the builder can see. The unconditional form cannot fail that way.

Cache-safety is a one-way ratchet, and every change here moves the safe direction: a token is
added to `Vary` (monotone under RFC 9111 §4.1 — reuse is a conjunction over the nominated names,
so more tokens can only shrink the set of requests an entry may serve) and a directive is added
that forbids storage outright. Nothing is made more cacheable and no `Vary` set shrinks.

### Two merge rules, deliberately not one mapping

`Vary` **unions** a caller's tokens; `Cache-Control` **replaces** them. #1365 declined a
`STANDING_HEADERS: Mapping[str, str]` for exactly this reason and the reasoning still holds — a
mapping would have to carry the rule as well as the name (`Mapping[str, MergeRule]`), a strictly
more complex type than two named constants and two named rules. What *is* shared is the mechanism:
`_folds_to(name, standing)` matches a header name case-folded for both, because `Vary` and `vary`
are one header on the wire and two keys in a dict.

Dropping a caller's `Cache-Control` whole is **not** directive-wise monotone, and the docstring
says so rather than claiming otherwise: `private` and `no-cache` are subsumed by `no-store`, but
`no-transform` and `must-understand` are not. It is chosen anyway, because preserving them means
parsing a header whose members take `=`-valued arguments and then ruling on whether `max-age=600`
contradicts `no-store` — a directive parser with a contradiction policy is a larger thing to get
wrong than the one directive it would rescue. No production call site passes a `Cache-Control`
today, so nothing pays for it. Prepending was rejected on evidence rather than taste: `no-store,
max-age=600` is self-contradictory, RFC 9111 does not say which half wins, and real caches resolve
it inconsistently.

`AUTHORIZATION_HEADER` moves to `creek_mcp/api/routes.py`, beside the two other published header
names. `auth.py` imports `errors.py`, so `errors.py` cannot import it back — but the real reason is
the invariant: the header the auth middleware **reads** and the header the `Vary` **names** must be
the same string, or the `Vary` declares a dependency on something no request carries.

## The P2-vs-P1 claim in the body does not hold

The issue is labelled **P1**; its body argues P2. I checked both clauses in code and left the label
alone.

| Clause | Verdict |
|---|---|
| "the default bind is loopback" | **True** — `DEFAULT_API_HOST = "127.0.0.1"` (`httpapi/cli.py:68`) |
| "a non-loopback bind already requires TLS" | **True** — `require_transport_confidentiality` (`transport_posture.py:89`), called at `cli.py:262` |
| "so there is no supported deployment today that puts a shared cache in the path" | **False** |

The conclusion does not follow from its premises. `docs/api.md` already documents fronting this
server with a reverse proxy as a supported deployment ("A reverse proxy, load balancer, or TLS
terminator placed in front of this server…"), and the TLS guard's own remediation text *recommends*
it: "Bind 127.0.0.1, terminate TLS in a reverse proxy". A reverse proxy is the canonical home of a
shared cache. Separately, `no-store` also governs **private** caches — a consumer's own HTTP client
writing tier-ceilinged vault content to disk — which a loopback bind does not mitigate at all.

**P1 is right; the body's own reasoning is what is wrong.** Not relabelled — flagging, not
deciding.

## Per-route decision

All five, because there is no unauthenticated route to spare:

| Route | Authenticated? | Stamped? |
|---|---|---|
| `GET /v1/capabilities` | Yes — below the bearer gate | Yes; body carries vault readiness |
| `PUT /v1/journal-entries/{external_id}` | Yes | Yes |
| `POST /v1/reflections` | Yes | Yes |
| `GET /v1/wheel` | Yes | Yes; ceiling-filtered aggregate |
| `GET /v1/health` | Yes | Yes |

`/v1/health` and `/v1/capabilities` look like the exceptions and are not — the bearer gate sits
above the router, so neither is reachable anonymously. Health is also the case where `no-store` is
affirmatively *correct* rather than a tolerated cost: a cached liveness probe reports "ok" for a
dead server, which is worse than no probe. No route loses a cache it was getting any value from.

`Vary: Authorization` **does** belong on unauthenticated responses too, and this is pinned by
`test_the_unauthenticated_refusal_carries_both_standing_tokens_and_no_store`. It reads backwards —
the caller sent no credential, so what can the answer vary on? — but reuse runs both directions: a
stored `401` that does not name `Authorization` may be served to a caller who *did* present a valid
token (a denial of service wearing the costume of a credential failure, undebuggable because the
token was never looked at), and a stored `200` may be served to one who presented nothing.

## Scope: two paths this does not reach

Both verified against the real app and filed rather than fixed here, and every claim in the new
prose is scoped to "every response **this builder constructs**" because of them:

- **#1369** — a trailing slash on any `/v1` path is answered by Starlette's router with a `307`
  before any handler runs: outside the published status set, past the contract-version gate, and
  carrying neither header.
- **#1370** — a fault in the outermost middleware escapes to Starlette's `ServerErrorMiddleware`
  and renders a bare `text/plain` `500`. Pre-existing; identical for `Vary` before this change.
- **#1371** — the published OpenAPI document declares no `securitySchemes`, though every route
  enforces a bearer. Found while relocating `AUTHORIZATION_HEADER`.

## Test plan

- **RED proven behaviourally**, not by ImportError: constants were added first with no wiring, then
  the suite run — **48 failures** against *rendered* response headers, including all five mounted
  routes and all five status classes through the real app.
- App-level tests drive `client(vault_path=vault)` and read `response.headers.get_list(...)` —
  `get_list`, never `headers[...]`, because httpx joins duplicate lines and would hide a doubled
  header or a `no-store, max-age=600` split across two lines. Builder-level tests read
  `raw_headers` and a reserialised header block.
- Parametrised over header **spellings** (canonical/lower/upper) for both `Cache-Control` and an
  echoed `Authorization` token — #1365's own battery had a mutant survive because a test used only
  lower-case.
- **Mutation battery: 11 mutants, 11 killed, zero survivors.** Seed loses `Authorization`; standing
  order reversed; dedupe `seen` seeded from only the first token; `Cache-Control` omitted; caller's
  value wins; unfolded name comparison; half-folded name comparison; `no-store` weakened to
  `no-cache, max-age=0`; `Cache-Control` unioned instead of replaced (killed by 12 assertions,
  including a `no-store, no-store` cell); `AUTHORIZATION_HEADER` near-miss spelling; directive
  stamped only on refusals.
- Twelve existing `Vary` assertions updated deliberately — byte-exact ones stayed byte-exact rather
  than being relaxed to set membership.
- **Gate 2: `./scripts/check-all.sh` exit 0** — 12/12 gates, 8948 passed, coverage 94.41%, mypy
  strict, xenon, interrogate, bandit, ruff clean. No bypasses added anywhere, tests included.

Gate 2.5 (`code-review-orchestrator`) returned two documentation blockers, both fixed and Gate 2
re-run to exit 0: the RFC 9110 §15.1 list had wrongly included `308` (it enumerates 11 codes, not
12 — the intersection was unaffected, but a wrong citation in *that* docstring is the one thing its
own text says is worse than silence), and `info.description` was the single place that dropped the
"this builder constructs" scoping used everywhere else. One non-blocking note accepted as-is:
`test_no_store_rides_on_every_mounted_route` asserts `status != 404` rather than a pinned status,
because reflections has no LLM provider in the hermetic environment and the `200` case is pinned
byte-exact elsewhere.

**OpenAPI golden: unchanged.** The byte-pinned artefacts are the component schemas under
`docs/contracts/adepthood-v1/schemas/`; no wire model, operation, parameter or status moved.
`info.description` prose changed (it stated the old one-token contract) and now interpolates from
the header constants so it cannot drift. Verified with `git diff --stat -- docs/contracts/`.

Closes #1129
Refs #1128

### DIFF
--- a/creek-tools/creek_mcp/api/openapi.py
+++ b/creek-tools/creek_mcp/api/openapi.py
@@ -39,6 +39,7 @@ from creek_mcp.api.models import (
     ErrorCode,
 )
 from creek_mcp.api.routes import (
+    AUTHORIZATION_HEADER,
     CEILING_HEADER,
     CONTRACT_VERSION_HEADER,
     IMPLEMENTED_CAPABILITIES,
@@ -67,11 +68,25 @@ DOCUMENT_TITLE: Final[str] = "Creek Adepthood /v1 application API"
 
 DOCUMENT_DESCRIPTION: Final[str] = (
     "Authenticated HTTP/JSON access to Creek's Adepthood-facing capabilities. "
-    f"Every response carries Vary: {CEILING_HEADER}, every error is the "
-    "ErrorEnvelope, and retryability is the static retry-policy.json table "
-    "keyed on the error code alone."
+    f"Every response this application builds carries Vary: {CEILING_HEADER}, "
+    f"{AUTHORIZATION_HEADER} and Cache-Control: no-store — do not enable "
+    "caching for /v1 on an intermediary. Every error is the ErrorEnvelope, and "
+    "retryability is the static retry-policy.json table keyed on the error "
+    "code alone."
 )
-"""``info.description``. Prose only; it states no fact about any vault."""
+"""``info.description``. Prose only; it states no fact about any vault.
+
+Interpolated from the header constants rather than spelled out, so the sentence
+a consumer reads and the header the server stamps cannot drift — the same
+reason :func:`_parameters` publishes the parameter names from constants.
+
+The caching half of it changed with #1129 and is a *description* change only:
+no operation, parameter, schema or status moved, so the byte-pinned component
+schemas under ``docs/contracts/adepthood-v1/schemas/`` are untouched and the
+contract minor does not advance. It documents a response header, which OpenAPI
+3.1 has no required place for on a document-wide basis, so prose is where it
+goes until the per-response ``headers`` objects are worth generating.
+"""
 
 _DEFS_KEY: Final[str] = "$defs"
 """The key Pydantic hoists nested definitions under, and OpenAPI does not."""
--- a/creek-tools/creek_mcp/api/routes.py
+++ b/creek-tools/creek_mcp/api/routes.py
@@ -52,8 +52,28 @@ without it, and those two must name the same header.
 CEILING_HEADER: Final[str] = "X-Creek-Tier-Ceiling"
 """Header carrying the caller's declared tier ceiling.
 
-Also the value of every response's ``Vary``, so an intermediary cache can never
-serve one caller's ceiling-filtered response to another.
+Also one of the two standing ``Vary`` tokens, so an intermediary cache can never
+serve one caller's ceiling-filtered response to another. The other is
+:data:`AUTHORIZATION_HEADER`.
+"""
+
+AUTHORIZATION_HEADER: Final[str] = "Authorization"
+"""The header a consumer presents its bearer token in.
+
+Declared in this framework-free module, rather than beside the middleware that
+reads it, because two modules now have to agree on the string and one of them
+cannot import the other. :mod:`creek_mcp.httpapi.auth` *reads* this header to
+authenticate; :mod:`creek_mcp.httpapi.errors` *names* it in every response's
+``Vary`` so a cache keys on the credential (#1129). Those two must be the same
+header or the ``Vary`` declares a dependency on something no request carries —
+and ``auth`` already imports ``errors``, so the constant cannot live there
+without a cycle.
+
+Unlike its two neighbours it is **not** published in the OpenAPI document: that
+document declares no ``securitySchemes`` and no bearer requirement at all,
+despite every route sitting behind the gate. That gap is tracked separately; it
+is a reason this constant will gain a third reader, not a reason to file it
+somewhere else.
 """
 
 OP_CAPABILITIES: Final[str] = "getCapabilities"
--- a/creek-tools/creek_mcp/httpapi/auth.py
+++ b/creek-tools/creek_mcp/httpapi/auth.py
@@ -37,6 +37,7 @@ from typing import TYPE_CHECKING, Final
 from starlette.datastructures import Headers
 
 from creek_mcp.api.models import ErrorCode
+from creek_mcp.api.routes import AUTHORIZATION_HEADER
 from creek_mcp.httpapi.context import HTTP_SCOPE, context_of, pass_through
 from creek_mcp.httpapi.errors import error_response
 from creek_mcp.remote_auth import (
@@ -49,9 +50,6 @@ if TYPE_CHECKING:
     from collections.abc import Mapping
 
     from starlette.types import ASGIApp, Receive, Scope, Send
-
-AUTHORIZATION_HEADER: Final[str] = "Authorization"
-"""The header a consumer presents its bearer token in."""
 
 BEARER_SCHEME: Final[str] = "Bearer"
 """The one accepted scheme spelling. Compared exactly, never case-folded."""
--- a/creek-tools/creek_mcp/httpapi/errors.py
+++ b/creek-tools/creek_mcp/httpapi/errors.py
@@ -1,7 +1,9 @@
 """The one place ``/v1`` builds a response, error or otherwise (#1074).
 
 Two invariants live here, and both are structural rather than behavioural —
-which is why they are worth concentrating in one small module.
+which is why they are worth concentrating in one small module. The second one
+carries most of the prose: the bold headings beneath it are the arguments for
+its shape, not a third and fourth invariant.
 
 **One error-envelope construction site.**
 :class:`~creek_mcp.api.models.ErrorEnvelope` is ``extra="forbid"`` with three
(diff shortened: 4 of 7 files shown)


Geoffe-Ga:
## Summary
Adds `Authorization` as a second standing `Vary` token and stamps `Cache-Control: no-store` on every `/v1` response built by the single `json_response` builder in `creek_mcp/httpapi/errors.py`, closing #1129. `AUTHORIZATION_HEADER` moves from `httpapi/auth.py` to the framework-free `api/routes.py` so `auth.py` (which reads the header) and `errors.py` (which now names it in `Vary`) share one source of truth without an import cycle. Docs, OpenAPI `info.description`, and both admission/hardening test suites are updated in step.

## Strengths
- The core mechanism (`_folds_to`, `_is_standing`, `_vary_tokens`, `_merged_headers`) is correct: `Vary` unions case-folded, deduplicated, standing-tokens-first; `Cache-Control` is replaced wholesale under any spelling; passthrough correctly excludes both standing names via `_is_standing` so the `|` merge can't be clobbered.
- No import cycle introduced: `routes.py` imports nothing from `auth.py`/`errors.py`, so hoisting `AUTHORIZATION_HEADER` there is the right fix for the auth↔errors dependency the PR body describes.
- Test coverage is genuinely thorough and matches the implementation: spelling-parametrized (canonical/lower/upper) cases for both headers, weakening-directive cases (`max-age`, `public, max-age`, echoed `no-store`), dedup-on-echo for the new `Authorization` token, a whole-stack test proving a handler can't override either header, and status-class sweeps (401/404/422/403/200) for both `Vary` and `Cache-Control`. Verified the referenced fixtures (`MOUNTED`, `_call`, `_UNAUTHENTICATED_STATUS`, etc.) all pre-exist in `tests/v1_api_support.py`/the test file, so nothing here is dangling.
- RFC 9110 §15.1 heuristically-cacheable status set `{200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501}` is cited correctly (11 codes, no `308`), and the PR body is candid that the correction was made after a review pass — a good sign the claim was actually checked rather than asserted.
- `docs/api.md`'s new "Caching" section is scoped honestly: it names the two known gaps (#1369 trailing-slash `307`, #1370 bare `500`) that bypass this builder rather than overclaiming coverage.
- `info.description` interpolates from the header constants instead of restating them as a literal, so the doc and the header can't drift independently — confirmed no golden/byte-pinned test asserts `info.description`, so this doesn't touch `docs/contracts/adepthood-v1/schemas/`.

## Security Concerns
🟢 LOW — this is itself a defense-in-depth hardening PR, not a change that introduces exposure. The one substantive claim to check — that the routing `404` is heuristically cacheable and is the one live-risk case even under the loopback default — checks out under RFC 9110 §15.1 and is independent of the deployment-posture debate the PR raises about the linked issue's P1/P2 label.

## Problems
None found. No behavioral bug, no missing test, no CLAUDE.md violation.

## Code Quality
- The module docstring in `errors.py` (and to a lesser extent `routes.py`'s new constant docstring) is extremely long — dozens of lines of argumentative prose per symbol. This matches the file's pre-existing style (the original docstring was already dense argument-laden prose before this PR), so it's consistent with established convention here rather than a new pattern, but it's worth the author's awareness that this module is trending toward being harder to skim than to read in full.
- `_folds_to`/`_is_standing`/`Mapping[str, MergeRule]` tradeoff is explained rather than just implemented — a good call given two genuinely different merge rules, and the PR is explicit about why a table-driven abstraction was rejected (2 members isn't enough repetition to justify it).

## CLAUDE.md Adherence
✅ Coverage/lint/types/complexity: PR body reports `check-all.sh` exit 0 (12/12 gates, 8948 passed, 94.41% coverage, mypy strict, ruff clean); could not independently execute the scripts in this sandbox (no `uv`/`pytest` available), but static review of the diff shows no obvious violation of strict typing, complexity, or the "no shortcuts" rule (no `noqa`/`type: ignore` added).
✅ TDD: PR body describes RED proven behaviorally (48 failures against rendered headers before wiring) rather than by import error — matches the Gate 1 requirement.
✅ Commits: branch-based (`issue/1129-issue`), not a direct-to-main commit.
✅ No duplicated logic — the two-header merge mechanism is centralized in one builder, consistent with the "single source of truth" principle.

## Testing
Strong. Byte-exact assertions (not membership tests) catch reordering/doubling regressions; `get_list` is used instead of `headers[...]` specifically to avoid httpx's header-joining masking a doubled or split header line — a subtlety that matters and is correctly handled. The one accepted non-pinned assertion (`status != 404` for the reflections mounted-route case, due to no LLM provider in the hermetic test env) is reasonable and explained.

## Documentation
Good. `docs/api.md`'s new "Caching" section is accurate, cross-links correctly to the existing `#request-logging` anchor, and is scoped to what the builder actually covers rather than overclaiming. Docstrings are thorough to the point of being verbose (see Code Quality), but factually consistent with the code they describe.



## Verdict: LGTM

The header-merge logic is correct and matches its extensive test suite (verified imports, fixtures, and control flow directly against the checked-out source, not just the diff); no import cycles, no missing tests, no factual errors in the RFC 9110 citation, and no CLAUDE.md violations. The only note — very verbose docstrings — is a pre-existing house style in this file, not a new problem, so it doesn't rise above a nit.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5249526627 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: NOT CURRENT

Issue #1377, opened 2026-08-11
fix(ingest): Discord ingest call sites surface a containment refusal as a raw traceback


Geoffe-Ga:
## Role

You are giving the Discord ingest call sites the same refusal banner every other ingest surface has.

## Goal

#1294 put a containment gate on `Ingestor.ingest`, so every caller now
fails closed on a source tree containing a symlink that escapes it. Three
call sites inherit the guard but never translate the exception, so the
refusal reaches the operator as a raw traceback instead of a message:

- `creek/ingest/discord_dispatch.py:292`
- `creek/ingest/discord.py:889`
- `creek/ingest/discord.py:914`

All three call `DiscordIngestor().ingest(...)` directly. Compare
`creek/cli.py::_run_ingest` and `creek/cli.py::_assert_ingest_source_contained`,
which print `Symlink containment: ...` and exit 1, and
`creek_mcp/tools/ingest.py`, which returns a structured
`refusal_response(...)` naming the link.

## Verified evidence

`creek._containment.EscapingSymlinkError` propagates out of
`Ingestor.ingest` by design — that is load-bearing, because swallowing it
would leave `run_ingest`'s `seen_keys` empty and let `tomb_missing_units`
soft-tomb every live ledger key (#1294). So the fix here is *translation
at the call site*, never a `try: ... except: pass`.

## Output format

A PR that:

1. Catches `EscapingSymlinkError` at each of the three call sites and
   renders the same `Symlink containment: <message>` shape the CLI uses,
   with a non-zero exit where the call site owns one.
2. Red-first: a test per call site asserting the operator-visible message
   names the link and does not name its resolved target.
3. Does not swallow the error into a collected-errors list — see above.

## Constraints

- Security posture is already correct and must not change: these paths
  fail closed today. This is presentation only.
- Do not weaken the #1294 gate or its propagation contract.
- No bypasses; `./scripts/check-all.sh` exits 0.

## Why P3

The behaviour is already safe — nothing leaks and nothing is written. The
staging directories these paths ingest live under the vault, so an
escaping symlink there needs an operator to have created one deliberately.
This is a legibility bug, not a correctness or security one.

Refs #1294

Geoffe-Ga:
## A third call site with the same shape — `creek_mcp/tools/upload.py`

Raised by the Gate 2.5 review of PR #1378 and folded in here rather than filed
separately, since it is the same defect on another surface.

[...]

Issue #1375, opened 2026-08-11
fix(ingest): CodeIngestor recurses into intra-root symlinked directories, duplicating fragments


Geoffe-Ga:
## Role

You are bounding `CodeIngestor`'s directory recursion.

## Goal

`CodeIngestor` is the only ingestor that walks its source tree by hand:
`creek/ingest/code.py:561-574 _discover_directory` does
`sorted(dir_path.iterdir())` and recurses on `item.is_dir()`. `is_dir()`
**follows symlinks**, so a symlinked directory pointing anywhere under the
source root is descended into — repeatedly, when it aliases an ancestor.

Every other ingestor uses `Path.rglob`, which does not descend into
symlinked directories on 3.11-3.13 (3.13 added `recurse_symlinks`,
defaulting to `False`), so none of them have this shape.

## Verified evidence

Measured on `main` at the tip of #1294's branch, with the #1294 containment
gate in place. The link stays **inside** the source root, so containment
admits it — correctly; #1294 refuses only links that *leave* the root:

```
src/
  a/
    README.md        <- the only real source file
    loop -> src      <- intra-root symlink to its own grandparent
```

```
discovered=16  fragments=16  distinct source_paths=16
  <src>/a/README.md
  <src>/a/loop/a/README.md
  <src>/a/loop/a/loop/a/README.md
  <src>/a/loop/a/loop/a/loop/a/README.md
  ...
```

One file becomes sixteen vault fragments, each with a distinct
`source_path` and therefore a distinct deterministic id, so dedup does not
collapse them. Recursion terminates only because the kernel eventually
raises `ELOOP` at around depth 8 — that is the bound, not anything in the
code.

A benign, non-cyclic alias has the milder version of the same bug: a
`latest -> 2026-08-01` link inside an export re-ingests that whole subtree
a second time under a second set of paths.

## Output format

A PR that:

1. Stops `_discover_directory` descending into symlinked directories, so
   its walk matches the `rglob` bound every sibling ingestor already has.
   `item.is_dir() and not item.is_symlink()` is the minimal form.
2. Decides explicitly whether an intra-root symlinked directory should be
   *skipped* (matching `rglob`) or *followed once* with a visited-set on
   resolved paths. Skipping is the smaller change and the one consistent
   with the other ten ingestors; say which you chose and why.
3. Red-first: a test that ingests the tree above and asserts exactly one
   fragment, with the count driven by the number of real files rather than
   a magic 16.
[...]

PR #1378, opened 2026-08-11
fix(ingest): confine every ingestor discovery walk to its source root

### LINKED ISSUE #1294: fix(ingest): ingestor discovery walks follow symlinks out of the source root

Geoffe-Ga:
## Summary

Every one of the 11 registered ingestors discovered its inputs with an unguarded walk and read whatever the walk yielded, so a symlink under the source tree pointing outside it was followed, read, and written into the vault as a durable fragment — which then feeds classification, embeddings, cloud LLM prompts, drafts and compiled pages.

This was not theoretical. Running the real `run_ingest` on `main` with a sentinel that exists **only** outside the source root:

```
markdown   written=2   LEAKED -> 01-Fragments/Notes/2026-08-10-Victim-notes.md
generic    written=2   LEAKED -> 01-Fragments/Unsorted/2026-08-11-link.md
code       written=3   LEAKED -> 01-Fragments/Technical/2026-08-10-Deep-victim.md
                       LEAKED -> 01-Fragments/Technical/2026-08-10-Victim.md
markdown   written=1   LEAKED -> 01-Fragments/Notes/2026-08-10-Victim.md   (named root is the link)
```

After this PR all four refuse and no vault file carries the sentinel.

`Deep-victim` is the one worth noticing: it comes from a *nested subdirectory* of the out-of-tree target. `CodeIngestor` is the only ingestor that walks by hand (`iterdir()` + `is_dir()`, and `is_dir()` follows links), so it walked the whole escaped subtree rather than just the named link.

### Where the premise needed correcting

The issue says "every ingestor discovers its inputs with an unguarded **recursive glob**". Unguarded is right — 11 of 11. Recursive glob is not: `chatgpt`/`claude` use non-recursive `glob("*.json")`, `discord` uses `iterdir()` under `messages/`, and `code` uses manual recursion. That matters, because a fix written only against `rglob` would have missed the worst offender.

| ingestor | discovery walk | guarded before? |
|---|---|---|
| `markdown` | `markdown.py:346` `rglob("*.md")`, no `is_file()` either | no |
| `generic` | `generic.py:199` `rglob("*")` | no |
| `document` | `documents.py:630` `rglob("*")` | no |
| `image` | `images.py:441` `rglob("*")` | no |
| `presentation` | `presentations.py:278` `rglob("*")` | no |
| `spreadsheet` | `spreadsheets.py:401` `rglob("*")` | no |
| `substack` | `substack.py:319` `rglob("*.html")` | no |
| `chatgpt` | `chatgpt.py:89` `glob("*.json")` — not recursive | no |
| `claude` | `claude.py:284` `glob("*.json")` — not recursive | no |
| `discord` | `discord.py:391` `iterdir()` under `messages/` | no |
| `code` | `code.py:561-574` manual `iterdir()` + `is_dir()` recursion | no — and descends into symlinked dirs |

## The three design questions

**Skip or refuse? Refuse.** #1360 set the rule by asking whether the caller writes: the read path (`redact --scan`) skips-and-counts, because refusing a whole scan over one bad link is a denial of service on the safety pass — the operator's only instrument for learning what is exposed. The write path (`--apply`/`--review`) refuses. Ingest writes, and writes the most consequential thing in the system. The DoS argument does not transfer: refusing an ingest removes no diagnostic capability (`creek redact --scan` still inspects the tree) and ingest is idempotent, so a refusal costs a retry rather than data. Decisively, `creek process` with the default `redaction.enabled: true` *already* refused this tree via #1087 — so skipping would have left the same tree behaving differently depending on a config toggle, which is precisely the defect this issue names.

**One chokepoint or N? One.** `Ingestor.ingest` is the single door all 11 go through, and the gate sits there, before `_discover_safe`. N per-ingestor copies would drift — the same argument #1360 used to make `resolves_within` public rather than write it twice. The chokepoint also buys a property copies never could: `test_every_registered_ingestor_refuses_an_escaping_link` drives straight off `INGESTOR_REGISTRY`, so a twelfth ingestor is covered without anyone remembering.

**Does a symlinked source root launder its children? Yes — same shape as #1360.** `assert_source_contained` therefore checks the named root with `named_path_escapes` **before** any `is_dir()`, which follows the link. Without that ordering the tree walk resolves the root to the link's target and then finds nothing escaping inside it.

## Two hazards the issue did not predict

Both are the #1360 lesson — enumerate the paths that *write*, not only those that read.

**A refusal that returns `[]` is a mass-deletion primitive.** `run_ingest` builds `seen_keys` from the discovered fragments, then calls `tomb_missing_units`, which soft-tombs every `ledger.live_keys() - seen_keys` for a ledger-backed directory. Markdown is ledger-backed. Had the refusal been collected into `IngestResult.errors` and turned into an empty discovery, one stray symlink would have orphaned every fragment previously ingested from that source. The refusal propagates out of `Ingestor.ingest` instead, and a test pins that the originals stay live.

**The consent gate read through the link first.** `creek ingest` calls `_gate_consent` before ingesting, which reaches `creek/consent.py:134 _build_source_summary` — an `rglob("*")` walk that `stat()`s every entry, following escaping links and folding out-of-tree file counts and byte totals into the very summary the operator is asked to approve. A guard sitting only downstream blocks the write and still performs that out-of-root read. The CLI now refuses above `_gate_consent`, and a recording spy asserts the summariser never runs.

## One predicate, three surfaces

The issue asked for "one shared confinement predicate rather than a third copy of the policy", so `resolves_within` moved out of `creek/redact/scanner.py` into a new stdlib-only `creek/_containment.py` and is re-exported; `_named_path_escapes` moved out of `creek/redact/cli_commands.py` and is rebound; and the redact CLI's `_assert_no_escaping_symlinks` now calls the shared `find_escaping_symlink` rather than keeping its own `os.walk`. A test asserts object identity across the modules so a fourth copy cannot reappear.

`_containment` is stdlib-only by design, following `creek/_fsio.py`: `creek.ingest.base` depends on it, and `creek.redact.scanner` compiles a large regex battery at import, so the dependency runs one way only.

## Notes

- **`recurse_symlinks` is pinned behaviourally, not by keyword.** The parameter does not exist on 3.11/3.12, so passing it would break two of the three CI versions. Confirmed on the interpreter that `rglob` does not descend into symlinked directories today; a test is the tripwire if a future default flips. The gate itself uses `os.walk(followlinks=False)` and inspects `dirnames` as well as `filenames`, which is what closes `CodeIngestor` without touching it.
- **`MarkdownIngestor` gained an `is_file()` filter** it uniquely lacked. Not the containment fix — a robustness fix for the same tombing hazard, reached via a dangling in-tree link.
- **#1087 is untouched.** `SymlinkEscapedSourceError` and the pipeline refusal stay; with redaction on it still fires first, and it remains the better-informed of the two because it carries the count of files the scan declined.
- **Security direction is one-way**: no `--force`, no environment variable, no config key.

## Test plan

- `./scripts/check-all.sh` exits **0** — 8956 passed, 5 skipped, coverage 94.40%, per-file gate green (257 files ≥ 80%).
- New `tests/test_ingest_symlink_containment.py` (39 tests). 30 were RED before the fix; 9 pass at HEAD by design as non-vacuity controls and over-breadth guards — 6 clean-control ingestor lanes proving each fixture actually writes fragments, plus intra-tree alias, symlinked-ancestor root, and the `rglob` tripwire. Without those controls every "sentinel absent" assertion would be trivially true.
- The primary RED asserts on the **written vault fragment** with `redaction.enabled = false`, not on a glob's return value.
- Non-vacuity: 6 ingestors × their real input shapes, plus a registry-driven lane over all 11.
- **Mutation battery, 9 mutations, 8 real ones all killed:** dropping the gate (23 tests fail), dropping the named-root check, ignoring `dirnames`, `followlinks=True`, dropping the CLI pre-consent gate, dropping the MCP translation, inverting `resolves_within`'s failure arm (42 fail), dropping the markdown `is_file()` guard. Two survivors were found and fixed rather than accepted: `followlinks=True` and the markdown guard were both untested, so I added tests for each. A third apparent survivor was an `except EscapingSymlinkError: raise` clause in `_discover_safe` that turned out to be **unreachable** — the gate runs before it — so I deleted the dead code and moved its rationale into a comment at the live call site.
- Neighbouring suites green: `test_pipeline.py`, `test_cli_redact.py`, `test_ingest*.py` — 300 passed.

## Review

Gate 2.5 (`code-review-orchestrator` over the full diff) came back **CLEAN — no blockers**. It independently confirmed the non-disclosure paths, the absence of an import cycle, the `logger.exception` → `logger.warning` swap being correct rather than a regression, that the exception cannot be swallowed before `tomb_missing_units`, and the 39/9-at-HEAD test accounting. Two non-blocking items, both handled:

- **Performance (should-fix).** `creek process` fans all 11 ingestors at one source, so the gate re-walks the tree 11 times — roughly 385k extra `lstat` calls on a 35k-file export. The reviewer's judgement, which I agree with, is to ship: this stage *already* performs 11 full `discover()` walks that read content and parse JSON, so an `lstat`-only pass is a constant factor on the same order, not a new asymptotic class. Hoisting it to `Pipeline._run_ingestion` would either need an "already verified" flag threaded through `Ingestor.ingest` — undercutting the whole point of a gate a subclass cannot weaken — or would drop the guarantee for every direct caller (`creek ingest`, the MCP tool, the two Discord paths). Deliberately not pre-empting without a profile at real scale.
- **Discord call sites (nit).** `discord_dispatch.py:292` and `discord.py:889,914` inherit the guard and fail closed, but render the refusal as a raw traceback instead of a banner. Presentation only, no security impact. Filed as #1377 (P3).

## Left alone

`CodeIngestor` still recurses into *intra-root* symlinked directories, so an alias pointing at its own ancestor turns one file into 16 fragments with distinct source paths, bounded only by `ELOOP`. Containment admits that link correctly — it never leaves the root — so it is a walk-bound bug rather than a security one. Filed as #1375 (P2) with the reproduction.

Closes #1294
Refs #1087, #1293, #1375, #1377

### DIFF
--- a/creek-tools/creek/_containment.py
+++ b/creek-tools/creek/_containment.py
@@ -29,6 +29,9 @@ The policy, unchanged from the shipped SEC-003 guard:
   would flag every child of a root reached *through* a symlinked component
   — ``/tmp`` -> ``/private/tmp`` on macOS — as escaping.
 * Unprovable containment IS an escape.
+* The resolution primitive is ``os.path.realpath``, NOT
+  :meth:`~pathlib.Path.resolve` — see :func:`_resolved_target` for why the
+  difference is load-bearing rather than stylistic.
 
 That policy is deliberately LEAF-ONLY, so a path reached through an
 escaping *ancestor* component (``<root>/linkdir/a.md``, where ``linkdir``
@@ -93,6 +96,53 @@ class EscapingSymlinkError(RuntimeError):
         )
 
 
+def _resolved_target(child: Path) -> Path | None:
+    """Resolve *child* to the location containment must be judged against.
+
+    The one place symlinks are followed, and deliberately NOT via
+    :meth:`pathlib.Path.resolve`. ``Path.resolve``'s behaviour on a symlink
+    **cycle** is not part of pathlib's contract and it changed under us:
+
+    * On 3.11/3.12, ``Path.resolve(strict=False)`` raises
+      ``RuntimeError("Symlink loop from ...")``.
+    * On 3.13 it delegates to ``os.path.realpath``, which for
+      ``strict=False`` treats a cycle as "stop unwinding here" and returns
+      a PARTIAL path with the looping component left unresolved.
+
+    That partial answer is worse than an error, because for a cycle that
+    closes back on its own starting point it erases the hops in between.
+    ``<root>/a.md -> <outside>/o.md -> <root>/a.md`` resolves, on 3.13, to
+    ``<root>/a.md`` — an in-root answer for a link whose very first hop
+    leaves the root. Judging containment on that answer admits the link.
+
+    ``os.path.realpath(..., strict=True)`` reports a cycle as
+    ``OSError(ELOOP)`` on every supported version, so it is the primitive
+    with the stable contract. Two arms, and the split is the whole policy:
+
+    * ``FileNotFoundError`` is NOT a containment failure. A dangling link
+      still names a candidate location worth comparing, so it falls back to
+      ``strict=False`` and is judged on where its target *would* sit —
+      which is what refuses a broken link pointing out of the root while
+      admitting a stale in-tree alias.
+    * Every other ``OSError`` — a cycle, an unreadable component — is a
+      containment that cannot be proven, and by this module's policy an
+      unprovable containment is an escape.
+
+    Args:
+        child: A path being tested for containment, as walked.
+
+    Returns:
+        The location to compare against the root, or ``None`` when
+        resolution failed in a way that leaves containment unproven.
+    """
+    try:
+        return Path(os.path.realpath(child, strict=True))
+    except FileNotFoundError:
+        return Path(os.path.realpath(child, strict=False))
+    except OSError:
+        return None
+
+
 def resolves_within(child: Path, resolved_root: Path) -> bool:
     """Report whether *child*'s symlink target stays under *resolved_root*.
 
@@ -102,15 +152,21 @@ def resolves_within(child: Path, resolved_root: Path) -> bool:
     drift into subtly different definitions of "inside".
 
     The failure arm is a deliberate classification, not a swallowed error:
-    a loop (``RuntimeError``), an unreadable link (``OSError``), and a
-    target outside the root (``ValueError`` from ``relative_to``) are all
-    cases where containment cannot be proven — and an unprovable
-    containment is an escape. Every caller logs and counts its rejections.
+    an unresolvable link (``None`` from :func:`_resolved_target` — a cycle,
+    an unreadable component) and a target outside the root (``ValueError``
+    from ``relative_to``) are both cases where containment cannot be
+    proven, and an unprovable containment is an escape. Every caller logs
+    and counts its rejections.
 
-    ``strict=False`` is deliberate: a dangling link still resolves to a
-    candidate location worth comparing, so a broken link pointing *outside*
-    is still refused rather than waved through on the technicality that its
-    target does not exist yet.
+    Resolution goes through :func:`_resolved_target` rather than
+    :meth:`~pathlib.Path.resolve` so that this classification is a decision
+    this module makes, not one inherited from whichever exception the
+    stdlib happens to raise that release. It was the latter until 3.13
+    changed the answer underneath it; see that function.
+
+    A dangling link is still judged on its candidate location, so a broken
+    link pointing *outside* is refused rather than waved through on the
+    technicality that its target does not exist yet.
 
     The exception object itself is deliberately dropped rather than logged:
     ``relative_to``'s message quotes the *resolved* target, which is exactly
@@ -125,7 +181,10 @@ def resolves_within(child: Path, resolved_root: Path) -> bool:
         ``True`` when the link's target is a descendant of *resolved_root*.
     """
     try:
-        child.resolve(strict=False).relative_to(resolved_root)
+        target = _resolved_target(child)
+        if target is None:
+            return False
+        target.relative_to(resolved_root)
     except (OSError, RuntimeError, ValueError):
         return False
     return True
--- a/creek-tools/creek/_containment.py
+++ b/creek-tools/creek/_containment.py
@@ -1,0 +1,243 @@
+"""Symlink containment for trees the operator names (#1087/#1293/#1294).
+
+Three surfaces need the same question answered — "does this path leave the
+root it was reached through?" — and until #1294 each answered it in its own
+words:
+
+* ``creek.redact.scanner._scannable_candidates`` — the redaction **read**
+  path, which skips an escaping child and counts it.
+* ``creek.redact.cli_commands`` — the redaction **write** path, which
+  refuses before it opens anything.
+* ``creek.ingest`` — every ingestor's discovery walk, which had no
+  containment check at all and read whatever its glob yielded.
+
+A predicate copied three times is three predicates, and they drift. This
+module is the single definition all three now import, so "inside" cannot
+come to mean one thing for the scanner and another for the ingestors.
+
+**Stdlib-only by design.** :mod:`creek.ingest.base` imports this module, and
+:mod:`creek.redact.scanner` compiles a large regex battery at import time;
+depending on the scanner from the ingest path would drag redaction into
+every ingest. The same reasoning that keeps :mod:`creek._fsio` underneath
+the writers keeps this module underneath both.
+
+The policy, unchanged from the shipped SEC-003 guard:
+
+* Resolve the ROOT exactly once.
+* ``lstat`` (i.e. :meth:`~pathlib.Path.is_symlink`) only the LEAF; never
+  resolve a child that is not itself a link. Resolving non-symlink children
+  would flag every child of a root reached *through* a symlinked component
+  — ``/tmp`` -> ``/private/tmp`` on macOS — as escaping.
+* Unprovable containment IS an escape.
+
+That policy is deliberately LEAF-ONLY, so a path reached through an
+escaping *ancestor* component (``<root>/linkdir/a.md``, where ``linkdir``
+is the link) is admitted by :func:`resolves_within` alone. For a walked
+tree that residual is closed by :func:`find_escaping_symlink`, which sees
+``linkdir`` in its own right; for a directly-named path it remains the
+documented residual recorded in ``docs/security/threat-model.md``.
+
+Security direction here is one-way: every function may only ever cause a
+caller to read *less*. There is no waiver — no ``--force``, no environment
+variable, no config key — per SEC-003's no-waiver precedent.
+"""
+
+from __future__ import annotations
+
+import itertools
+import logging
+import os
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+class EscapingSymlinkError(RuntimeError):
+    """A symlink under a named root resolves outside that root.
+
+    Raised by :func:`assert_source_contained` so that ingestion refuses a
+    source tree rather than reading through the link. Refusal rather than
+    skipping is the contract for this surface: ingest is a *write* path —
+    what it reads becomes a durable vault fragment that then feeds
+    classification, embeddings, cloud LLM prompts, drafts and compiled
+    pages — and #1360 established that the write path refuses while only
+    the read path (``redact --scan``, which writes nothing) may skip and
+    count.
+
+    Neither :attr:`path` nor the message ever names the link's *resolved*
+    target. Disclosing where the link points is the exfiltration oracle
+    #1087 closes, and it would be perverse for the refusal to leak the very
+    thing it refused to read.
+
+    Attributes:
+        path: The offending link exactly as walked or supplied — never
+            resolved, and never normalised, so the operator can find it by
+            the name they know it by.
+        root: The source root the operator named.
+    """
+
+    def __init__(self, path: Path, root: Path) -> None:
+        """Build a containment refusal naming the link and its root.
+
+        Args:
+            path: The offending symlink, exactly as walked or supplied.
+            root: The source root the operator named.
+        """
+        self.path = path
+        self.root = root
+        super().__init__(
+            f"Refusing to ingest {root}: {path} is a symlink whose target "
+            "resolves outside that tree, so ingestion would write content "
+            "from outside the source into the vault. Remove or re-point the "
+            "link, or ingest the target's own directory directly."
+        )
+
+
+def resolves_within(child: Path, resolved_root: Path) -> bool:
+    """Report whether *child*'s symlink target stays under *resolved_root*.
+
+    The single containment predicate for the whole codebase: the redaction
+    scanner's walked-child surface, the redaction CLI's named-path surface,
+    and the ingest discovery gate all call this one function, so they cannot
+    drift into subtly different definitions of "inside".
+
+    The failure arm is a deliberate classification, not a swallowed error:
+    a loop (``RuntimeError``), an unreadable link (``OSError``), and a
+    target outside the root (``ValueError`` from ``relative_to``) are all
+    cases where containment cannot be proven — and an unprovable
+    containment is an escape. Every caller logs and counts its rejections.
+
+    ``strict=False`` is deliberate: a dangling link still resolves to a
+    candidate location worth comparing, so a broken link pointing *outside*
+    is still refused rather than waved through on the technicality that its
+    target does not exist yet.
+
+    The exception object itself is deliberately dropped rather than logged:
+    ``relative_to``'s message quotes the *resolved* target, which is exactly
+    the out-of-root path this guard exists to keep out of the record.
+
+    Args:
+        child: A path being tested for containment, as walked.
+        resolved_root: The root, already resolved exactly once by the
+            caller.
+
+    Returns:
+        ``True`` when the link's target is a descendant of *resolved_root*.
+    """
+    try:
+        child.resolve(strict=False).relative_to(resolved_root)
+    except (OSError, RuntimeError, ValueError):
+        return False
+    return True
+
+
+def named_path_escapes(path: Path) -> bool:
+    """Report whether *path* is itself a symlink leaving its own parent.
+
+    The containment question for a path the operator names directly, as
+    opposed to one discovered while walking a tree.
+
+    Three properties carry the correctness argument:
+
+    * ``is_symlink()`` is an ``lstat`` on the leaf, so a path that is not
+      itself a link is never resolved and never compared. That is what
+      makes this guard behave identically on darwin and on Linux CI: a root
+      reached *through* a symlinked component (``/tmp`` -> ``/private/tmp``
+      on macOS) is not flagged.
+    * When the leaf *is* a link, BOTH sides are resolved — the target
+      inside :func:`resolves_within`, the parent here — so ``/tmp`` ->
+      ``/private/tmp`` cannot manufacture a spurious refusal for a link
+      that never left its own directory.
+    * The predicate is LEAF-ONLY, and deliberately so: a named path whose
+      escaping link is an ANCESTOR component is admitted. That is a known,
+      accepted residual — not full coverage of path traversal.
+
+    Args:
+        path: The path exactly as the operator supplied it.
+
+    Returns:
+        ``True`` when *path* is a symlink whose target is not a descendant
+        of its own resolved parent directory.
+    """
+    return path.is_symlink() and not resolves_within(
+        path,
+        path.parent.resolve(strict=False),
+    )
+
+
+def find_escaping_symlink(root: Path) -> Path | None:
+    """Return the first descendant of *root* that links out of it, if any.
+
+    Walks with ``followlinks=False`` so the walk itself never descends
+    through a link, and inspects BOTH ``dirnames`` and ``filenames``.
+    Checking directory entries is what closes
+    :meth:`creek.ingest.code.CodeIngestor._discover_directory`, which
+    recurses manually with ``is_dir()`` — and ``is_dir()`` follows symlinks,
+    so unlike the ``rglob`` ingestors it would otherwise walk an entire
+    out-of-tree subtree.
+
+    Returning the offender rather than raising is what lets the two callers
+    with different refusal mechanics — ``typer.Exit`` in
+    :mod:`creek.redact.cli_commands`, :class:`EscapingSymlinkError` in
+    :func:`assert_source_contained` — share one walk instead of keeping two
+    copies of it.
+
+    Args:
+        root: The tree to inspect, exactly as the operator supplied it.
+
+    Returns:
+        The first escaping entry found, as walked, or ``None`` when every
+        symlink under *root* resolves inside it.
+    """
+    resolved_root = root.resolve(strict=False)
+    for dirpath, dirnames, filenames in os.walk(root, followlinks=False):
+        for entry in itertools.chain(dirnames, filenames):
+            candidate = Path(dirpath) / entry
+            if candidate.is_symlink() and not resolves_within(candidate, resolved_root):
+                return candidate
+    return None
+
+
+def assert_source_contained(source_path: Path) -> None:
+    """Refuse a source tree that links out of itself (#1294).
+
+    The gate every ingestor passes through. Two checks, and the order
+    matters:
+
+    1. *source_path* itself, via :func:`named_path_escapes`. This must come
+       BEFORE any ``is_dir()``, which follows the link. #1360 found that
+       resolving a named symlinked directory launders every child as
+       in-root: the tree walk would resolve the root to the link's target
+       and then find nothing escaping inside it.
+    2. Every descendant, via :func:`find_escaping_symlink`.
+
+    A symlink whose target stays inside the root is admitted, so ordinary
+    intra-tree aliases keep working. Containment is about the target
+    escaping, not about the link existing.
+
+    Args:
+        source_path: The source file or directory as the caller supplied
+            it.
+
+    Raises:
+        EscapingSymlinkError: When *source_path* is, or contains, a symlink
+            whose target resolves outside the tree the operator named.
+    """
+    if named_path_escapes(source_path):
+        logger.warning(
+            "Refusing to ingest a source path that is a symlink escaping "
+            "its own parent: %s",
+            source_path,
+        )
+        raise EscapingSymlinkError(source_path, source_path)
+    if not source_path.is_dir():
+        return
+    escaping = find_escaping_symlink(source_path)
+    if escaping is not None:
+        logger.warning(
+            "Refusing to ingest %s: it contains a symlink that escapes the "
+            "source root: %s",
+            source_path,
+            escaping,
+        )
+        raise EscapingSymlinkError(escaping, source_path)
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -18,6 +18,7 @@ from rich.console import Console
 from rich.markup import escape
 from rich.table import Table
 
+from creek._containment import EscapingSymlinkError, assert_source_contained
 from creek.classify.privacy_filter import (
     PrivacyTierOverride,
     override_elevates,
@@ -430,6 +431,9 @@ def _run_ingest(
     except FileNotFoundError as exc:
         console.print(f"[red]Vault unavailable: {exc}[/red]")
         raise typer.Exit(code=1) from exc
+    except EscapingSymlinkError as exc:
+        console.print(f"[red]Symlink containment: {exc}[/red]")
+        raise typer.Exit(code=1) from exc
 
     if print_summary:
         console.print(
@@ -438,6 +442,27 @@ def _run_ingest(
             f"{result.skipped} skipped[/dim]",
         )
     return result.written, result.errors, result.discovered
+
+
+def _assert_ingest_source_contained(source_path: Path) -> None:
+    """Refuse an ingest source that links out of itself, before reading it.
+
+    Translates :class:`creek._containment.EscapingSymlinkError` into the
+    CLI's refusal idiom so the handler can call the gate early — above
+    ``_gate_consent`` — without letting a traceback reach the operator.
+
+    Args:
+        source_path: The ``--input`` path exactly as the operator gave it.
+
+    Raises:
+        typer.Exit: With code ``1`` when *source_path* is, or contains, a
+            symlink whose target resolves outside it.
+    """
+    try:
+        assert_source_contained(source_path)
+    except EscapingSymlinkError as exc:
+        console.print(f"[red]Symlink containment: {exc}[/red]")
+        raise typer.Exit(code=1) from exc
 
 
 def _warn_if_discovered_but_empty(
@@ -1203,6 +1228,13 @@ def process(
     except SymlinkEscapedSourceError as exc:
         console.print(f"[red]Symlink containment: {exc}[/red]")
         raise typer.Exit(code=1) from exc
+    except EscapingSymlinkError as exc:
+        # Reachable only with ``redaction.enabled: false``, which returns
+        # early from ``_run_redaction`` before the #1087 check. The ingestor
+        # gate is then what stops the run, and its refusal must read like a
+        # refusal rather than a traceback (#1294).
+        console.print(f"[red]Symlink containment: {exc}[/red]")
+        raise typer.Exit(code=1) from exc
 
     console.print(f"[bold]Files scanned:[/bold] {result.files_scanned}")
     console.print(f"[bold]Fragments created:[/bold] {result.fragments_created}")
@@ -1310,6 +1342,15 @@ def ingest(
     if not input.exists():
         console.print(f"[red]Input path not found: {input}[/red]")
         raise typer.Exit(code=2)
+
+    # Before ``_gate_consent``, and NOT redundant with the gate inside
+    # ``Ingestor.ingest``. The consent prompt calls
+    # ``creek/consent.py::_build_source_summary``, which walks ``rglob("*")``
+    # and ``stat()``s every entry — following an escaping link and folding
+    # out-of-tree file counts and byte totals into the very summary the
+    # operator is asked to approve. A guard that sits only downstream blocks
+    # the write and still performs that out-of-root read first (#1294).
+    _assert_ingest_source_contained(input)
 
     _gate_consent(
         source_path=input,
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -445,14 +445,28 @@ def _run_ingest(
 
 
 def _assert_ingest_source_contained(source_path: Path) -> None:
-    """Refuse an ingest source that links out of itself, before reading it.
+    """Refuse a source that links out of itself, before anything reads it.
 
     Translates :class:`creek._containment.EscapingSymlinkError` into the
-    CLI's refusal idiom so the handler can call the gate early — above
+    CLI's refusal idiom so a handler can call the gate early — above
     ``_gate_consent`` — without letting a traceback reach the operator.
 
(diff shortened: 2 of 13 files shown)


Geoffe-Ga:
## Summary
This PR closes a real, demonstrated vulnerability: all 11 registered ingestors discovered their inputs with an unguarded walk (`rglob`/`glob`/manual `iterdir`), so a symlink under a named source tree pointing outside it was read and written into the vault as a durable fragment. The fix adds a single shared containment predicate (`creek/_containment.py`), gates it at the `Ingestor.ingest` chokepoint (covering all 11 ingestors including any future ones via a registry-driven test), fixes a related mass-tombing hazard (a swallowed refusal would have orphaned the entire previously-ingested source), and consolidates three previously-drifting copies of the containment predicate into one. It is backed by an unusually thorough 39-test suite including non-vacuity controls, over-breadth guards, and non-disclosure assertions.

## Strengths
- **Real, measured vulnerability with a real, measured fix.** The PR body documents the leak reproduced on `main` and shows it closed after the change — not a theoretical hardening.
- **One chokepoint, not eleven.** Gating `Ingestor.ingest` rather than each `discover()` means a future 12th ingestor is covered automatically; `test_every_registered_ingestor_refuses_an_escaping_link` pins this by driving off `INGESTOR_REGISTRY` directly rather than a hardcoded list.
- **Found and fixed the mass-deletion hazard.** Recognizing that a swallowed `EscapingSymlinkError` inside `_discover_safe` would leave `run_ingest`'s `seen_keys` empty and cause `tomb_missing_units` to soft-tomb every previously-ingested fragment is a genuinely sharp catch, and it's pinned with `test_a_containment_refusal_does_not_tomb_previously_ingested_fragments`.
- **Non-disclosure is tested, not assumed.** `test_the_refusal_never_names_the_resolved_out_of_tree_target` checks both the exception message and the log records, including the Rich-truncation trap (asserting on the bare filename too, not just the full path).
- **One predicate, three surfaces, enforced by identity.** `test_the_containment_predicate_has_exactly_one_definition` asserts `scanner.resolves_within is _containment.resolves_within` rather than just "behaves the same," which is what actually prevents drift over time.
- **Honest about what's left alone.** `CodeIngestor`'s intra-root symlinked-directory recursion bug and the Discord traceback-vs-banner nit are explicitly scoped out and filed (#1375, #1377) rather than silently ignored or over-scoped into this PR.

## Security Concerns
🔴 **BLOCKING — `creek process` still performs the out-of-root consent-summary read the PR claims to have closed.**

The PR's own "Hazard #2" write-up states: *"The consent gate read through the link first... The CLI now refuses above `_gate_consent`."* That fix is real, but it was only wired into the `ingest` subcommand:

- `creek-tools/creek/cli.py` — in `ingest()` (~line 1346), `_assert_ingest_source_contained(input)` is called immediately before `_gate_consent(...)`.
- `creek-tools/creek/cli.py` — in `process()` (~line 1211), `_gate_consent(source_path=source_path, vault_path=vault_path, source_type="pipeline", assume_yes=yes)` is called with **no** preceding containment check.

Both commands funnel through the same `_gate_consent` helper (`cli.py:221`), which for a first-time (unconsented) source calls `manager.get_source_summary(source_path, exclusions=[])` → `creek/consent.py:118 _build_source_summary`, which does:
```python
files = [p for p in sorted(source_path.rglob("*")) if p.is_file() and ...]
total_size = sum(f.stat().st_size for f in files)
```
`Path.is_file()` and `Path.stat()` both follow symlinks by default, so a symlink escaping the source root is counted and its target's byte size is folded into `total_size` and `content_types` — exactly the out-of-root read the PR calls an oracle worth closing, printed straight to the operator's consent prompt (`Found: {file_count} file(s), {size} MB`).

Since `process()` is the primary "run everything" entry point (`creek process`), and its docstring states the same "first-time sources prompt for confirmation before ingestion" behavior as `ingest()`, this is not a hypothetical or secondary code path — it's the main one, and it still walks and stats through an escaping symlink before either the redaction-stage refusal (`SymlinkEscapedSourceError`, which only fires with `redaction.enabled: true`) or the new ingest gate (`EscapingSymlinkError`, which fires inside `pipeline.run()` → `_run_ingestion` → `Ingestor.ingest`) ever get a chance to run.

No test in the new suite or `test_pipeline.py` exercises `_gate_consent` from `process()` against an escaping-symlink tree — `test_cli_ingest_refuses_before_the_consent_gate_reads_the_source_tree` covers only the `ingest` subcommand, and `test_process_refuses_an_escaping_symlink_with_redaction_disabled` calls `pipeline.run()` directly, bypassing the CLI's `_gate_consent` call entirely.

**Suggested fix:** call `_assert_ingest_source_contained(source_path)` (or `assert_source_contained` directly) before `_gate_consent(...)` in `process()`, mirroring what `ingest()` already does, and add a `process()`-level CLI test analogous to the existing `ingest()` one.

🟢 LOW — `CodeIngestor`'s intra-root symlinked-directory recursion (an alias pointing at its own ancestor can multiply one file into many fragments, bounded only by `ELOOP`) is correctly identified as out of scope for a *containment* fix (the link never leaves the root) and is tracked as #1375. No objection to leaving it out of this PR.

## Problems
See the Security Concerns section above — the `process()` consent-gate ordering gap is the one blocking issue found in this review.

## Code Quality
- `creek/_containment.py` is a clean, well-justified extraction; the module-level comment explaining *why* it must stay stdlib-only (to avoid `creek.ingest.base` transitively importing the scanner's regex battery) is exactly the kind of non-obvious rationale worth keeping as a comment.
- The docstrings throughout (`resolves_within`, `named_path_escapes`, `find_escaping_symlink`, `assert_source_contained`) are thorough to the point of restating test rationale; that's a matter of taste, but it does make the security invariants easy to audit without cross-referencing the test file.
- `creek/redact/cli_commands.py`'s switch from `logger.exception` to `logger.warning` in `_assert_no_escaping_symlinks` is correct — no exception is in flight at that call site any more, since the walk moved into `find_escaping_symlink` which returns rather than raises.

## CLAUDE.md Adherence
- ✅ TDD: PR body documents RED tests written against the vulnerability before the fix (30 of 39 new tests RED pre-fix).
- ✅ Coverage: PR claims 94.40% aggregate, per-file gate green — could not independently verify in this review environment (no `uv`/dev environment available), but the new test file's breadth (non-vacuity controls, registry-driven lane, over-breadth guards) makes this plausible.
- ✅ Commit message follows Conventional Commits (`fix(ingest): ...`) with detailed rationale.
- ❓ "No shortcuts / fix root causes": mostly honored, but the incomplete `process()` fix above means the stated security guarantee ("the CLI now refuses above `_gate_consent`") doesn't hold everywhere it's claimed to.

## Testing
Exceptionally thorough for the surface it covers: registry-driven coverage of all 11 ingestors, non-vacuity controls proving each fixture is actually claimed, over-breadth guards for intra-tree and through-symlinked-ancestor cases, dangling-link and symlink-loop classification tests, non-disclosure tests on both the exception message and log records, an object-identity test preventing predicate drift, an MCP-surface test, and two mutation-survivor backfill tests. This is a strong bar.

The one gap: as detailed above, no test exercises `creek process`'s `_gate_consent` call against an escaping-symlink source, which is precisely the scenario `test_cli_ingest_refuses_before_the_consent_gate_reads_the_source_tree` exists to catch for `creek ingest`. That missing test is exactly what would have caught the blocking issue.

## Documentation
`docs/ingestion.md`, `docs/redaction.md`, and `docs/security/threat-model.md` are all updated consistently with the shipped behavior, including a concrete example of the refusal message and an explicit statement that intra-tree symlinks keep working. `docs/security/threat-model.md`'s SEC-003 entry is updated to reflect the three surfaces now sharing one predicate. Docs are accurate for what's implemented; they inherit the same overstatement as the PR description regarding `_gate_consent` (implying the CLI-level pre-consent guard is universal when it's `ingest`-only).



## Verdict: CHANGES_REQUESTED

The PR closes the ingest-time symlink-escape hole thoroughly, but it applies the "refuse before the consent summary reads through the link" fix (hazard #2 in the PR's own description) only to `creek ingest`, not to `creek process` — which calls the exact same `_gate_consent` helper and therefore still performs an out-of-root `rglob("*")`+`stat()` walk via `ConsentManager.get_source_summary` before any containment check fires. That is a real, easily reproduced gap in the very property this PR claims to have established, on the primary full-pipeline command, with no test covering it.

Geoffe-Ga:
**CI**: 9/12 Green
**Action**: pull comment 5249570951 to see in-depth feedback and continue iterating

Geoffe-Ga:
## The 3.13 job was right, and the test was not the thing that was wrong

`Tests & Type Checking (3.13)` failed `test_a_symlink_loop_under_the_root_is_refused` while 3.11 and 3.12 stayed green. That is a hole in the guard on 3.13, not a stale test premise — pushed as 6a984ba.

**What broke.** `resolves_within` took its "I cannot prove this" signal from `Path.resolve` *raising*, and pathlib never specified what `resolve` does with a symlink cycle. On 3.11/3.12 `resolve(strict=False)` raises `RuntimeError("Symlink loop from ...")` and the predicate refused. On 3.13 `resolve` delegates to `os.path.realpath`, which for `strict=False` stops unwinding at the cycle and returns a **partially resolved** path.

**Why that is a security bug and not a cosmetic divergence.** For a cycle that closes back on its own starting point, the partial answer erases the hops in between. Measured on 3.13 at this PR's HEAD:

```
<root>/a.md -> <outside>/o.md -> <root>/a.md
Path.resolve(strict=False)  ->  <root>/a.md      [IN-ROOT]
assert_source_contained      ->  ADMITTED
```

An in-root answer for a link whose very first hop leaves the root. The same tree is refused on 3.11/3.12. This corrects a claim in the description above: `resolves_within` was one predicate across the three *surfaces*, but not across the three *interpreters* CI runs.

No out-of-root content reached the vault even when admitted, because `open()` also fails `ELOOP`. That is the kernel refusing rather than Creek, and it stops holding the moment the cycle is broken at the far end — outside the named tree and outside anything Creek can see.

**The fix.** Resolution moved to `_resolved_target`, which uses `os.path.realpath(..., strict=True)` — a cycle is `OSError(ELOOP)` on every supported version. Only `FileNotFoundError` is non-fatal, falling back to `strict=False`, so a dangling link is still judged on its candidate location (refuse when it points out of root, admit a stale in-tree alias). Direction is one-way per SEC-003: this converts some admissions into refusals and no refusal into an admission.

**Evidence, by execution on both interpreters rather than from the changelog.** Re-installing the pre-fix primitive under 3.13 gives `3 failed, 38 passed`; the identical pre-fix guard under 3.12 gives `41 passed` — the divergence is exactly version-scoped. With the fix, 41 pass on both, and the whole suite passes on 3.13 (9327 passed, with `uv.lock`'s pins).

Two new regression tests, both version-independent, **no `skipif` on the version**: `test_a_symlink_loop_that_leaves_the_root_is_refused` pins the escape the partial answer laundered, and `test_containment_does_not_depend_on_path_resolve_raising` pins the primitive at unit level. The intra-root *directory* loop remains admitted and bounded — that is #1375's walk-bound bug, not containment's.

`docs/security/threat-model.md` records the finding under SEC-003, including the general form: a containment predicate must not take its "cannot prove this" signal from an unspecified stdlib exception.

Gate 2 locally: exit **0** (8958 passed).

Geoffe-Ga:
**CI**: 12/13 Green
**Action**: pull comment 5249570951 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Summary
This PR closes a real, demonstrated vulnerability (#1294): all 11 registered ingestors discovered their inputs with unguarded walks and would follow a symlink pointing outside the named source tree, writing the target's content into the vault. It adds a single shared containment predicate (`creek/_containment.py`), gates it at the one chokepoint every ingestor passes through (`Ingestor.ingest`), fixes a related mass-tombing hazard, fixes a Python 3.13 symlink-loop resolution regression, and adds an exceptionally thorough test suite (39 new tests) plus documentation/threat-model updates. The engineering quality here is very high — but the fix has a gap: `creek process`'s consent-summary walk still reads through an escaping symlink before any guard fires, which is the exact class of bug this PR explicitly identifies and fixes for `creek ingest`.

## Strengths
- Single shared predicate (`resolves_within`, `named_path_escapes`, `find_escaping_symlink`) consumed by all three surfaces (redact scanner, redact CLI, ingest), with a test (`test_the_containment_predicate_has_exactly_one_definition`) pinning object identity so a fourth copy can't silently reappear.
- Correctly identifies and fixes the "refusal collected into `_discover_safe` → empty discovery → `tomb_missing_units` mass-tombs the whole source" hazard, with a dedicated regression test.
- Finds and fixes a genuine Python-version-dependent correctness bug: `Path.resolve(strict=False)` on symlink cycles diverges between 3.11/3.12 (raises) and 3.13 (returns a partially-resolved, in-root-looking path), which would have silently reopened the vulnerability on 3.13 for a specific loop shape. Replaced with `os.path.realpath(..., strict=True)`, with tests for both interpreter behaviors.
- `EscapingSymlinkError` deliberately never discloses the resolved target, with a dedicated non-disclosure test checking both the exception message and captured log records.
- Ordering is handled correctly where addressed: `named_path_escapes` is checked before `is_dir()` (which follows links), and the CLI `ingest` command's containment check is placed above `_gate_consent` specifically because the consent-summary walk itself reads through symlinks.
- Registry-driven test (`test_every_registered_ingestor_refuses_an_escaping_link`) means a 12th ingestor is covered automatically.
- Very good non-vacuity discipline: separate "clean control" tests prove each fixture is actually claimed by its ingestor, so the "sentinel absent" assertions aren't vacuously true.

## Security Concerns
🔴 **BLOCKING — `creek process`'s consent gate still performs the out-of-root read this PR set out to eliminate.**

`creek/cli.py`'s `process` command calls `_gate_consent(source_path=source_path, ...)` (`cli.py:1211`) *before* `pipeline.run(...)`. For a first-time (unconsented) source, `_gate_consent` calls `manager.get_source_summary(source_path, exclusions=[])` → `creek/consent.py:118 _build_source_summary`, which does:

```python
files = [p for p in sorted(source_path.rglob("*")) if p.is_file() and ...]
total_size = sum(f.stat().st_size for f in files)
...
sample_filenames = [f.name for f in files[:_MAX_SAMPLE_FILENAMES]]
```

`rglob("*")`, `is_file()`, and `.stat()` all follow symlinks, and none of this is guarded by `assert_source_contained`. This is the *identical* hazard the PR explicitly names and fixes for `creek ingest` (test `test_cli_ingest_refuses_before_the_consent_gate_reads_the_source_tree`, and the `_assert_ingest_source_contained(input)` call inserted at `cli.py:1355`, above `_gate_consent`) — but no equivalent call was added before `_gate_consent` at `cli.py:1211` in `process`.

Concretely: running `creek process` for the first time against a source tree containing an escaping symlink prints `Sample: <out-of-tree-filename>` to the console and records an inflated `file_count` (counting out-of-tree files) into the persisted consent log — before the #1087 redaction check or the new #1294 ingest gate ever run, regardless of `redaction.enabled`. `docs/ingestion.md`'s new section claims "This applies to every `--type`, to `creek process`, and to the `creek.ingest` MCP tool" — which is true of the *ingest* gate itself, but not of this earlier consent-summary read that `process` performs first.

Fix: call `_assert_ingest_source_contained(source_path)` (or equivalent) before `_gate_consent` in the `process` command, mirroring what was done for `ingest`, and add a CLI-level regression test analogous to `test_cli_ingest_refuses_before_the_consent_gate_reads_the_source_tree` for `process`.

🟢 **LOW — minor resolution-primitive inconsistency.** `find_escaping_symlink` and `named_path_escapes` compute `resolved_root` via `Path.resolve(strict=False)`, while `resolves_within`'s target-side resolution deliberately switched to `os.path.realpath(..., strict=True/False)` specifically to avoid `Path.resolve`'s unspecified, version-dependent cycle behavior (per the module's own extensive docstring). In practice this is very unlikely to be exploitable, since the root is a real, non-cyclic path the operator supplied, but using two different resolution primitives for the two sides of the same comparison is a slight inconsistency the module's own stated design principle argues against.

## Problems
See the blocking security concern above — that is the one merge-blocking issue found. No other correctness bugs were found in `_containment.py`, `base.py`, `markdown.py`, `pipeline.py`, `cli_commands.py`, `scanner.py`, or the MCP tool wiring; the discord call sites that raise a raw traceback instead of a formatted refusal (filed as #1377) are a presentation-only nit, correctly triaged as non-blocking by the author.

## Code Quality
- Docstrings throughout `_containment.py` are exceptionally thorough — arguably close to the point of being essay-length for what are fairly small functions, but the reasoning captured (especially around the 3.13 `Path.resolve`/`os.path.realpath` divergence) is genuinely load-bearing and worth keeping.
- The `_named_path_escapes = named_path_escapes` re-binding in `cli_commands.py` for backward-compat naming is a reasonable, minimal shim given #1293's existing tests reference that name.
- Minor: consider unifying the root-resolution primitive in `find_escaping_symlink`/`named_path_escapes` with `_resolved_target`'s `os.path.realpath` approach for full consistency (see Security Concerns, LOW item) — non-blocking.

## CLAUDE.md Adherence
✅ TDD workflow evident from PR description (RED tests before fix, mutation-tested).
✅ Coverage: reported 94.40%, above the 90% branch-coverage threshold; per-file gate reported green.
✅ Docstring coverage: extensive Google-style docstrings on all new public/private functions.
✅ Commit branch is a feature branch (`issue/1294-issue`), not direct-to-main.
✅ `./scripts/check-all.sh` reported as passing (8956 passed, 5 skipped) per PR description — not independently re-run in this review.
❓ Cannot independently verify mypy strict / ruff zero-violations / complexity ≤10 without running the scripts in this session; nothing in the diff suggests an obvious violation (functions are reasonably small and fully typed).

## Testing
Testing is a clear strength of this PR: 39 new tests in `tests/test_ingest_symlink_containment.py` covering the primary defect (written vault fragment, not just ingestor return value), all 11 registered ingestors via the registry, the `CodeIngestor` directory-recursion case, the named-root-is-itself-a-link case, the mass-tombing hazard, CLI consent-gate ordering (for `ingest`), dangling/looping-link edge cases (including the 3.13-specific regression), non-disclosure, and object-identity of the shared predicate — plus over-breadth guards proving intra-tree aliases still work. Mutation testing (9 mutations, 8 killed, 2 survivors found and closed with new tests) is a notably rigorous practice.

The one gap: there is no test exercising the `creek process` CLI command's consent-gate ordering the way `test_cli_ingest_refuses_before_the_consent_gate_reads_the_source_tree` does for `ingest` — which is exactly the missing coverage that would have caught the blocking issue above.

## Documentation
`docs/ingestion.md`, `docs/redaction.md`, and `docs/security/threat-model.md` are all updated substantively and accurately describe the fixed behavior for the surfaces that were actually fixed. The `docs/ingestion.md` claim that containment "applies to ... `creek process`" is accurate for the ingest-level gate but glosses over the earlier, still-open consent-summary read described above; once that's fixed, the docs will be accurate as written.



## Verdict: CHANGES_REQUESTED

The `creek process` CLI command still performs an unguarded, symlink-following read (via `_gate_consent` → `_build_source_summary`'s `rglob("*")`/`stat()`) before any containment check runs — the exact class of "read-before-refuse" bug this PR explicitly identifies and fixes for `creek ingest` (test `test_cli_ingest_refuses_before_the_consent_gate_reads_the_source_tree`), but the equivalent `_assert_ingest_source_contained` call was never added before `_gate_consent` in `process` (`cli.py:1211`). This leaks out-of-tree filenames to the console and an inflated file count into the persisted consent log on a first-time run, undermining the PR's own stated goal.

Geoffe-Ga:
**CI**: 13/13 Green
**Action**: pull comment 5249797642 to see in-depth feedback and continue iterating

Geoffe-Ga:
Finding confirmed and fixed in 9f3288e. You were right on both counts — the gap and the reason it slipped.

**Verified.** `creek/cli.py:1211` called `_gate_consent` with no containment check anywhere above it on the `process` path; `creek/consent.py:118 _build_source_summary` does `sorted(source_path.rglob("*"))` + `f.stat().st_size`, both of which follow links.

**The durable half, measured at HEAD.** A refused `creek process --yes` still left this behind:

```json
{ "source_type": "pipeline", "source_path": ".../lane/src",
  "file_count": 2, "operator": "... (assume_yes)" }
```

against a source root holding exactly **one** file — the second was the escaping link, counted as a file and its out-of-tree target `stat()`ed. And with `--source` itself a link to an outside directory, the console printed `Sample: CANARY-INGEST-ESCAPED-1294-a7f3-filename.md` and the log recorded that directory's file count. `check_consent()` matches on `(source_type, source_path)`, so the record then suppresses the prompt on every later run — the leak outlives the process.

**Fix.** `_assert_ingest_source_contained(source_path)` above `_gate_consent`, same helper and same predicate as `ingest` — no second convention. Its rationale moved from a comment at the ingest call site into the helper's docstring so both callers share one copy.

**Third surface: there is none.** `_gate_consent` has exactly two callers in `creek/` and `creek_mcp/` (`process`, `ingest`), and `get_source_summary` / `ConsentManager(...)` have one construction site each, both inside `_gate_consent`. Stating it rather than leaving it implied, since "fixed one of two" is what happened last round.

**Rule, not instances.** `test_every_consent_gate_caller_checks_containment_first` parses `creek/cli.py`'s AST and fails if any `_gate_consent` call lacks a containment guard above it in the same function, so a *future* third caller cannot land unguarded in silence. Its limit is stated in its docstring: it judges by line order, so a guard hidden in a conditional would satisfy it — which is why the two behavioural tests exist alongside it.

**Mutation battery.** Reverting only the added call turns exactly the three new tests red; 9328 others pass.

**Docs.** `docs/ingestion.md`'s "applies to … `creek process`" is now true as written, and the section says *when* the refusal lands (above the consent prompt) and what that buys — no out-of-tree filename printed, no link-measured count persisted.

Gate 2 exit 0. Bypass sweep across the branch diff including tests: no `noqa`, `type: ignore`, `nosec`, skip/xfail, threshold change, or version gate added. The one pre-existing `# pragma: no cover` in the diff is on a test-only failure-message helper reached only when a test is already failing.

Geoffe-Ga:
## Summary
This PR closes a real, demonstrated symlink-escape vulnerability: all 11 registered ingestors discovered their inputs via unguarded walks (`rglob`, `glob`, `iterdir`) and read whatever a symlink pointed to, including targets outside the named source root — writing that content into the vault as durable fragments that then feed classification, embeddings, and cloud LLM prompts. The fix introduces one shared containment predicate (`creek/_containment.py`), gates it at the single `Ingestor.ingest` chokepoint (covering all ingestors, present and future, via `INGESTOR_REGISTRY`), and additionally closes two read-before-refuse gaps in the CLI's consent-summary path (`creek ingest` and `creek process`). A follow-up commit fixes a Python-3.13-specific soundness hole in the containment primitive itself, caught by CI.

## Strengths
- Concrete, reproducible proof of the vulnerability in the PR description (sentinel strings actually landing in vault fragments on `main`), not a theoretical concern.
- One containment predicate (`resolves_within`) shared by all three surfaces (ingest gate, redact scanner, redact CLI), with a test (`test_the_containment_predicate_has_exactly_one_definition`) asserting object identity so a fourth copy can't silently reappear.
- Correctly identifies and fixes the worst-case offender: `CodeIngestor`'s hand-rolled `iterdir()`/`is_dir()` recursion, which (unlike `rglob`) follows symlinked directories and would otherwise walk an entire out-of-tree subtree. The gate's `os.walk(followlinks=False)` inspecting both `dirnames` and `filenames` is what closes this without touching `code.py`.
- Identifies and fixes a genuinely subtle hazard: a swallowed refusal inside `_discover_safe` would leave `seen_keys` empty and cause `tomb_missing_units` to soft-tomb every previously-ingested fragment — a containment fix that would otherwise become a mass-deletion primitive. Pinned by `test_a_containment_refusal_does_not_tomb_previously_ingested_fragments`.
- Non-disclosure is treated as a first-class requirement: the refusal error never names the resolved out-of-tree target, only the link as walked — tested explicitly, including against log records, guarding against the exact oracle #1087 was meant to close.
- The Python 3.13 fix (commit 2) is a good catch: `Path.resolve(strict=False)` on a symlink cycle silently returns a partially-resolved, in-root-looking path on 3.13 vs. raising on 3.11/3.12. Switching to `os.path.realpath(..., strict=True)` gives a version-stable contract, and the regression tests assert on the verdict rather than on which exception type fires — genuinely version-independent rather than version-skipped.
- Commit 3 catches a real gap the first pass missed: `creek process` also calls `_gate_consent` (which walks the tree via `rglob` + `stat()` before anything refuses) but wasn't gated. A structural AST test (`test_every_consent_gate_caller_checks_containment_first`) now enforces the ordering rule against the function body itself, not just against the two call sites known today — good defense against a future third caller landing unguarded.
- Refusal is one-way with no bypass (no `--force`, env var, or config key), correctly matching the write-path-refuses / read-path-skips split established by #1293/#1360, with a clear rationale for why ingest belongs on the refuse side.
- Extensive over-breadth guards (intra-tree alias, root reached through a symlinked component, dangling links, `rglob` non-recursion tripwire) prevent the fix from becoming "refuse any symlink," which would break ordinary aliasing.
- Docs (`ingestion.md`, `redaction.md`, `threat-model.md`) and the `SEC-003` threat-model entry are updated to match the new behavior, not just the code.
- Honest, explicit acknowledgment of residual scope: the escaping-ancestor-component case and `CodeIngestor`'s intra-root symlink-loop bug are called out and filed separately (#1375) rather than silently left unhandled or overclaimed as fixed.

## Security Concerns
🟢 LOW — `creek_mcp/tools/upload.py`'s `_run_ingest_for_upload` (around line 530) calls `run_ingest(...)` and only catches `FileNotFoundError`, not the new `EscapingSymlinkError`. If it ever propagated, the MCP tool would crash with a transport-level exception instead of returning a structured `refusal_response`, the same failure mode the PR explicitly fixed for `creek_mcp/tools/ingest.py` and flagged (as a nit, #1377) for the Discord call sites. In practice this looks unreachable today — `staged` is always a path freshly written by `_stage_upload` inside the vault, never an attacker-influenced symlink — so this isn't a live vulnerability, just a defense-in-depth/consistency gap on a surface this PR otherwise made a point of hardening everywhere else.

## Problems
None found that block merge. The change is narrowly scoped to the stated defect, doesn't weaken any existing guard (#1087's pipeline-level refusal is explicitly preserved as a second line of defense, not replaced), and the security direction is one-way throughout.

## Code Quality
- The `upload.py` gap above is the one item worth a follow-up: either catch `EscapingSymlinkError` there for consistency with `creek_mcp/tools/ingest.py`, or fold it into the already-filed #1377 alongside the Discord call sites, since it's the same category of issue (guard inherited and fails closed, but surfaces as a raw exception instead of the surface's normal refusal idiom).
- Everything else — module boundaries (stdlib-only `_containment.py` to avoid dragging the regex-heavy `redact.scanner` into the ingest import path), naming, docstrings, and the deliberate choice to return the offending path from `find_escaping_symlink` rather than raise (so two different call sites can each apply their own refusal idiom) — is clean and well-reasoned.

## CLAUDE.md Adherence
✅ Conventional commit messages (`fix(ingest): ...`, `fix(cli): ...`) with clear rationale bodies.
✅ Per the PR's test-plan report: `./scripts/check-all.sh` exit 0, coverage 94.40% (above the 90% floor), per-file gate green — I could not independently re-run CI from this environment, but the reported numbers and the visible test file are consistent with the quality bar.
✅ No stdlib-vs-project dependency violations; `_containment.py` is explicitly kept stdlib-only with a documented rationale for the one-way dependency direction (`ingest.base` → `_containment`, `redact.scanner` → `_containment`, not the reverse).
✅ Docstrings are present and detailed throughout the new module and CLI helpers, consistent with the ≥95% interrogate requirement.

## Testing
Exceptionally thorough for the risk level. `tests/test_ingest_symlink_containment.py` (2028 lines, 39 tests) covers: the primary RED asserting on the actual written vault fragment (not just a glob's return value or an in-memory result); non-vacuity controls per ingestor (proving each fixture is actually claimed before asserting the sentinel is absent); a registry-driven test that automatically covers a 12th ingestor added later; the `CodeIngestor` symlinked-directory case; the named-root-is-itself-a-link case; the mass-tombing hazard; CLI ordering pins for both `creek ingest` and `creek process` (with a recording, non-raising spy so the ordering assertion can't pass for the wrong reason); MCP-surface refusal; dangling/looping-link edge cases; non-disclosure of the resolved target; and a structural AST-based test enforcing the ordering invariant going forward. The PR description reports a 9-mutation battery with 8 killed and the two survivors backfilled with new tests, plus one genuinely dead `except` clause found and removed. This is well above the bar the project sets for testing thoroughness.

## Documentation
`docs/ingestion.md`, `docs/redaction.md`, and `docs/security/threat-model.md` are all updated to describe the new refusal behavior, its ordering relative to the consent prompt, and the SEC-003 status change — and the threat-model narrative correctly distinguishes what's now resolved from what's still an open, separately-tracked residual (#1375). Nothing found stale or contradicted by the code.



## Verdict: COMMENTS

No blocking bugs, security holes, or test/doc gaps — the fix is well-scoped, extensively tested (including mutation testing and a structural regression test), and the security direction is one-way throughout. It stays out of LGTM only because of one concrete, worth-weighing suggestion: `creek_mcp/tools/upload.py`'s ingest call site doesn't catch the new `EscapingSymlinkError` the way `creek_mcp/tools/ingest.py` now does, which is inconsistent with how thoroughly this PR hardened every other entry point (it's not currently exploitable since `staged` is never attacker-controlled, but it's the same category of gap the author already flagged and filed for the Discord call sites as #1377).

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5249997409 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5249997409 to see in-depth feedback and continue iterating

Issue #1382, opened 2026-08-11
fix(link): a failed cluster-page write still stamps its wikilink onto every member fragment


Geoffe-Ga:
## Problem

In both `_run_eddies` and `_run_threads` (`creek/link/link_engine.py`), the order is:

1. `detector.assign_fragments_to_eddies(fragments, eddies)` — builds a `[[Title]]` wikilink for **every detected** cluster, unconditionally (`creek/link/eddies.py`, `assign_fragments_to_eddies`; same shape in `threads.py`).
2. `_materialise_link_models(...)` — writes one page per cluster, and **swallows `OSError` per model**, returning only the paths that succeeded.
3. `_persist_fragment_link_updates(...)` — rewrites member frontmatter using the fragment list from step 1.

Step 3 does not know which pages step 2 failed on. So if a page write fails — full disk, read-only volume, a permission problem on one file — the member fragments are still rewritten to cite `[[Title]]` for a page that does not exist. The result is a silent dangling wikilink in the vault.

The reported counts stay honest (`eddies_written`/`threads_written` count only successful writes, which is what `creek process`'s `Link artefacts persisted: N` is sourced from), so this is not an over-reporting bug. It is a vault-corruption-on-partial-failure bug: the summary correctly says "1 of 2 written" while 12 fragments now point at both.

## Why now

The line-level ordering predates #1303 — it is shared with `creek link`, where a human is watching the output and can act on a warning. #1303 wired the same code into `creek process`, which is the command people put on a schedule and do not watch. That raises the blast radius enough to be worth tracking rather than leaving implicit.

## Acceptance criteria

- [ ] `_persist_fragment_link_updates` only writes wikilinks for clusters whose page was actually materialised. The natural shape is to filter the detected models down to the successfully-written set *before* calling `assign_fragments_to_*`, so the in-memory fragment list never contains a link to a page that does not exist.
- [ ] A fragment that would have been linked only to a failed-write cluster is left byte-identical, not rewritten with an empty list.
- [ ] The `member_fragments_updated` count reflects the filtered reality.
- [ ] Test: force `write_eddy` to raise `OSError` for exactly one of two detected eddies; assert (a) the surviving eddy's members carry its link, (b) the failed eddy's members carry **no** `eddies:` entry, (c) no fragment cites a page missing from disk, (d) `eddies_written == 1`.
- [ ] Same for threads.
[...]

Issue #1381, opened 2026-08-11
fix(link): membership-derived cluster ids orphan a page on every membership change, and creek process now does it on a schedule


Geoffe-Ga:
## Problem

Eddy and thread page ids are derived from their sorted member fragment ids:

- `creek/link/eddies.py` — `_stable_eddy_id`: `sha256(",".join(sorted(f.id for f in frags)))[:8]`, prefixed `eddy-`.
- `creek/link/threads.py` — `_stable_thread_id`: the identical formula, prefixed `thread-`.

That is what makes re-running idempotent on an *unchanged* corpus: the same membership re-mints the same id, the same filename, and `VaultWriter`'s per-directory id-index dedupes. But when membership *changes* — you ingest one more note into an existing cluster — the id changes too. A **new** page is written and the previous one is left on disk, orphaned: no fragment's `eddies:`/`threads:` frontmatter points at it any more, because `_persist_fragment_link_updates` rewrote those lists to the new title.

## Why it matters more now

Before #1303, only `creek link` minted these pages, so the accumulation was bounded by how often an operator ran it by hand. #1303 made `creek process` persist its link stage, and `creek process` is the command people put on a schedule. On a vault that grows a little every day, `03-Eddies/` and `02-Threads/*/` will grow monotonically with dead pages, one per membership change per cluster.

`creek lint`'s broken-links check is the only current detector, and it detects the *symptom* (a fragment pointing at a missing page), not this shape (a page nobody points at).

## Why it was not fixed in #1303

Deleting vault files is destructive and irreversible from the tool's point of view — an operator may have edited an eddy page by hand, and a reaper that cannot tell "superseded by a new membership id" from "hand-curated page the detector no longer produces" would silently destroy work. That needs its own design (dry-run first, provenance check, probably an `archive/` move rather than an unlink), which is well outside the diff for a linking-persistence fix.

## Acceptance criteria

- [ ] Decide and document the disposal semantics: delete, or move to an archive location, or mark with frontmatter and leave in place. Default to the least destructive that still solves the problem.
- [ ] A page is only ever a reap candidate when (a) it lives under `03-Eddies/` or `02-Threads/{Active,Dormant,Resolved}/`, (b) its frontmatter `id` matches the `eddy-`/`thread-` membership-hash shape, and (c) **no** fragment in the vault cites its title in `eddies:`/`threads:`. A hand-written page with no membership-shaped id is never a candidate.
[...]

PR #1383, opened 2026-08-11
fix(pipeline): make `creek process` Stage 5 persist the link graph it computes

### LINKED ISSUE #1303: fix(pipeline): creek process Stage 5 computes the whole link graph, discards it, and reports 'Links found: N'
**Severity:** High
**Remediation:** decision-needed (delegate to `link_engine`, or persist) + e2e test

## Problem

`creek process` Stage 5 runs all four linkers, computes the entire link graph, throws every
result away, and then prints the count to the operator as if the links had been created.

`LinkingPipeline.run` (`creek/link/linker.py:63-146`) generates embeddings, finds
resonances, detects threads and eddies, calls `assign_fragments_to_threads` /
`assign_fragments_to_eddies` into a **local** `fragments` variable — and returns a
counts-only `LinkingResult`. It writes no thread page, no eddy page, no fragment
frontmatter, and no embeddings cache. `vault_path` is accepted and used only in a log line;
its own docstring concedes this at `linker.py:74`: `vault_path: Path to the Obsidian vault
(used for future I/O)`.

Meanwhile `creek/cli.py:1181` prints `Links found: {result.links_found}`.

Taxonomy family 3 (fake/aspirational implementation) + family 12 (architectural drift: two
orchestrators for one concept, only one of which is real).

**Current state:** the operator is told N links were found; `02-Threads/`, `03-Eddies/` and
every fragment's `threads:` / `eddies:` frontmatter are untouched. Because no embeddings
cache is written either, the next `creek link` re-embeds the whole corpus from scratch.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** `LinkingPipeline.run` over 6 fragments into a
scaffolded vault, with `generate_embeddings` stubbed to deterministic vectors:

```
REPORTED (what `Links found:` sums)  = 6
  resonances=6 temporal=0 threads=0 eddies=0
FILES WRITTEN UNDER VAULT           = []
```

Non-zero counts reported; zero bytes written. Repro:
```python
from unittest.mock import patch
from creek.link.linker import LinkingPipeline
from creek.config import EmbeddingsConfig, LinkingConfig
with patch("creek.link.linker.EmbeddingLinker.generate_embeddings", return_value=vecs):
    res = LinkingPipeline(EmbeddingsConfig(), LinkingConfig()).run(frags, vault)
assert res.resonance_count > 0
assert [p for p in vault.rglob("*") if p.is_file()] == []
```

**Signal 2 — structural, and there is a working twin.** The call chain is
`creek process` → `Pipeline.run` (`creek/pipeline.py:338-341`) → `Pipeline._run_linking`
(`creek/pipeline.py:765-798`) → `LinkingPipeline.run`. The *other* orchestrator,
`link_engine.run_link` (the `creek link` command's path), does all the persistence the
pipeline one skips: `_materialise_link_models` (`link_engine.py:254`, `:341`) and
`_persist_fragment_link_updates` (`link_engine.py:260`, `:347`), plus
`_load_or_compute_embeddings` (`link_engine.py:542`) reading and writing the parquet cache.
Issue #880 unified the two on *thresholds* only (`from_linking_config`); the persistence
contract still diverges completely.

**Signal 3 — no test covers it.** Grep over `tests/e2e/test_pipeline_markdown_e2e.py`,
`test_full_pipeline_markdown.py`, `test_full_pipeline_idempotency.py` finds zero assertions
about `02-Threads/`, `03-Eddies/`, or `links_found` from `creek process`.

Independently reached by two separate readers in this scan plus the reproducing run above.

## Scope

Covers `creek process` Stage 5 — `Pipeline._run_linking` and `creek/link/linker.py`.

Explicitly NOT in scope: the known `O(n^2)` hotspots in `link/temporal.py` and
`link/threads.py` (#857, #858), the temporal score clamp (#1216), or the
`similarity_edges`-not-cached claim on the `creek link` path (filed separately).

## Tasks

1. Decide the direction and record it on this issue:
   - **(a)** Have `Pipeline._run_linking` delegate to `creek.link.link_engine.run_link` per
     method, deleting `LinkingPipeline` — one orchestrator, one persistence contract; or
   - **(b)** Give `LinkingPipeline.run` the same persistence `link_engine` has
     (`_materialise_link_models` + `_persist_fragment_link_updates` + embeddings cache).
   (a) is preferred — it removes the drift rather than duplicating the fix.
2. Implement the chosen direction.
3. Delete the now-dead `LinkingPipeline.add_wikilinks` (`creek/link/linker.py:148`), which
   has zero production callers (only `tests/test_link.py:2053-2123`) and duplicates the
   dedup-append loop already in `threads.assign_fragments_to_threads` (`threads.py:769-773`)
   and `eddies.assign_fragments_to_eddies` (`eddies.py:767-771`).
4. Fix the stale `vault_path: … (used for future I/O)` docstring, or remove the parameter.

## Acceptance Criteria

- [ ] An e2e test drives `creek process` through Typer's `CliRunner` over a small markdown
      source and asserts that thread/eddy pages exist on disk and that at least one
      fragment's frontmatter carries a `threads:` or `eddies:` wikilink.
- [ ] `Links found: N` is non-zero only when N links were actually persisted.
- [ ] A second `creek process` run reuses the embeddings cache rather than re-embedding.
- [ ] `LinkingPipeline.add_wikilinks` is gone (or has a production caller).

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/pipeline.py` | Modify (`_run_linking`) |
| `creek-tools/creek/link/linker.py` | Modify or Delete |
| `creek-tools/tests/e2e/test_full_pipeline_markdown.py` | Modify (add the persistence assertions) |
| `creek-tools/tests/test_link.py` | Modify (drop `add_wikilinks` tests) |

Geoffe-Ga:
## Summary

`creek process` Stage 5 computed the entire link graph in memory and threw it away, then printed `Links found: N`. Reproduced at HEAD on a 12-file markdown corpus: **`links_found = 32`**, and the vault gained **zero** eddy pages, **zero** thread pages, **zero** `eddies:`/`threads:` frontmatter, and **no** `00-Creek-Meta/embeddings.parquet`. The only files that appeared under `02-Threads/` and `03-Eddies/` were `Thread-Index.md` and `Eddy-Map.md` — both written by Stage 6's `IndexGenerator`, neither anything to do with linking. Running `run_link` over the *same* vault immediately produced 2 eddy pages, 2 thread pages and 12 frontmatter rewrites. The twin worked; Stage 5 did not.

Stage 5 now delegates to `creek.link.link_engine.run_link`, the same entry point `creek link`, `creek fill` and `creek sync` already use. `creek/link/linker.py` — the second, count-only orchestrator — is deleted, so there is exactly one linking code path in the codebase.

### 1. Resonance and temporal are dropped, not relabelled — and AC #2 as written could not have been satisfied by delegation alone

The issue asked that "`Links found: N` is non-zero only when N links were actually persisted." **Delegating to `run_link` cannot deliver that.** `method="temporal"` returns a count and writes nothing whatsoever. `method="embeddings"` persists only the embedding *vector* parquet — the resonance edges it counts are computed in memory and dropped, because there is no `Resonance`/`TemporalLink` writer anywhere in the codebase and `Fragment` has no `resonances` field (only `threads` and `eddies`). In the reproduction the reported 32 broke down as 30 resonances + 2 eddies: the headline was *dominated* by the one quantity nothing can persist. A PR that delegated and stopped would ship the same overclaim wearing a different hat.

So `creek process` runs **two** methods, in this order: **`eddies`, then `threads`**. `embeddings` and `temporal` are dropped from the pipeline entirely — the strongest form of the honest-reporting criterion. The vector cache `embeddings` would have warmed is written by the eddies pass anyway (via `_load_or_compute_embeddings` → `_persist_cache`), and its `find_resonances` call is the O(n²) hotspot of #857/#858 that this issue declares out of scope. `creek link --method embeddings|temporal` is unchanged and still available.

The operator-facing output changes to match. `Links found: N` is replaced by the same per-stage phrasing `creek link` already uses (one shared `_format_link_summary`), followed by `Link artefacts persisted: N` sourced from `eddies_written + threads_written` — **`*_written`, never `*_detected`**, so a cluster the writer failed on cannot be reported as a success. Two arms of `_format_link_summary` were also overclaiming for `creek link` and are corrected: the embeddings arm said the *edges* were cached in the parquet (only the vectors are), and the temporal arm implied persistence where there is none.

### 2. Stage 5 now links the whole vault, deliberately

`run_link` calls `_load_fragments(vault_path)`, so `creek process` re-clusters the entire corpus rather than only the fragments this run ingested. **Accepted**, for two reasons: a newly ingested note must be able to join an eddy of notes ingested months ago, and cluster ids are membership-derived, so they only stay coherent whole-vault. The cost is one DBSCAN pass plus one union-find pass per `creek process` — exactly `creek link --method eddies` followed by `--method threads`, with no `find_resonances`. A run that ingested nothing short-circuits and does not re-cluster.

This also invalidated `result.local_model_processed += len(fragments)`, which billed the run for every fragment handed to the linker whether or not the model ran. It is now `+= summary.fragments_embedded`, i.e. vectors the sentence-transformer actually computed; cache hits are not counted. **This changes the numbers landing in `00-Creek-Meta/run-summary.jsonl`** via `PreLLMYieldSummary` — a warm-cache re-run now honestly reports near-zero local-model work instead of re-billing the corpus. The audit tests were checked: nothing asserts a specific `local_model_processed` value, only its presence and type.

### 3. The ordered two-call structure is a correctness requirement, not a style choice

`_persist_fragment_link_updates` rewrites frontmatter from the in-memory model's **full** dump (`raw.update(after.model_dump(mode="json"))`), so it overwrites *every* Fragment-owned key — including the link list the *other* stage owns. Any design that loads the corpus once and feeds both detectors off that single list would have the second stage write `threads: []` over the wiki-links the first stage had just persisted. It would still pass a page-existence assertion, and `member_fragments_updated` would still be non-zero: the data loss would ship green.

The per-call reload inside `run_link` is what makes it safe, and `_PROCESS_LINK_METHODS` carries that reasoning in its docstring. The guard is the `linked_both` block in the new e2e: every member fragment re-read **off disk** must carry both its `[[eddy]]` link in `eddies:` **and** its `[[thread]]` link in `threads:`, asserted together. Mutation-tested — memoising `_load_fragments` to simulate the shared-load design fails exactly that assertion (see below).

### 4. Deleted tests, one line each

The no-deleted-tests rule is waived only where the class under test ceases to exist.

| Deleted | Why | What carries the behaviour |
|---|---|---|
| `TestLinkingResult` (3 tests) | `LinkingResult` no longer exists | `LinkSummary` is covered throughout `tests/test_link_engine.py` |
| `TestLinkingPipeline::test_init_stores_configs` | class deleted | — |
| `TestLinkingPipeline::test_run_returns_linking_result` | class deleted | `test_run_linking_runs_eddies_then_threads` |
| **`TestLinkingPipeline::test_run_returns_zero_temporal_thread_eddy`** | **pinned the defect** — docstring "Pipeline.run should return zero counts for still-stubbed linkers" | the whole new e2e |
| `TestLinkingPipeline::test_run_empty_fragments` | class deleted | `test_run_linking_no_fragments`, now asserting `run_link` is not even called |
| `TestLinkingPipeline::test_run_logs_pipeline_stages` | class deleted | `test_run_linking_runs_eddies_then_threads` pins the stage list |
| 5 × `add_wikilinks` tests | `add_wikilinks` had **zero** production callers — the only references anywhere were these five tests | `assign_fragments_to_*` + `_persist_fragment_link_updates` in `tests/test_link_engine.py` |
| `test_creates_linking_pipeline` | asserted `pipeline.linking_pipeline is not None`; that attribute *was* the defect | replaced by `test_holds_no_second_linking_orchestrator`, which asserts the opposite |

Four of those `LinkingPipeline` tests passed `vault_path=Path("/fake/vault")`. None survive into a code path that does I/O.

Two existing assertions were changed rather than deleted:

- `tests/test_pipeline.py` — `0 <= links_found <= fragments_created**2` no longer holds now that Stage 5 links the whole vault and counts *pages*, not pairwise edges. Replaced with the exact identity `links_found == sum(eddies_written + threads_written)`, plus a new `test_links_found_counts_written_not_detected` that forces the two counts apart (3 detected, 1 written) so the identity cannot be satisfied by a detected-count regression.
- `tests/test_pipeline.py:747-748` (`Thread-Index.md` / `Eddy-Map.md` exist) is left alone but now carries a comment saying these are `IndexGenerator` artefacts and are **not** link-stage evidence — they were green throughout the bug.

### 5. One pre-existing flake this change started firing

`tests/e2e/test_full_pipeline_idempotency.py` counted every `.md` outside `00-Creek-Meta`, which includes `review-queue-<YYYY-MM-DD_HHMMSS>.md` — written to the vault root once per invocation, with a **second-resolution timestamp in its filename**. Two runs that straddle a second boundary therefore always leave two of them. The sentinel passed only while the pipeline was fast enough to finish both runs inside the same second; giving the link stage real work to do started tripping it (~2 in 5 runs). Proved deterministically by patching the clock forward one second: the count goes 16 → 17 with a correct pipeline. The counter now excludes that one filename, which narrows the assertion to what it always meant — *fragments and cluster pages* must not duplicate — and the counter now also covers the eddy/thread pages Stage 5 writes, so a link stage that re-minted a page per run would fail here.

### 6. Follow-ups filed

- **#1381** (P2) — eddy/thread ids are membership-derived, so growing a cluster by one fragment mints a new page and orphans the previous one. `creek process` on a schedule makes that accumulate. Not fixed here: deleting vault files is destructive, needs its own design (dry-run, hand-edit detection, archive-vs-unlink), and would blow this diff. Documented in `docs/linking.md` in the meantime.
- **#1382** (P2) — found during this PR's self-review. `assign_fragments_to_*` builds a wikilink for every *detected* cluster, but `_materialise_link_models` swallows a per-page `OSError`; the frontmatter rewrite that follows does not know which pages failed, so a partial write leaves member fragments citing a page that does not exist. The *counts* stay honest (`eddies_written` only counts successes), so this is vault corruption on partial failure, not over-reporting. The ordering predates this PR and is shared with `creek link`; what changes is that `creek process` is the unattended path. Fixing it properly means reordering the filter, which is out of scope here.

### 7. Gate 2.5 self-review

Dispatched `code-review-orchestrator` over the staged diff. It returned **one blocker**, which was **verified by execution and fixed**, plus several non-blocking findings, all addressed:

- **BLOCKER — `test_process_preserves_foreign_frontmatter_and_body` was vacuous.** It picked the first fragment carrying an eddy link, which is alphabetically a *topic-B* fragment, while the trigger note added before the second run is *topic-A*. Only the topic whose membership changes gets rewritten, so the file was **never touched** and the preservation assertions passed on an untouched file. Confirmed by execution (`REWRITTEN? False`). Fixed: the member is now selected from topic A, and the test asserts the eddy list actually changed *before* asserting preservation — that assertion alone would have caught the original bug.
- **Clobber guard was a lower bound.** `>= 6` over 12 fragments would still pass if one whole topic silently failed to link. Tightened to `== len(fragments)`, with an explicit `len(fragments) == 12` guard so a stray fragment cannot loosen it.
- **Dangling-wikilink check pooled the two title sets.** Now checked per key against that stage's own pages, plus an arity assertion (exactly one link per kind per fragment).
- **Misleading docstring** on the unscaffolded-vault test: it claimed to exercise `_materialise_link_models`'s `FileNotFoundError` swallow. Verified by execution that on a bare vault `_load_fragments` returns empty and `run_link` short-circuits before reaching it (`MATERIALISE_CALLS 0`). Docstring corrected to state what the test actually pins; the swallow itself is covered in `test_link_engine.py`.
- **Comment mischaracterised the model-unavailable exposure** as newly introduced. The deleted `LinkingPipeline` already loaded the model — the failure was always reachable, just never handled. Reworded.
- **`local_model_processed` docstring** still claimed "/ OCR" though nothing adds an OCR count. Reworded to say so explicitly.
- The reviewer confirmed clean on: the honesty ratchet (no path overstates what reached disk), the deleted `test_link.py` classes, the idempotency narrowing, and suppressions (none).

## Test plan

**RED first, on disk state.** `tests/e2e/test_full_pipeline_linking.py::test_process_persists_eddy_and_thread_pages` drives `creek process --source … --vault … --yes --no-llm` through Typer's `CliRunner` and fails at HEAD with:

```
AssertionError: creek process reported links but wrote no eddy page;
03-Eddies holds only ['.gitkeep', 'Eddy-Map.md'] (#1303).
```

Every assertion is against on-disk state, never a returned count — the whole defect was that the count lied about the disk.

**Both fake-green traps avoided, and the first one demonstrated.** `Eddy-Map.md` and `Thread-Index.md` exist after *every* `creek process`, bug or no bug. Removing the `!= "Eddy-Map.md"` exclusion and re-orphaning Stage 5 was run as an explicit mutation: the eddy assertion **passes**, and the test only fails later on threads. That is exactly the trap the exclusion prevents. Thread pages are found structurally via `02-Threads/*/*.md`, so the root-level `Thread-Index.md` can never satisfy them.

**Reachably green, and every fixture fact is load-bearing:**
- **Six notes per topic, not four** — `eddy_min_fragments` and `eddy_min_samples` are both 5, so a four-note cluster is DBSCAN noise and the test would stay red after a *perfect* fix.
- **Identical unit vectors per topic, zero jitter** — `EddyDetector._has_temporal_direction` discards any cluster whose |Spearman(chronological rank, cosine drift)| reaches 0.3. Identical vectors force every drift to 0.0 so the correlation short-circuits and the cluster survives; jitter makes that filter fire on a large fraction of orderings at n=6, i.e. jitter is what would make this flaky.
- **The topic keyword lives in the H1** — `fragment_embedding_text` returns the title only and the markdown ingestor derives the title from the H1, so the same word drives both the embedding and, at `TITLE_WEIGHT` 3 against `PRIMARY_THRESHOLD` 3, the rule classifier's `frequency.primary`. **Threads are reachable rules-only**, contrary to a claim made during triage: verified by execution, not argument — the 12-note corpus yields `frequency: {primary: F5}` / `{primary: F6}` and 2 threads with `--no-llm`.

The e2e also covers: the parquet exists; both frontmatter lists populated on the same fragment (the clobber guard); no dangling wiki-links; the eddy page's `id` re-derived from its on-disk membership as `sha256(",".join(sorted(member_ids)))[:8]`; the printed counts equal the pages counted on disk; a second run adds no page and rewrites no fragment byte; a warm cache means zero fresh vectors; and a foreign frontmatter key + the body survive a rewrite verbatim.

**Mutation battery** — 19 mutants, 17 killed, 2 verified equivalent (equivalence *checked*, not assumed):

| # | Mutation | Result |
|---|---|---|
| M1 | Stage 5 returns `[]` always (re-orphan the stage) | **killed** — 10 failures across e2e, pipeline, cli, wiring-contract |
| M2 | `_PROCESS_LINK_METHODS = ("eddies",)` | **killed** (5) |
| M3 | `_PROCESS_LINK_METHODS = ("threads",)` | **killed** (6) |
| M4 | order reversed to `("threads", "eddies")` | **killed** (3) — see note below |
| M5 | memoise `_load_fragments` = the single-shared-load design | **killed** (2) — the clobber guard fires |
| M6 | `except EmbeddingModelUnavailableError` → `except RuntimeError` | **EQUIVALENT** — verified: `EmbeddingModelUnavailableError(RuntimeError)`, so the broader clause catches it and behaviour is identical |
| M6b | the handler deleted outright | **killed** (1) |
| M7 | print the total but no per-stage lines | **killed** (2) |
| M8 | `local_model_processed += len(fragments)` | **killed** (1) |
| M9 | `if not fragments` guard removed | **killed** (1) |
| M10 | `links_found` from `*_detected` | **killed** (1) |
| M11 | `EmbeddingPass.computed` = total instead of misses | **killed** (2) |
| M12 | wiring-contract parquet tripwire vs. re-orphaned Stage 5 | **killed** (1) |
| M13 | frontmatter never rewritten | **killed** (2) |
| M14 | eddy id prefix changed | **killed** (1) |
| M15 | pages detected but never recorded as written | **killed** (1) |
| M16 | foreign frontmatter keys dropped on rewrite | **killed** (1) |
| M17 | fragment body dropped on rewrite | **survived → fixed → killed** |
| M18 | eddy link also filed under `threads:` | **EQUIVALENT in this fixture** — see below |

**On M4** — reversing the order is killed only by the explicit `["eddies", "threads"]` assertions, not by any vault-state assertion. That is correct and intended: with the per-call reload, *either* order produces the same on-disk result. What the order assertion pins is that the two calls remain two calls; the vault-state consequence of collapsing them is what M5 covers. Calling this out rather than implying the order is load-bearing on its own.

**On M17 — a real survivor, and why it survived.** Dropping the body on rewrite initially passed. Cause: the test captured `body_before` *after* run 1, and run 1 is itself a rewrite — so the mutation had already emptied the body before the baseline was taken, and `"" == ""` passed. A baseline captured downstream of the corruption proves nothing. Fixed by anchoring on the fixture's own text (`assert "corpus." in body_before`) before comparing; the mutant now dies.

**On M18 — verified equivalent, not assumed.** Filing an eddy's wikilink into `threads:` as well as `eddies:` produces a byte-identical vault *in this fixture*, because `naming` gives the eddy page and the thread page for a topic the **same title**, and `assign_fragments_to_threads` dedupes (`if link not in merged`, `threads.py:774`). The cross-filed link is literally the same string as the correct one and collapses into it, leaving arity 1. No assertion reading this vault can distinguish the two states because there is no difference on disk. The per-key check and the arity assertion added in review still kill this class of bug wherever the two titles differ; making them differ here would mean contorting the corpus and giving up the identical-vector property that keeps this test non-flaky, so it was not pursued.

**Gates.** `./scripts/check-all.sh` → **exit 0**. `./scripts/test.sh --all` → **9424 passed, 5 skipped** (the 5 are live-API smokes with no keys). `ruff check --no-cache` clean. `xenon --max-absolute B --max-modules B --max-average B creek/` (the repo gate) passes; no new B-or-worse block in `link_engine.py` or `pipeline.py`. Per-file coverage gate passes with `link/link_engine.py` and `pipeline.py` both COVERED.

**Not covered / not attempted, stated rather than glossed:**
- Resonance and temporal persistence is not built. They are dropped from `creek process` and labelled as non-persisting in `creek link`'s output; no writer was invented.
- The stale-page reaper is #1381, documented but not built.
- `docs/architecture/ADR/0008` still references `creek.link.linker`. ADRs are historical records of a decision at a point in time and were left unedited; the two live docstrings that referenced the deleted module were updated.
- `docs/wiring-contract.md` has no `process` artefact list to update.
- `EmbeddingPass.reused` has no production consumer — it is pinned by tests and documents the cache split, but no operator-facing "N served from cache" line was added.
- M18 above: cross-filing an eddy link into `threads:` is not detectable in this fixture, for the reason given.
- The dangling-wikilink-on-partial-write ordering (#1382) is tracked, not fixed.

Closes #1303
Refs #857, #858, #880, #1381, #1382


Geoffe-Ga:
## Summary
Fixes #1303: `creek process` Stage 5 used to compute the entire link graph in memory via a second orchestrator (`LinkingPipeline`) and discard it, printing an inflated `Links found: N` while writing zero eddy/thread pages. This PR deletes that dead orchestrator and delegates Stage 5 to `run_link` (the same entry point `creek link`/`fill`/`sync` use), running only the two methods that actually persist (`eddies`, `threads`), and replaces the misleading count with an honest `Link artefacts persisted: N` sourced from `*_written` counters. Verified by reading the full diff against the current tree, confirming no other production references to the deleted module remain, and checking live CI status (Pylint, Integration & E2E Tests, Code Complexity, Build Distribution all green; Tests & Type Checking / Static Analysis still in progress at review time).

## Strengths
- Root-cause fix rather than a patch: deletes the second linking code path entirely instead of teaching it to write, so there is exactly one orchestrator (`run_link`) going forward.
- The per-call `run_link` reload (rather than one shared fragment load feeding both detectors) is correctly identified as a correctness requirement, not a style choice — `_persist_fragment_link_updates` does a full-model dump, so a shared load would have one stage's writer clobber the other's frontmatter list. This is explicitly guarded by a "both eddies and threads present, read off disk" assertion in the new e2e test, and demonstrated failing under an explicit mutation (M5).
- `links_found` is now defined as `*_written`, never `*_detected`, with a dedicated test (`test_links_found_counts_written_not_detected`) that forces the two numbers apart so the identity can't hide behind coincidence.
- `local_model_processed` now bills only vectors actually computed (cache misses) rather than `len(fragments)`, fixing a real (if minor) accounting bug in `run-summary.jsonl`, and is tested for the double-count case across the two sequential `run_link` calls.
- New e2e test (`test_full_pipeline_linking.py`) asserts against on-disk state exclusively (never a returned count), explicitly avoids two identified "fake green" traps (the pre-existing `Eddy-Map.md`/`Thread-Index.md` artifacts from Stage 6, and asserting only the in-memory fragment model), and documents why each fixture parameter (6 notes/topic, identical vectors, H1-derived title) is load-bearing.
- Honest documentation of what's *not* covered: resonance/temporal persistence remains unbuilt (correctly dropped from `creek process` rather than silently kept as an overclaim), the stale-page reaper is filed as a separate follow-up (#1381), and a partial-write frontmatter-corruption risk is filed as #1382 rather than silently left.
- Fixes a real, unrelated pre-existing flake in the idempotency e2e test (second-resolution-timestamped review-queue filename) with a proof (deterministic clock-patch reproduction) rather than a guess.
- Two wording bugs in `_format_link_summary` (embeddings arm claiming edges were cached; temporal arm implying persistence) are corrected for `creek link` too, not just `creek process`.
- Verified independently: no remaining production imports of `creek.link.linker`/`LinkingPipeline`/`LinkingResult` anywhere outside historical ADR text (which is explicitly and correctly left alone as a historical record); `_PROCESS_LINK_METHODS` is deliberately distinct from `creek.surface_modes.LINK_METHODS` and that distinction is documented.

## Security Concerns
None identified. No new I/O surfaces, no new subprocess/network calls, no user input handling changes. 🟢 LOW (no findings)

## Problems
None found that block merge. Nothing here reads as a correctness regression; the design tradeoffs (dropping `embeddings`/`temporal` from `creek process`, whole-vault re-clustering cost, the known partial-write corruption path in #1382) are called out explicitly by the author with reasoning, rather than glossed over.

## Code Quality
- `_PROCESS_LINK_METHODS`'s trailing bare-string "docstring" (`pipeline.py`) is a Sphinx-recognized convention for module-level attribute docs but is inert at runtime — consistent with existing patterns elsewhere in this file, so not a new issue, just worth a mention if a reviewer isn't used to the convention.
- `EmbeddingPass.reused` is added but has no production consumer yet (per the author's own "not attempted" list) — harmless, but if it stays permanently unread it's a candidate for vulture/dead-code flagging down the line.

## CLAUDE.md Adherence
✅ Single conventional-commit message (`fix(pipeline): ...`). ✅ Scripts-based workflow implied throughout (check-all.sh, test.sh --all claimed run). ✅ No shortcut patterns observed (no `noqa`/`type: ignore` additions, no threshold changes). ✅ Root-cause fix (deletes the second orchestrator rather than papering over it) matches the "no shortcuts" and DRY principles. CI (as of review time): Pylint, Integration & E2E Tests, Code Complexity, Build Distribution, CrawDad Quality (3.11-3.13) all SUCCESS; Tests & Type Checking (3.11/3.12/3.13) and Static Analysis still IN_PROGRESS — not yet fully green, so Gate 3 isn't complete at review time, though nothing so far suggests it will fail.

## Testing
Exceptionally thorough for the size of the change: a new 525-line e2e module driving the actual CLI through `CliRunner`, unit tests for the new `EmbeddingPass` cache-hit/miss split, tests pinning the two-call ordering and its rationale, tests for the model-unavailable error path newly reachable from `creek process`, and a documented 19-mutant battery (17 killed, 2 verified — not assumed — equivalent, with one real survivor found and fixed during the author's own review, M17). Deleted tests are all accounted for: each deletion maps either to a class that no longer exists or to behavior now covered elsewhere, and one deletion (`test_run_returns_zero_temporal_thread_eddy`) is correctly identified as having pinned the original defect rather than a real behavior.

## Documentation
`docs/getting-started.md` and `docs/linking.md` are both updated to describe what `creek process`'s link stage actually persists (and explicitly does not), including the new "Stale pages on a growing corpus" caveat pointing at #1381. `docs/architecture/ADR/0008`'s stale reference to `creek.link.linker` is correctly left untouched as a historical record while the two live docstrings that referenced it were updated — a sensible, explicitly-reasoned distinction rather than an oversight.



## Verdict: LGTM

The fix is well-scoped, root-cause (deletes the dead second orchestrator rather than patching it), and backed by unusually rigorous on-disk-state testing plus a documented mutation-testing pass; independent verification confirms no dangling references to the deleted module and that CI is green on every check that has finished (Pylint, Integration & E2E Tests, Code Complexity, Build Distribution). The only asterisk is that Tests & Type Checking and Static Analysis were still in progress at review time, but nothing in the diff or the completed checks suggests they will fail, so this is ready to merge pending those finishing green.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5250653526 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1384, opened 2026-08-11
feat(vault): report pre-#1304 duplicate fragments left by the losing ingestor


Geoffe-Ga:
Split out of #1304, which fixed the cause but deliberately left the existing damage alone.

## Background

Before #1304, `creek process` ran every registered ingestor over every source file and kept
everything each returned, so a file claimed by two ingestors was written to the vault twice
or three times. #1304 arbitrates: one ingestor's output per source file.

That fixes new runs. It does nothing for vaults ingested before it. The **winning** ingestor
re-resolves to its own pre-existing id and de-duplicates cleanly; the **losing** ingestor's
fragments have different ids (the two disagree on both body and timestamp), so nothing
matches them, and now that the loser never runs against those files again nothing will ever
revisit them. They are ordinary fragment notes — indexed, linkable, indistinguishable from
real ones apart from being a second copy of a file the operator only has one of.

#1304's answer was **detect and report, delete nothing**: `PipelineResult.contested_sources`
names the affected source files and `creek process` prints them with a pointer to
`creek purge source --source-path <path> --match exact --dry-run`. That is deliberate —
deleting a fragment is not a call a processing run gets to make, and the operator may have
hand-edited or hand-linked either copy.

But it only names files the *current* run contested. It cannot see a stray whose source file
is no longer in the source tree, and the operator still has to run one purge dry-run per file
and eyeball the output.

## What is missing

A read-only, vault-wide duplicate report. Rough shape:

- Group every fragment under `01-Fragments/` by its `source.original_file` frontmatter key
  (the same key `creek purge source --source-path` already matches on).
- Report every group with more than one member, with each member's id, path, body length,
  and `source.platform`, so the operator can tell which is the pre-#1304 stray.
- Read-only by default. If a removal mode is offered at all it must reuse
  `creek.purge.engine` so reference scrubbing is not reimplemented, and must be explicit
  per group rather than a blanket sweep.

Worth deciding as part of this: whether it lands as `creek doctor` (a new umbrella command
for vault-integrity checks, which does not exist yet) or as a flag on an existing command.
`creek gdrive --check` is the closest precedent in the repo for a read-only doctor.

## Acceptance criteria

- [ ] A read-only command reports every source file with more than one fragment in the vault.
- [ ] Output is enough to act on: fragment id, vault path, `source.platform`, body length.
- [ ] Nothing is deleted or rewritten without an explicit, per-group opt-in.
[...]

PR #1387, opened 2026-08-11
fix(pipeline): one ingestor's output per source file, not every ingestor's

### LINKED ISSUE #1304: fix(pipeline): creek process runs every registered ingestor over every file, writing duplicate fragments
**Severity:** High
**Remediation:** decision-needed (route by source type, or make ingestor claims disjoint) + e2e test

## Problem

`Pipeline._run_ingestion` fans **every** registered ingestor across the **whole** source
tree with no type routing:

```python
for name, ingestor_cls in INGESTOR_REGISTRY.items():   # creek/pipeline.py:490
    ingestor = ingestor_cls()
    ingest_result = ingestor.ingest(source_path)
```

The ingestors' extension claims overlap. `GenericIngestor` excludes only `.md` and `.json`
(`creek/ingest/generic.py:39`), while `DocumentIngestor.SUPPORTED_EXTENSIONS` is
`{.docx, .pdf, .html, .htm, .txt, .rtf}` (`creek/ingest/documents.py:49-51`) and
`CodeIngestor` claims `.py`. Every file in that overlap is parsed **twice**, by two
ingestors that produce **different bodies** — and since `assemble_ingested_fragment` derives
the id from `source_path + timestamp + content` (`creek/ingest/base.py:441-445`), different
bodies mean different ids, so both land in the vault as separate fragments.

Taxonomy family 0 (correctness) + family 5 (architectural drift — two orchestrators, one
fans out, one routes).

**Current state:** every non-`.md` source file produces duplicate vault fragments on
`creek process`, with no error and no warning. Downstream this doubles classification and
LLM spend for those fragments.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** `Pipeline._run_ingestion` + `_write_to_vault` over a
temp source containing exactly `note.html` and `log.txt`:

```
4 fragments written, zero errors:
  01-Fragments/Unsorted/2026-08-09-log.md
  01-Fragments/Unsorted/2026-08-10-log.md
  01-Fragments/Unsorted/2026-08-09-note.md
  01-Fragments/Unsorted/2026-08-10-note.md

bodies differ per ingestor:
  note.html  Document -> "Hello world essay body."
             Generic  -> "```html\n<html>…"
  log.txt    Document -> "# plain text line one"
             Generic  -> "plain text line one"
```

Two source files in, four fragments out.

**Signal 2 — structural, and the correct contract exists next door.** The *other* ingest
path resolves exactly one ingestor: `run_ingest(ingestor_cls=…)`
(`creek/ingest/pipeline.py:328`), driven by `creek ingest --type`. So the repo already has a
single-ingestor routing contract; `creek process` simply does not use it.

**Signal 3 — why it survived.** The only e2e coverage
(`tests/e2e/test_full_pipeline_markdown.py`) uses a `.md`-only source dir — `.md` is the one
extension the `GenericIngestor` overlap excludes, so the duplication cannot show up there.

## Scope

Covers ingestor selection in `Pipeline._run_ingestion` (`creek/pipeline.py:482-510`) and the
extension-claim overlap between the registered ingestors.

Explicitly NOT in scope: the per-ingestor `authored_at`/mtime divergences (filed
separately), `code.py`'s LA_TZ mtime (#955), symlink-following discovery (#1294), or the
single-file discovery no-op in the ChatGPT/Claude ingestors.

## Tasks

1. Decide and record the direction:
   - **(a)** Route by source type in `_run_ingestion` — mirror `run_ingest`'s
     one-ingestor-per-source contract (preferred: it removes the drift); or
   - **(b)** Make the ingestors' extension claims disjoint so the fan-out is safe.
2. Implement it in `creek/pipeline.py:482-510`.
3. Add an e2e test over a **mixed-extension** source directory (at minimum one `.html`,
   one `.txt`, one `.md`) asserting one fragment per source file.

## Acceptance Criteria

- [ ] `creek process` over a source dir containing `note.html`, `log.txt` and `a.md`
      produces exactly 3 fragments.
- [ ] No fragment is written by two different ingestors for the same source file.
- [ ] An e2e test covers the mixed-extension case and fails before the fix.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/pipeline.py` | Modify |
| `creek-tools/tests/e2e/test_full_pipeline_markdown.py` | Modify (or add a mixed-source e2e) |

Geoffe-Ga:
## Summary

`Pipeline._run_ingestion` handed the whole source tree to all eleven entries in `INGESTOR_REGISTRY` and kept everything each one returned. The ingestors' claims overlap, so any contested file was written to the vault two or three times — never as an overwrite, because the vault writer de-collides a name clash with a `-N` suffix, so both copies land.

Measured by execution on this PR's base (`3796c39`), not inferred:

| source | before | after |
|---|---|---|
| `{note.html, log.txt, a.md}` | **5** fragment files, 0 errors | **3** |
| 7-file mixed tree (`README.md`, ADR `.md`, substack `.html`, plain `.html`, `.csv`, `.py`, `.json`) | **14** fragments | **7** |

Per-file, before: `README.md` 2 (code + markdown), ADR `0001-thing.md` 2, `163000000.my-post.html` **3** (substack + document + generic), `plain.html` 2, `sheet.csv` 2, `mod.py` 3.

Note the issue body is wrong on two points, both corrected here. It says `.md` is the one extension that cannot duplicate — it duplicates via `CodeIngestor`, which claims READMEs, `CLAUDE.md` and ADRs (`creek/ingest/code.py` `_is_relevant_file`). And it names `run_ingest(ingestor_cls=…)` as the routing contract that already exists; that is not a router, the caller supplies the class.

### The fix

Every ingestor still runs. `discord`, `chatgpt`, `claude` and `substack` recognise their input by its structure or content — a `messages/<channel>/messages.json` tree, a particular JSON envelope, a `<postid>.<slug>.html` filename — so no dispatch table can express their claims. An extension-keyed router would silently zero all four: their files are `.json`, `GenericIngestor` excludes `.json`, and nothing would pick up the slack. `tests/e2e/test_full_pipeline_mixed_sources.py::test_chat_exports_still_produce_fragments` is the guard for exactly that, green before and after.

What is new is `creek/ingest/routing.py`: a single total order `CLAIM_PRIORITY` over registry names, and `arbitrate()`, which groups the parsed fragments by `source_path` and keeps only the highest-priority ingestor that **actually produced output** for that path.

Arbitration runs **after** parsing rather than after `discover()`, deliberately. A claim staked at discovery has to be trusted before anyone has read the file, so when the winner's `parse()` then fails — a malformed `.csv`, a password-protected `.xlsx` — the loser is already discarded and the content is lost, where today it still lands as a generic fragment. Arbitrating on produced fragments makes that impossible by construction, and keeps today's memory profile (one ingestor's documents are freed before the next runs, rather than all eleven coexisting).

The invariant is **one ingestor per source file, never one fragment per source file**. `SpreadsheetIngestor` keeps emitting one fragment per non-empty sheet (this is #1305's model, and a winner keeps *all* its fragments for a path), `CodeIngestor` one per module and function, chat exports one per turn pair.

`creek ingest --type <x>` is unchanged by construction: `creek/ingest/base.py` is untouched, so every caller-chosen-class path still runs the ingestor the caller asked for over files the arbiter would have assigned elsewhere.

### Migration: the losing ingestor's already-written fragments

This is the mirror of what #1329 handled by pinning ids, and pinning does not help here. The *winner* re-resolves to its own pre-existing id and de-duplicates cleanly. The **loser's** fragments in an existing vault are real, indexed, linkable notes with different ids (the two ingestors disagree on both body and timestamp), and now that the loser will never run against those files again, nothing will ever revisit them.

**Decision: detect and report; delete nothing.** Deleting a fragment is not a call a processing run gets to make, and the operator may have hand-edited or hand-linked either copy. So the pipeline names the affected files instead of acting on them.

What an operator sees on the next `creek process`:

```
Contested sources: 3 (more than one ingestor claimed these; one won)
  A vault ingested before this release may still hold the losing ingestor's
  fragments. Nothing is deleted automatically; inspect with
  creek purge source --source-path <path> --match exact --dry-run
  /home/me/notes/log.txt
  ...
```

The recommended per-file sequence, documented in `docs/ingestion.md`, is to look first: `creek purge source --source-path <path> --match exact --dry-run` lists every vault fragment whose `source.original_file` is that path. More than one means a stray. Then either leave it (it is a duplicate note, not corruption) or purge-and-reprocess — with the caveat, stated in the doc, that purging scrubs references to `[purged]` across the vault and re-processing does not restore them.

`creek doctor`-style automated duplicate detection is a follow-up, filed as #1384.

### The `authored_at` consequence, named

The issue body defers per-ingestor `authored_at` divergence as separate. It is not fully separable — it is *why* the duplicates landed under different date-prefixed filenames, and the routing decision silently picks which timestamp survives. Concretely:

- Fragment ids hash `source_path + timestamp + content`, so the winner determines the body, the id **and** the `YYYY-MM-DD-` filename prefix.
- `CodeIngestor` stamps local time while the others stamp UTC. Between 16:00 and midnight Pacific those are different calendar days, which is why the original report shows `2026-08-10-log.md` and `2026-08-11-log.md` side by side.
- `GenericIngestor` writes an `authored_at` frontmatter key from the file mtime; `DocumentIngestor` leaves it unset. So `.txt`/`.html` fragments lose that key as a consequence of `document` winning.

All three are recorded in the `creek/ingest/routing.py` module docstring and the upgrade section of `docs/ingestion.md`.

### The priority order, and the two contests that are judgement calls

`discord · chatgpt · claude · substack` (sniffed) → `markdown · code · spreadsheet · presentation · image · document` → `generic` (the declared fallback, always last). One total order rather than a per-extension table, so it is reviewable at a glance and cannot disagree with itself. Every contested extension is justified in the module docstring; two deserve calling out here because I went against the pre-agreed plan on one of them:

- **`.md` → `markdown` over `code`** (the plan said `code`). `CodeIngestor` emits the file verbatim, so a README's YAML frontmatter stays inline in the fragment body — directly beneath the frontmatter block the vault writer adds. Verified: `code` yields `content='---\ntitle: My Readme\n…'` and `title='README.md'`; `markdown` splits the frontmatter, promotes `title: My Readme` and `tags`. Losing an `artifact_type` label costs less than a note with two frontmatter blocks in it, and "every `.md` goes to the markdown ingestor" is a rule an operator can hold in their head.
- **`.txt` → `document` over `generic`.** This is the weaker of the two: `DocumentIngestor` promotes the first line to an `#` heading (`# plain text line one`) where `generic` reproduces the text verbatim, and verbatim is arguably more faithful for a log file. But a total order cannot prefer `generic` on `.txt` and `document` on `.html`, and letting the declared fallback outrank a specialist anywhere makes the order incoherent. Named as a trade in the docstring rather than glossed.

### Riders

- **`GenericIngestor` no longer reads binaries it will discard.** `discover()` used to `read_bytes()` the whole file and `parse()` then dropped it; a tree of photos or Office documents was read cover to cover on every run. It now reads a bounded 8 KiB prefix and decides on that. **Output-neutral by proof, not by an extension list**: `_is_binary_content` returns True on a null byte *anywhere* or a >10% control-char ratio in the first 8192 bytes, so prefix-binary implies whole-file-binary implies `_try_decode` would have returned `None`. The UTF-16 BOM carve-out is replicated (UTF-16 text is null-heavy and is not binary) and locked by a test. This is why `.pdf` needed no special-casing despite being able to be pure ASCII: the predicate decides, not the extension.
- **No new silent drops.** `PipelineResult.unclaimed_sources` names files no ingestor produced anything from, with an explicit exclusion contract (directories; anything under a `SKIPPED_DIRECTORY_NAMES` or dot-directory; parent directories only, so a dotfile in the source root is still reported). This surfaces the *pre-existing* drop of a non-chat-export `.json`; fixing that drop is out of scope.
- **`_SKIP_DIRS` has one definition now.** `CodeIngestor` imports `routing.SKIPPED_DIRECTORY_NAMES`, so the discovery walk and the unclaimed report cannot drift into disagreeing about what was meant to be ingested.
- **Errors are never suppressed by arbitration.** Every ingestor's errors reach `result.errors`, losers included, and an ingestor that produced *zero* fragments still reports (that is the tesseract-missing case for `ImageIngestor`). This was a real hole the mutation battery found; it now has two tests.

## Test plan

**RED, proved on persisted vault state** — the defect is duplicate files on disk, so the assertions count files, not routing decisions. Fragments are attributed to their source via the `source.original_file` frontmatter key.

```
tests/e2e/test_full_pipeline_mixed_sources.py::test_process_writes_one_fragment_per_source_file
  E  assert 5 == 3
tests/e2e/test_full_pipeline_mixed_sources.py::test_mixed_extension_tree_routes_one_ingestor_per_file
  E  {'README.md': 2} != {'README.md': 1}
  E  {'0001-thing.md': 2} != {'0001-thing.md': 1}
  E  {'mod.py': 3} != {'mod.py': 2}
  E  {'sheet.csv': 2} != {'sheet.csv': 1}
  E  {'163000000.my-post.html': 3} != {'163000000.my-post.html': 1}
  E  {'plain.html': 2} != {'plain.html': 1}
```

`test_chat_exports_still_produce_fragments` passed at base and still passes — it is the guard that kills any extension-table design.

**New/changed tests**

- `tests/e2e/test_full_pipeline_mixed_sources.py` (new, 6 tests). A new module rather than an edit to the shared `test_full_pipeline_markdown.py`, which #1303 also wanted to extend; #1303 chose a new module (`test_full_pipeline_linking.py`) for the same reason.
- `tests/test_ingest_routing.py` (new, 19 tests). Includes `set(CLAIM_PRIORITY) == set(INGESTOR_REGISTRY)` as a forcing function: adding a twelfth ingestor without ranking it is a CI red, not a silent demotion behind `generic`.
- `tests/test_pipeline_source_coverage.py` (new, 11 tests). Exclusion contract, error preservation, and the rendered `creek process` output.
- `tests/test_ingest_generic.py` (+8). Binary skip at both discovery arms, UTF-16 survival, a >8 KiB text file round-tripping in full, an empty file still discovered, the late-binary edge, and a boundedness assertion (a 4 MB binary is judged on exactly 8192 bytes).
- `tests/e2e/test_full_pipeline_markdown.py`: the `assert result.fragments_created >= 1` at line 51 — a bound that cannot detect duplication *in principle*, and the reason this defect survived the whole life of the file — is now an exact count plus the exact on-disk filename, with the source mtime pinned so the date prefix is deterministic.
- **Zero edits** to the twelve `patch("creek.pipeline.INGESTOR_REGISTRY", …)` sites in `tests/test_pipeline.py` / `tests/test_pipeline_no_llm.py`. They pass untouched, which is a design constraint the approach was chosen to satisfy: the pipeline still calls `ingest()`, and their single synthetic `"mock"` key is unranked, uncontested, and wins silently.

**Mutation battery — 26 mutants, 24 killed, 2 survivors, both proven equivalent.**

Killed: reverting arbitration entirely; swapping `markdown`/`code`; ranking `generic` first; inverting the rank comparison; never recording a contest; ranking unranked names first; dropping the unranked warning; keeping every fragment; not sorting `contested`; dropping either half of the exclusion contract; excluding on the filename as well as the parents; suppressing either report field; dropping the UTF-16 carve-out; truncating reads at the prefix; never skipping binaries; always refusing to read; widening the prefix read; testing the whole file at discover; assembling from raw claims; never contesting; silencing the CLI report; suppressing a zero-output ingestor's errors.

Two of those mutants exist because the battery initially missed them. `suppressing a zero-output ingestor's errors` survived the first run — a genuine hole, since that is exactly the shape of `ImageIngestor` when `tesseract` is missing; it now has two tests. `testing the whole file at discover` was added on review feedback and covers the edge the output-neutrality argument actually turns on: a file whose first 8 KiB are clean text but which turns binary later.

Survivors, and why each is genuinely equivalent rather than assumed to be:

1. **`<` → `<=` in `arbitrate`.** The branch is guarded by `incumbent != name`, and `_rank` is injective (`CLAIM_PRIORITY` has no duplicates — asserted; unranked names get `len(CLAIM_PRIORITY) + offset` from distinct offsets over a sorted distinct name list). So `ranks[a] == ranks[b]` is unreachable for `a != b`. Confirmed by 20,000 randomised `_rank` calls over mixed known/unknown name sets: zero collisions.
2. **`claims` → `arbitration.winners` when building the `seen` set for `unclaimed_sources`.** Every source path in `claims` has exactly one owner, and that owner's fragments for that path are all kept and non-empty, so the two path sets are identical. Confirmed by 20,000 randomised arbitrations: zero differences. `claims` is kept anyway because "a loser still proves the file was seen" is the property that should hold if winners are ever filtered.

**Gates.** `./scripts/check-all.sh` → **exit 0**, 12/12 checks, 9090 passed, 94.44% aggregate coverage, per-file gate green (`creek/ingest/routing.py` 94.92%, `creek/pipeline.py` 99.31%, `creek/ingest/generic.py` 97.20%). `./scripts/test.sh --e2e` → 23 passed. No new `# noqa`, `type: ignore`, or skip anywhere, tests included.

**Pre-push self-review (Gate 2.5)** returned one blocker and three nits; all four are addressed in this branch:

1. *(blocker)* Two unjustified `# type: ignore[arg-type]` in a new test that monkeypatched `Path.open` to count bytes read. `tests/` is outside mypy's scope (#1366), so those would have been permanent debt no gate ever exercises. Fixed properly rather than tagged: the test now spies on `_is_binary_content` and asserts the predicate was handed exactly 8192 bytes, which is a more direct statement of the property anyway and needs no ignores.
2. `.rtf` was missing from the "each contested extension" enumeration in the `routing.py` docstring, which claims to be exhaustive. Added, along with the rest of `DocumentIngestor`'s set.
3. No test covered a file whose first 8 KiB are clean but which turns binary later — the edge the output-neutrality argument depends on. Added, and added as a mutant.
4. The extra `rglob` walk for `unclaimed_sources` and the `creek/pipeline.py` line count were raised as informational; the line count is in the "not satisfied" list below.

## Acceptance criteria not satisfied

Listed rather than glossed.

1. **"No source file is *parsed* by two ingestors in the same run."** This PR enforces the adjacent invariant: exactly one ingestor's output is **written** per source file. A contested *text* file is still parsed by every claimant and the losers' fragments are discarded before assembly — CPU only, within today's memory profile. The literal parsed-once invariant requires trusting a claim before parse, which is precisely what destroys content when the winner's parse then fails, and needs a `claims()` protocol on `Ingestor`; that is a larger PR and should not ride on this one. The binary sub-case the AC calls out explicitly — `GenericIngestor` reading `.docx`/`.pdf`/`.xlsx`/`.pptx`/images it will discard — **is** closed, at discover.
2. **Reusing `route_to_ingestor` / `_INGESTOR_BY_EXTENSION`** (`creek/ingest/gdrive.py`) is deliberately declined. It is extension-only and cannot express the sniffed claims of discord/chatgpt/claude/substack, and its missing `.py` key (would downgrade code ingestion to Generic) and missing `.json` key (would route to Generic, which excludes `.json`, yielding zero fragments) are what duplicating claim logic produces. It solves a different problem — naming an ingestor for one already-downloaded file with no directory discovery (`creek_mcp/tools/upload.py`). Leaving it alone keeps `upload.py`, `staged_names.py`, `connectors.py`, `journal_staging.py` and their tests entirely out of the blast radius. Unification is a follow-up.
3. **`creek/pipeline.py` grew 951 → 1123 lines**, crossing pylint's C0302 1000-line threshold. 19 such warnings already exist repo-wide and the pylint gate is score-based and passes, so I did not split the module in this PR; flagging it rather than letting it pass unremarked.

Closes #1304

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -11,7 +11,7 @@ import sys
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
-from typing import TYPE_CHECKING, Literal, NoReturn, cast
+from typing import TYPE_CHECKING, Final, Literal, NoReturn, cast
 
 import typer
 from rich.console import Console
@@ -159,6 +159,14 @@ app.add_typer(purge_app, name="purge")
 app.add_typer(skills_app, name="skills")
 app.add_typer(compost_app, name="compost")
 console = Console()
+
+_SOURCE_COVERAGE_SAMPLE: Final[int] = 10
+"""How many contested/unclaimed source paths ``creek process`` lists.
+
+The full lists live on :class:`~creek.pipeline.PipelineResult`; the
+summary only samples them so a large source tree cannot bury the rest of
+the report.
+"""
 
 
 def _consent_log_dir(vault_path: Path) -> Path:
@@ -1313,11 +1321,58 @@ def process(
         f"[bold]Residue:[/bold] {result.residue} "
         "(would go to LLM if Pass-3 enabled)"
     )
+    _print_source_coverage(result)
     error_count = len(result.errors)
     error_style = "red" if error_count else "dim"
     console.print(f"[bold {error_style}]Errors:[/bold {error_style}] {error_count}")
     for err in result.errors:
         console.print(f"  [dim]{err}[/dim]")
+
+
+def _print_source_coverage(result: PipelineResult) -> None:
+    """Report source files that were contested or that nobody ingested.
+
+    Silent when there is nothing to say, so a clean run's summary keeps
+    its current shape.
+
+    *Contested* is the migration notice for issue #1304. Before that
+    change several ingestors could each write a fragment for the same
+    file; now one wins. The loser's fragments in an existing vault are
+    left exactly where they are — deleting a fragment is not something a
+    processing run gets to decide — so this names the affected files and
+    the read-only command that finds the strays.
+
+    *Unclaimed* names files that produced nothing at all, which was
+    previously silent.
+
+    Args:
+        result: The finished pipeline result.
+    """
+    if result.contested_sources:
+        console.print(
+            f"[bold yellow]Contested sources:[/bold yellow] "
+            f"{len(result.contested_sources)} "
+            "(more than one ingestor claimed these; one won)"
+        )
+        console.print(
+            "  [dim]A vault ingested before this release may still hold the "
+            "losing ingestor's fragments. Nothing is deleted automatically; "
+            "inspect with[/dim]"
+        )
+        console.print(
+            "  [dim]creek purge source --source-path <path> --match exact "
+            "--dry-run[/dim]"
+        )
+        for path in result.contested_sources[:_SOURCE_COVERAGE_SAMPLE]:
+            console.print(f"  [dim]{path}[/dim]")
+    if result.unclaimed_sources:
+        console.print(
+            f"[bold yellow]Unclaimed sources:[/bold yellow] "
+            f"{len(result.unclaimed_sources)} "
+            "(no ingestor produced a fragment for these)"
+        )
+        for path in result.unclaimed_sources[:_SOURCE_COVERAGE_SAMPLE]:
+            console.print(f"  [dim]{path}[/dim]")
 
 
 @app.command()
--- a/creek-tools/creek/ingest/code.py
+++ b/creek-tools/creek/ingest/code.py
@@ -30,29 +30,20 @@ from creek.ingest.base import (
     normalize_encoding,
     parse_authored_at,
 )
+from creek.ingest.routing import SKIPPED_DIRECTORY_NAMES
 from creek.models import SourcePlatform
 
 logger = logging.getLogger(__name__)
 
 # ---- Constants ----
 
-_SKIP_DIRS: frozenset[str] = frozenset(
-    {
-        "node_modules",
-        "vendor",
-        ".git",
-        "__pycache__",
-        ".tox",
-        ".venv",
-        ".mypy_cache",
-        ".ruff_cache",
-        ".pytest_cache",
-        "dist",
-        "build",
-        ".eggs",
-    }
-)
-"""Directory names to skip during discovery."""
+_SKIP_DIRS: frozenset[str] = SKIPPED_DIRECTORY_NAMES
+"""Directory names to skip during discovery.
+
+Shared with the pipeline's unclaimed-source report (issue #1304) via
+:data:`creek.ingest.routing.SKIPPED_DIRECTORY_NAMES` — a file this walk
+declines to visit must not then be reported as an ingestor's oversight.
+"""
 
 _README_PATTERNS: tuple[str, ...] = (
     "README.md",
--- a/creek-tools/creek/ingest/documents.py
+++ b/creek-tools/creek/ingest/documents.py
@@ -581,10 +581,15 @@ class DocumentIngestor(Ingestor):
 
         Substack export directories (``posts.csv`` + per-post
         ``<id>.<slug>.html``) are claimed by ``SubstackIngestor``;
-        ``DocumentIngestor`` defers to it rather than double-emitting
-        every post as an opaque HTML document, so the auto-detect path
-        of ``creek process`` (which fans every registered ingestor at
-        the source) produces one fragment per essay, not two.
+        ``DocumentIngestor`` defers to it rather than emitting every
+        post a second time as an opaque HTML document. Since issue #1304
+        the pipeline would arbitrate that contest anyway — ``substack``
+        outranks ``document`` in
+        :data:`creek.ingest.routing.CLAIM_PRIORITY` — but narrowing the
+        claim here is still worth doing: it saves parsing a whole export
+        directory whose output is guaranteed to be discarded, and it
+        keeps ``creek ingest --type document`` honest, which the
+        arbiter never sees.
 
         Args:
             source_path: A file or directory path to search.
--- a/creek-tools/creek/ingest/generic.py
+++ b/creek-tools/creek/ingest/generic.py
@@ -47,6 +47,10 @@ _BINARY_CHECK_SIZE = 8192
 # If more than 10% of bytes are non-text control chars, treat as binary
 _BINARY_CONTROL_THRESHOLD = 0.10
 
+# UTF-16 byte-order marks. UTF-16 text is full of null bytes, so it trips
+# every binary heuristic here and has to be recognised before they run.
+_UTF16_BOMS: tuple[bytes, bytes] = (b"\xff\xfe", b"\xfe\xff")
+
 
 def _safe_file_mtime(path: Path) -> datetime | None:
     """Return *path*'s mtime as a tz-aware datetime, or ``None`` defensively.
@@ -113,7 +117,7 @@ def _try_decode(raw_bytes: bytes) -> str | None:
         return ""
 
     # Try UTF-16 first if BOM is present (contains null bytes so check before binary)
-    if raw_bytes[:2] in (b"\xff\xfe", b"\xfe\xff"):
+    if raw_bytes[:2] in _UTF16_BOMS:
         with suppress(UnicodeDecodeError, ValueError):
             return raw_bytes.decode("utf-16")
 
@@ -132,6 +136,39 @@ def _try_decode(raw_bytes: bytes) -> str | None:
 
     # Latin-1 as final fallback (can decode any byte sequence)
     return raw_bytes.decode("latin-1")
+
+
+def _read_unless_binary(file_path: Path) -> bytes | None:
+    """Read *file_path* whole, or return ``None`` once a prefix proves it binary.
+
+    ``parse`` has always discarded binary content (``_try_decode``
+    returns ``None`` for it), but ``discover`` used to slurp the entire
+    file into memory first — so a source tree of photos, videos or Office
+    documents was read cover to cover on every ``creek process`` run and
+    thrown away. This reads a bounded prefix, decides on that, and only
+    continues reading when the content might survive parsing.
+
+    Output-neutral by construction rather than by an extension list.
+    :func:`_is_binary_content` looks for a null byte anywhere and
+    otherwise scores only the first ``_BINARY_CHECK_SIZE`` bytes, so if
+    the prefix tests binary the whole file necessarily does too, and
+    ``parse`` would have dropped it. The UTF-16 BOM exemption is applied
+    here for the same reason ``_try_decode`` applies it: UTF-16 text is
+    riddled with nulls and is not binary.
+
+    Args:
+        file_path: File to read.
+
+    Returns:
+        The file's bytes, or ``None`` when it is binary and ``parse``
+        would have discarded it anyway.
+    """
+    with file_path.open("rb") as handle:
+        prefix = handle.read(_BINARY_CHECK_SIZE)
+        if prefix[:2] not in _UTF16_BOMS and _is_binary_content(prefix):
+            logger.debug("Skipping binary file without reading it: %s", file_path)
+            return None
+        return prefix + handle.read()
 
 
 class GenericIngestor(Ingestor):
@@ -170,12 +207,15 @@ class GenericIngestor(Ingestor):
             file_path: Path to the file.
 
         Returns:
-            A single-element list if the file is unclaimed, else empty.
+            A single-element list if the file is unclaimed and not
+            binary, else empty.
         """
         if file_path.suffix.lower() in _CLAIMED_EXTENSIONS:
             return []
 
-        raw_bytes = file_path.read_bytes()
+        raw_bytes = _read_unless_binary(file_path)
+        if raw_bytes is None:
+            return []
         _text, encoding = normalize_encoding(raw_bytes)
         return [
             RawDocument(
@@ -193,7 +233,8 @@ class GenericIngestor(Ingestor):
             dir_path: Directory path to search.
 
         Returns:
-            A list of ``RawDocument`` objects for unclaimed files.
+            A list of ``RawDocument`` objects for unclaimed, non-binary
+            files.
         """
         docs: list[RawDocument] = []
         for file_path in sorted(dir_path.rglob("*")):
@@ -202,7 +243,9 @@ class GenericIngestor(Ingestor):
             if file_path.suffix.lower() in _CLAIMED_EXTENSIONS:
                 continue
 
-            raw_bytes = file_path.read_bytes()
+            raw_bytes = _read_unless_binary(file_path)
+            if raw_bytes is None:
+                continue
             _text, encoding = normalize_encoding(raw_bytes)
             docs.append(
                 RawDocument(
--- a/creek-tools/creek/ingest/routing.py
+++ b/creek-tools/creek/ingest/routing.py
@@ -74,7 +74,10 @@ Each contested extension and the reason its winner wins:
   ``document`` on ``.html``, and letting the declared fallback outrank a
   specialist anywhere makes the order incoherent.
 * ``.csv``/``.xlsx`` -> **spreadsheet**, ``.pptx`` -> **presentation**,
-  images -> **image**, each over **generic**, for the same reason.
+  images -> **image**, ``.rtf``/``.docx``/``.pdf``/``.htm`` -> **document**,
+  each over **generic**, for the same reason. That completes
+  ``DocumentIngestor``'s extension set; ``generic`` claims everything
+  outside ``{.md, .json}``, so every one of them is contested.
 
 The winner is load-bearing beyond the body text: a fragment's id hashes
 ``source_path + timestamp + content`` (``creek.ingest.base``), and the
--- a/creek-tools/creek/ingest/routing.py
+++ b/creek-tools/creek/ingest/routing.py
@@ -1,0 +1,251 @@
+"""Decide which ingestor owns a source file when several claim it (#1304).
+
+``Pipeline._run_ingestion`` hands the whole source tree to every entry in
+:data:`creek.ingest.INGESTOR_REGISTRY`. That is deliberate — several
+ingestors identify their input by *sniffing* it, not by its extension —
+but it means a file claimed by two ingestors used to be written to the
+vault twice. Nothing de-duplicated it downstream: the vault writer
+de-collides a name clash with a ``-N`` suffix, so both copies landed.
+
+This module supplies the arbiter. Every ingestor still runs, and then
+:func:`arbitrate` groups the parsed fragments by
+:attr:`~creek.ingest.base.ParsedFragment.source_path` and keeps only the
+fragments of the highest-priority ingestor that actually produced output
+for that path.
+
+Why arbitrate after parsing rather than after ``discover()``
+------------------------------------------------------------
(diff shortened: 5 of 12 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real, well-diagnosed bug: `Pipeline._run_ingestion` ran every registered ingestor over the whole source tree and kept every fragment each one produced, so any file claimed by more than one ingestor (e.g. `.md` by both `code` and `markdown`, `.html` by `substack`/`document`/`generic`) was written to the vault two or three times. The fix adds `creek/ingest/routing.py`, a total-order arbiter (`CLAIM_PRIORITY` + `arbitrate()`) that runs *after* parsing and keeps only the highest-priority ingestor's output per source file, plus `contested_sources`/`unclaimed_sources` reporting on `PipelineResult` and in the `creek process` CLI output for migration visibility.

## Strengths
- Root-cause fix, not a patch: arbitrating on **produced fragments** rather than at `discover()` time is the right call — it means a claim that fails to parse (malformed `.csv`, encrypted `.xlsx`) forfeits to the next-best ingestor instead of silently dropping content. This is explained clearly in the `routing.py` module docstring and matches the code (`arbitrate` only compares fragments that were actually produced).
- I traced the `arbitrate()` running-min algorithm by hand (registry-order iteration, rank-based winner tracking) and confirmed it's order-independent and correctly finds the minimum-rank contender regardless of how `claims.items()` is iterated — verified against `test_fragment_order_within_an_ingestor_is_preserved` and the multi-contender parametrized test.
- `GenericIngestor`'s new bounded-prefix binary check (`_read_unless_binary`) is provably output-neutral: `_is_binary_content`'s null-byte check operates on whatever bytes it's handed, so a prefix that tests binary implies the full file would too; and the one case where the prefix *can't* know (clean-in-prefix, binary-later) is deliberately deferred to `parse()`, which still sees the whole file. I confirmed this by reading `_is_binary_content` directly (`creek/ingest/generic.py:79-100`) — the design and the `test_a_null_byte_past_the_prefix_is_left_to_parse` test agree.
- `SKIPPED_DIRECTORY_NAMES` is de-duplicated into `routing.py` and imported by `code.py`, closing a real drift risk between the discovery walk and the new unclaimed-source report.
- The PR body is unusually rigorous: it states measured before/after fragment counts, names the two mutation survivors and why each is genuinely equivalent, and lists three acceptance criteria explicitly *not* satisfied with reasoning rather than glossing over them. That's a high bar and it's met.
- Registry keys (`creek/ingest/__init__.py:55-67`) exactly match `CLAIM_PRIORITY`'s 11 entries — verified directly, not just via the `test_claim_priority_covers_the_whole_registry_exactly` forcing-function test.
- Sound "detect and report, delete nothing" call for pre-existing vault strays — deleting fragments during a processing run is the wrong layer for that decision, and the doc update explains the exact recovery path (`creek purge source --dry-run`).

## Security Concerns
None identified. 🟢 No new file I/O beyond the existing source tree, no new subprocess/network use, no injection surface. The bounded-read change in `generic.py` reduces attack surface (less binary data read into memory) rather than increasing it.

## Problems
None found that block merge. I manually traced the arbitration algorithm, the binary-prefix-skip logic, and the unclaimed/contested reporting flow (`_is_authored_content`, `_unclaimed_files`, `_record_source_coverage`) against the test suite and found the logic internally consistent and correctly reasoned in every case I checked, including edge cases (empty `claims`, unranked ingestor names, ties, single-file `source_path`, dotfile-vs-dot-directory).

**Caveat on verification**: this sandbox has no `uv`/`pytest`/`ruff`/`mypy` installed and installing them was out of scope for this review, so I could not execute `./scripts/check-all.sh` or the test suite myself to independently confirm the claimed "9090 passed, 94.44% coverage, exit 0" result. My review is based on careful manual static tracing of the diff, not execution. Given the precision of the PR description (exact before/after counts, exact mutant list, exact survivor justification) and that the logic held up under hand-tracing, I have reasonably high confidence in the claims, but this is worth flagging explicitly rather than silently treating the author's self-report as verified.

## Code Quality
- `_print_source_coverage` in `cli.py` (creek-tools/creek/cli.py:75-84) samples the first 10 contested/unclaimed paths but doesn't indicate truncation (e.g. "... and N more") when the list exceeds `_SOURCE_COVERAGE_SAMPLE`. The header count makes the truncation inferable, but an explicit "+N more" would be friendlier. Non-blocking.
- `_unclaimed_files` performs a full second `rglob("*")` walk of the source tree (on top of each ingestor's own discovery walk) purely for reporting. The author already flags this in the "acceptance criteria not satisfied" section as informational/accepted debt; agree it's not blocking for typical vault sizes, but worth watching if source trees get very large.
- `creek/pipeline.py` crossed pylint's C0302 1000-line soft threshold (951→1123); explicitly acknowledged by the author as an unaddressed but non-blocking issue since the pylint gate is score-based, not line-count-based.

## CLAUDE.md Adherence
✅ Conventional commit messages (`fix(pipeline): ...`, `test(ingest): ...`) with `Closes #1304`.
✅ No new `# noqa`, `# type: ignore`, or `pylint: disable` anywhere in the diff (verified by grep against the full diff).
✅ TDD structure evident: new test modules (`test_ingest_routing.py`, `test_pipeline_source_coverage.py`, `test_full_pipeline_mixed_sources.py`) each open with the RED assertion / bug being guarded against.
✅ Single-source-of-truth principle followed for `SKIPPED_DIRECTORY_NAMES` (§1.2 DRY).
✅ Root-cause fix rather than a shortcut — no lowered thresholds, no skipped tests, no suppressed checks.
⚠️ Coverage/lint/type-check/complexity thresholds are self-reported in the PR body, not independently verified in this review (see caveat above under Problems).

## Testing
Extensive and well-targeted: unit tests directly on the arbiter's contract (ties, unranked ingestors, ordering, empty input), e2e tests asserting on persisted vault state (not just routing decisions, which is the right level given the bug was "duplicate files on disk"), and a dedicated CLI-output test class. The RED assertions shown in the PR body (`assert 5 == 3`, exact per-file dict mismatches) demonstrate the tests actually caught the described bug before the fix. The claimed mutation-testing pass (26 mutants, 24 killed, 2 proven-equivalent survivors with explicit justification) is a strong signal, though — per the caveat above — I could not re-run it myself. No insufficient-testing concerns; this would not be a blocking gap even under strict review.

## Documentation
`docs/ingestion.md` gets a substantial, clearly-written new section explaining the arbitration model, the priority order, and — importantly — a concrete migration runbook for vaults ingested before this change (`creek purge source --dry-run` workflow, with the caveat about `[purged]` reference-scrubbing being irreversible). The `routing.py` module docstring is unusually thorough, justifying every contested-extension decision inline. Docstrings on new public surface (`Arbitration`, `arbitrate`, `_record_source_coverage`, `PipelineResult` field docs) are complete and accurate against the implementation I traced.



## Verdict: LGTM

The fix is well-targeted at the actual root cause, the arbitration algorithm and binary-prefix-skip logic both check out under manual tracing (including edge cases), test coverage is extensive and demonstrably catches the original bug, documentation and migration guidance are thorough, and CLAUDE.md conventions (commits, no suppressions, DRY) are followed. The only remarks are non-blocking nits (CLI sample truncation indicator, an acknowledged extra rglob walk, a pylint line-count soft-threshold the author already flagged) — none of which I'd hold up merge for.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5251233290 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1393, opened 2026-08-11
fix(ingest): an edited spreadsheet is never detected as changed (ledger hashes the empty string)


Geoffe-Ga:
## What

On the ledgered ingest path, `record_in_ledger` stores `ledger.content_hash(parsed.content)` and `write_fragment_idempotent` decides changed-vs-unchanged by comparing that hash.

`SpreadsheetIngestor.parse` sets `content=""` for every fragment — the rendered table is produced later, in `convert_to_markdown`. So **every** spreadsheet unit records the SHA-256 of the empty string, forever.

Consequence: an uploaded workbook whose cells the operator has edited re-ingests as `unchanged`. The new values are never written to the vault. The run reports success.

`PresentationIngestor.parse` has the identical `content=""` shape and is presumed to have the identical defect; it is not ledger-reachable today via `creek.upload` only because of extension routing, so confirm before fixing.

## Reproduce

Upload a `.xlsx` through `creek.upload`, change a cell, re-upload under the same `external_id`. Expected `action: updated` with the new table on disk; actual `action: unchanged` with the original table.

## Not caused by #1305, and specifically not fixable the obvious way

#1305 gave each sheet its own ledger key, so each sheet now has its own record — but every one of those records still holds the empty-string hash. #1305 neither caused nor cured this.

**Do not fix it by moving the sheet discriminator into `parsed.content`.** That was measured during #1305's design and it is data loss: a discriminator in `content` changes the content hash while leaving all N sheets on one ledger key, which makes `write_fragment_idempotent` take the changed branch, reassign `fragment.id = record.fragment_id`, and overwrite sheet 1 with sheet 2 and then sheet 3 — reported as `created=1 updated=2`, which reads like work was done.

## Shape of a real fix

Hash the *rendered* content the fragment will actually carry, not `parsed.content`. That means either running `convert_to_markdown` before the ledger record is written, or having such ingestors populate `content` with the rendered body and adjusting the id derivation accordingly — the latter moves fragment ids and therefore needs the same migration care #1305 and #1329 both needed.

## Acceptance

- [ ] Editing a cell in a ledgered workbook and re-uploading reports `updated` and writes the new table.
- [ ] Re-uploading an UNMODIFIED workbook still reports `unchanged` (no spurious rewrite).
- [ ] Whether `PresentationIngestor` shares the defect is established, and covered either way.
- [ ] Any change to id derivation carries the pin discipline #1329 established.

Discovered while implementing #1305.

Issue #1392, opened 2026-08-11
fix(models): Fragment silently discards ingestor frontmatter it does not model


Geoffe-Ga:
## What

`Fragment` (`creek/models.py`) sets `model_config = ConfigDict(use_enum_values=True)` and does **not** set `extra`, so pydantic's default `extra="ignore"` applies. `assemble_ingested_fragment` builds the frontmatter dict from whatever the ingestor emitted and then calls `Fragment.model_validate(frontmatter_dict)` — which **silently discards every key the model does not declare**.

`SpreadsheetIngestor.generate_frontmatter` emits `sheet`, `rows` and `columns`. None of the three ever reaches a vault file.

Verified by dumping a written fragment: `grep -cE '^sheet:|^rows:|^columns:'` returns 0.

`tests/test_spreadsheet_presentation_ingestors.py::test_frontmatter_records_platform_and_dimensions` asserts all three keys — on the **dict the ingestor returns**, which is true, and which never becomes a file. So the suite looks like it covers this and does not.

## Why it matters now

Before #1305 this was cosmetic. After #1305 the sheet name is load-bearing provenance: a multi-sheet workbook produces one fragment per sheet, and the *only* per-sheet markers that survive to disk are the `title`, the filename derived from it, the `# <file> — <sheet>` body heading, and (on the ledgered path) `source.origin_key`. The structured `sheet` field an automated consumer would want is dropped.

#1305 decided explicitly to **accept the drop** rather than widen the model inside a bugfix, and pinned the current behaviour in `test_sheet_rows_and_columns_do_not_survive_to_disk` so a future change is a deliberate one.

## Options

1. Declare the fields on a spreadsheet-aware model or a typed `extra_frontmatter` passthrough on the writer.
2. Set `extra="allow"` on `Fragment` — **rejected as stated**: it would let every ingestor's typos become vault frontmatter.
3. Keep the drop and document it in the ingestor contract.

Note the trap that rules out the obvious shortcut: `_write_model` serialises with `model.model_dump(mode="json")` and no `exclude_none`, so adding a nullable field to `FragmentSource` appends `unit: null` (or `sheet: null`) to the `source:` block of **every fragment written by every ingestor**, not just spreadsheets.

## Acceptance

- [ ] A decision is recorded (surface the fields, or document the drop in the `Ingestor` contract docstring).
- [ ] If surfaced: no new key appears in the frontmatter of fragments from ingestors that do not set it.
[...]

PR #1394, opened 2026-08-11
fix(ingest): give every sheet of a workbook its own identity

### LINKED ISSUE #1305: fix(ingest): every sheet of a multi-sheet workbook gets the same fragment id — all but the last are lost
**Severity:** High
**Remediation:** fix (give the id hash a per-sheet discriminator) + regression test

## Problem

Every sheet of a multi-sheet workbook collapses into a single vault fragment. Only the last
sheet survives; the rest are silently lost.

`SpreadsheetIngestor` emits one `ParsedFragment` per sheet, but sets `content=""`
(`creek/ingest/spreadsheets.py:450`, comment: `# Markdown rendered in convert_to_markdown.`)
with the same `timestamp` (`:437`) and the same `source_path` (`:458`).
`assemble_ingested_fragment` derives the fragment id from `source_path + timestamp +
content` (`creek/ingest/base.py:441-445`) — and `content` is the only per-sheet
discriminator that hash sees. `generate_frontmatter` also emits no `title`
(`spreadsheets.py:479-504`), so `base.py:447` falls back to the file stem for every sheet
and the vault filename collides too.

Taxonomy family 0 (silent data loss). The payload refactor (#166) moved sheet content onto
`ParsedFragment.payload` and left `content` empty; the id derivation was never updated.

**Current state:** an XLSX with N sheets yields N fragments carrying **one** id, which
`VaultWriter` collapses to one file containing only the last sheet.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** A 3-sheet stub workbook through
`SpreadsheetIngestor(backend=stub).ingest(path)` then `assemble_ingested_fragment`:

```
('frag-095913511f6c', 'Budget')
('frag-095913511f6c', 'Notes')
('frag-095913511f6c', 'Q3')
distinct ids: 1 of 3
```

Driving the same three fragments through `VaultWriter` writes **1 file**
(`01-Fragments/Data/2026-08-09-book.md`) containing only the last sheet.

**Signal 2 — structural, and the sibling is safe.** `presentations.py:310` sets
`content=""` the same way, but emits exactly **one** fragment per file, so only the
spreadsheet ingestor is exposed to the collision. That asymmetry is the tell: the empty
`content` is only a defect where multiple fragments share a `(source_path, timestamp)`.

Reachable from both `creek ingest --type spreadsheet` and `creek process`
(`creek/pipeline.py:498`).

## Scope

Covers per-sheet fragment identity and titling in `creek/ingest/spreadsheets.py`.

Explicitly NOT in scope: the openpyxl-absent / CSV-encoding coverage gaps (#870), the
`has_header` plumbing, or the shared `discover()` duplication with `presentations.py`.

## Tasks

1. Give the id a per-sheet discriminator — either include `sheet.name` in
   `ParsedFragment.content`, or extend `generate_fragment_id`
   (`creek/ingest/base.py:330-346`) with an optional `discriminator` argument and pass the
   sheet name from `spreadsheets.py:448-462`.
2. Emit a per-sheet `title` (e.g. `"<file stem> — <sheet name>"`) from
   `generate_frontmatter` (`spreadsheets.py:479-504`) so the vault filename stops colliding.
3. Add a regression test: a 3-sheet workbook produces 3 distinct ids, 3 distinct titles, and
   3 files on disk after a `VaultWriter` round-trip.

## Acceptance Criteria

- [ ] An N-sheet workbook yields N fragments with N distinct ids.
- [ ] After writing to a vault, N files exist and each contains its own sheet's content.
- [ ] Re-ingesting the same workbook is idempotent (same ids, no new files).

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/ingest/spreadsheets.py` | Modify |
| `creek-tools/creek/ingest/base.py` | Modify (only if the `discriminator` route is chosen) |
| `creek-tools/tests/test_spreadsheet_presentation_ingestors.py` | Modify |

Geoffe-Ga:
## Summary

Every sheet of a multi-sheet workbook derived the same fragment id, so all but one were silently dropped. This gives each sheet its own identity — and closes the three ratchets that opening that door creates.

**The issue body is wrong about the symptom, and correcting it matters.** It says twice that "only the last sheet survives". It is the **first**. `VaultWriter._write_model` takes the lock, calls `_find_existing_locked` and returns the existing path, so writes 2..N are silent no-ops. A test written to the body's wording is red before *and* after the fix and certifies nothing.

### The four-combination matrix, re-measured at this branch's base

```
                                    run  created updated unchanged  files  headings on disk
HEAD,  unledgered (creek ingest)    1      3       0        0         1    Budget
HEAD,  ledgered  (creek.upload)     1      1       0        2         1    Budget
                                    2      0       0        3         1    Budget
FIXED, unledgered (creek ingest)    1      3       0        0         3    Budget, Notes, Q3
FIXED, ledgered  (creek.upload)     1      3       0        0         3    Budget, Notes, Q3
                                    2      0       0        3         3    Budget, Notes, Q3
```

The `FIXED, ledgered` row is the one that needed work. A per-sheet **fragment id alone does not produce it** — `derive_source_key` keys on the file, so all N sheets share one `origin_key` and one `LedgerRecord`, and `write_fragment_idempotent` then reassigns `fragment.id = record.fragment_id` and overwrites sheet 1 with sheet 2 and then sheet 3. That intermediate state reports `created=1 updated=2`: two of three sheets still lost, wearing a success message. Closing it needs a per-sheet **source key** as well.

### What changed

1. **`creek/ingest/source_unit.py`** (new) — `SOURCE_UNIT_SEPARATOR`, `compose_source_unit`, `split_source_unit`. One spelling, shared by the ingest derivation, the ledger, and the purge engine. No heavy imports, so `creek.purge` can take it without pulling `creek.classify`.

2. **`ParsedFragment.source_unit` + `identity_key`** — the discriminator rides on a new optional field, and `identity_key` is the single string every identity derivation reads. `source_path` is left as the whole file. That is a deliberate departure from the planned design, and the reason is below.

3. **`SpreadsheetIngestor.parse`** sets a per-sheet unit **only when the workbook yields two or more non-empty sheets**, plus an explicit per-sheet `title` in `generate_frontmatter`.

4. **The ledgered path** — `attach_origin_key` composes the unit into `source.origin_key`; `SourceLedger.records_for` enumerates a base key's units; `creek_mcp/tools/upload.py`'s `_resolve_fragment_id` becomes `_resolve_fragment_ids` and all four consumers are updated.

5. **The purge sweep** — `PurgeEngine._resolve_staged_source` tries the whole key, then its path half, re-applying both containment guards.

### The design departs from the plan, because the plan predates `routing.py`

The judged plan put the discriminator directly in `ParsedFragment.source_path`. That was correct against the tree it was written for. It is not correct against this one.

`creek/ingest/routing.py` was added by #1387 (issue #1304) and **did not exist** at the commit the plan was written against — `git cat-file -e fdfc39f:creek-tools/creek/ingest/routing.py` returns `ABSENT`. `arbitrate` decides which ingestor owns a source file by grouping fragments on `fragment.source_path`. Overloading that string would give a 3-sheet workbook three distinct "source files", so a contested extension would no longer be arbitrated at all — both the specialist's fragments *and* the generic fallback's would survive, which is precisely the double-writing #1304 had just fixed.

Two further hazards the overload carries, both measured rather than reasoned: `Path("book.xlsx#Budget").suffix` is `'.xlsx#Budget'`, so any suffix dispatch on the string mis-routes; and five production consumers of `.source_path` (`creek/fragment.py:213/252/290/302/362`) were outside the plan's stated audit.

Putting the unit on its own field gets everything the overload got — per-sheet fragment id, per-sheet provenance, per-sheet ledger key — with none of that. `arbitrate` needed no change at all. This is **not** the rejected `FragmentSource.unit` approach: `ParsedFragment` is an in-memory parse-stage model that is never serialised, so no `unit: null` appears in any vault file.

### Migration: the #1304 answer, reused

Two populations were at risk.

**Every existing CSV and single-sheet XLSX** — pinned **by construction**. A workbook with exactly one non-empty sheet emits no unit, so it hashes exactly what it hashed before. A ledger backfill was unavailable (`ledger_for_source` returns `None` for every source type but markdown, and #1368's `pin_ids.py` is markdown-only by construction), so the pin had to be in the derivation. Proven, not asserted: `git archive HEAD` into a scratch tree, verified to import pre-fix code, then the same scenarios run there —

| source | pre-fix id | post-fix id |
|---|---|---|
| one-sheet `book.xlsx` | `frag-2ae23100e7a8` | `frag-2ae23100e7a8` |
| `book.csv` | `frag-e2f9d18e46af` | `frag-e2f9d18e46af` |
| `deck.pptx` | `frag-439246060197` | `frag-439246060197` |

The **filename** is pinned too, which is why `generate_frontmatter` emits the bare stem when there is no unit rather than naming the lone sheet: an id pin that lets the filename churn is only half a pin.

**The one collapsed fragment per multi-sheet workbook** — **detected, named, and left alone**, which is the answer #1304 gave the same shape of problem (`contested_sources`: name the strays, point at the read-only `creek purge source --dry-run` recipe, delete nothing). `collapsed_unit_warning` recomputes the superseded id — exactly what the fragment in hand hashes to with no unit — and looks it up in the writer's id index. It rides the existing `IngestRunResult.warnings` + `on_warning` channel #1329 built, so `creek ingest` prints it with no new CLI surface.

That detector is strictly better than the frontmatter-recomputation one the plan proposed: no vault walk, no `platform` scoping heuristic, no sibling check, and it sidesteps both frontmatter traps (`ingested` reads back as a `str`; its trailing `Z` hashes differently from `+00:00`) by never reading frontmatter. It is self-clearing and silent on unit-less sources.

I did **not** ship a tombing migration. Deleting operator vault content on the strength of a heuristic is the one-way operation this repo has now twice declined to automate, and I saw no reason #1305 differs.

### Two ratchets this fix opens, both closed here

Neither is optional; `SpreadsheetIngestor.parse` alone would be two privacy regressions.

- **RTBF purge.** A sub-unit `origin_key` resolves inside the staging root but names no file, so `is_file()` returned early and the staged workbook **survived an erasure request**. `_resolve_staged_source` now falls back to the path half — re-applying *both* guards, and only ever widening deletion inside the staging roots. Whole-key-first means a genuinely `#`-named file (`report#1.xlsx`) still resolves to itself.
- **The #970 tier gate.** `_resolve_fragment_id` looked up the bare staged key and returned `None`, which every caller reads as "this id owns nothing": the gate would compute `existing=None` and **admit a caller at `ceiling=open` to overwrite intimate content**; `_discard_unreferenced` would **unlink a workbook three live fragments still reference**; `_refuse_extension_conflict` would fork one `external_id` onto two fragments. All four now read the id list.

### The two surfaces that bypass `run_ingest`

Both were named in the issue's blast radius and neither is reachable by a `run_ingest` test, so both got their own:

- The **`creek.ingest` MCP tool** audit-logged the same id three times in `affected_fragment_ids` — a false audit record, and audit records are what an RTBF request is answered from. Now three distinct ids. Fixed (upstream in `parse`) **and asserted**.
- **`creek process`** classified all N fragments and then discarded N−1 when the writer deduped, paying N−1 LLM calls for nothing. That waste is gone as a side effect; flagging it for #1303/#1304 so nobody re-plans around a number that has already moved.

## Test plan

**RED, correctly stated.** `test_multi_sheet_workbook_writes_one_file_per_sheet` asserts persisted vault state — three files, and each of `# book.xlsx — Budget` / `— Notes` / `— Q3` in exactly one file — not a returned id. Its docstring states the *first*-sheet symptom. No `frag-<12hex>` literal appears in any multi-sheet test: ids hash the absolute `tmp_path`, so the issue body's `frag-095913511f6c` is unreproducible (three runs gave three different values). Literals appear only in the pin tests, under `monkeypatch.chdir` with a relative path and a forced mtime.

**Each hunk proven load-bearing** by reverting only it:

| reverted | result |
|---|---|
| `parse` emits no unit | 19 failures across all four surfaces |
| `attach_origin_key` drops the unit | 9 failures — ledgered path only, the piece a per-sheet id does not give you |
| purge sub-unit fallback | 2 failures — the workbook survives RTBF |
| upload plural resolution | 4 failures, incl. `fragment_id is None` (gate ratchet) and a destroyed staged workbook |
| advisory | 1 failure |
| *(restored)* | 120 passed |

**Mutation battery: 26 mutants at first pass** (extended to 33 after review — see below). The first run reported 26/26 with a scope list containing a filename that does not exist — pytest exited 4 on every run, so every "kill" was false. Fixed the scope, added a fail-fast baseline-green guard, and re-ran: **4 real survivors**, all genuine coverage gaps rather than equivalent mutants, each closed with a new test:

- `M9` provenance id reverted to `source_path` — the second, independent `generate_fragment_id` call site was unasserted. → `test_each_sheet_gets_its_own_provenance_entry`.
- `M15` `records_for` unordered — `fragment_id` reduces N ids to one, and insertion order is not a promise the ledger makes. → `test_multi_sheet_fragment_id_is_stable_under_sheet_order`, with anti-alphabetical sheet names so the two orders disagree.
- `M23` advisory dedup removed — would overstate stray content by a factor of N. → `test_collapsed_advisory_names_each_superseded_id_once`.
- `M24` advisory fires for unit-less fragments — would accuse the operator on **every** re-ingest of every CSV. → `test_single_sheet_re_ingest_raises_no_collapsed_advisory`.

**Gate 2:** `./scripts/check-all.sh` exit code **0**, 12/12 checks. Aggregate branch coverage 94.43%; `creek/ingest/spreadsheets.py` 89.77% (up from 89.50%, floor 80%); `source_unit.py` and `ingest/base.py` 100%. Integration + e2e: 380 passed, including #1304's `test_mixed_extension_tree_routes_one_ingestor_per_file` — its `sheet.csv` is single-sheet, so the carve-out keeps it at one fragment and the one-ingestor-per-file invariant intact.

### Gate 2.5 found two real security defects, both fixed

Self-review returned `BLOCKERS`. Both were genuine, both were introduced by this change, and both are now fixed at the root with a regression test proven to catch them.

**1. A sheet name containing `#` minted an unreadable key.** Excel permits `#` in a sheet title (openpyxl round-trips one), so `Rev#2` composed `…/book.xlsx#Rev#2`. `split_source_unit` splits at the *last* separator, giving base `…/book.xlsx#Rev` — a file that does not exist — so the RTBF purge deleted the fragment and **silently left the staged workbook on disk**, `journal_staged_removed == 0`, no error. A regression versus pre-#1305, where no unit existed and the whole-key lookup matched.

Fixed at mint time, not in the reader: `sanitize_unit` rewrites the separator, so "the unit half of a composed key never contains a separator" is an invariant of the key space rather than something each reader must rediscover. Applied **before** de-duplication — sanitising after would map `Rev#2` and `Rev-2` onto one unit and one id, reintroducing this issue's own defect (pinned by `test_separator_bearing_sheet_names_do_not_collide_after_sanitising`).

**2. Purge one sheet, then re-upload at `ceiling=open`, and the tier gate opened.** Purging one sheet of an N-sheet intimate upload deletes the shared staged workbook (that is the RTBF contract) while the other N−1 fragments stay live, intimate, and ledgered. The #970 gate surveyed the staging *directory*, saw it empty, concluded the `external_id` owned nothing, and admitted the caller as a **creation**. The re-upload restaged to the same deterministic path, `write_fragment_idempotent` matched the survivors by their intact ledger keys, and overwrote intimate content with the low-ceiling caller's bytes — **a privacy-tier bypass reachable through two individually legitimate operations.**

Fixed by surveying the **ledger** (`_ledgered_names_for_id`), which outlives the file it points at, for both the tier gate and the extension-conflict check. Uses `all_keys()` rather than `live_keys()`: a soft-tombed fragment still exists under `10-Liminal/Orphaned/`, and for a gate whose failure mode is *admitting*, over-counting is the safe error.

Reverting either fix turns `refused` back into `ok` on the new tests — verified.

Also from the review, all applied: five stale `path:line` citations in the new tests replaced with symbol references (`writer.py:1258-1261` had drifted into docstring `Args:` text; `pipeline.py:336` likewise — ironic in a change whose own docstring true-up was about exactly this); an unasserted in-vault decoy in the purge traversal test now asserted; and `no_llm=True` on the `creek process` e2e so it cannot reach a live model on a machine with ollama up.

**Mutation battery re-run and extended to 33 mutants** covering both fixes and the two defensive branches they introduced: **33/33 killed, zero survivors.** The two that initially survived (`all_keys` vs `live_keys`; the staging-dir scope filter) are unreachable-today defensive branches — rather than delete them or assert equivalence, both now have a direct ledger-seeding test, so the safety property is pinned for the day it becomes reachable.

**No existing assertion was weakened.** `git diff` over all three modified test files shows **zero removed lines**; they are pure additions. `test_frontmatter_records_platform_and_dimensions` is untouched — it asserts `sheet`/`rows`/`columns` on the dict the *ingestor returns*, which is true. The new `test_sheet_rows_and_columns_do_not_survive_to_disk` pins the other half: those keys never reach a file.

**Decision recorded:** `sheet`/`rows`/`columns` are dropped by `Fragment.model_validate` (pydantic's default `extra="ignore"`). Accepted, pinned by test, tracked in #1392 — widening what survives validation is a repo-wide frontmatter change, not a bugfix. `FragmentSource.unit` was rejected outright: `_write_model` dumps with no `exclude_none`, so a nullable field there appends `unit: null` to every fragment from every ingestor.

**Docstring true-up.** `assemble_ingested_fragment` cited four `path:line` callers; **all four had rotted** (`creek/pipeline.py:498` had drifted into `_scan_for_redaction`) and a real caller, the `creek.ingest` MCP tool, was missing. Replaced with symbol references, which cannot rot.

## Follow-ups filed

- #1392 — `Fragment` silently discards ingestor frontmatter it does not model.
- #1393 — an edited spreadsheet is never detected as changed, because `parsed.content` is `""` so the ledger's content hash is constant. **Pre-existing**; this change neither causes nor cures it, and it must specifically **not** be "fixed" by moving the discriminator into `content` — that is exactly what produces the `created=1 updated=2` data loss above.

## Not done

- No tombing/cleanup CLI migration (see *Migration* above — detect-and-report instead, deliberately).
- `creek/fragment.py`'s `FragmentSplitter` still *groups* by `source_path` (grouping by file is arguably right there, and it has no production caller). Its parent-id derivation now uses `identity_key`, which removes a latent sibling collision and is a no-op for every fragment without a unit.

Closes #1305
Refs #1304, #1329, #1392, #1393

### DIFF
--- a/creek-tools/creek/fragment.py
+++ b/creek-tools/creek/fragment.py
@@ -286,8 +286,12 @@ class FragmentationEngine:
         Returns:
             List of child fragments with parent_id metadata.
         """
+        # ``identity_key``, not ``source_path``, so a parent that addresses a
+        # sub-unit of a file (a workbook sheet) derives a distinct parent id
+        # rather than colliding with its siblings (#1305). Identical to
+        # ``source_path`` for every fragment that carries no unit.
         parent_id = generate_fragment_id(
-            parent.source_path,
+            parent.identity_key,
             parent.timestamp,
             parent.content,
         )
--- a/creek-tools/creek/ingest/base.py
+++ b/creek-tools/creek/ingest/base.py
@@ -28,6 +28,7 @@ from pydantic import BaseModel, ConfigDict, Field
 
 from creek._containment import assert_source_contained
 from creek.classify.tags_pass import apply_tags
+from creek.ingest.source_unit import compose_source_unit
 from creek.models import Fragment, FragmentLevel
 
 logger = logging.getLogger(__name__)
@@ -91,7 +92,31 @@ class ParsedFragment(BaseModel):
     """
 
     source_path: str
-    """Path to the original source file."""
+    """Path to the original source file.
+
+    Always the *whole file*, verbatim, even for an ingestor that emits
+    several fragments from it. Downstream consumers rely on that: the
+    per-file titles in :mod:`creek.ingest.code` and :mod:`creek.ingest.generic`,
+    the extension dispatch in :meth:`Ingestor.discover`, and — load-bearing
+    since #1304 — :func:`creek.ingest.routing.arbitrate`, which decides which
+    ingestor owns a file by grouping fragments on exactly this string. A
+    sub-unit discriminator belongs on :attr:`source_unit`, never here.
+    """
+
+    source_unit: str | None = None
+    """The sub-unit of :attr:`source_path` this fragment addresses (#1305).
+
+    ``None`` for the overwhelmingly common one-fragment-per-file case, and
+    for the *single*-sheet workbook — a CSV or a one-sheet XLSX derives
+    exactly the identity it derived before this field existed, which is what
+    keeps every such fragment already in an operator's vault from being
+    re-minted and duplicated on its next ingest.
+
+    Set only where one file genuinely contains several independently
+    identifiable units: today that is ``SpreadsheetIngestor``'s sheets. It
+    reaches identity through :attr:`identity_key` and nothing else, so
+    adding one cannot disturb a consumer of :attr:`source_path`.
+    """
 
     timestamp: datetime
     """Timestamp associated with this fragment."""
@@ -111,6 +136,25 @@ class ParsedFragment(BaseModel):
     markdown, html, code, …) that don't need structured payloads are
     not forced to set a slot they don't use.
     """
+
+    @property
+    def identity_key(self) -> str:
+        """Return the string that identifies this fragment's source (#1305).
+
+        The **only** input any identity derivation may take from a parsed
+        fragment's provenance: :func:`generate_fragment_id`, the provenance
+        entry in :meth:`Ingestor._process_fragment`, and the ledger key in
+        :func:`creek.ingest.pipeline.attach_origin_key` all read this rather
+        than :attr:`source_path`.
+
+        Equal to :attr:`source_path` whenever :attr:`source_unit` is unset,
+        so every ingestor that predates #1305 derives byte-identical ids.
+
+        Returns:
+            ``source_path`` composed with ``source_unit``, or ``source_path``
+            unchanged when there is no unit.
+        """
+        return compose_source_unit(self.source_path, self.source_unit)
 
 
 class ProvenanceEntry(BaseModel):
@@ -404,14 +448,23 @@ def assemble_ingested_fragment(parsed: ParsedFragment) -> IngestedFragment:
 
     Step 4 lives here rather than in each ingestor because this function
     is the **universal ingest chokepoint**: every adapter and every CLI
-    surface funnels through it (``creek/pipeline.py:498`` for ``creek
-    process``, ``creek/ingest/pipeline.py:309`` for ``creek ingest`` /
-    ``creek sync`` / the MCP tool, ``creek/ingest/discord.py:839`` for
-    Discord capture). Wiring it once here is what makes ``tags``
-    populated for markdown, Discord, ChatGPT, Substack and the rest in
-    one place, and it is why the ``creek process`` path needs no separate
-    call further down the pipeline — nothing between here and the write
-    mutates the body the hashtags come from.
+    surface funnels through it —
+    :meth:`creek.pipeline.Pipeline._run_ingestion` for ``creek process``,
+    :func:`creek.ingest.pipeline.run_ingest` for ``creek ingest`` /
+    ``creek sync`` / the ``creek.upload`` MCP tool,
+    :func:`creek_mcp.tools.ingest.ingest_tool` for the ``creek.ingest`` MCP
+    tool, and :mod:`creek.ingest.discord` for Discord capture. Wiring it
+    once here is what makes ``tags`` populated for markdown, Discord,
+    ChatGPT, Substack and the rest in one place, and it is why the ``creek
+    process`` path needs no separate call further down the pipeline —
+    nothing between here and the write mutates the body the hashtags come
+    from.
+
+    Callers are named by *symbol*, not by file and line. The four
+    ``path:line`` citations this paragraph used to carry had every one of
+    them rotted by the time #1305 read them (``creek/pipeline.py:498`` had
+    drifted into an unrelated function), and the MCP ingest tool — a real
+    caller — was missing entirely.
 
     Args:
         parsed: A parsed fragment produced by an ingestor's four-stage
@@ -440,7 +493,7 @@ def assemble_ingested_fragment(parsed: ParsedFragment) -> IngestedFragment:
     body: str = str(parsed.metadata["markdown"])
 
     frontmatter_dict["id"] = generate_fragment_id(
-        parsed.source_path,
+        parsed.identity_key,
         parsed.timestamp,
         parsed.content,
     )
@@ -709,8 +762,10 @@ class Ingestor(abc.ABC):
             ingestor_name: The class name of this ingestor.
             now: The current timestamp for provenance.
         """
+        # ``identity_key``, not ``source_path``: a provenance entry per sheet,
+        # matching the one fragment id per sheet the same key mints (#1305).
         frag_id = generate_fragment_id(
-            fragment.source_path, fragment.timestamp, fragment.content
+            fragment.identity_key, fragment.timestamp, fragment.content
         )
 
         # Stage 3: Convert to markdown
--- a/creek-tools/creek/ingest/ledger.py
+++ b/creek-tools/creek/ingest/ledger.py
@@ -21,6 +21,8 @@ from dataclasses import dataclass
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 
+from creek.ingest.source_unit import SOURCE_UNIT_SEPARATOR
+
 if TYPE_CHECKING:
     from pathlib import Path
 
@@ -151,6 +153,55 @@ class SourceLedger:
     def live_keys(self) -> set[str]:
         """Return the tracked source keys that are not tombed (#674)."""
         return {key for key, rec in self._records.items() if not rec.tombed}
+
+    def all_keys(self) -> set[str]:
+        """Return every tracked source key, tombed or not (#1305).
+
+        The counterpart to :meth:`live_keys`, for the readers that must not
+        treat a tombed record as absence. A soft-tombed unit's fragment was
+        moved to ``10-Liminal/Orphaned/``, not deleted, so a gate deciding
+        whether an identity *owns anything* has to count it — ``live_keys``
+        answers a different question, and answering this one with it is how
+        a gate comes to admit a caller to content that still exists.
+
+        Returns:
+            All tracked source keys.
+        """
+        return set(self._records)
+
+    def records_for(self, base_key: str) -> list[tuple[str, LedgerRecord]]:
+        """Return every record keyed on *base_key* or a sub-unit of it (#1305).
+
+        :meth:`get` answers "what is this file's fragment", which stopped
+        being a question with one answer once one file could hold several
+        independently identified units — the sheets of a workbook. A caller
+        holding only the file's path (the ``creek.upload`` tool holds a
+        staged path, never a sheet name) has no way to enumerate the units
+        without this.
+
+        Both the exact key and its ``<base_key>#<unit>`` children are
+        returned, so a single-unit source still answers with its one record
+        and callers need no special case for the two shapes.
+
+        The ``#`` scan cannot mis-fire for the upload ledger, its only
+        caller: staged filenames come from ``creek_mcp.staged_names.safe_stem``,
+        whose slug regex excludes ``#`` outright, so no staged path can look
+        like a composed key.
+
+        Args:
+            base_key: The whole file's source key.
+
+        Returns:
+            ``(source_key, record)`` pairs ordered by key, so a caller that
+            reduces the list to one id — a response's ``fragment_id`` —
+            picks the same one on every run.
+        """
+        prefix = f"{base_key}{SOURCE_UNIT_SEPARATOR}"
+        return sorted(
+            (key, record)
+            for key, record in self._records.items()
+            if key == base_key or key.startswith(prefix)
+        )
 
     def __len__(self) -> int:
         """Return the number of distinct source keys tracked."""
--- a/creek-tools/creek/ingest/pipeline.py
+++ b/creek-tools/creek/ingest/pipeline.py
@@ -20,12 +20,13 @@ from dataclasses import dataclass, field
 from pathlib import Path  # runtime use: resolving recorded source paths
 from typing import TYPE_CHECKING, Final
 
-from creek.ingest.base import assemble_ingested_fragment
+from creek.ingest.base import assemble_ingested_fragment, generate_fragment_id
+from creek.ingest.source_unit import compose_source_unit
 
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Sequence
     from datetime import datetime
 
     from creek.ingest.base import Ingestor, ParsedFragment
@@ -236,10 +237,25 @@ def attach_origin_key(
     fragment: Fragment,
     vault_path: Path,
 ) -> None:
-    """Stamp the fragment's ``source.origin_key`` before it is written (#672)."""
+    """Stamp the fragment's ``source.origin_key`` before it is written (#672).
+
+    The key addresses the *unit* the fragment came from, not merely the file
+    (#1305). Composed after :func:`derive_source_key` rather than by handing
+    it a pre-composed string, so that function keeps resolving a path that
+    actually exists on disk and the vault-relative/bare-filename choice is
+    made on the real file.
+
+    This is the half of #1305 a per-sheet fragment id does not reach.
+    ``derive_source_key`` keys on the file, so without the unit every sheet
+    of a workbook shares one ``origin_key`` and one ledger record, and
+    :func:`write_fragment_idempotent` then reassigns ``fragment.id`` to the
+    record's id and overwrites sheet 1 with sheet 2 — reporting ``updated``
+    while losing the content.
+    """
     if ledger is None:
         return
-    fragment.source.origin_key = derive_source_key(parsed.source_path, vault_path)
+    base_key = derive_source_key(parsed.source_path, vault_path)
+    fragment.source.origin_key = compose_source_unit(base_key, parsed.source_unit)
 
 
 def record_in_ledger(
@@ -493,6 +509,85 @@ def unpinned_vault_warning(
     return _UNPINNED_VAULT_WARNING
 
 
+_COLLAPSED_UNIT_WARNING_TEMPLATE = (
+    "This vault holds {count} fragment(s) written before #1305, when every "
+    "sheet of a multi-sheet workbook shared one fragment id and only the "
+    "FIRST reached disk. Each one is now superseded by the per-sheet "
+    "fragments this run writes, and is left in place: {sample}. Nothing is "
+    "deleted automatically — inspect with `creek purge source "
+    "--source-path <path> --match exact --dry-run`, and purge only what you "
+    "recognise. Links pointing at a superseded fragment keep resolving to "
+    "it until you do."
+)
+"""Advisory for a vault still holding pre-#1305 collapsed fragments."""
+
+_COLLAPSED_UNIT_SAMPLE = 3
+"""How many superseded ids the advisory names before it stops listing."""
+
+
+def collapsed_unit_warning(
+    writer: VaultWriter,
+    fragments: Sequence[ParsedFragment],
+) -> str | None:
+    """Return the pre-#1305 collapsed-fragment advisory, or ``None`` (#1305).
+
+    Before #1305 every sheet of a multi-sheet workbook derived the same
+    fragment id, so ``_write_model``'s first-writer-wins dedup kept sheet 1
+    and silently dropped the rest. This run writes each sheet under its own
+    id, which is the fix — but it also means the single fragment the old
+    derivation left in an operator's vault is now superseded by N new ones
+    and is not overwritten by any of them.
+
+    **Detected, reported, and left alone.** That is deliberately the same
+    answer #1304 gave the same shape of problem: naming the strays and
+    handing over a read-only inspection command, rather than deleting vault
+    content on the strength of a heuristic. Deleting a fragment is not
+    something an ingest run gets to decide, and the id recomputation below,
+    however exact, cannot know whether the operator has since edited that
+    file, linked to it, or curated it into a thread.
+
+    The detection needs no vault walk, no frontmatter parsing and no
+    scoping heuristic, because the superseded id is *exactly* what this
+    same fragment would have hashed to with no unit — same source path,
+    same timestamp, same content. So it is recomputed from the parsed
+    fragment in hand and looked up in the writer's id index. That also
+    sidesteps the two traps a frontmatter-based detector hits (``ingested``
+    reads back as a ``str``, and its trailing ``Z`` hashes differently from
+    the ``+00:00`` the derivation emits) by never reading frontmatter.
+
+    It follows that the advisory is self-clearing: once the operator purges
+    the superseded fragments, the lookup misses and the run goes quiet. It
+    is also silent on a fresh vault and on every single-unit source, which
+    keeps no unit and therefore has nothing to supersede.
+
+    Args:
+        writer: The vault writer, consulted read-only via
+            :meth:`~creek.vault.writer.VaultWriter.find_fragment`.
+        fragments: This run's parsed fragments, before any are written.
+
+    Returns:
+        The advisory text, or ``None`` when no superseded fragment is
+        present.
+    """
+    superseded: list[str] = []
+    for parsed in fragments:
+        if parsed.source_unit is None:
+            continue
+        legacy_id = generate_fragment_id(
+            parsed.source_path, parsed.timestamp, parsed.content
+        )
+        if legacy_id in superseded:
+            continue
+        if writer.find_fragment(legacy_id) is not None:
+            superseded.append(legacy_id)
+    if not superseded:
+        return None
+    sample = ", ".join(sorted(superseded)[:_COLLAPSED_UNIT_SAMPLE])
+    if len(superseded) > _COLLAPSED_UNIT_SAMPLE:
+        sample = f"{sample}, …"
+    return _COLLAPSED_UNIT_WARNING_TEMPLATE.format(count=len(superseded), sample=sample)
+
+
 def run_ingest(
     *,
     ingestor_cls: type[Ingestor],
@@ -573,6 +668,13 @@ def run_ingest(
     unpinned = unpinned_vault_warning(source_type, vault_path)
     if unpinned is not None:
         warn(unpinned)
+
+    # Same reason as above: checked before the write loop, so the operator
+    # is told about superseded fragments before the run that supersedes
+    # them finishes rather than after (#1305).
+    collapsed = collapsed_unit_warning(writer, ingest_result.fragments)
+    if collapsed is not None:
+        warn(collapsed)
 
     errors: list[str] = [f"[{source_type}] {err}" for err in ingest_result.errors]
     written = 0
--- a/creek-tools/creek/ingest/source_unit.py
+++ b/creek-tools/creek/ingest/source_unit.py
@@ -1,0 +1,120 @@
+"""Compose and split the sub-unit half of a source key (#1305).
+
+Most ingestors map one source *file* to one fragment, so the file's path is
+a sufficient identity. Some do not: ``SpreadsheetIngestor`` emits one
+fragment per sheet of a workbook, and every one of those sheets shares a
+single file. Until #1305 they also shared a single fragment id, a single
+provenance entry and a single ledger record, so ``_write_model``'s
+first-writer-wins dedup (``creek/vault/writer.py``) silently dropped every
+sheet after the first.
+
+The fix is to give such a fragment a *sub-unit* discriminator, and this
+module owns the one spelling of how that discriminator is joined to and
+recovered from a path. Three unrelated layers have to agree on it — the
+ingest derivation that mints it, the upload ledger that stores it, and the
+RTBF purge sweep that has to find the underlying file again — and a
+separator that three modules each spell for themselves is a separator that
+will eventually be spelled two ways.
+
+Why ``#``
+---------
+It is not legal in a ``safe_stem`` slug (``creek_mcp.tools.upload``'s
+``_SLUG_RE`` is ``[^A-Za-z0-9._-]+``), so no staged upload path can contain
+one and the upload ledger's key space stays unambiguous. It is legal in a
+POSIX filename, though, which is why :func:`split_source_unit` is documented
+as a *hypothesis* rather than a parse: a caller must try the whole key first
+and fall back to the split only when the whole key resolves to nothing.
+
+Deliberately free of heavy imports. :mod:`creek.purge.engine` and
+:mod:`creek_mcp.tools.upload` both need it, and neither should be made to
+pull :mod:`creek.classify` in behind it.
+"""
+
+from __future__ import annotations
+
+from typing import Final
+
+SOURCE_UNIT_SEPARATOR: Final[str] = "#"
+"""Joins a source path to the sub-unit of it a fragment addresses."""
+
+_UNIT_SEPARATOR_REPLACEMENT: Final[str] = "-"
+"""What a separator inside a unit name is rewritten to. See :func:`sanitize_unit`."""
+
+
+def sanitize_unit(unit: str) -> str:
+    """Return *unit* with every separator rewritten, so it cannot be ambiguous.
+
+    The separator is legal in the things units are named after — Excel
+    permits ``#`` in a sheet title — and a unit containing one produces a
+    key that cannot be read back. ``book.xlsx#Rev#2`` splits at the *last*
+    separator to ``("book.xlsx#Rev", "2")``, naming a file that does not
+    exist; the RTBF purge sweep then finds no staged source and silently
+    leaves the operator's document on disk.
+
+    Rewriting at mint time is the fix, rather than teaching every reader to
+    try successive splits: it makes "the unit half of a composed key never
+    contains a separator" an invariant of the key space instead of a
+    property each reader has to rediscover. One rpartition is then exact
+    for every key this module produced.
+
+    Callers that derive a unit from user-controlled text must apply this
+    **before** de-duplicating, not after — two distinct names that sanitize
+    to the same string (``Rev#2`` and ``Rev-2``) must be seen as a
+    collision by the de-duplicator, or they silently mint one id for two
+    units, which is the very defect #1305 exists to fix.
+
+    Args:
+        unit: The raw unit name.
+
+    Returns:
+        *unit* with each separator replaced by
+        :data:`_UNIT_SEPARATOR_REPLACEMENT`.
+    """
+    return unit.replace(SOURCE_UNIT_SEPARATOR, _UNIT_SEPARATOR_REPLACEMENT)
+
+
+def compose_source_unit(path: str, unit: str | None) -> str:
+    """Return the source key addressing *unit* within *path*.
+
+    The unit is passed through :func:`sanitize_unit` so a composed key can
+    always be split back apart. That is enforced here, at the single place
+    keys are minted, rather than trusted to each caller.
+
+    Args:
+        path: The source key or path of the whole file.
+        unit: The sub-unit discriminator, or ``None`` for the whole file.
+
+    Returns:
+        ``path`` unchanged when *unit* is ``None`` or empty, else
+        ``f"{path}#{sanitized}"``. An empty unit is treated as no unit
+        deliberately: a key ending in a bare separator would be a second
+        spelling of "the whole file", and two spellings of one identity is
+        how the same fragment comes to be written twice.
+    """
+    if not unit:
+        return path
+    return f"{path}{SOURCE_UNIT_SEPARATOR}{sanitize_unit(unit)}"
+
+
+def split_source_unit(key: str) -> tuple[str, str | None]:
+    """Split *key* into its path half and its sub-unit half, if it has one.
+
+    This is a **hypothesis, not a parse**. ``#`` is a legal character in a
+    POSIX filename, so ``report#1.xlsx`` is indistinguishable from unit
+    ``1.xlsx`` of a file named ``report``. Callers must therefore try the
+    whole key first and consult this only when that resolved to nothing —
+    which is exactly what the purge engine's staged-source sweep does, so
+    a genuinely ``#``-named file keeps winning.
+
+    Args:
+        key: A source key, with or without a sub-unit suffix.
+
+    Returns:
+        ``(path, unit)``, with *unit* ``None`` when *key* carries no
+        separator or when the text after the last separator is empty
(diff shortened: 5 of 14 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real data-loss bug (#1305): every sheet of a multi-sheet workbook derived the same fragment id, so `VaultWriter._write_model`'s first-writer-wins dedup silently dropped all but the first sheet. The fix introduces a `source_unit`/`identity_key` discriminator (deliberately *not* overloading `source_path`, to avoid breaking `routing.arbitrate` from #1304), threads it through the ledger, the RTBF purge sweep, and the upload MCP tool's tier gate, and ships an extensive, well-reasoned test suite. The PR description is unusually rigorous — it documents a migration analysis, two additional privacy defects found and fixed during self-review, and a per-hunk revert-to-prove-load-bearing table.

## Strengths
- Correctly avoids overloading `ParsedFragment.source_path` after recognizing that `creek/ingest/routing.py` (added by #1304, after the original plan was written) groups fragments by that exact string — overloading it would have reintroduced double-writing. This is a genuinely sharp catch.
- Migration handled *by construction*: a single-sheet workbook emits no `source_unit`, so every existing CSV/single-sheet XLSX keeps a byte-identical id and filename (pinned by `TestSingleUnitIdentityIsUnchanged`, with literal `frag-<12hex>` pins under `monkeypatch.chdir` + forced mtime).
- Found and fixed two real security/privacy regressions during self-review (RTBF purge leaving a staged workbook on disk when its `origin_key` names a sub-unit; the #970 tier gate being blindable by purging one sheet of an N-sheet upload and then re-uploading at a lower ceiling) — both with dedicated regression tests, verified to fail when the fix is reverted.
- `sanitize_unit` closes a real hole (Excel permits `#` in sheet titles) applied *before* de-duplication rather than after, avoiding a re-introduction of the exact defect being fixed.
- Extensive test suite across four surfaces (`run_ingest` ledgered/unledgered, the `creek.ingest` MCP tool, `creek process`), including order-stability, tombed-record, and out-of-scope-key regression tests.
- `collapsed_unit_warning` is a good design choice — detects pre-fix collapsed fragments without a vault walk or frontmatter parsing, is self-clearing, and deliberately does not auto-delete anything (consistent with #1304's prior answer to the same shape of problem).

## Security Concerns
🟢 LOW — No new vulnerabilities identified. The two privacy-tier-bypass hazards this change itself introduced (RTBF purge miss, tier-gate blinding via purge-then-reupload) were caught and fixed within the PR, each with a regression test that fails when the fix is reverted. `_resolve_staged_source`'s fallback re-applies both containment guards (`_resolve_pointer_in_vault`, `_in_any_staging_root`) on the split path, and only ever widens deletion *inside* the staging roots — verified by `test_purge_still_refuses_an_out_of_scope_key_with_a_unit_suffix`.

## Problems
🟡 **Duplicate-named sheets get identical titles/headings despite distinct ids** — `SpreadsheetIngestor.generate_frontmatter` (`creek/ingest/spreadsheets.py:592-597`) computes `label = sheet_name or fragment.source_unit`, preferring the raw (un-deduplicated) `sheet.name` over the deduplicated `source_unit`. For two sheets that share a literal name (only reachable via a hand-built/non-Excel-produced workbook — `_sheet_unit_keys` explicitly handles this case with a `~<n>` occurrence suffix, e.g. `"Q"` and `"Q~2"`), both fragments get `title = "book — Q"` and both bodies render `# book.xlsx — Q`. The underlying fragment ids/files remain distinct (no data loss), but the very identity the PR set out to make visible per-sheet is not actually visible for the duplicate-name case its own code defends against. `test_duplicate_sheet_names_still_get_distinct_ids` checks id/file-count distinctness but never asserts on `title`/heading, so this gap isn't caught by the suite. Non-blocking (real-world Excel files cannot produce duplicate sheet names), but worth a one-line fix (e.g. `label = fragment.source_unit if <name occurred more than once> else sheet_name`, or simply prefer `source_unit`) before merge.

## Code Quality
- `_ledgered_names_for_id` (`creek_mcp/tools/upload.py:225-266`) has a duplicated `Args:` section in its docstring (once before "Tombed records are counted too...", once after) — malformed Google-style docstring; harmless to `interrogate`'s presence check but would render oddly under Sphinx napoleon. Easy fix: merge into one `Args:` block.
- The `_sheet_unit_keys` order-stability guarantee ("inserting a differently named sheet leaves every existing sheet's unit untouched") is correct as stated but doesn't extend to inserting a sheet with a name that duplicates an existing one earlier in workbook order — that reassigns the occurrence suffix (and hence the id) of the previously-unique sheet. This is implicitly acknowledged by scoping the claim to "differently named," but could be called out more explicitly given it's the flip side of the very case the code handles.

## CLAUDE.md Adherence
✅ TDD workflow is clearly followed and documented (RED tests described, per-hunk revert table proving each hunk load-bearing).
✅ Mutation testing performed and iterated on (26 → 33 mutants, 0 survivors after adding 4 tests for real gaps) — well beyond what's required, in the spirit of `mutation-testing` skill guidance.
✅ Coverage claimed at 94.43% aggregate / 89.77% for `spreadsheets.py` (floor 80%), well above the 90% branch-coverage bar; `source_unit.py` and `ingest/base.py` at 100%.
⚠️ CI was still in progress at review time (`Tests & Type Checking` 3.11/3.12/3.13, `Static Analysis`, `Pylint` all `IN_PROGRESS`); `Integration & E2E`, `CrawDad Quality` (×3), `Code Complexity`, and `Build Distribution` had already reported `SUCCESS`. The PR body's own Gate 2 numbers (`check-all.sh` exit 0, 12/12) are self-reported and should be confirmed against final CI before merge, per the "Stay Green" 4-gate process.

## Testing
Testing is a real strength here, not a checkbox: four new/extended test files cover the ledgered and unledgered ingest paths, the two MCP surfaces that bypass `run_ingest`, and the purge engine's sub-unit key handling — including adversarial cases (a sheet literally named with `#`, an out-of-vault/out-of-staging traversal attempt with a unit suffix, tombed-record survey, decoy ledger entries outside the staging dir). The PR explicitly calls out and avoids the trap the issue's own bug report fell into (asserting the wrong symptom — "last sheet survives" vs. actually "first"). The one gap is the duplicate-title assertion noted above.

## Documentation
Docstrings are thorough to the point of being reference material — `source_unit.py`'s module docstring, `identity_key`, and `collapsed_unit_warning` all explain *why*, not just *what*, and stale `path:line` citations were proactively replaced with symbol references. Only ding is the duplicated `Args:` block noted in Code Quality.



## Verdict: COMMENTS

Nothing here blocks merge — the core fix is correct, thoroughly tested (including two self-caught privacy regressions with regression tests), and well-documented — but there is a substantive, non-blocking gap worth the author's attention: duplicate-named sheets (the exact edge case `_sheet_unit_keys` was written to handle) get distinct fragment ids yet identical titles/headings because `generate_frontmatter` prefers the raw, un-deduplicated sheet name over the deduplicated `source_unit`, and no test catches it. A minor docstring duplication in `_ledgered_names_for_id` is a secondary nit.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5252026198 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5252026198 to see in-depth feedback and continue iterating

Geoffe-Ga:
## Review finding addressed — the disambiguation stopped at the index

The finding is **real**, and it is fixed in `bcbcac7` rather than filed, because it sits in this PR's own headline feature.

### Verified

`generate_frontmatter` derived the title from the raw, un-deduplicated sheet name:

```python
sheet_name = str(fragment.metadata.get("sheet", "")).strip()
label = sheet_name or fragment.source_unit   # raw name wins
title = f"{stem} — {label}" if fragment.source_unit else stem
```

`metadata["sheet"]` is `sheet.name` verbatim, so two sheets both named `Data` carry the identical string. `_sheet_unit_keys` had already resolved them into `Data` and `Data~2`; nothing consumed that answer.

### RED first — identical titles on disk

New test, two sheets named `Data`, asserting on file bytes after a real `VaultWriter` round-trip:

```
>       assert titles == ["book — Data", "book — Data~2"]
E       AssertionError: assert ['book — Data', 'book — Data'] == [...]
E         At index 1 diff: 'book — Data' != 'book — Data~2'
```

Two fragments, distinct ids, two files, and text an operator cannot tell apart. Worth noting the filenames collided too — `VaultWriter` had to invent a `-1` de-collision suffix, so the only distinguishing marks were a `frag-<12hex>` id and that suffix.

### Sibling consumers of the raw sheet name

Swept every reader of `metadata["sheet"]` and `sheet.name`. `SpreadsheetIngestor` is the only producer of `source_unit` in the tree, so the blast radius is one module:

| Site | Verdict |
|---|---|
| `generate_frontmatter` → `title` | **The finding.** Fixed. |
| `convert_to_markdown` → the body's `# ` heading | **Same bug, same cause.** Fixed. This is the other string an operator actually reads. |
| `parse` → `metadata["sheet"] = sheet.name` | Correct as-is. The raw name is the honest provenance record. |
| `generate_frontmatter` → the `sheet` frontmatter key | Left alone. Raw provenance, and it never reaches a file anyway (dropped by `extra="ignore"`; #1392). |
| `collapsed_unit_warning` advisory | Names fragment ids, not sheet names. Not affected. |
| `PresentationIngestor` | Emits one fragment per *deck*, carries no `source_unit`. Not affected. |

### The fix

Both derivations now go through one `_sheet_label`, which returns `source_unit` when there is one. No special-casing of duplicates — the dedup already happened upstream; this is only about consuming it. A side benefit: the visible label and the `…#<unit>` source key become the same string, so a fragment matches by eye to the ledger and purge records that name it.

The no-unit fallback (a CSV, a one-sheet XLSX — nothing to be confused with) is the previous expression **verbatim**, so no fragment already in a vault is re-titled and the `_PIN_*` identity/filename pins are untouched.

### Mutation battery

Each mutation ran behind a baseline-green guard, so a pytest exit code meaning "harness broken" could not be misread as a kill:

| Mutant | Result |
|---|---|
| Exact verbatim revert of the title change | baseline exit=0 → **exactly the new test** red (1 failure) → restore exit=0 |
| Heading reverted to the raw name | baseline exit=0 → **exactly the new test** red (1 failure) → restore exit=0 |
| Title reverted *and* blank-name fallback dropped | 2 red, including the pre-existing filename test |

### Also

Dropped the duplicated `Args:` block in `_ledgered_names_for_id`.

### Gates

`./scripts/check-all.sh` → **exit 0**, 12/12 checks, 9131 passed, coverage 94.43%. Bypass sweep over the whole branch diff *including tests*: zero `noqa` / `type: ignore` / `pragma: no cover` / skip / xfail added, and no threshold, waiver, script or CI file touched by the branch at all.

Out of scope and deliberately untouched: the migration decision, the `sheet`/`rows`/`columns` drop (#1392), and #1393.

Geoffe-Ga:
## Summary
This PR fixes a data-loss bug where every sheet of a multi-sheet workbook derived the same fragment id, so the vault writer's first-writer-wins dedup silently dropped all but the first sheet. It introduces a `source_unit` discriminator on `ParsedFragment`, threads it through fragment-id generation, the upload ledger, frontmatter titles/headings, and the RTBF purge sweep, while preserving byte-identical identity for every existing single-sheet source (the migration story). Two additional security-relevant regressions the fix itself would have opened — an RTBF purge that left the staged workbook on disk, and a #970 tier-gate bypass reachable via purge-then-reupload — are also closed in the same change.

## Strengths
- The root-cause analysis is exact and independently verifiable: `write_fragment_idempotent` (creek/ingest/pipeline.py:352) really does `fragment.id = record.fragment_id` unconditionally, which is exactly why a per-sheet fragment id alone would not have fixed the ledgered upload path — the PR correctly identifies and fixes both halves (fragment id *and* ledger key).
- The migration design (no discriminator for single-non-empty-sheet sources) is sound and pinned by dedicated regression tests (`_PIN_XLSX_SINGLE_SHEET_ID`, `_PIN_CSV_ID`, `_PIN_PRESENTATION_ID`) that would catch any accidental re-minting of ids already living in real vaults.
- The purge fallback (`_resolve_staged_source` / `_staged_key_candidates`) correctly re-applies both containment guards on the fallback path and tries the whole key before splitting — I traced `_resolve_pointer_in_vault`'s containment check and `_in_any_staging_root` and confirmed the traversal/out-of-scope test cases (`test_purge_still_refuses_an_out_of_scope_key_with_a_unit_suffix`) are handled correctly by the actual code, not just asserted to be.
- The #970 tier-gate fix (surveying the ledger via `_ledgered_names_for_id` / `all_keys()` rather than the staging directory) is the right call — a ledger record outliving the staged file is precisely the condition that made the old directory-listing survey a privacy bypass after a partial purge.
- Extensive, well-targeted new tests covering the two bypassed `run_ingest` surfaces (the `creek.ingest` MCP tool and `creek process`), sheet-name collision/sanitization, and the RTBF purge sub-unit path.
- `source_unit` defaults to `None` and every other ingestor (`markdown.py`, `discord.py`, `code.py`, etc.) is unaffected — confirmed no other `ParsedFragment(...)` call site sets it.

## Security Concerns
🟢 LOW — No new vulnerabilities found. Both security-relevant fixes in this PR (RTBF purge fallback, tier-gate ledger survey) were independently verified against the actual containment-guard code, not just the PR's own description, and both hold up. One pre-existing, unrelated stem-matching quirk (`name.startswith(f"{stem}.")` in `_ledgered_names_for_id`, mirrored from the pre-existing `_staged_for_id`) could theoretically let an external_id whose stem is a dotted prefix of another id's stem pick up its ledger records, but this pattern already existed in `_staged_for_id` before this PR and is not a regression introduced here — not blocking.

## Problems
None found that block merge.

## Code Quality
- **Dead code**: `_staged_for_id` (creek_mcp/tools/upload.py:280) is no longer called anywhere — `_refuse_unadmitted_overwrite` and `_refuse_extension_conflict` were both migrated to `_ledgered_names_for_id`. Verified via repo-wide grep: only its own definition remains. It's a fully-documented ~30-line function that should be removed as part of this PR rather than left as refactor debris (and it's untested, which is a small hole in what the coverage numbers actually mean).
- The `collapsed_unit_warning` advisory design (recompute the legacy id, look it up read-only) is clever and appropriately conservative — detects and reports rather than auto-deleting, consistent with the project's prior stance on #1304.

## CLAUDE.md Adherence
✅ TDD workflow is evident throughout (tests document RED-before-fix state with exact pre-fix numbers).
✅ Commit messages follow Conventional Commits (`fix(ingest): ...`).
❓ Coverage/lint/typecheck/complexity claims (94.43% branch coverage, 12/12 checks, mutation testing) are asserted at length in the PR body but I was not able to independently re-run the full `check-all.sh` suite within this review to confirm them; the diff itself is consistent with those claims (thorough docstrings, typed signatures, no obvious complexity red flags).
❌ Minor gap: leftover dead code (`_staged_for_id`) is exactly the kind of refactor debris `check-all.sh`'s docstring/complexity gates wouldn't catch but `vulture` (lint-extended.sh) would — worth a follow-up pass before merge.

## Testing
Testing is a clear strength of this PR. New/modified test files (`test_ingest_ledger.py`, `test_mcp_upload.py`, `test_multi_sheet_surfaces.py` [new], `test_purge_sheet_units.py` [new], `test_spreadsheet_presentation_ingestors.py`) are all pure additions (confirmed via the diff — zero deleted lines in existing test files), covering: the ledgered and unledgered ingest paths, sheet-name collisions and sanitization, filename/title/heading de-duplication, RTBF purge sub-unit resolution and containment, the #970 tier gate under purge-then-reupload, and the two `run_ingest`-bypassing surfaces (MCP `creek.ingest` tool, `creek process`). The migration carve-out is pinned with literal pre-computed fragment-id hashes rather than relative assertions, which is a meaningfully stronger guarantee against silent regression.

## Documentation
Docstrings throughout the new and modified code are unusually thorough — each explains not just what a function does but why, including references to the specific defect it closes (#1305) and prior related work (#672, #970, #1304, #1329). The PR description itself is exceptionally detailed, including a measured before/after data table and a full self-review discovering and fixing two additional defects before this review even started.



## Verdict: COMMENTS

The fix is correct and thoroughly tested — I independently traced the core defect (shared fragment id/ledger key across sheets), the two follow-on security fixes (RTBF purge sub-unit fallback with re-applied containment guards, and the #970 tier gate's switch to a ledger-based survey), and the single-sheet migration carve-out, and all hold up against the actual code. Nothing here blocks merge. The one substantive item worth the author's attention: `_staged_for_id` in `creek_mcp/tools/upload.py` (~line 280) is now dead code — its callers were all migrated to `_ledgered_names_for_id`, and nothing in the module or test suite calls it anymore.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5252251557 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: COMMENTS
**Action**: pull comment 5252251557 to see in-depth feedback and continue iterating

Geoffe-Ga:
Thanks — good catch, and it was my debris, so I've closed it here rather than filing it.

## `_staged_for_id` deleted

Confirmed dead before removing it. On `origin/main` it had two callers (`upload.py:258` and `upload.py:331`); this PR migrated both to `_ledgered_names_for_id`, leaving zero. A repo-wide search across `creek/`, `creek_mcp/` and `tests/` returned exactly one hit — the definition itself:

```
creek_mcp/tools/upload.py:280:def _staged_for_id(vault_path: Path, external_id: str) -> list[Path]:
```

No test referenced it. Removed in full (33 lines with its docstring). Its two dependencies, `safe_stem` and `UPLOAD_STAGING_RELDIR`, are still used by `_ledgered_names_for_id`, so nothing else was stranded by the removal.

## Swept the rest of the diff for the same failure mode

Fair prompt to check — a survey-to-ledger migration can strand more than one helper. I enumerated every symbol this PR adds and counted non-definition references across `creek/`, `creek_mcp/` and `tests/`:

| new symbol | refs |
|---|---|
| `identity_key` | 7 |
| `all_keys` | 2 |
| `records_for` | 1 |
| `collapsed_unit_warning` | 1 |
| `sanitize_unit` | 7 |
| `compose_source_unit` | 15 |
| `split_source_unit` | 13 |
| `find_fragment` | 2 |
| `_sheet_unit_keys` | 4 |
| `_sheet_label` | 4 |
| `_resolve_staged_source` | 1 |
| `_staged_key_candidates` | 1 |
| `_resolve_fragment_ids` | 2 |
| `_ledgered_names_for_id` | 2 |

Plus a zero-caller sweep over every module-level `def _…` and every private method in the eight source files this PR touches. `_staged_for_id` was the only one. (The sweep also surfaced `creek/vault/writer.py::_generate_filename` as zero-caller, but it is identical on `origin/main` and untouched by this diff — pre-existing, not this PR's debris, so I've left it alone.)

## Why the gate didn't catch it

You're right that `vulture` would have flagged this, and it's worth being precise about why it didn't, because the answer is both of the things one might guess — and the second is broader than this PR.

**1. `vulture` runs in no blocking gate.** It is absent from `scripts/check-all.sh`, and `.github/workflows/ci.yml` has no `vulture` step, no `lint-extended.sh` step, and no `pre-commit` step at all. `scripts/lint-extended.sh:4` says so in its own header — *"Status: OPTIONAL — not currently invoked from check-all.sh or CI."* The only workflow hits for `vulture` are `scan-dead-code.yml` / `_claude-scan.yml` (scheduled advisory scans, not PR gates). So the pre-commit hook is the sole place it runs, and only locally.

**2. Even where it runs, two independent suppressions make a zero-caller function invisible.** `.pre-commit-config.yaml:119-126`:

```yaml
- repo: https://github.com/jendrikseipp/vulture
  rev: v2.10
  hooks:
  - id: vulture
    args:
    - creek-tools/creek/
    - --min-confidence
    - '80'
```

- **Path scope:** the survey is `creek-tools/creek/` only. This file is `creek-tools/creek_mcp/tools/upload.py` — outside it. `creek_mcp/` gets no dead-code check whatsoever. `scripts/lint-extended.sh:86` matches (`vulture creek/ --min-confidence 80`), and lines 75-78 state the intent: those tools "stay on `creek/` until their `creek_mcp` backlog is groomed."

- **`--min-confidence 80` categorically excludes the entire dead-code tier.** Vulture scores *unused function / method / class / attribute* at **60%**; unused imports at 90%; unreachable code and unused variables at 100%. A floor of 80 therefore drops precisely and only the 60% tier. Demonstrated against this very file:

  ```
  $ vulture creek_mcp/tools/upload.py --min-confidence 80
  (no output, exit 0)

  $ vulture creek_mcp/tools/upload.py --min-confidence 60
  creek_mcp/tools/upload.py:280: unused function '_staged_for_id' (60% confidence)
  ```

So the honest characterisation is that **as configured, the `vulture` hook cannot report a dead function, method, class or attribute anywhere in the repo — including inside `creek/`, where it does run.** What remains in scope for it is unused imports and unused variables, which ruff's `F401`/`F841` already gate in `check-all.sh`. That is a general suppression, not a quirk of this path, so I'm flagging it rather than quietly working around it.

One caveat against a naive fix: dropping straight to `--min-confidence 60` is not the answer. The same run also flags `upload_tool` (live — called at `creek_mcp/server.py:494`, exported at `creek_mcp/tools/__init__.py:27`), because vulture is single-module and cannot see cross-module callers, so every public entry point reads as dead. The threshold bump looks like a blunt substitute for a whitelist. A real fix would be a 60% run plus a whitelist file, or a 60% run scoped to `_`-prefixed module-private symbols where the false-positive class largely disappears.

I have **not** changed any gate configuration on this PR — that's a separate decision, and this branch is not the place for it.

## Verification

- `./scripts/check-all.sh` → **exit 0**, 12/12 passed, 9131 passed / 5 skipped, total coverage 94.45%.
- Per-file gate holds: **259 files ≥ 80%**, 2 waivered ≥ 65%.
- Coverage moved the right way. Measured apples-to-apples on `tests/test_mcp_upload.py`, before vs after:

  | | statements | missing | branches | coverage |
  |---|---|---|---|---|
  | before | 146 | 11 | 46 | 91.67% |
  | after | 140 | 6 | 44 | **95.11%** |

  Six statements left the denominator; five of them (the old lines 302-306) were uncovered, and the sixth was the `def` line itself, executed at import. Net **+3.44pp** on `creek_mcp/tools/upload.py`.
- Bypass sweep over the full branch diff including tests (#1366): no added `noqa`, `type: ignore`, `nosec`, `pragma: no cover`, `skip`/`xfail`, threshold edits, or `|| true`.

Geoffe-Ga:
## Summary
Fixes #1305: every sheet of a multi-sheet workbook previously derived the same fragment id, so `VaultWriter._write_model`'s first-writer-wins duplicate check silently dropped every sheet after the first (the issue body said "last sheet survives"; this PR correctly identifies and proves it's the *first*). The fix adds a `ParsedFragment.source_unit` / `identity_key` discriminator (kept off `source_path` deliberately, since `routing.arbitrate` groups ingestor ownership on that exact string), threads it through the upload ledger (`attach_origin_key`, `SourceLedger.records_for`), and closes two privacy ratchets the change would otherwise open in the RTBF purge sweep and the #970 tier-overwrite gate.

## Strengths
- Every material claim in the PR description was independently verified against the current source: `_write_model`'s lock-held existing-id early return (confirms "first sheet wins"), `routing.arbitrate`'s grouping on `fragment.source_path` (justifies keeping the discriminator off that field), and `ledger_record`'s lookup by `fragment.source.origin_key` (confirms the per-sheet ledger key is what makes the ledgered path per-sheet).
- Migration handled by construction rather than backfill: a workbook with exactly one non-empty sheet emits `source_unit=None`, so every existing CSV/single-sheet-XLSX fragment keeps a byte-identical id and filename — pinned by dedicated tests with forced mtime and relative paths.
- Two security regressions this change would otherwise introduce are caught and fixed in the same PR (per the description, found during the author's own Gate 2.5 self-review): the RTBF purge sweep silently leaving a staged workbook on disk when a fragment's `origin_key` names a sub-unit, and the #970 tier gate being blinded by a staging-directory listing after a sibling sheet's purge deletes the shared staged file. Both are fixed at the right layer (survey the ledger, which outlives the file) with targeted regression tests.
- `sanitize_unit` closes a real edge case (`#` legal in Excel sheet titles) applied *before* de-duplication, with a test proving the ordering matters (`Rev#2` / `Rev-2` don't collide).
- Extensive test suite across all four ingest surfaces (`run_ingest` ledgered/unledgered, the `creek.ingest` MCP tool, `creek process`), plus a self-clearing "un-migrated vault" advisory that detects (never deletes) pre-fix collapsed fragments, consistent with the read-only-detection precedent set by #1304.
- Confirmed `_staged_for_id` was fully removed with no leftover references anywhere in the tree.

## Security Concerns
None outstanding — both defects identified above (🟢 the purge/RTBF gap and the tier-gate bypass) are fixed in this PR with regression tests, not merely reported.

## Problems
None found blocking merge.

## Code Quality
- Non-blocking observation: `_ledgered_names_for_id` and `_resolve_fragment_ids` both call `SourceLedger.load` independently within the same request path (e.g., `_refuse_unadmitted_overwrite` then later `upload_tool`'s audit resolution) — a minor duplicate I/O cost, not a correctness issue, and likely already how the pre-existing code worked.
- The `collapsed_unit_warning` recomputation walks all `ingest_result.fragments` doing an id-index lookup per multi-unit fragment; fine at expected vault/ingest scale, flagged only for awareness if very large batch ingests become common.

## CLAUDE.md Adherence
✅ TDD workflow followed and documented (RED tests described, each hunk proven load-bearing by targeted revert). ✅ Scripts-based workflow referenced (`./scripts/check-all.sh` reported 0/12 checks failing, 94.43% aggregate branch coverage). ✅ No shortcuts/suppressions introduced. ✅ Conventional commit messages, feature branch. ⚠️ CI was still `IN_PROGRESS` on several required jobs (Static Analysis, Pylint, Tests & Type Checking 3.11/3.12/3.13) at review time — Integration & E2E, Complexity, and Build jobs had already passed. This isn't a code issue, just worth confirming green before merge per the "Stay Green" gate.

## Testing
Very strong. New/modified test files: `test_ingest_ledger.py`, `test_mcp_upload.py`, `test_multi_sheet_surfaces.py` (new), `test_purge_sheet_units.py` (new), `test_spreadsheet_presentation_ingestors.py`. Coverage spans the migration pin (byte-identical ids for single-unit sources), duplicate/blank sheet-name disambiguation (both id and user-visible title/heading), ledger stability under sheet reordering, the two security-fix regressions, and mutation-testing to 33/33 killed on the newly introduced logic. `git diff` on test files shows pure additions — no weakened assertions.

## Documentation
Docstrings are unusually thorough and explain *why*, not just *what* (e.g., `source_unit.py` module docstring explaining the separator choice and why it can't collide with staged filenames). The PR description itself functions as a de facto ADR for the identity-key design decision and correctly documents the deviation from the originally-judged plan (routing.py's introduction in #1387 postdates the plan). `assemble_ingested_fragment`'s stale `path:line` citations were also proactively replaced with symbol references to prevent rot — a nice quality-of-life improvement beyond the ticket's scope.


## Verdict: LGTM

The fix is correct, verified against the actual source (VaultWriter._write_model's first-writer-wins dedup, arbitrate's source_path grouping, ledger_record's origin_key lookup all behave exactly as the PR describes), the migration carve-out is sound and independently confirmed, and the two security regressions the change would otherwise have introduced (RTBF purge miss, #970 tier-gate bypass) are fixed with direct regression tests. Test coverage is exceptionally thorough (mutation-tested, each hunk proven load-bearing by revert) and no dead code or loose ends remain (`_staged_for_id` fully removed). Nothing here rises to a blocking or even a substantive non-blocking concern.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5252602344 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1354, opened 2026-08-11
fix(author): a vault-deployed medium contract can raise the privacy ceiling and disarm the HARD leak gate


Geoffe-Ga:
Found during the Gate 2.5 review of #1310 (PR #1352). **Pre-existing, not introduced there** — `run_author` (and therefore the already-correct MCP verb) has trusted a vault-deployed contract's `default_privacy_tier` since before #1310. That PR only made the CLI consistent with the same trust boundary.

## Problem

`check_privacy_compliance` (`creek/author/checks.py`) takes its ceiling from `contract.default_privacy_tier` and flags a cited fragment only when `rank > ceiling`. `load_medium_contract` prefers the **vault-deployed** `00-Creek-Meta/Skills/mediums/<medium>.MEDIUM.md` over the packaged template (`creek/author/contracts.py:100-117`).

So a contract deployed into the vault declaring:

```yaml
default_privacy_tier: intimate
```

raises the ceiling to rank 2. Cited INTIMATE fragments are also rank 2, so `rank > ceiling` is false and the verbatim-leak check never fires for them. Only `unclassified` (the fail-closed rank 3) would still be caught. The HARD gate is disarmable from inside the vault by editing one YAML line — a file that is not obviously security-relevant to whoever edits it, and that `creek skills sync` may not overwrite.

Note the sibling property is already defended: `ReflectionNode.review` does `del rubric` (`creek/author/reflection.py:102`), so a contract cannot weight `privacy_compliance` to `0.0` and buy a clean verdict. #1310 added `tests/test_reflection.py::test_rubric_cannot_soften_a_verdict` to pin that. The ceiling is the remaining lever.

## Proposed fix

Give `check_privacy_compliance` an independent floor that a contract cannot widen — e.g. from `creek_config.yaml` — and take the **more restrictive** of the floor and the contract's declared tier. A vault-deployed contract could then only narrow what a draft may reproduce, never widen it.

Worth deciding explicitly whether a legitimate medium ever *should* be allowed to reproduce intimate text (a private journal-summary medium, say). If so the escape hatch belongs in config the operator sets deliberately, not in a skill file that deploys by default.

## Acceptance criteria

- [ ] A vault-deployed contract declaring `default_privacy_tier: intimate` does not stop the HARD gate flagging verbatim reproduction of an INTIMATE fragment.
- [ ] The floor's source is documented in `docs/writing-desk.md`.
[...]

PR #1396, opened 2026-08-11
fix(skills): guard medium contracts from silent overwrite, at the primitive

### LINKED ISSUE #1306: fix(skills): creek skills sync silently overwrites hand-edited mediums/*.MEDIUM.md — the drift guard only checks *.SKILL.md
**Severity:** High
**Remediation:** fix (extend the drift guard to the second file class) + regression test

## Problem

`creek skills sync` silently overwrites hand-edited `mediums/*.MEDIUM.md` contracts. The
drift guard that exists to prevent exactly this only inspects `*.SKILL.md`.

`deploy_skills` deploys **two** file classes — `*.SKILL.md` (`creek/scaffold.py:98`) and
`mediums/*.MEDIUM.md` (`creek/scaffold.py:101-107`). `detect_drifted_skills`
(`creek/scaffold.py:131`) globs only `SKILLS_TEMPLATE_DIR.glob("*.SKILL.md")`.
`creek/cli.py:3846-3856` uses `detect_drifted_skills` as the **only** gate before an
unconditional `shutil.copy2`, so a locally-edited MEDIUM contract is neither detected nor
protected by `--force`.

Taxonomy family 0 (correctness) — silent loss of user-authored vault content, which is the
precise thing the `--force` gate exists to prevent. FEAT-041 §5 medium contracts are
user-tunable by design.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Deploy to a temp vault, hand-edit
`mediums/essay.MEDIUM.md`, then:

```
detect_drifted_skills(vault) -> []          # sync proceeds without --force
deploy_skills(vault)         -> the edit is gone
  file now reads: '---\nmedium: essay\nstructure:…'   (the canonical template)

control: editing lint.SKILL.md IS detected -> ['…/00-Creek-Meta/Skills/lint.SKILL.md']
```

The control case proves the guard works for the class it covers, and that the MEDIUM class
falls straight through it.

**Signal 2 — structural.** The asymmetry is between two functions ~40 lines apart in one
module: `deploy_skills` iterates two globs, `detect_drifted_skills` iterates one. Nothing in
either docstring says MEDIUM contracts are intentionally unprotected — `detect_drifted_skills`
describes itself generically as returning *"deployed skill paths whose bytes diverge from the
canonical template"*.

## Scope

Covers drift detection for the second deployed file class in `creek/scaffold.py` and the
`creek skills sync` gate that consumes it.

Explicitly NOT in scope: the privacy-ceiling threading defects in `creek skills generate`
(#971, #1280, #1281, #1284), or the ontology-spec deploy path.

## Tasks

1. Extend `detect_drifted_skills` (`creek/scaffold.py:131-140`) to also walk
   `SKILLS_TEMPLATE_DIR / "mediums"` for `*.MEDIUM.md`, comparing against
   `<vault>/00-Creek-Meta/Skills/mediums/<name>`.
2. Confirm `creek/cli.py:3846-3856` surfaces the MEDIUM paths in its drift message so the
   operator sees which files `--force` would overwrite.
3. Add regression tests for both classes: an edited `*.SKILL.md` and an edited
   `mediums/*.MEDIUM.md` are each reported as drifted, and each survives a sync without
   `--force`.

## Acceptance Criteria

- [ ] A hand-edited `mediums/essay.MEDIUM.md` is returned by `detect_drifted_skills`.
- [ ] `creek skills sync` without `--force` refuses (and names the file) when only a MEDIUM
      contract is drifted.
- [ ] `creek skills sync --force` still overwrites both classes.
- [ ] Tests cover both file classes and fail before the fix.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/scaffold.py` | Modify |
| `creek-tools/creek/cli.py` | Modify (drift message, if needed) |
| `creek-tools/tests/test_skills.py` | Modify (add both regression tests) |

Geoffe-Ga:
## Summary

`creek skills sync` silently destroyed hand-edited `<vault>/00-Creek-Meta/Skills/mediums/*.MEDIUM.md`. `deploy_skills` wrote 14 files — 8 `*.SKILL.md` **and** 6 `mediums/*.MEDIUM.md` — while `detect_drifted_skills` globbed only the first class. Two enumerations of "what is a skill file" had diverged, and the guard was watching one of them.

Scoping the fix to that glob would have left the destruction half-closed, so this PR goes further. Four things were verified broken **by execution**, not by reading:

1. A hand-edited `mediums/essay.MEDIUM.md` yielded `detect_drifted_skills(vault) -> []`; `creek skills sync` exited **0**, printed `Synced 14 skill file(s)`, and replaced the operator's bytes. (Control: editing `lint.SKILL.md` *was* detected.)
2. `creek init --refresh` reached `deploy_skills` through a **second, entirely unguarded call site** (`deploy_canonical`) and destroyed **both** classes plus `AGENTS.md` — including the `*.SKILL.md` class the guard supposedly already covered. Measured: `SKILL survived refresh? False / MEDIUM survived refresh? False / AGENTS survived refresh? False`.
3. `docs/getting-started.md` actively recommended that path with the words "without touching user content". False.
4. `README.md` claimed sync already "refuses to overwrite locally-modified skill files". False for the medium class.

### The fix

- **One enumeration.** `_canonical_skill_pairs()` is now the module's only answer to "what is a skill file"; drift detection and deployment both read it, so they cannot disagree again. That divergence *was* this bug.
- **The guard moved into the destructive primitive.** `deploy_skills(vault, *, force: bool = False)` raises `DriftedSkillsError` itself. `force=False` is the default, so a future third caller is safe by construction rather than by convention. Leaving the guard in the CLI would have re-created the exact defect one layer up.
- **`deploy_canonical` pre-flights the check**, so a refused `creek init --refresh` deploys *nothing* — no folders, no ontology spec, no `AGENTS.md`. A refusal that had already overwritten `AGENTS.md` would leave the operator unable to tell what landed.
- **The refusal names the file unambiguously**, as a path relative to the Skills root (`mediums/essay.MEDIUM.md`), one per line. It previously printed a bare `p.name`, which could not distinguish a medium contract from a top-level skill.

## The `--force` policy — decided deliberately, not inherited

The question this issue forces is *what happens to an edit `--force` cannot preserve*. Silent discard is what `*.SKILL.md` did before this PR. For a data-loss issue that is the wrong default to leave unexamined, so it is changed:

> Canonical skill deployment **refuses** when a deployed skill file's bytes diverge from its canonical template. `--force` overwrites, but **preserves the operator's version alongside as `<name>.bak`** before the canonical bytes land. A later `--force` overwrites that backup. The policy is **identical for `*.SKILL.md` and `mediums/*.MEDIUM.md`**, and **identical for `creek skills sync` and `creek init`**, because the guard lives inside `deploy_skills`.

That policy is recorded in four places that the operator can actually reach: the `--force` help on both commands, the `creek init` docstring, the refusal message itself, and `docs/getting-started.md` + `README.md`.

**Honest limits of the `.bak` scheme** (probed by execution, not assumed):
- A pre-existing operator-authored `essay.MEDIUM.md.bak` **is** overwritten by the first `--force` — not just a backup from an earlier `--force`. The single deterministic name is the trade: numbered backups were rejected because unbounded `.bak.1`, `.bak.2` litter in a browsed Obsidian vault has no lifecycle and nothing in creek ever lists or cleans it. This is now disclosed in the refusal message itself ("replacing anything already at that path") and in `_preserve_local_edits`'s docstring, rather than left for someone to discover.
- Verified non-interference three ways by test: after a forced sync `detect_drifted_skills(vault) == []`; neither `*.SKILL.md` nor `mediums/*.MEDIUM.md` globs match a `.bak`; and `creek lint`'s skill-size-budget sweep produces no finding naming one. The sidecar is deliberately named so it cannot re-enter `creek/lint/checks/skill_size_budget.py` or `creek/author/contracts.py:_resolve_contract_path`.

## `AGENTS.md` is intentionally forfeit — and the pin is left alone

`tests/test_vault_sovereignty.py::test_init_refresh_overwrites_canonical_material` pins the *opposite* doctrine: it hand-edits `AGENTS.md` and asserts the canonical bytes win. **That test is deliberately untouched and still green.** `AGENTS.md` is the machine-facing agent contract whose canonical version is the intended one; a vault `MEDIUM` contract, by contrast, changes what `creek author` actually emits (`creek/author/contracts.py` prefers the vault copy over the template), so an edit there is load-bearing on output. The two are not the same kind of file and do not get the same guarantee. Every `creek init` run still overwrites `AGENTS.md`, and the help text and docs now say so instead of claiming user content is untouched.

## Consequence for #1354 — stated plainly

[#1354](https://github.com/Geoffe-Ga/Creek-Vault/issues/1354) reports that a vault-deployed medium contract can raise the privacy ceiling and disarm the HARD leak gate. Today a tampered `mediums/*.MEDIUM.md` is *incidentally* reverted the next time anyone runs `creek skills sync`. **After this change, sync refuses and the tampered contract persists.**

This PR therefore **strictly increases the persistence of a ceiling-widening contract**, and makes #1354's independent, config-sourced floor more load-bearing, not less. That is accepted: a bug that sometimes clobbers tampering is not a security control, and the new refusal actually *names the drifted file*, which is better detection than silent reversion. **Nothing anywhere should now cite "sync will restore the canonical contract" as a mitigation for #1354 — after this merges it is no longer true.**

## Two smaller decisions worth surfacing

- **Two explicit globs, not an `rglob` walk.** A whole-tree sweep would ship any stray `.md` dropped into `creek/templates/skills/` into every user vault *and* silently promote it into the watched set. It would also discard the stripped-wheel `is_dir()` guard, whose behaviour depends on pathlib globbing that changed in 3.13 — and CI runs 3.11/3.12/3.13. The two-glob form mirrors the correct in-tree precedent at `creek/lint/checks/skill_size_budget.py`.
- **`Refusal.mutate` added to the wiring-contract harness.** A drift gate could not be declared at all before: `Refusal` had no way to create pre-existing local state, and a vault fresh from `creek init` has no drift by definition. An undeclarable gate is an unwatched one — that is precisely how this bug survived in a suite whose job is to catch unwired surfaces. The `skills sync` surface also gains `00-Creek-Meta/Skills/mediums/*.MEDIUM.md` in `effect.writes`; declaring only the schema skills is what let the guard cover one class for a whole release unnoticed.

## Gate 2.5 self-review: verdict CLEAN, four non-blocking findings fixed anyway

- **The refusal named the wrong command.** `creek init` printed "Refusing to **refresh**" even without `--refresh` (reachable when the config file is missing but a drifted skill tree exists), and then advised re-running with `--refresh --force` — which would have silently skipped `generate_default_config`, so the suggested "fix" would have done *less* than the command it replaced. The verb and the remedy are now supplied by the caller and match the flags actually passed.
- **`.bak` collision disclosed** (above) rather than left implicit.
- **An overstated atomicity claim retracted.** `deploy_canonical`'s docstring said "the whole deployment is all-or-nothing". True of a *drift refusal*; not true of an I/O error partway through the four sequential steps. Scoped to what is actually guaranteed — the point of this PR is to stop shipping prose that overpromises.
- **A test-hardening gap closed.** The whole-vault digest assertion could not distinguish "the pre-flight ran before `scaffold_vault`" from "`scaffold_vault` ran but was a no-op on an already-scaffolded vault". So deleting `deploy_canonical`'s seemingly-redundant pre-flight would have broken *no test* while quietly letting a refused refresh create canonical folders. `test_init_refresh_refusal_precedes_the_folder_scaffold` now pins it with a template tree shipping a folder the vault lacks, and carries a control half so the assertion cannot pass vacuously. Verified: deleting the pre-flight reddens exactly that one test.

One finding accepted and not fixed: `deploy_canonical` computes `detect_drifted_skills` and `deploy_skills` recomputes it, a duplicate glob+hash pass per `init`. Harmless at 14 files, and keeping the primitive self-guarding is the entire point of the design.

## Test plan

**Gate 1 RED, proven by running it — not asserted.** Before any production change, the lead test failed on real behaviour, not a missing symbol:

```
FAILED test_skills_sync_refuses_a_drifted_medium_contract_and_writes_nothing
  AssertionError: Synced 14 skill file(s) to .../vault/00-Creek-Meta/Skills.
  assert 0 == 1
```

11 of the 13 tests written at that point were red (10 behavioural, 1 on the genuinely-absent `DriftedSkillsError`). The other 2 are guard rails against the *fix* — the `*.SKILL.md` no-regression control and the missing-directory tolerance — and cannot be red without changing what they assert. A 14th test came out of self-review.

**Mutation-tested, three ways.** Reintroducing exactly this bug into the fixed code (blinding the detector to the mediums class, every new symbol still present) turns **9 of the new tests red**, plus the new wiring-contract refusal gate:

```
FAILED test_declared_gate_refuses_and_leaves_the_vault_untouched[skills sync-surface14]
  AssertionError: skills sync: refusal expected exit 1, got 0
```

Restoring the fix returns green. The tests that stay green under that mutation are the ones not about the mediums class. Separately, deleting `deploy_canonical`'s pre-flight reddens exactly the new folder-scaffold pin and nothing else.

**Coverage of the behaviour, not just the line:**
- refusal exits 1, names `mediums/essay.MEDIUM.md`, operator bytes intact, **and every other skill file byte-identical** — the reason-printed half alone is passed by a surface that refuses everything
- no regression on the `*.SKILL.md` class (which had **zero** test coverage before this PR — the existing guard was itself untested)
- both classes in one call, asserted as an **exact ordered list** (`lint.SKILL.md`, `mediums/book-report.MEDIUM.md`, `paradox.SKILL.md`) — the assertion a naive two-glob concatenation fails, where a set or a count would not
- operator-authored extras (`mine.SKILL.md`, `mediums/custom.MEDIUM.md`) survive and are never reported as drift — the pin against a `copytree`-style "fix"
- stripped wheel with no `mediums/` template dir; missing `mediums/`; missing `Skills/`
- `init --refresh` refusal is atomic: the **whole-vault digest map** is identical afterwards
- `.bak` sidecar non-interference, three ways
- the `init --refresh` refusal precedes the *folder scaffold*, not merely the skill copy

**Gate 2:** `./scripts/check-all.sh` — literal exit code **0**, 12/12 checks. 9145 passed, 5 skipped (env-gated live smokes). Aggregate coverage 94.46%; `creek/scaffold.py` 95.28% and complexity grade A; per-file gate 259 files ≥80%. Integration + e2e lanes pass (24). Ruff re-verified with `--no-cache`. No `# noqa`, `# type: ignore`, `skip`, or threshold change anywhere in the diff — swept, test files included.

Closes #1306

### DIFF
--- a/creek-tools/CHANGELOG.md
+++ b/creek-tools/CHANGELOG.md
@@ -69,3 +69,17 @@ to locate the originating commit for any reference below.
   `01-Fragments/` … `10-Liminal/` or `00-Creek-Meta/`. User vault
   content lives outside this repository (default suggested location:
   `~/Obsidian/Creek-Vault/`, scaffolded by `creek init`).
+
+### Fixed
+
+- **The drift guard that refuses to overwrite a hand-edited skill file
+  now covers medium contracts too, not just schema skills, and it now
+  lives inside the shared deployment primitive itself** (#1306).
+  `<vault>/00-Creek-Meta/Skills/mediums/*.MEDIUM.md` is refused on
+  divergence exactly like `*.SKILL.md`; because the guard moved into
+  the deploy primitive, `creek init --refresh` is covered by the same
+  check, and a refusal there is atomic — nothing at all is deployed,
+  not just the skill tree. `--force` no longer silently discards the
+  operator's edits: it now preserves the local version alongside as
+  `<name>.bak` (e.g. `mediums/essay.MEDIUM.md.bak`) before
+  overwriting.
--- a/creek-tools/README.md
+++ b/creek-tools/README.md
@@ -118,7 +118,7 @@ Every command is also documented under [`docs/`](docs/) with end-to-end examples
 | Command | Purpose | Doc |
 |---------|---------|-----|
 | `creek skills generate` | Generate the **Voice Skill Tree** under `<vault>/creek-skills/` (one `SKILL.md` per frequency, phase, mode, register, plus thread/eddy/meta skills). Previously `creek skills`; see [`CHANGELOG.md`](CHANGELOG.md) for the FEAT-019 rename. | [generation](docs/generation.md#voice-skill-tree) |
-| `creek skills sync` | Re-deploy the canonical schema-skill tree from `creek-tools/creek/templates/skills/` into `<vault>/00-Creek-Meta/Skills/`. Refuses to overwrite locally-modified skill files unless `--force` is passed. | [generation](docs/generation.md#voice-skill-tree) |
+| `creek skills sync` | Re-deploy the canonical schema-skill tree and medium contracts from `creek-tools/creek/templates/skills/` into `<vault>/00-Creek-Meta/Skills/` (including `mediums/*.MEDIUM.md`). Refuses to overwrite a locally-modified *canonical* file of either class unless `--force` is passed, in which case the local version is preserved alongside as `<name>.bak`. Files you authored yourself are never deployed over or reported as drift. | [generation](docs/generation.md#voice-skill-tree) |
 | `creek mine`     | Mine blog/essay seed ideas from the vault using four discovery strategies (liminal cross-eddy, thread terminus, resonance chain, wavelength-phase window). | [generation](docs/generation.md#mining) |
 | `creek draft`    | Draft an essay from a mined idea using the activated skill stack. Saves to `07-Voice/Drafts/` with full provenance. | [generation](docs/generation.md#drafting) |
 
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -43,7 +43,7 @@ logger = logging.getLogger(__name__)
 
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Sequence
 
     from creek.author.conductor import VoiceClientFactory
     from creek.author.models import AuthoredDraft
@@ -569,21 +569,95 @@ def _guard_vault_path(vault: Path, allow_in_repo: bool) -> None:
     )
 
 
+def _print_skill_drift_refusal(
+    labels: Sequence[str],
+    *,
+    verb: str,
+    remedy: str,
+    whole_deployment: bool,
+) -> None:
+    """Explain a refused canonical-skill deployment and the way past it.
+
+    ``creek skills sync`` and ``creek init`` reach the same destructive
+    primitive, so they owe the operator the same explanation (issue
+    #1306). Each label gets its own line: Rich hard-wraps at the console
+    width, and a path folded across a wrap is both unreadable and
+    unsearchable. The labels are vault-relative on purpose — the bare
+    filename this used to print could not tell
+    ``mediums/essay.MEDIUM.md`` apart from a top-level skill.
+
+    Both the verb and the suggested command are supplied by the caller
+    rather than derived here. Printing a fixed menu of escapes is how a
+    refusal ends up advising ``--refresh`` to someone who never typed
+    it — which for a plain ``creek init`` would also skip seeding
+    ``creek_config.yaml``, so the "fix" would quietly do less than the
+    command it replaced.
+
+    Args:
+        labels: Vault-relative paths of the drifted files, as
+            ``creek.scaffold.DriftedSkillsError.labels`` renders them.
+            Plain strings so this module needs no import of the
+            exception type (``creek.scaffold`` is imported lazily inside
+            the commands that use it).
+        verb: What this invocation was trying to do, for the first line.
+        remedy: The exact command to re-run, already ``--force``-bearing
+            and matching the flags the operator actually passed.
+        whole_deployment: Whether the refusal aborted more than the skill
+            tree. An explicit flag, not a test on *verb*: what the
+            operator is told must not depend on how a word was spelled.
+    """
+    console.print(
+        f"[red]Refusing to {verb}: local changes detected in "
+        f"{len(labels)} skill file(s):[/red]",
+        highlight=False,
+    )
+    for label in labels:
+        console.print(f"[red]  - {escape(label)}[/red]", highlight=False)
+    if whole_deployment:
+        # A drifted skill aborts the *whole* canonical deployment, not
+        # just the skill tree. Say so, or the operator is left guessing
+        # whether the ontology spec and AGENTS.md landed.
+        console.print(
+            "[red]Nothing at all was deployed: no folders, no ontology "
+            "spec, no AGENTS.md.[/red]",
+            highlight=False,
+        )
+    console.print(
+        "[yellow]Revert the edit, or re-run to overwrite it:\n"
+        f"  {escape(remedy)}[/yellow]",
+        highlight=False,
+    )
+    console.print(
+        "[dim]--force keeps your version alongside the canonical file as "
+        "<name>.bak, replacing anything already at that path; a later "
+        "--force overwrites it again.[/dim]",
+        highlight=False,
+    )
+
+
 @app.command()
 def init(
     vault: Path = typer.Option(..., "--vault", help="Vault root to initialise"),
     force: bool = typer.Option(
         False,
         "--force",
-        help="Overwrite an existing creek_config.yaml.",
+        help=(
+            "Overwrite an existing creek_config.yaml, and overwrite "
+            "locally-modified canonical skill files (your version is kept "
+            "alongside as <name>.bak; a later --force overwrites that "
+            "backup)."
+        ),
     ),
     refresh: bool = typer.Option(
         False,
         "--refresh",
         help=(
-            "Re-copy canonical templates (ontology, AGENTS.md, schema "
-            "skills, scaffold) into an existing vault. User data and "
-            "creek_config.yaml are preserved."
+            "Re-copy canonical material (folder scaffold, ontology spec, "
+            "AGENTS.md, schema skills, medium contracts) into an existing "
+            "vault without seeding a fresh creek_config.yaml. AGENTS.md is "
+            "always overwritten; a locally-edited skill or medium contract "
+            "makes the command refuse unless --force. User content "
+            "elsewhere in the vault is untouched."
         ),
     ),
     allow_in_repo: bool = typer.Option(
@@ -609,11 +683,19 @@ def init(
     By default ``creek init`` refuses to scaffold inside a git repo;
     pass ``--allow-in-repo`` to override.
 
-    ``--refresh`` re-copies canonical material into an existing vault
-    without touching user-edited content or ``creek_config.yaml``.
+    Every run — with or without ``--refresh`` — restores the canonical
+    material: the folder scaffold, the ontology spec, ``AGENTS.md``, the
+    schema skills and the medium contracts. ``AGENTS.md`` is *always*
+    overwritten; it is the machine-facing agent contract, not
+    operator-authored content. A locally-edited skill or medium contract
+    makes the command refuse outright — deploying nothing — unless
+    ``--force`` is passed, which keeps the operator's bytes alongside as
+    ``<name>.bak``. ``--refresh`` differs from a plain run only in that
+    it leaves ``creek_config.yaml`` alone instead of seeding a fresh
+    one. User content elsewhere in the vault is never touched.
     """
     from creek.config import generate_default_config
-    from creek.scaffold import deploy_canonical
+    from creek.scaffold import DriftedSkillsError, deploy_canonical
 
     _guard_vault_path(vault, allow_in_repo)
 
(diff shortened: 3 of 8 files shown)


Geoffe-Ga:
## Summary
This PR fixes issue #1306: `deploy_skills` wrote both `*.SKILL.md` and `mediums/*.MEDIUM.md`, but `detect_drifted_skills` only globbed the former, so `creek skills sync` silently destroyed hand-edited medium contracts. The fix unifies both enumerations behind a single `_canonical_skill_pairs()` helper, moves the drift guard into the `deploy_skills` primitive itself (`force=False` default), adds a `--force`-preserving `.bak` backup mechanism, pre-flights the same guard in `deploy_canonical` so a refused `creek init --refresh` deploys nothing at all, and updates CLI help/docs/CHANGELOG to match the new (more honest) behavior.

## Strengths
- Root-causes the bug correctly: two divergent enumerations of "what is a skill file" were the actual defect, and the fix collapses them into one (`_canonical_skill_pairs`), which both `detect_drifted_skills` and `deploy_skills` now read — they structurally cannot disagree again.
- Puts the guard inside the destructive primitive (`deploy_skills`) rather than only in the CLI caller, closing the exact hole that let `creek init --refresh` bypass it entirely via a second, unguarded call site.
- `deploy_canonical`'s drift check is a genuine pre-flight (runs before `scaffold_vault`), verified by a dedicated regression test (`test_init_refresh_refusal_precedes_the_folder_scaffold`) with a real control half, not just a "no exception raised" check.
- `.bak` backups are written before the destructive overwrite (`_preserve_local_edits` runs before the copy loop in `deploy_skills`), so no data-loss window exists between detecting drift and preserving it.
- Backup suffix (`.bak`) is deliberately excluded from both skill globs and the `skill-size` lint sweep, and this is verified by test rather than merely asserted.
- Very thorough test additions covering: both-classes drift, global sort order (not per-glob concatenation), operator-authored extras surviving a forced deploy, stripped-wheel installs with no `mediums/` dir, atomicity of a refused refresh, and the `.bak` sidecar's invisibility to other globs/lint.
- Docs (CHANGELOG, README, `--help` text, docstrings) are updated to state the *actual* new behavior, including retracting an overstated "all-or-nothing" atomicity claim in `deploy_canonical`'s docstring — a good instance of not overpromising.
- `AGENTS.md`'s different treatment (always overwritten, no drift guard) is a deliberate and clearly-justified asymmetry, not an oversight, and the existing pinning test for that behavior is left untouched.
- PR description honestly surfaces the consequence for a related open issue (#1354) — that sync no longer incidentally reverts a tampered medium contract — rather than glossing over it.

## Security Concerns
None. All paths involved are derived from fixed, repo-shipped template filenames or CLI-supplied vault roots, not attacker-controlled input; no injection surface introduced. `escape()` is correctly applied before interpolating any path into Rich markup output.

## Problems
None found. I traced `deploy_skills`, `detect_drifted_skills`, `deploy_canonical`, and both CLI call sites (`init`, `skills sync`) end-to-end and did not find a logic bug: drift is always computed via the same shared enumeration, backups always happen before overwrite, and I confirmed via `grep` that no other call site in the codebase (`creek_mcp/`, other `creek/*` modules) still calls the old un-forced signatures.

## Code Quality
- Non-blocking: the printed remedy commands (e.g. `f"creek init --vault {vault}{refresh_flag} --force"`) aren't shell-quoted, so a vault path containing a space would produce a copy-pasteable-but-broken suggestion. Minor UX nit, not worth blocking on.
- Non-blocking (acknowledged in the PR description itself): `deploy_canonical` and `deploy_skills` each independently call `detect_drifted_skills`, a duplicate glob+hash pass per `init` invocation. Explicitly called out as an accepted tradeoff for keeping `deploy_skills` self-guarding; reasonable at the current file count (14).

## CLAUDE.md Adherence
✅ TDD workflow followed and documented (RED test output shown in PR body, plus a mutation-testing pass reintroducing the original bug to confirm 9 new tests catch it).
✅ No `# noqa`, `# type: ignore`, or threshold changes anywhere in the diff (verified by grep).
✅ Docstrings are thorough and Google-style with Args/Returns/Raises throughout the new/changed public API (`deploy_skills`, `detect_drifted_skills`, `deploy_canonical`, `DriftedSkillsError`, `SkillDeployment`).
✅ Single source of truth respected: `_canonical_skill_pairs` is explicitly built to eliminate the duplicate-enumeration root cause (DRY principle, §1.2).
⚠️ I could not execute `./scripts/check-all.sh` myself in this review sandbox (command execution required approval I wasn't granted), so the ≥90% coverage / mypy strict / ruff-zero-violations claims in the PR body are taken on the author's word rather than independently re-verified here. The commit message, PR body, and Gate 2.5 self-review section are detailed and specific (exact coverage percentages, exact test counts, exact failing-test output), which is consistent with an actual run rather than a fabricated claim, but this is still a review-process caveat worth flagging.

## Testing
Insufficient testing is supposed to be blocking per the instructions, but this PR is the opposite case — testing is unusually thorough and specifically targets the failure modes that let the original bug ship (global sort order asserted as an exact list, not a set; operator-authored extras pinned against a `copytree`-style over-fix; atomicity of the refusal pinned with a control test that would fail if the pre-flight were "helpfully" removed as apparently-redundant). No test-coverage gaps identified.

## Documentation
CHANGELOG, README's command table, `docs/getting-started.md`, `docs/wiring-contract.md`, and both `--help` texts are all updated consistently with the new behavior, including the previously-false "without touching user content" and "refuses to overwrite locally-modified skill files" claims that motivated this fix per the PR description. `docs/wiring-contract.md`'s new `Refusal.mutate` field is documented with a clear rationale (declaring a drift-triggered gate that can't exist in a freshly-scaffolded vault).



## Verdict: LGTM

The fix correctly root-causes the reported bug (two divergent skill-file enumerations) by unifying them into one shared helper, closes the second unguarded call site (`creek init --refresh`) that made the original bug worse, and adds a well-designed `.bak` preservation mechanism — all verified by tracing the actual control flow, not just reading the PR's claims. Test coverage is exceptionally thorough and targets the specific failure modes that let the original bug ship undetected. No blocking bugs, security issues, or missing tests found; the only remarks are two minor non-blocking nits (unquoted remedy paths in CLI output, an acknowledged duplicate drift-detection pass) that don't warrant blocking merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5253281276 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1398, opened 2026-08-11
fix(redact): --apply over a vault rewrites creek_config.yaml, corrupting the config that governs redaction


Geoffe-Ga:
## Signal

`.yaml` is in the default `redaction.supported_extensions`
(`creek-tools/creek/config.py:781-792`) and `exclude_patterns` defaults to
`['.git', 'node_modules']` (`:795-797`), so
`creek redact --apply --source <vault> --vault <vault>` treats
`<vault>/00-Creek-Meta/creek_config.yaml` as an ordinary candidate and rewrites
it in place.

## Why this is worse than losing a config file

`RedactionConfig.false_positive_allowlist` exists to hold strings that *look
like* secrets but are not — which is precisely the set of strings the scanner
will match. A vault-wide apply therefore replaces allowlist entries with
`[REDACTED:...]` markers, and every subsequent scan then flags the very
false positives the operator had already triaged. The config that governs
redaction is silently widened by running redaction.

Custom `patterns`, `replacement_template`, and any tokens in
`exclude_patterns` are exposed the same way.

## Not fixed by #1308

#1308 excludes `00-Creek-Meta/audit/` and the legacy
`00-Creek-Meta/Processing-Log/purge-log.json` from the apply set, because those
are hash-chained compliance artifacts whose in-place mutation is undetectable.
`creek_config.yaml` is a different failure (config corruption, loudly wrong
rather than silently signed) and was left out of that change deliberately
rather than by oversight — see `PROTECTED_AUDIT_RELPATHS` in
`creek-tools/creek/redact/cli_commands.py`.

## Options

1. Add `00-Creek-Meta/creek_config.yaml` to `PROTECTED_AUDIT_RELPATHS` — one
   line, same ratchet direction, but the constant is then no longer only about
   audit artifacts and should be renamed.
2. Exclude `00-Creek-Meta/` from the apply set wholesale and enumerate what
   *should* be redactable inside it. Broader, and needs a look at whether any
   vault content legitimately lives there.
3. Refuse `--apply` when `--source` contains `--vault`'s `00-Creek-Meta`
   without an explicit opt-in flag.

## Acceptance

- A test that seeds `false_positive_allowlist` with a secret-shaped string,
  runs `--apply --source <vault> --vault <vault>`, and asserts
  `creek_config.yaml` is byte-identical.
- Mutation-check it: remove the guard and confirm the test goes red.

Refs #1308

[...]

Issue #1397, opened 2026-08-11
fix(redact): a secret leaked into a filename is now unremediable — audit lines are excluded from --apply


Geoffe-Ga:
## Context

#1308 excludes `<vault>/00-Creek-Meta/audit/` and the legacy
`00-Creek-Meta/Processing-Log/purge-log.json` from the `creek redact --apply`
rewrite set, unconditionally. That closes a real hazard (a run rewriting its own
hash-chained history, with `verify()` still passing because later entries
re-anchor the chain onto the mutated line).

## The residual it creates

`RedactionAuditEntry.source_path` records the path as the pipeline saw it. If a
secret leaked into a *filename* — `/exports/ssn-123-45-6789.md`, which is
exactly the shape of the reproduction in #1308 — that secret is now written into
`redact.jsonl` and can no longer be remediated by `creek redact --apply`,
because the file holding it is protected from rewriting.

Before #1308 it was *technically* remediable, but only by the same mechanism
that silently corrupted the chain, so this is not a regression in any usable
sense — it is a gap that was previously masked by a bug.

## Why it is purge-shaped, not redact-shaped

The audit log is a hash-chained compliance record. You cannot rewrite one line
in place without invalidating every hash after it. Remediating it means either

1. a chain-aware rewrite that re-signs the tail and records that it did so
   (the `creek purge` model — see `PurgeAuditLog._migrate_legacy_if_needed` for
   the existing precedent of replaying a log into a new chain), or
2. a preventive fix: hash or elide the filename component of `source_path`
   when the filename itself matches a redaction pattern, so the secret never
   reaches the log.

Option 2 is cheaper and probably correct, but it changes what `source_path`
means to every existing reader, so it wants its own design pass.

Documented as a known residual in `creek-tools/docs/redaction.md`.

Refs #1308

PR #1399, opened 2026-08-11
fix(redact): record each rewrite as it happens, so a mid-batch abort cannot destroy bytes unaudited

### LINKED ISSUE #1308: fix(redact): --apply rewrites files before writing any audit record, so a mid-batch failure leaves an unaudited destructive change
**Labels:** `de-slop`, `creek-pipeline`, `security`, `priority-high`
**Severity:** High
**Remediation:** fix (port the purge engine's intent/outcome audit pairing) + regression test

## Problem

`creek redact --apply` rewrites files **before** it writes any audit record, and converts a
mid-batch `OSError` into `typer.Exit(1)` — which unwinds past the audit write. The result is
a destructive, irreversible rewrite with **zero** audit lines, which is precisely the
incident you would later need the audit log to investigate.

- `creek/redact/cli_commands.py:590-596` — `run_apply` calls `_apply_redactions(...)`
  (destructive) before `_write_redaction_audit(...)`.
- `creek/redact/cli_commands.py:502-507` — `_apply_redactions` raises `typer.Exit` from `OSError`.
- `creek/redact/audit.py:24-48` — `RedactionAuditEntry` has no `phase` / `operation_id` /
  `status` / `failure_reason`, so the schema cannot even express a partial outcome.

Taxonomy family 0 (correctness on a compliance path) + family 5 (architectural drift).

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Two files with `ssn` findings; the second
`_atomic_write` raises `OSError(ENOSPC)`:

```
run_apply exited with code 1
a.md on disk now: 'ssn [REDACTED:ssn] here\n'   # rewritten
audit log exists: False
audit entries recorded: 0
```

**Signal 2 — structural, and the sibling subsystem already solved it.**
`creek/purge/engine.py:1701-1772` `_run_audited` writes an `intent` line *before* the first
destructive op and an `outcome` line with `status="partial"` + `failure_reason` on
exception — documented there as "the engine's recovery contract" (GAP-002). The redact path
has no equivalent.

**Signal 3 — the claim it breaks.** `creek/redact/cli_commands.py:546-549` and
`docs/redaction.md:158` both assert: *"Every `creek redact --apply` invocation appends one
JSONL entry per touched file."*

Not covered by #1293 / #1288 / #1087 / #946 / #945 / #900 / #901 / #886, which are all about
scan/apply *matching* or symlink handling.

## Scope

Covers audit sequencing and the audit-entry schema for the `creek redact --apply` path.

Explicitly NOT in scope: scan/apply parity for line-anchored patterns (#900), marker
re-redaction (#945), token-boundary snapping (#946), symlink guards (#1293, #1087, #1298).

## Tasks

1. Extend `RedactionAuditEntry` (`creek/redact/audit.py:24-48`) with `phase`
   (`intent` | `outcome`), `operation_id`, `status` (`ok` | `partial` | `failed`) and
   `failure_reason`, mirroring `creek/purge/audit.py`.
2. Write the `intent` line before the first destructive write in `run_apply`
   (`creek/redact/cli_commands.py:590-596`).
3. Wrap the apply body so any exception still writes an `outcome` line with
   `status="partial"` and the failure reason, then re-raises — the shape
   `PurgeEngine._run_audited` already uses.
4. Add a regression test injecting an `OSError` on the second file and asserting the log
   contains an `intent` line plus a `partial` `outcome` line.

## Acceptance Criteria

- [ ] A mid-batch `OSError` during `--apply` leaves an audit log containing an `intent`
      entry and a `partial` `outcome` entry naming the failure.
- [ ] A successful `--apply` still writes one entry per touched file plus the
      intent/outcome pair.
- [ ] The documented claim in `docs/redaction.md:158` is true, or is updated to match.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/redact/audit.py` | Modify |
| `creek-tools/creek/redact/cli_commands.py` | Modify |
| `creek-tools/tests/test_redact_audit.py` | Modify |

Geoffe-Ga:
## Summary

`creek redact --apply` rewrote every file and only then wrote its audit entries, so a mid-batch failure destroyed bytes and left no record of it. Reproduced at the base commit: an `OSError(ENOSPC)` on the second of two files leaves `a.md` rewritten to `contact: [REDACTED:ssn]` and `redact.jsonl` **not even created** — zero entries for the one file whose bytes were actually destroyed.

The fix moves the per-file append **inside** the rewrite loop, next to each file's `_atomic_write`, and brackets the run with an `intent` line (written before the first byte) and an `outcome` line. Same N appends, re-ordered.

### The issue body's own proposed fix does not close its own acceptance criterion

The body proposes adding an intent line and a `partial` outcome line but leaves `_write_redaction_audit` called after `_apply_redactions` returns. I implemented exactly that as mutation **M1** and ran it: the aborted run logs `['intent', 'outcome']` and still records **nothing** about the destroyed file. `docs/redaction.md`'s "one JSONL entry per touched file" would have stayed false for the only file that was touched. The per-file interleave is the load-bearing part.

### Two findings the issue never mentions, both fixed here

**An unwritable audit destination was discovered only after the destruction.** With the vault `chmod 0o500`, the old code rewrote the file and *then* died with an unhandled `PermissionError` and zero audit lines. Writing intent first turns "destroy, then discover you cannot log it" into "fail before touching anything."

**The audit log lives inside the tree redaction rewrites, and was only accidentally spared.** `00-Creek-Meta` is not in `exclude_patterns`; `redact.jsonl` survived only because `.jsonl` is absent from the *user-editable* `supported_extensions`. Add it and a run rewrites its own first log line — and `verify()` still passes, because the entries appended afterwards re-anchor the hash chain onto the mutated line. Silent, undetectable mutation of a tamper-evident log, reachable by editing a config field. The whole `00-Creek-Meta/audit/` directory is now excluded from the apply set unconditionally.

Reviewing that exclusion surfaced a third: `00-Creek-Meta/Processing-Log/purge-log.json` sits *outside* `audit/`, `.json` is in the default extension list, and `PurgeAuditLog._migrate_legacy_if_needed` replays that file into the chained `purge.jsonl`. Redacting it would get mutated records chain-**signed** into the purge trail as authentic. It is protected too, derived from `LEGACY_PURGE_LOG_RELPATH` rather than a restated literal.

## Design notes

**`failure_reason` is the exception type name only, never `str(exc)`.** An `OSError` message embeds the offending path, and filenames here demonstrably carry secrets — the reproduction used `ssn-123-45-6789.md`. Asserted on the **raw JSONL bytes**, not the parsed model, so a leak through any other field is caught too.

**The Gate 2.5 self-review caught that same leak one handler away from the fix** (second commit). The `except OSError` branch still printed `{exc}`, rendering `[Errno 28] No space left on device: '/vault/ssn-123-45-6789.md'` to the console — which is not a private channel: it reaches CI logs, terminal scrollback and shared screens, all of which outlive the run. The privacy test asserted only on the log's bytes while its docstring claimed it caught "a leak through any other field", so a green gate said nothing about it. It now asserts on `result.output` too. The message reports the **errno symbol** rather than dropping the detail: `ENOSPC` is the actionable part and is OS-derived, never path-derived, so an operator still learns the disk is full.

**The outcome write lives inside the exception handlers, not in an outer wrapper.** `_apply_redactions` converts `OSError` into `typer.Exit`, which subclasses `click.exceptions.Exit` → `RuntimeError`. A purge-style wrapper around the function would catch the *converted* exception and record the literal string `"Exit"`. The test asserts `failure_reason == "OSError"` by exact equality for that reason.

**Three-tier audit-write policy**, because an audit append fails for the same reason the rewrite did (ENOSPC on the shared volume) and the right answer differs by phase:

| Phase | Policy | Why |
|---|---|---|
| `intent` | fails **closed** (`typer.Exit`) | nothing destroyed yet, so refusing is free |
| `file` | fails **fast** (`_AuditWriteError`, loop stops) | bytes are gone; continuing destroys more files that also cannot be recorded |
| `outcome` | **best effort**, never raises | runs from inside `except`; must not displace the original error |

**Schema.** One model, not a discriminated union: `read()`'s return type and the `creek/redact/__init__.py` export are unchanged. `source_path` becomes optional so intent/outcome lines can omit it, and a `model_validator` re-imposes the invariant the required field used to carry (`phase=="file"` implies `source_path is not None`). `phase` defaults to `"file"` — note this differs from `creek/purge/audit.py`, which defaults to `"outcome"`, because a legacy *redaction* line genuinely was a per-file record. `append` now uses `exclude_none=True`, matching purge, so per-file lines do not carry `"status": null`.

**Structural forcing function.** `_rewrite_and_record` is the only caller of `_atomic_write`, and an `ast`-based test asserts exactly one call site, inside that function, alongside `record_file`. Reintroducing the batched ordering is a test failure rather than a review catch.

## Two contracts written down, not implied

**What intent-without-outcome means.** Documented in `docs/redaction.md` and both docstrings: every file with a `file` entry *was* rewritten; **at most one** further file may have been rewritten without its record (the one in flight at kill time); nothing outside the intent line's `files` list was touched. A missing outcome line also means the best-effort outcome write may itself have failed — the console warning is the only thing separating that from a real abort, so it is asserted.

**Durability scope.** `AuditLog.append` fsyncs inside the flock window, so the intent line's bytes are durable before the first rewrite. **Not** claimed: directory-level fsync of a freshly created log, or of `_atomic_write`'s `os.replace`. Both residuals fail safe — intent present, rewrite lost.

## Two existing tests changed deliberately

The issue body's "no existing tests break" is unachievable; both assertions below are load-bearing and were updated with the rationale in the diff, not deleted or weakened.

- `test_cli_redact_apply_writes_audit_entry` — `len(entries) == 1` → the ordered `["intent", "file", "outcome"]`. A lone `file` line cannot distinguish "rewritten, run finished" from "rewritten, process died next" — exactly the ambiguity this issue is about. The per-file assertions now index the `file` entry rather than `entries[0]`.
- `test_cli_redact_dry_run_writes_audit_entry` — dry runs emit the full three-phase grammar, all `dry_run=True`, matching purge. A bare `file` line would leave readers with orphan entries belonging to no run and no way to tell one three-file preview from three one-file previews.

`test_cli_redact_no_findings_writes_no_audit_entry` is untouched and green — it is the guard that intent is not written before the no-findings return.

## Test plan

```
./scripts/check-all.sh   # exit 0 — 12/12 gates, 9161 passed, 5 skipped,
                         # 94.46% branch coverage, per-file gate clean,
                         # ruff --no-cache, mypy strict, tryceratops,
                         # xenon --max-absolute B all clean
```

14 new tests in `tests/test_redact_audit.py`. RED proven before any production edit — the headline test failed on its first assertion, `"a.md's bytes were rewritten and no audit line survives the abort"`, a behavioural failure rather than a missing symbol.

**Mutation checks — each reverted after, with the observed failure:**

- **M1** (batch the appends after the loop — the issue body's own fix): abort test RED, `assert ['intent', 'outcome'] == ['intent', 'file', 'outcome']`. The destroyed file has no record. The `ast` guard also fired.
- **M2** (move `record_intent` after the rewrite loop): chmod test RED, `"the file was destroyed before the audit destination was proven"` — `contact: [REDACTED:ssn]` where the original was expected. This also proves the fixture is non-vacuous: it builds a vault with **no** `00-Creek-Meta` directory, because `AuditLog.append`'s `mkdir(exist_ok=True)` short-circuits when `audit/` already exists and the test would then pass identically before and after the fix.
- **M3** (delete the exclusion call): both artifact tests RED — `redact.jsonl`'s first line rewritten, and `purge-log.json`'s `fragment-123-45-6789` → `[REDACTED:ssn]`.
- **M4** (flip `_is_protected`'s fail-closed arm to `return False`): RED — the unresolvable candidate stays in the rewrite set.
- **M5** (restore the leaky `{exc}` interpolation): RED — `"the operator-facing error echoed the secret-bearing path"`.

`_exclude_audit_artifacts` now also fails closed on **both** sides of its comparison: the candidate path and the protected roots themselves. A symlink loop at `00-Creek-Meta/audit` previously raised `RuntimeError` as a raw traceback, with the audit trail excluded from nothing.

Zero SKIPPED in `tests/test_redact_audit.py` locally. The single `skipif` is `os.geteuid() == 0` (root ignores mode bits); CI is non-containerised `ubuntu-latest`, so it runs there.

`run_apply` is still radon B(8) — no complexity regression against the xenon ceiling.

## Follow-ups filed

- **#1397** (P3) — excluding the audit directory means a secret leaked into a *filename*, already recorded in `source_path`, can no longer be remediated by `creek redact --apply`. Rewriting a hash-chained log is purge-shaped work. Documented as a known residual.
- **#1398** (P2) — `--apply` over a vault still rewrites `creek_config.yaml`, which can redact your own `false_positive_allowlist` entries: running redaction silently widens the config that governs redaction. Deliberately out of scope here.

## Process note

The plan came from the pre-built dossier at `.ralph/dossiers/issue-1308.md` — an independent premise-and-design audit that reproduced every claim by execution at this branch's base commit — plus a `chief-architect` verification pass over it against the live code. That pass is what caught the `verify()`-returns-`None` trap in the plan, the three collateral tests beyond the two the dossier named, and the legacy purge-log gap. A `code-review-orchestrator` pass then found the console leak. Every RED proof, mutation result and gate number above is from my own runs, not a specialist's static read.

Closes #1308


Geoffe-Ga:
## Summary
This PR fixes #1308: `creek redact --apply` used to rewrite every candidate file and only afterward batch-write audit entries, so a mid-batch `OSError` (e.g. `ENOSPC`) could destroy file contents while leaving zero audit trail for the destroyed file. The fix introduces a three-phase audit grammar (`intent` → `file` → `outcome`) written by a new `_ApplyAudit` class, interleaves the per-file audit append immediately after each file's `_atomic_write` (enforced structurally by `_rewrite_and_record` plus an AST-based single-call-site test), and additionally protects the vault's own compliance artifacts (`00-Creek-Meta/audit/`, the legacy `purge-log.json`) from ever being rewritten by `--apply`.

## Strengths
- **Root-cause fix, not a patch.** The core defect was ordering — appends batched after the rewrite loop — and the fix directly inverts that ordering rather than adding a band-aid retry/rollback layer.
- **Structural regression guard.** `test_atomic_write_has_exactly_one_call_site_and_it_records` uses `ast` parsing to pin `_atomic_write`'s sole caller to `_rewrite_and_record`, which also calls `record_file` in the same body — reintroducing the batched-append bug becomes a hard test failure, not something that depends on a future reviewer noticing.
- **Correctly distinguishes failure classes.** `_AuditWriteError` (bytes rewritten, audit append failed) is kept structurally distinct from a bare `OSError` (bytes untouched) via a dedicated exception type carrying `__cause__`, and the three-tier write policy (`intent` fails closed, `file` fails fast, `outcome` is best-effort and never raises) is a sound, well-justified design for the different risk profiles at each phase.
- **Self-critical, verifiable PR description.** The body identifies that the issue's own proposed fix (batch append with `partial` status) is insufficient and proves it via a reverted mutation (M1); a second commit fixes a real secret-leak-to-console regression caught by the author's own Gate 2.5 review. I independently verified several of the PR's more surprising factual claims against the referenced source (`AuditLog.append`'s fsync-inside-flock window at `creek/audit/log.py:230`, `resolves_within`'s fail-open semantics, `PurgeAuditLog._migrate_legacy_if_needed` replaying the legacy log into the hash chain) — all checked out.
- **New protections address a real, previously-undetected gap**: the audit log itself sat inside the tree `--apply` walks and was only accidentally excluded via an extension-list coincidence; this PR makes that exclusion explicit and unconditional, and extends it to the legacy purge log to prevent chain-signing mutated records as authentic.

## Security Concerns
🟢 LOW — None blocking. The PR itself fixes a real information-leak (secret-bearing filename echoed to console via `str(OSError)`) as part of its second commit, and adds `_safe_error_detail` to consistently strip path data from operator-facing messages while preserving the actionable errno symbol. The `_is_protected` fail-closed behavior on both the candidate and the protected-root side (guarding against `RuntimeError` from a symlink loop) is a good defensive addition.

## Problems
None found that block merge.

## Code Quality
- `_ApplyAudit` is a reasonably large class (five methods) but each method has a single, clearly-scoped responsibility and the docstrings are unusually precise about invariants (durability scope, disclosure notes, three-tier policy) — appropriate given this is compliance/audit-trail code where precision matters more than brevity.
- Minor: the PR body itself is extremely long (arguably disproportionate to the diff), but that verbosity produced verifiable, checked claims rather than filler, so it's a matter of taste rather than a quality issue with the code.

## CLAUDE.md Adherence
✅ TDD workflow evidenced (RED-proof-first methodology described and mutation-tested, per the PR body's test plan).
✅ Commit messages follow Conventional Commits (`fix(redact): ...`) with detailed rationale bodies.
✅ Coverage/complexity/lint claims in the PR body (94.46% branch coverage, `run_apply` radon B(8), zero ruff violations) are consistent with the repo's stated thresholds in CLAUDE.md, though I was unable to independently execute `./scripts/check-all.sh` in this review environment (sandboxed Bash lacked permission to run `pytest`/`uv`) — I relied on static tracing of the diff and cross-referencing claims against the actual source rather than a live test run.
✅ Uses feature branch (`issue/1308-...`), not a direct commit to `main`.

## Testing
Strong. 14 new tests in `tests/test_redact_audit.py` cover: the core regression (mid-batch abort names the destroyed file), the audit-directory self-protection, the legacy purge-log protection, unwritable-audit-destination fail-closed behavior (with a `_make_bare_vault` fixture specifically constructed to avoid a vacuous test), console/log leak prevention (asserting on raw JSONL bytes AND `result.output`), `KeyboardInterrupt` handling, audit-write-failure mid-batch stop, declined-consent no-op, and a structural AST guard against reintroducing the original ordering bug. Two existing tests were deliberately updated (not weakened) with clear rationale for why the old assertions no longer capture the correct invariant. Not independently re-run in this review due to sandbox restrictions, but the test code itself is well-reasoned and internally consistent with the implementation it verifies.

## Documentation
Excellent — `docs/redaction.md` gets a full new "The audit trail" section documenting the three-phase grammar, exact conclusions a reader may and may not draw, durability scope, and the two known residuals (filename-embedded secrets in protected files, `creek_config.yaml` still being rewritten) with links to their tracking issues (#1397, #1398). `docs/security/threat-model.md` and `docs/cleaning-and-purge.md` are updated consistently with the new behavior. Docstrings throughout `cli_commands.py` and `audit.py` explain the *why* behind non-obvious decisions (e.g., why the outcome write can't live in an outer wrapper, why `_is_protected` can't reuse `resolves_within`) rather than restating the *what*.



## Verdict: LGTM

The fix correctly moves the per-file audit append inside the rewrite loop (verified via the `_rewrite_and_record`/`_atomic_write` single-call-site AST test), the exception hierarchy in `_apply_redactions` correctly separates `_AuditWriteError` (bytes rewritten, unrecorded) from `OSError` (bytes untouched) from `BaseException` (e.g. Ctrl-C), and every non-obvious factual claim in the PR body (AuditLog's fsync-inside-flock, `resolves_within`'s fail-open semantics, legacy purge-log migration replaying into the hash chain) checks out against the actual referenced source. Tests are thorough, include mutation-tested regression coverage, and a genuine console-leak bug was caught and fixed by the PR's own second commit. No blocking issues found in static review.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5254171347 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1401, opened 2026-08-11
fix(prompt): detect_ontology fails soft with no succeeded signal, so a dead provider silently yields an empty draft ontology


Geoffe-Ga:
## Summary

`detect_ontology` (`creek-tools/creek/classify/prompt.py:308`) has the exact defect PR #1358 just fixed on the fragment-classification path: it fails soft to a bare `PromptOntology(prompt=prompt)` at three separate sites and gives the caller **no way to distinguish "the model detected no ontology" from "the call never happened"**.

Found while fixing #1309 (which widened the weighted YAML schema shared by both paths). Deliberately not folded into that PR — it is an independent defect on an independent entry point.

## Evidence

Three failure returns, all identical and all indistinguishable from a genuine empty verdict:

- `creek/classify/prompt.py:330` — provider unavailable
- `creek/classify/prompt.py:340` — transport / invoke failure
- `creek/classify/prompt.py:358` — parse failure

Compare the fragment path, which PR #1358 fixed by introducing a frozen `WeightedClassificationResult(classification, succeeded)` wrapper (`creek/classify/weighted.py`), mirroring the `LLMClassificationResult` the single-pick path already used (#744).

`detect_ontology` shares `parse_weighted_yaml` and `_ALLOWED_TOP_LEVEL_KEYS` with the fragment path — the same parser, the same injection boundary, but without the success signal.

## Impact

`detect_ontology` feeds draft generation:

- `creek/cli.py:4817` wraps it as the detector closure
- `creek/generate/drafts.py:1913` — `ontology = detect_ontology(section.seed_text)`
- `creek/generate/outline.py:75`

A provider outage therefore yields an empty ontology per section, and `creek draft` proceeds to select and arrange fragments against a profile that detects nothing — silently, with no error surfaced to the operator. The draft is produced, looks complete, and is built on no ontological signal at all.

Lower severity than #1358's fragment-path twin: nothing is persisted to fragment frontmatter, so there is no on-disk corruption and no lying `classification_method` stamp. The damage is a silently degraded draft, recoverable by re-running once the provider is back.

## Suggested fix

Mirror #1358 exactly:

- Return a frozen `PromptOntologyResult(ontology, succeeded)` (or add `succeeded` to `PromptOntology`, if no caller round-trips it to disk).
- Have `drafts.py` surface a real error, or at minimum a warning the operator sees, rather than proceeding on an empty profile.

## Acceptance criteria

[...]

Issue #1400, opened 2026-08-11
fix(classify): RuleClassifier rebuilds voice and wavelength from scratch, destroying persisted evidence


Geoffe-Ga:
## Summary

`RuleClassifier._build_updates` (`creek-tools/creek/classify/rules.py:785-795`) rebuilds `WavelengthClassification` and `VoiceClassification` **from scratch** rather than merging, so a partial keyword match silently destroys persisted classification evidence that the rule pass has no opinion about.

This is the rules-layer twin of #1309 (weighted path). It was found while fixing #1309 and deliberately **not** folded into that PR — its blast radius across the rules test suite is much larger. #1309 works around it by using a rule-inert test body plus an explicit precondition assertion, and discloses the residual.

## The two destruction sites

```python
if phase != Phase.UNCLASSIFIED or mode != Mode.UNCLASSIFIED:
    updates["wavelength"] = WavelengthClassification(
        phase=phase,
        mode=mode,
    )                       # drops orientation, dosage, color, descriptor

if voice_register is not None or confidence is not None:
    updates["voice"] = VoiceClassification(
        voice_register=voice_register,
        confidence=confidence,
    )                       # OR guard: one match wipes the other field
```

Both constructors are unconditional replacements, and both are guarded by `OR`. So matching **one** field of the pair blanks the other, and matching phase-or-mode blanks all four of the wavelength fields the rule pass never inspects.

## Verified reproducer (at `ccb88ee`)

A fragment carrying `voice={voice_register: analytical, confidence: conviction}` plus the body:

> "I have to admit something. I confess I was afraid, and honestly I still am."

`RuleClassifier().classify(fragment, content=body)` alone returns:

```
voice: voice_register='confessional' confidence=None
```

The `conviction` is gone. `_match_voice_register` fired on "I confess"; `_match_confidence` did not; the `OR` guard rebuilt the whole object and dropped the field it had nothing to say about.

(In this particular case wavelength survived, because neither phase nor mode matched. A body that trips the phase matcher destroys `descriptor`, `color`, `orientation` and `dosage` the same way.)

## Why it matters — this is a fail-open privacy hole, same class as #1309

`PrivacyClassifier.classify_tier`'s third INTIMATE trigger is `voice_register == CONFESSIONAL and confidence == CONVICTION`.

[...]

PR #1402, opened 2026-08-11
fix(classify): give the weighted path the confidence axis, and stop it erasing the evidence

### LINKED ISSUE #1309: fix(classify): weighted_classification voids the INTIMATE escalation and nulls a persisted voice.confidence
**Severity:** High
**Remediation:** fix (carry `Confidence` through the weighted profile) + parity test

## Problem

With `weighted_classification: true`, the INTIMATE privacy escalation is structurally dead
and a `voice.confidence` value already on disk is overwritten with `null`.

`WeightedFragmentClassification.to_legacy()` hard-codes `confidence=None`
(`creek/classify/weighted.py:417-420`), documenting the drop as a harmless round-trip
limitation (`weighted.py:352-354`). It is not harmless: `voice.confidence == CONVICTION` is
one of exactly two inputs to `PrivacyClassifier`'s third INTIMATE trigger
(`creek/classify/privacy.py:204-214`, `_is_high_confidence_confessional`). And
`_classify_one_weighted` (`creek/classify/classify_engine.py:1569-1577`) assigns the derived
`voice` object **wholesale** onto the fragment, so a prior `conviction` is destroyed rather
than merely ignored.

Taxonomy family 0 (correctness) + family 12 (confident wrongness).

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Same model verdict, two classification paths:

```
single-pick  (voice_register=confessional, confidence=conviction) -> classify_tier: intimate ; reassess: intimate
weighted     (voice_registers=[confessional@0.95])                -> classify_tier: open     ; reassess: open
```

and the destructive half:

```
before: voice_register='analytical' confidence='conviction'
after : voice_register='analytical' confidence=None
persisted frontmatter voice: {'voice_register': 'analytical', 'confidence': None}
```

Recipe: build a self-authored non-journal `Fragment` with no recovery keyword; call
`_classify_one_weighted(frag, body, LLMConfig())` with
`creek.classify.classify_engine.classify_weighted` stubbed to return
`WeightedFragmentClassification(voice_registers=(WeightedDimension(VoiceRegister.CONFESSIONAL, 0.95),), overall_confidence=0.95)`;
then `reassess(result, body, baseline=PrivacyTier.OPEN, classifier=PrivacyClassifier())`.
Compare against the same fragment carrying `VoiceClassification(CONFESSIONAL, CONVICTION)`.

**Signal 2 — the invariant it breaks is written down.**
`creek/classify/classify_engine.py:157-172` ("Known residual") rests the whole privacy story
on this trigger firing on the *next* run: *"The escalation still prevents every future
egress for that fragment."* On the weighted path there is no next run — the evidence the
next run would read has been nulled.

Side effect: every weighted-classified fragment permanently satisfies
`ReviewQueueGenerator.needs_review`'s `voice.confidence is None` branch
(`creek/classify/review.py:71`).

`weighted_classification` defaults to `False` (`creek/config.py:678`), so this is an
advertised opt-in mode rather than the default path — hence High, not Critical.

## Scope

Covers `Confidence` propagation through the weighted classification path and the privacy
reassessment that depends on it.

Explicitly NOT in scope: `emotional_texture` on the weighted path (#1208 — same root-cause
class, different field), the pre-#974 mis-stamping remediation (#1106), or the
process↔classify parity detector (#1107).

## Tasks

1. Carry `Confidence` through `WeightedFragmentClassification` — either add a `confidences`
   weighted dimension, or have `to_legacy()` accept and preserve an incoming
   `voice.confidence` when the weighted profile has none
   (`creek/classify/weighted.py:417-420`).
2. Make `_classify_one_weighted` (`creek/classify/classify_engine.py:1569-1577`) merge into
   the existing `voice` object rather than replacing it, so a persisted `conviction` is
   never nulled.
3. Correct the "harmless round-trip limitation" note at `weighted.py:352-354`.
4. Add a parity test asserting the weighted and single-pick paths agree on `classify_tier`
   for a confessional/conviction fragment.

## Acceptance Criteria

- [ ] A confessional, high-confidence fragment classifies to `intimate` on **both** the
      weighted and single-pick paths.
- [ ] Running the weighted path over a fragment whose frontmatter already carries
      `confidence: conviction` leaves that value intact on disk.
- [ ] A parity test covers both and fails before the fix.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/classify/weighted.py` | Modify |
| `creek-tools/creek/classify/classify_engine.py` | Modify |
| `creek-tools/tests/test_weighted.py` | Modify (add the parity test) |

Geoffe-Ga:
## Summary

`PrivacyClassifier.classify_tier`'s third INTIMATE trigger is `voice_register == CONFESSIONAL and confidence == CONVICTION`. On the weighted classification path it could never fire, and the run destroyed the evidence a later run would have needed.

**This is a missed escalation — fail-open — not a lowered tier.** I traced every writer of `privacy_tier`: `escalate` takes the max, `apply_tier` assigns outright only when `needs_tier`, `reassess` double-guards, and `vault/writer.py` mirrors it. Nothing lowers a stored tier, even under `--force`. Fragments that should have been INTIMATE simply stayed OPEN — and because the nulling recurred, they routed to the cloud on **every** run rather than being protected after the first. That falsifies the "Known residual" note in `classify_engine.py` promising the escalation "still prevents every *future* egress"; the note is corrected here.

The practical consequence for testing: **an assertion that the tier "did not go down" passes at HEAD and proves nothing.** Every test below asserts the tier goes **UP**, at parity with the single-pick path.

### Three independent causes, all closed

1. **No axis.** The weighted prompt never asked for an author stance, and `_ALLOWED_TOP_LEVEL_KEYS` would have rejected one had the model volunteered it.
2. **`to_legacy()` hard-coded `confidence=None`**, documented as a benign limitation. It was not benign — it was the trigger's second operand.
3. **`_classify_one_weighted` assigned the derived fields wholesale**, overwriting a persisted `conviction` with `None`, `wavelength.descriptor` with `""`, and `mode` with `unclassified`, on every successful run.

The issue body proposes only cause 2's fix (preserve an incoming confidence). That cannot satisfy its own AC #1: a fragment never single-pick-classified has no confidence to preserve, so the weighted path still yields `None` and still misses the escalation. Both halves are required — the real axis for fresh fragments, the merge for persisted ones.

### `merge_onto`

`to_legacy` stays a pure collapse; a new `merge_onto` does the overlay. It rests on one invariant, stated in the code: **every legacy classification field's default IS its "not determined" sentinel** (`UNCLASSIFIED` / `""` / `None`), so `model_dump(exclude_defaults=True)` is exactly the subset the model spoke to. A test pins the delta dict so a future field with a non-sentinel default breaks the build loudly rather than silently resuming the overwrite.

Post-#1358 every profile reaching `merge_onto` is a genuine verdict — all four soft-failure modes return upstream — so a silent dimension means "detected nothing", never "the call died". That is what licenses the rule, and it is recorded as a precondition on the method.

## The privacy judgement, stated plainly for review

Emitting `conviction` from the weighted prompt hands the model a trigger under a ratchet where nothing ever lowers a stored tier, so a manufactured escalation is **permanent**. That deserves scrutiny rather than a silent merge. The reasoning:

- This reaches **parity** with the single-pick path, which has shipped the identical trigger from the same model for its whole life. It removes an inconsistency; it does not create a new class of burial.
- **`overall_confidence` is never mapped to `Confidence`.** They are different quantities — the model's self-rating of its own work versus ontology axis 9, the author's stance. Conflating them would manufacture escalations at scale. The prompt now says so explicitly, and a test pins that `overall_confidence: 0.99` with no `confidences:` section never yields `conviction`.
- **No weight floor, deliberately.** The single-pick path has none, so a floor would make the weighted path *less* likely to surface confessional+conviction than the legacy one — the forbidden direction for a privacy control, and it would break the parity criterion.
- **Guard tests** pin that `musing` and `exploring` do not reach INTIMATE, in memory and on disk. Under a one-way ratchet, over-burying is as much a defect as under-burying.

### RED, proven before the fix

Both reds are wrong values on real objects, not import errors.

**Parity (primary).** Same model verdict, two code paths:

```
assert single_tier == PrivacyTier.INTIMATE      # PASSED at HEAD
>   assert weighted_tier == PrivacyTier.INTIMATE
E   AssertionError: assert <PrivacyTier.UNCLASSIFIED: 'unclassified'> == <PrivacyTier.INTIMATE: 'intimate'>
```

The single-pick arm passes at HEAD while the weighted arm does not — one verdict, two tiers. At HEAD this goes red for **two** independent reasons, and the log shows the first: `Weighted fragment classification failed (unexpected top-level keys ... ['confidences']); returning empty result`. The schema rejects the axis; and even a profile carrying it would be flattened by the hard-coded `confidence=None`.

**Isolating red (merge half alone, HEAD-valid payload, no schema change involved):**

```
assert updated.voice.confidence == Confidence.CONVICTION   # HEAD: None
assert updated.wavelength.descriptor == "Social Anxiety"   # HEAD: ""
assert updated.wavelength.mode == Mode.EXPRESS             # HEAD: 'unclassified'
```

**On disk**, reverting the three source guards turns 5 of 6 tests in the new file red, including `assert None == 'conviction'` (proving `confidence: null` really lands in frontmatter — the write is `model_dump(mode="json")` with no `exclude_none`) and `assert 'open' == 'intimate'` (the escalation genuinely absent).

### Mutation testing

Each guard reverted alone, confirming exactly the intended tests go red:

| Mutation | Result |
|---|---|
| `to_legacy` re-hardcodes `confidence=None` | 7 red |
| `merge_onto` replaces voice wholesale | 4 red |
| `merge_onto` replaces wavelength wholesale | 3 red |
| allow-list drops `confidences` (SEC-004) | 11 red |
| frequency asymmetry guard removed | 2 red |
| holonic JSD drops the dimension | 1 red |

The last one is why it is in the table: **on the first pass it turned nothing red.** `combine` and `decompose_prior` still threaded the axis, so every other test stayed green while children disagreeing maximally on author stance contributed zero divergence, silently overstating the parent's `overall_confidence`. `test_opposed_confidences_dampen_parent_confidence` was added to bind it.

### Wavelength — not in the issue body

The issue names only `voice` as destroyed. The same `model_copy` also blanked `wavelength.descriptor` and overwrote phase/mode; verified empirically. Fixing only voice would have left an identical defect one line above. It also matters downstream: `generate/voice.py::_is_fully_classified` requires `wavelength.mode != UNCLASSIFIED`, so the voice-exemplar corpus test passes only because the wavelength half was fixed too.

### One existing test inverted, deliberately

`test_widen_drops_voice_confidence` asserted `voice_out.confidence is None` — it pinned the defect in place. It is **inverted and renamed** to `test_widen_round_trips_voice_confidence`, not deleted and not left passing. **The issue's "no existing tests break" criterion is therefore unachievable as literally written**, and that is the correct outcome: a test was holding a fail-open privacy hole open. `test_empty_collapses_to_unclassified`, `test_widen_drops_descriptor` and `test_widen_drops_pre_existing_color` all still pass **verbatim**, because `to_legacy` remains a pure collapse.

### SEC-004 and a new hazard #1358 created

The allow-list is an injection boundary. Exactly one key added, pinned by equality (not membership) at 8 keys; the singular `confidence:` is still rejected, and so is `privacy_tier` riding alongside a legal `confidences:` section.

Beyond that, a **structural** drift guard: `test_every_key_the_prompt_requests_is_on_the_allow_list` extracts the top-level keys from the built prompt's YAML skeleton and asserts they are a subset of the allow-list. This exists because #1358 changed the consequence of prompt/validator drift from loud to silent — before it, a mismatch blanked every field on every fragment (catastrophic but obvious); after it, the call fails soft, the rule verdict is preserved, an honest `rules` stamp is written, and the run reports zero errors while the weighted feature is dead on every fragment. A hand-maintained key list cannot catch that, because whoever adds the key updates the list they can see.

### Gate 2

`./scripts/check-all.sh` exit code **0**, 12/12 checks. 9196 passed, 5 skipped, 94.46% branch coverage, per-file TEST-002 gate green, mypy strict clean, ruff verified with `--no-cache` (#1119). The 5 skips are all pre-existing environment skips (Ollama unreachable, live API keys absent) — **zero skipped** across every privacy-related suite (`test_weighted`, `test_holonic`, `test_privacy`, `test_privacy_pass`, `test_prompt_ontology`, the bubble-up pair, and #1358's provenance suite): 281 passed, 0 skipped.

No `# noqa`, `# type: ignore`, `# pylint: disable`, `@pytest.mark.skip` or `--no-verify` introduced; `pyproject.toml` and `scripts/` untouched. #1358's `tests/test_classify_failed_weighted_provenance.py` is untouched, and the only two assertions removed anywhere are the deliberate inversion above and a `_widen_voice_register` → `_widen_optional` rename that gained coverage.

## Residual — please read before merging

**The on-disk acceptance criterion holds for rule-inert bodies only.** `RuleClassifier._build_updates` rebuilds `VoiceClassification` from scratch under an `OR` guard, so for any body its voice matcher fires on, a persisted `confidence` is destroyed *before* the weighted classifier runs and no downstream merge can recover it. Verified: a fragment carrying `{analytical, conviction}` plus "I confess I was afraid" comes back from `RuleClassifier.classify` alone as `{confessional, None}`.

This PR closes the weighted-path defect completely for the fragments the rule pass leaves alone, and makes the boundary **visible rather than implied** — every affected test asserts the precondition explicitly. That guard earned its place during development: an earlier draft titled a fixture "Evidence bearing fragment", the word "evidence" tripped the rules voice matcher, and the precondition assertion caught it rather than letting the test fail for an unrelated reason.

Filed rather than folded in:
- **#1400** (P1) — the rules-layer twin above.
- **#1401** (P2) — `detect_ontology` fails soft with no `succeeded` signal, the same defect #1358 fixed on the fragment path, so a dead provider silently yields an empty draft ontology.

Closes #1309

### DIFF
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -170,6 +170,15 @@ voice-proxy generation, the Writing Desk), which is a strict improvement
 on the prior behaviour where 100% of intimate content routed to the
 cloud, every time. Fixing the residual properly needs a local
 pre-classification pass, which is out of scope here.
+
+That "once, not every time" promise was true only of the single-pick
+path until #1309. The weighted path could not keep it: it had no
+author-stance axis to detect ``conviction`` with, and it overwrote
+``voice.confidence`` with ``None`` on every run — so the escalation
+never fired, and even a tier escalated by some other route lost the
+evidence explaining it. Both paths now reach the same tier for the
+same model verdict, and both leave the evidence in place across
+re-runs, so the residual above is genuinely one egress on either.
 """
 
 from __future__ import annotations
@@ -1489,10 +1498,11 @@ def _classify_one(
             dispatches through
             :func:`creek.classify.weighted.classify_weighted`, persists
             the resulting weighted profile to
-            :attr:`Fragment.weighted`, and derives the legacy
-            single-pick fields via
-            :meth:`WeightedFragmentClassification.to_legacy` so
-            downstream consumers stay synchronised. Has no effect on
+            :attr:`Fragment.weighted`, and merges the derived legacy
+            single-pick fields over the fragment via
+            :meth:`WeightedFragmentClassification.merge_onto` — a
+            dimension the profile is silent about keeps its prior
+            value rather than being reset. Has no effect on
             the rules path or on ``--method llm`` runs where the rule
             classifier already cleared the confidence floor; for those
             fragments :attr:`Fragment.weighted` stays ``None``.
@@ -1540,10 +1550,16 @@ def _classify_one_weighted(
     :attr:`ClassificationConfig.weighted_classification` is set.
     Populates :attr:`Fragment.weighted` with the full weighted
     profile and derives the legacy single-pick fields from the
-    profile's top entries via
-    :meth:`WeightedFragmentClassification.to_legacy` so existing
-    consumers (lint, compile, voice-skill generation) stay
-    synchronised. When the underlying call fails soft to an empty
+    profile's top entries, so existing consumers (lint, compile,
+    voice-skill generation) keep reading a canonical pick.
+
+    Those derived fields **merge** over the fragment's existing
+    classification rather than replacing it wholesale — a dimension
+    the profile is silent about leaves the prior value standing. The
+    merge rule, and why the wholesale form was a privacy defect, live
+    on :meth:`WeightedFragmentClassification.merge_onto` (#1309).
+
+    When the underlying call fails soft to an empty
     profile (whitespace-only body, provider unavailable, transport
     error, malformed YAML) nothing is derived from it: the input
     fragment is handed back untouched — rule verdict intact,
@@ -1581,15 +1597,7 @@ def _classify_one_weighted(
         # ``_record_if_preserved`` forever (#744, #1330).
         return fragment, True, ""
     weighted = result.classification
-    freq, wave, voice = weighted.to_legacy()
-    updated = fragment.model_copy(
-        update={
-            "weighted": weighted,
-            "frequency": freq,
-            "wavelength": wave,
-            "voice": voice,
-        },
-    )
+    updated = weighted.merge_onto(fragment)
     return updated, False, weighted.reasoning
 
 
--- a/creek-tools/creek/classify/holonic.py
+++ b/creek-tools/creek/classify/holonic.py
@@ -72,6 +72,7 @@ from creek.classify.weighted import (
     WeightedFragmentClassification,
 )
 from creek.models import (
+    Confidence,
     Dosage,
     Frequency,
     Mode,
@@ -189,6 +190,11 @@ def combine(
             effective_weights,
             top_k=top_k,
         ),
+        confidences=_combine_dimension(
+            [c.confidences for c in children],
+            effective_weights,
+            top_k=top_k,
+        ),
         overall_confidence=confidence,
         reasoning=reasoning,
     )
@@ -235,6 +241,7 @@ def decompose_prior(
         orientations=parent.orientations,
         dosages=parent.dosages,
         voice_registers=parent.voice_registers,
+        confidences=parent.confidences,
         overall_confidence=parent.overall_confidence * 0.5,
         reasoning=parent.reasoning,
     )
@@ -387,6 +394,12 @@ def _mean_normalised_jsd(
         ("orientations", Orientation),
         ("dosages", Dosage),
         ("voice_registers", VoiceRegister),
+        # Author stance counts toward the divergence penalty like any
+        # other dimension. Omitting it would let children that disagree
+        # maximally about how firmly the writer holds their claim
+        # contribute zero divergence, silently overstating the bubbled-up
+        # parent's ``overall_confidence`` (#1309).
+        ("confidences", Confidence),
     )
     divergences: list[float] = []
     for attr, enum_type in dimension_accessors:
--- a/creek-tools/creek/classify/prompt.py
+++ b/creek-tools/creek/classify/prompt.py
@@ -291,6 +291,17 @@ def parse_prompt_ontology_response(
     # Parsing lives on :func:`creek.classify.weighted.parse_weighted_yaml`
     # so the prompt-level and fragment-level paths share one source of
     # truth; this wrapper only adds the ``prompt`` echo onto the result.
+    #
+    # One dimension is intentionally not copied across. #1309 gave the
+    # fragment-level profile a ``confidences`` axis — the *author's*
+    # stance toward their own claim — and :class:`PromptOntology`
+    # deliberately does not gain it: an operator-supplied essay seed has
+    # no author whose stance could be detected, so the axis is undefined
+    # rather than merely unmeasured, and inventing a value for it would
+    # feed a privacy control (the INTIMATE escalation) a fabricated
+    # reading. The shared parser tolerating the extra key costs nothing
+    # here precisely because this field copy is hand-written: an axis
+    # this path has no meaning for is simply left behind.
     parsed = parse_weighted_yaml(response)
     return PromptOntology(
         prompt=prompt,
--- a/creek-tools/creek/classify/weighted.py
+++ b/creek-tools/creek/classify/weighted.py
@@ -19,7 +19,10 @@ The legacy single-pick fields on :class:`creek.models.Fragment`
 downstream consumers (compile, lint, voice-skill generation). When
 ``weighted`` is populated, the two stay synchronised via the
 :meth:`WeightedFragmentClassification.from_single_pick` and
-:meth:`WeightedFragmentClassification.to_legacy` adapters.
+:meth:`WeightedFragmentClassification.to_legacy` adapters — or, on a
+real classification run,
+:meth:`WeightedFragmentClassification.merge_onto`, which layers the
+collapse over the fragment instead of replacing it.
 """
 
 from __future__ import annotations
@@ -52,6 +55,7 @@ from creek.classify.llm.prompts import (
 )
 from creek.models import (
     Color,
+    Confidence,
     Dosage,
     Frequency,
     FrequencyClassification,
@@ -110,6 +114,14 @@ class WeightedDimension(Generic[_DimT]):
 # classifier so the schema is enforced from one place. ``reasoning``
 # is populated outside the YAML path (via :func:`_split_reasoning_and_yaml`)
 # and therefore is not a permitted top-level YAML key here.
+#
+# This set is the SEC-004 injection boundary: it is what stops a model —
+# or a prompt injection riding in on fragment text — from writing an
+# arbitrary :class:`~creek.models.Fragment` field, ``privacy_tier`` above
+# all, straight out of an LLM response. Widening it is always a deliberate
+# act; #1309 added exactly one key, ``confidences``, and like every other
+# member it is named after the *dimension* rather than the Fragment field
+# it eventually feeds (``voice.confidence``).
 _ALLOWED_TOP_LEVEL_KEYS: frozenset[str] = frozenset(
     {
         "frequencies",
@@ -118,6 +130,7 @@ _ALLOWED_TOP_LEVEL_KEYS: frozenset[str] = frozenset(
         "orientations",
         "dosages",
         "voice_registers",
+        "confidences",
         "overall_confidence",
     },
 )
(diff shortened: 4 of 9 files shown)


Geoffe-Ga:
## Summary
Fixes a real fail-open privacy bug: the weighted classification path never asked the model for author "stance" (`Confidence`), and `to_legacy()` hard-coded `confidence=None`, so the `confessional + conviction → INTIMATE` escalation could never fire on that path — and the wholesale-assignment in `_classify_one_weighted` also destroyed persisted `voice.confidence` and `wavelength.descriptor`/`mode` on every run, so a fragment that *should* stay protected re-exposed itself to the cloud on every re-run instead of escalating once. The PR adds the `confidences` axis end-to-end (prompt, schema allow-list, `WeightedDimension`, holonic combine/decompose/JSD), and replaces the wholesale-overwrite with a new `merge_onto` that layers only the dimensions the model actually spoke to (via an `exclude_defaults` invariant) over the fragment's prior classification.

## Strengths
- Root-caused into three independent, correctly-diagnosed defects (missing axis, lossy `to_legacy`, wholesale overwrite) and fixed all three rather than patching the one named in the linked issue.
- `merge_onto`'s `exclude_defaults=True` trick is elegant and its precondition (every legacy field's default *is* its "not determined" sentinel) is pinned by an explicit test (`test_exclude_defaults_delta_is_exactly_the_determined_subset`) so a future non-sentinel default fails loudly instead of silently reintroducing the bug.
- Correctly threads the new axis through `holonic.combine`/`decompose_prior`/`_mean_normalised_jsd` — I verified `dimension_accessors` in `holonic.py` now includes `("confidences", Confidence)`, and the PR includes a regression test proving that specific line matters (`test_opposed_confidences_dampen_parent_confidence`).
- The prompt explicitly disambiguates `overall_confidence` (model's self-rating) from `confidences` (author's stance) and this is tested directly — a real and easy-to-miss privacy hazard (conflating the two would manufacture INTIMATE escalations at scale under a one-way ratchet) that the PR calls out and guards against.
- SEC-004 allow-list widened by exactly one key, pinned by equality rather than membership, with tests for the singular `confidence:` field-name collision and a `confidences:` + `privacy_tier:` smuggling attempt — both still correctly rejected. I confirmed `_ALLOWED_TOP_LEVEL_KEYS` in `weighted.py` matches.
- Honest about a residual: `RuleClassifier._build_updates`'s OR-guard (rules.py:791, confirmed by reading) still destroys `voice.confidence` before the weighted path runs for any body the rules voice matcher fires on. This is disclosed, tests assert the precondition explicitly rather than assuming it, and it's filed as a separate follow-up (#1400) rather than silently left uncovered.
- Deliberately inverts (not deletes) `test_widen_drops_voice_confidence` → `test_widen_round_trips_voice_confidence`, since the old test was pinning the defect in place — the right call, clearly explained.
- Backward-compatibility tests confirm pre-existing `weighted:` frontmatter blocks without `confidences` still validate and round-trip.
- I independently traced `_classify_one_weighted` in `classify_engine.py` and confirmed the early-return soft-failure paths all precede the `merge_onto` call site, which is what licenses `merge_onto`'s documented precondition ("only ever reached with a genuine profile").

## Security Concerns
🟢 LOW — None found. The SEC-004 injection boundary is correctly widened and tested both for the new field and for smuggling attempts. Directionally, the only way the new axis affects privacy is by making the escalate-only tier *more* likely to correctly fire (never lowering a tier), which is the intended and safe direction for a privacy control, and is specifically tested (musing/exploring must not reach INTIMATE).

## Problems
None found blocking merge. I read the full diff and cross-checked the non-trivial claims (privacy trigger condition, `_ALLOWED_TOP_LEVEL_KEYS`, model field defaults underpinning the `exclude_defaults` merge invariant, the rules.py OR-guard residual, and that no other `WeightedFragmentClassification`-shaped constructors were missed) directly against the checked-out source, and everything the PR body claims matches the actual code.

## Code Quality
- Non-blocking: `merge_onto`'s frequency-vs-wavelength/voice asymmetry (frequency replaced wholesale, others merged field-wise) is well-justified in a comment (secondaries are a list, can't merge field-wise) but is a subtle enough asymmetry that a future reader modifying this method should re-read that comment carefully; consider whether a short note in the class docstring (not just the method) would help future maintainers who only skim the class-level docs.
- Non-blocking: `OntologyAnalysis` in `author/models.py` has its own independent `voice_registers` field with no `confidences` twin — correctly out of scope here (it's a corpus-level author-voice analysis, not privacy-relevant), but worth a one-line note if a future reader wonders why it wasn't touched.

## CLAUDE.md Adherence
✅ Commit message uses Conventional Commits format (`fix(classify): …`).
✅ No `# noqa`, `# type: ignore`, `# pylint: disable`, or `@pytest.mark.skip` introduced (verified via grep over the full diff).
✅ TDD workflow followed per PR description (RED tests shown failing at HEAD, mutation-testing table included).
✅ PR body reports `./scripts/check-all.sh` exit 0, 12/12 checks, 94.46% branch coverage — I could not independently re-run CI/local checks in this sandboxed review environment (tooling restricted `gh pr checks` and shell `cd`/pytest invocations), so this is taken on the PR author's word but is consistent with the code I did read.
✅ No `pyproject.toml`/`scripts/` changes, per the PR body.

## Testing
Excellent and unusually rigorous for this codebase's already-high bar:
- In-memory unit tests for the merge logic, holonic threading, prompt/schema drift guard, and privacy-tier parity between the weighted and single-pick paths.
- A new on-disk end-to-end test file (`test_classify_weighted_evidence_on_disk.py`) proving the fix through the real frontmatter-write path, including idempotency across re-runs and backward-compat with pre-fix frontmatter.
- A structural (not hand-maintained) test verifying the prompt's YAML skeleton keys are always a subset of the schema allow-list, specifically motivated by the fail-soft consequence #1358 introduced.
- Deliberate, well-labeled inversion of a test that pinned the original defect.
No insufficient-testing concerns.

## Documentation
Docstrings on `merge_onto`, `to_legacy`, `from_single_pick`, and the module-level docstring in `classify_engine.py` are updated and internally consistent with the new behavior (verified by reading the actual post-merge source, not just the diff). The privacy-relevant reasoning (why no weight floor, why `overall_confidence` must never be conflated with `confidences`) is documented in-line at the exact place a future maintainer would need it, not just in the PR description.



## Verdict: LGTM

This is a well-scoped, thoroughly tested fix for a genuine fail-open privacy defect; I independently verified the core claims (privacy trigger condition, SEC-004 allow-list, the exclude_defaults invariant's underlying model defaults, and the disclosed rules.py residual) against the actual checked-out source and found no discrepancies or bugs. The only items I'd raise are non-blocking clarity nits, not correctness concerns, so this is ready to merge as-is.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5254922981 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 12/13 Green
**VERDICT**: LGTM
**Action**: pull comment 5254922981 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1406, opened 2026-08-11
chore(lint): audit the 40 blind `except Exception` sites, then enable ruff BLE


Geoffe-Ga:
## Problem

`except Exception:` is invisible to this repo's lint gate. `creek-tools/pyproject.toml` `[tool.ruff.lint]` has:

```toml
select = ["E","W","F","I","N","UP","B","C4","SIM","TCH","RUF"]
ignore = []
```

`BLE` (flake8-blind-except) is absent, so nothing flags a blind catch.

This is not hypothetical. #1312 was caused by exactly one such line:

```python
        except Exception:
            logger.warning("Failed to parse consent log, starting fresh")
            return ConsentLog()
```

in `creek/consent.py`. It turned every possible read failure — a torn write, a permissions error, bad bytes on disk — into "no consent was ever recorded", and the next write then destroyed the append-only audit record it had just failed to read. No gate saw it.

## Why this was declined inside #1312

Turning `BLE` on is not a bugfix, it is an epic. Counts measured at `dd58593`:

- **30** `except Exception` sites in `creek/` (`cli.py` alone has 7) — of which #1312 removes exactly one, leaving **29**
- **11** in `creek_mcp/`
- **40** remaining after #1312 lands

Since `lint.ignore` is empty, enabling `BLE` means 40 audited rewrites or 40 issue-referenced `# noqa`s in one PR. Smuggling that into a data-integrity fix would have buried the actual change and produced an unreviewable diff.

## Acceptance criteria

- [ ] Triage all 40 sites into: (a) genuinely needs the blind catch and gets an issue-referenced `# noqa: BLE001`, (b) can be narrowed to specific exception types, (c) is an actual latent bug like #1312 and gets its own issue.
- [ ] Land the narrowings in reviewable batches (by subpackage), not one mega-PR.
- [ ] Only once the count is zero, add `"BLE"` to `select` in `creek-tools/pyproject.toml` so the gate holds the line.
- [ ] Consider `crawdad/` separately — it has its own quality gate and its own config.
- [ ] `./scripts/check-all.sh` exits 0 at every step.

## Note

A cheaper, narrower forcing function already shipped in #1312: its regression tests go red the instant anyone restores `except Exception: return ConsentLog()` in `creek/consent.py`. That protects the one site that actually bled, but it does not generalise — hence this issue.

Refs #1312.

Issue #1405, opened 2026-08-11
refactor(fsio): converge three atomic-write implementations onto creek._fsio


Geoffe-Ga:
## Problem

There are now **three** atomic-write implementations in `creek-tools/creek/`, all doing tempfile + `os.replace` with slightly different cleanup semantics:

- `creek/vault/writer.py:392` — `_atomic_write_text(path, content)`
- `creek/redact/cli_commands.py:531` — `_atomic_write(file_path, content)`
- `creek/_fsio.py` — `atomic_write_text(path, content)` (added by #1312)

`_fsio.py` is the designated stdlib-only home for these primitives (it already owns `write_all` and `create_exclusive` from #987). Three copies means three places to audit when the next torn-write bug lands, and they do not agree:

- `_fsio.atomic_write_text` leaves the file at **0o600** (mkstemp's mode) and deliberately does not `mkdir`.
- `vault/writer.py:_atomic_write_text` has different `finally` cleanup and reconstructs `Path(tmp_name)`.
- `redact/cli_commands.py:_atomic_write` differs again.

## Why this was not done in #1312

#1312 fixed a live data-integrity defect (a corrupt consent log silently truncating an append-only audit record). Converging two other already-green modules onto a brand-new helper is a refactor with a different risk profile — notably the **file-mode divergence**: `_fsio.atomic_write_text` yields 0o600, which is correct for an operator-private consent record but is very likely wrong for vault notes, which today land at 0o644 via `_NEW_FILE_MODE`. Converging without deciding that per-caller would silently tighten permissions on every note in the vault.

## Acceptance criteria

- [ ] Decide the mode contract: either `atomic_write_text` grows an explicit `mode` parameter defaulting to `_NEW_FILE_MODE` (0o644) with consent passing 0o600, or each caller chmods explicitly. Do not let mkstemp's default decide it implicitly.
- [ ] `creek/vault/writer.py` and `creek/redact/cli_commands.py` call `creek._fsio.atomic_write_text` instead of their private copies; the private copies are deleted.
- [ ] Existing mode expectations for vault notes are pinned by a test **before** the switch, so a permissions regression cannot land silently.
- [ ] The `mkdir` contract is explicit: `_fsio` helpers document "parent must already exist"; any caller relying on implicit mkdir keeps its own.
- [ ] `./scripts/check-all.sh` exits 0; no new `# noqa` / `# type: ignore`.

Refs #1312, #1307, #987.

[...]

PR #1407, opened 2026-08-11
fix(consent): quarantine a corrupt consent log instead of overwriting it (#1312)

### LINKED ISSUE #1312: fix(consent): a single unparsable read silently truncates the 'append-only' consent log
**Severity:** High
**Remediation:** fix (narrow the except, quarantine rather than discard) + regression test

## Problem

The consent log advertises itself as append-only, but a single unparsable read silently
discards every prior consent record.

`ConsentManager._load_log` (`creek/consent.py:255-262`) wraps the read + parse in
`except Exception:`, logs `"Failed to parse consent log, starting fresh"`, and returns an
empty `ConsentLog()`. `_save_log` (`creek/consent.py:264-276`) then unconditionally rewrites
the whole file — so the next `record_consent` call persists only the new entry and the
previous consent history is gone.

The broad `except` also swallows `PermissionError` / `OSError` / `UnicodeDecodeError` from
`read_text`, so a transient read failure destroys the audit trail identically to a genuinely
corrupt file.

`creek/consent.py:8-9` and the class docstring at `:158-160` both promise "append-only JSON".
Taxonomy family 0 (correctness on a consent/audit path) + family 12 (over-broad error
handling that hides a bug).

Ruff's `BLE` family is not in this repo's `select` list, so no existing gate sees this.

## Evidence (corroboration)

```
record_consent("claude", ...)          -> log contains claude
truncate the JSON file by 5 characters
check_consent("claude", ...)           -> False
record_consent("chatgpt", ...)         -> log now contains ONLY chatgpt
```

The prior consent record is unrecoverable from the file.

**Signal 2 — structural.** `_load_log`'s recovery path and `_save_log`'s unconditional
whole-file rewrite are 10 lines apart in one class; nothing quarantines the unparsable file
or refuses the write. The "append-only" claim appears twice in the module and is not
implemented anywhere.

## Scope

Covers the load/save error contract in `creek/consent.py`.

Explicitly NOT in scope: the `ConsentManager` docstrings that describe prompting UI it does
not own (that lives in `cli._gate_consent`), or the unused `SourceSummary.content_types`
field — both are Low and belong in a tidy-up.

## Tasks

1. Narrow the `except Exception:` at `creek/consent.py:260` to the parse failures it means to
   handle (`json.JSONDecodeError`, `pydantic.ValidationError`), and let read errors
   (`OSError`, `PermissionError`) propagate rather than presenting as "no consent recorded".
2. On a genuine parse failure, **quarantine** the unparsable file (rename to
   `<name>.corrupt-<timestamp>`) before starting fresh, so the record is recoverable, and log
   at WARNING with the quarantine path.
3. Add regression tests: (a) a truncated log is quarantined and the new entry is written;
   (b) an unreadable log raises rather than silently reporting no consent.

## Acceptance Criteria

- [ ] A corrupt consent log is preserved on disk under a `.corrupt-*` name; the original
      bytes are recoverable.
- [ ] A `PermissionError` on read surfaces to the caller instead of being reported as
      "no consent recorded".
- [ ] `check_consent` never returns `False` because of an I/O error without saying so.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/consent.py` | Modify |
| `creek-tools/tests/test_consent.py` | Modify |

Geoffe-Ga:
## Summary

The consent log advertised itself as append-only in two docstrings and was neither append-only nor durable. `ConsentManager._load_log` wrapped read-and-parse in `except Exception:`, logged `"Failed to parse consent log, starting fresh"` — naming neither the file nor where the bytes went — and returned an empty log; `_save_log` then rewrote the whole file with a bare, non-atomic `write_text`. Ten lines apart, those two made a single unparsable read permanently destroy the record of every prior grant.

**This is `creek-tools`, not `crawdad`.** `crawdad/crawdad/consent.py` is an unrelated in-memory `PendingBatchStore` for Discord staging with zero file I/O and zero `except` clauses; `creek/classify/llm/consent.py` is env-var cloud-egress consent. Neither is touched, and crawdad's separate quality gate is not involved.

### What was actually broken — three failure modes, measured at `dd58593`

```
# 1. The prior grant is destroyed, unrecoverably.
record_consent("claude", "/data/claude")   -> log holds claude
truncate consent-log.json by 5 bytes       -> torn write, as a crash would leave it
check_consent("claude", "/data/claude")    -> False
record_consent("chatgpt", "/data/chatgpt")
  files on disk: ['consent-log.json']      <- no quarantine sibling
  records now:   ['chatgpt']
  '/data/claude' recoverable anywhere?     False
```

```
# 2. Two contradictory error contracts, four lines apart.
#    `if not self._log_file.exists()` sat OUTSIDE the try.
unreadable PARENT dir -> Path.exists() RAISES PermissionError -> bare traceback
unreadable log FILE   -> read_text() PermissionError -> swallowed -> "no consent"
```

```
# 3. The worst shape: a real grant reported as no grant.
record_consent("claude", "/data/claude")
chmod 0o000 consent-log.json
check_consent("claude", "/data/claude")    -> False
#   Consent EXISTS on disk. The gate says it does not, and under `--yes`
#   the auto-grant path then rewrites the log with one entry and prints
#   only "Consent auto-granted" — completely silent for scripted callers.
```

Nothing observable distinguished *"no consent was ever given"* from *"the record of it was destroyed"*. For an audit artifact that is the worst possible failure shape.

### After

```
=== truncated log ===
  files on disk: ['consent-log.json', 'consent-log.json.corrupt-20260811T103025-r42b8myl']
  quarantined bytes identical to torn bytes: True
  '/data/claude' recoverable from quarantine: True
  fresh log records: ['chatgpt']
  WARNING Consent log <path> was unparsable (<pydantic error>); the existing
          record has been quarantined to <path>.corrupt-<ts>-<rand> and a
          fresh log started — prior grants must be re-confirmed

=== unreadable parent dir ===   ConsentLogUnavailableError: consent log <path> is unavailable: [Errno 13] ...
=== unreadable log file ===     ConsentLogUnavailableError: consent log <path> is unavailable: [Errno 13] ...
=== quarantine itself fails === ConsentLogUnavailableError raised; corrupt bytes intact: True
```

The direction of failure is preserved throughout: every new error path **denies consent or aborts**, and none can turn a denial into a grant.

## The change

**`creek/consent.py`**
- `ConsentLogUnavailableError(OSError)`. `OSError`, not `RuntimeError`, on evidence: there are **zero** `except OSError` handlers anywhere on the consent path, while `cli.py` catches bare `RuntimeError` in four places — so `RuntimeError` is the strictly *more*-absorbed base. Single-arg `super().__init__(msg)` keeps `errno` at `None` and satisfies TRY003.
- `_quarantine_corrupt_log()` reserves a `<name>.corrupt-<timestamp>-<rand>` sibling with `tempfile.mkstemp` and commits with `os.replace`. Same-directory, so `EXDEV` is unreachable and the inode simply moves — bytes *and* mode survive with no copy step that could itself be truncated. Two quarantines in one clock second cannot collide, and an existing `.corrupt-*` can never be overwritten. On failure the corrupt bytes are left exactly where they are and the empty reservation is removed.
- `_load_log()` is now EAFP; the `exists()` pre-check is **deleted** (it was failure point #2 and a TOCTOU window — `read_text` raises `FileNotFoundError` for a missing file *and* a missing parent, the same answer the guard gave). Four arms, in this order: `FileNotFoundError` → empty log; `OSError` → raise typed; `(UnicodeDecodeError, ValidationError)` → quarantine, warn, empty log; `else: return log`.
- `_save_log()` writes through `creek._fsio.atomic_write_text` and raises the typed error, so an unwritable vault no longer tracebacks out of `record_consent`.

**`creek/_fsio.py`** — new `atomic_write_text()` (tempfile + `os.replace`), the same hardening merged for the vault index in #1307. Reusing the existing stdlib-only primitives module rather than adding a fourth copy. Recovering from torn writes while leaving the tear un-prevented would be half a fix: this removes the mechanism that manufactures the corrupt file in the first place.

**`creek/cli.py`** — `_gate_consent` split into `_gate_consent_unguarded` (body unchanged) plus a thin wrapper that turns the typed error into a red message and `typer.Exit(code=1)`. One wrapper covers all three consent calls and **both** call sites — `process` and `ingest`. `ingest` never constructs a `Pipeline`, so a handler at `pipeline.run` could not have covered it. Exit 1, not 2 (2 is usage error here); the decline paths and the `code=2` source-not-found path are untouched because `typer.Exit` is not an `OSError`.

**`creek/pipeline.py`** — docstring only, zero logic change.

## Judgement calls, stated plainly

- **`json.JSONDecodeError` is dead code and is deliberately NOT caught.** Task 1 of the issue asks for it. `ConsentLog.model_validate_json` never raises it — verified on pydantic 2.13.4 that `'{"records": 1'`, `'not json'`, `''` and `'{"records":"a"}'` all raise `pydantic_core.ValidationError`, and `isinstance(exc, json.JSONDecodeError)` is `False`. `json` is not even imported in the module; importing it to catch a never-raised exception would be new slop.
- **`UnicodeDecodeError` is reclassified as corruption, against the issue body**, which groups it with `PermissionError`/`OSError` as a transient read failure that should propagate. Bad bytes on disk do not heal, and re-reading them can only reproduce the failure — so it quarantines. It is a `ValueError`, not an `OSError`, so the handler ordering is safe.
- **A 0-byte file is quarantined** under the same uniform "unparsable, so quarantine" rule rather than special-cased to "empty". Special-casing empty would silently accept precisely the shape a torn `write_text` produces — the bug. (`'{}'` is different: it validates to an empty record list and is not quarantined.)
- **Quarantine is a deliberate write side effect of a read**, and the reason `_load_log` can raise on what looks like a read-only path. The alternative is the previous behaviour, which destroyed the evidence on the next grant.
- **File mode tightens 0o644 → 0o600.** `mkstemp` creates owner-only and `os.replace` does not restore the target's mode. Deliberate and pinned by a test: the consent log names source paths and operator identity, so owner-only is right for it. Flagged here because a reviewer will notice.
- **`_run_pipeline_or_exit` extraction was forced, not cosmetic.** Adding a fifth handler to `process` pushed it to xenon rank C (CC 11, gate is `--max-absolute B`). Extracting the handler block returns `process` to B (6) with the helper at A (5). Behaviour-preserving; the two symlink arms printed an identical message and now share one arm.
- **One honest residual.** A *corrupt* log is quarantined and still yields `has_consent = False`, so `pipeline.run` still logs `"Consent not granted — skipping ingestion."` What distinguishes destruction from never-granted in that case is the WARNING naming both paths (visible on stderr via `logging.lastResort`; `creek/cli.py` never calls `basicConfig`) plus the on-disk `.corrupt-*` artifact. Recorded in `_check_consent`'s docstring rather than papered over.
- **Ruff `BLE` is declined here** and filed as #1406. Enabling it means 40 audited rewrites (29 in `creek/` after this PR, 11 in `creek_mcp/`) or 40 issue-referenced noqas — an epic, and smuggling it into a data-integrity fix would make the diff unreviewable. The cheaper forcing function ships here: the regression tests go red the instant anyone restores `except Exception: return ConsentLog()`.
- **Atomic-write convergence** across `vault/writer.py`, `redact/cli_commands.py` and `_fsio.py` is filed as #1405, not done here — their cleanup semantics and mode contracts differ, and converging two already-green modules is a different risk profile.

## Test plan

**Every test below was proven to fail without the fix** (standing repo rule). Each guard was reverted *alone* — source copied aside, never `git stash`, which is shared across live worktrees — and the intended tests confirmed red:

| Mutation applied alone | Result |
|---|---|
| Restore the blind `except Exception: return ConsentLog()` | **9 red** — quarantine, both propagate arms, pipeline abort |
| Revert `_save_log` to the bare non-atomic `write_text` | **1 red** — `test_save_log_failure_leaves_the_previous_log_intact` |
| Restore the `exists()` pre-check outside the try | **1 red** — parent-dir test dies with a raw `PermissionError` |
| Drop `UnicodeDecodeError` from the quarantine arm | **1 red** — non-UTF-8 test |
| Remove the quarantine call | **5 red** — all preservation/collision tests |
| Strip tempfile staging from `atomic_write_text` | **3 red** — including the 0o600 mode pin |
| Remove the mid-run `ConsentLogUnavailableError` CLI handler | **1 red** |

Source restored byte-identical after each; suite green again.

New coverage — 21 tests:
- `tests/test_consent.py::TestConsentLogRecovery` (12): truncated log quarantined with byte-identical preservation and the prior grant recoverable; directory-at-path raises; unreadable file raises; unreadable **parent** raises through the same contract; non-UTF-8 quarantined without raising; 0-byte quarantined; two quarantines in one second do not collide and a pre-seeded `.corrupt-*` is untouched; a failed quarantine never lets `_save_log` clobber the corrupt bytes; read-only-directory quarantine failure preserves them; a missing log is a first run, not a corruption; a failed save leaves the previous log intact with no `.tmp` residue; a valid log is never quarantined.
- `tests/test_fsio.py` (5 added, all 8 pre-existing `write_all`/`create_exclusive` tests retained): round-trip, wholesale replace, no `.tmp` residue, owner-only mode, original preserved when the rename fails.
- `tests/test_cli.py` (3): `process --yes` and `ingest --yes` against a broken log both exit 1 with a message naming the log and **without** printing "Consent auto-granted"; plus a log that breaks *mid-run* (readable at the gate, gone by Stage 0) reported rather than tracebacking.
- `tests/test_pipeline.py` (1): an unreadable log aborts `pipeline.run` instead of taking the "skip ingestion, still index" branch as if the operator had declined.

Chmod-based tests carry a root/Windows `skipif` because `chmod` no-ops for root; the propagate and quarantine-failure arms are *also* anchored by root-proof tests (directory-at-path, monkeypatched `os.replace`) so both arms stay covered wherever CI runs. `0o500` is deliberately never used for a read-blocking test — I verified it is a no-op for directory traversal; only `0o600`/`0o000` provoke `PermissionError`.

**Gate 2: `./scripts/check-all.sh` exit code 0**, 12/12 checks.
- 9233 passed, **5 skipped** — all pre-existing environment skips (Ollama unavailable, live API keys unset). **Zero skips in the consent suites.**
- `creek/consent.py` **100%**, `creek/_fsio.py` **100%**, `creek/cli.py` 90.17% — no coverage waiver was added or needed.
- Ruff verified with `--no-cache` (lint and format), mypy strict, bandit, xenon `--max-absolute B`, refurb 0, tryceratops 0, interrogate ≥95%, pylint 9.71.

No `# noqa`, no `# type: ignore`, no `# pylint: disable`, no `@pytest.mark.skip`, no `--no-verify`, no lowered thresholds, no deleted tests. The one `# noqa: TC003` that *was* in `consent.py` is **removed**, because `Path` is now genuinely used at runtime. I swept the whole diff including tests before pushing (#1366).

One note on the issue's final acceptance criterion, "All pre-commit hooks pass (`pre-commit run --all-files`)": I did not run that command. It has previously edited unrelated files in this repo, including stripping the BOM from a fixture whose entire purpose is having one. `./scripts/check-all.sh` is the gate and it exits 0.

Closes #1312

### DIFF
--- a/creek-tools/creek/_fsio.py
+++ b/creek-tools/creek/_fsio.py
@@ -1,4 +1,4 @@
-"""Short-write-safe primitives for raw file-descriptor I/O (#987).
+"""Torn-write-safe file primitives: short writes (#987), atomic replace (#1307).
 
 ``os.write`` is allowed to write *fewer* bytes than it was handed: the
 POSIX ``write(2)`` contract only promises some forward progress, so a
@@ -9,10 +9,26 @@ Both :mod:`creek.vault.writer` and :mod:`creek.save.writer` create notes
 by writing straight to the *final* path (``O_CREAT | O_EXCL``, with no
 tempfile + ``os.replace`` staging step), so a short write there does not
 merely spoil a scratch file — it files a half-written note under the
-real name, with no exception raised for anyone to notice. These two
-helpers live here instead of being triplicated across those call sites,
-so the drain loop and the partial-file cleanup have exactly one
-implementation to audit.
+real name, with no exception raised for anyone to notice.
+:func:`write_all` and :func:`create_exclusive` live here instead of being
+triplicated across those call sites, so the drain loop and the
+partial-file cleanup have exactly one implementation to audit.
+
+:func:`atomic_write_text` covers the *overwrite* case the two creation
+helpers deliberately do not: ``Path.write_text`` truncates the target
+before it writes, so an interrupted rewrite leaves a half-file under the
+real name and the previous contents gone. Staging into a tempfile and
+committing with ``os.replace`` means a reader sees either the whole old
+file or the whole new one, never a splice of the two — the same
+hardening merged for the vault index in #1307.
+
+Near-identical private implementations already exist at
+``creek/vault/writer.py:392`` (``_atomic_write_text``) and
+``creek/redact/cli_commands.py:531`` (``_atomic_write``). Converging
+them onto this helper is deliberate follow-up work, not part of #1312:
+each carries call-site-specific behaviour (a ``mkdir`` in the former, a
+symlink-materialisation contract in the latter) that has to be unpicked
+against its own tests.
 
 Stdlib-only by design: this module sits underneath the writers and must
 not drag any of the package in behind it.
@@ -23,10 +39,8 @@ from __future__ import annotations
 import contextlib
 import errno
 import os
-from typing import TYPE_CHECKING
-
-if TYPE_CHECKING:
-    from pathlib import Path
+import tempfile
+from pathlib import Path
 
 _EXCLUSIVE_CREATE_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_EXCL
 """Create-or-fail open flags: never truncates or clobbers an existing file."""
@@ -122,3 +136,70 @@ def create_exclusive(path: Path, data: bytes) -> None:
         with contextlib.suppress(OSError):
             path.unlink()
         raise
+
+
+def atomic_write_text(path: Path, content: str) -> None:
+    """Replace *path*'s contents with *content* in a single atomic step.
+
+    *content* is staged in a uniquely named tempfile inside *path*'s own
+    directory and committed with ``os.replace``, so a concurrent reader
+    sees either the whole previous file or the whole new one and never a
+    splice of the two. If the write or the rename fails, the original
+    file is left byte-intact and the stage file is removed — unlike
+    ``Path.write_text``, which truncates the target *before* it writes
+    and so turns any interruption into permanent data loss.
+
+    Two consequences of ``mkstemp`` + ``os.replace`` that callers must
+    know about, both measured rather than assumed:
+
+    - The resulting file mode is **0o600**, inherited from
+      :func:`tempfile.mkstemp` and deliberately tighter than this
+      module's ``_NEW_FILE_MODE`` (0o644). The first consumer is an
+      operator-private audit record naming source paths and operator
+      identity, so owner-only is the right default; ``write_text``
+      would have produced 0o644.
+    - ``os.replace`` does *not* preserve the target's existing mode, so
+      a file that was 0o644 tightens to 0o600 on its next write through
+      this helper. The change is a one-way narrowing — it can never
+      widen a file that was already private.
+
+    Deliberately does not ``mkdir``: as with :func:`create_exclusive`,
+    the parent directory must already exist. The stage file has to live
+    beside the target anyway (a cross-filesystem ``os.replace`` raises
+    ``EXDEV``), and conjuring a directory tree here would silently
+    absorb a caller's wrong path.
+
+    Args:
+        path: File to overwrite. Its parent directory must already
+            exist and be writable, because the stage file is created
+            there.
+        content: The full text to write, encoded as UTF-8.
+
+    Raises:
+        OSError: Propagated untouched from the staging create, the
+            write, or the rename — an unwritable directory
+            (``EACCES``), a full disk (``ENOSPC``), and so on. Any
+            stage file left behind is unlinked first, and a failure to
+            unlink it is suppressed so the caller still sees the
+            original error rather than the cleanup's.
+    """
+    # ``mkstemp`` runs outside the ``try`` on purpose: if it raises there
+    # is no descriptor and no stage file to clean up, and ``tmp_name``
+    # would be unbound in the ``finally``.
+    fd, tmp_name = tempfile.mkstemp(
+        prefix=f".{path.name}.", suffix=".tmp", dir=path.parent
+    )
+    try:
+        # ``fdopen`` takes ownership of ``fd``, so the ``with`` closes it
+        # exactly once on every path — including the write failing.
+        with os.fdopen(fd, "w", encoding="utf-8") as handle:
+            handle.write(content)
+        os.replace(tmp_name, path)
+    finally:
+        # After a successful ``os.replace`` the stage name is gone, so
+        # the existence check is what makes this cleanup a no-op on the
+        # happy path rather than a spurious ``FileNotFoundError``.
+        tmp_path = Path(tmp_name)
+        if tmp_path.exists():
+            with contextlib.suppress(OSError):
+                tmp_path.unlink()
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -26,7 +26,7 @@ from creek.classify.privacy_filter import (
     record_privacy_override,
 )
 from creek.config import AIStyleConfig, load_config, resolve_config_path
-from creek.consent import ConsentManager
+from creek.consent import ConsentLogUnavailableError, ConsentManager
 from creek.models import PraxisPotential, PrivacyTier
 from creek.pipeline import (
     Pipeline,
@@ -228,7 +228,7 @@ def _format_summary(file_count: int, size_bytes: int) -> str:
     return f"{file_count} file(s), {mb:.1f} MB"
 
 
-def _gate_consent(
+def _gate_consent_unguarded(
     *,
     source_path: Path,
     vault_path: Path,
@@ -256,7 +256,15 @@ def _gate_consent(
     Raises:
         typer.Exit: With code ``1`` when the operator declines, or
             when consent has not been recorded and the caller is
-            non-interactive without ``--yes``.
+            non-interactive without ``--yes``. With code ``2`` when
+            *source_path* does not exist.
+        ConsentLogUnavailableError: Propagated from the consent-log
+            read (``check_consent``) or write (``record_consent``)
+            when the log cannot be read or safely set aside. This
+            function deliberately does not handle it; the
+            :func:`_gate_consent` wrapper turns it into a legible
+            refusal so an unreadable log can never be mistaken for a
+            source the operator simply never approved.
     """
     log_dir = _consent_log_dir(vault_path)
     manager = ConsentManager(log_dir=log_dir)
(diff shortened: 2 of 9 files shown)


Geoffe-Ga:
## Summary
Fixes a real data-integrity bug: `ConsentManager._load_log` caught bare `Exception`, silently discarded unparsable consent-log bytes, and returned an empty log — which `_save_log` (a non-atomic `write_text`) would then permanently overwrite on the next grant, destroying prior consent records with no recovery path and no distinguishable signal from "never granted." This PR replaces that with a typed `ConsentLogUnavailableError(OSError)`, a byte-preserving quarantine of unparsable logs, an atomic `_fsio.atomic_write_text` primitive for the rewrite, and CLI/pipeline handling that turns log-unavailability into an explicit refusal rather than a silent `False`.

## Strengths
- Correct exception-ordering: `FileNotFoundError` (subclass of `OSError`) is caught before the general `OSError` arm; `UnicodeDecodeError`/`ValidationError` (not `OSError` subclasses) are handled separately and reclassified as corruption with justification.
- The `exists()` pre-check removal closes a real TOCTOU/handler-bypass bug (`Path.exists()` sat outside the `try`, letting `PermissionError` on an unreadable parent escape untyped) — verified this against the actual pre-change code via the diff context.
- `atomic_write_text` and `_quarantine_corrupt_log` both stage via `tempfile.mkstemp` in the *same* directory as the target, so `os.replace` is same-filesystem (no `EXDEV`) and atomic; failure paths clean up the empty reservation while leaving the original bytes untouched — traced this through both the implementation and the corresponding failure-injection tests (`monkeypatch` on `os.replace`) and it holds.
- I manually traced both `process` and `ingest` call sites (`cli.py:1495`, `cli.py:1697`) to confirm both go through the *wrapped* `_gate_consent`, not the unguarded inner function, and that `pipeline.run()` has no intervening `except OSError`/`except Exception` between `_check_consent` and `_run_pipeline_or_exit`'s explicit `ConsentLogUnavailableError` handler — the error propagates exactly as claimed, with no accidental swallow point.
- The `process`-function extraction (`_run_pipeline_or_exit`) is behavior-preserving; the two symlink-error arms it collapsed already printed identical messages.
- Honest, explicit judgement calls in the PR body (dead `json.JSONDecodeError`, `UnicodeDecodeError` reclassification, 0-byte quarantine, the corrupt-log residual ambiguity, deferred `BLE` lint adoption as #1406, deferred atomic-write convergence as #1405) — these read as genuine engineering tradeoffs, not scope-dodging, and each is backed by a specific reason.

## Security Concerns
🟢 LOW — File mode for the consent log tightens from 0o644 to 0o600 (inherited from `mkstemp`), which is a hardening, not a regression, and is deliberate/tested. No new attack surface identified; quarantine/atomic-write logic stays within the log's own directory, so no path-traversal or symlink concerns beyond what already existed.

## Problems
None found blocking merge. I traced the full exception-handling chain (consent.py → cli.py → pipeline.py) by hand against the new tests and did not find a path where a genuine I/O failure or corrupt log could still be silently reported as "no consent" or silently overwrite a prior grant.

## Code Quality
- `ConsentLogUnavailableError.__init__` deliberately uses single-arg `super().__init__` to keep `errno` at `None` (documented rationale: satisfies TRY003, keeps `OSError.__new__`'s errno-subclass magic from firing on a subclass with a custom `__init__`). This is a reasonable, well-justified choice, though a future reader relying on `.errno` on this exception type should know it's always `None` — the original errno is only reachable via `__cause__`. Not blocking, just worth flagging for anyone building on this exception later.
- The two nearly-identical private atomic-write implementations in `vault/writer.py` and `redact/cli_commands.py` are correctly *not* converged here (filed as #1405) — good scope discipline given their differing `mkdir`/symlink contracts.

## CLAUDE.md Adherence
✅ TDD workflow followed — the PR body documents a revert-and-confirm-red methodology per guard (9+1+1+1+5+3+1 red counts), matching the "write tests first" / "every test proven to fail without the fix" house rule.
✅ No forbidden shortcuts — no `noqa`/`type: ignore`/`pylint: disable`/skipped tests/lowered thresholds are introduced; an existing `# noqa: TC003` is correctly *removed* now that `Path` is genuinely used at runtime.
✅ Commit message follows Conventional Commits (`fix(consent): ...`).
✅ Coverage claims (100% on `consent.py`/`_fsio.py`, 90.17% on `cli.py`) meet the ≥90% aggregate / per-file thresholds if accurate — I could not execute `./scripts/check-all.sh` in this review sandbox (bash execution was blocked), so this is taken on the PR's word rather than independently re-verified; the level of self-reported detail (per-mutation red counts, exact tool versions) is consistent with genuine execution rather than a fabricated claim.
✅ Docs updated (`docs/ingestion.md`) with an operator-facing section on quarantined logs — consistent with "Documentation Standards" requiring updates when behavior changes.

## Testing
Strong. 21 new tests across `test_consent.py`, `test_fsio.py`, `test_cli.py`, `test_pipeline.py` cover: truncated-log quarantine + recoverability, directory-at-path, unreadable file/parent (with root/Windows `skipif` for chmod-based tests), non-UTF-8, 0-byte, quarantine-collision avoidance, quarantine-failure preserving corrupt bytes, save-failure leaving the previous log intact, CLI refusal messaging (including the "Consent auto-granted" negative assertion), and a mid-run log-disappearance scenario distinguishing the gate-time read from the pipeline's Stage-0 re-read. I independently verified the fixture wiring (`manager`/`consent_dir`) and the exception-flow assertions in several of these tests against the implementation and found them accurate to what the code does.

## Documentation
Docstrings are extensive and accurate to the implementation for every touched function (`_load_log`, `_save_log`, `_quarantine_corrupt_log`, `atomic_write_text`, `ConsentLogUnavailableError`, `_gate_consent`/`_gate_consent_unguarded`, `_run_pipeline_or_exit`, `_check_consent`). The `docs/ingestion.md` addition correctly tells operators not to blindly delete `.corrupt-*` files, which is the exact hazard this fix protects against.



## Verdict: LGTM

No blocking correctness, security, or test-coverage issues surfaced after a line-by-line trace of the exception-handling chain across consent.py, _fsio.py, cli.py, and pipeline.py against the new tests; the fix is well-scoped, the atomicity/quarantine mechanics check out, and the CLAUDE.md TDD/no-shortcuts/documentation requirements are all satisfied based on the evidence available.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5256772036 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1410, opened 2026-08-11
fix(voice): voice-check's fallback build_fingerprint ignores voice_audience_weighting


Geoffe-Ga:
## Summary

`_resolve_voice_fingerprint` in `creek-tools/creek/cli.py` calls
`build_fingerprint(vault_path, ai_style)` with no `audience_weighting=`, so its
fallback-built fingerprint silently uses the flat-average path instead of the vault's
`voice_audience_weighting` settings.

`build_fingerprint`'s `audience_weighting` parameter defaults to `None`, which means
**off** on that path — so an omitted keyword is not a neutral omission, it disables the
graduated audience-authority model shipped in epic #632/#633/#634.

Found during the Gate 2.5 review of #1313. It is **pre-existing, not a regression from
that PR**, and was outside its stated inventory (which covered `VoiceProfileGenerator`,
`VoiceExemplarCollector` and `generate_register_samples` on the exemplar path, plus the
two `_generate_fingerprint` sites).

## Why it is low severity

- Only `creek voice-check` reaches it (one caller).
- It is a **fallback**: `_resolve_voice_fingerprint` prefers the persisted
  `<vault>/00-Creek-Meta/voice-fingerprint.json` and only builds when that is absent. So
  it bites only on a vault that has never run `creek report --type fingerprint`.
- `voice-check` is a pure read-only diagnostic — it writes no vault artifact and copies
  no fragment bodies, so this is not privacy-relevant.

The inconsistency is still worth closing: the same vault can report one voice distance
before its first `report --type fingerprint` and a different one after, with nothing in
the output explaining why.

## Suggested fix

`voice-check` already holds the full `CreekConfig` where it derives `ai_style`. Thread
`voice_audience_weighting` through `_resolve_voice_fingerprint` into `build_fingerprint`,
matching `_report_fingerprint`.

Consider also whether `build_fingerprint` belongs in the `_AUDIENCE_WEIGHTING_CALL_NAMES`
structural guard in `tests/test_mcp_report_tier_ceiling.py` once every production caller
states the keyword — #1313 deliberately left it out because this site did not yet.

## Acceptance criteria

- [ ] `_resolve_voice_fingerprint` accepts and forwards the vault's
      `voice_audience_weighting`.
- [ ] A regression test runs `creek voice-check` on a vault with **no persisted
      fingerprint** and an edited `voice_audience_weighting`, from a config-less working
[...]

Issue #1409, opened 2026-08-11
fix(mcp): bare load_config() reads the cwd, not the vault — remaining report/server sites


Geoffe-Ga:
## Summary

`creek_mcp/tools/report.py` and `creek_mcp/server.py` read configuration through a bare
process-wide `load_config()` in several places. That call resolves `creek_config.yaml`
against the **current working directory** (`creek/config.py`, the `config_path is None`
branch) and never looks at `<vault>/00-Creek-Meta/creek_config.yaml`.

The consequence is silent: a vault-scoped setting is ignored, the run logs
"Config file creek_config.yaml not found; running with built-in defaults", and the tool
proceeds with built-in defaults while the operator's vault config says something else.
Nothing fails, so nothing is noticed.

This is the same class of defect as #1313, which fixed it only for the voice/fingerprint
path (`_generate_voice`, `_generate_rhetorical_patterns`, `_generate_fingerprint`) by
introducing a lazy `_vault_config(request)` helper in `creek_mcp/tools/report.py`. The
remaining sites read **other** config sections and were left out of that PR's scope
deliberately.

## Affected sites

`creek_mcp/tools/report.py` (all read `.embeddings`):

- `_generate_unnamed` — `load_config()` for the unnamed-digest embedding linker
- `_generate_paradox` — `load_config().embeddings`
- `_generate_synchronicity` — `load_config().embeddings`

`creek_mcp/server.py` — three further bare reads (around the mine/reflect/upload wrappers).

Each carries an inline comment pointing at this issue.

## Suggested fix

Route each through the existing `_vault_config(request)` helper in
`creek_mcp/tools/report.py` (or the equivalent
`load_config(resolve_config_path(vault_path, None), warn_on_missing=False)` spelling used
by `creek_mcp/tools/draft.py` and `creek/cli.py::_load_config_for_vault`).

Worth considering as part of this: the incantation now has five spellings in the tree
(`creek_mcp/tools/draft.py`, `creek/author/conductor.py`, `creek/author/agents.py`'s
private `_load_config`, `creek/cli.py::_load_config_for_vault`, and
`creek_mcp/tools/report.py::_vault_config`). Consolidating them into one shared helper is
the natural home for that cleanup — #1313 declined to do it there because it would have
widened the highest-blast-radius file in that diff for zero tests that could fail.

## Acceptance criteria

[...]

Issue #632, opened 2026-06-21
EPIC: Build the voice from audience-facing documents + discerning AI-tell scrub


Geoffe-Ga:
## Summary
The creek voice fingerprint is built from a **diluted corpus** — ~31k fragments dominated by private chats/journals — so it encodes the user's *texting average*, not their *audience-facing voice*. Proof: the fingerprint reports `em_dash_density` 0.39/1000 words, but the user's real Substack essay uses ~8/1000 (20×). A draft that matches the diluted average scores `voice_distance` **0.00** ("perfect") while not sounding like the user.

Re-base the **voice** on **audience-facing documents** (essays, Substack, published course material) while keeping **all** documents in the knowledge/retrieval corpus, and make AI-tell scrubbing **discerning** (strip genuine tells, keep authentic signatures).

## Goals
1. Voice fingerprint built primarily from audience-facing docs; private chats/journals down-weighted/excluded from VOICE but unchanged for retrieval/lore.
2. A real audience-facing classifier (today's signal is platform-grep and doesn't scope the fingerprint).
3. Discerning scrub: keep authentic signatures (curly quotes already protected; reinforce others), strip only genuine tells.

## Non-goals
- Blog apparatus (links/CTAs/subheads) — added manually by the author.
- Per-subagent voice injection into the FEAT-041 desk — deferred; the active path is single-pass `creek draft`.

## Build on existing (don't duplicate)
- FEAT-040.x AI-style subsystem: `generate/ai_style/` (fingerprint, features, vault-relative tell framework + distance, typography/structure sanitizers, revise guard, voice-targets preamble).
- FEAT-041 attribution: `voice_weight` + `representativeness` on `Fragment`; `VoiceAudienceWeightingConfig` (config.py:1165-1186); `audience_authority` already weights register exemplars (generate/voice.py).
- Curly-quote/em-dash signature protection already works (sanitize_typography.py only straightens if the fingerprint shows the user does NOT use them).

## Children (tracer-sequenced)
1. Scope the voice fingerprint to audience-facing documents (v0 using existing signals).
2. Buff the audience-facing classifier beyond grep + wire it into the fingerprint scoping.
3. Discerning scrub: per-user polarity + reinforce currently-unmeasured authentic signatures.

## Success
[...]

PR #1411, opened 2026-08-11
fix(voice): wire voice_audience_weighting into the exemplar path

### LINKED ISSUE #1313: fix(voice): voice_audience_weighting is documented and shipped but inert — the exemplar path never receives it
**Severity:** High
**Remediation:** fix (thread the config into the exemplar path) + regression test

## Problem

`voice_audience_weighting` is a shipped, documented config knob that the entire voice-exemplar
path ignores. Setting `enabled: false` changes nothing — the 1.5x OPEN authority multiplier
is still applied.

`docs/configuration.md:373` states: *"Each voice exemplar's ranking score is multiplied by
`privacy_tier_authority[tier] x representativeness_authority[value]` … Setting `enabled:
false` makes every authority `1.0` (the pre-weighting ranking)."*

But no production construction site on the exemplar path passes `audience_weighting=`:
`creek/generate/voice.py:2520` (`generate_register_samples` → `VoiceExemplarCollector(override=override)`),
`creek/cli.py:2749` (`VoiceProfileGenerator(override=override)`),
`creek_mcp/tools/report.py:145` and `:190`, `creek/generate/lexicon.py:787`.
So the default `VoiceAudienceWeightingConfig()` (`creek/config.py:1583`, `enabled: bool = True`)
is always used. The only two config-sourced reads in the tree are on the *fingerprint* path
(`creek/cli.py:3011`, `creek_mcp/tools/report.py:277`).

Taxonomy family 3 (fake/aspirational feature flag) — a privacy-shaped knob an operator can
set and reasonably believe. Issues #554 and #557 (which shipped the knob and its docs) are
both closed.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Vault config `voice_audience_weighting: {enabled: false}`;
two conviction/300-word fragments, `a-personal` (PERSONAL) and `b-open` (OPEN). With
weighting genuinely off both score 4 and tie-break ascending-id, so `a-personal` should rank
first. Actual output of `_report_voice`:

```
config voice_audience_weighting.enabled = False
_Summary.md ranking: ['- [[b-open|b-open]]', '- [[a-personal|a-personal]]']   # 1.5x OPEN multiplier still applied
instrumented constructor state: [('VoiceExemplarCollector', True), ('VoiceProfileGenerator', True), ('VoiceExemplarCollector', True)]
```

**Signal 2 — structural.** `grep -rn "audience_weighting" creek creek_mcp` returns exactly two
config-sourced call sites, both `build_fingerprint`. Every exemplar-path constructor takes the
default.

## Scope

Covers threading `config.voice_audience_weighting` into the voice-exemplar ranking path.

Explicitly NOT in scope: `_eligible_register` ignoring `voice_proxy_eligible` (#1213), the
default-ceiling consent gate (#1212), `_is_safe_sample_stem` hardening (#1214), or the
`_report_voice` double walk (#1220).

## Tasks

1. Thread the loaded `CreekConfig.voice_audience_weighting` into `VoiceExemplarCollector` and
   `VoiceProfileGenerator` at all five construction sites listed above.
2. Also thread it into `build_voice_authenticity_report`
   (`creek/generate/voice_authenticity.py:315-342`), which currently constructs a **fresh
   default** config at `:224` and therefore reports `weighting: ON` unconditionally. Its
   module docstring (`voice_authenticity.py:14-15, 26-27`) still describes `weighting_active`
   as *"a stub `False`"*, which is stale in the opposite direction — fix both together, since
   fixing this issue alone turns that docstring into an outright lie.
3. Add a regression test asserting `enabled: false` restores id-order tie-breaking in
   `_Summary.md`.

## Acceptance Criteria

- [ ] With `voice_audience_weighting.enabled: false`, two equal-scoring exemplars rank by
      ascending id, not by tier authority.
- [ ] `creek voice-check` reports `weighting: OFF` when the vault config disables it.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/generate/voice.py` | Modify |
| `creek-tools/creek/generate/voice_authenticity.py` | Modify |
| `creek-tools/creek_mcp/tools/report.py` | Modify |
| `creek-tools/tests/test_voice_audience_authority.py` | Modify |

Geoffe-Ga:
## Summary

`voice_audience_weighting` shipped with epic #632/#633/#634, is written into every vault by `creek init`, and is documented in two places as governing how voice exemplars are ranked. On the **exemplar** path it did nothing at all.

Every production construction of `VoiceProfileGenerator`, `VoiceExemplarCollector` and `generate_register_samples` omitted the keyword, and `VoiceExemplarCollector.__init__` turned the resulting `None` into a fresh `VoiceAudienceWeightingConfig()` — an *enabled* one. So a vault with `enabled: false` on disk still had the 1.5x `open` multiplier applied to its ranking, and `creek voice-authenticity` reported `weighting: ON` regardless, because `_probe_audience_mix` constructed its own config object and reported that default back as an observation.

This wires the vault's config through to the ranking, and corrects two adjacent sites that only *looked* wired.

### The trap: the two "already-working" sites were broken too

`creek/cli.py::_report_fingerprint` and `creek_mcp/tools/report.py::_generate_fingerprint` did read the config — via a bare `load_config()`. That call resolves `creek_config.yaml` against the **current working directory** and never looks at `<vault>/00-Creek-Meta/creek_config.yaml`. So `creek report --type fingerprint --vault X` silently ignored X's settings and ran on built-in defaults.

Copying that pattern to the exemplar sites would have shipped a fix that is still inert for `creek report --type voice --vault X` — the primary invocation form — while passing any regression test that set `CREEK_CONFIG` or chdir'd into the vault. Every site now uses the vault-scoped `load_config(resolve_config_path(vault_path, None), warn_on_missing=False)` spelling, the precedent already in the tree at `creek_mcp/tools/draft.py`.

**Every test in the new module runs from a config-less working directory with `CREEK_CONFIG` unset.** That is the half of the suite that tells a real fix from a fake-green one. Mutation M3 below proves it earns its place.

## What changed

| File | Change |
|---|---|
| `creek/generate/voice.py` | `VoiceExemplarCollector._audience_weighting` → public `audience_weighting`; `VoiceProfileGenerator` stops keeping a second copy and reads `self._collector.audience_weighting`; `generate_register_samples` gains the keyword and forwards it |
| `creek/cli.py` | `_load_config_for_vault` gains `warn_on_missing`; `_report_voice` and `_report_rhetorical_patterns` resolve the vault's config and state it; `_report_fingerprint` drops bare `load_config()` |
| `creek_mcp/tools/report.py` | new lazy `_vault_config(request)` helper, used by `_generate_voice`, `_generate_rhetorical_patterns`, `_generate_fingerprint` |
| `creek/generate/voice_authenticity.py` | `_probe_audience_mix` takes the real config; `build_voice_authenticity_report` resolves it from the `vault_path` it already owns; stale module docstring corrected |
| `creek/generate/ai_style/fingerprint.py` | docstrings only — the `0.0` semantics and the default asymmetry |
| `creek/generate/lexicon.py` | docstring only — the recorded decision not to thread it |
| `docs/configuration.md`, `docs/generation.md` | the two undocumented maps, which path reads which, the `0.0` split, the privacy note |

### Single storage location

`VoiceProfileGenerator` previously stored `self._audience_weighting` **and** passed it to its default collector. When a collector is injected the two disagreed: `_rank_exemplars` and `_stream_into` read the generator's copy while `save_exemplars` used the collector's. That contradicts the design the same constructor's docstring already spells out for `override`. Nothing in the repo passes `collector=`, so this refactor is behaviourally inert today — it just stops the contradiction being reachable.

### Deliberate deviations from the acceptance criteria

**1. Five outer production sites, not six.** The AC counts `creek/generate/lexicon.py`'s collector as a sixth site. Threading it would be a no-op no test could fail on, and an untestable edit is worse than an honest omission because it reads as working wiring. The proof is three links long:

- `collect_all_exemplars` performs no ranking and no capping — it returns the whole eligible corpus, so the multiplier has nothing to reorder or cut.
- `extract_patterns` is called there with no `weights=`, and `weights` reaches only `citation_density`.
- `build_lexicon` consumes `patterns` solely via `_build_metaphors`, which reads only `metaphor_families`.

`generate_lexicon` is therefore absent from the audience-weighting guard set while remaining in the override set, and carries a `Note:` recording why. What guards it instead is a behavioural **invariance tripwire** asserting the lexicon output is identical across enabled/disabled vaults — which, unlike a guard exclusion, goes red the day the lexicon starts consuming a weighted metric.

**2. The AC names the wrong command.** It says `creek voice-check` prints the weighting line. `voice-check` is the unrelated voice-distance scanner; the command is `creek voice-authenticity`.

**3. `build_voice_authenticity_report` gains no keyword.** A `None`-defaulting kwarg would fall straight back to the fabricated default this fix removes, and no structural guard covers that function — so a future caller could silently reintroduce the exact bug. Resolving internally from the `vault_path` it already owns leaves no call site to get wrong, and keeps all 17 existing invocations working unchanged.

## Behaviour changes, stated plainly

**For operators who never touched the knob: provably nothing.** The YAML `creek init` writes is byte-identical to the code default the exemplar path was already using — verified field by field, including `platform_authority` and `audience_authority`.

**For operators who deliberately edited it: their edit now takes effect** — different exemplars ranked, different fragment bodies persisted verbatim under `07-Voice/Register-Samples/` (with stale ones pruned), and a different draft voice downstream. That is the behaviour those edits asked for. It is not gated behind an opt-in.

**A second behaviour change rides along.** Fixing `_report_fingerprint`'s config source also moves the `ai_style` section onto the vault's config for `report --type fingerprint`. It is the same one-line defect and the same correct fix, but it is a second change in one PR and reviewers should see it named. No existing test depended on the cwd behaviour.

### Privacy

**This knob cannot widen corpus membership, and I verified that rather than assuming it.** Membership is decided upstream and independently by `within_ceiling(raw, self.override)` and `_eligible_register(fragment)`; the weighting is read in exactly three places, all of them ranking (`_score`, `_exemplar_score`, and the `weights=` passed to `extract_patterns`), none of them a gate. `rank_exemplars` returns `scored[:max]` regardless of score, so no setting — even `intimate: 10.0` — can admit an intimate or above-ceiling fragment.

**But turning it OFF widens how much private or borrowed prose is copied verbatim into the vault.** Register samples are source files copied byte for byte, and the knob changes *which* fragments survive the top-N cut. With 20 `reference` fragments and one `self`-authored one competing for a cap of 20, `enabled: true` keeps the self-authored sample; `enabled: false` drops it and the corpus becomes pure borrowed material, which then seeds drafts. Operator-directed and permissible — but it was undocumented, and now is not.

### The `0.0` asymmetry, resolved

A zero authority is a membership **gate** on the fingerprint path (`if weight > 0.0`) but only **de-ranks** on the exemplar path. I kept both and documented the difference rather than unifying them, because each unification is a regression:

- Making `0.0` non-excluding on the fingerprint path would stop `intimate: 0.0` — the shipped default — keeping intimate content out of the fingerprint. Worth being precise about when that actually bites, and the docs now are: `_eligible_texts` already skips intimate fragments unless its caller passes `include_intimate=True`, so on a default run the zero authority is a redundant second gate. It becomes the operative one exactly when an operator opts intimate content in — at which point the shipped `0.0` quietly overrides that opt-in. That is surprising enough to document, which `docs/configuration.md` now does, including how to raise it deliberately.
- Making `0.0` excluding on the exemplar path would let a ranking knob silently override an operator's explicit `allow_intimate` consent, turning a preference into a consent gate.

The two paths mean different things by "weight": the fingerprint computes a weighted **mean**, where a zero-weight member is meaningless and would inflate `fragment_count`; the exemplar path computes a **ranking**, where selection is positional. Documented in both docstrings and in `docs/configuration.md`.

The **default** asymmetry (`build_fingerprint`'s `None` means off, the collector's means on) is likewise documented rather than changed — flipping the collector's would silently disable the shipped epic, and flipping the fingerprint's is a behaviour change for library callers with no issue behind it. It is now unreachable in production anyway, because the structural guard requires every production caller to state the value explicitly.

## Test plan

`./scripts/check-all.sh` exits **0** — 12/12 checks, 9261 passed, 94.55% branch coverage, ruff `--no-cache`. Integration lane 357 passed; e2e lane 25 passed. The only 5 skips are pre-existing environment skips (Ollama, live API keys); none come from this PR.

New module `tests/test_voice_audience_weighting_wiring.py` — 21 tests (23 cases), every one under a config-less cwd with `CREEK_CONFIG` unset:

| Test | Guards |
|---|---|
| `..._ranks_unweighted_when_vault_config_disables_it` / `..._enables_it` | the two-vault ordering diff; one config value, opposite orderings |
| `test_weighting_changes_which_bodies_are_persisted_verbatim` | top-N **set** membership, not just order |
| `test_enabling_the_weighting_prunes_the_sample_that_left_the_cut` | the displaced fragment's verbatim copy is deleted, not orphaned |
| `test_profile_sample_passages_depend_on_the_vault_weighting` | CLI `VoiceProfileGenerator`; `z-open` named, not merely "different" |
| `test_custom_privacy_tier_authority_inverts_the_ranking` | the whole config object is threaded; the expected order is one **no other configuration can produce** |
| `test_the_vault_config_beats_a_cwd_config_that_names_the_vault` | precedence when the two resolutions diverge |
| `test_weighting_can_never_widen_corpus_membership` | privacy, at both ceilings |
| `test_shipped_default_config_is_byte_identical_to_the_code_default` | the no-op claim, field by field |
| `test_untouched_default_config_produces_unchanged_voice_output` | whole-tree equality vs a config-less vault |
| `test_lexicon_output_is_invariant_to_the_audience_weighting` | the deliberate non-threading decision |
| `test_cli_rhetorical_patterns_...` / `test_mcp_rhetorical_patterns_...` | both move tallies pinned by value |
| `test_cli_fingerprint_...` / `test_mcp_fingerprint_...` | `fragment_count` pinned 1 vs 2 |
| `test_mcp_voice_report_reads_the_vault_config` | MCP profile; also pins that MCP writes no register samples (#1204) |
| `test_voice_authenticity_reports_the_vaults_weighting_state` | `ON`/`OFF` and the JSON payload |
| `test_fill_voice_step_honours_the_vault_audience_weighting` | `creek fill`'s unattended `report/voice` step |

Three previously **vacuous** assertions were de-vacuumed rather than left green: `tests/test_voice_authenticity.py` (x2) and `tests/e2e/test_voice_authenticity_epic_e2e.py` asserted `weighting_active is True` on a config-less tmp vault, which passed both before and after the fix. Each is now parametrized over both values, and each new `False` case fails against the old code.

Two vacuous assertions were caught *during* development and are worth naming, because both are the kind that ship silently:

- an inequality over the rhetorical-patterns note, which embeds `datetime.now()` — so the two files always differed and the assertion passed unconditionally. It survived a mutation deleting the very keyword it guarded. Now parsed into an integer tally and pinned by value.
- an inequality over the fingerprint JSON on a corpus where every feature extractor returns identical values, and where the shipped authority map has no *reachable* zero on that path. Now uses a `personal: 0.0` map and pins `fragment_count`.

Plus a structural guard in `tests/test_mcp_report_tier_ceiling.py` — `_states_override`/`_guarded_calls` generalised to take a keyword name (the existing override guard is semantically untouched and still passes), with a **separate** `_AUDIENCE_WEIGHTING_CALL_NAMES` set. `creek.generate.voice` is in its module list because `VoiceExemplarCollector` is constructed only there; without that entry, deleting the forward inside `generate_register_samples` leaves the guard green.

### Mutation matrix

Each revert applied alone; the named test must go red and the code must otherwise look fine.

| # | Mutation | Test that went red |
|---|---|---|
| M1 | `voice.py` `generate_register_samples` drops the forward to its collector | ordering, membership, prune, custom-map, divergent-config, fill (6) |
| M2 | `cli.py` drops `audience_weighting=` on `generate_register_samples` | same 6 |
| M3 | `cli.py` `_report_voice` sources from **bare `load_config()`** | 7 tests — and the run logs "Config file creek_config.yaml not found; running with built-in defaults", which is exactly the fake-green this suite exists to catch |
| M4 | `voice.py` generator reads a fabricated default instead of `self._collector.audience_weighting` | profile, CLI + MCP rhetorical, MCP voice |
| M5 | `cli.py` `_report_voice` drops it on `VoiceProfileGenerator` | profile passages |
| M6 | `cli.py` `_report_rhetorical_patterns` drops it | CLI rhetorical tally |
| M7 | `cli.py` `_report_fingerprint` back to bare `load_config()` | CLI fingerprint (`2 == 1`) |
| M8 | `report.py` `_generate_voice` drops it | MCP voice |
| M9 | `report.py` `_generate_rhetorical_patterns` drops it | MCP rhetorical tally |
| M10 | `report.py` `_generate_fingerprint` back to bare `load_config()` | MCP fingerprint (`2 == 1`) |
| M11 | `voice_authenticity.py` back to a fabricated default | `voice-authenticity[False-OFF]` **plus all three de-vacuumed assertions** |
| M12 | `cli.py` takes `ai_style` from a bare `load_config()` again | fingerprint lands at the wrong path |
| M13 | MCP resolves config eagerly for a config-less type | `report_type="tags"` raises the predicted `ValidationError` |
| MG1 | M1 applied — does the *guard* see it? | RED, naming `('VoiceExemplarCollector', 2603)` |
| MG2 | drop `creek.generate.voice` from the guard's module list | RED — the module list earns its place |

## What the pre-push review changed

A security pass and a test pass both ran over the diff. One blocker, fixed:

**The docstrings claimed a guarantee the guard does not make.** `fingerprint.py` and `voice.py` said "every production caller now states the vault's config explicitly under the structural guard". False for `build_fingerprint` — the symbol is not in `_AUDIENCE_WEIGHTING_CALL_NAMES`, and `cli.py`'s `_resolve_voice_fingerprint` still takes the `None` default, where `None` means *off*. A docstring promising a guarantee no test enforces is worse than no docstring. Both now name exactly which symbols the guard covers, and name #1410 for the one it does not.

Then two coverage gaps in this PR's own change, and three tests that could not have failed:

| Found | Fix |
|---|---|
| the `ai_style` half of the fingerprint change was threaded but untested — a mutant reverting only it survived | pinned via a non-default `fingerprint_path` (mutation M12) |
| `_vault_config`'s laziness was asserted in its docstring and enforced by nothing | `report_type="tags"` over a config with an unknown key (mutation M13) |
| the no-widening privacy test made only non-membership claims, and `_persisted_stems` returns `set()` for a missing directory — an outage satisfied it | positive control added |
| the lexicon tripwire compared two corpora differing only in near-identical bodies, so its sensitivity was hypothetical | `z-open` now carries borrowed glossary terms, which `_build_borrowed_terms` copies into `glossary.md` verbatim — confirmed present by execution |
| the parametrized voice-authenticity rows discriminated against a bare-`load_config()` regression only because the ambient cwd happens to hold no config | they now `chdir` |

Two accuracy corrections to claims I had over-stated, both in privacy-relevant prose:

- the docs named only `Register-Samples/` as the verbatim surface. Rendered profiles embed up to 10 **full exemplar bodies**, and that is the *only* surface the MCP voice tool writes — so an MCP-only operator was reading a warning that appeared not to apply to them.
- "`intimate: 0.0` keeps intimate content out of the fingerprint" is true but bites only when a caller passes `include_intimate=True`; on a default run it is a redundant second gate. Now stated precisely, including the surprising corollary that the shipped `0.0` silently overrides an operator who opts intimate content in.

## Follow-ups filed

- **#1409** — the same bare-`load_config()`-instead-of-vault-config defect for **other** config sections, at three more sites in `creek_mcp/tools/report.py` and three in `creek_mcp/server.py`. Each carries an inline comment pointing at it.
- **#1410** — `voice-check`'s fallback `_resolve_voice_fingerprint` calls `build_fingerprint` with no `audience_weighting=`. Pre-existing, not a regression, and outside this issue's inventory: reached only on a vault that has never run `report --type fingerprint`, and `voice-check` writes no vault artifact.
- **#1412** — the authority maps accept `NaN`/`inf`/negative values. A `NaN` makes the ranking sort key a non-total order, so the set of verbatim-copied bodies becomes arbitrary. Not a membership escalation (the gates still hold, and the other two consumers fail closed).
- **#1413** — both structural guards scan a hand-maintained module allowlist with no completeness check, so a guarded call in a *new* module would be invisible. Pre-existing for the `override` guard; this PR sharpened it by establishing that generator-internal modules belong in scope.

Closes #1313


Geoffe-Ga:
## Summary
Fixes issue #1313: `voice_audience_weighting` was defined in `CreekConfig`, seeded into every vault by `creek init`, and honoured on the AI-style *fingerprint* path — but every construction on the *exemplar* path (`VoiceProfileGenerator`, `VoiceExemplarCollector`, `generate_register_samples`) silently substituted a fresh, enabled default, so the vault's own setting never reached exemplar ranking. Two "already working" fingerprint call sites (`creek/cli.py::_report_fingerprint`, `creek_mcp/tools/report.py::_generate_fingerprint`) were also fixed: they read config via a bare `load_config()`, which resolves against the process cwd and never opens `<vault>/00-Creek-Meta/creek_config.yaml`, so `--vault X` silently ignored X's config. This PR threads the vault-scoped config through all real call sites, adds a structural AST guard preventing future regressions, and adds an extensive behavioural test module that runs from a config-less cwd (the condition that would have caught the original defect and the "looks-fixed-but-isn't" fingerprint variant).

## Strengths
- **Verified, not asserted, correctness.** I cross-checked the PR body's claims against the actual source (`creek/generate/voice.py`, `creek/generate/ai_style/fingerprint.py`, `creek/cli.py`, `creek_mcp/tools/report.py`) — every described call site, rename, and semantic distinction (e.g. `0.0` as membership gate on the fingerprint path vs. de-rank-only on the exemplar path) matches the code exactly.
- **Single storage location fix.** `VoiceProfileGenerator` no longer keeps a second, potentially-divergent copy of `audience_weighting`; `_rank_exemplars`/`_stream_into` now read `self._collector.audience_weighting`, eliminating a latent contradiction reachable only once `collector=` injection is used.
- **Structural regression guard.** `test_production_voice_callers_always_state_an_audience_weighting` in `tests/test_mcp_report_tier_ceiling.py` AST-scans the three producing modules and fails if any call site omits the keyword — directly modeled on the original defect shape (5 of 6 sites wired, 1 silently forgotten). It also asserts the guard's own name set isn't stale (`test_every_audience_weighted_symbol_is_actually_called_somewhere`).
- **Rigorous, well-reasoned test suite.** 18 new tests in `tests/test_voice_audience_weighting_wiring.py`, all deliberately run from a config-less cwd with `CREEK_CONFIG` unset — the exact condition that would let a cwd-scoped "fix" pass. Assertions target set membership (not just ordering) for the privacy-relevant register-samples path, verbatim-copy pruning, and precedence between vault config and cwd config.
- **Honest scope discipline.** The PR explicitly declines to thread the keyword into `generate_lexicon` with a three-step proof it would be a no-op, backed by a real invariance test (`test_lexicon_output_is_invariant_to_the_audience_weighting`) rather than a guard exclusion — a genuinely careful call.
- **Privacy analysis done properly.** The PR body proves the knob can't widen corpus membership (upstream `within_ceiling` + `_eligible_register` gates), and separately documents that disabling it *can* change which fragment survives the top-N cut, affecting what's copied verbatim into the vault — this is reflected in both code comments and `docs/configuration.md`.
- **De-vacuumed prior tests.** Two previously-vacuous assertions (a `datetime.now()`-based inequality that always passed, and a fingerprint inequality with no reachable zero) were identified and fixed to be genuinely load-bearing — a level of test-suite hygiene beyond what the issue asked for.
- **Documentation is thorough and precise**, including the `0.0` asymmetry, the two-maps-vs-four-maps table, and an upgrade note confirming byte-identical defaults.

## Security Concerns
🟢 LOW — No new vulnerabilities. The privacy-adjacent behavior change (disabling the weighting can cause more borrowed material to be copied verbatim into `07-Voice/Register-Samples/`) is an existing, operator-directed capability that is now correctly documented rather than newly introduced. Membership/consent gates (`within_ceiling`, `_eligible_register`) are verified independent of this ranking knob and are asserted never to be bypassable even by an absurd `intimate: 10.0` authority (`test_weighting_can_never_widen_corpus_membership`).

## Problems
None found blocking merge.

## Code Quality
- Non-blocking: the inline comments in `creek/cli.py` and `creek_mcp/tools/report.py` are quite long relative to the code they annotate (e.g. the 16-line comment on `_vault_config`). This matches the repo's evident house style of heavily-justified changes, but a future reader may find some of the "why we didn't do X" reasoning better suited to the PR description than permanent inline comments (e.g. the full mutation-matrix-style justification embedded at `_report_voice`).
- Non-blocking: `creek_mcp/tools/report.py::_generate_paradox` and `_generate_synchronicity` retain the bare, cwd-scoped `load_config().embeddings` call, now flagged with a `# ...tracked in #1409` comment. This is a reasonable scope boundary (different config section, follow-up filed), not a defect in this PR.

## CLAUDE.md Adherence
✅ Coverage: PR reports 94.55% branch coverage and 9259 passing tests locally (`./scripts/check-all.sh` reported green per the PR body); I could not independently execute the test suite in this review sandbox (command execution was restricted), so this is taken on the author's stated evidence plus my own static verification of the test logic and code wiring, which is internally consistent and specific enough to be credible (exact line numbers, exact assertion values, named mutation reverts).
✅ Docstrings: every new/changed function carries Args/Returns/Raises sections consistent with the repo's strict docstring style; the interrogate ≥95% gate is very likely satisfied given the density of docstrings added.
✅ Typing: all new signatures are fully typed (`VoiceAudienceWeightingConfig | None`, etc.), consistent with MyPy strict mode.
✅ Patterns: follows the existing `resolve_config_path` + `load_config(..., warn_on_missing=...)` precedent already established in `creek_mcp/tools/draft.py`, rather than inventing a new resolution mechanism.
✅ Commits/branch: feature branch `issue/1313-...`, not a direct push to main.

## Testing
Insufficient testing is supposed to be blocking, but this PR is the opposite case: the new test module is unusually rigorous, with an explicit mutation matrix (11 named mutations + 2 guard mutations, each mapped to the specific test(s) that catch it) documented in the PR body and reflected in the actual test names and assertions I read. Tests assert exact values/sets rather than loose inequalities, and two previously-vacuous existing tests were fixed as part of this change. I was unable to execute `pytest` directly in this review environment to get a live pass/fail signal, but the test code itself, read in full, is internally coherent, uses fixtures correctly, and its assertions genuinely depend on the changed production code paths (confirmed by tracing the config flow through `voice.py`, `cli.py`, and `report.py`).

## Documentation
`docs/configuration.md` and `docs/generation.md` are updated with a formula table distinguishing which report types read `platform_authority`/`audience_authority` vs. only the two original maps, a clear explanation of the `0.0` gate-vs-derank asymmetry, an explicit "this is not a privacy control" callout, and an upgrade note confirming the shipped defaults are a no-op for untouched vaults. Docstrings throughout the changed modules (`voice.py`, `fingerprint.py`, `voice_authenticity.py`, `lexicon.py`) were updated in lockstep with the code and accurately describe current behavior — no stale documentation found during review.



## Verdict: LGTM

The fix is correctly and completely wired (verified by reading the actual source, not just the diff): all five documented production call sites (VoiceProfileGenerator ×2 in cli.py, VoiceProfileGenerator ×2 in report.py, generate_register_samples, VoiceExemplarCollector's default construction) now thread the vault-scoped config, no stray references to the removed private attribute remain, the deliberate lexicon exclusion is backed by a real invariance test, and the new structural guard/test suite is well-targeted with a documented mutation matrix. No blocking issues found; only a couple of minor non-blocking observations.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5257834413 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Geoffe-Ga:
## Summary
This PR wires the vault's `voice_audience_weighting` config into the voice-exemplar ranking path, which had silently ignored it since the epic (#632/#633/#634) shipped — every production construction of `VoiceProfileGenerator`/`VoiceExemplarCollector`/`generate_register_samples` fell back to a fabricated *enabled* default, and `creek voice-authenticity` reported `ON` regardless of the vault's setting. It also fixes two "already-working" sites (`_report_fingerprint` in `cli.py` and `creek_mcp/tools/report.py`) that used a bare, cwd-scoped `load_config()` instead of the vault-scoped resolution, and adds a large, well-targeted test module plus doc updates.

## Strengths
- Root-caused correctly: traced the defect to *two* distinct bugs (fabricated collector default + cwd-scoped config resolution) and fixed both, rather than papering over symptoms.
- `VoiceExemplarCollector._audience_weighting` → public `audience_weighting`, with `VoiceProfileGenerator` reading it back off `self._collector` instead of keeping a second, potentially-divergent copy — a real single-source-of-truth fix (`creek-tools/creek/generate/voice.py:2191`, `:2280`, `:2508`).
- The new structural guard (`test_production_voice_callers_always_state_an_audience_weighting`) plus a completeness check that the guarded names are actually called (`test_every_audience_weighted_symbol_is_actually_called_somewhere`) closes the exact hole that let the original bug ship silently.
- Deliberate, well-argued scope boundaries: `generate_lexicon` is proven (not asserted) to be unaffected by the weighting and is guarded by a behavioural invariance tripwire instead of a no-op kwarg; `build_voice_authenticity_report` resolves the config internally rather than exposing a forgettable keyword.
- Every new test runs from a config-less cwd with `CREEK_CONFIG` unset — the discipline that actually distinguishes "reads the vault's config" from "reads *a* config, coincidentally."
- Privacy analysis is concrete and verified against the code (three-hop proof that the knob is ranking-only and cannot widen corpus membership), and the docs are corrected to state the `0.0` gate-vs-derank asymmetry precisely rather than round it off.
- The PR body documents its own pre-push review catching and fixing an overstated docstring guarantee and two coverage gaps (the `ai_style` half of the fingerprint fix, `_vault_config`'s laziness) — evidence of genuine self-verification, not just narration.
- Four legitimately out-of-scope defects found along the way were filed as follow-ups (#1409, #1410, #1412, #1413) rather than either scope-crept into this PR or silently dropped.

## Security Concerns
🟢 LOW — No new attack surface. Config is read via `load_config`/`resolve_config_path`, both pre-existing and pydantic-validated (`CreekConfig` forbids extra keys). The MCP `_vault_config` helper is documented as inheriting the pre-existing `CREEK_CONFIG`-outranks-vault-file precedence rather than changing it, and the blast radius is explicitly bounded to ranking (never membership).

## Problems
None found. I traced the wiring end-to-end (`cli.py` → `voice.py` → `VoiceExemplarCollector`/`_rank_exemplars`/`_iter_register_patterns`; `report.py`'s `_vault_config` → the three voice generators; `voice_authenticity.py`'s `_probe_audience_mix`) and it is internally consistent — no site was left silently defaulting.

## Code Quality
- Minor, non-blocking: `_report_voice` and `_report_rhetorical_patterns` in `cli.py` each independently re-resolve the vault's config via `_load_config_for_vault`, on top of the resolution the `report` command already did for dispatch — up to three YAML loads per invocation. This is explicitly deliberate (documented in the `_load_config_for_vault` docstring, with `warn_on_missing=False` to avoid repeat warnings) and the files are tiny, so it's not worth blocking on.
- `test_voice_audience_weighting_wiring.py` is a large (1185-line) single file; splitting by concern (ranking/membership vs. CLI/MCP surface vs. authenticity) could aid future navigation, but the existing organization (helpers up top, grouped by call-site) is easy to follow as-is.

## CLAUDE.md Adherence
✅ Conventional Commits used correctly (`fix(voice): ...`, `test(voice): ...`).
✅ TDD workflow evident (new tests explicitly designed to fail against each described mutation).
✅ Docstrings are extensive and precise, including on the *reasoning* behind each design decision — exceeds the 95% interrogate bar in spirit.
✅ PR reports `./scripts/check-all.sh` exit 0 (12/12 checks, 9261 passed, 94.55% branch coverage) and integration/e2e lanes green; I was unable to independently re-run the suite in this environment (no `uv`/project venv available), so this is taken on the PR's self-report, which is unusually well-substantiated (mutation matrix, named tests per mutation).
✅ No scope creep — deviations from the linked issue's acceptance criteria are named and justified individually.

## Testing
Exceptionally thorough. Beyond ordinary coverage, the PR provides:
- A mutation matrix (13 named code mutations + 2 guard mutations) with the specific test(s) that catch each.
- Set-membership assertions (not just ordering) for the privacy-relevant top-N cut, including verbatim-copy pruning.
- A positive control added to the "cannot widen membership" test after noticing `_persisted_stems` returns `{}` for a missing directory (would otherwise let an outage pass silently).
- De-vacuumed three previously-vacuous assertions (one compared strings containing `datetime.now()`, guaranteed to differ regardless of the code under test) — a good catch, and the kind of thing that's easy to miss in review.
No test-coverage concerns; nothing here is blocking.

## Documentation
`docs/configuration.md` and `docs/generation.md` are updated with a formula/reads table, a precise statement of the `0.0` gate-vs-derank asymmetry (including *when* it actually bites), and an explicit "this is not a privacy control, but affects what gets duplicated verbatim" section. Docstrings across `voice.py`, `fingerprint.py`, `lexicon.py`, and `voice_authenticity.py` were corrected during the PR's own review pass to stop overstating what the structural guard covers (`build_fingerprint` is explicitly named as *not* covered, with #1410 tracking it). Documentation is accurate to the implementation, not aspirational.



## Verdict: LGTM

The wiring fix is correct and internally consistent end-to-end, is backed by an unusually rigorous test suite (including a mutation matrix and structural guards against regression), and the author's own pre-push review already caught and fixed the kind of issues a reviewer would normally flag (overstated docstrings, untested coverage gaps, vacuous assertions). No blocking bugs, security issues, or missing tests found; the only remarks are minor non-blocking style notes.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5257967779 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1414, opened 2026-08-11
feat(cli): warn (do not fail) when --vault points at something that is not a vault


Geoffe-Ga:
Split out of #1318, where the question came up and was deliberately answered "not here".

## Problem

`_resolve_vault` (`creek-tools/creek/cli.py:5922`) returns whatever `--vault` was given, unchanged:

```python
def _resolve_vault(vault: Path | None) -> Path:
    if vault is not None:
        return vault
    return _load_config_for_vault(vault).vault_path
```

Nothing anywhere checks that the resulting path *looks like a Creek vault*. Every report generator then scaffolds its own output folder on demand — the #1231 convention, and after #1318 that now includes `00-Creek-Meta/` from the Tag Garden. So a mistyped `--vault ~/Documnets/vault` does not fail; it quietly creates a new directory tree and writes a report into it, and the operator finds out later when the real vault turns out to be unchanged.

This is a genuine cost of the on-demand-creation convention, and the convention is the right one — the alternative (each generator refusing to create its own parent) is exactly the defect #1318 fixed. The check belongs one layer up, where it can be made once for all ~30 commands instead of 17 times.

## Why it was out of scope for #1318

Making `report --type tags` alone fail-closed would have replaced one sibling inconsistency with its mirror image: an operator whose `report --type wavelength` succeeded on a path would then get a refusal from `--type tags` on the same path. The folder name `00-Creek-Meta` is a hard-coded literal in the generator, not operator input, so the blast radius of any single generator is one known subdirectory of an already-resolved root — the risk is entirely in *which root* got resolved, which is this issue.

## Suggested shape

**Warn, do not fail.** A hard refusal breaks `creek init`-adjacent flows and any legitimate empty-vault bootstrap, and there is no reliable positive test for "this is a vault" that a first run can satisfy.

- In `_resolve_vault`, when an explicit `--vault` was supplied and the target has neither `00-Creek-Meta/creek_config.yaml` nor any of the numbered top-level folders (`01-Fragments/`, `02-Threads/`, …), print a warning naming the resolved absolute path and suggesting `creek init --vault <path>`.
- Keep it to one warning per invocation. `_load_config_for_vault` already prints "Config file creek_config.yaml not found" — and my #1318 sweep showed it printing *twice* for `report --type paradox` and `--type synchronicity`, so the duplicate-warning problem is real and worth folding in here.
- Say nothing when the path came from config rather than the flag: that path was already validated when the config was written.

[...]

PR #1415, opened 2026-08-11
fix(generate): create 00-Creek-Meta/ on demand for the Tag Garden

### LINKED ISSUE #1318: fix(generate): creek report --type tags crashes with FileNotFoundError when 00-Creek-Meta/ does not exist
**Severity:** High
**Remediation:** fix (one `mkdir`) + regression test

## Problem

`TagGardenGenerator.generate_garden` is the only one of seventeen report write sites that
does not create its target directory. On a vault without `00-Creek-Meta/`, `creek report
--type tags` dies with a raw `FileNotFoundError` traceback.

`creek/generate/tags.py:361-362`:
```python
output_path = self.vault_path / "00-Creek-Meta" / "Tag-Garden.md"
output_path.write_text(_build_note(front, body), encoding="utf-8")
```

The same module mkdirs for its *other* output (`tags.py:442`), and the sibling writing into
the *same folder* mkdirs too (`creek/generate/indexes.py:679`, `:714`). Across the report
family, **16 of 17 write sites** call `parent.mkdir(parents=True, exist_ok=True)` —
`wavelength.py:704, 823, 885, 1461`; `decisions.py:217, 271`; `paradox.py:588`;
`lexicon.py:605, 638`; `synchronicity.py:345`; `unnamed.py:387, 460`;
`indexes.py:602, 637, 679, 714, 756`. One does not.

Taxonomy family 0 (correctness) + family 5 (inconsistent error contract across siblings).

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** A vault directory containing only `01-Fragments/` with
one tagged fragment:

```
TagGardenGenerator(vault_path=v).generate_garden()
-> FileNotFoundError [Errno 2] '.../v/00-Creek-Meta/Tag-Garden.md'

IndexGenerator(vault_path=v).generate_temporal_index()
-> OK (created 00-Creek-Meta/ on demand)
```

**Signal 2 — the convention is documented, and this site was missed.**
`creek/generate/indexes.py:8-13` documents exactly this fix as issue **#1231**: *"a missing
`06-Frequencies/<F*>` made the whole command exit 0 having written nothing… Every write
target is created on demand."* `tags.py` was never brought into that convention. #1231 is
closed and scoped to `indexes.py`, so this is a distinct site, not a re-file.

The CLI handler at `creek/cli.py:3402` has no `try`/`except`, so the operator gets the raw
traceback.

## Scope

Covers the single missing `mkdir` at `creek/generate/tags.py:361`.

Explicitly NOT in scope: the unguarded/unbounded tag history JSON, the O(T^2) similar-tag
scan, or the `type: ignore` entries (#851) — all separate findings in the same module.

## Tasks

1. Add `output_path.parent.mkdir(parents=True, exist_ok=True)` before the write at
   `creek/generate/tags.py:361`.
2. Add a regression test that runs `generate_garden` against a vault with no
   `00-Creek-Meta/` directory and asserts the note is written.

## Acceptance Criteria

- [ ] `creek report --type tags` succeeds on a vault that has no `00-Creek-Meta/` directory.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/generate/tags.py` | Modify |
| `creek-tools/tests/test_tag_garden.py` | Modify |

Geoffe-Ga:
## Summary

`TagGardenGenerator.generate_garden` wrote `<vault>/00-Creek-Meta/Tag-Garden.md` without creating its parent directory, so on a vault that had never been scaffolded both `creek report --type tags` and the MCP `creek.report` tool (`creek_mcp/tools/report.py:175`) died with a raw `FileNotFoundError` traceback. This adds the one missing `mkdir`.

**Reproduced before fixing.** Vault containing only `01-Fragments/` with one tagged fragment:

```
TagGardenGenerator(vault_path=v).generate_garden()
  File "creek/generate/tags.py", line 362, in generate_garden
    output_path.write_text(_build_note(front, body), encoding="utf-8")
FileNotFoundError: [Errno 2] No such file or directory: '.../00-Creek-Meta/Tag-Garden.md'

IndexGenerator(vault_path=v).generate_temporal_index()
  -> OK; created 00-Creek-Meta/ on demand
```

The sibling writing into *the same folder* succeeds, which is the whole shape of the bug.

**The census was re-derived at HEAD rather than taken from the issue.** All 32 `.write_text(` sites in `creek/generate/*.py` were enumerated and each checked for a guard. Exactly one is unguarded: `tags.py:362`. Four sites look unguarded on a naive scan and are not — `indexes.py:559` is guarded by `_ensure_frequency_subdir` at `:756`, `voice.py:1004`/`:1061` by `register_dir.mkdir` at `:888`, `wavelength.py:1493` at `:1461` — and the rest (`decisions.py:266`/`:676`, `drafts.py:2266`/`:2327`, `state.py:1708`) rewrite files that already exist. An empirical sweep running all eleven `REPORT_TYPES` against bare vaults confirms `tags` is the only one that crashes. So this is a genuine single-site oversight, not a fix that closes one of several.

The issue's own line numbers were ~15-20 lines stale (it cites `:361-362`; the write is at `:362` at this HEAD and ends up at `:372` after the docstring change, with the new guard on `:371`), and its "16 of 17" enumeration lists 17 `mkdir` lines rather than 17 write sites. The count is directionally right and the conclusion holds.

### Two design calls, made deliberately

**The guard goes at the call site, not in a shared write helper.** This looks like a case for pushing the guard down to a primitive so an eighteenth site inherits it, and the repo has such a primitive — but `creek/_fsio.py:166-170` already records the opposite decision in writing, for `atomic_write_text` (#1312):

> Deliberately does not `mkdir`: … the parent directory must already exist. … conjuring a directory tree here would silently absorb a caller's wrong path.

`creek/generate/*` does not route through that helper at all, and ten sibling modules inline this exact two-line idiom. A private helper inside `tags.py` alone would remove the asymmetry within one module by creating a new one against its ten siblings. The durable guard against a future eighteenth site is the behavioural test below, not an abstraction nothing forces callers through.

**No vault-shape validation is added here.** `mkdir(parents=True)` on a mistyped `--vault` will happily build a tree, and nothing upstream checks: `_resolve_vault` (`creek/cli.py:5922`) returns the path unchanged. Adding a refusal to *this* generator alone would replace one sibling inconsistency with its mirror image — an operator whose `report --type wavelength` succeeded on a path would then be refused by `--type tags` on the same path. The folder name is a hard-coded literal, not operator input, so the blast radius of any one generator is one known subdirectory of an already-resolved root; the risk lives in which root got resolved. Filed as #1414 (P3) against `_resolve_vault`, where one check serves all ~30 commands.

## Test plan

`tests/test_tag_garden.py` gains a `bare_vault` fixture (only `01-Fragments/` — the existing `vault` fixture pre-creates `00-Creek-Meta/` *and* `Processing-Log/`, which is precisely why twenty-odd `generate_garden` tests passed over a live bug) and `TestGenerateGardenOnAnUnscaffoldedVault`:

- `test_writes_the_garden_when_00_creek_meta_is_absent` — asserts the folder's absence *before* the call so the test cannot rot into a duplicate of the happy path, then asserts the note is written.
- `test_history_is_persisted_when_00_creek_meta_is_absent` — `_save_history` runs *after* the crash site and creates its own parent, so it never got the chance. This pins the whole method completing, not just the patched line.
- `test_an_existing_creek_meta_folder_and_its_contents_survive` — runs the generator twice against a pre-existing `00-Creek-Meta/` holding a sentinel file, asserting exact bytes. Forbids a guard that omits `exist_ok` or that recreates the directory.
- `test_the_only_directory_created_is_under_the_vault_root` — pins the blast radius: the vault's parent gains no entry, and the vault's directory inventory is exactly `01-Fragments`, `00-Creek-Meta`, `00-Creek-Meta/Processing-Log`. A "fix" that called a full vault-scaffolder would fail this.

**Mutation-tested, both directions.**

- Reverting the `mkdir` line alone: exactly the 3 new bare-vault tests go red with the real `FileNotFoundError` from `pathlib.write_text`; the other 268 tests in `test_tag_garden.py` + `test_cli.py` + `test_mcp_report_tier_ceiling.py` stay green.
- Dropping `exist_ok=True`: 16 tests go red on `FileExistsError`, including the dedicated no-clobber test.

**Gate 2:** `./scripts/check-all.sh` exit code `0` — 12/12 checks, 9265 passed, 5 skipped, 94.56% coverage (ruff runs `--no-cache`).

**Anti-bypass sweep** over the full diff including tests: no `noqa`, `type: ignore`, `pylint: disable`, `mark.skip`, `xfail`, or `--no-verify`; no threshold touched in `pyproject.toml` or `scripts/`; no test or code deleted (`git diff --numstat` = `+10/-0` and `+153/-0`).

### Process note

The `chief-architect` pass was independent and is what settled the call-site-versus-helper question. Two agents ended up writing `tests/test_tag_garden.py` concurrently after the first appeared to hang; the collision was caught, the duplicate class removed, and the surviving block is a merge of both. Verified afterwards that zero pre-existing lines were deleted or reworded and that all 37 original tests are still present and passing.

Closes #1318

### DIFF
--- a/creek-tools/creek/generate/tags.py
+++ b/creek-tools/creek/generate/tags.py
@@ -338,6 +338,15 @@ class TagGardenGenerator:
         and suggestion generation, then writes the output to
         ``00-Creek-Meta/Tag-Garden.md`` and persists history.
 
+        ``00-Creek-Meta/`` is created on demand (#1318). This was the one
+        report write site that assumed its folder already existed, so on a
+        vault that had never been scaffolded ``creek report --type tags``
+        died with a raw ``FileNotFoundError`` while every sibling generator
+        succeeded — the #1231 convention, "every write target is created on
+        demand", had never been applied here. The folder name is a literal,
+        not operator input, so nothing beyond one known subdirectory of the
+        already-resolved vault is ever conjured.
+
         Returns:
             Path to the generated Tag-Garden.md file.
         """
@@ -359,6 +368,7 @@ class TagGardenGenerator:
         }
 
         output_path = self.vault_path / "00-Creek-Meta" / "Tag-Garden.md"
+        output_path.parent.mkdir(parents=True, exist_ok=True)
         output_path.write_text(_build_note(front, body), encoding="utf-8")
 
         self._save_history(scan)
--- a/creek-tools/tests/test_tag_garden.py
+++ b/creek-tools/tests/test_tag_garden.py
@@ -50,6 +50,23 @@ def generator(vault: Path) -> TagGardenGenerator:
     return TagGardenGenerator(vault)
 
 
+@pytest.fixture()
+def bare_vault(tmp_path: Path) -> Path:
+    """Create a vault holding ``01-Fragments/`` and nothing else.
+
+    This fixture deliberately omits ``00-Creek-Meta/`` — and with it
+    ``00-Creek-Meta/Processing-Log/``. That omission is the whole point: the
+    ``vault`` fixture above pre-creates both, so it cannot express the state a
+    real vault is in before anything has scaffolded the meta folder, which is
+    exactly the state in which ``generate_garden`` used to die on a missing
+    parent directory (issue #1318). ``01-Fragments/`` is present because
+    ``_write_fragment`` writes into it, and because a vault with no content at
+    all would make the assertions about the seeded tag vacuous.
+    """
+    (tmp_path / "01-Fragments").mkdir()
+    return tmp_path
+
+
 def _write_fragment(vault: Path, name: str, tags: list[str]) -> Path:
     """Write a mock fragment markdown file with frontmatter tags.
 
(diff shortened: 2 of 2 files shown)


Geoffe-Ga:
## Summary
Fixes a `FileNotFoundError` crash in `TagGardenGenerator.generate_garden` (`creek/generate/tags.py`) that occurred when writing `Tag-Garden.md` to a vault where `00-Creek-Meta/` had never been scaffolded. The fix is a single added line — `output_path.parent.mkdir(parents=True, exist_ok=True)` — plus a docstring note and four new tests covering the previously-unexercised bare-vault path.

## Strengths
- Minimal, surgical diff: one line of production code fixes the actual bug, nothing else is touched.
- The fix exactly matches the existing convention used by every sibling write site in `creek/generate/*.py` (spot-checked `indexes.py:602/637/679/714`, which use the identical `note_path.parent.mkdir(parents=True, exist_ok=True)` idiom immediately before `write_text`).
- Excellent regression coverage: a new `bare_vault` fixture (only `01-Fragments/`, deliberately omitting what the existing `vault` fixture pre-creates) exposes the defect, and four tests each pin a distinct property — the crash is fixed, `_save_history` still runs to completion (it sits *after* the old crash site), an existing `00-Creek-Meta/` and its unrelated contents survive a second run untouched (guards against a fix that drops `exist_ok=True` or recreates the dir), and the blast radius is exactly the two expected subdirectories with nothing created outside the vault root.
- The PR description's supporting claims were independently verified against the repo rather than taken on faith: `_fsio.py:166-170`'s "deliberately does not mkdir" rationale for `atomic_write_text` checks out verbatim, `indexes.py:559`'s write is indeed guarded upstream via `_ensure_frequency_subdir` (line 546→756), and `decisions.py:266`/`:676` are confirmed rewrites of files located via `_find_decision_by_id` rather than fresh writes. The audit backing "this is the one unguarded site" holds up.
- Docstring addition explains the *why* (a historical convention gap, not a design flaw) rather than restating the code, consistent with this repo's comment philosophy.
- Two design calls are explicitly deferred with reasoning rather than scope-crept into this PR: no shared write-helper (would create asymmetry with the `_fsio.py` precedent and the other 10 inline sites), and no vault-path validation (filed separately as #1414, since the literal subfolder name isn't attacker/operator-controlled input — the real risk surface is `_resolve_vault`, which affects ~30 commands, not just this one).

## Security Concerns
None. 🟢 The created path segment (`00-Creek-Meta`) is a hard-coded literal, not derived from user input, so `mkdir(parents=True)` here doesn't introduce a path-traversal or arbitrary-directory-creation vector beyond what already exists via the resolved vault root.

## Problems
None found blocking merge.

## Code Quality
- Nit (non-blocking): `tags.py` writes via plain `output_path.write_text(...)`, not the `atomic_write_text` helper documented in `_fsio.py` for crash-safe writes — pre-existing and out of scope for this fix, but worth a future look given the module explicitly discusses `_fsio.py`'s conventions in its own docstring now.

## CLAUDE.md Adherence
✅ Follows the project's "fix root causes, no shortcuts" principle — real mkdir fix, no suppression or workaround.
✅ Docstring updated per the ≥95% interrogate / Google-style docstring standard, includes an issue reference (#1318) per the repo's forbidden-pattern rules for inline references.
✅ Test-first shape is evident (tests target the exact previously-uncovered code path, with an explicit "asserts absence before the call" anti-regression pattern).
Could not independently run `./scripts/check-all.sh`, `pytest`, ruff, or mypy in this sandbox (tool execution was restricted to read-only inspection this session), so CI-gate-level verification wasn't possible here — but static review shows the change is consistent with formatting/complexity/style conventions used elsewhere in the same file.

## Testing
Strong. Four new tests target exactly the four properties that matter for this class of fix: does it stop crashing, does the whole method now complete (not just the patched line), is it idempotent/non-destructive, and is the blast radius contained. This is well beyond a token regression test.

## Documentation
Docstring update to `generate_garden` is clear, accurate, and appropriately scoped (explains the historical convention gap without restating the one-line fix). PR description itself is unusually rigorous — includes a reproduction, a full write-site audit, and explicit rationale for the two design calls it declines to make.



## Verdict: LGTM

The fix is a correct, minimal, single-line mkdir guard that exactly matches the established pattern used by sibling generators in the same package (e.g. indexes.py:602,637,679,714), verified directly against the referenced source files. Test coverage is thorough (crash repro, post-crash-site history persistence, idempotency/non-destruction of an existing folder, and blast-radius containment), and the PR's factual claims about sibling write sites and the _fsio.py precedent were independently spot-checked and hold up.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5258258556 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1417, opened 2026-08-11
refactor(generate): one read-back idempotency scan, not three that drift — and never a shared filename builder


Geoffe-Ga:
**Found by:** the #1320 lane. Filed as a deliberate follow-up rather than done inline, so the fix for #1320 stays a bug fix.

## Problem

After #1320 there are **three** near-identical read-back idempotency scans in `creek/generate/`, each written independently and each subtly different:

| Function | File:line | Key | Except tuple |
|---|---|---|---|
| `_existing_synchronicity_pairs` | `synchronicity.py:454-475` | `frozenset` of `fragments:` frontmatter | `(OSError, ValueError)` — **incomplete, see #1416** |
| `_existing_decision_fragment_ids` | `decisions.py:1003-1037` | first `## Source Fragments` bullet | `(OSError, ValueError, yaml.YAMLError)` |
| `_paradox_notes_by_pair` / `_recorded_pair` | `paradox.py` (added by #1320) | `frozenset` of `fragments:` frontmatter | `(OSError, ValueError, yaml.YAMLError)` |

Each has its own `is_dir()` guard, its own `glob("*.md")`, its own bracket-stripping, and its own skip-on-error policy. #1416 exists precisely because one of the three drifted. The paradox one additionally returns a `dict[key, list[Path]]` so the same pass yields a duplicate census — a capability the other two would want and do not have.

Their call sites are also structurally identical (`already = …` / `if key in already: continue` / `already.add(key)`), see `synchronicity.py:443-451`, `decisions.py:1080-1086`, and `generate_paradoxes`.

## Scope — read the constraint before designing

Extract **the read-back/skip scan only**. Do **not** extract a shared filename builder.

That boundary is not stylistic. #1334 asks for a shared filename disambiguator and its surviving remedy is `VaultWriter._atomic_create`'s counter-suffix retry — which never overwrites, so a paradox generator adopting it would mint `a-b-2.md`, `a-b-3.md`, … on every run, institutionalising #1320 instead of fixing it. Separately, #1334 must not fold `source_id` into a filename, because for a compost project candidate the `source_id` *is* the tag, which can be intimate-derived — the exact string #1404 stopped leaking into the filesystem. A "shared disambiguation helper" is the shape that would reintroduce it.

So: the identity **key** differs per generator by design (pair set vs source fragment id vs title), and only the *mechanism* around it is shared.

## Sketch

`creek/generate/_idempotency.py`, something like:

```python
[...]

Issue #1416, opened 2026-08-11
fix(generate): a malformed synchronicity note crashes the read-back scan the docstring says skips it


Geoffe-Ga:
**Found by:** the #1320 lane, while mirroring the sibling read-back scan into `creek/generate/paradox.py`.

## Problem

`_existing_synchronicity_pairs` (`creek-tools/creek/generate/synchronicity.py:466-469`) guards its frontmatter read with:

```python
        try:
            post = frontmatter.loads(note.read_text(encoding="utf-8"))
        except (OSError, ValueError):
            continue
```

`frontmatter.loads` raises `yaml.parser.ParserError` on malformed YAML. That class inherits `yaml.YAMLError` → `Exception` — it is **not** a `ValueError` and **not** an `OSError`, so the guard does not catch it and one hand-broken note in `10-Liminal/Synchronicities/` takes down the whole command.

The docstring says unreadable notes are skipped. They are not.

`decisions.py:1026` gets this right (`except (OSError, ValueError, yaml.YAMLError)`), and `paradox.py`'s new `_recorded_pair` (added in #1320) follows `decisions.py`.

## Reproduction (run at `ca1b9c3`)

```python
from creek.generate.synchronicity import _existing_synchronicity_pairs
d = vault / "10-Liminal" / "Synchronicities"; d.mkdir(parents=True)
(d / "sync-bad.md").write_text("---\nfragments: [a, b\ntype: synchronicity\n---\nbody")
_existing_synchronicity_pairs(vault)
```

```
CRASH: ParserError while parsing a flow sequence
```

Type check on the same input:

```
(<class 'yaml.parser.ParserError'>, <class 'yaml.error.MarkedYAMLError'>,
 <class 'yaml.error.YAMLError'>, <class 'Exception'>, ...)
is ValueError: False | is OSError: False | is YAMLError: True
```

## Impact

`creek report --type synchronicity` (`creek/cli.py:_report_synchronicity`) and the MCP `report` tool both call `generate_synchronicities`, which calls this function on every run. `10-Liminal/` is an operator-editable Obsidian folder — a stray keystroke in one note makes the command unrunnable, with a raw `ParserError` traceback rather than the documented skip.

## Tasks

1. Add `yaml.YAMLError` to the except tuple at `synchronicity.py:468`, matching `decisions.py:1026` and `paradox.py`'s `_recorded_pair`.
[...]

PR #1404, opened 2026-08-11
fix(compost): screen withheld fragments at the detection boundary, once (#1311)

Geoffe-Ga:
## Summary

`skip_intimate` / `skip_paradox` guarded only one of `CompostTracker`'s three
detection paths, so dormant-thread and disappeared-project compost notes copied
intimate fragments' titles and `[[ids]]` straight into the operator's vault.

**The fix the issue body proposes does not close the leak, and the body's own
acceptance criterion passes under it.** I implemented Tasks 2 and 3 exactly as
written — filter `fragment_ids` and `energy_excerpts` in `_detect_threads` and
`_project_candidate` — on a scratch copy of HEAD and ran the real
`run_compost_scan` over a vault whose only two fragments are
`privacy_tier: intimate` and both tagged `affair-recovery`. Output:

```
FILES WRITTEN: ['2026-04-01-affair-recovery.md', '_Compost-Report.md']
title: affair-recovery
original_project: affair-recovery
reason: Project silent for 700 days
## What it was
A project titled **affair-recovery** that has since composted back into the vault.
## Related Fragments
_No related fragments were recorded._
--- _Compost-Report.md ---
- [[2026-04-01-affair-recovery|affair-recovery]]

body's own AC -- no intimate fragment id or title appears: True
but the intimate-only tag names a file in the vault: True
```

A project candidate's `source_id` and `title` **are the tag**, and that tag
reaches the filename (`compost.py` builds it from `candidate.title`), the
frontmatter, the body, and the rollup report. Filtering the emitted *lists*
cannot suppress an identity that only intimate fragments brought into
existence. That project exists solely because two intimate fragments carry
that tag.

So the guard goes at the boundary instead, once:

- **`_AdmittedFragments`** — a frozen dataclass holding the screened fragment
  tuple plus `withheld_thread_titles: frozenset[str]`. `CompostTracker._screen()`
  runs at the top of `detect_compost_candidates`, and all three detectors are
  retyped to take it, so an unscreened detector is a mypy `arg-type` error
  rather than a privacy leak. The now-redundant per-path guard is deleted.
  Titles rather than `Fragment` objects deliberately: the thread detector is the
  one proven to copy `frag.title`/`frag.id` into a note, and precomputing the
  set once is O(withheld) instead of O(threads x withheld) regex re-parses on a
  35k-fragment vault.
- **Silent omission.** A thread named only by withheld fragments yields no
  candidate. A note reading `_No related fragments were recorded._` announces
[...]

PR #1418, opened 2026-08-11
fix(generate): key paradox notes on the fragment pair, not the calendar day (#1320)

### LINKED ISSUE #1320: fix(generate): paradox notes duplicate on every new calendar day despite the docstring's idempotency claim
**Severity:** High
**Remediation:** fix (key the note on the fragment pair, or add the sibling read-back scan)

## Problem

Paradox notes are **not** idempotent across calendar days, though two docstrings say they are.
Running `creek report --type paradox` weekly for a year leaves 52 copies of every paradox in
`10-Liminal/Paradoxes/`.

- `creek/generate/paradox.py:263` — `detected_date: date = field(default_factory=date.today)`
- `creek/generate/paradox.py:744-750` — `_build_paradox` constructs `Paradox(...)` and never
  passes `detected_date`, so the factory fires on every run.
- `creek/generate/paradox.py:798-804` — `_filename` embeds
  `paradox.detected_date.isoformat()` in the note name.
- `creek/generate/paradox.py:815-817` — *"**Idempotent** — each note's path is stable per
  paradox, so re-running overwrites in place rather than duplicating."* Same claim on
  `create_paradox_note` at `:576-578`.

The path is stable per paradox *per day*, which is not the same thing.

The two sibling generators both enforce idempotency by reading the output folder back:
`synchronicity.py:454-475` (`_existing_synchronicity_pairs`) and `decisions.py:1003-1037`
(`_existing_decision_fragment_ids`). Paradox has no such scan and relies purely on the
date-stamped filename.

Taxonomy family 0 (correctness) + family 3 (aspirational claim) + family 5 (divergent
sibling contract).

## Evidence (corroboration)

```
day 1: wrote 2026-08-10-frag-1-frag-2.md
day 2: wrote 2026-08-11-frag-1-frag-2.md
ON DISK for ONE unchanged fragment pair:
  ['2026-08-10-frag-1-frag-2.md', '2026-08-11-frag-1-frag-2.md']
```

Recipe: a vault with two fragments sharing `frequency.primary: F3`, one
`wavelength.dosage: medicine` and one `toxic` (fires the dosage rule). Call
`generate_paradoxes(vault, EmbeddingsConfig())`; re-run `create_paradox_note` on the same
detector output with `detected_date` advanced one day — exactly what `date.today()` returns
on the next run.

**Signal 2 — structural, against the sibling contract.** `_build_paradox` is the only
construction site on the production path and omits `detected_date`; the two sibling
generators in the same package both carry a read-back idempotency scan that paradox lacks.

Note the trigger is also timezone-sensitive: `date.today()` is system-local, so on a non-UTC
host the day rolls at local midnight (see the related `wavelength.py` `date.today()` sites).

## Scope

Covers idempotency of paradox note creation in `creek/generate/paradox.py`.

Explicitly NOT in scope: `_is_paradox` reading `tags` instead of `emotional_texture` (#1210),
or the `date.today()` sites in `wavelength.py`.

## Tasks

1. Choose and implement one:
   - **(a)** Key `_filename` (`paradox.py:798-804`) on the fragment-pair only, dropping the
     date from the name — matching how `synchronicity` keys on the pair set; or
   - **(b)** Add an `_existing_paradox_pairs` read-back scan mirroring
     `synchronicity._existing_synchronicity_pairs` (`synchronicity.py:454-475`), and skip
     pairs already on disk.
   (a) is simpler and makes the stated "stable per paradox" claim literally true.
2. Correct the two docstrings at `paradox.py:815-817` and `:576-578` to describe the
   behaviour that actually ships.
3. Add a regression test: generate, advance the clock a day, generate again, assert exactly
   one note exists for the pair.

## Acceptance Criteria

- [ ] Re-running paradox generation on an unchanged vault on a later date produces no new
      file for an already-recorded pair.
- [ ] The idempotency docstrings match the implementation.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/generate/paradox.py` | Modify |
| `creek-tools/tests/test_paradox.py` | Modify |

Geoffe-Ga:
## Summary

Paradox notes are named `<YYYY-MM-DD>-<fragA>-<fragB>.md`. `Paradox.detected_date` is `field(default_factory=date.today)` and `_build_paradox` never overrides it, so every run stamps today and the path is stable per paradox *per day*. Running `creek report --type paradox` weekly for a year left 52 copies of every paradox in `10-Liminal/Paradoxes/`.

Reproduced at `ca1b9c3`, one unchanged fragment pair, real listing:

```
day 1 (2026-08-10): wrote 2026-08-10-frag-1-frag-2.md
day 2 (2026-08-11): wrote 2026-08-11-frag-1-frag-2.md
ON DISK for ONE unchanged fragment pair:
  ['2026-08-10-frag-1-frag-2.md', '2026-08-11-frag-1-frag-2.md']
```

All four line cites in the issue are exact at `ca1b9c3`. One half of the issue's claim is overstated, though: `generate_paradoxes:815-817` ("re-running overwrites in place rather than duplicating") is flatly false, but `create_paradox_note:576-578` ("writing the same `Paradox` twice produces the same file path") is *literally true* — same object, same date — just badly misleading beside the first. Both are rewritten; only one was a lie.

`generate_paradoxes` now reads `10-Liminal/Paradoxes/` back, keys each note on `frozenset(fragments[:2])` from its frontmatter, and **skips** a pair already recorded. That is the shape both siblings already use — `synchronicity._existing_synchronicity_pairs` and `decisions._existing_decision_fragment_ids` — so this closes a divergent-sibling-contract gap, not just a bug.

### Why the read-back scan and not the issue's preferred remedy

The issue rated remedy (a) — drop the date from the filename — "simpler". It is a silent-data-loss regression, and three independent facts say so:

1. **It clobbers the operator's writing.** `create_paradox_note` ends in an unconditional `write_text`. Make the path stable and every weekly run overwrites the note in place — destroying whatever reflection the operator wrote under `REFLECTION_PROMPT`. That prompt *is* the product; the note exists to be annotated. Trading 52 harmless copies for silent loss of the operator's own prose is the worse trade, and it is exactly the class this repo has declined three times. This is pinned by `test_the_operators_reflection_survives_the_next_run`, which goes red under a mutation implementing (a) with the byte diff showing the prose gone.
2. **It breaks `[[wikilinks]]` with no fallback.** Nothing in production writes a link *at* a paradox note, but `vault/links.py:174` registers every note's stem in the link index, and paradox notes emit no `title:` and no `aliases:` — the stem is the only handle a hand-written `[[2026-04-12-frag-a-frag-b]]` can use. Every other date-prefixed generator (threads, eddies) puts the human-readable name in `aliases:`; paradox is the outlier that never got it. Renaming without an alias breaks the link silently.
3. **It does not converge an existing vault.** With identity living only in the filename, an existing `2026-04-12-a-b.md` is unrecognisable to a new scheme, so the next run writes *one more* file before stabilising.

Two further options were considered and rejected: recovering the recorded `detected_date` and rewriting at the original path (re-introduces the clobber), and deriving `detected_date` from fragment timestamps (makes a field named `detected_date` lie, and still adds a file to every existing vault).

The filename is therefore unchanged. That also protects the state report's Liminal Watch (`generate/state.py:942`), which renders the paradox file stem verbatim as operator-facing text and uses `p.name` as its sort tiebreak.

### Migration — every existing vault already holds duplicates

Fixing the generator stops the growth; it cannot undo what is on disk. New public `duplicate_paradox_warning(vault_path) -> str | None` **detects and reports**: the count, the number of distinct pairs, and a sample-capped list of paths. It is delivered through a new keyword-only `on_warning` callback on `generate_paradoxes` plus `logger.warning`, and `creek report --type paradox` prints it.

**Nothing is deleted.** Which copy to keep is a judgement about the operator's own writing — one of them may carry their reflection — and a generator does not get to make it on a heuristic. This is the fourth instance of the shape #1304, #1305 and #1329 each established; the funnel is lifted from `creek/ingest/pipeline.py`'s. The advisory is self-clearing (remove the strays and the run goes quiet), silent on a clean vault, and names **paths only** — no fragment title, excerpt, or tag, per #1404.

`creek report --type paradox`'s empty-result line said "no contradictory fragment pairs found", which this change would newly turn into a lie (pairs exist; they are already recorded). It now covers both cases honestly. `_report_synchronicity` has the same pre-existing imprecision and is deliberately left alone.

### Relationship to #1334 — declined, with reasons (commented on the issue)

#1334 asks that its fix be "shaped so #1320 can adopt it". Same smell, opposite failure mode: #1320 over-produces harmlessly; #1334 *destroys* (the second same-titled candidate overwrites the first and the index oscillates `b -> a -> b` forever). #1320's identity key was already unique and already on disk, so a read-back scan sufficed; #1334's key is a title, which is not unique and needs a filename **disambiguator**. Proof they are disjoint: `decisions.py` already has a read-back scan and it does not help.

Flagged loudly on #1334: its Task 1's `source_id` option is already ruled out (for a compost project candidate the `source_id` *is* the tag, which can be intimate-derived — the string #1404 just stopped leaking into the filesystem), and its surviving option, `VaultWriter._atomic_create`'s counter-suffix retry, **never overwrites** — so a paradox generator "adopting" it would mint `a-b-2.md`, `a-b-3.md`, … every run, institutionalising this bug rather than fixing it. The only shareable abstraction across these generators is the read-back scan, never the filename builder. Filed as #1417 with that constraint written into its scope.

### Follow-ups filed

- **#1416** (P1, bug) — `_existing_synchronicity_pairs` catches `(OSError, ValueError)`, but `frontmatter.loads` raises `yaml.ParserError` on malformed YAML, which is neither. Reproduced: one hand-broken note takes down `creek report --type synchronicity`. `decisions.py:1026` gets it right and this PR's `_recorded_pair` follows `decisions.py`.
- **#1417** (P2, refactor) — extract the now-three near-identical read-back scans; read-back only, never a shared filename builder.

## Test plan

New `tests/test_paradox_idempotency.py` (22 tests, zero skipped) plus a strengthened `test_liminal_report_generators.py::TestGenerateParadoxes::test_idempotent` and three new CLI tests.

**Clock injection.** There is no freezegun/time_machine here, and patching `creek.generate.paradox.date` does **nothing** — `default_factory=date.today` captured the bound method at class-definition time. Verified empirically: with the module attribute patched to a 1999 fake, `Paradox().detected_date` still returned the real today. "Yesterday's run" is therefore staged by writing its note through the same public path with an explicit past `detected_date` — byte-for-byte the bytes that run left. A fixed 2020 date is used so no assertion can straddle midnight. `TestSecondRunSameDay` needs no clock manipulation at all.

**RED proven at `ca1b9c3`** before implementing — 4 behavioural failures, including the headline `assert 2 == 1` on files-on-disk.

**Mutation tested**, three mutants, each restored:

| Mutation | Result |
|---|---|
| Delete the `if pair in already: continue` guard | 8 red — 6 new, 1 strengthened, 1 new CLI. No pre-existing test noticed, which is why the bug survived. |
| Force the advisory to `None` | 8 red — exactly the advisory tests |
| The issue's remedy (a): dateless filename, no guard | 14 red, including the reflection test, whose byte diff shows the operator's prose replaced by the regenerated body |

Also covered: pair key is order-insensitive; a genuinely new pair is still written alongside an old one (blocks a lazy "folder non-empty → do nothing"); malformed YAML, a short `fragments:` list, an unreadable `*.md` directory entry, and a `creek save --target paradox` note (which records provenance under `saved_from.contributing_fragments`, not a top-level `fragments:`) are each skipped **while a legitimate paradox is still written in the same run** — so an over-broad `except` cannot pass; a missing `10-Liminal/Paradoxes/` folder; advisory counted once per pair, silent when clean, self-clearing, sample-capped, and carrying no fragment prose.

**Gate 2:** `./scripts/check-all.sh` exit code **0** — 9286 unit tests, ruff (`--no-cache`), mypy strict, pylint 9.71, bandit, pip-audit, refurb, tryceratops, interrogate, complexity, 90% aggregate + per-file coverage.

**Bypass sweep** over the whole diff including tests: no `# noqa`, `# type: ignore`, `# pylint: disable`, `@pytest.mark.skip/xfail`, or `--no-verify` added; `pyproject.toml` and `scripts/` untouched. The one refurb finding this change produced (`FURB179`) was fixed at the root with `itertools.chain.from_iterable`, not suppressed.

Closes #1320

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -455,29 +455,38 @@ def _parse_since_arg(text: str) -> datetime:
     return parsed if parsed.tzinfo is not None else parsed.replace(tzinfo=UTC)
 
 
-def _print_ingest_warning(message: str) -> None:
-    """Show one ingest advisory to the operator, as it is detected (#1329).
+def _print_advisory(message: str) -> None:
+    """Show one operator advisory on the console, as it is detected (#1329).
 
-    Warnings go to the same stdout console as every other ``creek ingest``
-    advisory (``_warn_if_discovered_but_empty``, ``_print_pin_findings``)
-    rather than to stderr: this command's output is human prose, not data on a
-    pipe, and an advisory routed to a stream none of its neighbours use is one
-    an operator can lose.
+    Advisories go to the same stdout console as every other advisory the
+    surrounding command prints, rather than to stderr: these commands' output
+    is human prose, not data on a pipe, and an advisory routed to a stream
+    none of its neighbours use is one an operator can lose.
 
-    The exit code is deliberately left alone. A warning reports vault state
+    The exit code is deliberately left alone. An advisory reports vault state
     that will cause trouble later, not a failure of this run — and
-    ``creek sync`` shares this path, so escalating would turn every scheduled
-    pass over an un-migrated vault into a hard failure.
+    ``creek sync`` shares the ingest path, so escalating would turn every
+    scheduled pass over an un-migrated vault into a hard failure.
 
     ``escape`` and ``soft_wrap=True`` are load-bearing for the same reasons as
-    in :func:`_print_pin_findings`: the text carries a shell command the
-    operator is meant to copy, and rich would otherwise be free to wrap it
-    across lines (or read a bracketed path as markup).
+    in :func:`_print_pin_findings`: the text carries paths and commands the
+    operator is meant to read or copy, and rich would otherwise be free to
+    wrap them across lines (or read a bracketed path as markup — a paradox
+    advisory names ``[[wikilink]]`` literally).
+
+    Args:
+        message: The advisory text from the pipeline that detected it.
+    """
+    console.print(f"[yellow]{escape(message)}[/yellow]", soft_wrap=True)
+
+
+def _print_ingest_warning(message: str) -> None:
+    """Show one ingest advisory to the operator, as it is detected (#1329).
 
     Args:
         message: The advisory text from the ingest pipeline.
     """
-    console.print(f"[yellow]{escape(message)}[/yellow]", soft_wrap=True)
+    _print_advisory(message)
 
 
 def _run_ingest(
@@ -3882,7 +3891,13 @@ def _report_paradox(vault_path: Path, override: PrivacyTierOverride) -> None:
 
     Wires the implemented ``ParadoxDetector`` to a runnable command: scans the
     vault for contradictory pairs and writes one neutral note per paradox into
-    ``10-Liminal/Paradoxes/``. Idempotent; deterministic (no LLM).
+    ``10-Liminal/Paradoxes/``. Deterministic (no LLM), and idempotent because
+    the generator skips pairs an existing note already records (#1320) — so a
+    re-run on a later day writes nothing rather than a second copy.
+
+    Any pre-#1320 duplicate copies already in the vault are reported through
+    ``on_warning`` as they are detected, before the empty/success line, and are
+    never deleted.
 
     Args:
         vault_path: Vault root.
@@ -3898,11 +3913,20 @@ def _report_paradox(vault_path: Path, override: PrivacyTierOverride) -> None:
     from creek.generate.paradox import generate_paradoxes
 
     config = _load_config_for_vault(vault_path)
-    written = generate_paradoxes(vault_path, config.embeddings)
+    written = generate_paradoxes(
+        vault_path,
+        config.embeddings,
+        on_warning=_print_advisory,
+    )
     if not written:
+        # Deliberately covers both empty cases: no contradictory pair exists,
+        # and every pair that exists is already recorded. `generate_paradoxes`
+        # returns `[]` for both and the difference does not change what the
+        # operator should do — the notes they want are in the folder either way.
         console.print(
-            "[yellow]No paradox notes generated: "
-            "no contradictory fragment pairs found.[/yellow]",
+            "[yellow]No paradox notes generated: no contradictory fragment "
+            "pairs beyond those already recorded in "
+            "10-Liminal/Paradoxes/.[/yellow]",
         )
         return
     console.print(
--- a/creek-tools/creek/generate/paradox.py
+++ b/creek-tools/creek/generate/paradox.py
@@ -25,6 +25,7 @@ deliberately neutral: no resolution, prescription, or judgment.
 from __future__ import annotations
 
 import itertools
+import logging
 import math
 import re
 from dataclasses import dataclass, field
@@ -32,16 +33,19 @@ from datetime import date
 from typing import TYPE_CHECKING
 
 import frontmatter
+import yaml
 
 from creek.link.neighbours import cosine_neighbours
 from creek.models import Confidence, Dosage, Frequency, Phase
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable
+    from collections.abc import Callable, Iterable
     from pathlib import Path
 
     from creek.config import EmbeddingsConfig
     from creek.models import Fragment
+
+logger = logging.getLogger(__name__)
 
 
 CONTRADICTION_KEYWORDS: tuple[str, ...] = (
@@ -131,6 +135,47 @@ _CONTRADICTION_DESCRIPTIONS: dict[str, str] = {
 
 
 _FILENAME_SANITIZER = re.compile(r"[^\w\-]+")
+
+
+_PARADOX_SUBPATH: tuple[str, str] = ("10-Liminal", "Paradoxes")
+"""Vault-relative folder paradox notes are written to and read back from."""
+
+
+_PAIR_SIZE: int = 2
+"""Fragments per paradox — the arity both the filename and the pair key use."""
+
+
+_WIKILINK_BRACKETS = re.compile(r"[\[\]]")
+"""Strips ``[[…]]`` from a frontmatter fragment reference.
+
+:meth:`ParadoxDetector.create_paradox_note` writes bare ids, but the sibling
+Synchronicity notes write wikilinks and an operator may normalise a paradox
+note by hand. Stripping costs nothing and makes the pair key tolerant of both,
+exactly as ``synchronicity._existing_synchronicity_pairs`` is.
+"""
+
+
+_DUPLICATE_SAMPLE: int = 3
+"""How many redundant note paths the pre-#1320 advisory names before stopping."""
+
+
+_DUPLICATE_WARNING_TEMPLATE: str = (
+    "This vault holds {count} paradox notes recording only {pairs} distinct "
+    "fragment pair(s). Before #1320 the note filename embedded the detection "
+    "date, so every re-run on a new calendar day wrote a fresh copy of an "
+    "unchanged paradox. This run wrote no further copy. The redundant notes "
+    "are left exactly where they are: {sample}. Nothing is deleted "
+    "automatically — open them in your vault and keep whichever one carries "
+    "the reflection you wrote. Any [[wikilink]] pointing at one of them keeps "
+    "resolving until you act."
+)
+"""Advisory for a vault still holding pre-#1320 per-calendar-day copies.
+
+Names paths and nothing else. A paradox note's excerpts are fragment prose and
+its tags can be intimate-derived, so neither belongs in a console line (#1404);
+the filename is already on disk and already surfaced by the state report's
+Liminal Watch, so naming it discloses nothing new.
+"""
 
 
 def _cosine_similarity(a: list[float], b: list[float]) -> float:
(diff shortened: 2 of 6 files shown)


Geoffe-Ga:
## Summary
Fixes #1320: `Paradox.detected_date` defaults to `date.today()` and is never overridden, so the note filename (which embeds that date) was stable only per calendar day — a weekly `creek report --type paradox` run left one duplicate copy of every unchanged paradox per week. `generate_paradoxes` now reads `10-Liminal/Paradoxes/` back, keys each note on its `fragments:` frontmatter pair, and skips a pair already recorded. It also adds a non-destructive `duplicate_paradox_warning`/`on_warning` advisory channel to surface pre-existing duplicate copies without deleting anything.

## Strengths
- The chosen fix (read-back scan keyed on content, not filename) mirrors the existing `synchronicity._existing_synchronicity_pairs` / `decisions._existing_decision_fragment_ids` shape exactly — closes a divergent-sibling-contract gap rather than inventing a new pattern.
- The PR explicitly considered and rejected the "simpler" fix (dropping the date from the filename), with concrete reasoning: it would silently clobber operator-written reflections (`create_paradox_note`'s unconditional `write_text`), break hand-written `[[wikilink]]`s (no `title:`/`aliases:` fallback), and not converge existing vaults. This reasoning is verified correct by reading `create_paradox_note` and `_filename`.
- `_recorded_pair` catches `(OSError, ValueError, yaml.YAMLError)` — I confirmed `yaml.YAMLError` does *not* subclass `ValueError`, so this is a real, necessary broadening versus the sibling scan in `synchronicity.py` (which the PR correctly flags as a separate bug, filed as #1416).
- Migration story is sound: nothing is deleted, the advisory is self-clearing, silent on a clean vault, sample-capped, and names paths only (no fragment prose/tags per #1404) — verified against the template string and tests.
- Docstrings are unusually thorough and explain *why*, not just *what* (e.g. `create_paradox_note`'s "This method clobbers" callout, `_filename`'s rationale for keeping the date prefix).
- Test suite is excellent: covers cross-day dedup, same-day dedup, pair-order independence, a genuinely-new pair still being written, malformed YAML, an unreadable directory entry masquerading as `*.md`, a `creek save --target paradox` note's differently-shaped frontmatter, a missing folder, and the full advisory lifecycle (reported, capped, self-clearing, silent-when-clean, never proposes deletion).
- Backward compatible: `on_warning` is a new keyword-only parameter defaulting to `None`, so the MCP caller (`creek_mcp/tools/report.py:346`) keeps working unchanged (still gets the advisory via `logger.warning`).

## Security Concerns
None. No new file writes outside the existing `10-Liminal/Paradoxes/` folder, no new external input parsing beyond what `frontmatter.loads` already did, and the advisory text is explicitly scoped to exclude fragment prose/tags (#1404 constraint), which the tests pin (`test_advisory_names_no_fragment_prose`). 🟢 LOW / none.

## Problems
None found blocking merge.

## Code Quality
Two minor, non-blocking observations:
- `duplicate_paradox_warning` is a public function but currently has no production caller — `generate_paradoxes` computes the advisory itself via the internal `_advisory_for` helper (deliberately, to avoid a second folder scan, per its own docstring). Right now it's exercised only by tests. That's a reasonable "read-only entry point for future use" choice as documented, but worth a follow-up if there's no near-term plan to wire it into something like `creek doctor`.
- The MCP `_generate_paradox` tool (`creek_mcp/tools/report.py:346`) doesn't pass `on_warning`, so an agent driving the vault over MCP never sees the duplicate-copies advisory in its tool response — only the server-side log gets it. This is explicitly acknowledged in the PR description as a deliberate scope cut, but it's worth flagging since MCP is often the primary interface for agent-driven vault maintenance.

Neither blocks merge.

## CLAUDE.md Adherence
✅ TDD workflow followed (RED proven before implementation per PR description). ✅ Docstrings are Google-style and thorough (well above the 95% interrogate floor). ✅ Type hints throughout, `Callable` imported correctly under `TYPE_CHECKING`. ✅ No `# noqa`/`# type: ignore`/skip markers added (per the PR's own bypass sweep, and confirmed no such markers appear in the diff). ✅ Follows DRY — the read-back-scan pattern is reused/mirrored rather than duplicated ad hoc, and the two related read-back scans are flagged for a proper extraction in a filed follow-up (#1417) rather than done ad hoc here. ✅ Commit message and PR description are exceptionally detailed on rationale. No `scripts/` or `pyproject.toml` changes.

## Testing
Strong. `tests/test_paradox_idempotency.py` (466 new lines) covers the headline regression, same-day/cross-day behavior, pair-order insensitivity, folder-scan robustness against three distinct malformed-input shapes, and the full duplicate-advisory lifecycle. `test_liminal_report_generators.py::test_idempotent` was correctly updated (the old assertion "re-running writes the same note in place" was only true within a single day and is now honestly described as superseded). New CLI-level tests (`test_cli.py`) pin the end-to-end behavior through `creek report --type paradox`, including the empty-result message wording and duplicate-reporting output. No blocking gaps found in the diff.

## Documentation
`docs/emergence.md` is updated with a clear, accurate paragraph on the idempotency behavior and the duplicate-note advisory. Docstrings across the changed module are extensive and explain design rationale, not just mechanics.



## Verdict: LGTM

The fix is correct (verified the read-back-scan key logic, the exception handling in `_recorded_pair` including confirming `yaml.YAMLError` is not a `ValueError` subclass so the broader except clause is actually necessary, and the CLI/MCP call sites), matches the established sibling pattern in synchronicity.py/decisions.py, and ships with extensive, well-reasoned tests (22 new tests plus updated existing ones) covering the regression, robustness of the folder scan, and the migration advisory. No bugs, security issues, or CLAUDE.md violations found; only two non-blocking observations (noted in Code Quality) that don't warrant blocking merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5258843152 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1419, opened 2026-08-11
docs(generation): the report table points `unnamed` at a path the code has never written


Geoffe-Ga:
`docs/generation.md`'s report-type table documents the `unnamed` report as writing to:

```
10-Liminal/Unnamed-Digest-<period>.md
```

The code has never written there. `UnnamedDigestGenerator._write_digest`
(`creek-tools/creek/generate/unnamed.py:386-391`) writes to:

```
10-Liminal/Unnamed/Digests/<iso-year>-W<week>.md
```

i.e. a `Digests/` subfolder *inside* `10-Liminal/Unnamed/`, named by ISO year and
ISO week — not a flat file under `10-Liminal/` named by an unspecified `<period>`.

The module docstring at `creek-tools/creek/generate/unnamed.py:15-17` and
`creek-tools/docs/emergence.md:21` both state the real path correctly, so
`docs/generation.md:21` is the lone outlier.

## Why it matters

The table is the surface a user consults to find out where `creek report --type
unnamed` put its output. Pointing at a path that does not exist sends them
looking in the wrong folder, and it is the one row in that table a reader has no
way to falsify without reading the source.

Note the `Digests/` subfolder is not incidental: `load_unnamed_fragments` skips it
by name (`_DIGESTS_SUBFOLDER`) so digests are not re-ingested as fragments. A
reader who trusts the docs would not know that.

## Fix

One-line correction to the `unnamed` row of the table in
`creek-tools/docs/generation.md:21`.

## Provenance

Found while reading `creek/generate/unnamed.py` for #1321 (the digest's quadratic
re-parse). Deliberately not fixed there — unrelated to that change's diff.

PR #1420, opened 2026-08-11
perf(generate): parse the Unnamed tree once per digest, not once per fragment (20,500 -> 200 at N=200)

### LINKED ISSUE #1321: perf(generate): the Unnamed digest re-parses the whole Unnamed tree once per fragment (20,500 parses at N=200)
**Labels:** `de-slop`, `creek-pipeline`, `performance`, `priority-high`
**Severity:** High
**Remediation:** fix (return `(path, fragment, body)` from the single existing walk)

## Problem

The Unnamed weekly digest re-walks and re-parses the entire `10-Liminal/Unnamed/` tree once
per fragment. Parse count grows as `N + N(N+1)/2 + N`.

`load_unnamed_fragments` (`creek/generate/unnamed.py:214-237`) already walks
`sorted(unnamed_dir.rglob("*.md"))` and calls `frontmatter.load` on every file — then
**discards both `md_file` and `post.content`**, returning only `Fragment` models. Those two
discarded values are exactly what the next two passes reconstruct quadratically:

- `_build_excerpt_map` (`unnamed.py:665-682`) loops the week's fragments, calling
  `_find_fragment_file` per fragment;
- `_find_fragment_file` (`unnamed.py:610-629`) does a full `rglob("*.md")` +
  `frontmatter.load` on every file until the id matches;
- `_excerpt_from_file` (`unnamed.py:588-607`) does a **third** `frontmatter.load` of the file
  just found, to read the body pass #1 already had in `post.content`.

Reached from `creek report --type unnamed` (`creek/cli.py:2718`) and the MCP path
(`creek_mcp/tools/report.py:_generate_unnamed`) via `_render_body` (`unnamed.py:357`).

Taxonomy family 5 (per-stage vault re-reads) + family 7 (needless work).

## Evidence (corroboration)

**Signal 1 — measured artifact.** Instrumented `frontmatter.load` counter over
`generate_weekly_digest`:

```
N=  50 fragments ->    1,375 frontmatter.load calls (0.36s)
N= 100 fragments ->    5,250 frontmatter.load calls (1.00s)
N= 200 fragments ->   20,500 frontmatter.load calls (5.00s)
```

Exactly `N + N(N+1)/2 + N`; clean 4x per doubling. Extrapolating, 2,000 unnamed fragments is
~2.0M parses and ~8 minutes for one weekly digest.

This is a measurement, not a "could be faster" claim.

**Signal 2 — structural.** The single-pass shape already exists in the repo:
`creek.vault.reader.iter_vault_fragments` returns `(path, fragment, body, raw)` for
`01-Fragments`, which is precisely the tuple `load_unnamed_fragments` throws away.

## Scope

Covers the excerpt-resolution path in `creek/generate/unnamed.py`.

Explicitly NOT in scope: the corrupt-history / invalid-fragment / excerpt-boundary coverage
gaps (#871), or the narrow `frontmatter.load` guards at `unnamed.py:541/601/626` (#924).

## Tasks

1. Change `load_unnamed_fragments` (`unnamed.py:214-237`) to return
   `(path, fragment, body)` tuples, mirroring `creek.vault.reader.iter_vault_fragments`.
2. Thread those through `_render_body` (`unnamed.py:357`) into `_build_excerpt_map`
   (`unnamed.py:665-682`), which then needs no lookup at all.
3. Delete `_find_fragment_file` (`unnamed.py:610-629`) and reduce `_excerpt_from_file`
   (`unnamed.py:588-607`) to a pure string-truncation helper over the already-loaded body.
4. Add a test that counts `frontmatter.load` invocations and asserts one parse per file.

## Acceptance Criteria

- [ ] `generate_weekly_digest` parses each `10-Liminal/Unnamed/` file exactly once.
- [ ] The parse-count test fails before the fix.
- [ ] Digest output is byte-identical to pre-change output for the same vault.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/generate/unnamed.py` | Modify |
| `creek-tools/tests/test_unnamed_digest.py` | Modify |

Geoffe-Ga:
## Summary

The Unnamed weekly digest walked and re-parsed the entire `10-Liminal/Unnamed/`
tree three times over, once more per fragment. `load_unnamed_fragments` already
parsed every file and then **discarded `post.content`** — which is exactly what
the next two passes rebuilt quadratically: `_build_excerpt_map` called
`_find_fragment_file` per fragment (a full rescan each time), and
`_excerpt_from_file` then re-loaded the file that scan had just opened, to read a
body pass #1 had already held.

The fix keeps the body from the walk that was happening anyway.

### Measured: before and after

Parse count is the primary evidence because it is exact and machine-independent.
Instrumented `frontmatter.load` around `generate_weekly_digest`, all N fragments
inside the target week:

| N | before | after | wall clock before → after |
|---|--------|-------|---------------------------|
| 25 | 375 | **25** | 0.39s → 0.04s |
| 50 | 1,375 | **50** | 0.18s → 0.01s |
| 100 | 5,250 | **100** | 0.70s → 0.02s |
| 200 | 20,500 | **200** | 2.70s → 0.03s |

The before column reproduces the issue's `N + N(N+1)/2 + N` to the digit
(the issue reported 1,375 / 5,250 / 20,500). After: exactly **N** — one parse per
unnamed markdown file for the whole digest run. Quadratic → linear; ~90x wall
clock at N=200, and the gap widens with N. Extrapolated at 2,000 unnamed
fragments: ~2.0M parses → 2,000.

### Output equivalence

Generated a digest before and after on the same fixture vaults and diffed the
bytes (`generated:` stamp normalised, since it is wall-clock by design).

- Clean vaults at N=25/50/100/200: **byte-identical**, all four.
- One adversarial fixture (nested subdirs, empty body, >200-char body, multi-line
  body, `Digests/` file, duplicate ids, a non-fragment stray): **exactly one line
  differs**, and it is a bug being fixed — see below. Nothing else moved,
  including digest section order and the history JSON.

## Behaviour changes — both deliberate, both tested

**1. A non-fragment file can no longer supply a fragment's excerpt.** This is the
one byte difference. `_find_fragment_file` matched on raw frontmatter `id` across
*every* markdown file, including ones that are not `type: fragment` and ones that
fail `Fragment.model_validate`. On the fixture, a stray:

```
---
type: not-a-fragment
id: frag-empty
---
STRAY BODY.
```

caused the digest to render, under the *real* `frag-empty` fragment's heading:

```
### The Real Fragment
- Link: [[frag-empty]]
- Excerpt: STRAY BODY.
```

An unrelated file's body published under a fragment's title. Excerpts now resolve
only from files that validate as fragments, so that line correctly disappears.
Pinned by `test_non_fragment_file_never_supplies_a_fragment_excerpt`, which is RED
at HEAD.

**2. Duplicate ids resolve deterministically.** `_find_fragment_file` returned the
first **unsorted** `rglob` match, so which of two same-id files supplied the
excerpt depended on filesystem iteration order. The map is now built with
`excerpts.setdefault(...)` during the `sorted()` walk: first-wins, the
deterministic successor to "first match wins". Plain assignment would be
last-wins — the same arbitrary pick aimed the other way. Pinned by
`test_duplicate_ids_resolve_to_the_sorted_first_file`.

## Shape: why not the signature change the issue proposed

The issue proposed returning `(path, fragment, body)` from `load_unnamed_fragments`.
Not done, for two reasons:

- **It moves a declared public API.** `load_unnamed_fragments` sits under this
  module's `# ---- Public API ----` banner. Its signature, name, and contract are
  unchanged here, and its five tests are byte-untouched — which is the cheapest
  available proof that this is a pure performance fix with no API drift.
- **It retains the wrong thing.** That method loads *all* unnamed fragments (the
  history record's `total_unnamed` depends on it), not just the week's, so a list
  of `(path, fragment, body)` holds every unnamed body live at once.
  `10-Liminal/Unnamed/` is by design the vault's accumulating residue — the digest
  exists to report that it grows. Trading a CPU quadratic for an unbounded memory
  retention on the same call path is not a fix.

Instead:

- `_iter_unnamed_records(vault_path) -> Iterator[tuple[Fragment, str]]` — a **lazy
  generator**, the single definition of the Unnamed walk (directory guard,
  `sorted(rglob("*.md"))`, `Digests/` skip). No `path` and no `raw` in the tuple, a
  deliberate deviation from `creek.vault.reader.iter_vault_fragments`: nothing
  downstream consumes them once `_find_fragment_file` is gone.
- `_load_unnamed_with_excerpts` consumes it once, building the fragment list and
  the id→excerpt map in the same loop. Only the ≤200-char excerpts are retained;
  bodies fall out of scope as the generator advances. At 35k fragments that is
  ~7MB of excerpts instead of >100MB of bodies.
- `load_unnamed_fragments` becomes a projection over the same generator.
- `_load_fragment` now returns `tuple[Fragment, str] | None` — the shape
  `creek.generate.skills._load_fragment` already uses for the same job in this
  package.
- `_excerpt_from_file` → pure `_excerpt_from_body(body: str) -> str`;
  `_find_fragment_file` and `_build_excerpt_map` deleted.

One honest cost, stated plainly: excerpts are now built for every unnamed fragment
rather than just the week's, because `filter_fragments_in_week` runs *after* the
load — restricting the map would mean either duplicating that tested predicate or
holding all bodies until it runs. The old `_build_excerpt_map` was week-scoped, so
this is a small new cost the previous code did not pay: bounded at
`_EXCERPT_MAX_CHARS` (200) per out-of-week fragment, no extra parses, and the
entries are simply never looked up. It is dwarfed by what it replaces.

### Other callers

`load_unnamed_fragments`: one production caller (`generate_weekly_digest`) plus
five tests, all unchanged. `_find_fragment_file`, `_build_excerpt_map` and
`_excerpt_from_file` had no callers outside this module. Both entry points —
`creek report --type unnamed` (`creek/cli.py:_report_unnamed`) and MCP
`_generate_unnamed` (`creek_mcp/tools/report.py`) — reach this only through
`generate_weekly_digest`, whose signature is unchanged, so neither file is touched.

## Test plan

`./scripts/check-all.sh` → **exit 0**. 12/12 checks, 9298 passed, coverage 94.60%,
per-file coverage gate passed.

New tests, all in `tests/test_unnamed_digest.py`:

- `test_generate_weekly_digest_parses_each_unnamed_file_exactly_once[2,8]` — the
  durable regression guard. Spies on `frontmatter.load` and asserts one parse per
  file plus no file parsed twice. RED at HEAD: `7 == 2` and `52 == 8`.
- `test_iter_unnamed_records_parses_lazily_one_file_at_a_time` — draws a single
  record and asserts exactly one file was opened. Guards the memory property:
  collapsing the generator into a list is otherwise completely invisible.
- Excerpt-formatting equivalence net (GREEN before and after): 199 chars + `…`
  truncation, empty body emits no excerpt line, newlines collapse to spaces.
- The two behaviour-change tests described above.

**Mutation-tested** — each mutation kills exactly the intended test and nothing
else:

| mutation | result |
|---|---|
| revert to the multi-walk (i.e. HEAD) | parse-count test RED: `7 == 2`, `52 == 8` |
| add back a second walk | parse-count test RED: `4 == 2`, `16 == 8` |
| `setdefault` → plain assignment | duplicate-id test RED: `{'ZZZ BODY.'} == {'AAA BODY.'}` |
| lazy generator → eager list | laziness test RED — **and all 43 others stay green** |

That last row is the point of having the laziness test at all.

## Scope

Out of scope and deliberately untouched: the `yaml.YAMLError` gap in
`_load_fragment` (#924) — its `except (OSError, ValueError)` and its
`# noqa: TRY101` are verbatim from HEAD, and no new exception handling was added.
Note two of #924's three cited call sites (`unnamed.py:601` and `:626`) are deleted
outright by this change; only the guarded load inside `_load_fragment` survives, so
#924 wants rescoping. Commented there.

Also out of scope: the corrupt-history / excerpt-boundary coverage gaps (#871), and
a docs drift found while reading (`docs/generation.md` points the `unnamed` report
at a path the code has never written) — filed as #1419 rather than fixed here.

Closes #1321

### DIFF
--- a/creek-tools/creek/generate/unnamed.py
+++ b/creek-tools/creek/generate/unnamed.py
@@ -43,6 +43,7 @@ from creek.models import Fragment
 from creek.time import effective_authored_at, effective_authored_date
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
     from pathlib import Path
 
     from creek.link.embeddings import EmbeddingLinker
@@ -183,16 +184,16 @@ class UnnamedDigestGenerator:
         Returns:
             Path to the generated digest markdown file.
         """
-        all_unnamed = self.load_unnamed_fragments(vault_path)
+        all_unnamed, excerpts = _load_unnamed_with_excerpts(vault_path)
         this_week = self.filter_fragments_in_week(all_unnamed, week_start)
         clusters = self.detect_unnamed_clusters(this_week)
         previous = self._load_history(vault_path, week_start)
         body = self._render_body(
-            vault_path,
             this_week,
             clusters,
             week_start,
             previous=previous,
+            excerpts=excerpts,
         )
         digest_path = self._write_digest(vault_path, week_start, body)
         self._append_history(
@@ -223,18 +224,10 @@ class UnnamedDigestGenerator:
         Returns:
             List of Fragment objects parsed from the vault.
         """
-        unnamed_dir = vault_path.joinpath(*_UNNAMED_SUBPATH)
-        if not unnamed_dir.is_dir():
-            return []
-
-        fragments: list[Fragment] = []
-        for md_file in sorted(unnamed_dir.rglob("*.md")):
-            if _DIGESTS_SUBFOLDER in md_file.relative_to(unnamed_dir).parts:
-                continue
-            fragment = _load_fragment(md_file)
-            if fragment is not None:
-                fragments.append(fragment)
-        return fragments
+        # A projection of the shared walk: each body is bound for one loop
+        # iteration and discarded, so this costs no more memory than the
+        # per-file parse it always did.
+        return [fragment for fragment, _body in _iter_unnamed_records(vault_path)]
 
     @staticmethod
     def filter_fragments_in_week(
@@ -333,28 +326,31 @@ class UnnamedDigestGenerator:
 
     def _render_body(
         self,
-        vault_path: Path,
         fragments: list[Fragment],
         clusters: list[list[Fragment]],
         week_start: date,
         *,
         previous: _HistoryEntry | None,
+        excerpts: dict[str, str],
     ) -> str:
         """Render the complete markdown body for the digest.
 
         Args:
-            vault_path: Path to the root of the Obsidian vault.
             fragments: Fragments observed this week.
             clusters: Similarity clusters detected this week.
             week_start: Monday of the target ISO week.
             previous: Previous history entry, or ``None`` when no prior
                 run has been recorded.
+            excerpts: Mapping of fragment ID to a short body excerpt, built
+                by :func:`_load_unnamed_with_excerpts` during the single
+                pass over the Unnamed tree. It covers every unnamed
+                fragment, not just *fragments*; the extra entries are
+                simply never looked up.
 
         Returns:
             The markdown body (without frontmatter).
         """
         week_end = week_start + timedelta(days=_WEEK_LENGTH_DAYS - 1)
-        excerpts = _build_excerpt_map(vault_path, fragments)
         sections = [
             f"# Unnamed Digest — {week_start.isoformat()} to {week_end.isoformat()}\n",
             _render_fragments_section(fragments, excerpts),
@@ -527,15 +523,20 @@ def _upsert_history_entry(
     return merged
 
 
-def _load_fragment(md_file: Path) -> Fragment | None:
-    """Parse a markdown file into a :class:`Fragment`, or ``None`` on failure.
+def _load_fragment(md_file: Path) -> tuple[Fragment, str] | None:
+    """Parse a markdown file into ``(fragment, body)``, or ``None`` on failure.
+
+    Returns the markdown body alongside the validated model so a caller that
+    needs an excerpt already has it and never reopens the file. The shape is
+    the one :func:`creek.generate.skills._load_fragment` already sets for the
+    same job in this package.
 
     Args:
         md_file: Path to the markdown file.
 
     Returns:
-        The parsed Fragment, or ``None`` when the file is not a valid
-        fragment record.
+        ``(fragment, body)`` when the file is a valid fragment record, or
+        ``None`` when it is not.
     """
     try:
         post = frontmatter.load(str(md_file))
@@ -546,7 +547,7 @@ def _load_fragment(md_file: Path) -> Fragment | None:
     if metadata.get("type") != "fragment":
         return None
     try:  # noqa: TRY101  # Separate failure modes: file IO vs schema validation each return None with distinct debug logs.
-        return Fragment.model_validate(metadata)
+        return Fragment.model_validate(metadata), post.content
     except ValidationError:
         logger.debug("Skipping invalid fragment frontmatter: %s", md_file)
         return None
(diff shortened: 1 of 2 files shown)


Geoffe-Ga:
## Summary
Fixes a real quadratic-parse bug in `generate_weekly_digest`: the Unnamed tree was walked/parsed up to `N + N(N+1)/2 + N` times per digest run. The fix introduces a single lazy generator (`_iter_unnamed_records`) that yields `(fragment, body)` once per file, consumed once by `_load_unnamed_with_excerpts` to build both the fragment list and the id→excerpt map in the same pass. `_find_fragment_file` and `_build_excerpt_map` are deleted; `load_unnamed_fragments`'s public signature is preserved as a thin projection over the same generator.

## Strengths
- Correctly diagnoses and eliminates the quadratic: verified by reading the deleted `_find_fragment_file`/`_build_excerpt_map` and confirming they had no callers outside this module (`grep` confirms no other file references them).
- Keeps `load_unnamed_fragments`'s public contract unchanged rather than widening it to `(path, fragment, body)` as the linked issue proposed — a good call, since that would have forced holding every unnamed body live at once, trading a CPU quadratic for unbounded memory in a directory that is *by design* meant to grow unbounded.
- The laziness is genuinely load-bearing (not decorative): `_iter_unnamed_records` is a real generator, and `test_iter_unnamed_records_parses_lazily_one_file_at_a_time` pins that drawing one record parses exactly one file. A `list`-comprehension "cleanup" would silently reintroduce the memory failure mode while every other test stayed green — that's exactly the gap this test closes.
- Two behavior changes (non-fragment files can no longer donate an excerpt; duplicate IDs resolve to the sorted-first file deterministically) are called out explicitly, each pinned by its own test, and each is a strict correctness improvement over the prior filesystem-order-dependent / type-blind behavior.
- Verified `_excerpt_from_body` is a faithful extraction of the old `_excerpt_from_file`'s formatting logic (strip → collapse newlines → truncate at 200 chars with ellipsis) — behavior-preserving, not just structurally similar.
- Verified the "same shape as `creek.generate.skills._load_fragment`" claim against that file — accurate.

## Security Concerns
None found. No new I/O, subprocess, or user-input handling introduced; same `frontmatter.load`/`Fragment.model_validate` boundary as before.

## Problems
None found. I did not find a blocking correctness, security, or test-coverage issue in this diff.

## Code Quality
- 🟢 Minor: `excerpts.setdefault(fragment.id, _excerpt_from_body(body))` evaluates `_excerpt_from_body(body)` unconditionally even when the key already exists (Python evaluates call args before `setdefault` checks membership), so a duplicate-ID's excerpt is computed and discarded. Trivial cost (a string slice), not worth a special case.
- 🟢 Minor: the new `_iter_unnamed_records` docstring says `creek.vault.reader.iter_vault_fragments` "yields `(path, fragment, body, raw)`," but that function actually returns an eagerly-built `list`, not a generator — the tuple shape it describes is correct, but "yields" slightly overstates that sibling function's laziness. Cosmetic only.

## CLAUDE.md Adherence
✅ TDD workflow followed — PR body documents RED-at-HEAD tests for the parse-count and both behavior-change assertions, GREEN-before-and-after equivalence tests for excerpt formatting.
✅ `./scripts/check-all.sh` reported exit 0 (9298 passed, 94.60% coverage) per PR description — not independently re-run in this review environment (test/CI execution was unavailable here), but the diff is internally consistent with that claim on manual trace-through.
✅ No forbidden patterns (`noqa`/`type: ignore` additions, threshold changes, disabled tests).
✅ Commit message follows Conventional Commits (`perf(generate): ...`), closes the linked issue.
✅ Scope discipline: explicitly declines to touch the out-of-scope `yaml.YAMLError` gap (#924) and docs drift (filed as #1419) rather than scope-creeping the fix.

## Testing
Strong. New tests in `tests/test_unnamed_digest.py` cover: exact one-parse-per-file with a no-double-parse assertion (anchored to the actual file count, not a hardcoded number, avoiding vacuous pass); generator laziness via a call-count probe on a single `next()`; excerpt-formatting equivalence (truncation, empty body, multiline collapse); and both deliberate behavior changes with clear rationale for why each was RED at HEAD. The mutation table in the PR body (revert-to-multi-walk, partial-fix, setdefault→assignment, generator→list) each map to a specific test going red, which is good evidence the tests aren't just incidentally passing.

## Documentation
Docstrings are thorough and Google-style throughout the new/changed functions (`_iter_unnamed_records`, `_load_unnamed_with_excerpts`, `_excerpt_from_body`, `_load_fragment`), each explaining the *why* (laziness as a memory property, first-wins as the deterministic successor to the old accidental behavior) rather than restating the code. No doc drift found against the `creek.generate.skills._load_fragment` comparison it makes.



## Verdict: LGTM

The change correctly fixes a real quadratic-parse bug with a well-scoped, memory-conscious design (a genuinely lazy generator, verified by a test that would catch the tempting "list comprehension" regression); both incidental behavior changes are deliberate, disclosed, and individually tested; no dangling references to the deleted helpers exist elsewhere in the codebase. Only cosmetic nits were found (a redundant excerpt computation on duplicate IDs, a slightly loose docstring claim about a sibling function's laziness) — neither warrants blocking or even a must-fix comment.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5259259437 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1422, opened 2026-08-11
fix(classify): _is_accepted cannot distinguish an inherited wavelength phase from a derived one


Geoffe-Ga:
`_is_accepted` cannot tell a phase the fragment *inherited* from a phase the
classifier *derived*, so re-atomization accepts or rejects a child on evidence
that is not about that child.

## The defect

`creek/classify/reatomize.py:441`:

```python
return fragment.wavelength.phase != Phase.UNCLASSIFIED.value
```

A split child inherits its parent's whole wavelength block
(`creek/atomize/split.py:254-262`), so `phase` is very often already populated
before any classifier looks at the child's own text. `_is_accepted` reads that
inherited value as though the pass had just derived it, and the FEAT-023
recursion decision — accept this atom, or split it again — turns on it.

## How #1331 surfaced it

Before #1331 the rule pass rebuilt the whole wavelength block, so a child whose
keyword pass fired *mode but not phase* had its inherited phase wiped and was
rejected, while a child with **no** keyword hits at all kept its inherited phase
and was accepted. Those two cases disagreed for no defensible reason. #1331
made the pass merge, so they now agree — both keep the inherited phase, both are
accepted. That is a consistency improvement and it is pinned by
`tests/test_rules_preserves_evidence.py::TestReatomizeAcceptance`.

It does not fix the underlying question, which #1331 deliberately did not
answer: **should an inherited phase count as "classified" for acceptance at
all?** Arguably not — the acceptance bar is meant to ask whether *this* atom
has been understood, and inheriting the parent's phase is not evidence that it
has. But changing that flips recursion depth for real vaults, so it needs its
own decision rather than riding along on a merge fix.

## Bounded blast radius

`classification.reatomize` is opt-in and off by default
(`creek/config.py`), so no default run is affected.

## Acceptance criteria

- [ ] Decide, and write down, whether `_is_accepted` should require a phase this
      pass derived or accept an inherited one; the decision belongs in the
      `_is_accepted` docstring, which today explains the *choice of dimensions*
      but not the provenance question.
- [ ] If provenance must be tracked, it needs a mechanism — the child currently
      carries no marker distinguishing inherited classification from derived.
      Say what that mechanism is before implementing it.
- [ ] `frequency.primary` gets the same audit; it is the other half of the same
      `_is_accepted` conjunction and children inherit it the same way.
- [ ] Whatever is chosen, `TestReatomizeAcceptance` in
[...]

Issue #1421, opened 2026-08-11
fix(classify): _apply_wavelength rebuilds the wavelength block wholesale on the single-pick LLM path


Geoffe-Ga:
`_apply_wavelength` rebuilds `WavelengthClassification` from scratch on the
single-pick LLM path, so a persisted `descriptor` / `color` / `mode` /
`orientation` / `dosage` is destroyed whenever the model is silent or
below the FEAT-017 confidence floor.

## The defect

`creek/classify/llm/calibration.py:144` builds a whole
`WavelengthClassification(...)` from the parsed response and
`creek/classify/llm/orchestrator.py:304` assigns it wholesale via
`fragment.model_copy(update=updates)`. Every wavelength axis the model did not
speak to is therefore reset to its "not determined" default rather than left
standing.

Two ways it fires:

1. **The model omits a field.** `descriptor` has no gloss requirement, so a
   response without one blanks a descriptor an earlier run (or the operator)
   recorded.
2. **`_biased_enum` gates the pick.** FEAT-017 deliberately downgrades `mode`,
   `orientation` and `dosage` to `UNCLASSIFIED` when the model's self-reported
   confidence is below `unclassified_threshold`. The *intent* is "do not adopt
   a noisy pick". The *effect* is "erase what we already knew", which is a
   strictly stronger and unintended action.

## Why this is a separate issue from #1331

#1331 fixed the same wholesale-rebuild defect at the rules layer
(`RuleClassifier._build_updates`) and at `_apply_voice`, the two places where
the destroyed field feeds the INTIMATE privacy escalation. `classify_tier`
(`creek/classify/privacy.py:135-141`) reads only authorship, platform, recovery
keywords, `voice.voice_register` and `voice.confidence` — **no wavelength
field**. So this one is metadata loss, not a fail-open privacy path, and it
carries a design question #1331 deliberately did not answer on its own
authority: whether the FEAT-017 bias should mean "do not adopt" (merge, prior
stands) or keep meaning "erase" (replace).

## Suggested fix

Reuse `creek.classify.evidence.layer_determined_over`, the shared primitive
#1331 added and which `WeightedFragmentClassification.merge_onto`
(`creek/classify/weighted.py:475`) and `RuleClassifier.classify` already use.
Give `_apply_wavelength` the fragment's prior wavelength the way `_apply_praxis`
(#877), `_apply_texture` (#878) and `_apply_voice` (#1331) already take theirs.

[...]

PR #1423, opened 2026-08-11
fix(classify): merge classifier verdicts instead of rebuilding the block

### LINKED ISSUE #1331: fix(classify): RuleClassifier rebuilds whole sub-models, wiping wavelength orientation/dosage/color/descriptor and voice.confidence
**Severity:** High
**Remediation:** fix in place

## Problem

`RuleClassifier._build_updates` (`creek/classify/rules.py:751`) scores five
things — frequency, phase, mode, voice register, confidence — but writes back
**whole Pydantic sub-models built from scratch**:
`WavelengthClassification(phase=phase, mode=mode)` and
`VoiceClassification(voice_register=..., confidence=...)`.

Every field it does not score is reset to its default. The fragment's existing
values are never read. Family 0 (correctness — silent data loss).

**Current state:** `wavelength.orientation`, `wavelength.dosage`,
`wavelength.color` and `wavelength.descriptor` are wiped whenever the rules pass
writes, and `voice.confidence` is nulled. This is reachable on the main path:
`_classify_one` runs `rules.classify()` first on **every** `--method llm`
fragment (`classify_engine.py:1510`), so `creek classify --method llm --force`
over an LLM-classified vault zeroes those four axes on every rule-confident
fragment.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Run against the real classifier:

```
  before: {'phase': 'rising', 'mode': 'express', 'orientation': 'do_feel',
           'dosage': 'medicine', 'color': 'red', 'descriptor': 'hand written'}
  after : {'phase': 'rising', 'mode': 'unclassified', 'orientation': 'unclassified',
           'dosage': 'unclassified', 'color': 'unclassified', 'descriptor': ''}

E  AssertionError: orientation wiped by rules pass
E  assert <Orientation.UNCLASSIFIED: 'unclassified'> == 'do_feel'
```

(body: `"power dominance control conquest force. Emerging building growing momentum."`)
Note `mode` is also reset whenever `phase` alone crosses the gate at
`rules.py:785`. The voice half behaves the same way: for
`"I will analyze and examine the evidence systematically with data."`,
`voice_register` `prophetic -> analytical` and `confidence` `conviction -> None`.

**Signal 2 — structural + intent.** `_build_updates`' own docstring says it
"Only includes fields whose classification produced a result above the
configured thresholds" — it does the opposite for the four axes inside
`WavelengthClassification`. `rg -n "orientation|dosage|descriptor|color"
tests/test_classify.py` shows every hit is about the LLM *prompt* text; no test
pins rules-path preservation, which is why the wipe is invisible to the suite.
The nulled `voice.confidence` is also one of the two inputs to
`PrivacyClassifier._is_high_confidence_confessional`
(`creek/classify/privacy.py:204-214`) — the INTIMATE trigger that #1309 protects
on the weighted path and that is unprotected here.

## Scope

`_build_updates` and its two sub-model constructions. Related to but not covered
by #1309 (weighted path / `to_legacy`) or #1317 (test falsifiability).

## Tasks

1. In `creek/classify/rules.py:751`, merge scored axes onto the fragment's
   existing `wavelength` / `voice` sub-models (`model_copy(update=...)`) instead
   of constructing fresh ones.
2. Ensure `mode` is only written when the mode score itself clears its gate —
   not as a side effect of `phase` clearing (`rules.py:785`).
3. Add tests asserting that `orientation`, `dosage`, `color`, `descriptor` and
   `voice.confidence` survive a rules pass that scores only phase/mode/register.

## Acceptance Criteria

- [ ] The repro above passes: unscored wavelength axes are preserved.
- [ ] `voice.confidence` is not nulled by a rules pass that did not score it.
- [ ] A mutation that reintroduces whole-sub-model replacement fails a test.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/classify/rules.py` | Modify |
| `creek-tools/tests/test_classify.py` | Modify |

Geoffe-Ga:
## Summary

**In one line:** a classifier pass that was silent about an axis used to erase
that axis, which on the voice block meant erasing half of the `INTIMATE`
privacy trigger — a fail-open. Passes now merge instead of replacing, through
one shared rule.

**Which paths this closes**, because the issue title names only one function
and the real reach is wider:

- `creek classify --method rules` — `_build_updates`, the named defect.
- `creek process` — same rule pass (`pipeline.py:938`), and the sharpest
  consequence: the tier is assigned *before* the provider is chosen, so a
  fragment that should have been `INTIMATE` was seen as `PERSONAL` and routed
  to a **cloud** provider.
- `creek classify --method llm`, single-pick — `_apply_voice`, the identical
  fail-open one layer on. This is the default LLM method, and it is a
  deliberate extension beyond the issue title; the reasoning is in
  [Scope](#scope-one-extension-beyond-the-issue-title-disclosed) below.
- `creek classify --method llm`, weighted — already fixed by #1309; this PR
  re-points it at the shared rule so the two cannot drift apart.

Not closed, deliberately: `_apply_wavelength` on the single-pick LLM path
(metadata loss, not privacy — filed as #1421).

---

`RuleClassifier._build_updates` built a whole fresh `WavelengthClassification`
and a whole fresh `VoiceClassification` from the keyword matchers' verdict, and
`classify` assigned them onto the fragment. Every axis the matchers were silent
about was therefore reset to its default — including axes no matcher can ever
write.

Reproduced at HEAD (`4efc66c`), title `"A quiet note"`, verified rule-inert:

**Wavelength**, body `"The peak arrived. At its peak the whole thing reached a peak of stillness."` (fires `Phase.PEAKING`, nothing else):

```
BEFORE {'phase':'rising','mode':'express','orientation':'feel',
        'dosage':'medicine','color':'green','descriptor':'Social Anxiety'}
AFTER  {'phase':'peaking','mode':'unclassified','orientation':'unclassified',
        'dosage':'unclassified','color':'unclassified','descriptor':''}
```

Five of six axes destroyed — the four the issue names, plus `mode`, because
`phase` and `mode` share one `OR` guard so a match on either rebuilds both.

**Voice**, body `"I have to admit something. I confess I was afraid, and honestly I still am."` (fires `confessional`, no confidence keyword):

```
BEFORE {'voice_register':'analytical','confidence':'conviction'}
AFTER  {'voice_register':'confessional','confidence':None}
```

### Why this is a fail-open and not lossy bookkeeping

Every writer of `privacy_tier` is escalate-only, so nothing here is being
*lowered*. `PrivacyClassifier._is_high_confidence_confessional`
(`creek/classify/privacy.py:205-214`) escalates to `INTIMATE` only on
`voice_register == confessional` **and** `confidence == conviction`, and
short-circuits to `False` if either is `None`.

On a confessional body the rule pass supplies the missing half of that trigger
and destroys the half already on disk **in the same operation**. Verified
end to end with `PrivacyClassifier().classify_tier`:

| | tier |
|---|---|
| the fragment as persisted (`analytical` + `conviction`) | `OPEN` |
| after the rule pass at HEAD (`confessional` + `None`) | `OPEN` |
| with the evidence merged (`confessional` + `conviction`) | **`INTIMATE`** |

So this is a **missed escalation**, not a lowered tier. Every privacy assertion
in this PR therefore asserts the tier goes **UP** to `INTIMATE`. An assertion
that "the tier did not go down" is true at HEAD, true after the fix, and proves
nothing.

### The fix: one merge rule, shared

PR #1402 already merged the right shape for the weighted path
(`WeightedFragmentClassification.merge_onto`). Rather than write a second one,
this PR **extracts** it into `creek.classify.evidence.layer_determined_over` and
routes every classifier pass that writes a legacy block through it:

```python
def layer_determined_over(*, prior: _ModelT, determined: _ModelT) -> _ModelT:
    return prior.model_copy(update=determined.model_dump(exclude_defaults=True))
```

It rests on one invariant, now documented in one place: **every legacy
classification field's default *is* its "not determined" sentinel**
(`UNCLASSIFIED` / `[]` / `""` / `None`), so `exclude_defaults=True` yields
exactly the axes the pass actually spoke to. A test pins that invariant
executably for all three models, so a future field with a non-sentinel default
goes red here instead of silently reintroducing the bug.

Arguments are **keyword-only**, and that is a safety property rather than a
style choice: both operands share a type, so a positional swap typechecks
perfectly while inverting the merge — on the voice block that silently inverts a
privacy control and no gate in this repo would catch it.

`_build_updates` stays a **pure** producer of "what the matchers determined",
mirroring `to_legacy`; the new `_layer_over_fragment` does the layering,
mirroring `merge_onto`. Its docstring now says outright that assigning its
output wholesale is the defect.

**The frequency asymmetry is preserved deliberately.** `frequency` is replaced
wholesale when the matchers pick a primary (clearing stale secondaries, because
`secondary` is a list and cannot be merged field-wise) and left entirely alone
when they do not. This matches `merge_onto` on purpose, and a test pins it so
nobody later "completes" the merge by making frequency behave like the scalar
blocks.

### Scope: one extension beyond the issue title, disclosed

The issue names `_build_updates` only. This PR also fixes `_apply_voice`
(`creek/classify/llm/parsing.py`), the identical wholesale rebuild on the
single-pick LLM path, where a response carrying `voice: {voice_register: ...}`
and no `confidence` nulled a persisted `conviction`.

The chief-architect recommended deferring that to a follow-up; I overrode it,
and the reasoning should be visible rather than buried:

- Fixing only the rules layer would leave the identical fail-open open on
  `--method llm`, the default LLM method — the rule pass would preserve
  `conviction` and the parse would destroy it a moment later. The PR would then
  claim to close a fail-open that stays open.
- That same function already takes the fragment's prior value to merge `praxis`
  (#877) and `emotional_texture` (#878), for exactly this reason. This is the
  third instance of an established local pattern, not a new one.
- Measured: zero existing tests change behaviour as a result.

The cut is at the **privacy boundary**. `_apply_wavelength`
(`creek/classify/llm/calibration.py`) has the same wholesale-rebuild shape and
is **deliberately untouched**: it is entangled with the FEAT-017 `_biased_enum`
downgrade, which resets mode/orientation/dosage on a low self-reported
confidence *on purpose*, so whether that should come to mean "do not adopt the
noisy pick" or keep meaning "erase what we knew" is a design call that deserves
its own issue. No wavelength axis feeds any privacy control, so that one is
metadata loss rather than a fail-open. Filed as #1421.

### Direction: this can only cause more escalations, never fewer

`classify_tier` reads authorship, platform, recovery keywords, and the two voice
axes — no wavelength or frequency field at all. Its only voice branch is a
conjunction of two equality tests against non-default values that short-circuits
to `False` on a `None`. The merge only ever replaces a "not determined" sentinel
with prior evidence; a non-default fresh verdict always wins outright. So the
predicate can only move `False -> True`, and `escalate` / `apply_tier` /
`reassess` are all monotone in the candidate. No path is made less restrictive.

Honest boundary: the rules path is not made *globally* monotone. A prior
`{confessional, conviction}` with a body firing `analytical` still lands
`{analytical, conviction}` and drops out of the INTIMATE trigger. That is
correct — the matchers genuinely determined `analytical`, which is a verdict,
not silence. Only silence now defers.

### Scope also reaches further than the issue says

The issue and the consolidated #1400 comment both describe this as firing on
`creek classify --method llm`. It also fires on `creek classify --method rules`
(`classify_engine.py:1518` returns `rules.classify(...)` directly) and on
`creek process` (`pipeline.py:938`). On `creek process` the consequence is
sharper still, because `apply_tier` runs *after* the rule pass and the provider
is then chosen by `tier_classifiers.for_tier(...)`: a fragment that should have
been `INTIMATE` was seen as `PERSONAL` and routed to a cloud provider. This PR
closes that as a side effect.

## Test plan

New module `tests/test_rules_preserves_evidence.py` — 26 tests, 0 skipped.
Proven RED at HEAD before the fix: **10 failed, 5 passed**, every failure a
wrong value on a real object (no `ImportError`/`AttributeError`). The 5 that
passed at HEAD are declared regression guards, not padding.

- Per-axis wavelength survival, all six axes asserted explicitly on every case —
  the recurring failure on this backlog is a fix that restores the two axes the
  body emphasises and keeps resetting a third.
- Cross-destruction both ways: a phase-only body keeps `mode`, a mode-only body
  keeps `phase`.
- Both halves of the voice `OR` guard.
- **The headline test** asserts the baseline `OPEN` first — so it proves its own
  premise — then asserts `INTIMATE` after the pass.
- **Parity**: a fragment already holding both halves, and a fragment whose rule
  pass supplies one of them, must reach the same tier. At HEAD they diverge.
- **No manufactured escalation**: a `musing`/`exploring` stance under a
  confessional verdict must not reach `INTIMATE` — *and* the test asserts the
  stance survived, without which it would pass at HEAD for the wrong reason
  (the field having been destroyed rather than respected).
- **On disk**, through `run_classify(method="rules")`: the write is
  `model_dump(mode="json")` with no `exclude_none`, so a nulled confidence
  really lands as `confidence: null`; an in-memory assertion cannot see that.
  Asserted on the reloaded frontmatter, including `privacy_tier: intimate`.
- **Idempotence**: run twice; tier and evidence both still stand.
- The merge primitive's own units, including correctness over a prior that is a
  mix of `str` and enum (which `use_enum_values=True` produces), and
  non-mutation of both operands.
- The executable invariant guard: all three classification models dump to `{}`
  under `exclude_defaults`.
- A characterization test for the one downstream behaviour this moves —
  `reatomize._is_accepted`. Filed as #1422; the change makes two previously
  inconsistent cases agree, and `classification.reatomize` is off by default.

**Mutation-tested**, five guards reverted one at a time. Each killed exactly its
intended tests and nothing else:

| reverted guard | tests killed |
|---|---|
| rules wavelength layering | 3 |
| rules voice layering | 8 |
| frequency asymmetry (merged it instead) | 1 |
| single-pick LLM voice merge | 1 |
| `merge_onto` layering | 5, all pre-existing #1309 tests |

That last row is the check that the extraction is behaviour-preserving: the
weighted path's own tests still fail if the shared primitive stops merging.

**Gate 2**: `./scripts/check-all.sh` exit code **0**, 12/12 checks. 9325 passed,
5 skipped — all five are pre-existing environment gates (Ollama unreachable, API
keys unset) in files this diff does not touch. Ruff runs `--no-cache`
throughout.

### Existing tests changed

No test assertion anywhere was changed, weakened, or deleted; zero test lines
were deleted. The only test-file edit is to
`tests/test_classify_weighted_evidence_on_disk.py`, and it is **docstrings
only**.

That module's `_assert_rule_pass_preserves_evidence`, its `_RULE_INERT_BODY` /
`_RULE_INERT_TITLE` constants, and its module docstring were written by the
#1309 lane as the visible boundary of this defect while it was still open. The
helper and every one of its assertions are **kept and still called**; what
changed is what they mean. They were a fixture-dependent precondition hedging a
known gap; they are now a tripwire on a guarantee the rule pass makes
unconditionally. The inert fixture stays for a better reason than the original
one: those tests measure the *weighted* merge, and a rule-firing body would let
the rules layer explain a passing result.

Notably, no existing test needed updating for the behaviour change itself.
That is not luck and it is worth stating: every fixture that reaches a rule pass
starts at model defaults, so the two conditions this fix changes — matchers fire
**and** the fragment already carries non-default evidence — never co-occurred
anywhere in the suite. 260 tests in `test_classify.py` and not one could see
this bug.

### Not fixed here, noted

The rule matchers gate on `primary >= PRIMARY_THRESHOLD`, so a body producing
secondary frequencies but no primary writes nothing and the secondaries are
discarded. Pre-existing, unrelated to evidence destruction, and I have not
verified it beyond reading the threshold logic — recorded here rather than filed
as an issue I did not confirm.

This PR stops the bleeding; it does not resurrect evidence already destroyed on
disk by earlier runs. A repair pass would be a separate change.

Closes #1331
Refs #1309, #1400, #1421, #1422

### DIFF
--- a/creek-tools/creek/author/agents.py
+++ b/creek-tools/creek/author/agents.py
@@ -527,6 +527,12 @@ class _FragmentSignal:
         and confidence from ``title + body`` keyword signal; the rule
         classifier does not detect dosage, so dosage is read from the
         fragment's existing frontmatter (``wavelength.dosage``) instead.
+        Reading the *original* fragment for that used to be load-bearing:
+        the rule pass rebuilt the whole wavelength block and blanked the
+        dosage on its way past, so ``classified.wavelength.dosage`` was a
+        sentinel. Since #1331 the pass merges instead, and the two agree —
+        the explicit read is kept because it still states the intent, not
+        because it is the only thing that works.
         ``voice_register`` and ``confidence`` are ``None`` when no keyword
         fired (neither enum has an ``UNCLASSIFIED`` member).
 
--- a/creek-tools/creek/classify/classify_engine.py
+++ b/creek-tools/creek/classify/classify_engine.py
@@ -1506,6 +1506,10 @@ def _classify_one(
             the rules path or on ``--method llm`` runs where the rule
             classifier already cleared the confidence floor; for those
             fragments :attr:`Fragment.weighted` stays ``None``.
+            The *merge* itself is no longer particular to this flag:
+            since #1331 the rule pass and the single-pick LLM parse
+            layer their verdicts the same way, through the one shared
+            rule in :mod:`creek.classify.evidence`.
 
     Returns:
         ``(updated_fragment, skipped, reasoning)``. ``skipped`` is
@@ -1557,7 +1561,11 @@ def _classify_one_weighted(
     classification rather than replacing it wholesale — a dimension
     the profile is silent about leaves the prior value standing. The
     merge rule, and why the wholesale form was a privacy defect, live
-    on :meth:`WeightedFragmentClassification.merge_onto` (#1309).
+    on :func:`creek.classify.evidence.layer_determined_over`, which
+    :meth:`WeightedFragmentClassification.merge_onto` calls (#1309) and
+    which the rule pass upstream of it now calls too (#1331) — so by
+    the time a fragment reaches here its rule-derived evidence is
+    intact rather than already half destroyed.
 
     When the underlying call fails soft to an empty
     profile (whitespace-only body, provider unavailable, transport
--- a/creek-tools/creek/classify/evidence.py
+++ b/creek-tools/creek/classify/evidence.py
@@ -1,0 +1,116 @@
+"""Sparse-verdict layering for the legacy classification blocks.
+
+A classifier pass produces a *sparse verdict*: it speaks to the axes its
+evidence actually covers and stays silent about the rest. Assigning such
+a verdict onto a fragment wholesale erases every axis it was silent
+about, which is how one defect earned two issue numbers — once against
+the weighted path (#1309) and once against the rules path (#1331).
+:func:`layer_determined_over` is the single implementation of the merge
+rule both paths share, so neither can drift back into the bug alone.
+
+**THE** ``exclude_defaults`` **INVARIANT this rests on.** Every legacy
+classification field's default *is* its "not determined" sentinel
+(``UNCLASSIFIED`` / ``[]`` / ``""`` / ``None``). Dumping a verdict with
+``exclude_defaults=True`` therefore yields exactly the subset of axes the
+classifier actually spoke to, and anything it was silent about is simply
+absent from the update — leaving the fragment's prior evidence standing.
+**A future field whose default is not a "not determined" sentinel breaks
+this silently**: such a field would be dumped whenever a pass left it at
+some *other* value and omitted whenever the pass agreed with the
+non-sentinel default, so a classifier that never considered the axis at
+all would start stamping that default over prior evidence again, with no
+test failing on the way past. Give any new classification field a
+sentinel default, or teach this function about it explicitly.
+
+**Why this module imports nothing from** :mod:`creek`. Every caller sits
+inside a delicate import cycle: :mod:`creek.models` defers its import of
+:mod:`creek.classify.weighted` all the way to the bottom of the file
+(``creek/models.py:1149``, carrying a ``noqa: E402`` and explained at
+``creek/models.py:29`` and ``creek/models.py:813``). Importing only
+:mod:`typing` and :mod:`pydantic` — never a :mod:`creek` module — takes
+this helper out of that reasoning entirely.
+
+To be accurate about how load-bearing that is: it is not the *only*
+workable placement. All three callers already import names straight from
+:mod:`creek.models` at module-load time, so defining this above line 1149
+there would have worked too. The standalone module is chosen because it
+is independently better — single-purpose, trivially unit-testable, and
+no new edge into a 1200-line module — with cycle-immunity as a bonus
+rather than a necessity.
+
+:class:`~pydantic.BaseModel` is imported at runtime rather than behind
+:data:`typing.TYPE_CHECKING` deliberately: third-party imports are not
+what the cycle is about, and a guarded import here would be an
+unexecutable line in a module small enough that one of those drops it
+under the per-file coverage floor.
+
+Callers:
+
+- :meth:`creek.classify.rules.RuleClassifier.classify`, through its
+  ``_layer_over_fragment`` step (#1331).
+- :meth:`creek.classify.weighted.WeightedFragmentClassification.merge_onto`
+  (#1309).
+- ``creek.classify.llm.parsing._apply_voice``, the single-pick LLM
+  path's voice block (#1331).
+
+One writer of a legacy block is deliberately **not** a caller:
+``creek.classify.llm.calibration._apply_wavelength`` still rebuilds its
+block wholesale. Routing it through here is entangled with the FEAT-017
+``_biased_enum`` downgrade, which resets mode/orientation/dosage on a low
+self-reported confidence *on purpose*, so whether that should come to
+mean "do not adopt the noisy pick" or keep meaning "erase what we knew"
+is a design call tracked on its own issue. No wavelength axis feeds a
+privacy control, so unlike the voice block that one is metadata loss
+rather than a fail-open.
+"""
+
+from __future__ import annotations
+
+from typing import TypeVar
+
+from pydantic import BaseModel
+
+_ModelT = TypeVar("_ModelT", bound=BaseModel)
+
+
+def layer_determined_over(
+    *,
+    prior: _ModelT,
+    determined: _ModelT,
+) -> _ModelT:
+    """Overlay only the axes *determined* actually decided onto *prior*.
+
+    The merge rule described in this module's docstring: fields of
+    *determined* left at their "not determined" sentinel default are
+    absent from the update, so the corresponding evidence already on
+    *prior* survives. Fields the pass did decide win outright.
+
+    **The arguments are keyword-only on purpose, and it is a safety
+    property rather than a style choice.** Both operands have the same
+    type, so swapping them typechecks perfectly while inverting the
+    merge — the fresh verdict would become the base and the stale
+    evidence would overwrite it. On the voice block that silently
+    inverts a privacy control, and no gate in this repo would catch it.
+    Naming the roles at every call site is the only thing that can.
+
+    Deliberately generic over :class:`~pydantic.BaseModel` rather than
+    typed to the three classification models, because it states
+    something about merging sparse verdicts, not something about those
+    models. That generality is also its one sharp edge: it is only
+    *sound* for a model whose every default is a "not determined"
+    sentinel. :class:`~creek.models.Fragment` itself does not qualify —
+    ``voice_weight`` defaults to ``1.0``, a real value and not an
+    absence — so do not reach for this to merge whole fragments.
+
+    Args:
+        prior: The block already on record — whatever a previous run,
+            the rule pass, or the operator established. Never mutated.
+        determined: This pass's verdict for the same block, carrying a
+            sentinel default on every axis it did not decide.
+
+    Returns:
+        A copy of *prior* with the non-default fields of *determined*
+        applied. An entirely default *determined* yields an unchanged
+        copy of *prior*.
+    """
+    return prior.model_copy(update=determined.model_dump(exclude_defaults=True))
--- a/creek-tools/creek/classify/llm/orchestrator.py
+++ b/creek-tools/creek/classify/llm/orchestrator.py
@@ -284,7 +284,11 @@ class LLMClassifier:
             updates,
             unclassified_threshold=self.config.unclassified_threshold,
         )
-        _apply_voice(data, updates)
+        # Issue #1331: the voice block merges axis-by-axis, so it needs the
+        # one the fragment already carries — a response that names only
+        # ``voice_register`` must not null a persisted ``confidence``, which
+        # is half the INTIMATE privacy trigger.
+        _apply_voice(data, updates, fragment.voice)
         # Issue #877: ``praxis_potential`` is monotone, so the merge needs
         # the verdict the fragment already carries — a model answering
         # ``latent`` for an ``explicit`` fragment must lose. ``_apply_praxis``
--- a/creek-tools/creek/classify/llm/parsing.py
+++ b/creek-tools/creek/classify/llm/parsing.py
@@ -18,6 +18,7 @@ from typing import TypeVar
 
 import yaml
 
+from creek.classify.evidence import layer_determined_over
 from creek.classify.praxis_pass import PRAXIS_POTENTIAL_KEY, escalate
 from creek.models import (
     Confidence,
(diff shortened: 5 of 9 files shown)


Geoffe-Ga:
## Summary
This PR fixes a real privacy fail-open: `RuleClassifier._build_updates` (and the single-pick LLM `_apply_voice`) rebuilt whole `WavelengthClassification`/`VoiceClassification` sub-models from a sparse keyword/LLM verdict, silently resetting axes the pass never spoke to — including `voice.confidence`, half of the `confessional + conviction` trigger that escalates a fragment to `INTIMATE`. The fix extracts the merge rule PR #1402 already established for the weighted path into a single shared `layer_determined_over` primitive in a new `creek/classify/evidence.py`, and routes the rules pass and the single-pick LLM voice parse through it, closing the fail-open on all three classification methods (`rules`, `llm` single-pick, `llm` weighted) plus `creek process`.

## Strengths
- The root cause and blast radius are precisely diagnosed: the PR distinguishes "lossy" from "fail-open" and proves the fail-open with a concrete before/after privacy-tier table (`OPEN` → still `OPEN` at HEAD → `INTIMATE` after the fix), so the severity claim is falsifiable rather than asserted.
- Correctly identifies and fixes the actual reach of the bug beyond the issue title (`--method rules`, `--method llm` single-pick, and `creek process`'s pre-provider-selection tier assignment), while explicitly and reasonably declining to touch `_apply_wavelength` (entangled with an unrelated FEAT-017 design question, filed separately as #1421) — a well-justified, disclosed scope boundary rather than scope creep.
- `layer_determined_over`'s keyword-only arguments are a genuine safety property, not a style nit: both operands share a type, so a positional swap would silently invert a privacy control with no static check catching it.
- I independently verified the core mechanics against the current source (not just the diff): `_build_updates` has exactly one caller and it is routed through `_layer_over_fragment`; `_apply_voice` has exactly one caller and it now receives `fragment.voice`; `_BIASED_DIMENSIONS` in `calibration.py` indeed excludes `voice_register`/`confidence`, confirming the claim that `_apply_wavelength` and `_apply_voice` don't need the same FEAT-017 interaction reasoning; `_is_high_confidence_confessional` in `privacy.py` matches the described `confessional AND conviction` short-circuit-on-`None` logic exactly.
- Test coverage is unusually rigorous for this class of bug: per-axis wavelength assertions (not just the two axes the issue names), both directions of the phase/mode cross-destruction, both halves of the voice `OR` guard, an on-disk round-trip through `run_classify` (catching what an in-memory assertion would miss, since the write path has no `exclude_none`), idempotence, and a parity test showing two different provenances of the same evidence now reach the same tier.
- The one existing test file that was touched received docstring-only edits (confirmed in the diff — no assertions added, changed, or removed), keeping the "did not weaken any existing test" claim honest and checkable.
- `evidence.py`'s docstring documents the `exclude_defaults` invariant its soundness depends on and *why* the module avoids importing from `creek` (an existing import-cycle constraint), which will save the next reader from relitigating both.

## Security Concerns
None outstanding — this PR fixes a privacy fail-open (🟢, resolved). No new external input handling, subprocess use, or secret handling is introduced. The fix strictly narrows the conditions under which a fragment can incorrectly avoid `INTIMATE` classification; it cannot lower an existing tier (verified: `layer_determined_over` only ever fills in *missing* evidence from `prior`, never overwrites a `determined` non-default value, and every `privacy_tier` writer is escalate-only per the existing `PrivacyClassifier` code I read).

## Problems
None found. I checked the specific places most likely to hide a subtle regression (call-site signatures, the frequency-block asymmetry being left deliberately unmerged, the untouched `_apply_wavelength` path, the `_BIASED_DIMENSIONS` interaction) against the actual source, and each matched the PR's description.

## Code Quality
- The docstrings across `evidence.py`, `rules.py`, and `weighted.py` are long and repeat some of the same invariant explanation in multiple places (module docstring, function docstring, inline comments at call sites). This is a stylistic choice favoring self-containedness over DRY documentation; not a blocker, but a future editor touching this invariant will need to update it in several places.
- `layer_determined_over` is generic over `BaseModel` but is only sound for models whose every field default is a "not determined" sentinel — this is documented clearly in the docstring with the `Fragment.voice_weight` counter-example, so the sharp edge is disclosed rather than hidden.

## CLAUDE.md Adherence
✅ Coverage/testing — PR claims `./scripts/check-all.sh` exit 0, 9325 passed, and per-file coverage per repo gates; I could not execute the suite myself in this sandboxed review environment (test commands were blocked by the approval layer), but I did verify the test file's logic and assertions read correctly against the actual implementation, and the described RED→GREEN progression is consistent with the code changes.
✅ Commit conventions — commit message follows Conventional Commits (`fix(classify): ...`), on a feature branch (`issue/1331-...`), not direct-to-main.
✅ Docstring coverage — every new/changed function carries a full Google-style docstring with Args/Returns, consistent with the `creek-tools/CLAUDE.md` interrogate ≥95% requirement.
✅ Type hints — `layer_determined_over` is fully typed with a bound `TypeVar`; no obvious mypy-strict violations spotted (e.g., no untyped `Any` introduced).
✅ Complexity — the new logic (`_layer_over_fragment`, `layer_determined_over`) is straightforward branching, well within the ≤10 cyclomatic complexity threshold.

## Testing
Strong. 26 new tests in `tests/test_rules_preserves_evidence.py` cover per-axis preservation, cross-axis destruction (both directions), both halves of the voice OR-guard, privacy-tier escalation (asserting the baseline first so the test proves its own premise), a non-escalation guard against manufactured `INTIMATE` results, the merge primitive's own unit tests (including a mixed str/enum prior from `use_enum_values=True`, and non-mutation of both operands), an on-disk round trip, idempotence, the single-pick LLM voice path, and a characterization test for a downstream `reatomize` behavior change. The PR also reports a mutation-testing pass (five guards reverted individually, each killing exactly its intended tests) — strong evidence the tests aren't just coincidentally green.

## Documentation
Thorough to the point of being unusually dense — every new function documents not just what it does but why it's shaped that way, what alternatives were rejected, and what it deliberately does not do (e.g., `_apply_wavelength`, the frequency asymmetry). This exceeds the repo's docstring-coverage bar and leaves a clear paper trail for future maintainers weighing whether to "simplify" the merge back into a defect.



## Verdict: LGTM

The bug is real, precisely diagnosed, and correctly fixed at every call site I checked directly against the source (single caller of `_build_updates` and `_apply_voice`, correct `_BIASED_DIMENSIONS` exclusion, matching `_is_high_confidence_confessional` logic, escalate-only tier writers). Test coverage is exceptionally thorough (per-axis assertions, on-disk round trip, mutation-testing report) and the one existing test file touched received only docstring changes. No blocking issues, no security concerns, and nothing substantive enough to withhold approval for.

Geoffe-Ga:
**CI**: 8/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5259875483 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1424, opened 2026-08-11
fix(vault): borrowed fragments in 11-Other-Authors are invisible to every id lookup


Geoffe-Ga:
## Problem

`VaultWriter._find_in_fragments_locked` — the lookup behind `tomb_fragment`,
`find_fragment` and (since #1332) `find_tombed_fragment`'s sibling — searches
`01-Fragments/` only. A **borrowed** fragment (`source.author_slug` set) routes
to `11-Other-Authors/<slug>/` via `_fragment_target_dir`, so no lookup can ever
find one.

`restore_fragment` is unaffected: it routes through `_fragment_target_dir` and
resolves the borrowed destination correctly. So tomb and restore disagree for
borrowed fragments in exactly the way #1332 fixed for nested platform
subfolders — the nested-path half is closed, this half is not.

## Why it is filed rather than fixed in #1332

Not currently reachable through the ledgered tomb sweep: `TOMBING_SOURCES` is
`{"markdown"}`, and no ingestor sets `source.author_slug` (the only producer of
borrowed fragments is `creek/save/writer.py`, which is not ledger-backed). And
since #1332 a tomb that finds nothing reports the miss instead of recording a
tomb that did not happen, so the failure mode is now loud rather than silent.

The fix also has a different shape from #1332's. The `01-Fragments` search set
is bounded by `_PLATFORM_SUBFOLDER` (12 directories). `11-Other-Authors/` is
bounded only by the number of authors in the vault, so a lookup that sweeps it
is unbounded in a way the current one is not. That needs a design decision
(index the author dirs? require the caller to supply the slug? a vault-wide id
index?), not a wider `for` loop.

## Scope

- `creek/vault/writer.py` — `_find_in_fragments_locked` / `find_fragment` /
  `tomb_fragment`. `find_fragment`'s docstring already documents the exclusion
  since #1332 and should cite this issue once it exists.
- Decide whether `tomb_missing_units` should be able to tomb a borrowed
  fragment at all, or whether borrowed content is out of the tomb contract by
  design.

## Acceptance criteria

- [ ] A decision is recorded (in the docstring or an ADR) on whether borrowed
      fragments are in or out of the tomb/lookup contract.
- [ ] If in: `tomb_fragment` round-trips for a fragment with `author_slug` set,
      and the search cost is bounded by something other than the author count.
- [ ] If out: the exclusion is asserted by a test, not just documented, so a
[...]

Issue #674, opened 2026-06-26
feat(ingest): soft-tomb orphaned fragments on source deletion


Geoffe-Ga:
## Role

You are a Senior Python engineer working in this repo's `creek/ingest/` +
`creek/vault/` modules.

## Goal

Detect ledgered `source_key`s that are **no longer present** in the source on a
markdown (journal) ingest run and **soft-tomb** their fragments to
`10-Liminal/Orphaned/`, so the vault reflects deletions; a subsequently
re-added source unit restores a live fragment.

## Context

- **Parent epic:** #668
- **Predecessor issue(s):** #673 (must be merged first — the changed-branch
  + ledger-driven ingest path this builds the "gone" branch onto).
- **SPEC section:**
  `creek-tools/plans/2026-06-26_SPEC_continuous-idempotent-ingest.md` — R1 (§3,
  the **gone** branch, lines 119–122: *mark the fragment orphaned (soft-delete to
  `10-Liminal/Orphaned/` or tomb it)*). Resolved decision #1: a deletion
  soft-tombs, no version history (§6, lines 235–237). Success criterion: §1 line
  16 (no corruption on re-run) and §4 sketch line 205.
- **Files involved:**
  - `creek/ingest/markdown.py` — at end of a full-source ingest pass, diff the
    ledger's `source_key` set against the `source_key`s seen this run; the
    difference is the "gone" set.
  - `creek/ingest/ledger.py` (from #672/02) — enumerate ledger entries; mark
    or remove tombed entries so a second run does not re-tomb.
  - `creek/vault/writer.py` — add a soft-tomb move: relocate the fragment file
    (by id, via `.id-index.jsonl`, line 120) to `10-Liminal/Orphaned/` and stamp
    a tomb marker in frontmatter (e.g. `lifecycle: orphaned`, `orphaned_at`).
  - `creek/config.py` — `SourcePaths` (line 1244) / vault layout for the
    `10-Liminal/Orphaned/` target dir.
  - `tests/` — deletion + re-appearance tests.
- **Prior decisions:**
  - Soft-tomb (move + mark), **not** hard-delete; no version history (SPEC §6 #1).
  - Tomb destination is `10-Liminal/Orphaned/` (epic Context).
  - "Gone" is scoped to a **full-source** ingest pass (the whole journal dir is
    scanned); a single-file `--input` run must NOT tomb everything else — only a
    whole-source scan computes the gone set. (Decision: single-file input does
[...]

PR #1425, opened 2026-08-11
fix(vault): descend into nested platform subfolders, and stop recording tombs that did not happen

### LINKED ISSUE #1332: fix(vault): tomb_fragment cannot find Substack fragments, and tomb_missing_units records them tombed anyway
**Severity:** High
**Remediation:** fix in place

## Problem

`_PLATFORM_SUBFOLDER[SourcePlatform.SUBSTACK] = "Writing/Substack"`
(`creek/vault/writer.py:86`) is the only **two-level** value in the map, so
Substack fragments are written to `01-Fragments/Writing/Substack/`.
`_find_in_fragments_locked` (`creek/vault/writer.py:871`) — the lookup behind
`tomb_fragment` — enumerates only the **immediate** children of `01-Fragments`
via `fragments_root.iterdir()`, so it probes `01-Fragments/Writing`'s (empty)
index and never descends into `Substack`. Family 0 (correctness).

**Current state:** `tomb_fragment` returns `None` for every Substack fragment.
Worse, `creek/ingest/pipeline.py:293` (`tomb_missing_units`) **ignores the return
value**: it calls `writer.tomb_fragment(...)` and then unconditionally does
`ledger.record(..., tombed=True)` and `tombed += 1`. A deleted Substack export
therefore marks the ledger tombed and reports N tombed to the operator while the
fragment stays live in the vault with no `lifecycle: orphaned` — permanently
divergent ledger/vault state. The failed scan also leaves a zero-byte
`01-Fragments/Writing/.id-index.jsonl` behind.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Same test, two platforms:

```
substack: written to 01-Fragments/Writing/Substack/2026-08-10-T-frag-substack01.md
substack: tomb_fragment -> None
substack: original still live = True
journal:  written to 01-Fragments/Journal/2026-08-10-T-frag-journal001.md
journal:  tomb_fragment -> /tmp/.../10-Liminal/Orphaned/2026-08-10-T-frag-journal001.md

E  AssertionError: tomb_fragment returned None for substack
1 failed, 1 passed
```

**Signal 2 — structural + intent.** The method's own docstring claims it "Scans
each platform subfolder's id index so a tomb does not need to know which
subfolder a fragment landed in" — false for the one platform subfolder that is
nested. `restore_fragment`, which routes via `_fragment_target_dir`, resolves the
nested path correctly, so the two halves of #674 disagree with each other.

## Scope

The `iterdir()` scan in `_find_in_fragments_locked`, and the ignored return value
in `tomb_missing_units`. Distinct from #1302/#1325 (destination collision and
relocation failure modes) — this is a *lookup* miss, not a write conflict.

## Tasks

1. Make `_find_in_fragments_locked` enumerate the actual set of platform
   subfolders (derive from `_PLATFORM_SUBFOLDER.values()`) rather than assuming
   one level, so nested values resolve.
2. Have `tomb_missing_units` (`creek/ingest/pipeline.py:293`) treat a `None`
   return as a failure: do not record `tombed=True`, do not increment the count,
   and surface the miss to the operator.
3. Avoid persisting a rebuilt index for a directory with no fragments (the
   zero-byte `Writing/.id-index.jsonl` artefact).
4. Add tests: `tomb_fragment` round-trips for a Substack fragment; a `None`
   return never marks the ledger tombed.

## Acceptance Criteria

- [ ] `tomb_fragment` returns a path for every platform in `_PLATFORM_SUBFOLDER`,
      including `substack`.
- [ ] A failed tomb never records `tombed=True` in the ledger.
- [ ] A test would fail if another nested subfolder value were added without
      updating the scan.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/vault/writer.py` | Modify |
| `creek-tools/creek/ingest/pipeline.py` | Modify |
| `creek-tools/tests/test_vault_writer.py` (or nearest existing) | Modify |

Geoffe-Ga:
## Summary

Two defects, and the second is the serious one.

**The lookup miss.** `_PLATFORM_SUBFOLDER[SUBSTACK] = "Writing/Substack"` is the
only two-level value in the routing map, and `_find_in_fragments_locked`
enumerated only `01-Fragments`'s immediate children — so it probed the empty
`Writing` index and never descended into `Substack`. `tomb_fragment` and the
public `find_fragment` returned `None` for every Substack fragment, while
`restore_fragment` (routing through `_fragment_target_dir`) resolved the nested
path correctly. The two halves of #674 disagreed with each other.

**The lie.** `tomb_missing_units` discarded `tomb_fragment`'s return value and
did `ledger.record(..., tombed=True)` and `tombed += 1` unconditionally. A tomb
that found nothing therefore reported a tomb to the operator and wrote a
permanent falsehood into the ledger — permanent because `live_keys()` never
revisits a tombed record. Reproduced against `origin/main` before any code was
written:

```
reported tombed count        = 1
errors                       = []
fragment still live on disk  = True (01-Fragments/Writing/Substack/2026-08-11-T.md)
lifecycle key in frontmatter = False
orphan dir contents          = []
ledger raw bytes:
  {"source_key": "posts/gone.md", ..., "tombed": false}
  {"source_key": "posts/gone.md", ..., "tombed": true}
reloaded live_keys()         = set()
```

A fix that only makes the lookup descend leaves the unconditional-success path
intact, and it would lie again the next time any lookup failed for any reason.
Both are fixed.

### What changed

- **`VaultWriter._fragment_search_dirs`** (new) returns the **union** of every
  distinct `_PLATFORM_SUBFOLDER` value joined a segment at a time, and the
  immediate on-disk child directories. The declared half makes nested values
  reachable and makes a future *third* level free; the on-disk half is today's
  behaviour, kept because narrowing to the declared set would newly lose
  fragments in any directory the map no longer names — the same defect in a
  tidier map. Deduplicated on the resolved `Path` (`Writing` is both a declared
  value and an on-disk child, and visiting it twice could fire
  `_repair_index_locked`'s append twice) and derived per call, because the
  routing map is read at call time by `_fragment_target_dir` and monkeypatched
  by tests — a frozen copy would answer for a map no longer in force.

  Rejected: flattening `_PLATFORM_SUBFOLDER[SUBSTACK]` to `"Writing"`. It is the
  one-character alternative and it is a silent data migration — every existing
  Substack fragment in every live vault would be orphaned from its routing
  directory.

- **`_tomb_one_unit`** (new) carries the per-record body out of
  `tomb_missing_units` and separates three outcomes:
  - *relocated* — record the tomb, count it;
  - *already tombed* — the new public `VaultWriter.find_tombed_fragment` finds
    the fragment in `10-Liminal/Orphaned/`, so the ledger is reconciled on
    evidence the vault is showing, but nothing moved so nothing is counted;
  - *not found* — append `no fragment found to tomb …` to `errors`, write no
    ledger record, count nothing, and leave the unit live so the next
    full-source pass retries. Same shape as the existing `(OSError, KeyError)`
    branch, which is non-fatal and printed to the operator.

  The already-tombed branch is not politeness. Tomb-then-record is the correct
  crash ordering — the reverse leaves `tombed=True` on a live fragment, which
  `live_keys` makes permanent — and this branch is what lets the surviving
  state (tombstone on disk, ledger record still live) converge instead of
  erroring on every future pass forever.

- **The count is now the number of relocations.** It is printed verbatim as
  `N tombed`, so it may only name work that happened. A reconciliation moved
  nothing and contributes zero.

- **Lookups stopped writing to the vault.** `_load_index_locked` persisted an
  empty rebuild, so scanning a directory with nothing to index left a zero-byte
  `.id-index.jsonl` behind — which made `find_fragment`'s docstring claim that
  "asking cannot change anything" false, in exactly the directories this PR
  teaches it to visit. Guarded by `if index:`, and the claim is now an
  executable test.

### Migration: nothing to migrate, and here is why

The divergent state a migration would repair — `tombed=True` while the fragment
is still live under `01-Fragments` — **has no producer**. Tombing arms only for
`TOMBING_SOURCES == {"markdown"}`; the markdown ingestor's `_infer_platform` can
only yield `journal`/`essay`/`technical`/`markdown`, every one of them a
single-level subfolder. The ledgered sweep therefore cannot reach a nested
fragment, so the lookup defect is **latent, not live**, and no shipped vault is
corrupted by it. The one lie that *is* reachable today (a fragment moved out of
`01-Fragments` by hand) leaves a fragment `find_fragment` cannot see, so it is
undetectable as well as unrepairable.

A detect-and-report advisory was designed and tested, then cut on that
reasoning: its only firing population would have been the operator manually
restoring a fragment from `10-Liminal/Orphaned/` — precisely the case it would
have to take no action on. `TestLedgeredTombingCannotMintANestedPlatform`
asserts the reachability argument instead of leaving it in a commit message; it
goes red the day someone arms tombing for a source type that can mint
`substack`, which is the day the migration question becomes real.

Nothing in this PR deletes, moves, or rewrites operator vault state.

### The zero-byte index files

Not created any more, and existing ones need no cleanup: `_read_index_records`
parses an empty file to `{}` with `torn_lines=0`, identical to a rescan of an
empty directory. They are stale memos, not damage, and this PR does not delete
files from anyone's vault to tidy them.

### Notes

- Filed #1424 for the sibling gap this PR does **not** close: a *borrowed*
  fragment lives in `11-Other-Authors/<slug>/` and no id lookup can see it. Not
  reachable from the tomb sweep (no ingestor sets `author_slug`), and its fix
  needs a search set bounded by something other than the author count. Cited
  from `find_fragment`'s docstring.
- Pre-existing and out of scope: the index-repair path can raise `ValueError`,
  which `tomb_missing_units`'s `except (OSError, KeyError)` does not catch
  (adjacent to #1325's `RuntimeError` note).
- **Process disclosure:** an independent chief-architect pass was run and its
  verdicts are reflected here — in particular it rejected the migration
  advisory, and its argument is the one given above. The test and
  implementation specialists stalled, so those were written by the conductor
  rather than by independent specialist passes; a `code-review-orchestrator`
  pass was run over the finished diff.

## Test plan

- Both defects reproduced against `origin/main` first, asserting on-disk state:
  `tomb_fragment -> None` for substack against a real path for journal, the
  fragment still present with no `lifecycle` key, the orphan directory empty,
  and the ledger's raw JSONL bytes carrying `"tombed": true`.
- 14 new test functions in `tests/test_ingest_orphan_tomb.py` across four
  classes — 27 collected cases with the platform sweep expanded (18 -> 45
  in that file). Every
  assertion is on-disk (file presence, frontmatter keys, raw ledger bytes, a
  before/after byte-map of the whole vault) rather than lookup-shaped — a reader
  that repairs what it reads can make a return-value assertion vacuous.
- The platform sweep is parametrized over `sorted(_PLATFORM_SUBFOLDER)` and
  guarded by a companion test asserting the map still covers all 14
  `SourcePlatform` members and still holds a multi-level value, so an emptied
  parametrize list cannot pass silently. Zero tests skipped.
- **Mutation-tested**, one guard broken at a time; each mutant killed by a
  distinct test: revert the traversal to immediate children (killed by the
  substack sweep case); narrow the union to declared-only (killed by the
  undeclared-directory test); restore the empty-index persist (killed by the
  byte-identical-vault test); disable the not-found guard (killed by the
  raw-ledger-bytes test); make the count always 1 (killed by the
  no-inflation test); remove the already-tombed branch (killed by the
  convergence test); make `find_tombed_fragment` always answer yes (killed by
  its own test).
- `./scripts/check-all.sh` exit **0** — 12/12 checks, 9352 passed, 5 skipped
  (all pre-existing environment skips), coverage 94.60%.
- No `# noqa`, `# type: ignore`, `pytest.mark.skip`, `--no-verify`, threshold
  change, or deleted test in the diff. Every deleted line is a docstring line
  rewritten in place or code moved into `_tomb_one_unit`; no assertion was
  removed.

Closes #1332

### DIFF
--- a/creek-tools/creek/ingest/pipeline.py
+++ b/creek-tools/creek/ingest/pipeline.py
@@ -398,7 +398,16 @@ def tomb_missing_units(
 
     A tomb that fails on I/O is collected into *errors* and the unit is left
     live in the ledger so the next full-source pass retries it, rather than
-    crashing the whole run. Returns the number of units actually tombed.
+    crashing the whole run. A tomb that finds nothing to move is treated the
+    same way (#1332) — see :func:`_tomb_one_unit`.
+
+    Returns:
+        The number of fragments this call actually relocated into
+        ``10-Liminal/Orphaned/``. Not the number of units considered, and
+        not the number of ledger records written: a unit whose fragment was
+        already tombed has its record reconciled but moved nothing, so it
+        does not count. The figure is printed to the operator verbatim, so
+        it may only ever name work that happened.
     """
     if ledger is None or not input_path.is_dir():
         return 0
@@ -409,22 +418,80 @@ def tomb_missing_units(
         record = ledger.get(source_key)
         if record is None:
             continue
-        try:
-            writer.tomb_fragment(record.fragment_id)
-        except (OSError, KeyError) as exc:
-            errors.append(
-                f"[{source_type}] failed to tomb {record.fragment_id}: {exc}",
-            )
-            continue
-        ledger.record(
-            source_key,
-            record.fragment_id,
-            record.content_hash,
-            last_seen=record.last_seen,
-            tombed=True,
-        )
-        tombed += 1
+        tombed += _tomb_one_unit(ledger, writer, record, errors, source_type)
     return tombed
+
+
+def _tomb_one_unit(
+    ledger: SourceLedger,
+    writer: VaultWriter,
+    record: LedgerRecord,
+    errors: list[str],
+    source_type: str,
+) -> int:
+    """Soft-tomb one vanished unit; return 1 only if this call moved it.
+
+    The three outcomes are kept distinct because conflating them is the
+    #1332 defect. Before it, the caller discarded
+    :meth:`~creek.vault.writer.VaultWriter.tomb_fragment`'s return value and
+    recorded ``tombed=True`` unconditionally, so a lookup that found nothing
+    wrote a permanent lie into the ledger — ``live_keys`` never revisits a
+    tombed record — and reported a tomb the operator could disprove by
+    opening the vault.
+
+    - **Relocated.** The fragment moved. Record it and count it.
+    - **Already tombed.** ``tomb_fragment`` returns ``None`` both when it
+      found nothing and when the fragment is already in the orphan
+      directory, so the two are separated by asking
+      :meth:`~creek.vault.writer.VaultWriter.find_tombed_fragment`. This is
+      the state a crash between the file move and the ledger append leaves
+      behind — the tomb-then-record ordering is deliberate, because the
+      reverse leaves ``tombed=True`` on a live fragment, which
+      ``live_keys`` guarantees is permanent. Recording it here is not a
+      guess: the vault is showing the tombstone. Nothing moved, so nothing
+      is counted.
+    - **Not found.** No live fragment and no tombstone. The record stays
+      live so the next full-source pass retries, and the miss is reported.
+      Recording success here is the behaviour this function exists to
+      remove; recording a third ledger state instead would need every
+      reader of ``tombed`` to learn a new case to answer the same question.
+
+    Args:
+        ledger: The ledger to reconcile; written only on a proven tomb.
+        writer: Vault writer performing the relocation and the lookups.
+        record: The ledgered unit that this pass no longer saw.
+        errors: Collector for ``[<source_type>] …`` operator messages.
+        source_type: Registry key, used to prefix those messages.
+
+    Returns:
+        ``1`` when this call relocated the fragment, else ``0``.
+    """
+    try:
+        moved = writer.tomb_fragment(record.fragment_id)
+        # ``or`` short-circuits, so the orphan directory is consulted only
+        # when nothing was relocated. A ``Path`` is always truthy, so this
+        # falls back exactly when *moved* is ``None``.
+        tombstone = moved or writer.find_tombed_fragment(record.fragment_id)
+    except (OSError, KeyError) as exc:
+        errors.append(
+            f"[{source_type}] failed to tomb {record.fragment_id}: {exc}",
+        )
+        return 0
+    if tombstone is None:
+        errors.append(
+            f"[{source_type}] no fragment found to tomb for "
+            f"{record.fragment_id} (source {record.source_key}); left live "
+            f"in the ledger for the next full-source pass",
+        )
+        return 0
+    ledger.record(
+        record.source_key,
+        record.fragment_id,
+        record.content_hash,
+        last_seen=record.last_seen,
+        tombed=True,
+    )
+    return 1 if moved is not None else 0
 
 
 def should_skip_unit(
--- a/creek-tools/creek/vault/writer.py
+++ b/creek-tools/creek/vault/writer.py
@@ -868,17 +868,78 @@ class VaultWriter:
             self._log_provenance_locked(fragment.id, fragment.type, existing)
         return existing
 
+    @staticmethod
+    def _fragment_search_dirs(fragments_root: Path) -> list[Path]:
+        """Return every directory under *fragments_root* a fragment may sit in.
+
+        The union of two sets, and both halves earn their place:
+
+        - **Declared** — every distinct value of :data:`_PLATFORM_SUBFOLDER`,
+          joined a segment at a time. This is what makes a *nested* value
+          reachable. ``substack -> "Writing/Substack"`` is the only two-level
+          entry today, and enumerating immediate children alone probed the
+          empty ``Writing`` index and never descended (#1332). Deriving the
+          set from the routing map means a future third level is covered the
+          day it is added rather than the day someone remembers the scan.
+        - **On disk** — the immediate child directories, which is what this
+          method used to return outright. Kept, because narrowing to the
+          declared set would newly *lose* fragments in any directory the map
+          no longer names — a hand-made folder, or one an older release
+          routed to. Shipping that inside a fix for "the lookup misses
+          things" would be the same defect wearing a tidier map.
+
+        Derived per call rather than memoised at import: the routing map is
+        monkeypatched by tests and read at call time by
+        :meth:`_fragment_target_dir`, so a frozen copy would answer for a map
+        that is no longer in force. The cost is a dozen path joins against an
+        ``iterdir`` syscall this method already pays.
+
+        Args:
+            fragments_root: The ``01-Fragments`` directory, which must exist.
+
+        Returns:
+            Deduplicated, sorted directories. Declared entries that do not
+            exist on disk are included and create nothing: the index load
+            returns an empty mapping for a non-directory. It does *cache*
+            that emptiness against the path, which is new — before #1332
+            only directories that already existed were ever visited — and
+            within one :class:`VaultWriter` it is harmless, because
+            :meth:`_write_model` mutates the same cached dict in place. It
+            does widen, by the width of one lookup, the cross-process index
+            staleness already declared out of scope for BUG-006 in the
+            ``self._lock`` docstring: another process could create and
+            populate a declared directory between this process caching it
+            empty and a later write in the same run.
+        """
+        declared = (
+            fragments_root.joinpath(*subfolder.split("/"))
+            for subfolder in dict.fromkeys(_PLATFORM_SUBFOLDER.values())
+        )
+        on_disk = (p for p in fragments_root.iterdir() if p.is_dir())
+        # Deduplicate on the resolved Path, not the string: ``Writing`` is
+        # both a declared value (ESSAY) and an on-disk child, and visiting it
+        # twice would let one lookup fire ``_repair_index_locked``'s append
+        # twice.
+        return sorted({*declared, *on_disk})
+
     def _find_in_fragments_locked(self, fragment_id: str) -> Path | None:
         """Return the live fragment file for *fragment_id* under 01-Fragments.
 
         Caller must hold ``self._lock``. Scans each platform subfolder's id
         index so a tomb does not need to know which subfolder a fragment
-        landed in.
+        landed in — including the nested ones, which is what
+        :meth:`_fragment_search_dirs` is for.
+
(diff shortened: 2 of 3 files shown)


Geoffe-Ga:
## Summary
Fixes two related defects in the fragment tombing path (#1332): (1) `_find_in_fragments_locked` only scanned immediate children of `01-Fragments`, so it never descended into the nested `Writing/Substack` subfolder, making `tomb_fragment`/`find_fragment` blind to Substack fragments; and (2) `tomb_missing_units` discarded `tomb_fragment`'s return value and recorded `tombed=True` unconditionally, so a lookup miss silently wrote a permanent false record into the ledger. Both are fixed with new helpers (`VaultWriter._fragment_search_dirs`, `VaultWriter.find_tombed_fragment`, `pipeline._tomb_one_unit`) and a large, on-disk-state-asserting test suite.

## Strengths
- Both bugs were reproduced against `origin/main` first (per the PR description) with concrete on-disk evidence, not just a suspicion.
- `_fragment_search_dirs` derives its search set from the routing map itself rather than hardcoding `Writing/Substack`, so a future third-level route is covered automatically — verified this claim with the `test_a_three_level_subfolder_resolves_without_a_code_change` test, which monkeypatches a 3-level path in and confirms it resolves with no code change.
- The three-way outcome split in `_tomb_one_unit` (relocated / already-tombed / not-found) is exactly right: I traced it end-to-end — `moved or find_tombed_fragment(...)` correctly discriminates "nothing there" from "already moved," the ledger is only written on proven evidence, and the returned count (`1 if moved is not None else 0`) can never count a no-op reconciliation as a tomb.
- The "no migration needed" argument in the PR body is independently verifiable and I verified it: `TOMBING_SOURCES == {"markdown"}`, and `_infer_platform`/`_infer_platform_from_path` can only produce `JOURNAL`/`ESSAY`/`CODE`/`MARKDOWN`, all single-level in `_PLATFORM_SUBFOLDER`. The ledgered tomb sweep genuinely cannot reach a nested platform today, so shipping no migration is correct, not just convenient.
- Fixing the zero-byte `.id-index.jsonl` side effect (`_load_index_locked`'s new `if index:` guard) is a good catch — it directly follows from widening the search set, since more of a lookup's candidate directories are now empty ones.
- Test suite is unusually rigorous: raw-JSONL-byte assertions (not just `SourceLedger.load()`'s collapsed view) directly targeting the "wrote a lie" defect, a byte-identical-vault snapshot proving lookups are side-effect-free, and a companion test guarding the parametrized sweep itself from going vacuous (`test_the_routing_map_is_total_and_holds_a_nested_value`).
- Docstrings are unusually thorough and — I checked several of the more surprising claims against the code (in-place index-dict mutation in `_write_model`, the `source_key`/`record.source_key` equivalence via `SourceLedger`'s dict keying) — accurate.

## Security Concerns
None. No new I/O surface, no new external input, no path traversal risk (`_fragment_search_dirs` only joins static routing-map segments and on-disk child names already scoped under `01-Fragments`).

## Problems
None found. This is a correctness-critical fix and I traced the control flow carefully looking for a case where a real tomb could still go uncounted or a miss could still be recorded — didn't find one.

## Code Quality
- Nit: `_fragment_search_dirs`'s docstring says "Deduplicate on the resolved `Path`," but the implementation dedupes on `Path.__eq__`/`__hash__` (structural equality of the joined segments), not `Path.resolve()`. It happens to be correct here because both `declared` and `on_disk` paths are built the same way (joined under the same `fragments_root`), but the wording could mislead a future reader into thinking symlinks or `..` segments are normalized. Not worth blocking on.
- Nit: `_tomb_one_unit` takes `source_type` only to prefix error strings — a small amount of parameter-threading that's already well-precedented by the pre-existing `(OSError, KeyError)` branch it mirrors, so leaving as-is is reasonable.

## CLAUDE.md Adherence
✅ TDD workflow followed (tests described as written to reproduce the defects first). ✅ No `# noqa`, `# type: ignore`, `pytest.mark.skip`, or threshold changes present in the diff. ✅ Docstrings are Google-style with Args/Returns/Raises sections throughout the new code. ✅ Commit target is a feature branch (`issue/1332-...`), not directly to main. ✅ PR description reports `./scripts/check-all.sh` exit 0, 12/12 checks, coverage 94.60% — consistent with the repo's ≥90% coverage / ≥95% docstring-coverage bars. CI on the PR was still finishing the 3.11/3.12/3.13 test matrix at review time, but every job that had completed (Static Analysis, Pylint, Integration & E2E, Complexity, Build) was green.

## Testing
Excellent. 14 new test functions (27 collected cases with the platform sweep) across four well-scoped classes, each targeting on-disk state (file presence, frontmatter keys, raw ledger JSONL bytes, before/after byte-maps) rather than only return values — which specifically avoids the "return-value assertion is vacuous because the code also self-heals" trap. The claimed mutation-testing pass (one guard broken at a time, each killed by a distinct test) is plausible given the specificity of the assertions I read (e.g., the raw-ledger-bytes check in `test_a_tomb_that_found_nothing_is_not_recorded_tombed` is precisely shaped to kill the original unconditional-`tombed=True` bug).

## Documentation
Docstrings for all new/changed public and private members (`_fragment_search_dirs`, `find_tombed_fragment`, `_tomb_one_unit`, `_load_index_locked`, `tomb_missing_units`, `find_fragment`) are detailed, explain rejected alternatives, and cross-reference the related issue (#1332) and the still-open gap (#1424, borrowed fragments under `11-Other-Authors/`). This is well above the repo's docstring bar and genuinely useful for a future reader trying to understand why the code is shaped this way.


## Verdict: LGTM

Both defects are real, correctly fixed, and thoroughly tested — I independently verified the return-value threading in `_tomb_one_unit` (relocated/already-tombed/not-found are correctly distinguished and only the first is counted), the `_fragment_search_dirs` union-and-dedup logic against `_find_existing_locked`/`_load_index_locked`, and the no-migration-needed claim by tracing `_infer_platform`/`_infer_platform_from_path` to confirm markdown ingest can never mint a nested `SourcePlatform`. No blocking issues; nothing rises above a nit.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5260238057 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1428, opened 2026-08-12
verify(ingest): does ClaudeIngestor accept a real claude.ai export, or only a synthetic role/content shape?


Geoffe-Ga:
**Verify the premise before doing any work on this one.** I could not check it from inside the repo; it needs one real `conversations.json` from a claude.ai data export.

## The observation

Issue #1333 shipped a reproduction written against this shape:

```python
{"sender": "human", "text": "part one of my question"}
```

`ClaudeIngestor` does not read that shape. It keys on `role` and `content`:

- `creek/ingest/claude.py::_pair_turns` compares `msg.get("role", "")` against `"human"` / `"assistant"` / `"system"`.
- `creek/ingest/claude.py::_extract_text` reads `msg["content"]` (a string, or a list of `{"type": "text", "text": ...}` parts).
- `_normalize_messages` maps only `timestamp` -> `created_at`. There is no `sender` -> `role` or `text` -> `content` mapping anywhere in the package (`grep -r sender creek/` finds nothing).
- `_is_claude_export` accepts a top-level **dict** with a `conversations` key, or a dict with `messages` plus `conversation_id`/`uuid`.

Run the issue's dict through the current code and you get `[]` — not because of the bug it was filing, but because no role ever matched. Every fixture in `tests/test_ingest_claude.py` uses `role`/`content`, so the tests agree with the code and neither agrees with whatever produced that repro.

## The question

Does a real claude.ai export match what the ingestor accepts? If the export is a top-level JSON **array** of conversations whose messages live under `chat_messages` with `sender`/`text` fields, then:

- `discover()` rejects the file outright (`_is_claude_export` requires a dict), so nothing is ingested at all; or
- if it is reached some other way, every message has role `""` and produces zero fragments.

Either way the `--type claude` path would be non-functional against real exports and only ever exercised by synthetic fixtures — which would explain why the fixtures and the code agree so comfortably.

The counter-possibility is that the accepted shape is a **staging** format some pre-ingest transform produces (the repo has precedent: Discord/Substack/ChatGPT raw exports need pre-ingest transforms), in which case the ingestor is correct and only the docs are missing.

## Steps

1. Take one real claude.ai export (`conversations.json` from Settings -> Export data). Record its top-level type and one message's keys **in this issue** — that single paste settles it.
[...]

Issue #1427, opened 2026-08-12
fix(clean): ChatbotFilter reshapes turn runs — short-human drops create adjacencies its regeneration collapse then discards


Geoffe-Ga:
**Found while fixing #1333** (which merged consecutive same-role messages into one turn). Verified by reading `creek/clean/filters/chatbot.py`, not by running the filter.

`ChatbotFilter` runs *before* turn pairing in both chat ingestors. Two of its behaviours interact badly with the message-run structure it hands downstream.

## 1. Short-human-turn filtering manufactures consecutive-human runs

`chatbot.py:260-267` drops any human message shorter than `min_human_turn_length` (default 20 characters). Drop the middle message of `human("ok"), human("here is the real question"), assistant(...)` and the pairing sees a normal turn — fine. But drop the *only* human of a turn and the surrounding humans become adjacent, so the filter reshapes the conversation into runs that were never sent as runs.

Post-#1333 that merges two unrelated questions into one fragment body instead of dropping one of them; pre-#1333 it dropped one of them. Neither is what the operator sent. The threshold is also a blunt instrument on its own terms: "Yes, exactly — do that." is 24 characters of the operator's voice and "no, the other one" is 17 and gone.

## 2. Regeneration collapse can discard a real response

`chatbot.py:273-319` (`_collapse_regenerations`) reduces every run of consecutive assistant messages to its **last** element, on the theory that the earlier ones are regenerations. That theory is only sound if the run really is a set of alternatives. When phase 1 drops the human message *between* two assistant messages, phase 2 sees a run that the export never contained, and discards a genuine earlier response as if it were a superseded draft.

The two phases are independent today; nothing tells phase 2 that an adjacency it is looking at was created by phase 1 rather than by the source.

## Note on scope

#1333 deliberately did **not** change either behaviour. The new merge in `creek/ingest/turns.py` cannot conflict with `_collapse_regenerations`, because the filter runs first and the merge then sees a single assistant message and returns it byte-identically. This issue is about the filter's own semantics.

`ChatbotFilter` is opt-in and, as far as this reading goes, is not wired into the default `creek ingest` path — worth confirming as the first step, since it changes the severity.

## Acceptance criteria

- [ ] Confirm (and record in the issue) whether `ChatbotFilter` is reachable from any default CLI/MCP ingest path.
- [ ] Phase 2 does not collapse an assistant adjacency that phase 1 created; a real response separated from another by a dropped human message survives.
[...]

Issue #1426, opened 2026-08-12
fix(generate): split chat human turns are silently excluded from the AI-style fingerprint


Geoffe-Ga:
**Found while fixing #1333.** Verified by reading the code, not by running the generator.

## Problem

`creek/generate/ai_style/fingerprint.py` builds its corpus from a fragment's body, and for conversation platforms it insists on a marker that per-turn-split fragments do not carry — so **every Claude and ChatGPT human turn written since per-turn attribution landed is silently absent from the AI-style fingerprint**.

The chain:

- `fingerprint.py:39` — `_CONVERSATION_PLATFORMS = frozenset({"chatgpt", "claude"})`.
- `fingerprint.py:104` (`_user_text`) — for those platforms, returns `extract_user_turns(body)` instead of the whole body.
- `extract_user_turns` — walks for a `**User**:` / `**Assistant**:` blockquote structure and returns `None` when no `**User**:` marker is present ("the body cannot be split cleanly, so it is excluded rather than risk fingerprinting AI text").
- `fingerprint.py:202-203` — `text = _user_text(...)`; `if not text: continue`. The fragment drops out of the corpus.

But a per-turn-split fragment's body has no such marker. `ClaudeIngestor.convert_to_markdown` emits the human turn as a plain blockquote (`> line`, per line, nothing else), and `ChatGPTIngestor.convert_to_markdown` emits `# {title}` plus a plain blockquote. The `**User**:` / `**Assistant**:` shape belongs to the **pre-split merged** era — which is also what the surviving test fixture uses (`tests/generate/ai_style/test_fingerprint.py:45`), so the tests corroborate the old format rather than catching the gap.

Net effect: the exclusion path that exists to keep AI prose *out* of the fingerprint now keeps the operator's own chat prose out too. Chat is a high-volume source, so this is a large, silent hole in the sample.

## Why it is worth fixing rather than accepting

The safety intent (never fingerprint AI text) is already satisfied *upstream* for split fragments: the human turn is `source.author = self` and the AI turn is `source.author = ai`, and `fingerprint.py:196` already filters on `source.author != "self"`. For a split fragment the whole body is the human turn by construction, so no marker parsing is needed to be safe.

## Suggested shape (not prescriptive)

[...]

PR #1429, opened 2026-08-12
fix(ingest): merge split chat turns, and stop dropping the operator's own words

### LINKED ISSUE #1333: fix(ingest): ClaudeIngestor._pair_turns drops all but the last of consecutive human messages
**Severity:** High
**Remediation:** fix in place

## Problem

`_pair_turns` (`creek/ingest/claude.py:175`) sets `pending_human = msg` on each
human message, so an earlier human message in a consecutive run is overwritten
and never reaches a fragment. Its own docstring (`claude.py:157`) says
"Consecutive human messages are **merged** (only the last is kept)" — the two
halves of that sentence contradict each other, and the code does the second.
Family 0 (correctness — silent data loss), corroborated by family 9 (a
tautological test).

**Current state:** splitting a thought across two messages before the model
replies is ordinary Claude usage. The first half of the user's own words never
enters the vault — and therefore never enters the voice corpus, which is built
from self-authored fragments.

## Evidence (corroboration)

```python
_pair_turns([
    {"sender": "human",     "text": "part one of my question"},
    {"sender": "human",     "text": "and part two"},
    {"sender": "assistant", "text": "answer"},
])
```
```
  _pair_turns -> []                       # nothing printed: no pair carries part one
E AssertionError: first human message silently dropped
E assert 'part one of my question' in '[]'
```

**Signal 2 — reading/intent.** The only test covering this,
`tests/test_ingest_claude.py:1062 test_consecutive_human_messages`, asserts only
`len(fragments) >= 1` with the comment "Should handle gracefully" — it cannot
fail for any implementation that returns anything, and never asserts the first
message survives. The docstring's own word "merged" names the intended
behaviour: concatenate the run, do not drop it.

## Scope

`_pair_turns` in the Claude ingestor and its test. The equivalent ChatGPT
pairing path should be checked for the same shape but is not assumed defective
here.

## Tasks

1. Change `_pair_turns` (`creek/ingest/claude.py:175`) to accumulate consecutive
   human messages (join their text) rather than overwrite `pending_human`,
   matching the docstring's "merged".
2. Decide and document the join separator (a blank line reads best in a
   fragment body).
3. Replace `tests/test_ingest_claude.py:1062`'s `len(fragments) >= 1` with an
   assertion that both human messages' text is present in the emitted fragment.
4. Check `creek/ingest/chatgpt.py`'s pairing for the same pattern and fix or
   explicitly pin it.

## Acceptance Criteria

- [ ] The repro above passes: "part one of my question" reaches a fragment.
- [ ] `test_consecutive_human_messages` fails if the drop is reintroduced.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/ingest/claude.py` | Modify |
| `creek-tools/tests/test_ingest_claude.py` | Modify |

Geoffe-Ga:
## Summary

A turn in a chat export is not a message. Splitting a thought across two sends before the model replies is ordinary usage — and until now the first half never reached a fragment.

`ClaudeIngestor._pair_turns` kept exactly one message from each run of consecutive same-role messages and dropped the rest, silently: no error, no counter, no trace in the vault. Its docstring said *"Consecutive human messages are merged (only the last is kept)"* — two different behaviours in one sentence, and the code did the second.

**Why this is worse than a lost message.** Human turns are written as `source.author = self` fragments, and self-authored fragments are the entire training sample for the voice fingerprint (`VoiceExemplarCollector._eligible_register`). So the fingerprint was not learning from how the operator writes; it was learning from a *silently filtered* view of how the operator writes — the second half of every split thought, never the first, with no signal that anything was missing.

This merges each run into one turn, in both chat ingestors, and makes the docstrings true.

## Where the issue was wrong, and where it understated the defect

**The reproduction in the issue does not reproduce the bug it describes.** It passes `{"sender": "human", "text": ...}`; the ingestor keys on `role` and `content`. Run verbatim it returns `[]` — but because *no role ever matched*, not because of the overwrite. Re-run against the shape the code actually reads, the defect is real:

```
_pair_turns([{'role':'human','content':'part one of my question'},
             {'role':'human','content':'and part two'},
             {'role':'assistant','content':'answer'}])
-> [({'role': 'human', 'content': 'and part two'}, {'role': 'assistant', 'content': 'answer'})]
```

That mismatch is itself worth a look and is filed as #1428: no code anywhere in the package maps `sender`/`text`, and every Claude fixture uses `role`/`content`, so the tests agree with the ingestor and neither has ever met a real export. Settling it needs one paste of a real `conversations.json`, which is why it is a separate issue rather than a guess made here.

**The defect is symmetric, and the issue scoped it to humans only.** The same function drops the 2nd and later messages of an *assistant* run, because `pending_human` is already cleared:

```
_pair_turns([{'role':'human','content':'q'},
             {'role':'assistant','content':'reply part one'},
             {'role':'assistant','content':'reply part two'}])
-> [({'role':'human','content':'q'}, {'role':'assistant','content':'reply part one'})]
```

Fixed here too.

**ChatGPT is defective the same way**, which the issue listed as "check, but not assumed defective". `_pair_messages_to_fragments` was an index walk with strict adjacency lookahead, and it lost content three ways: `user, user, assistant` skipped the first user; `user, assistant, assistant` swept the second assistant into the "any other role" arm; and `user, system, assistant` produced **zero** fragments, because a system node broke adjacency — contradicting that function's own docstring, which already claimed system messages were skipped.

## The design decision: merge, and join with a blank line

Merging is the right reading — two messages before a reply are one utterance split by the send button — but the separator is a real choice, and `"\n".join` is the wrong one. `TURN_TEXT_SEPARATOR` is `"\n\n"` because:

- A blank line is the only separator CommonMark treats as a **block** boundary. Under a single newline, an indented code block or a list sent as its own message becomes a lazy continuation of the preceding paragraph, and two sentences run together into one.
- `ClaudeIngestor.convert_to_markdown` prefixes every line of a human turn with `"> "`. A blank separator becomes a quoted blank line and keeps two paragraphs inside one blockquote; a soft break would have fused them.
- No *visible* delimiter (`---`, `[continued]`) is inserted. Whatever goes there is indistinguishable downstream from text the operator wrote — it gets embedded, classified, and weighed by the voice fingerprint as their prose.

**Merging the assistant side does not fight `ChatbotFilter._collapse_regenerations`**, which reduces an assistant run to its last message on the theory that the earlier ones are regenerations. That filter runs *before* pairing, so when it is configured the merge only ever sees a single assistant message and returns it byte-identically. When it is not configured, ingest has no evidence about which of two assistant messages the operator meant to keep, and preserving both is the only non-lossy answer. (The filter has its own turn-shaping problems: #1427.)

## No ids shift, by construction

Fragment ids are `SHA-256(identity_key, timestamp, content)`. Neither chat ingestor sets `source_unit`, so `turn_index` is **not** in the id — and pair count is invariant with the old implementation for every input that previously produced pairs, so `turn_index` and the `(turn N)` titles do not renumber either. The single exception is ChatGPT `user, system, assistant`, which went from 0 pairs to 1.

Two deliberate choices keep the blast radius to the text actually being recovered:

- **`merge_turn_texts` returns a one-element run byte-identically** — not stripped, not normalised. That is the load-bearing case, not the degenerate one: any change to an already-correct single-message turn would re-mint its id and orphan the fragment already on disk. Two tests pin it and a stripping mutation reds four.
- **The human carrier of a merged run is its *last* message.** `_resolve_timestamp` / `_resolve_authored_at` read the human dict, and *both* fragments of the turn are stamped from it. The pre-fix code already used the last human, so keeping it means the AI fragment — whose text has not changed at all — keeps its timestamp and therefore its id. Same reasoning for `user_run[-1].get("create_time")` on the ChatGPT side.

## Migration: re-ingest genuinely recovers it, and nothing is deleted

Vaults ingested from Claude/ChatGPT exports are missing that text today. The first thing to check was whether re-ingesting is a no-op — the shape #1393 is open for — and it is not: `ledger_for_source` serves a ledger only for `markdown`, so the chat sources are unledgered and nothing records those turns as already processed. `creek ingest --type claude --input <export dir> --vault <vault>` recovers the text.

It is not free of residue, and the honest version is in `docs/ingestion.md`:

- The recovered turn has different content, so it hashes to a **new** id and is written *alongside* the truncated fragment, which stays exactly where it is. Expect near-duplicates on affected turns.
- Aggregate parents (`creek atomize`, FEAT-022) hash their ordered child ids, so a recovered child mints a new parent and leaves the old one behind the same way.

No vault content is deleted or rewritten on a heuristic, and no new migration command is added. The seventh detect-and-report in a row.

## Downstream check

`--refresh-ai-chat` (`resplit_merged_ai_chat`) decides whether a fragment is a *merged pre-split* turn by looking for non-blockquoted prose in the body. A merged human run is now multi-paragraph, so this needed checking: it is safe, because `convert_to_markdown` blockquotes every line and `_parse_claude_merged` discards blank lines from the "assistant prose" bucket. A regression test now builds its body by calling `convert_to_markdown` rather than hand-writing it, so a future change that lets any line through unprefixed fails the test instead of silently making the migration destructive — it would otherwise re-attribute the second half of the operator's own words to the AI.

## The tautological test

`tests/test_ingest_claude.py::test_consecutive_human_messages` ended at:

```python
# Should handle gracefully; at least one fragment produced
assert len(fragments) >= 1
```

No implementation that returns anything can fail that. It named the exact scenario and asserted nothing about it — which is why the bug survived being tested. It is **inverted in place, not deleted**: same name, same fixture messages, now asserting that both messages' text reaches the human fragment separated by a blank line, with a comment recording that the old assertion corroborated the drop.

## Test plan

- `./scripts/check-all.sh` → **exit 0**. 9377 passed, 5 skipped (all pre-existing environment skips: Ollama, live API keys), 94.60% coverage, per-file gate clean, ruff `--no-cache` clean, refurb clean.
- New `tests/test_ingest_turns.py` (18 tests) pins both primitives directly, including the byte-identity invariant and the indented-code-block separator case.
- Mutation-tested, each half independently:
  - revert `claude.py` alone → exactly 2 tests red (the two Claude drop tests);
  - revert `chatgpt.py` alone → exactly 3 tests red (the three ChatGPT tests);
  - separator `"\n\n"` → `"\n"` → 7 red;
  - remove the single-element short-circuit → 1 red (the whitespace-only case, precisely the one it exists for);
  - strip each part before joining → 4 red, including both id-stability guards.
- Bypass sweep over the whole diff including tests: no `# noqa`, `# type: ignore`, `pylint: disable`, skip/xfail, `--no-verify`, threshold edits, or deleted tests. The only deletions are the two rewritten function bodies and the three lines of the inverted assertion.

## Follow-ups filed

- #1426 (P1) — split chat human turns are silently excluded from the AI-style fingerprint: `extract_user_turns` requires a `**User**:` marker that per-turn-split bodies do not carry, so it drops them. Same class of silent filtering as this bug, one layer up.
- #1427 (P2) — `ChatbotFilter`'s short-human-turn drop manufactures assistant adjacencies that its regeneration collapse then discards.
- #1428 (P2) — verify `ClaudeIngestor` against a real claude.ai export before trusting the `role`/`content` shape.

Closes #1333

### DIFF
--- a/creek-tools/creek/ingest/chatgpt.py
+++ b/creek-tools/creek/ingest/chatgpt.py
@@ -28,6 +28,7 @@ from creek.ingest.base import (
     RawDocument,
     normalize_encoding,
 )
+from creek.ingest.turns import group_turn_runs, merge_turn_texts
 from creek.models import Authorship
 
 if TYPE_CHECKING:
@@ -540,12 +541,24 @@ def _pair_messages_to_fragments(
     conversation_id: str | None = None,
     conversation_create_time: float | None = None,
 ) -> list[ParsedFragment]:
-    """Pair consecutive user+assistant messages into fragments.
+    """Pair user+assistant turns into fragments.
 
-    Iterates through the ordered message list, pairing each user
-    message with the following assistant message. System messages
-    and unpaired user messages at the end are skipped. Each fragment
-    title includes a turn index to prevent collisions.
+    Segments the ordered message list into ``(user_run, assistant_run)``
+    turns via :func:`creek.ingest.turns.group_turn_runs`, merging each
+    run's text. System and tool messages are skipped; an unanswered user
+    turn at the end is discarded. Each fragment title includes a turn
+    index to prevent collisions.
+
+    Until #1333 this was an index walk that paired a user message only
+    with a *strictly adjacent* assistant message, which lost content
+    three ways, all silently: the first of two consecutive user messages
+    was skipped entirely, the second of two consecutive assistant
+    messages was swept up by the "any other role" arm, and a system node
+    sitting between a question and its answer suppressed the whole
+    exchange — despite this docstring already claiming system messages
+    were skipped. The user-side loss is the costly one: those fragments
+    are ``source.author=self`` and are what the voice fingerprint trains
+    on.
 
     FEAT-031: each pair carries ``authored_at`` derived from the
     user message's ``create_time``. When the per-message field is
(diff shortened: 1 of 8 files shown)


Geoffe-Ga:
## Summary
Fixes issue #1333: `ClaudeIngestor._pair_turns` and ChatGPT's `_pair_messages_to_fragments` silently dropped all but one message from a run of consecutive same-role messages (human *and* assistant sides), which meant split human thoughts never reached the voice-training corpus. The fix extracts a shared `group_turn_runs`/`merge_turn_texts` pair into a new `creek/ingest/turns.py` module, used by both ingestors, joining merged runs with a blank line (`"\n\n"`) rather than `"\n"`.

## Strengths
- Root-caused correctly and fixed symmetrically: both the Claude human-drop and the less-obvious assistant-drop, plus a third independent ChatGPT bug (`user, system, assistant` producing zero fragments) that the original issue didn't even scope.
- Extraction of `group_turn_runs`/`merge_turn_texts` into one shared module is the right move — avoids fixing the same bug twice, differently, and is well-isolated/testable (`tests/test_ingest_turns.py`, 18 focused tests).
- Careful id-stability engineering: `merge_turn_texts` returns single-message runs byte-identically (verified against `generate_fragment_id`, which hashes `source:timestamp:content` — no `source_unit` is set by either ingestor, so `turn_index` was never in the id and pair count is provably invariant for previously-working inputs). I traced this claim through `creek/ingest/base.py` and it holds.
- The blank-line separator choice is justified with a concrete CommonMark failure mode (lazy continuation under a bare `\n`) and is checked against `ClaudeIngestor.convert_to_markdown`'s per-line `"> "` blockquoting.
- The `--refresh-ai-chat` safety argument (a merged human run must not be mis-parsed as containing "AI prose") is verified against `_parse_claude_merged` in `refresh.py`, and a new regression test (`test_merged_run_human_fragment_is_skipped`) builds its fixture via the real `convert_to_markdown` rather than hand-writing it, so future formatting drift fails loudly instead of silently corrupting attribution.
- The previously-tautological `test_consecutive_human_messages` (`assert len(fragments) >= 1`, which nothing could fail) is rewritten in place to assert actual content and pairing — with a comment explaining why the old assertion was worthless.
- Migration story is explicit and honest in `docs/ingestion.md`: re-ingest recovers text non-destructively (chat sources are unledgered), old truncated fragments and orphaned aggregate parents are left behind by design, not silently cleaned up.
- Grepped the changed/new files for `noqa`, `type: ignore`, `pytest.mark.skip`, `xfail` — none found, consistent with CLAUDE.md §1.3/§6.2 (no bypass shortcuts).

## Security Concerns
🟢 LOW — none identified. Pure data-shape transformation over already-trusted export JSON; no new I/O, subprocess, or path handling.

## Problems
None found that block merge. I traced `group_turn_runs`'s state machine by hand against every documented behavior (leading-assistant drop, trailing-human drop, non-participating-role pass-through mid-run, multi-turn re-pairing) and it matches both the tests and the docstring in every case I checked, including cases not explicitly tested (e.g., a non-participating role appearing *inside* an already-open human or assistant run).

## Code Quality
- `creek/ingest/claude.py::_collapse_run`: for the assistant side, the carrier is `assistant_run[0]` (first message), while the human side uses `human_run[-1]` (last message) — asymmetric on purpose per the docstring (`content` is fully replaced by the merge either way, and no code currently reads other per-message fields off the assistant dict downstream). Correct today, but it's a latent sharp edge: if a future change reads e.g. `assistant_msg.get("id")` or `.get("stop_reason")` off the collapsed dict, it'll silently get the *first* message's value for a merged run. Not worth blocking on, but a one-line comment already covers the timestamp case — might as well cover this one too.
- Not part of this diff, but noticed in passing: `claude.py`'s `convert_to_markdown` renders a blank line as `"> "` (trailing space) while `chatgpt.py`'s renders it as `">"` (no trailing space) — a pre-existing inconsistency between the two ingestors, unaffected by this change and not worth raising as a fix here.

## CLAUDE.md Adherence
✅ TDD workflow: PR description documents red→green with a per-file mutation-testing breakdown (revert `claude.py` alone → 2 red, `chatgpt.py` alone → 3 red, separator swap → 7 red, etc.) — I couldn't execute this myself (sandbox blocks running pytest/scripts in this review environment), but the trace-through of the code against the tests as written is consistent with these claims.
✅ No forbidden shortcuts (verified via grep — see Strengths).
✅ Conventional commit + Closes #1333, follow-ups filed as separate issues (#1426–#1428) rather than scope creep.
✅ Docs updated (`docs/ingestion.md`) alongside the behavior change.
⚠️ I was unable to independently run `./scripts/check-all.sh` in this review session (tool execution was blocked by the sandbox's approval gate), so the claimed "9377 passed, 94.60% coverage" is not independently verified here — CI should be treated as the authoritative confirmation before merge, per Gate 3.

## Testing
Strong. New `tests/test_ingest_turns.py` pins the shared primitives directly (byte-identity invariant, blank-line separator, indented-code-block case, leading-assistant/trailing-human drop, non-participating-role pass-through). `test_ingest_claude.py` and `test_chatgpt_ingest.py` add turn-level integration tests for both the human-run and assistant-run merge cases plus the ChatGPT `system`-in-the-middle case that used to yield zero fragments. `test_ingest_resplit.py` adds a regression test tying the merge format to the `--refresh-ai-chat` safety invariant. The rewritten `test_consecutive_human_messages` now asserts real content instead of the previous no-op assertion.

## Documentation
`docs/ingestion.md` gets a new "Split messages are one turn" section covering the behavior change, the recovery procedure, and its residue (near-duplicate fragments, orphaned aggregate parents) — matches CLAUDE.md's expectation that behavior changes ship with doc updates. New/changed docstrings are thorough (arguably more prose than usual, but accurate and directly tied to the bug history, not filler).



## Verdict: LGTM

No correctness, security, or CLAUDE.md-adherence issues survived a careful hand-trace of the new `group_turn_runs`/`merge_turn_texts` logic against its tests and the id-stability/resplit-safety claims made in the PR description (both verified directly against `generate_fragment_id` and `_parse_claude_merged`); the one code-quality note (asymmetric carrier selection for merged assistant runs) is a latent nit with no live effect, not something worth blocking on.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5260768346 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1432, opened 2026-08-12
fix(generate): note-path resolvers follow dangling symlinks and race between exists() and write


Geoffe-Ga:
**Found by:** the security pass on the #1334 lane.

## Problem

#1334 gave `create_decision_note` and `create_compost_note` an ownership-resolving note-path helper. Both resolvers decide a path is available with `Path.exists()` and then write it with `write_text`:

```python
note_path = target_dir / f"{stem}{suffix}.md"
if not note_path.exists() or _recorded_..._source(note_path) == identity:
    return note_path
...
note_path.write_text(frontmatter.dumps(post), encoding="utf-8")
```

Two consequences follow from `exists()` + `write_text`, neither of which #1334 introduced but one of which it **widens**:

**1. A dangling symlink reads as free and is written through.** `Path.exists()` follows symlinks and returns `False` for a broken one, so a symlink at a candidate path is judged available; `write_text` then follows it and creates the file at the link target — potentially outside the vault. Before #1334 this was reachable only at the natural (ordinal-0) path; the resolver now probes `-1`, `-2`, … too, so the surface is wider.

Worth recording the counterweight: for an *occupied* symlink #1334 is a net **improvement** — the identity check now fails and the resolver steps over it, where the old code wrote straight through and destroyed the target.

**2. TOCTOU.** The gap between `exists()` and `write_text` is a real race. It is a **correctness** problem, not a security boundary — every principal that could win the race already has vault write access, and the write stays inside `target_dir` regardless. It bites under two concurrent `creek` processes on one vault, or a sync client (Obsidian Sync / Dropbox / iCloud) materialising a file mid-run. The outcome is a lost note.

Severity is low: both require write access to the vault directory, which is already inside the trust boundary. Path traversal itself is **closed** — `_sanitize_title` / `_sanitize_filename` strip `.`, `/`, `\` and NUL via `re.sub(r"[^\w\s-]", "", title)`, and both empty-title fallbacks are date-prefixed, so no bare `.md`, dot-entry or leading dash is reachable.

## Suggested fix

One change closes both: replace `exists()` + `write_text` on the *free* branch with an exclusive, non-following create — `os.open(path, os.O_CREAT | os.O_EXCL | os.O_WRONLY | os.O_NOFOLLOW)` — and treat `FileExistsError` as "occupied, go check ownership" and `OSError/ELOOP` as "not mine, advance".

[...]

Issue #1431, opened 2026-08-12
fix(decisions): creek fill writes an intimate fragment's title into an 08-Decisions filename


Geoffe-Ga:
**Found by:** the security pass on the #1334 lane.

## Problem

`generate_decisions`' `override` parameter defaults to `PrivacyTierOverride.ALL`, and `within_ceiling(raw, ALL)` short-circuits to `True` — "no ceiling declared". The MCP report surface always states an override explicitly (`creek_mcp/tools/report.py`, pinned by `tests/test_mcp_report_tier_ceiling.py::test_production_report_callers_always_state_an_override`), but **`creek fill` passes `unfiltered = PrivacyTierOverride.ALL` deliberately** (`creek/cli.py`).

So on the flagship command, an **intimate-tier fragment's title is sanitised into a filename** at `08-Decisions/Active/<date>-<title>.md`, and into the note's `title:` frontmatter.

`generate_decisions`' own docstring already flags the stakes — "a generated note's `title:` frontmatter *and* its filename are a source fragment's title verbatim" — and `creek/generate/tags.py` carries a related acknowledgement. But flagging is not enforcing.

## Why this is worth an issue

This is the exact channel **#1311 / PR #1404** closed on the *compost* side: intimate-derived strings must not reach the filesystem, because a filename is visible in Finder, in Obsidian's quick-switcher, in backups, and in any sync client's metadata — regardless of what the note body says.

Compost fixed it structurally: `CompostTracker._screen` removes withheld fragments at the detection boundary, and `run_compost_scan` forces `skip_intimate=True` non-defeasibly. **Decisions has no equivalent.** The two sibling generators disagree about whether an intimate title may become a filename.

#1334 did **not** change this either way — the pre-fix code had the same behaviour, and #1334's ordinal disambiguator is provably identity-free. This is the surrounding gap, filed separately so #1334 stays a bug fix.

## Scope

Decide and enforce one answer for decisions:

- **Option A** — screen at the detection boundary, mirroring `CompostTracker._screen`, so an intimate fragment never becomes a decision candidate at all.
- **Option B** — admit the candidate but write a title-free filename/frontmatter for intimate tiers, mirroring the `creek save` intimate-stub shape.

Option A matches the compost precedent and is the smaller change. Note that `privacy_tier` is a one-way ratchet in this repo: check the revisit predicate before choosing a fail-closed default.

[...]

Issue #1430, opened 2026-08-12
fix(decisions): update_decision_phase renames over a same-named archived note, destroying it


Geoffe-Ga:
**Found by:** the #1334 lane (Gate 2.5 self-review), while fixing the sibling defect in `create_decision_note`.

## Problem

`DecisionDetector.update_decision_phase` moves a note between `08-Decisions/Active/` and `08-Decisions/Archive/` with a bare rename:

```python
if source_path.parent != target_dir:
    dest_path = target_dir / source_path.name
    source_path.rename(dest_path)
    return dest_path
```

`Path.rename` **silently replaces** an existing destination on POSIX. If `Archive/` already holds a note with the same filename, archiving the second one destroys the first — no warning, no return-value signal.

This is the **same bug class as #1334**: a path derived from `<date>-<sanitised-title>` is treated as though it were an identity, and the write is unguarded. #1334 fixed it for the two `create_*_note` writers; this rename was deliberately left out of that PR's scope.

## Reachability

Two same-titled decisions opened on the same day now legitimately coexist as `<date>-<title>.md` and `<date>-<title>-1.md` (that is #1334's fix). Archive them in the wrong order and no collision occurs — but the older path is also *freed* in `Active/` when the first is archived, so a later decision can take the natural stem and then collide in `Archive/` with the one already there. #1334 makes that sequence marginally more reachable, not less.

There is no CLI or MCP surface for `update_decision_phase` today, so this is currently library-only.

## Suggested fix

Reuse the shape #1334 established: resolve the destination by **ownership**, not occupancy. `decisions._resolve_decision_note_path(target_dir, source_path.stem, fragment_id)` already does exactly this — probe until a path is free or already records this decision's source fragment, never overwrite a stranger. The extractor `_recorded_decision_source` is public-within-module and ready.

Do **not** reach for `VaultWriter._atomic_create`'s counter-suffix (see #1417's Scope and #1320): it advances on occupancy and would mint a new archive copy on every call.

## Acceptance criteria

- [ ] Archiving a decision never overwrites a differently-identified note already in `Archive/`.
- [ ] A test seeds a same-named stranger in `Archive/`, archives, and asserts the stranger's **bytes are unchanged** and both notes exist.
- [ ] `./scripts/check-all.sh` exits 0.

| File | Action |
|------|--------|
| `creek-tools/creek/generate/decisions.py` | Modify |
| `creek-tools/tests/test_decisions.py` | Modify |

PR #1433, opened 2026-08-12
fix(generate): resolve note paths by identity ownership, not by name (#1334)

### LINKED ISSUE #1334: fix(generate): decision and compost notes are keyed on <today>-<title>, so same-titled candidates overwrite each other and never converge
**Severity:** High
**Remediation:** fix in place

## Problem

Two liminal generators name their output files `<today>-<sanitized title>.md`
and index them by something else, so two candidates that share a title collide:

- `creek/generate/decisions.py:220` (`create_decision_note`), read back by
  `_existing_decision_fragment_ids` (`decisions.py:1003`)
- `creek/generate/compost.py:504` (`create_compost_note`), read back by
  `load_composted_source_ids` (`compost_scan.py:144`)

Family 0 (correctness — silent loss + non-convergence). Same class as the
already-filed #1320 (paradox notes duplicating per calendar day), different
modules and a different failure mode: here the *second* candidate overwrites the
first, and the read-back index then only sees whichever survived, so the
generator never converges.

**Current state:** one of the two source fragments loses its note entirely, and
the surviving note is rewritten on every run with a fresh `decision-<uuid>` /
compost id — breaking any `[[decision-…]]` wikilink an operator has made. For
compost, when a verifier is wired, every oscillating run re-spends an LLM call.

## Evidence (corroboration)

**Signal 1 — reproducing artifact (decisions).** Two fragments, same title,
three consecutive runs:

```
run1: written=2 unique_paths=1 files=['2026-08-10-Should-I-move-to-Portland.md'] src=['- frag-b']
run2: written=1 unique_paths=1 files=['2026-08-10-Should-I-move-to-Portland.md'] src=['- frag-a']
run3: written=1 unique_paths=1 files=['2026-08-10-Should-I-move-to-Portland.md'] src=['- frag-b']

E AssertionError: expected 2 decision notes, got ['2026-08-10-Should-I-move-to-Portland.md']
```

`generate_decisions` returned two paths on run 1 that are the **same** `Path`,
and the source bullet flips b -> a -> b forever.

**Signal 1b — same shape for compost.** `run_compost_scan` with
`similarity_fn=lambda t: 0.99` and two fragments both titled `Letting this go`:

```
run1: review_queued=2 skipped=0 files=['2026-08-10-Letting-this-go.md'] sources=['[[frag-b]]']
run2: review_queued=1 skipped=1 files=['2026-08-10-Letting-this-go.md'] sources=['[[frag-a]]']
run3: review_queued=1 skipped=1 files=['2026-08-10-Letting-this-go.md'] sources=['[[frag-b]]']
```

**Signal 2 — reading/intent.** `generate_decisions`' docstring
(`decisions.py:1050-1051`) states it "writes one Decision note per candidate …
so re-running is idempotent and never duplicates notes for one fragment" — both
halves are false. `run_compost_scan`'s docstring (`compost.py:364-366`) says
dropped candidates are "what makes a re-scan idempotent and keeps a second run
from re-spending LLM calls". `compost_scan.py`'s module docstring additionally
pins "Costs are quoted before they are incurred", which also breaks:
`plan.llm_calls` counts `fresh` fragment candidates, so a colliding pair charges
two calls and writes one note. Note the repo already owns the right primitive —
`VaultWriter._atomic_create`'s counter-suffix retry.

## Scope

The filename derivation in the two `create_*_note` functions. Does **not** cover
paradox (already #1320) — but the fix should be shaped so #1320 can adopt it.

## Tasks

1. Include the stable source id in the filename stem (e.g.
   `<date>-<slug>-<source-id-suffix>.md`), or route the write through a
   collision-retrying create, so two same-titled candidates cannot share a path.
2. Apply the same change in `creek/generate/decisions.py:220` and
   `creek/generate/compost.py:504`.
3. Add tests: two same-titled fragments produce two distinct notes; runs 2 and 3
   write nothing (`skipped_existing == 2` for compost); `generate_decisions`
   returns distinct paths.
4. Correct the two docstrings' idempotency claims to match the implementation.

## Acceptance Criteria

- [ ] The decisions repro yields two distinct notes and run 2 writes nothing.
- [ ] The compost repro yields `skipped_existing == 2` on run 2 and two notes on disk.
- [ ] A note's `id:` is stable across runs.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/generate/decisions.py` | Modify |
| `creek-tools/creek/generate/compost.py` | Modify |
| `creek-tools/tests/test_decisions.py` | Modify |
| `creek-tools/tests/test_compost_scan.py` | Modify |

Geoffe-Ga:
## Summary

`create_decision_note` and `create_compost_note` both ended in an unconditional `note_path.write_text(...)` at a path derived from `<date>-<sanitised title>`, while their read-back indexes keyed on something else entirely — the source fragment id, and the `original_*` frontmatter pair. **A title is not an identity.** Two candidates sharing one destroyed the other, the index then only saw the survivor, and the generator oscillated forever.

Reproduced at base `b55ed1b`, three consecutive runs, real listings:

```
DECISIONS
run1: written=2 unique_paths=1 files=['2026-08-11-Should-I-move-to-Portland.md'] src=['- frag-b'] ids=['decision-31eecda0']
run2: written=1 unique_paths=1 files=['2026-08-11-Should-I-move-to-Portland.md'] src=['- frag-a'] ids=['decision-1dbcf6ef']
run3: written=1 unique_paths=1 files=['2026-08-11-Should-I-move-to-Portland.md'] src=['- frag-b'] ids=['decision-0be4a97a']

COMPOST  (similarity_fn=lambda t: 0.99, both fragments titled "Letting this go")
run1: review_queued=2 skipped=0 files=['2026-04-01-Letting-this-go.md'] sources=['[[frag-b]]']
run2: review_queued=1 skipped=1 files=['2026-04-01-Letting-this-go.md'] sources=['[[frag-a]]']
run3: review_queued=1 skipped=1 files=['2026-04-01-Letting-this-go.md'] sources=['[[frag-b]]']
```

`generate_decisions` returns two paths on run 1 that are the **same** `Path`, and the source bullet flips `b -> a -> b` forever.

### The fix is in the write, not in the name

The name being non-unique is only *how* the collision is reached; the loss happens because the writer clobbers a note it does not own. Both writers now resolve their path by **identity ownership**:

1. natural name free → write there;
2. occupied by a note recording **the same identity** → write there (a legitimate refresh);
3. occupied by a **different** identity, or by a note whose identity cannot be *positively established* → advance an ordinal (`-1`, `-2`, …) and re-apply the rule.

**Why this is categorically not `VaultWriter._atomic_create`'s counter-suffix**, which #1320/#1418 and #1417's Scope rule out: `_atomic_create` asks *"is this path free?"* — a function of history, so it mints a file every run and never converges. This asks *"is this path mine?"* — a function of the candidate, so the file count is bounded by the number of distinct identities. **Branch 2 is precisely the branch `_atomic_create` structurally lacks**, and it is pinned by `test_recreating_the_same_candidate_refreshes_the_one_note` in both test modules, which a naive port of the occupancy counter would break.

**No identity-derived string reaches a filename**, per #1311/PR #1404. The disambiguator is a bare ordinal integer — provably a *count of foreign occupants* and nothing else. `fragment_id` / the compost identity pair appear in the resolvers exactly once each, inside a boolean comparison that can only advance or stop the loop; neither is ever concatenated into a path. This matters most for a compost **project** candidate, where `source_id` *is* the tag and can be intimate-derived.

**No existing note is renamed.** The natural name is unchanged for the ordinary one-candidate-per-title case, so no `[[wikilink]]` breaks and nothing needs migrating.

### The identity is a *pair*, not a bare id

`_unwrap_source_id` strips `[[ ]]`, so `original_fragment: [[X]]` and `original_project: X` both yield `"X"`. Comparing through it would let a project tagged `X` judge itself the owner of the note recording fragment `X` and overwrite it. The resolver therefore compares `(source_key, raw_value)` as written. `_candidate_source_pair` is now the single place `source_type` maps onto a key — `_build_metadata` writes the pair, `_compost_source_of` reads it back — so writer and reader cannot drift, which is the failure class this issue *is*.

### Line cites, and what moved under #1404

`decisions.py:220` and `:1003` were exact. The compost cites were stale — `compost.py:504` is now `:650` (`create_compost_note` at `:607`) and `compost_scan.py:144` is now `:152`. Both were exact against the pre-#1404 tree; `dd58593` shifted them without touching the defect.

### The consequences the issue names — two confirmed, one overstated

- **A source fragment loses its note entirely.** Confirmed.
- **Compost re-spends an LLM call every oscillating run.** Confirmed with a counting verifier: 2 candidates cost 2 calls on run 1 and **one more on every run thereafter, forever** (5 runs → 6 calls).
- **"breaking any `[[decision-…]]` wikilink"** — **overstated.** `vault/links.py:176` registers each note under its stem, `title` and `aliases`, never `id:`, so a `[[decision-<uuid>]]` link was never resolvable. The real harm is worse and is confirmed: the surviving file is destroyed and replaced every run with a fresh draft belonging to the *other* candidate. An operator edit measured directly — `status: deliberating` reverted to `sensing`, their prose gone, `id:` churned `765ea527 → 2f7b9e99`. `_find_decision_by_id` can no longer find the decision either.

### Migration — the vault self-heals; nothing is deleted

This repo has chosen detect-and-report-never-delete six times, but the right answer here is better and simpler: **there is nothing to report.** The damage is an *absence*, not a stray on disk — nothing to census, no self-clearing predicate. The lost candidate is absent from the index, so the next run re-detects it and files it at the next free ordinal; the survivor **is** in the index, so it is skipped and left byte-for-byte alone. Pinned by `test_the_survivor_is_untouched_and_the_lost_note_reappears`, which asserts the survivor's bytes, `id:`, `status: deliberating` and operator prose all survive while the missing note reappears at `-1`, and that the run after that writes nothing.

What genuinely cannot be recovered — prose in copies destroyed by earlier runs, and any `decision-…` id recorded before it churned — is undetectable from disk, and the PR says so rather than guessing.

### On #1417

Worth doing, and cheaper now. Each module needed its identity extractor callable per-note (for the writer) as well as per-folder (for the index), so both now match the `paradox.py::_recorded_pair` + folder-scan shape #1417 wants to hoist. The ordinal loop is **deliberately duplicated** with an in-code comment reserving the hoist — #1417's Scope forbids a shared *filename builder*, and only the read-back scan is shareable.

### Two widenings beyond the strict bug, both argued

- `decisions._load_post` now catches `yaml.YAMLError`. Required: `_existing_decision_fragment_ids` carried that handler inline and now routes through the shared helper. Its docstring explicitly **declines** to claim the tuple is exhaustive — a non-string YAML key still raises `TypeError` past it; that gap is pre-existing and belongs on #1416.
- `compost.py::_load_existing_compost_notes` / `_load_active_threads` likewise. This one is caused by the fix: the resolver now **preserves** malformed neighbours instead of clobbering them, so `creek fill`'s report run newly meets them. Without it both parametrizations of the new test die on a real `yaml.parser.ParserError`.

### Docs

`docs/decisions.md` updated. `docs/emergence.md` claimed compost notes are named `<id>.md` — **wrong before this change and independent of it**; corrected.

### One caveat stated rather than hidden

"A note this candidate already owns is refreshed in place" is true **at today's stem only** — the stem embeds the date, so a direct call on a later day writes a second note rather than refreshing. `paradox.py` carries the same caveat for the same reason (#1320). Production is unaffected: the index spans every date and both folders, so the candidate is skipped before a stem is built. Both `create_*_note` docstrings now say this outright.

### Follow-ups filed (none block this)

- **#1430** (P2) — `update_decision_phase` renames over a same-named archived note. Same bug class as this one, deliberately out of scope.
- **#1431** (P2) — `creek fill` writes an intimate fragment's title into an `08-Decisions/` filename. Pre-existing and unchanged here; the compost side was closed by #1404 and decisions never got the equivalent.
- **#1432** (P3) — the resolvers follow dangling symlinks and race between `exists()` and write. `O_CREAT|O_EXCL|O_NOFOLLOW` closes both.
- Commented on **#1416** — the `(OSError, ValueError, yaml.YAMLError)` tuple still misses `TypeError` from a non-string YAML key; this PR's `_load_post` docstring explicitly declines to claim otherwise.
- Commented on **#1417** — its two constraints held, and the hoist is now cheaper because all three generators finally share the per-note-extractor shape.

## Test plan

`./scripts/check-all.sh` → **exit 0**, 12/12 gates, 9408 passed / 5 skipped (pre-existing Ollama + live-key skips), coverage 94.63%. Ruff re-verified with `--no-cache` (#1119).

31 new tests, **zero skipped, zero existing assertions touched, zero lines deleted from any test file**. RED proven at `b55ed1b` first: **25 behavioural failures**, no import errors.

Coverage includes the three-run oscillation per generator (asserting the set of *identities recorded on disk*, not file counts — two notes both recording `frag-b` would satisfy `len(files) == 2` and still be the bug), `id:` stability, the damaged-vault self-heal, the 80-char truncation collision, both empty-title fallbacks, malformed/identity-less neighbours (asserted **byte-identical** afterwards), a title literally ending `-1`, the fragment-`X`-vs-project-`X` cross-type collision, the ordinal cap, and the LLM re-spend. Identity assertions read raw bytes via `frontmatter.load` rather than the module's own extractor, so a self-healing reader cannot make them vacuous.

**Mutation-tested — 9 mutants, every one killed, each restored:**

| Mutant | Kills |
|---|---|
| decisions ownership predicate inverted | 11 |
| decisions occupancy-only (`_atomic_create` semantics) | 1 — the refresh guard |
| decisions resolver bypassed (original bug) | 10 |
| decisions ordinal cap forced to 1 | 9 |
| compost ownership predicate inverted | 14 |
| compost reader collapsed | 1 |
| compost resolver bypassed (original bug) | 13 |
| compost identity pair collapsed in the comparison only | 1 — the cross-type guard |
| report `yaml.YAMLError` tuple narrowed back | 2 |

The single-kill mutants are the load-bearing ones, because each isolates one guard: "occupancy-only" proves the `_atomic_create` semantics are genuinely excluded, and "pair collapsed in the comparison only" proves the identity really is compared as a pair. An earlier, cruder version of that last mutant collapsed the writer and the reader at once and killed 22 tests for the wrong reason (it also changed what `_build_metadata` writes) — it was discarded and replaced with one that touches only the resolver's comparison.

Gate 2.5 self-review returned no blockers and four docstring/comment corrections, all applied: a `run_compost_scan` sentence that read as though a write could *still* destroy another source's note, the same-day caveat above, six stale `file:line` cites in test docstrings (replaced with symbol names, since line numbers rot — the diff now contains none), and one reference to `_SOURCE_KEYS` under its pre-move module. One nit is knowingly left: four tests derive a stem from `date.today()` before the call under test, a ~10⁻⁶ midnight-straddle window that the test author documented as a considered trade-off.

Closes #1334


Geoffe-Ga:
## Summary
Fixes #1334: `create_decision_note` and `create_compost_note` named output files `<date>-<sanitised-title>.md`, while their read-back indexes keyed on source-fragment identity. Two candidates sharing a title collided on one filename, the second `write_text` silently destroyed the first, the index only ever saw the survivor, and the generator oscillated forever (re-minting IDs, re-spending LLM calls, discarding operator edits). The fix resolves note paths by *identity ownership*: probe `stem`, `stem-1`, `stem-2`, … and take the first path that is free or already records the candidate's own identity; a path recording a different (or unestablishable) identity is skipped, never overwritten. Docs, tests, and two related exception-handling widenings (`yaml.YAMLError`) round out the change.

## Strengths
- **Root-cause fix, not a name-tweak.** Correctly identifies that the defect is in the *write* (occupancy-blind overwrite), not in filename generation, and explicitly rules out reusing `VaultWriter._atomic_create`'s counter-suffix (which advances on occupancy and would never converge) in favor of an ownership predicate bounded by distinct identities.
- **Identity is a pair, not a bare id.** Correctly recognizes that `_unwrap_source_id` collapses `original_fragment: [[X]]` and `original_project: X` onto the same string, and compares `(source_key, raw_value)` to keep a fragment and a project with the same tag from clobbering each other — pinned by `test_a_fragment_and_a_project_sharing_an_id_get_separate_notes` / `test_a_settled_vault_stops_re_spending_llm_calls`.
- **Self-healing migration with no destructive cleanup.** Rather than scanning-and-deleting, the fix lets an already-damaged vault heal naturally: the survivor is left byte-for-byte alone, the previously-lost candidate reappears at the next free ordinal. Verified by `test_the_survivor_is_untouched_and_the_lost_note_reappears` / `test_a_vault_already_in_the_post_collision_state_heals_without_loss`.
- **Excellent regression coverage**, including a dedicated test that guards against the "tempting shortcut" of reusing `_atomic_create` semantics (`test_recreating_the_same_candidate_refreshes_the_one_note`), an 80-char truncation collision, both empty-title fallbacks, a title that sanitises to another identity's ordinal-suffixed name, and a bounded/loudly-failing ordinal cap.
- **Honest scope discipline.** Explicitly declines to fix adjacent bugs (#1430, #1431, #1432) and instead files them, and states the one caveat it doesn't close ("refresh" is keyed to today's stem only) rather than hiding it.
- Docs (`docs/decisions.md`, `docs/emergence.md`) updated to match, including a stale claim in `emergence.md` that predates this PR.

## Security Concerns
🟢 LOW — None found. The resolver only ever probes deterministic, non-identity-derived filenames (a bare ordinal), consistent with the #1311 constraint against leaking intimate data into filenames. The known `exists()`/write TOCTOU race and dangling-symlink following are acknowledged and filed as follow-up #1432 rather than silently left; reasonable to defer given this is a local single-writer CLI pipeline, not a multi-tenant service.

## Problems
None blocking.

## Code Quality
- `compost_scan.py` now imports `_recorded_compost_source` and `_unwrap_source_id` — both underscore-prefixed — from `compost.py`. This is a deliberate DRY move (writer and reader share one identity extractor so they can't drift) and is well-justified in the docstrings, but crossing a module boundary via private names is worth a second look if `check-all.sh`'s pylint/ruff private-access rules don't already catch it.
- The ordinal-probing loop (`_resolve_compost_note_path` / `_resolve_decision_note_path`) is intentionally duplicated between `compost.py` and `decisions.py` rather than hoisted, with an inline comment reserving that for #1417. Reasonable given the stated scope boundary, but two independent implementations of the same probe-and-compare logic are two places a future edit can drift.
- Docstrings are unusually long (multi-paragraph justifications embedded in `Args`/`Returns`-style docstrings). Functionally fine and clearly intentional given the review-friendliness goals stated in the PR description, but denser than the repo's typical docstring style.
- `create_decision_note`'s "refresh" branch is documented as destructive (mints a fresh `id:`, resets `status:` to `sensing`) for direct callers outside the index-gated `generate_decisions` path. This is pre-existing behavior, not a regression, and is now explicitly documented — but it remains a footgun for any future direct caller; worth a tracked follow-up if not already covered by #1417's hoist.

## CLAUDE.md Adherence
✅ TDD — commit message and PR description describe RED-first development (25 failing behavioural tests at base commit) before the GREEN implementation.
✅ Coverage — PR reports 94.63% aggregate coverage with `./scripts/check-all.sh` exit 0, well above the 90% floor; 31 new tests added, zero existing assertions touched.
✅ Complexity — new functions (`_resolve_compost_note_path`, `_resolve_decision_note_path`, `_compost_source_of`, `_decision_source_fragment_id`) are small, single-purpose loops/branches, well under the 10-complexity ceiling.
✅ Docstrings — Google-style, Args/Returns/Raises present throughout; docstring coverage claim (part of the 12/12 gate report) is plausible given the density observed.
✅ No forbidden shortcuts — no `# noqa`, `# type: ignore`, or skipped tests observed in the diff.
✅ Commit conventions — `fix(generate): ...` conventional-commit message, references issue #1334, co-authored trailer present.
I could not independently execute `./scripts/check-all.sh` in this review environment (no `uv`/pytest available), so the specific pass/fail numbers are taken on the PR author's report rather than independently reproduced — flagged for transparency, not as a blocker given the depth of test evidence in the diff itself.

## Testing
Strong. Both `compost.py` and `decisions.py` get parallel test suites covering: the headline 3-run oscillation (asserting the *set of identities on disk*, not just file counts, which correctly avoids a vacuous pass), same-identity refresh (regression guard against the `_atomic_create`-style shortcut), truncation-induced collisions, empty-title collisions, ordinal-suffix hijacking, a damaged-neighbour matrix (malformed YAML, foreign `type:`, no source key) asserted byte-identical after a write, ordinal-cap exhaustion and boundary, and vault self-heal from a pre-existing damaged state. Identity is read back via raw `frontmatter.load` rather than the module's own extractors, correctly avoiding a self-healing-reader trap that would make assertions vacuous. Reported mutation testing (9/9 mutants killed, including single-kill mutants isolating the refresh guard and the pair-vs-bare-id comparison) is a good-faith signal of assertion strength, though not independently re-run here.

## Documentation
Both `docs/decisions.md` and `docs/emergence.md` updated to reflect the `[-N]` ordinal suffix and the ownership-resolution behavior. The `emergence.md` fix (correcting a `<id>.md` claim that was already wrong pre-PR) is a reasonable drive-by correction directly relevant to the area touched. Docstrings on all new/changed functions are thorough to the point of restating design rationale inline — good for this review, arguably more than typical inline documentation needs going forward.



## Verdict: LGTM

The fix correctly replaces path-occupancy filename generation with identity-ownership resolution, closing a real data-loss bug (colliding titles clobbering notes and causing infinite oscillation); it's backed by thorough new tests (three-run oscillation, damaged-vault self-heal, cross-type identity collisions, ordinal-cap boundaries, malformed-neighbour matrix) and reported mutation-testing evidence, with no correctness, security, or coverage issues found. Only minor non-blocking style notes (verbose docstrings, cross-module import of underscore-prefixed helpers) — neither rises to a reason to withhold merge.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5261878664 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1437, opened 2026-08-12
fix(sync): render a Tier-A cadence systemd and cron can actually keep

### LINKED ISSUE #1336: fix(sync): tier_a_interval_minutes >= 60 emits an OnCalendar systemd refuses to parse, so Tier A never runs
**Severity:** High
**Remediation:** fix in place

## Problem

`SyncConfig.tier_a_interval_minutes` is validated only as `ge=1`
(`creek/config.py:1644`), but the systemd and cron renderers interpolate it into
a **minute field**, which is 0–59:

- `creek/sync/schedule.py:177` → `OnCalendar=*:0/{tier_a_minutes}`
- `creek/sync/schedule.py:200` → `*/{tier_a_minutes} * * * *`

Family 0 (boundary error) + family 11 (config hygiene).

**Current state:** any hourly-or-slower cadence — a natural choice the config's
own docs invite — emits a `.timer` systemd refuses to parse, so
`systemctl --user enable --now creek-sync-tier-a.timer` fails and Tier A
silently never runs. The launchd renderer (`StartInterval` seconds) handles the
same value correctly, so the three adapters disagree about one config key.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Rendered output vs. `systemd-analyze` on
this host:

```
tier_a_minutes=  30 -> ['OnCalendar=*:0/30']
tier_a_minutes=  59 -> ['OnCalendar=*:0/59']
tier_a_minutes=  60 -> ['OnCalendar=*:0/60']
tier_a_minutes= 120 -> ['OnCalendar=*:0/120']

*:0/30       Original form: *:0/30
*:0/59       Original form: *:0/59
*:0/60     Failed to parse calendar specification '*:0/60': Invalid argument
*:0/120    Failed to parse calendar specification '*:0/120': Invalid argument
```

**Signal 2 — reading/intent.** `render_systemd_units`' docstring promises "Tier A
fires every `tier_a_minutes`". For the whole upper half of the accepted domain it
fires never. `tests/test_sync_schedule.py` only ever passes 15 and 30
(`:86`, `:104`, `:125`) and asserts string equality; the file docstring claims it
"verifies the emitted units are valid", but only launchd is actually parsed
(`ET.fromstring`, `:40-41`). The module docstring frames these as artifacts the
operator installs manually, which is exactly why the emitter is the last place
that can catch the value.

## Scope

The systemd and cron renderers plus the config bound. Does not change launchd.

## Tasks

1. Either bound `SyncConfig.tier_a_interval_minutes` at `le=59`, or (preferred)
   render whole-hour intervals as an hour spec — e.g.
   `OnCalendar=*-*-* 0/{hours}:00:00` and `0 */{hours} * * *` — and reject values
   that are neither `<60` nor a whole number of hours.
2. Apply the same to the cron renderer (`schedule.py:200`).
3. Extend `tests/test_sync_schedule.py` to cover 59, 60, 90 and 120, and to
   actually validate the systemd unit (parse the `OnCalendar` value rather than
   comparing strings).

## Acceptance Criteria

- [ ] `render_systemd_units(..., tier_a_minutes=60)` emits a spec
      `systemd-analyze calendar` accepts (or the config rejects the value at load).
- [ ] Boundary values 59/60 are both covered by tests.
- [ ] The launchd, systemd and cron adapters agree on the same config value.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/sync/schedule.py` | Modify |
| `creek-tools/creek/config.py` | Modify |
| `creek-tools/tests/test_sync_schedule.py` | Modify |

Geoffe-Ga:
## Summary

`sync.tier_a_interval_minutes` was interpolated straight into a systemd minute field (`OnCalendar=*:0/N`, `schedule.py:177`) and a cron minute field (`*/N * * * *`, `:200`). Both fields are 0–59, and a repetition step restarts at the enclosing field's boundary.

**The issue understated the defect.** Measured on real systemd 257.13-1~deb13u1 and vixie cron 3.0pl1-197 (Debian trixie container, TZ=UTC — observed, not derived):

| N | systemd `*:0/N` | cron `*/N` | launchd |
|---|---|---|---|
| 30 | accepted, uniform 30 min | accepted, correct | 1800 s, exact |
| **59** | **accepted — fires :00 and :59** (gaps 59, 1) | **accepted, same wrong cadence** | 3540 s, exact |
| 60 | rejected: `Failed to parse calendar specification` | rejected: `Step size 60 higher than possible maximum of 59` | 3600 s, exact |
| 90 | rejected | rejected | 5400 s, exact |
| 1440 | rejected | rejected | 86400 s, exact |

The filed issue names only the loud half (`>= 60`). The silent half — `{25, 40, 45, 50, 59}` — parses fine and keeps the wrong cadence with nothing to notice: `*:0/45` fires at :00 and :45 of every hour, 45 minutes then 15. **The real boundary is not `>= 60`; it is "the step must divide its field".**

At install, systemd *ignores* the bad directive, leaving the timer with no trigger, then refuses the unit (`Timer unit lacks value setting. Refusing.`). So `enable --now` does fail loudly — but only after `creek sync --install-schedule` already reported success and printed the activation command.

**The issue's own preferred remedy is also defective.** `*-*-* 0/H:00:00` is faithful only when H divides 24: H=5 gives 00,05,10,15,20 with a four-hour gap at midnight; H=7 gives 07,14,21,00; and H=24 (a daily cadence) is **rejected outright**. Rendering whole hours as an hour step, exactly as filed, would have shipped a new silent-wrongness bug.

### The fix

One private classifier decides the form, and `render_systemd_units` becomes **total**:

| cadence | systemd | cron |
|---|---|---|
| divides 60 (1,2,3,4,5,6,10,12,15,20,30) | `OnCalendar=*:0/N` + `Persistent=true` | `*/N * * * *` |
| whole hours dividing 24 (60…720) | `OnCalendar=*-*-* 0/H:00:00` + `Persistent=true` | `0 */H * * *` |
| daily (1440) | `OnCalendar=*-*-* 00:00:00` + `Persistent=true` | `0 0 * * *` |
| anything else (45, 90, 300, 2880…) | monotonic `OnBootSec`/`OnUnitActiveSec`, no `Persistent=` | **refused, with an explanation** |

Daily is a separate case on purpose: 24 divides 24, so a naive divisor rule would emit `0/24:00:00`, which systemd rejects — reintroducing the bug through the fix.

**Interpolation vs. validation.** Guarding at the primitive, not per caller (precedent: #1306, #1311, #1332). Validating the rendered output — e.g. shelling out to `systemd-analyze` at install — was considered and rejected: `*:0/45` *parses fine*, so a grammar check is blind to the worse, silent half and only catches what already fails loudly at `enable`. Faithfulness is decidable only where intent (minutes) becomes representation. It would also break the module's stated invariant that Creek never shells out to `launchctl`/`systemctl`/`crontab`, and would be unavailable on macOS and in CI.

Rather than fail on inexpressible cadences, the systemd renderer was made able to express all of them: the monotonic pair is exact at any interval and is the precise analogue of launchd's `StartInterval`, so the three adapters finally agree. It omits `Persistent=` (inert without `OnCalendar=`) and says in a `#` comment what that costs. **At the default N=30 the emitted bytes are unchanged.**

cron genuinely cannot be made total, so `render_crontab` raises `UnrepresentableCadenceError` naming the cadence and the nearest expressible ones. It refuses rather than guessing because vixie rejects the *whole* crontab on an out-of-range step — one bad Tier-A line also costs the operator their Tier-B line. `--install-schedule` prints the explanation, still writes all four units, and still exits 0.

**The config bound is deliberately unchanged at `ge=1`.** Capping at `le=59` (the issue's first option) would not fix 45/59 and would break an operator whose launchd 90-minute schedule works correctly today. launchd is untouched.

Also closed, found in self-review: the renderers are public API but had no lower bound of their own, so `tier_a_minutes=0` emitted `OnBootSec=0min` (loads, re-fires with no delay) and a negative value emitted `OnBootSec=-5min`, which systemd refuses — the original defect reproduced through the new path.

### Migration

Operators with a broken timer are covered in `docs/sync-scheduling.md`, split by which failure they hit, because recovery differs:

- **N >= 60** — the unit was refused and was never active. No running trigger, no double-fire risk.
- **N < 60, non-divisor (25/40/45/50/59)** — the unit *loaded and has been running at the wrong cadence*. `list-timers` shows an entry, which is why nothing looked wrong.

Re-running `--install-schedule systemd` fixes both: the four unit filenames are unchanged, so the content is overwritten in place — **no stale unit lingers and there is nothing to delete**. Then `daemon-reload` (required — systemd caches a refused unit, and omitting it is the likeliest reason a real fix looks like a no-op), `enable --now`, and a `restart` to re-arm an already-active timer. The docs instruct no deletion of anything on the operator's system.

## Test plan

Gate 2 `./scripts/check-all.sh` → **exit 0**, 12/12 checks, 9671 passed, coverage 94.66%, per-file gate passed, mypy strict clean, ruff clean with `--no-cache`.

`tests/test_sync_schedule.py`: 227 → 1360 lines, **268 pass, 0 skipped**. All 17 pre-existing tests survive unmodified.

Verified by putting the output **back through the schedulers**, not just through a model:

- All 19 emitted `OnCalendar` specs on real systemd: every one parses **and** fires on uniform gaps equal to the configured interval (60 elapses each).
- Full rendered unit sets for N ∈ {7,25,45,50,59,90,100,300,2880,30,60,120,1440} pass `systemd-analyze verify` with no output.
- All 19 emitted cron lines accepted by real vixie cron.
- Sweep 1..2880: exactly 19 calendar cadences, 2861 monotonic, cron agreeing exactly.

The suite's fire-time expander is an independent oracle, not a restatement of the renderer: it models systemd's "step restarts at the field boundary" semantics with no divisibility test anywhere, and is pinned to 13 transcribed `systemd-analyze calendar` runs. It expands over two days so the midnight discontinuity is visible — without that wrap every hourly assertion would be vacuous. The systemd invariant has **two** clauses (PARSES *and* FAITHFUL) because a uniform-gap check alone blesses `*:0/60`: `range(0,60,60) == [0]` models a perfectly uniform hourly cadence that real systemd refuses to load.

**Mutation battery: 18 mutants, all killed.** Dropping each divisibility guard; the `<60` and `<=23` bounds; `==1440` → `>=` (2880 must not be silently halved); dropping `OnBootSec`; re-adding inert `Persistent=`; deleting the disclosure comment; cron returning `*/N` instead of raising; dropping the value or the alternatives from the message; dropping `Persistent=` from the calendar branch; rendering DAILY as `0/24`; and five on the new positive-cadence guard. Source restored and verified byte-identical by SHA-256 after every run.

No `# noqa`, `# type: ignore`, skip/xfail, or threshold change anywhere in the diff, tests included. `creek/config.py`, `pyproject.toml`, and `scripts/` are untouched.

Closes #1336

### DIFF
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -8,6 +8,7 @@ import logging
 import os
 import re
 import sys
+import textwrap
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
@@ -166,6 +167,15 @@ _SOURCE_COVERAGE_SAMPLE: Final[int] = 10
 The full lists live on :class:`~creek.pipeline.PipelineResult`; the
 summary only samples them so a large source tree cannot bury the rest of
 the report.
+"""
+
+_CRON_NOTICE_WIDTH: Final[int] = 74
+"""Wrap width for the ``--install-schedule`` cron-refusal prose.
+
+Wrapped here rather than by the console so every line keeps its ``#``
+comment marker: the block is meant to be read as one commented note, and
+a console-folded continuation line would lose the marker. 74 plus the
+four-character indent stays inside Rich's 80-column default.
 """
 
 
@@ -1081,6 +1091,46 @@ def _install_launchd(
     console.print(f"# To activate: launchctl load {tier_a} && launchctl load {tier_b}")
 
 
+def _print_cron_alternative(vault_path: Path, minutes: int, hour: int) -> None:
+    """Print the equivalent crontab, or say honestly why there isn't one.
+
+    An irregular Tier-A cadence is a supported configuration, not a failure:
+    the systemd units written alongside this are correct. Only the cron *hint*
+    is unavailable, so this prints an explanation and the command still exits
+    0. No cron line at all is printed in that case — not even the correct
+    Tier-B one — because vixie rejects a crontab as a whole, so a half-file
+    pasted into ``crontab -e`` would install nothing.
+
+    Args:
+        vault_path: The vault the emitted commands target.
+        minutes: The configured Tier-A cadence, in minutes.
+        hour: The configured Tier-B nightly hour.
+    """
+    from creek.sync import UnrepresentableCadenceError, render_crontab
+
+    try:
+        cron = render_crontab(
+            vault=vault_path,
+            tier_a_minutes=minutes,
+            tier_b_hour=hour,
+        )
+    except UnrepresentableCadenceError as exc:
+        console.print("# No crontab is printed for this cadence:")
+        for line in textwrap.wrap(
+            str(exc),
+            width=_CRON_NOTICE_WIDTH,
+            break_on_hyphens=False,
+        ):
+            console.print(f"#   {line}")
+        console.print("#   The systemd timer above is monotonic instead:")
+        console.print("#   it keeps the interval, but ticks missed while the")
+        console.print("#   host was off get no catch-up.")
+    else:
+        console.print("# Or as cron (crontab -e):")
+        for line in cron.splitlines():
+            console.print(f"#   {line}")
+
+
 def _install_systemd(
     vault_path: Path,
     minutes: int,
@@ -1088,7 +1138,7 @@ def _install_systemd(
     out_dir: Path | None,
 ) -> None:
     """Write systemd service+timer units (and print the cron alternative)."""
-    from creek.sync import render_crontab, render_systemd_units
+    from creek.sync import render_systemd_units
 
     units = render_systemd_units(
         vault=vault_path, tier_a_minutes=minutes, tier_b_hour=hour
@@ -1103,14 +1153,14 @@ def _install_systemd(
     ):
         (base / fname).write_text(content, encoding="utf-8")
         console.print(f"# wrote {base / fname}")
+    # daemon-reload first: systemd caches a unit it previously refused to load,
+    # so re-installing over a broken timer looks like a no-op without it.
+    console.print("# To activate: systemctl --user daemon-reload")
     console.print(
-        "# To activate: systemctl --user enable --now "
+        "#   then: systemctl --user enable --now "
         "creek-sync-tier-a.timer creek-sync-tier-b.timer",
     )
-    console.print("# Or as cron (crontab -e):")
-    cron = render_crontab(vault=vault_path, tier_a_minutes=minutes, tier_b_hour=hour)
-    for line in cron.splitlines():
-        console.print(f"#   {line}")
+    _print_cron_alternative(vault_path, minutes, hour)
 
 
 def _install_schedule(host: str, vault: Path | None, out_dir: Path | None) -> None:
--- a/creek-tools/creek/sync/__init__.py
+++ b/creek-tools/creek/sync/__init__.py
@@ -3,11 +3,16 @@
 Emits launchd / systemd / cron artifacts that run the two-tier sync on a
 schedule. These are *config artifacts* the operator installs manually — Creek
 never shells out to ``launchctl``/``systemctl``/``crontab``.
+
+Not every cadence survives every host's grammar (#1336): cron refuses one it
+cannot state exactly, raising :class:`UnrepresentableCadenceError`, while
+systemd falls back to a monotonic timer and launchd is exact throughout.
 """
 
 from creek.sync.schedule import (
     LaunchdPlists,
     SystemdUnits,
+    UnrepresentableCadenceError,
     render_crontab,
     render_launchd_plists,
     render_systemd_units,
@@ -16,6 +21,7 @@ from creek.sync.schedule import (
 __all__ = [
     "LaunchdPlists",
     "SystemdUnits",
+    "UnrepresentableCadenceError",
     "render_crontab",
     "render_launchd_plists",
     "render_systemd_units",
--- a/creek-tools/creek/sync/schedule.py
+++ b/creek-tools/creek/sync/schedule.py
@@ -14,12 +14,25 @@ Vault paths may contain spaces (common on macOS, e.g. ``~/Documents/My
 Notes``): launchd receives the command as an argument *list* (one ``<string>``
 per token, so the path stays whole), while systemd ``ExecStart`` and the cron
 line *quote* the path.
+
+Cadence fidelity (#1336). A repetition step in a systemd calendar spec or a
+cron field restarts at the *enclosing* field's boundary, so a step that does
+not divide its field silently keeps the wrong cadence: ``*:0/45`` fires at :00
+and :45 of every hour, never every 45 minutes. A step at or above the field's
+own maximum is worse still — ``*:0/60`` and ``0/24:00:00`` are refused at parse
+time ("Failed to parse calendar specification"), after which the timer unit
+will not load at all. Every Tier-A cadence therefore goes through
+:func:`_classify_cadence` before it is rendered: the cadences a calendar field
+can keep faithfully get an ``OnCalendar=`` timer (keeping ``Persistent=true``
+catch-up), and the rest get a monotonic systemd timer plus, for cron, an
+explicit refusal. launchd is unaffected: ``StartInterval`` is exact seconds.
 """
 
 from __future__ import annotations
 
 import html
 from dataclasses import dataclass
+from enum import Enum, auto
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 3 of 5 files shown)


Geoffe-Ga:
## Summary
This PR fixes #1336: `sync.tier_a_interval_minutes` was interpolated directly into systemd's `OnCalendar=*:0/N` and cron's `*/N * * * *` minute-step fields, both of which silently misfire for any N that doesn't divide the field's modulus (e.g. `*:0/45` fires at :00 and :45, not every 45 minutes) and are rejected outright for N ≥ 60. The fix introduces a `_classify_cadence` classifier in `creek/sync/schedule.py` that routes each cadence to the one calendar form that can state it exactly (sub-hourly divisor of 60, whole-hour divisor of 24, or literal daily midnight), falls back to a monotonic `OnBootSec`/`OnUnitActiveSec` systemd timer for everything else, and makes cron refuse inexpressible cadences via a new `UnrepresentableCadenceError` rather than emit a wrong or crontab-breaking line. `docs/sync-scheduling.md` gets a supported-cadence matrix and a recovery runbook for operators who already installed a broken timer.

## Strengths
- The classifier is provably total and correct by construction: `_CALENDAR_SPEC_TEMPLATES`/`_CRON_SCHEDULE_TEMPLATES` omit `IRREGULAR` by construction, so the renderers can't drift from the divisibility rule; I traced every branch (sub-hourly, hourly, the 24-divides-24-but-must-not-match ordering trick for `DAILY`, and the fallthrough to `IRREGULAR`) by hand against the stated boundary cases (60, 1440, 2880) and it's correct.
- The test suite calibrates its own oracle against *recorded* real-scheduler output (`systemd-analyze calendar`, vixie `crontab -`) rather than re-deriving the renderer's own formula, and explicitly tests that a uniform-gap check alone is insufficient (`*:0/60` looks uniform but real systemd refuses it) — this is exactly the kind of independent verification that prevents a test suite from becoming a mirror of a buggy implementation.
- Byte-for-byte compatibility at the existing default (N=30) is preserved and tested, and Tier B's timer/cron line is asserted unchanged across the whole 1..1440 sweep — a real regression-safety net for a change this invasive.
- The docs addition is unusually good: it distinguishes the two different failure modes an already-affected operator may have hit (loud refusal vs. silent wrong-cadence) and gives cadence-appropriate recovery steps, including the non-obvious `daemon-reload` requirement.
- Zero-and-negative cadence handling was caught in self-review and closed with its own guard + tests, which is exactly the kind of scope the PR description says was "found in self-review" rather than smuggled in silently.

## Security Concerns
None. 🟢 No new subprocess/shell invocation, no user-controlled data reaches a format string in an unsafe way (the vault path was already quoted; the new cadence fields are integers formatted via `{step}`/`{minutes}` templates, not concatenation of arbitrary strings), and the module's existing invariant (never shells out to `launchctl`/`systemctl`/`crontab`) is preserved.

## Problems
None found. I traced `_classify_cadence`, both template dicts, `_nearest_cron_cadences`, `_require_positive_cadence`, and the CLI wiring (`_install_systemd` → `_print_cron_alternative`, which catches `UnrepresentableCadenceError` internally so the command still exits 0) line by line against the PR's own claims and found no discrepancy. I was not able to execute `./scripts/check-all.sh` or `pytest` in this review environment (tool invocations for anything beyond `gh`/read-only shell utilities were not approved), so the 268-tests/94.66%-coverage/mutation-testing claims in the PR description are unverified by me directly, but the diff is internally consistent and the test code itself (independently reading the rendered `OnCalendar=`/`OnBootSec=` directives and modeling scheduler semantics from scratch) is convincing.

## Code Quality
Non-blocking observations only:
- `_print_cron_alternative` does a local `from creek.sync import ... ` on every call; this matches the file's existing lazy-import convention for `creek.sync` symbols in `cli.py`, so not worth changing.
- `_nearest_cron_cadences` only offers a lower-bound alternative once `minutes` exceeds 1440 (no cron cadence is longer than a day), so extremely large cadences get a one-sided suggestion — correct behavior, just worth knowing it's asymmetric by design rather than an oversight.

## CLAUDE.md Adherence
✅ No `# noqa`, `# type: ignore`, skip/xfail — confirmed by inspection of the diff.
✅ Docstrings are Google-style with Args/Returns/Raises throughout the new public and private functions.
✅ `creek/config.py`, `pyproject.toml`, and `scripts/` are untouched, matching the PR's own claim.
✅ Commit message follows Conventional Commits (`fix(sync): ...`).
✅ Guard-at-the-primitive design (validate where intent becomes representation, not per-caller) matches the repo's stated precedent (#1306, #1311, #1332 per the PR description).
Coverage/lint/mypy/complexity claims in the PR body could not be independently executed in this review sandbox; nothing in the diff itself suggests a threshold would be missed.

## Testing
Very strong. The new `tests/test_sync_schedule.py` additions build an independent test-local model of systemd/cron field-restart semantics (not a restatement of `schedule.py`'s divisibility rule), pin it to transcribed real-scheduler output, sweep the entire 1..1440-minute space for both "never emits a spec systemd rejects" and "every monotonic timer discloses its lost catch-up," and include explicit anti-vacuity tests guarding against an accidentally-emptied parametrize list. CLI-level tests confirm the `--install-schedule systemd` exit code stays 0 for an irregular cadence and that no cron line is ever printed for an inexpressible one. This is well above what's needed to block merge — nothing here is insufficient.

## Documentation
`docs/sync-scheduling.md` is updated with a supported-cadence table, an explanation of the monotonic fallback's catch-up tradeoff, and a dedicated "Fixing an already-installed broken timer" section that correctly distinguishes the loud (N≥60) and silent (non-divisor <60) failure modes operators may already be experiencing, with per-mode recovery commands. This is complete and accurate against the code changes.



## Verdict: LGTM

The classifier, both render paths, and the CLI wiring are logically correct by manual trace (I verified the sub-hourly/hourly/daily/irregular boundary cases, the zero/negative guard, and the exit-code-0 cron-refusal path), the change preserves byte-identical output at the existing default, and the test suite is unusually rigorous with an independently-calibrated oracle rather than a restatement of the implementation. I found no bugs, no security concerns, and no CLAUDE.md violations in the diff — this is ready to merge as-is.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5262982140 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #338, opened 2026-05-25
bug(link): linker output counts don't match what materializes in the vault (eddies reports `1 link` with empty 03-Eddies/; similarity edges aren't visible)


Geoffe-Ga:
## Symptom

After running all three linker passes on a freshly-classified vault:

```
$ creek link --vault <vault> --method embeddings
Embeddings linker: 188 fragment(s), 1212 link(s).

$ creek link --vault <vault> --method temporal
Temporal linker: 188 fragment(s), 0 link(s).

$ creek link --vault <vault> --method eddies
Eddies linker: 188 fragment(s), 1 link(s).
```

The reported counts do not match what's actually visible in the vault:

| Linker     | Reported | Materialized in vault                                                            |
|------------|----------|----------------------------------------------------------------------------------|
| embeddings | 1212     | `00-Creek-Meta/embeddings.parquet` (460KB) exists; **no per-fragment links** in any frontmatter |
| temporal   | 0        | `02-Threads/{Active,Dormant,Resolved}/` empty (just `.gitkeep`); every fragment has `threads: []` |
| eddies     | 1        | `03-Eddies/` empty (just `.gitkeep`); every fragment has `eddies: []`            |

The temporal "0 links" is consistent and probably correct (Substack essays are too sparse in time for the default 168h window with `thread_min_fragments: 3`, and many fragments lack `authored_at`).

The other two are mismatches worth fixing.

### Problem A: eddies linker reports `1 link(s)` but materializes nothing

The eddies output says one was found, yet:
- `03-Eddies/` contains no eddy markdown file
- No fragment's `eddies:` frontmatter field references any eddy

Either the eddy was found but the writer is silently failing to persist it, or the "1 link" count refers to something internal that isn't an eddy in the user-facing sense.

### Problem B: linker count semantics are opaque

The `Embeddings linker: 188 fragment(s), 1212 link(s).` output is meaningful internally (presumably the count of pairwise similarity edges stored in `embeddings.parquet`) but it implies to the user that 1,212 *links* now exist somewhere visible. They don't — at least not in any frontmatter field, and not as files. Same vocabulary ("link") is being used for two different things: parquet-internal similarity edges, and user-facing per-fragment `resonances`/`eddies`/`threads` cross-references.

This matters because the user has no way to verify what the linker actually did. A 1212-edge parquet is unreviewable; per-fragment links would be reviewable. The current output suggests the latter while delivering the former.

## Expected

[...]

Issue #1441, opened 2026-08-12
fix(link): embeddings.parquet is rewritten in place, so a mid-write failure destroys a good cache


Geoffe-Ga:
**Found during:** Gate 2.5 self-review of #1337. Pre-existing; not a regression that PR introduces.

## Problem

`EmbeddingLinker.save_cache` (`creek-tools/creek/link/embeddings.py`) writes the vector
cache with `pq.write_table(table, path, compression="snappy")` — straight onto the live
`<vault>/00-Creek-Meta/embeddings.parquet`, with no temp-file-plus-rename.

`_persist_cache` (`creek-tools/creek/link/link_engine.py`) then catches `OSError` and
degrades gracefully. That handles a write that fails *before* it starts (read-only volume,
permission denied). It does not handle a write that fails *part-way*: the disk filling up
mid-`write_table`, or the process being killed. In that window the previously-good cache
has already been truncated, so the failure destroys a working artifact rather than leaving
it untouched.

The blast radius is bounded — `_safe_load_cache` catches `OSError`/`ValueError` on read and
falls through to a full recompute, so a corrupt parquet costs re-embedding the corpus
rather than producing wrong answers. On a large vault that is still expensive
(`docs/linking.md` puts a 10k-fragment rebuild at ~10-20 minutes), and it is silent apart
from a `WARNING`.

## Why not fixed in #1337

#1337 was about the *operator-facing sentence* matching the artifact, and it fixed the
reporting side of exactly this failure: the CLI now says `no vectors written to
00-Creek-Meta/embeddings.parquet` instead of claiming the vectors were cached. That
statement stays literally true in the partial-write case too, because a truncated parquet
yields zero readable rows on the next load. So the honesty defect is closed; the
durability defect is a separate concern with a separate fix, and making it atomic would
have widened that diff into I/O semantics it did not otherwise touch.

## Tasks

1. Write the parquet to a sibling temp file in the same directory, then `os.replace` it
   over the target — atomic within a filesystem, so a failed or interrupted write leaves
   the previous cache intact.
2. Clean up the temp file on failure so a series of failed runs cannot litter
   `00-Creek-Meta/`.
3. Test it: simulate a mid-write failure and assert the *previous* cache is still readable
   afterwards and still holds its original rows — asserting on the bytes on disk, not on a
   returned value.
4. Check whether the same pattern applies to the other writers under `00-Creek-Meta/`
[...]

Issue #1438, opened 2026-08-12
fix(mine): the no-synchronicities hint cites 'embeddings.parquet edges' and points at a command that does not populate them


Geoffe-Ga:
**Found during:** #1337 (`creek link --method embeddings` output honesty). Verified at HEAD `26a69c9` by reading the source; the empty-state hint itself was not exercised.

## Problem

`creek/generate/mining.py:108-112` defines the empty-state hint the mining surface shows when there are no synchronicity records:

```python
_NO_SYNCHRONICITIES: str = (
    "no synchronicity records on disk under 10-Liminal/Synchronicities; "
    "embeddings.parquet edges do not contribute until they are written as "
    "synchronicity notes (run `creek link` to populate)"
)
```

Two problems, both operator-facing:

1. **"embeddings.parquet edges"** — the parquet has no edge column. Its schema, read
   back from a real vault at HEAD, is exactly `fragment_id`, `content_hash`,
   `model_name`, `embedding` (`list<float>`), `computed_at`; `num_rows` equals the
   fragment count. Resonance edges are computed in memory and dropped — there is no
   resonance writer anywhere in the codebase and `Fragment` has no `resonances` field.
   This is the same family-3 claim #1303 fixed in `_format_link_summary` and #1337
   finished on the `creek link` path; this string was missed by both.

2. **"run `creek link` to populate"** — unverified, and probably false. No
   `creek link` method writes anything under `10-Liminal/Synchronicities/`.
   Synchronicity notes are produced by
   `creek.generate.synchronicity.SynchronicityDetector`, run via
   `creek report --type synchronicity` (`docs/linking.md:122`). If that is right, the
   hint sends the operator to a command that cannot resolve the empty state it is
   explaining.

## Tasks

1. Verify by execution which command actually writes `10-Liminal/Synchronicities/`.
2. Reword `_NO_SYNCHRONICITIES` to drop the "parquet edges" framing and to name the
   command that genuinely populates the directory.
3. Add a test that ties the hint to the command it names — asserting the named command
   is a real registered CLI entry point and that running it is what makes the empty
   state go away, rather than asserting the hint's substring.

## Why not fixed in #1337

Different command surface (`creek mine`, not `creek link`), and the second half needs
its own wiring verification rather than a wording change. #1337 deliberately kept its
diff to the linker's own output.

Refs #1337, #1303

PR #1442, opened 2026-08-12
fix(link): make the embeddings summary true of the parquet, and stop building 43GB of edges to count them

### LINKED ISSUE #1337: fix(link): creek link reports N 'similarity edge(s) cached in embeddings.parquet' — the parquet has no edge column and the resonances are discarded
**Severity:** High
**Remediation:** fix in place

## Problem

`creek link --method embeddings` tells the operator:

```
Embeddings linker: {fragments} fragment(s) embedded, {summary.similarity_edges}
similarity edge(s) cached in 00-Creek-Meta/embeddings.parquet.
```

(`creek/cli.py:1931-1935`), and `LinkSummary.similarity_edges`'s docstring
(`creek/link/link_engine.py:80`) repeats it: "Pairwise cosine-similarity edges
cached in the embeddings parquet."

Nothing persists resonance edges. `_run_embeddings`
(`creek/link/link_engine.py:534-539`) calls `find_resonances(...)`, returns
`len(resonances)`, and every `Resonance` — with its `similarity`, `from_level`,
`to_level` — is then garbage. Family 3 (fake/aspirational claim) + family 0
(silent loss).

**Current state:** the parquet holds embeddings only, so an operator who runs
`creek link --method embeddings` believing the edges are cached will find nothing
to reuse, and every downstream consumer recomputes from scratch
(`creek/generate/synchronicity.py:441` does exactly that).

## Evidence (corroboration)

**Signal 1 — structural.** `EmbeddingLinker.save_cache`
(`creek/link/embeddings.py:596-620`) writes exactly five columns:
`fragment_id`, `content_hash`, `model_name`, `embedding`, `computed_at`. There is
no edge column. `rg resonance` across `creek/vault/` and `creek/models.py`
returns zero persistence sites.

Reproduce: `creek link --vault V --method embeddings` on any vault with
resonances → stdout claims N edges cached; then
`pq.read_table(V/"00-Creek-Meta/embeddings.parquet").schema` contains no edge
column and `num_rows` equals the fragment count, not N.

**Signal 2 — reading/intent.** `LinkSummary`'s class docstring
(`link_engine.py:66-71`) says the structured fields exist "so the CLI can phrase
what actually happened". This is the one field that doesn't. Closed issue #338
proposed this exact wording as the fix for opaque counters; the wording shipped
but its premise was never true, so #338's Problem B is live again under a more
specific and more misleading sentence. At the documented 35k-fragment scale
(`prompts/scans/perf.md:27`), `_resonances_exact` also materialises one Pydantic
`Resonance` per above-threshold pair into a single list purely to call `len()` on
it — the 188-fragment vault in #338 produced 1,212 edges (~6.9% of pairs).

Distinct from open #1303, which covers `LinkingPipeline` / `creek process`
Stage 5. This is the `link_engine` / `creek link` path and the parquet claim.

## Scope

The CLI sentence, the `LinkSummary` docstring, and (if edges are wanted) the
persistence. Does not change the resonance computation itself.

## Tasks

1. Decide: either persist resonance edges alongside the embeddings, or stop
   claiming they are cached.
2. If not persisting: reword `creek/cli.py:1931-1935` and
   `link_engine.py:80` to describe what actually happened (edges *found*, not
   cached), and count them without materialising the full list.
3. If persisting: add an edges table/column to `save_cache`/`load_cache` with a
   schema version bump, and have `generate/synchronicity.py:441` reuse it.
4. Add a test asserting the CLI sentence matches the parquet's actual contents.

## Acceptance Criteria

- [ ] The operator-facing sentence is true of the artifact on disk.
- [ ] A test fails if the claim and the parquet schema diverge again.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/link/link_engine.py` | Modify |
| `creek-tools/creek/link/embeddings.py` | Modify (only if persisting) |
| `creek-tools/tests/test_link_engine.py` | Modify |

Geoffe-Ga:
### The issue's headline premise is refuted at HEAD — and it says so itself, wrongly

#1337 is titled "`creek link` reports N 'similarity edge(s) cached in embeddings.parquet'". It also says "Distinct from open #1303." **#1303 is closed**, merged as PR #1383 (commit `3796c39`, in this branch's base), and that PR already reworded both sites this issue names. Observed at HEAD `26a69c9`:

```
Embeddings linker: 12 fragment(s) embedded, 66 similarity edge(s) computed in memory
(not persisted — no resonance writer exists); vectors cached in 00-Creek-Meta/embeddings.parquet.
```

The edge-caching claim is gone. `LinkSummary.similarity_edges`'s docstring (now `link_engine.py:81-85`, not the `:80` the issue cites) already reads "**Not persisted**". Every line number in the issue is stale.

**But three lies of exactly the same family survive, all inside that one sentence,** and #1383 built the field that fixes one of them without wiring it up. I ran a census of all four linker arms; the temporal, eddies and threads arms are honest and untouched here.

| # | State | What it said | What was on disk |
|---|---|---|---|
| 1 | Warm cache (2nd identical run) | `12 fragment(s) embedded` | `fragments_embedded == 0` — the model computed nothing |
| 2 | Empty vault | `vectors cached in 00-Creek-Meta/embeddings.parquet` | **no parquet exists** — `run_link` returns early at `link_engine.py:190-191` |
| 3 | Cache write fails | same caching claim, exit 0 | **no parquet exists** — `_persist_cache` swallows `OSError` into a log line |

Observed CLI sentence and observed parquet schema, side by side, at HEAD:

```
SAYS : ... 12 fragment(s) embedded ... vectors cached in 00-Creek-Meta/embeddings.parquet.
DISK : fragment_id: string | content_hash: string | model_name: string
       embedding: list<element: float> | computed_at: timestamp[us, tz=UTC]
       num_rows: 12   schema.metadata: None
```

The only existing test of this sentence, `test_link_embeddings_output_phrases_similarity_edges`, ran on an **empty vault where no parquet was ever written** and asserted `"embeddings.parquet" in result.output`. It would have passed with the cache deleted from the codebase.

### The decision on task 1: reword. Persist rejected.

The issue asks for a persist-vs-reword call. **Reword**, for five reasons — and this is not the cheap option talking, because the expensive half of "reword" (counting without materialising) is most of this diff.

1. **#1383 already made this call on the neighbouring path**, explicitly: "no writer was invented." Re-deciding it here re-forks the contract, which is the architectural-drift family #1303 closed.
2. **Invalidation is the killer.** The parquet's freshness model is per-fragment `content_hash`. An edge is a function of *two* fragments, so one edit invalidates every edge touching it. Bolting edges into this file means either a full recompute on any edit (the cache never helps) or a per-edge freshness protocol that does not exist.
3. **The parquet is documented-deletable derived data** — `--rebuild` unlinks it, `delete_embeddings_cache` exists, `creek purge vault` removes it, and `docs/cleaning-and-purge.md` / `docs/security/threat-model.md` both describe it as regenerable. A Resonance is one of the five ontological primitives. Putting a primitive inside a file whose documented lifecycle is "safe to delete" is a category error no schema version can fix.
4. **Measured scale.** Using the issue's own density datum (#338: 188 fragments → 1,212 edges = 6.9% of pairs), 35k fragments is 42,261,292 edges ≈ **0.9 GB of parquet minimum**, built from ~43 GB of intermediate objects. That artifact needs its own design — streaming writer, pruning policy, size budget — not a column bolted onto a vector cache.
5. **No consumer is wired.** `generate/synchronicity.py:441` computes over its own fragment set at its own config; reuse needs threshold agreement, which is a separate design.

**Migration answer: none is needed, and that is a property of the decision, not a gap.** The on-disk format is byte-for-byte unchanged, so there is no schema version bump, no `load_cache` compatibility branch, and no possibility of an old parquet being misread as "zero edges found" — because nothing ever reads edges from it.

### What changed

The sentence is now derived from what happened, in all four states (each observed by running the CLI and reading the parquet back):

```
cold  : 12 fragment(s) scanned, 12 vector(s) computed, 66 similarity edge(s) computed in memory
        (not persisted — no resonance writer exists); 12 vector(s) cached in 00-Creek-Meta/embeddings.parquet.
warm  : 12 fragment(s) scanned,  0 vector(s) computed, ... 12 vector(s) cached in ...
empty :  0 fragment(s) scanned,  0 vector(s) computed, ... no vectors written to ...   (parquet absent)
OSError:  4 fragment(s) scanned,  4 vector(s) computed, ... no vectors written to ...   (parquet absent, exit 0)
```

`LinkSummary` gains `vectors_persisted` (rows this run landed) and `EmbeddingPass` gains a **required** `persisted` — required, not defaulted, because the number must be measured and a default is precisely how a future call site under-reports silently. `_persist_cache` returns the count and `0` from its `except OSError`. The swallow is deliberate and stays: losing the cache only costs a recompute, and **the honest count is the fix for the swallowed error** — it is now the only operator-visible trace of a failure that previously reached the log alone.

The wording split is load-bearing: only the *edges* clause says "not persisted", so an assertion on that phrase unambiguously pins the edge claim and cannot be satisfied by the cache clause.

### The 43 GB the issue's sub-point asked about

`_run_embeddings` called `find_resonances` and took `len()` of the result, so every above-threshold pair became a Pydantic `Resonance` that was counted and dropped. Measured with `tracemalloc`, stable across three scale points:

| N | edges | bytes/edge |
|---|---|---|
| 800 | 13,953 | 1,017 |
| 1,500 | 302 | 1,019 |
| 2,000 | 87,408 | 1,016 |

At 42.3M edges that is **~43 GB of heap to call `len()`** — an out-of-memory failure, not a slowdown.

The exact pair scan is now a true generator and `count_resonances` streams it. Measured after: peak memory is **independent of edge count** (N=800 → 2.62 MB, N=2000 → 8.33 MB; the residual scales with the row-block matmul, i.e. with N, not edges), against `find_resonances`'s steady ~1,060 B/edge. Counting is also 5.7x faster (23.9 ms vs 135.7 ms at N=2000) because it skips Pydantic construction.

**Returning `list[tuple[int, int, float]]` instead would not have been a fix** — at ~120 B/edge that is still ~5 GB at 35k. The laziness is the point, and it is tested.

**Stated plainly: time is still O(n²).** This removes an OOM; it does not make 35k fragments fast. `_resonances_topk` is a generator for signature parity only — dedup-then-sort is inherently eager and its pick list is bounded at O(n·k) — and its docstring says so rather than implying it streams.

**Equivalence proved by execution, not argument.** `find_resonances` output was compared against `git show HEAD:`'s version across exact/topk × hierarchy/no-hierarchy: **identical edge tuples in identical order**, including the `float(row[j])` values. Ragged input raises `ValueError` identically from `find_resonances` and the new `count_resonances`. Wall-clock cost of the generator indirection on the list path: **+1.1% (N=2000), +2.3% (N=3000)**.

### Census of the rest of the linker's CLI output

Asked for one sentence; swept all four arms. `temporal` ("computed in memory, not persisted" — observed zero new files) and `eddies`/`threads` (which report `*_written`, never `*_detected`) are honest and are **not touched**. `vectors_persisted` is populated on the eddies and threads paths too, so the struct is never a lie for a programmatic consumer, but it is not surfaced in their sentences — they make no parquet claim, so there is no false claim to correct, and widening the diff there would buy nothing.

One adjacent lie found and **filed rather than fixed**: `creek/generate/mining.py:108-112` tells the operator that "embeddings.parquet edges do not contribute until they are written as synchronicity notes (run `creek link` to populate)" — same false framing, plus a second claim that `creek link` populates `10-Liminal/Synchronicities/`, which needs its own wiring verification. **#1438** (P2).

Also filed from self-review: **#1441** (P2) — `save_cache` rewrites the parquet in place with no temp-file-plus-rename, so a mid-write failure destroys a good cache. Pre-existing, bounded (a corrupt file forces a clean recompute), and orthogonal to this issue's honesty defect.

### One documentation correction, in the same family

`docs/linking.md:25` said a resonance is emitted when similarity "exceeds" the threshold. The code compares with `>=` and `EmbeddingsConfig.similarity_threshold`'s own docstring says "**Minimum**". A pair sitting exactly on the threshold *does* resonate. Corrected, and now pinned by a test so the doc and the code cannot drift again.

## Test plan

**RED first, proven by running it — not asserted.** Twelve failures, four of them behavioural on observed CLI output rather than a missing symbol:

```
test_link_embeddings_output_phrases_similarity_edges  AssertionError: no cached-vector count in output:
                                                      Embeddings linker: 4 fragment(s) embedded, 6 similarity edge(s) co...
test_link_embeddings_warm_run_reports_zero_vectors_computed
                                                      assert '4 fragment(s) scanned, 4 vector(s) computed,' in '...4 fragment(s) em...'
test_link_embeddings_empty_vault_admits_nothing_was_cached
                                                      assert 'no vectors written to 00-Creek-Meta/embeddings.parquet' in '...'
test_link_embeddings_cache_write_failure_is_admitted  same
test_embedding_pass_requires_an_explicit_persisted_count   Failed: DID NOT RAISE TypeError
```
The remaining seven are `AttributeError` on the new `vectors_persisted` / `count_resonances`, which a new field cannot avoid; the behavioural five are what prove the defect.

**The invariant: the claim is tied to the artifact, not to a string.** `test_link_embeddings_output_phrases_similarity_edges` was strengthened **in place** (its #338/#1303 vocabulary assertions kept verbatim — the accurate wording has to survive the rewording, not be replaced by it), moved onto a seeded vault so a parquet genuinely exists, and now:

- parses the cached count out of the CLI output with a regex and asserts it `== pq.read_table(cache).num_rows` — **derived, never hardcoded**, so growing the fixture cannot make it vacuous;
- asserts `set(table.column_names)` is **exactly** the five vector/freshness columns;
- asserts that *because* there is no edge column, the sentence must not claim edges are cached.

That is a two-way tie: **add an edge column later and the schema assertion fires; claim edge caching without adding the column and the wording assertion fires.** Neither half can move alone. The empty-vault and OSError tests each put `not cache_path.exists()` and the wording assertion **in the same test**, so they cannot be weakened independently — which is exactly how the claim and the artifact drifted apart originally.

**Mutation battery — 12 mutants, 12 killed.** Snapshot-restore by file bytes with hash verification after every mutant, never `git checkout` (that would delete this lane's own uncommitted work). Two mutants initially **survived and are the reason two tests exist**:

| Mutant | Result |
|---|---|
| CLI computed count reverts to `fragment_count` | killed |
| CLI cache clause always positive | killed |
| `scanned` reverts to `embedded` | killed |
| `_persist_cache` reports rows after a failed write | killed |
| `vectors_persisted` aliased to `fragments_embedded` | killed |
| `EmbeddingPass.persisted` gains a default | killed |
| **call site reverts to `find_resonances`** | **survived → new test → killed** |
| **`count_resonances` uses `len(list(...))`** | **survived → new test → killed** |
| exact path emits hierarchy-suppressed pairs | killed |
| `_prepare_scan` loses its `<2 ids` short-circuit | killed |
| topk loses dedup-then-sort | killed |
| threshold `>=` → `>` | **survived → new test → killed** |

The two survivors mattered. Nothing pinned that `_run_embeddings` *calls* `count_resonances` — the entire memory fix could be reverted at the call site with every count still correct — so `test_run_link_embeddings_counts_edges_without_building_them` spies on the `Resonance` constructor through a real `run_link` and asserts zero objects for 15 edges. And `len(list(...))` builds zero `Resonance` objects, so the constructor spy could not see it; `test_count_resonances_peak_memory_does_not_hold_the_edge_set` measures `tracemalloc` peak on a corpus where edges vastly outnumber fragments (**3.3 MB streaming vs 18.3 MB via `list()`**, ceiling set between them with ~2.4x headroom). The threshold survivor was a genuinely untested boundary in code this PR rewrote, now pinned inclusively to match the config docstring.

**Drift alarm.** `count_resonances(...) == len(find_resonances(...))` parametrised over `exact`/`topk` × hierarchy/no-hierarchy, each asserting `len(edges) > 0` first so the property cannot go vacuous at `0 == 0`.

**Gates, all run by me:**
- `./scripts/check-all.sh` → **exit 0**, 12/12 gates, **9687 passed / 5 skipped**, coverage **94.67%**, per-file gate green. The 5 skips are the pre-existing live-API and Ollama smokes; **zero SKIPPED** across every touched test file (asserted with `-rs`).
- Integration lane 357 passed; e2e lane 25 passed.
- `ruff check --no-cache` clean (verified without the cache — a stale one produced a false-clean here in #1119); `ruff format` clean; mypy strict clean; `xenon --max-absolute B --max-modules B --max-average A creek/` exit 0.

**Substituted for `pre-commit run --all-files`, which the issue's acceptance criteria ask for.** It silently edits unrelated files in the worktree — it has stripped the BOM from `tests/fixtures/encoding/utf8_bom.csv` before. `check-all.sh` is the gate here and covers the same ground.

**Anti-bypass sweep over the whole diff:** no `# noqa`, no `# type: ignore`, no `# pylint: disable`, no skip/xfail, no threshold touched in `pyproject.toml` or `scripts/`. Five `# type: ignore` comments the first pass introduced in my own new tests were **removed**, not justified. No test deleted: all 15 deleted lines are one rewritten docstring, the body of the one strengthened-in-place test, one stub constructor gaining `persisted=0`, and two import lines. No duplicate `def test_`/`class Test` names (checked by AST, module level and per class).

**Gate 2.5:** `code-review-orchestrator` over the diff returned **CLEAN, no blockers**. Its one actionable nit was acted on: `_ResonanceScan` is `@dataclass(frozen=True)` holding a numpy array, so the generated `__eq__`/`__hash__` advertised behaviour they could not honour — **verified by execution** (`ValueError: truth value of an array... is ambiguous` and `TypeError: unhashable type`) and fixed with `eq=False`. Its other finding became #1441.

**Not covered, stated rather than glossed:**
- Resonance persistence is not built, deliberately — see the decision above. `generate/synchronicity.py:441` still recomputes; nothing changed there.
- Time complexity is unchanged at O(n²). Only the memory failure is fixed.
- `test_exact_pair_scan_yields_before_finishing_the_traversal` is necessarily white-box (it calls the private `_prepare_scan` and monkeypatches the private `_is_hierarchy_suppressed`) because laziness has no black-box-observable return value; it is coupled to those two names.
- The `tracemalloc` ceiling is absolute. If it ever flakes on a specific interpreter in the 3.11/3.12/3.13 matrix, the fix is a wider ceiling or a relative streaming-vs-list comparison — not weakening the property.

Closes #1337
Refs #1303, #338, #1438, #1441

### DIFF
--- a/creek-tools/CHANGELOG.md
+++ b/creek-tools/CHANGELOG.md
@@ -83,3 +83,21 @@ to locate the originating commit for any reference below.
   operator's edits: it now preserves the local version alongside as
   `<name>.bak` (e.g. `mediums/essay.MEDIUM.md.bak`) before
   overwriting.
+- **`creek link --method embeddings` now reports what actually happened on
+  disk, not an assumption about it** (#1337). The summary line used to claim
+  a single "N fragment(s) embedded" count regardless of cache state; it now
+  reports three independent numbers — `scanned` (fragments loaded from the
+  vault), `computed` (vectors the model actually produced this run, i.e.
+  cache misses only — legitimately zero on a warm cache), and `cached`
+  (rows in `00-Creek-Meta/embeddings.parquet` after the run). When nothing
+  reached disk — an empty vault, or a cache write that failed — the line
+  now says `no vectors written` instead of silently repeating a stale count;
+  a failed cache write was previously swallowed into a log line only, with
+  the CLI still exiting 0. Separately, counting resonance edges no longer
+  materializes a `Resonance` object per pair
+  (`EmbeddingLinker.count_resonances` walks the same traversal as
+  `find_resonances` without allocating them), making peak memory
+  independent of edge count. Resonance edges are still not persisted, and
+  the `embeddings.parquet` schema is unchanged (`fragment_id`,
+  `content_hash`, `model_name`, `embedding`, `computed_at`) — no on-disk
+  format changed, and no migration is needed.
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -2665,6 +2665,14 @@ def _format_link_summary(summary: LinkSummary) -> str:
     to disk (eddies). Per-method strings name the actual side effect so
     operators can correlate the number with what they see on disk.
 
+    The embeddings sentence carries three separate numbers because they
+    are three separate facts and every pairing of them can differ:
+    *scanned* is the corpus size, *computed* is the local-model work this
+    run paid for (zero on a warm cache), and *cached* is what the parquet
+    holds after the write (the whole corpus on that same warm run, or zero
+    if the write failed). The similarity edges sit outside all three —
+    nothing persists them at all (#1337).
+
     Args:
         summary: The link summary returned by
             :func:`creek.link.link_engine.run_link`.
@@ -2681,11 +2689,18 @@ def _format_link_summary(summary: LinkSummary) -> str:
         # the parquet. Only the vectors are — the similarity edges are
         # computed in memory and dropped, because there is no Resonance
         # writer and Fragment has no ``resonances`` field.
+        # #1337: the remaining three claims are now measured rather than
+        # assumed. "embedded" conflated the corpus with the vectors this
+        # run computed (a warm cache computes none), and the cache clause
+        # was a hardcoded assertion that survived an empty vault and a
+        # swallowed OSError alike. Only the *edges* clause says "not
+        # persisted", so that phrase unambiguously pins the edge claim.
         body = (
-            f"Embeddings linker: {fragments} fragment(s) embedded, "
+            f"Embeddings linker: {fragments} fragment(s) scanned, "
+            f"{summary.fragments_embedded} vector(s) computed, "
             f"{summary.similarity_edges} similarity edge(s) computed in memory "
-            f"(not persisted — no resonance writer exists); vectors cached in "
-            f"00-Creek-Meta/embeddings.parquet."
+            f"(not persisted — no resonance writer exists); "
+            f"{_format_embeddings_cache_clause(summary)}."
         )
     elif summary.method == "eddies":
         body = (
@@ -2717,6 +2732,41 @@ def _format_link_summary(summary: LinkSummary) -> str:
         msg = f"Unknown link method: {summary.method!r}"
         raise ValueError(msg)
     return f"[bold green]{body}[/bold green]"
+
+
+def _format_embeddings_cache_clause(summary: LinkSummary) -> str:
+    """Render the parquet clause for the embeddings sentence.
+
+    Issue #1337: the clause used to be a constant asserting that vectors
+    had been cached, which stayed true-looking on the two paths where
+    nothing reached disk — a fragmentless vault (the cache write never
+    runs) and a swallowed ``OSError`` from a full or read-only volume.
+    Both now say so out loud, which is the only operator-visible trace of
+    that deliberately-swallowed error.
+
+    Extracted rather than inlined for the same reason as
+    :func:`_format_cluster_stats`: xenon's ``--max-modules B`` bites on
+    this module's aggregate, so branches live in their own small function.
+
+    Args:
+        summary: The link summary being rendered.
+
+    Returns:
+        The clause naming the row count the run persisted, or the explicit
+        admission that it persisted none.
+    """
+    # Named from the canonical constants rather than a literal, so a
+    # rename of the cache file cannot leave the sentence pointing at a
+    # path that no longer exists.
+    from creek.link.embeddings import (
+        EMBEDDINGS_CACHE_DIR,
+        EMBEDDINGS_CACHE_FILENAME,
+    )
+
+    cache_label = f"{EMBEDDINGS_CACHE_DIR}/{EMBEDDINGS_CACHE_FILENAME}"
+    if summary.vectors_persisted > 0:
+        return f"{summary.vectors_persisted} vector(s) cached in {cache_label}"
+    return f"no vectors written to {cache_label}"
 
 
 def _format_cluster_stats(summary: LinkSummary) -> str:
--- a/creek-tools/creek/link/embeddings.py
+++ b/creek-tools/creek/link/embeddings.py
@@ -25,7 +25,7 @@ from creek.models import (
 )
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Mapping
+    from collections.abc import Iterable, Iterator, Mapping
     from pathlib import Path
 
     import numpy as np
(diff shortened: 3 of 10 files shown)


Geoffe-Ga:
## Summary
Rewrites the `creek link --method embeddings` summary sentence so every number in it (`scanned`, `computed`, `cached`) is derived from what the run actually did to disk, rather than three separate claims that could go stale independently. It also replaces the `len(find_resonances(...))` edge-count call — which materialized one Pydantic `Resonance` per pair (~1KB each, ~43GB at documented 35k-fragment scale) — with a true streaming `count_resonances` that shares a `_prepare_scan`/`_scan_pairs` traversal with `find_resonances` so the two can't drift. Includes a doc correction (`>=` not `>` at the similarity threshold) and a CHANGELOG entry.

## Strengths
- `LinkSummary.vectors_persisted` and the new required (no-default) `EmbeddingPass.persisted` field are threaded end-to-end from `_persist_cache`'s actual `len(entries)`/`0`-on-`OSError` return value up to the CLI clause — I traced this through `link_engine.py` and it's consistent: a fully warm run reports the whole corpus as `cached` while `computed` is `0`, matching the stated intent.
- The `find_resonances`/`count_resonances` split via a shared `_ResonanceScan`/`_prepare_scan`/`_scan_pairs` is a clean refactor — it makes divergence between the two entry points structurally hard, and both raise identically on ragged input since validation is shared.
- `_resonances_exact` is a genuine generator (row-block matmul, yields per pair); `_resonances_topk` stays eager but is documented as eager rather than falsely presented as streaming (dedup-then-sort is inherently eager, and its O(n·k) bound is small).
- `_ResonanceScan` correctly sets `eq=False` on the frozen dataclass since it holds a numpy array — the default generated `__eq__`/`__hash__` would raise; nice catch, called out in the docstring with the actual exception text.
- Test coverage is unusually rigorous for this kind of change: schema-derived (not hardcoded) assertions on the parquet row count and column set, a defaultless-field contract test, a call-count spy proving zero `Resonance` construction, a hierarchy-check-count spy proving true laziness (not just an eager list disguised as a generator), and a `tracemalloc`-bounded peak-memory test distinguishing "streams" from "drains into a list" — the PR description's claimed mutation-testing pass (12/12 killed) matches what these tests would actually catch.
- Docs (`docs/linking.md`) gained a table explaining the three counts and clarified `>=` vs `exceeds`, backed by `test_similarity_threshold_is_inclusive_at_the_boundary`.
- CHANGELOG entry is accurate and specific about what changed and what didn't (schema unchanged, no migration).

## Security Concerns
None. No new I/O, network, subprocess, or path-handling surface. 🟢 LOW (nothing found).

## Problems
None blocking. I checked for stale callers of the removed bound methods (`EmbeddingLinker._resonances_exact`/`_resonances_topk`, now module-level functions) and found only docstring references, no live call-site breakage. The `_persist_cache` return-value plumbing (`entries` = fresh-reused ∪ new, `len(entries)` on success, `0` on `OSError`) matches the PR's stated semantics.

## Code Quality
- Minor: `_format_embeddings_cache_clause` does a local `from creek.link.embeddings import ...` inside the function rather than a module-level import in `cli.py`; the docstring explains this is deliberate (xenon module-complexity budget forces small extracted functions), which is a reasonable trade given the stated constraint — not something to change.
- The CLI sentence is now three clauses long, which the tests handle by whitespace-normalizing/ANSI-stripping output rather than matching raw Rich-wrapped text — a sensible way to keep the assertions robust to terminal width, not a smell.

## CLAUDE.md Adherence
✅ TDD workflow described and evidenced (RED-first failures shown in the PR body before GREEN).
✅ Coverage: PR reports 94.67% overall with per-file gate green, comfortably above the ≥90% threshold.
✅ Linting/formatting/mypy strict/complexity (`xenon --max-absolute B --max-modules B --max-average A`) all reported clean, and the diff itself shows no `# noqa`/`# type: ignore`/skip markers introduced.
✅ Conventional commit message (`fix(link): ...`) on a feature branch, not a direct push to `main`.
✅ Uses `./scripts/check-all.sh` as the gate per CLAUDE.md workflow rather than ad hoc tool invocation.

## Testing
Extensive and well-targeted: parity tests (`count_resonances == len(find_resonances(...))`) parametrized across `exact`/`topk` × hierarchy/no-hierarchy with a non-zero vacuity guard; CLI-level tests tying the printed count to `pq.read_table(...).num_rows` for cold/warm/empty/OSError-failure states; a dataclass-contract test for the defaultless `persisted` field; laziness proven via hierarchy-check call counting rather than an unfalsifiable return-type check; a memory-ceiling test distinguishing streaming from list-draining. This meets and exceeds the bar for "insufficient testing is blocking" — nothing here is under-tested.

## Documentation
`docs/linking.md` gained a clear table explaining `scanned`/`computed`/`cached` and corrected the threshold-comparison wording, now pinned by a test so it can't drift again. CHANGELOG entry is detailed and accurate. Docstrings throughout the new/changed functions (`_ResonanceScan`, `_ScanStats`, `count_resonances`, `_persist_cache`) explain *why*, not just *what*, consistent with repo conventions.



## Verdict: LGTM

The change is well-scoped, correct, and thoroughly tested: it replaces a hardcoded, sometimes-false CLI sentence with values derived from the real parquet write, and fixes a genuine O(edges) memory blowup by making the counting path a true generator. Every claim in the PR body is verifiable in the diff (schema-tied test assertions, dataclass-field defaultless-ness test, generator-laziness test via call-count spying, tracemalloc-bounded streaming test), and CI is green on all completed jobs. No blocking issues found; the two follow-ups it deliberately deferred are filed as separate issues rather than silently dropped.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5263802648 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1452, opened 2026-08-12
fix(config): delete the timezone key nothing read, and declare LA_TZ once

### LINKED ISSUE #1339: fix(config): the timezone key is validated at load and read by nothing — America/Los_Angeles is hard-coded
**Labels:** `de-slop`, `creek-pipeline`, `dead-code`, `priority-high`
**Severity:** High
**Remediation:** decision-needed (wire it through `creek/time.py`, or delete the key and correct the docs)

## Problem

`CreekConfig.timezone` (`creek/config.py:1751`) is validated against `ZoneInfo`
at load (`config.py:1819`) — real work that signals the value matters — but
**nothing ever reads it**. Every timestamp the pipeline writes is anchored to a
hard-coded `America/Los_Angeles` constant. Family 3 (dead config) + family 6
(lying documentation).

**Current state:** `docs/configuration.md:48` states the key is "**Used for every
`datetime` written into frontmatter.**" It is used for nothing. An operator
outside the Pacific timezone who sets `timezone: Europe/London` gets LA-anchored
`created`/`ingested` timestamps, and the naive-frontmatter repair path
(`Fragment._normalise_timestamp` → `ensure_aware`) attaches the **wrong** offset
to their own wall-clock readings — `ensure_aware` attaches rather than converts,
so this is genuinely lossy.

## Evidence (corroboration)

**Signal 1 — structural.**
```
$ grep -rn --include=*.py "\.timezone\b" creek creek_mcp crawdad
(no matches)
```
The only `timezone` occurrences repo-wide are the field definition, its
validator, and unrelated docstrings. The actual normalisation targets are two
hard-coded module constants, themselves declared twice:
`creek/time.py:29 LA_TZ = ZoneInfo("America/Los_Angeles")` and
`creek/ingest/base.py:36` (same value, independently declared), plus
`creek/ingest/chatgpt.py:329 .astimezone(LA_TZ)`.

**Signal 2 — reading/intent.** A field that is *validated* at load teaches the
operator it is load-bearing; a `ZoneInfo` check that rejects a typo and then
discards the value is worse than no key at all, because the operator gets
positive feedback that the setting took effect. `creek/time.py`'s own module
docstring states the LA anchoring as an ontology mandate (§8.3), which is a
coherent design — but then the config key should not exist, or should be
constrained to the values the design supports.

## Scope

The `timezone` key, its validator, and the two docs that describe it. Does not
attempt to make the whole pipeline timezone-configurable unless that is the
chosen direction.

## Tasks

1. **Decide** the direction and record it (a one-paragraph note in
   `docs/configuration.md` is enough):
   - (a) *Wire it*: have `creek/time.py` resolve its anchor from the loaded
     config, thread the config to the ingest-time helpers, and de-duplicate the
     second `LA_TZ` declaration in `creek/ingest/base.py:36`; **or**
   - (b) *Delete it*: remove `CreekConfig.timezone` and its validator, and state
     in `docs/configuration.md` that Creek anchors every timestamp to
     America/Los_Angeles by design.
2. Implement the chosen direction.
3. Add a test that fails if the key is reintroduced without a reader (or, under
   (a), that a non-LA `timezone` actually changes a written frontmatter offset).

## Acceptance Criteria

- [ ] `docs/configuration.md:48` matches the code.
- [ ] Either `timezone` has at least one production reader, or it no longer exists.
- [ ] `LA_TZ` is declared once.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/config.py` | Modify |
| `creek-tools/creek/time.py` | Modify |
| `creek-tools/creek/ingest/base.py` | Modify |
| `creek-tools/docs/configuration.md` | Modify |
| `creek-tools/tests/test_config.py` | Modify |

Geoffe-Ga:
## Summary

`CreekConfig.timezone` was validated against `ZoneInfo` at load — real work that
teaches an operator the setting matters — and then read by **nothing**. Every
timestamp the pipeline writes was anchored to a hard-coded `America/Los_Angeles`
constant, which was itself declared twice.

The issue offered a fork: wire the key through, or delete it. **This PR deletes
it**, and the reason is not "wiring was expensive" — it is that wiring would
have been actively harmful.

`generate_fragment_id` (`creek/ingest/base.py`) computes
`f"{source}:{timestamp.isoformat()}:{content}"`, and `isoformat()` renders the
UTC offset. One instant therefore hashes differently per anchor — verified by
execution:

```
2023-11-14T14:13:20-08:00  ->  frag-f8d2701a16e9   (America/Los_Angeles)
2023-11-14T22:13:20+00:00  ->  frag-b79caa69dda2   (Europe/London)
```

Same moment, two ids. A settable anchor would make every fragment's identity a
function of a mutable operator setting — reopening the exact `#1329` bug class
this repo already paid for with a dedicated vault migration module
(`creek/ingest/pin_ids.py`) and a whole regression suite
(`tests/test_ingest_id_host_independence.py`). It would also dissolve
`ensure_aware`'s correctness argument (naive frontmatter is an *LA* wall-clock
reading) and make `chatgpt.py`'s `_MISSING_TIMESTAMP_SENTINEL` a moving instant.
Deleting the key is what `creek/time.py`'s own §8.3 ontology mandate already
said.

### The migration this needed (the part the issue did not mention)

`CreekConfig` is `extra='forbid'`, and `generate_default_config` writes
`CreekConfig().model_dump()` — so **every `creek_config.yaml` `creek init` has
ever produced contains a `timezone:` line**. A naive field deletion would have
turned 100% of existing operator configs into a hard `ValidationError` at load.

So deletion ships with a deliberately narrow shim,
`CreekConfig._drop_obsolete_timezone` — a `model_validator(mode="before")` that
removes that one key **by name** (returning a copy, never mutating the caller's
parsed YAML) and logs one WARNING telling the operator the key is obsolete,
ignored, and safe to delete.

`extra='forbid'` is untouched. Flipping it to `extra='ignore'` would have been
the easy migration and a safety regression — a typo'd `redction:` block would
silently become a no-op. Two tests pin that unknown keys still hard-fail, one of
them specifically with a stale `timezone:` sitting right beside the typo.

Environment vars needed no migration: pydantic-settings never collects a
`CREEK_*` var that matches no field, so a leftover `export CREEK_TIMEZONE=...`
is inert. Pinned by test so it cannot regress into a crash.

### The guard against the key coming back

The invariant "the config cannot advertise a setting nothing honours" was
already enforced here — `tests/test_config_contract.py` (#1042) AST-scans
`creek/` + `creek_mcp/` and fails any field that is neither read nor
allowlisted. `timezone` was sitting in its `DORMANT_CONFIG_FIELDS` allowlist;
that entry is now removed, and the allowlist only ever shrinks.

On top of it this PR adds
`test_timezone_field_may_not_return_without_a_production_reader`, which
deliberately **does not consult the allowlist**, so it cannot be silenced by
re-declaring the field dormant. The only legitimate way to make it pass with the
field present is to bring it back *with a real production reader*.

Mutation-proved, not asserted:

| mutant | result |
| --- | --- |
| reintroduce `timezone` field, no reader | **7 tests fail** |
| reintroduce it *and* re-add a `DORMANT_CONFIG_FIELDS` entry | generic gate goes green, **the new guard still fails** |

Both mutants applied via byte snapshots and restored by verified SHA-256 (never
`git checkout`, which would have destroyed uncommitted lane work).

### `LA_TZ` declared once

Was two independent declarations (`creek/time.py:29`, `creek/ingest/base.py:38`)
whose duplication `creek/time.py`'s module docstring explicitly *defended*. Now
one, in `creek/time.py`; `creek/ingest/base.py` imports and re-exports it so the
historical `from creek.ingest.base import LA_TZ` path keeps working, and the
five other importers point at the declaration site. Verified:

```
$ grep -rn '^\s*LA_TZ\s*[:=]' --include='*.py' creek creek_mcp
creek/time.py:31:LA_TZ = ZoneInfo("America/Los_Angeles")
```

Note an identity check (`base.LA_TZ is time.LA_TZ`) **cannot** prove this —
`ZoneInfo` caches instances by key, so it was already `True` before the fix. The
load-bearing assertion is a source scan.

Three lying docstrings fixed along the way: `creek/time.py`'s paragraph
defending the duplicate constant, `creek/ingest/discord.py:668`'s "in the
configured timezone", and `CreekConfig`'s class docstring advertising
`CREEK_TIMEZONE` as an example env var.

### Decisions recorded

- **Consent records stay pinned.** `creek/consent.py` stamps consent evidence
  with `LA_TZ`. These must never be operator-configurable — that is an argument
  *for* this direction. The retarget there is import-source-only; record naming
  and timestamps are byte-identical.
- Deleted-vs-converted tests: several `tests/test_config.py` cases used
  `timezone` only as an arbitrary key to exercise `load_config`'s `CREEK_CONFIG`
  path resolution. Those were **converted** to `vault_path`, not deleted —
  deleting them would have been a silent coverage regression. Only three tests
  that genuinely tested the deleted field were removed.

### What Gate 2.5 review caught

Worth recording, because two of the three were *my own* new tests failing to
earn their keep:

1. **Blocker, fixed.** The shim's docstring claimed it lives on the model
   because "MCP, direct construction and the environment all bypass
   `load_config`". The MCP half is false — `creek_mcp/` has **27** `load_config`
   calls and **zero** direct `CreekConfig()` constructions. On an issue whose
   entire premise is documentation that lied, that could not ship. Reworded to
   the two bypasses that are real.
2. **A test of mine was unfalsifiable.** I had added
   `test_shim_does_not_mutate_the_callers_mapping` to pin the "copy, not a pop"
   comment. Mutation-testing it revealed the in-place `del` mutant *survives* —
   because as a `BaseSettings`, this model receives a dict its settings sources
   freshly merged, never the caller's object. The test could never fail, and the
   code comment asserting "``data`` is the caller's parsed YAML" was itself
   false. Test deleted; comment corrected to say the copy is defensive rather
   than load-bearing.
3. **A test of mine probed the wrong thing.** The non-mapping guard test passed
   `["not", "a", "mapping"]` — but `in` works fine on a list, so it survived
   removal of the `isinstance` guard. Now parametrised over `None` and `42`,
   which are genuinely non-iterable; both fail with `TypeError` when the guard
   is removed.

Every guard in this PR was mutation-verified after fixing, not just asserted.

## Test plan

- `./scripts/check-all.sh` — **exit 0**; 12/12 checks, **9693 passed**, coverage
  **94.66%** (gate 90%), per-file gate green, mypy strict clean, ruff clean
  verified with `--no-cache`.
- The 5 skips are pre-existing environmental ones (Ollama unreachable, absent
  API keys) — no test skipped by this change.
- Gate 1 RED proved by running: 9 failing tests before implementation, all
  behavioural (no missing-symbol errors).
- Mutation battery above.
- Bypass sweep over added lines: no `noqa`, `type: ignore`, `pylint: disable`,
  `skip`/`xfail`, `pragma: no cover`, `nosec`. No `pyproject.toml`, `scripts/`,
  pre-commit or CI files touched.

**Substitution noted:** the issue's acceptance criteria list
`pre-commit run --all-files`. That was **not** run — it reformats unrelated
files across the worktree. `./scripts/check-all.sh` is the repo's gate and runs
the same ruff/mypy/bandit/interrogate checks; it passed at exit 0.

Closes #1339

### DIFF
--- a/creek-tools/creek/classify/review_runner.py
+++ b/creek-tools/creek/classify/review_runner.py
@@ -28,8 +28,8 @@ from creek.classify.constants import (
     MANUAL_METHOD,
 )
 from creek.classify.review import ReviewQueueGenerator
-from creek.ingest.base import LA_TZ
 from creek.models import Fragment, Frequency, FrequencyClassification
+from creek.time import LA_TZ
 from creek.vault.reader import try_load_fragment
 
 if TYPE_CHECKING:
--- a/creek-tools/creek/config.py
+++ b/creek-tools/creek/config.py
@@ -15,7 +15,6 @@ import re
 from functools import cached_property
 from pathlib import Path
 from typing import TYPE_CHECKING, Literal
-from zoneinfo import ZoneInfo
 
 import yaml
 from pydantic import BaseModel, Field, field_validator, model_validator
@@ -1746,12 +1745,22 @@ class DiscordSourceConfig(BaseModel):
         return value
 
 
+_OBSOLETE_TIMEZONE_KEY = "timezone"
+"""Config key deleted by #1339, still present in every pre-#1339 vault config.
+
+Creek anchors every timestamp it writes to America/Los_Angeles by ontology
+mandate §8.3 (:data:`creek.time.LA_TZ`); the knob had no production reader, and
+wiring one would have made fragment ids config-dependent (see
+:meth:`CreekConfig._drop_obsolete_timezone`).
+"""
+
+
 class CreekConfig(BaseSettings):
     """Top-level Creek configuration.
 
     Values are loaded from a YAML file and can be overridden by
     environment variables prefixed with ``CREEK_`` (e.g.
-    ``CREEK_VAULT_PATH``, ``CREEK_TIMEZONE``).
+    ``CREEK_VAULT_PATH``, ``CREEK_SOURCE_DRIVE``).
     """
 
     model_config = SettingsConfigDict(
@@ -1763,9 +1772,6 @@ class CreekConfig(BaseSettings):
 
     source_drive: Path = Path()
     """Path to the mounted source drive containing raw exports."""
-
-    timezone: str = "America/Los_Angeles"
-    """IANA timezone for timestamp normalisation."""
 
     llm: LLMRoutingConfig = Field(default_factory=LLMRoutingConfig)
     """Per-stage LLM routing (#642). A legacy flat ``llm`` block is promoted to
@@ -1832,26 +1838,61 @@ class CreekConfig(BaseSettings):
     discord: DiscordSourceConfig = Field(default_factory=DiscordSourceConfig)
     """Discord ingest mode toggles (data_package | exporter | bot_capture)."""
 
-    @field_validator("timezone")
+    @model_validator(mode="before")
     @classmethod
-    def validate_timezone(cls, v: str) -> str:
-        """Validate that *v* is a recognised IANA timezone.
+    def _drop_obsolete_timezone(cls, data: object) -> object:
+        """Discard the pre-#1339 ``timezone:`` key, loudly (#1339).
+
+        The field was deleted rather than wired: it had no production reader,
+        and :func:`creek.ingest.base.generate_fragment_id` hashes
+        ``timestamp.isoformat()`` — whose rendered UTC offset changes with the
+        zone — so an operator-settable anchor would mint a second id for every
+        fragment on re-ingest, reopening the bug #1329 had to migrate vaults
+        out of.
+
+        A shim is required because this model forbids extra keys while
+        :func:`generate_default_config` dumps the whole model, so every
+        ``creek_config.yaml`` ``creek init`` has ever written carries a
+        ``timezone:`` line. Without this, deletion would turn every existing
+        config into a hard ``ValidationError`` at load. It lives on the model
+        rather than in :func:`load_config` because direct construction reaches
+        validation without passing through that loader's YAML dict —
+        :func:`generate_default_config` builds its own ``CreekConfig()``, and
+        the environment-variable settings source populates the model whether or
+        not a config file was ever read.
+
+        Exactly one key is removed, by name: ``extra='forbid'`` still rejects
+        everything else, so a typo stays a loud error rather than a silent
+        no-op.
 
         Args:
-            v: Timezone string to validate.
+            data: The raw input to validation — a ``dict`` for YAML and
+                settings-source loads, but any object at all when a caller
+                hands one straight to ``model_validate``.
 
         Returns:
-            The validated timezone string.
-
-        Raises:
-            ValueError: If the timezone is not recognised by ``zoneinfo``.
+            *data* unchanged, or a new ``dict`` without the obsolete key.
+            Anything that is not a ``dict`` passes straight through, so it
+            still fails with pydantic's own ``ValidationError`` rather than
+            an ``AttributeError`` raised from inside this validator.
         """
-        try:
-            ZoneInfo(v)
-        except KeyError as exc:
-            msg = f"Invalid timezone: {v}"
-            raise ValueError(msg) from exc
-        return v
+        if not isinstance(data, dict) or _OBSOLETE_TIMEZONE_KEY not in data:
+            return data
+        logger.warning(
+            "Ignoring obsolete `%s` key in your Creek config: it is no longer "
+            "a setting. Creek anchors every timestamp it writes to "
+            "America/Los_Angeles by design, so the key had no effect and can "
+            "be deleted from your config file (issue #1339).",
+            _OBSOLETE_TIMEZONE_KEY,
+        )
+        # A copy, not a pop. As a ``BaseSettings``, this model is handed a
+        # dict its settings sources freshly merged, so an in-place ``del``
+        # would not actually reach the caller's parsed YAML today — but that
+        # is an implementation detail of pydantic-settings, not a promise.
+        # Copying keeps the validator side-effect-free regardless.
+        remaining = dict(data)
+        del remaining[_OBSOLETE_TIMEZONE_KEY]
+        return remaining
 
     @cached_property
     def model_router(self) -> "ModelRouter":
--- a/creek-tools/creek/consent.py
+++ b/creek-tools/creek/consent.py
@@ -39,7 +39,7 @@ from pathlib import Path
 from pydantic import BaseModel, Field, ValidationError
 
 from creek._fsio import atomic_write_text
-from creek.ingest.base import LA_TZ
+from creek.time import LA_TZ
 
 logger = logging.getLogger(__name__)
 
--- a/creek-tools/creek/ingest/base.py
+++ b/creek-tools/creek/ingest/base.py
@@ -31,12 +31,13 @@ from creek.classify.tags_pass import apply_tags
 from creek.ingest.source_unit import compose_source_unit
 from creek.models import Fragment, FragmentLevel
 
+# ``LA_TZ`` — the target timezone for all normalized timestamps — is declared
+# once in :mod:`creek.time` and imported here (#1339). This module re-exports
+# it so the historical ``from creek.ingest.base import LA_TZ`` path keeps
+# resolving for out-of-tree callers.
+from creek.time import LA_TZ
+
 logger = logging.getLogger(__name__)
-
-# ---- Constants ----
-
-LA_TZ = ZoneInfo("America/Los_Angeles")
-"""Target timezone for all normalized timestamps."""
 
 # ---- Pydantic Models ----
 
--- a/creek-tools/creek/ingest/chatgpt.py
+++ b/creek-tools/creek/ingest/chatgpt.py
@@ -22,7 +22,6 @@ from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
 
 from creek.ingest.base import (
-    LA_TZ,
     Ingestor,
     ParsedFragment,
     RawDocument,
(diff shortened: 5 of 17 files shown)


Geoffe-Ga:
## Summary
Deletes the dormant `CreekConfig.timezone` field (never read by production code) rather than wiring it up, because doing so would have made `generate_fragment_id`'s hashed `timestamp.isoformat()` — and therefore fragment identity — a function of a mutable operator setting, reopening the #1329 bug class. Ships a narrow `model_validator(mode="before")` shim that silently-but-loudly drops a stale `timezone:` key from pre-#1339 configs (with a WARNING), consolidates the previously-duplicated `LA_TZ` constant into `creek/time.py` with `creek/ingest/base.py` re-exporting it for backward compatibility, and adds a mutation-tested anti-regression test that closes the exact loophole (re-declaring the field "dormant") that let it survive this long.

## Strengths
- The shim is precisely scoped: only the `timezone` key is popped by name; `extra='forbid'` is left untouched, so a genuine unknown-key typo (e.g. `redction:`) still hard-fails — verified in code that this is a real, non-blanket removal (`creek/config.py:1879-1892`), and pinned by `test_genuinely_unknown_key_is_still_rejected` / `test_unknown_key_rejected_even_beside_a_stale_timezone`.
- `LA_TZ` consolidation is real, not just claimed: `zoneinfo.ZoneInfo` is still imported and used elsewhere in `creek/ingest/base.py` (line 372, for a different purpose), so that import wasn't left dangling; the re-export at `creek/ingest/base.py:38` correctly preserves the historical import path used by 5 other modules.
- `test_config_contract.py`'s new `test_timezone_field_may_not_return_without_a_production_reader` is verified trustworthy: I confirmed `creek/config.py` is genuinely excluded from `_consumed_config_fields()`'s source scan (`_EXCLUDED_SOURCES`), so the shim's own docstring/log message mentioning `"timezone"` cannot spuriously satisfy the "has a reader" check — the guard closes the actual loophole (re-adding a `DORMANT_CONFIG_FIELDS` entry) that let this field survive from #1041 to #1339.
- PR description is unusually rigorous: documents mutation-testing results, explicitly flags a self-authored test that turned out unfalsifiable and was deleted rather than kept for optics, and separately corrects a false claim in its own shim docstring caught during self-review. This is the kind of self-critical diligence that should be encouraged.
- Docstrings that described behavior the code never had (`creek/time.py`, `creek/ingest/discord.py:668`, `CreekConfig`'s env-var example) are fixed alongside the functional change rather than left to rot.

## Security Concerns
None. 🟢 LOW — no injection, path traversal, or secrets handling in this diff; the shim only manipulates an in-memory dict during config validation.

## Problems
None blocking.

## Code Quality
- `creek/config.py:1888-1892` — the comment on the `dict(data)` copy is now framed as "defensive... not load-bearing," which per the PR description was itself a correction after mutation-testing showed the original claim (\"data is the caller's parsed YAML\") was false. Worth double-checking in a future pydantic-settings upgrade that this assumption (settings sources always hand the validator a freshly-merged dict, never the caller's object) still holds — a one-line comment pointer to this PR/issue would help a future reader who hits an in-place-mutation bug trace back the reasoning quickly. Non-blocking.
- The `_OBSOLETE_TIMEZONE_KEY` module-level constant plus its own docstring is slightly more ceremony than a private local string would need, but it does make the AST-scan-based contract test's job easier to reason about and is consistent with the file's existing style of well-documented module constants.

## CLAUDE.md Adherence
✅ Coverage/linting/types per PR description (94.66% aggregate, mypy strict clean, ruff clean, `--no-cache` verified) — could not independently re-run CI in this sandbox (`gh pr checks` required interactive approval not available here), but the claims are consistent with everything independently verified by reading the merged source.
✅ TDD: PR description states 9 failing tests were run before implementation (Gate 1 RED).
✅ No forbidden shortcuts: no `noqa`/`type: ignore`/`skip`/`xfail` in the diff; verified by direct read of changed files.
✅ Commit message follows Conventional Commits (`fix(config): ...`) and closes #1339.
✅ Single-source-of-truth (§1.2): `LA_TZ` duplication eliminated, consistent with the DRY principle.

## Testing
Strong. `TestTimezoneFieldDeleted` in `tests/test_config.py` covers: field absence, stale-key tolerance with correct residual-config survival, WARNING emission (content-checked, not just presence), validator-removal (no more `ZoneInfo` rejection), unknown-key-still-rejected (twice, including alongside a stale `timezone`), inert leftover `CREEK_TIMEZONE` env var, silence on modern configs, and non-mapping `model_validate` input (parametrized over genuinely non-iterable `None`/`42`, correctly avoiding the `str`/`list` false-negative the PR description says an earlier version of this test had). `tests/test_time.py`'s new `test_la_tz_is_declared_exactly_once_in_production` correctly uses a source-text scan rather than an identity check, since `ZoneInfo` caches instances by key and an identity check would pass even against the duplicated-constant bug this PR fixes.

## Documentation
`docs/configuration.md` clearly explains the removal, the WARNING behavior, and links the reasoning to #1339/#1329. The stray `timezone: "America/Los_Angeles"` example in `docs/Ontology/creek_ontology_agent_prompt.md` was also removed, keeping the canonical ontology spec in sync.



## Verdict: LGTM

The deletion is correct, the migration shim is narrow and well-justified (verified: `extra='forbid'` is preserved, `ZoneInfo` import in `creek/ingest/base.py` remains legitimately used elsewhere, `logger` is defined before the validator uses it, and the contract test's `config.py` exclusion is real, so the new anti-regression guard is trustworthy). Test coverage of the migration path (stale key present/absent, non-mapping input, env-var leftovers, neighbor-key rejection) is thorough and each claim in the PR description was spot-checked against the actual merged files. No blocking issues found; only minor optional nits below.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5265356247 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1455, opened 2026-08-12
fix(purge): hostile frontmatter can veto a purge — RecursionError/MemoryError escape _FRONTMATTER_LOAD_ERRORS


Geoffe-Ga:
Found while reviewing #1340.

## Problem

`creek/purge/engine.py::_FRONTMATTER_LOAD_ERRORS` is
`(OSError, TypeError, ValueError, yaml.YAMLError)` — carefully chosen, and
it covers the realistic hand-edited-vault cases (see its docstring and
#847). It does **not** cover two failures PyYAML can still raise on
attacker- or corruption-authored frontmatter:

- **`RecursionError`** from deeply nested flow collections (`[[[[[…`).
  PyYAML recurses on nested flow nodes; `RecursionError` is a
  `RuntimeError` and is caught nowhere in the purge path.
- **`MemoryError` / unbounded time** from alias expansion. `SafeLoader`
  honours anchors and aliases, so a billion-laughs frontmatter in a single
  fragment is a live denial-of-service.

## Why it matters for RTBF specifically

One corrupt or hostile fragment file can permanently block a purge — and
`creek purge vault` is the operator's fallback when everything else has
failed. "Content that cannot be erased because a file it shares a directory
with will not parse" is a denial-of-erasure primitive.

This is pre-existing for `purge fragment` / `source` / `daterange`, which
have always parsed every fragment in `01-Fragments/`. #1340 extends the
same exposure to `purge vault`, which previously parsed nothing.

Mitigating context, so this is not over-rated: since #1340 the vault
census runs strictly *before* any deletion and the counts are committed
strictly *after* the wipe, so an abort here destroys nothing and the
`status="partial"` audit line correctly reports `fragments_deleted: 0`. The
failure is a refusal to act, not a half-erased vault or a false record.

## Tasks

Widen tolerance at the *call sites that only ever delete*, not by loosening
the shared `_FRONTMATTER_LOAD_ERRORS` constant — the write-safe loader
must keep its narrow tuple, because there a parse failure protects the
operator's bytes.

A fragment whose frontmatter cannot be parsed by any means should be
**counted** (the wipe destroys it regardless) and contribute **no id**,
which is already the documented asymmetry for the `id`-less case.

## Acceptance criteria

- [ ] A fragment with deeply nested flow YAML does not abort
      `purge vault`, `purge source`, or `purge daterange`.
- [ ] It is still counted in `fragments_deleted` and still deleted.
[...]

Issue #1454, opened 2026-08-12
fix(purge): bound the MCP deleted_files payload and add the missing containment guard on _list_fragment_files


Geoffe-Ga:
Found while reviewing #1340. Two adjacent hardening items in the purge
package that were deliberately left out of that PR to keep its diff to the
counting bug it fixes.

## 1. `creek.purge.vault`'s MCP payload forwards an uncapped path list

`creek_mcp/tools/purge.py::_result_payload` forwards
`**result.model_dump(mode="json")`, which is a deliberate design (#1246:
derive the payload from the model rather than hand-picking fields, because
hand-picking silently dropped six fields including `voice_body_undecodable`).

Since #1340, `PurgeResult.deleted_files` for a vault purge holds one entry
per destroyed file instead of one per top-level folder. Those filenames are
title-derived, so they are content — and on a 35k-fragment vault the tool
returns a multi-megabyte list of them into the calling model's context.
`creek/cli.py` caps its own table at `_MAX_DELETED_FILE_ROWS` for exactly
this reason; the MCP surface has no equivalent bound.

Note this is not unique to the vault operation — `purge source` on a large
vault already returned one path per fragment before #1340 — but #1340 makes
the vault path behave the same way, which is what surfaced it.

The fix has to respect the #1246 principle. Two options:
- cap in `_result_payload` and add an explicit `deleted_files_total: int`,
  documented in the same "truncated, and says so" terms as
  `_MAX_DELETED_FILE_ROWS`; or
- bound it on `PurgeResult` itself and carry the true total as a field, so
  the payload stays a straight model dump.

## 2. `_list_fragment_files` has no containment guard

`creek/purge/engine.py::_list_fragment_files` is
`(vault/"01-Fragments").rglob("*.md")` with no resolve-then-contain check —
the only walk in that module without one. Every other pointer-following path
(`_resolve_pointer_in_vault`, `_in_any_staging_root`,
`_contained_voice_artifact`) is fail-closed.

So `01-Fragments/x.md -> /somewhere/outside/notes.md` is matched by name and
opened; if the target declares a string `id`, that id is appended to
`affected_fragment_ids` and copied verbatim into `purge.jsonl`, which
survives every purge. It also inflates `fragments_affected` by a file the
wipe only ever removes as an alias.

This is pre-existing for `purge fragment` / `source` / `source-path` /
`daterange`, which all iterate the same list; #1340 extends it to
[...]

Issue #1453, opened 2026-08-12
fix(purge): the ingest ledger survives a whole-vault purge, mapping source paths to fragment ids


Geoffe-Ga:
Found while reviewing #1340 (the RTBF vault-purge audit fix). Not caused by
that change; it is larger than anything in that diff and it is the premise
the #1340 privacy decision rests on.

## Problem

`purge_vault` wipes `01-` … `10-` (`_VAULT_CONTENT_FOLDERS`), the two
Adepthood staging roots (`ADEPTHOOD_STAGING_RELDIRS` — journal + uploads),
and `embeddings.parquet`. **Everything else under `00-Creek-Meta/` is
preserved by design.** That includes the ingest ledger at
`00-Creek-Meta/State/ingest/<source>.jsonl` (`creek/ingest/ledger.py`),
which carries per record:

- `source_key` — the original source file path
- `fragment_id`
- a content hash

So a whole-vault right-to-be-forgotten purge already leaves behind a
durable map from **original source path → fragment id → content hash** for
every fragment the vault ever held. `Processing-Log/provenance.jsonl`
survives on the same terms.

This is strictly more identifying than the fragment-id roster #1340 adds to
`purge.jsonl`: the ids alone are truncated SHA-256 digests that confirm
membership only to someone holding the exact original, whereas the ledger
*joins those ids to the source paths that produced them*.

## Why it matters

The audit log is deliberately non-purgeable and that is a defensible,
documented decision. The ingest ledger has no such rationale recorded — it
appears to survive because `purge_vault` preserves `00-Creek-Meta/`
wholesale rather than because anyone decided this data should outlive an
erasure request.

## Tasks

1. Decide, and write down, whether the ingest ledger and
   `Processing-Log/provenance.jsonl` are compliance records (keep, and say
   so where `purge.jsonl`'s non-purgeability is documented) or operational
   state (wipe them in `purge_vault`, like the staging roots).
2. Implement whichever it is. If they are wiped, they need the same
   counted-and-audited treatment the staging roots already get.
3. Extend `docs/cleaning-and-purge.md` to state exactly what survives a
   `creek purge vault` under `00-Creek-Meta/`, as an explicit list. Today a
   reader has to derive it from `_VAULT_CONTENT_FOLDERS`.

## Acceptance criteria

- [ ] A documented decision exists for each surviving `00-Creek-Meta/`
      artifact that names a fragment id or a source path.
[...]

PR #1456, opened 2026-08-12
fix(purge): count the fragments a vault purge destroys, and make a dry run predict its apply

### LINKED ISSUE #1340: fix(purge): purge vault records the folder count as fragments_deleted, so the RTBF audit log is wrong by orders of magnitude
**Severity:** High
**Remediation:** fix in place

## Problem

`purge_vault` sets `result.fragments_affected = len(result.deleted_files)`
(`creek/purge/engine.py:654`), and `_wipe_folder_contents` (`:1681`) appends only
the **top-level** entries of each content folder (`folder_path.iterdir()`,
non-recursive) — so a directory holding 500 fragments contributes exactly one
entry. `_write_outcome_audit` then writes that number as `fragments_deleted`
because `"vault"` is in `_FILE_DELETING_OPERATIONS`, and
`affected_fragment_ids` is never populated for the vault path at all.
Family 0 (correctness) + family 8 (contract erosion).

**Current state:** the right-to-be-forgotten compliance log for the most
destructive operation in the product is wrong by orders of magnitude. Every
fragment `creek` writes lands in `01-Fragments/<subfolder>/`
(`creek/vault/writer.py:80 _PLATFORM_SUBFOLDER` maps every platform to a
subfolder), so the top-level entry count is *always* the folder count, never the
fragment count.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** 500 fragments under
`01-Fragments/Conversations/`:

```
fragments_affected: 1
deleted_files: ['/tmp/.../01-Fragments/Conversations']
affected_fragment_ids: []
audit fragments_deleted: 1   affected: []
remaining fragments on disk: []
```

500 fragments erased; the compliance log says 1.

**Signal 2 — structural + intent.** `PurgeAuditEntry.fragments_deleted` is
documented as "Number of fragment files removed from disk"
(`creek/purge/audit.py:75`) and `PurgeResult.fragments_affected` as "Number of
fragments directly affected". `creek-tools/docs/cleaning-and-purge.md:200-225`
documents the vault outcome line as
`"affected_fragments": ["frag-…", "frag-…"], "fragments_deleted": 47` — a shape
the code cannot produce. The existing test asserts only
`result.fragments_affected >= 1` (`tests/test_purge.py:1941`, `:1959`), so it
passes at 1 and can never catch this. The CLI repeats the wrong number verbatim
(`creek/cli.py:5394` "Fragments affected: 1").

**Related, same package:** `_purge_single` (`engine.py:732`) scrubs references
then `unlink`s the fragment only `if not self.dry_run`, so in a dry run the
already-"purged" files stay on disk and later scrub passes keep finding
references inside them. Reproduced: `dry provenance_scrubbed: 2` vs
`apply provenance_scrubbed: 1`. A preview whose numbers the real operation
cannot reach is not a preview, and both runs land in the tamper-evident log
where they will not reconcile.

## Scope

The vault-purge counting path and the dry-run reference-scrub divergence. Does
not change what gets deleted.

## Tasks

1. Have `_wipe_folder_contents` (or `purge_vault`) count and record **fragment
   files** recursively, not top-level directory entries.
2. Populate `affected_fragment_ids` for the vault path, matching the shape
   `docs/cleaning-and-purge.md:200-225` documents.
3. Make the dry run predict the apply: track fragments as logically removed
   during a dry run so subsequent scrub passes do not re-count references inside
   them.
4. Replace the `>= 1` assertions in `tests/test_purge.py:1941,1959` with exact
   counts, and add a dry-run/apply reconciliation test.

## Acceptance Criteria

- [ ] Purging a vault with 500 fragments records `fragments_deleted == 500` and
      500 `affected_fragment_ids`.
- [ ] `provenance_scrubbed` is identical between a dry run and the apply it
      previews.
- [ ] The audit entry shape matches `docs/cleaning-and-purge.md`.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/purge/engine.py` | Modify |
| `creek-tools/tests/test_purge.py` | Modify |

Geoffe-Ga:
## Summary

The right-to-be-forgotten compliance log for `creek purge vault` was wrong by two orders of magnitude, and a dry run could not predict the apply it previewed.

**Reproduced at HEAD before changing anything** (500 fragments in `01-Fragments/Conversations`):

```
fragment .md files on disk before: 500
result.fragments_affected: 3          <- three DIRECTORIES, two of them empty
len(result.affected_fragment_ids): 0
remaining fragments on disk: []
audit outcome: fragments_deleted=3 affected_fragments=[]
```

`purge_vault` set `fragments_affected = len(deleted_files)`, and `_wipe_folder_contents` filled `deleted_files` with each content folder's top-level `iterdir()` entries. The existing test asserted `fragments_affected >= 1`, so it passed at 3 and could never have caught this.

After: `fragments_affected: 500`, all 500 ids, `fragments_deleted=500` in the audit line.

### The audit schema did not change

`PurgeAuditEntry.affected_fragments` and `fragments_deleted` already existed and were already wired at the `_write_outcome_audit` call site; `"vault"` was already in `_FILE_DELETING_OPERATIONS`. Only the values were wrong. This matters because the log is a sha256 hash chain: `AuditLog.verify()` hashes the **stored line bytes**, so adding a field would not have broken existing logs — but *renaming* one would have silently dropped data on read (pydantic v2 `extra='ignore'`, and `_coerce_legacy_entry` maps no old→new names). No field was added or renamed.

### Dry run now predicts its apply, exactly

Four counters diverged, from **two independent causes**:

| counter | dry | apply |
|---|---|---|
| `intimate_stubs_removed` | 2 | 1 |
| `provenance_scrubbed` | 6 | 1 |
| `voice_artifacts_removed` | 6 | 4 |
| `wikilinks_removed` | 4 | 1 |

1. *Deletion-driven* — a counted-only `unlink` leaves the artifact on disk for a later pass to re-count.
2. *Rewrite-driven* — a counted-only `write_text` leaves the **old bytes** on disk, so a later pass matches references the real run had already scrubbed. Reproducible with two same-titled fragments and one surviving note, where **nothing is deleted at all**; a path-only fix cannot close it.

New `creek/purge/dryrun.py` records both. It is populated only in a dry run and read only in a dry run, so an apply run cannot reach it.

### Deletion behaviour is unchanged — verified, not argued

I ran a real `purge_source` and a real `purge_vault` over the same rich vault under HEAD's source and this branch's, and diffed a manifest of every surviving file plus its sha256. **Byte-identical.** The only differing line is the reporting counter:

```
- COUNTS deleted=10 frags=10 ...   (HEAD: 10 top-level entries, miscounted as fragments)
+ COUNTS deleted=11 frags=2  ...   (branch: 11 real files, 2 real fragments)
```

## Two defects the security review caught in my own fix

Both were introduced by the first cut, are now fixed, and each is pinned by a mutation-proven test.

- **An aborted vault purge over-claimed.** The census must run before the wipe (afterwards the files are gone), but committing it up front handed the `except` clause a full count: the first `rmtree` fails, nothing is deleted, and the surviving `status="partial"` line still read `fragments_deleted: 1` naming `frag-A` — with `Alpha.md` on disk. HEAD under-claimed here (`0`); for an erasure record, over-claiming is the dangerous direction. Counts now commit only after the destructive section.
- **`_regular_files_under` walked through symlinked directories.** `rglob` scandirs a symlinked directory it is anchored on, so `01-Fragments/ext -> /elsewhere` put out-of-vault paths into `deleted_files` and previewed deletions `shutil.rmtree` refuses to make. Now refused: a symlink to a file is named (the alias really is unlinked), a symlink to a directory contributes nothing.

## The privacy decision, made deliberately

This writes N fragment ids into `purge.jsonl`, which `purge_vault` preserves — so the ids outlive the erasure. **I concluded the ids belong there**, and the security reviewer argued the other side hard enough that it deserves recording:

- `affected_fragments: []` was never a control anyone set; it is downstream of the counting bug. `_scrub_references`' own docstring states the opposite intent in writing: *"the audit's `affected_fragments` list keeps the real ID for compliance reconstruction."*
- The three sibling operations already write ids — including intimate-tier ones — verbatim into the same non-purgeable log.
- A fragment id is `frag-` + 12 hex of `sha256(source:timestamp:content)`: non-invertible, a membership oracle only for someone who already holds the exact original.
- No ceiling, guard, or default is relaxed.

**The reviewer's counter, which I want on record:** for `vault` the criteria are already a total predicate (`{"scope": "entire vault"}`), so a roster answers no question the criteria do not — and the roster is knowingly a *subset* of the count, since unparseable fragments have no id. The narrower alternative (count, no roster) is a one-line change if you prefer it; I did not take it because it fails the issue's stated acceptance criterion and the documented audit shape, and because a bare count with nothing to reconcile against is exactly what let this bug live.

That decision rests on a fact worth knowing: **the ids already survive a vault purge in a worse form** — `00-Creek-Meta/State/ingest/*.jsonl` maps source path → fragment id → content hash and is preserved. Filed as **#1453**; the docs now say so rather than implying the roster is the only survivor.

## Scope notes

- The voice sweep deliberately does **not** consult the ledger. Its needle and haystack are both read from disk and rewritten together, so disk-vs-disk already agrees; overlaying only the haystack measured dry 0 / apply 1. The half-overlay was in the first cut, is removed, and is pinned against reintroduction.
- `creek/cli.py`'s "Deleted files" table is now capped at 20 rows with an "and N more" line — a vault purge printed 3 rows before this change and would now print one per fragment.
- **`pre-commit run --all-files` (acceptance criterion 5) was deliberately not run**: it silently edits unrelated files in a shared worktree. `./scripts/check-all.sh` is the gate and covers the same hooks.

## Follow-ups filed

| # | Priority | What |
|---|---|---|
| #1453 | P1 · security | The ingest ledger survives a whole-vault purge, mapping source paths to fragment ids |
| #1454 | P2 · security | Uncapped MCP `deleted_files` payload; missing containment guard on `_list_fragment_files` |
| #1455 | P2 · security | Hostile frontmatter (`RecursionError`/`MemoryError`) can veto a purge |

## Test plan

- **Gate 1 RED proved by running it**: 16 failures before the fix, each for the right reason (`3 == 500`, `[] == ['frag-A']`, `12 == 2`, full-model-dump mismatch, `ModuleNotFoundError`).
- **Gate 2**: `./scripts/check-all.sh` → **exit 0**, 12/12 checks, 9717 passed, coverage 94.66%, per-file gate green. Ruff re-verified with `--no-cache`; xenon clean.
- **Gate 2.5**: `code-review-orchestrator` → **CLEAN**, zero blockers, all dimensions clean.
- Five weak `>= ` assertions replaced with exact counts (`:1943`, `:1962`, `:3725`, `:3749`, `:3767`); `:3749` was hiding a 2 behind `>= 1`. Three others left deliberately — two are "at least one warning fired" log assertions, one carries an inline rationale about YAML flow-vs-block serialisation; an exact assertion on the result object was added alongside that third one.
- **Mutation-proven**, snapshotting file bytes and restoring by hash (never `git checkout`): re-adding the voice overlay, removing the symlink guard, and committing the census early each turn their pinning test red.
- New: 20 tests across `tests/test_purge.py`, `tests/test_purge_dryrun.py`, `tests/test_mcp_purge.py`. Every deleted line in `test_purge.py` accounted for individually — 6 assertion lines I authorised, plus one call rewritten to capture its result.

Closes #1340


Geoffe-Ga:
## Summary
Fixes #1340: `creek purge vault` reported `fragments_affected` as the count of top-level directory entries it walked (e.g. `3` for a vault with 500 fragments), and its `--dry-run` preview diverged from the real apply run on four counters because it never modeled the world a real deletion/rewrite would leave for later passes to see. The PR adds an accurate pre-wipe fragment census (`_census_fragments_for_vault_purge`), a new `creek/purge/dryrun.py` `DryRunLedger` that records what an apply run *would* delete/rewrite so later passes in the same dry run see consistent state, a symlink-safe recursive file enumerator (`_regular_files_under`) for the deletion record, and a capped/annotated CLI "Deleted files" table.

## Strengths
- Root-cause fix, not a patch: separates *reading* the census (must happen before the wipe) from *committing* it to the result (must happen after the destructive section), so an aborted vault purge can no longer certify an erasure that didn't happen — verified by `test_an_aborted_vault_purge_certifies_no_deletion_it_did_not_make`.
- The `DryRunLedger` safety property (populated every run, queried only in dry-run) is enforced structurally through a single gate method (`_skip_as_removed`) rather than repeated at each call site, and is proven — not just argued — by `test_an_apply_run_ignores_the_dry_run_ledger_entirely`, which monkeypatches the ledger to lie and asserts the filesystem is still correct.
- Correctly distinguishes the two independent causes of dry-run drift (deletion-driven vs. rewrite-driven) and explicitly tests the case where nothing is deleted at all (two same-titled fragments, one surviving note) — the case a paths-only ledger would miss.
- Symlink handling in the new deletion-record walker is deliberate and tested: a symlink to a file is named (real `unlink` target), a symlink to a directory contributes nothing (never walked through), preventing out-of-vault paths from leaking into a compliance record.
- The PR body documents a genuine, well-reasoned privacy trade-off (keeping fragment ids in the roster vs. a bare count) including the counter-argument the author didn't take, and separately filed the ingest-ledger survival issue (#1453) rather than silently expanding scope.
- Weak `>=` test assertions were tightened to exact values project-wide in the touched areas, which is exactly how this bug hid in the first place (the old test asserted `fragments_affected >= 1` and passed at the wrong value, 3).
- Excellent, load-bearing docstrings throughout `dryrun.py` and `engine.py` that explain *why*, not what — matches the project's own documentation bar.

## Security Concerns
🟢 LOW — Two real defects (over-claiming on an aborted vault purge; symlink traversal into out-of-vault content) were caught by the author's own security review pass and fixed within this same PR, each pinned by a mutation-proven regression test. Two follow-up security-tagged issues were filed for genuinely separate scope (#1453 ingest ledger survives a vault purge; #1454 uncapped MCP payload / missing containment guard) rather than bundled in — appropriate scoping, not a gap in this PR.

## Problems
None found that block merge.

## Code Quality
- `creek-tools/creek/purge/engine.py:1595` — the docstring on `_wipe_adepthood_staging` cross-references `` :meth:`_count_fragments_for_vault_purge` ``, but the method is actually named `_census_fragments_for_vault_purge` (defined at `engine.py:815`, called at `engine.py:772`). This is a broken Sphinx-style xref/stale rename artifact — trivial, non-blocking, but worth a one-line fix given how much this PR leans on cross-references being correct.

## CLAUDE.md Adherence
✅ TDD workflow followed per the PR's own account (16 RED failures reproduced before the fix, each tied to a specific wrong number).
✅ `./scripts/check-all.sh` reported exit 0 locally (12/12 checks, 9717 passed, 94.66% branch coverage, per-file gate green) — self-reported but consistent with CI's completed jobs (Integration & E2E, CrawDad Quality ×3, Code Complexity Analysis, Build Distribution) all showing SUCCESS at review time.
⏳ `Static Analysis`, `Pylint`, and `Tests & Type Checking` (3.11/3.12/3.13) were still `IN_PROGRESS` as of this review; nothing has failed. Confirm these finish green before merge, but there's no indication they won't.
✅ Commit/PR conventions and scope discipline (follow-up issues filed rather than scope creep) match repo norms.

## Testing
Very strong. 20 new tests across `test_purge.py`, `test_purge_dryrun.py`, and `test_mcp_purge.py`, asserting against the filesystem and audit log (not just internal counters), including: the 500-fragment acceptance-criterion corpus, nested-directory placement (defeats a top-level-only regression), empty-directory-produces-no-file-entry, symlink non-traversal, partial-failure non-over-claim, full dry/apply field-for-field equality via `model_dump`, the zero-deletion rewrite-only divergence case, ledger-poisoning safety test, and ledger-reset-between-operations test. This is well above the bar for a compliance-critical fix.

## Documentation
`docs/cleaning-and-purge.md` was updated thoroughly and accurately: the asymmetry between `fragments_deleted` and `affected_fragments`, the census-timing/partial-failure contract, the dry-run/apply parity guarantee, and the ingest-ledger caveat (pointing to #1453) are all spelled out for an operator reading the compliance log. One stale internal method-name reference in a docstring (noted above) is the only inaccuracy found.



## Verdict: LGTM

The two-orders-of-magnitude undercount bug and the dry-run/apply divergence are fixed with a correct, well-reasoned design (census-before-wipe/commit-after-wipe for compliance-safe partial-failure semantics; a per-operation DryRunLedger that is populated on every run but only ever queried in dry-run mode). Tests are exhaustive and verify against the filesystem and audit log rather than just internal counters, including an explicit poisoned-ledger test proving apply runs can't be influenced by dry-run bookkeeping. CI is green everywhere it has completed (nothing failing), and the only issue found — a stale docstring cross-reference to a renamed method — is a trivial nit, not a blocker.

Geoffe-Ga:
**CI**: 9/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5266947874 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

PR #1459, opened 2026-08-12
fix(generate): stop glossing ordinary English, and make the drift guard able to fail

### LINKED ISSUE #1343: fix(generate): unglossed_jargon flags ordinary English verbs (rising, absorb, express, collaborate), so a Wavelength essay can never PASS the author desk
**Severity:** High
**Remediation:** fix in place

## Problem

`_phase_terms` (`creek/generate/ontology_glossary.py:125`) and `_mode_terms`
(`:138`) register each enum's raw `.value` as an alias — `"rising"`,
`"peaking"`, `"withdrawal"`, `"absorb"`, `"express"`, `"collaborate"`,
`"integrate"`. `creek/generate/jargon.py:92` matches aliases case-sensitively on
a word boundary, so ordinary English verbs match. Family 0 (correctness) with
family 6 (lying docstring).

`jargon.py`'s module docstring claims the opposite, naming the exact failing
case: *"Surface forms are matched case-sensitively on a word boundary so an
ordinary lower-case word (`rising` the verb) … never trips the scan."*

**Current state:** every `unglossed_jargon` finding is a `MID` finding
(`creek/author/checks.py:384`), which forces `decision = "REVISE"` in
`ReflectionNode.reflect` (`creek/author/reflection.py:124`), and
`AuthorConductor` (`creek/author/conductor.py:403-430`) loops on REVISE up to
`max_rounds` then `ESCALATE`s to a human. An essay about the Archetypal
Wavelength necessarily contains the word "rising"; unless a gloss marker happens
to fall inside the 90-char window, that draft can **never PASS**.

## Evidence (corroboration)

**Signal 1 — reproducing artifact**, run against the current tree:

```
$ python -c "from creek.generate.jargon import detect_unglossed_jargon; \
    print([f.term for f in detect_unglossed_jargon(
      'The bread was rising in the pan while I waited. I could not absorb what he said. '
      'We collaborate on Tuesdays and express ourselves badly.')])"
['rising', 'absorb', 'collaborate', 'express']
```

Four false positives in three sentences of ordinary prose.

**Signal 2 — reading/intent.** The detector is documented as "deliberately
biased toward **false negatives**"; the bias is inverted.
`tests/test_ontology_glossary.py:169` has a false-positive guard for a
*capitalised* non-term (`test_ordinary_capitalised_word_is_not_flagged`, body
`"On Tuesday I walked to London…"`) and **no** test for the lower-case case the
docstring actually promises — coverage theater on precisely the guard that fails.

**Related, same module (bundle into this fix):** `ontology_term_registry`
(`ontology_glossary.py:195`) collides the three `UNCLASSIFIED` sentinels —
`Frequency.UNCLASSIFIED` (label `"unclassified"`), `Phase.UNCLASSIFIED` and
`Mode.UNCLASSIFIED` (label `"Unclassified"`, alias `"unclassified"`) all register
under the same keys, so Mode wins and 30 terms in produce 28 out. The module
docstring promises "A drift-guard test asserts every enum member resolves to an
entry, so a new enum member without a gloss seed fails the build" — for the
Frequency and Phase sentinels that assertion passes by resolving to a **Mode**
term, so the guard reports green on entries that no longer exist.

## Scope

Alias registration for phase/mode terms and the sentinel collision. Does not
change the 90-char gloss window or the finding severity.

## Tasks

1. Stop registering a term's bare lower-case `.value` as a matchable alias when
   it is an ordinary dictionary word — either restrict aliases to capitalised
   surface forms, or gate `.value` aliases behind an explicit opt-in per term.
2. Exclude the `UNCLASSIFIED` sentinels from `ontology_term_registry` entirely —
   they are internal sentinels, not terms a newcomer needs glossed.
3. Add the missing lower-case false-positive test the docstring promises (the
   repro above), and a registry test asserting no two terms collide on a key.

## Acceptance Criteria

- [ ] `detect_unglossed_jargon("The bread was rising in the pan.")` returns `[]`.
- [ ] An essay containing "rising", "absorb", "express" and "collaborate" as
      ordinary verbs can reach `PASS` in the author desk.
- [ ] `ontology_term_registry` produces one entry per registered term (no silent
      collisions), and the drift guard fails for a genuinely missing gloss.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/generate/ontology_glossary.py` | Modify |
| `creek-tools/creek/generate/jargon.py` | Modify |
| `creek-tools/tests/test_ontology_glossary.py` | Modify |

Geoffe-Ga:
## Summary

`_phase_terms` and `_mode_terms` registered each enum's raw lower-case `.value`
as a matchable alias, so `detect_unglossed_jargon` flagged ordinary English:

```
detect_unglossed_jargon("The bread was rising in the pan while I waited. I could
not absorb what he said. We collaborate on Tuesdays and express ourselves badly.")
  -> ['rising', 'absorb', 'collaborate', 'express']
```

Every `unglossed_jargon` finding is `MID` (`creek/author/checks.py:575` — the
issue's `:384` is stale), `ReflectionNode.reflect` turns *any* finding into
`REVISE` (`creek/author/reflection.py:126`), and `AuthorConductor` loops on
REVISE to `max_rounds` then rewrites the verdict to `ESCALATE`
(`creek/author/conductor.py:403-436`). I confirmed the whole chain by running
it: that bread sentence returns `decision == "REVISE"` with four findings, all
`unglossed_jargon`. An essay about the Archetypal Wavelength necessarily
contains "rising", so it could never PASS the author desk.

`jargon.py`'s own module docstring named the exact failing case as something
that could not happen: *"an ordinary lower-case word (`rising` the verb) …
never trips the scan."*

### The design fork (issue task 1)

Chose **capitalised surface forms only**, expressed as a structural invariant
rather than a per-term list: a surface form must carry a capital or be a
multi-word phrase. `OntologyTerm.__post_init__` raises on a bare all-lower-case
single word, so a *future* enum member cannot reintroduce the bug. The dropped
aliases carried no detection power the labels did not already have — "Rising"
and "Absorb" are the forms a draft actually writes; `.value` is a wire value.

**Rejected: per-term opt-in for `.value` aliases.** It makes someone adjudicate
"is `peaking` an ordinary word?" thirteen times, and whatever default a new enum
member inherits is either the bug (default on) or dead config nobody maintains
(default off). The invariant also satisfies the detector's stated bias toward
false negatives; opt-in does not, because the default decides for you.

`"non-dual substrate"` stays legal and detectable: it is multi-word, so it is
vocabulary rather than a serialised value.

### The registry collision (issue task 2)

The real numbers, counted: `iter_ontology_terms()` returned **30** terms but the
registry held only **28** distinct values across **76** keys — every surface form
was registered twice, exact and lower-cased. (The issue's "30 in, 28 out" counts
distinct terms; 76 counts keys. Both denominators are right about different
things.) Exactly **3** collisions, all on the `UNCLASSIFIED` sentinels:
`registry["unclassified"]` and `registry["Unclassified"]` both returned the
**Mode** term, silently burying the Frequency and Phase ones.

The three sentinels now earn no term at all — "we could not classify this" is a
wire value, not vocabulary. The registry holds exact surface forms only and
`_build_registry` *raises* on a duplicate rather than last-writer-wins.
Post-change: **27** terms, **37** unique surface forms, zero collisions.

### The vacuous guard — the part that mattered most

The module docstring promised "a drift-guard test asserts every enum member
resolves to an entry, so a new enum member without a gloss seed fails the
build". It asserted `member.value in registry` — mere *string* resolvability —
and `Frequency.UNCLASSIFIED` / `Phase.UNCLASSIFIED` "passed" by resolving to the
**Mode** term. Green on entries that no longer existed: a guard that could not
fail for the reason it existed.

Identity is load-bearing here in a way that is easy to get wrong. All three
taxonomies are `StrEnum`, so their members compare **and hash** as bare strings
*across classes*:

```
frozenset({Frequency.UNCLASSIFIED, Phase.UNCLASSIFIED, Mode.UNCLASSIFIED})  # len 1
{Frequency.UNCLASSIFIED: 'freq', Phase.UNCLASSIFIED: 'phase', Mode.UNCLASSIFIED: 'mode'}
  # {<Frequency.UNCLASSIFIED>: 'mode'} — one entry
```

A guard keyed by the enum member would have reproduced the exact bug it exists
to catch. So each term carries `source`, uniqueness is keyed by
`(type(source).__name__, source.value)` via `_source_key`, and every sentinel
comparison uses `is`, never `==`.

## Test plan

Gate 1 RED proved by running it: 11 failures, each for the stated reason. The
headline one is the test the docstring already promised and nobody had written —
`test_ordinary_capitalised_word_is_not_flagged` guarded the *capitalised* case
while the lower-case case had no test at all.

Mutation battery — 6 mutations, byte-snapshot restore, hash-verified after each
(never `git checkout`):

| Mutation | Detected by |
|---|---|
| re-add `aliases=(phase.value,)` | `__post_init__` refuses it at import |
| re-add the aliases **and** neuter the invariant (the exact HEAD defect) | exactly 6 tests fail, no others |
| reintroduce the Phase+Mode `Unclassified` collision | `_build_registry`: "two ontology terms claim the surface form 'Unclassified'" |
| point a Phase term at a `Mode` source | `_check_source_collisions`: "two ontology terms derive from Mode.express" |
| delete a live `_MODE_SENSES` entry | `KeyError` at build |
| silently drop `Phase.PEAKING` (no collision, no KeyError) | drift guard: "no gloss-seed term sourced from ('Phase', 'peaking')" |

The last one is the proof the guard now detects identity rather than
resolvability, and the second is the proof the new tests detect the original
defect.

Gate 2 `./scripts/check-all.sh` → **exit 0**: 12/12 checks, 9639 passed,
coverage 94.65%, per-file gate 261 files >= 80%
(`ontology_glossary.py` 97.67%, `jargon.py` 100%). ruff verified with
`--no-cache`; xenon `--max-absolute B --max-modules B --max-average A` clean.

Per this repo's standing guidance I ran `./scripts/check-all.sh` as the gate
rather than `pre-commit run --all-files` (which edits unrelated files in a
worktree); the acceptance criterion's hook list is covered by check-all's 12
gates plus the pre-commit hooks that run on the commit itself.

### Tests changed, by name

Three per-enum guards — `test_every_frequency_member_has_a_registry_entry`,
`test_every_phase_member_has_a_registry_entry`,
`test_every_mode_member_has_a_registry_entry` — were **merged into one**
strictly stronger `test_every_non_sentinel_member_resolves_to_its_own_term`,
which matches each member to the term declaring it as `source` and asserts the
category matches. `test_named_concepts_are_registered` and
`test_altitude_terms_are_registered` **keep their names**; their bodies move
from `x.lower() in registry` to the exact form plus a category assertion.
Nothing else was deleted: 22 removed lines = 21 content + 1 blank, all
accounted for above. Ten tests added.

## Out of scope

The 90-char `_GLOSS_WINDOW` and the `MID` severity are untouched, per the
issue's own Scope section.

Closes #1343

### DIFF
--- a/creek-tools/creek/generate/jargon.py
+++ b/creek-tools/creek/generate/jargon.py
@@ -14,6 +14,17 @@ holds any gloss signal — an em-dash, a parenthetical, an appositive comma, a
 keywords. Surface forms are matched case-sensitively on a word boundary so an
 ordinary lower-case word (``rising`` the verb) or a capitalised non-term
 (``London``) never trips the scan.
+
+Case-sensitivity only buys that where the registry holds no bare lower-case
+form to match. Every registered surface form is prose a draft writes — a label
+("Rising", "Ultraviolet") or a proper name ("Agency") — and a serialised wire
+value (``rising``, ``absorb``) can never be registered at all: constructing an
+:class:`~creek.generate.ontology_glossary.OntologyTerm` around a single
+all-lower-case word raises, in its ``__post_init__``. That guard is what stands
+between this module and the #1343 defect: re-adding a ``.value`` alias would,
+absent it, put the wire values back in the scan's path and make this module flag
+ordinary English again. Weakening the guard and re-adding the alias together is
+the one change that silently restores the bug.
 """
 
 from __future__ import annotations
--- a/creek-tools/creek/generate/ontology_glossary.py
+++ b/creek-tools/creek/generate/ontology_glossary.py
@@ -11,8 +11,20 @@ This module is the **single source of truth** for that vocabulary. Every
 :class:`OntologyTerm` is *derived* from the existing taxonomy constants — the
 ``Frequency`` / ``Phase`` / ``Mode`` enums and the ``FREQUENCY_*`` / ``PHASE_*``
 description maps — rather than re-typed by hand, so the glossary cannot silently
-drift from the taxonomy. A drift-guard test asserts every enum member resolves
-to an entry, so a new enum member without a gloss seed fails the build.
+drift from the taxonomy. Each derived term names the taxonomy member it came
+from in its ``source`` field, and the drift-guard test matches every
+non-sentinel member to the term declaring it **by identity**. Identity is
+load-bearing: all three taxonomies are ``StrEnum`` classes, so their members
+compare *and hash* as bare strings across classes, and all three
+``UNCLASSIFIED`` sentinels are the one string ``"unclassified"``. The older
+guard asserted only that ``member.value`` resolved to *some* registry key, which
+one enum's term could satisfy on another enum's behalf — that is how the
+sentinels slipped through (#1343).
+
+The three ``UNCLASSIFIED`` sentinels earn no term at all: "we could not classify
+this" is a wire value, not vocabulary a draft ever writes. Each builder excludes
+its own sentinel with ``is``, never ``==``, which would also match the other two
+taxonomies' sentinels.
 
 The seeds are short hints (a clause's worth), not dictionary definitions: they
 exist to ground the model so it can phrase the gloss *in the owner's voice* on
@@ -36,18 +48,27 @@ from creek.generate.wavelength import _PHASE_DESCRIPTIONS
 from creek.models import Frequency, Mode, Phase
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping
+    from collections.abc import Mapping, Sequence
 
 #: Plain-English activation sense for each engagement mode, adapted from the
 #: ``_mode_activation_phrase`` base map in :mod:`creek.generate.skills`. Kept
-#: here as a short gloss seed (not the full activation phrasing).
+#: here as a short gloss seed (not the full activation phrasing). Deliberately
+#: keyed only by the modes the glossary glosses — the ``UNCLASSIFIED`` sentinel
+#: is not vocabulary and is excluded from the glossary — so a future ``Mode``
+#: member added without a gloss seed raises ``KeyError`` when the terms are
+#: built. That build-time failure is the drift detection working, not a bug.
+#:
+#: Keying a dict by an enum member is the shape #1343 warns about, and it is
+#: safe *here only* because this map is mono-taxonomy and excludes the sentinel:
+#: no two ``Mode`` members share a value, and ``UNCLASSIFIED`` — the one value
+#: that also belongs to ``Frequency`` and ``Phase`` — is not a key. Do not
+#: generalise the pattern to a map spanning more than one taxonomy.
 _MODE_SENSES: dict[Mode, str] = {
     Mode.INHABIT: "living inside the frequency rather than performing it",
     Mode.EXPRESS: "enacting the frequency outward, for others to feel",
     Mode.COLLABORATE: "working a frequency alongside other people",
     Mode.INTEGRATE: "metabolising a frequency into a coherent whole",
     Mode.ABSORB: "receiving a frequency rather than producing it",
-    Mode.UNCLASSIFIED: "no dominant engagement mode is in play",
 }
 
 #: Developmental-altitude colour terms surfaced in prose (e.g. "Ultraviolet").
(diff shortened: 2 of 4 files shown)


Geoffe-Ga:
## Summary
This PR fixes issue #1343: `ontology_term_registry()` was registering a lower-cased shadow key for every surface form, which turned enum wire values like `"rising"`, `"absorb"`, `"collaborate"`, and `"express"` into matchable "jargon" terms. Since `unglossed_jargon` findings are always `MID` and `ReflectionNode` turns any finding into `REVISE`, ordinary English words made essays about the ontology unable to ever `PASS`. The fix replaces the lower-cased alias scheme with a structural invariant (`OntologyTerm.__post_init__` rejects bare lower-case single-word surface forms), removes the `.value` wire-form aliases for `Phase`/`Mode`, drops the three `UNCLASSIFIED` sentinels from the glossary entirely, and replaces the old "does `member.value` resolve to *some* registry key" drift guard — which could be silently satisfied by a *different* enum's term because all three taxonomies are `StrEnum` and compare/hash as bare strings — with an identity-keyed guard via a new `source` field and `_source_key`.

## Strengths
- Root-cause fix, not a symptom patch: the invariant lives in `OntologyTerm.__post_init__`, so a future contributor cannot silently reintroduce the bug by re-adding a `.value` alias — they'd hit a `ValueError` at construction time.
- Correctly diagnoses and fixes a *second*, independent bug: the old drift guard (`member.value in registry`) was vacuous for the `UNCLASSIFIED` sentinels because `Frequency.UNCLASSIFIED == Phase.UNCLASSIFIED == Mode.UNCLASSIFIED` as `StrEnum` members — the guard could pass while a member's own term was silently shadowed. The new `_source_key` (`(type(source).__name__, source.value)`) and consistent `is`-based sentinel checks (never `==`) correctly account for this.
- `_build_registry` and `_check_source_collisions` fail loudly (raise `ValueError`) on any surface-form or source collision rather than silently last-writer-wins, closing the exact failure mode that caused #1343 (silent overwrite of the Frequency/Phase `unclassified` term by the Mode term).
- Verified by direct execution (per PR description) and by reading the code: I confirmed statically that `ontology_term_registry()` is consumed only by `jargon.py`, so dropping the lower-cased shadow keys can't silently break another caller that relied on case-insensitive lookups.
- Test suite is unusually rigorous: it collapses three redundant per-enum drift-guard tests into one strictly stronger identity-keyed test, adds dedicated tests for the surface-form invariant (bare lower-case rejected, multi-word lower-case phrase allowed), registry-collision detection, sentinel exclusion, and the literal reproduction case from the issue (`test_ordinary_english_verbs_do_not_block_pass` in `test_reflection.py`, asserting the full `PASS`/empty-findings outcome, not just "no jargon finding").
- Docstrings are extensive and explain *why*, including the subtle cross-taxonomy `StrEnum` hashing hazard — exactly the kind of non-obvious invariant worth documenting.

## Security Concerns
🟢 LOW — None found. This is a text-matching/registry-construction change with no I/O, network, subprocess, or user-input handling implications.

## Problems
None found that block merge.

## Code Quality
- `_altitude_terms()` still has the pre-existing `# pragma: no cover` defensive branch (`if FREQUENCY_COLORS[freq] != label.lower(): continue`) — untouched by this PR and out of scope, just noting it's still there.
- Minor: `OntologyTerm.__post_init__`'s check (`surface.islower() and " " not in surface`) treats any non-space-separated lower-case token as invalid, including hyphenated single "words" like `bottoming-out` if such a form ever existed — not an issue today since the current data doesn't produce that, but worth keeping in mind if a future term's wire value uses a hyphen instead of underscore.
- The module docstring changes and inline comments are long, but given the subtlety of the cross-`StrEnum` identity bug being guarded against, the verbosity is justified here rather than being noise.

## CLAUDE.md Adherence
✅ TDD: PR description states Gate 1 RED was proven (11 failures) before implementation.
✅ `./scripts/check-all.sh` reported exit 0 with 12/12 checks, coverage 94.65%, per-file coverage for touched files at 97.67%/100%.
✅ Root-cause fix, no shortcuts, no `noqa`/`type: ignore` suppressions introduced.
✅ Docstrings are thorough (Google-style, Args/Returns/Raises present on new/changed public and private functions).
✅ No new dependencies, no config threshold changes.
I was unable to execute `./scripts/check-all.sh` or pytest myself in this review sandbox (command execution required approval that wasn't available), so I relied on static reading of the diff and surrounding code plus the PR's own detailed, itemized test-plan/mutation-table evidence, which is unusually well corroborated (mutation table naming which test/guard catches each of 6 reintroduced-bug mutations).

## Testing
Strong. Coverage includes: the exact issue reproduction (bread/rising sentence) at both the `detect_unglossed_jargon` level and the full `ReflectionNode.review` → `AuthorConductor` level; the surface-form invariant (bare lower-case rejected in both `label` and `aliases`, multi-word lower-case phrase allowed); registry-collision detection for both surface-form and source collisions; sentinel exclusion checked with `is` semantics; and the strengthened identity-based drift guard. This is not a case of "insufficient testing" — if anything it's a model example of pinning the exact defect and its fix with named, mutation-justified tests.

## Documentation
Docstrings are updated in both `jargon.py` and `ontology_glossary.py` to describe the new invariants and why they matter, including the specific hazard (StrEnum cross-class equality/hashing) that made the old guard vacuous. The module docstring in `jargon.py` now explicitly states the one change that would silently reintroduce the bug (weakening the guard *and* re-adding a `.value` alias together), which is valuable guidance for future maintainers.



## Verdict: LGTM

The change is a well-targeted root-cause fix backed by a structural invariant that prevents regression, an identity-based drift guard that correctly closes the StrEnum cross-class hazard that caused the original bug, and a thorough, well-corroborated test suite including the exact issue-reproduction case at both the detector and reflection-pipeline levels. No security, correctness, or CLAUDE.md-adherence issues were found; nothing here rises to a blocking concern or even a substantive non-blocking suggestion beyond a couple of already-noted minor observations.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5268696140 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

Issue #1461, opened 2026-08-12
docs(skills): lint.SKILL.md points agents at a lint report path that never exists


Geoffe-Ga:
## Problem

`creek/templates/skills/lint.SKILL.md:85` tells the agent:

> Output lives at `00-Creek-Meta/Processing-Log/lint/YYYY-MM-DD.md`.

`creek/lint/runner.py:242` actually writes:

```python
target = log_dir / f"lint-{report.today.isoformat()}.md"
```

i.e. `00-Creek-Meta/Processing-Log/lint-<iso-date>.md` — a `lint-` filename
prefix, not a `lint/` subdirectory. `docs/lint.md`'s Output section has it right;
the deployed skill does not.

This matters more than an ordinary doc typo: `lint.SKILL.md` is deployed into
every vault by `creek init` / `creek skills sync`, so it is what an agent reads
when asked to find the last lint report. An agent following it looks in a
directory that never exists and concludes no lint has run.

## Found while

Working #1344, auditing which files quote the lint report path — the
`Processing-Log` carve-out in `creek.vault.links.iter_link_sources` is
directory-level and so is robust to either spelling. Not a blocker there.

## Also check

The same line claims "previous reports are not overwritten". `LintRunner.__init__`'s
docstring says runs in the same day *do* overwrite that day's file (prior days are
untouched). Reword to match.

## Acceptance criteria

- [ ] `lint.SKILL.md:85` states the real path, `00-Creek-Meta/Processing-Log/lint-<iso-date>.md`.
- [ ] The overwrite claim matches `LintRunner` behaviour.
- [ ] A test pins the skill's stated path against the path `LintRunner.write`
      returns, so the two cannot drift again.

Refs #1344.

Issue #1460, opened 2026-08-12
fix(clean): link extraction is not code-span aware, so documented syntax examples scan as broken links


Geoffe-Ga:
## Problem

`_WIKILINK_PATTERN` and `_RELATIVE_LINK_PATTERN` (`creek/clean/hygiene.py:149,159`)
match link syntax anywhere in a file, including inside inline code spans and
fenced code blocks. Documentation that *describes* wiki-link syntax is therefore
scanned as if it *used* it.

Measured on a fresh `creek init` vault (32 markdown files), scanning the whole
vault reports exactly one finding:

```
- `00-Creek-Meta/Ontology/creek_ontology_agent_prompt.md` → `[[note-name]]`
```

That line is `docs/Ontology/creek_ontology_agent_prompt.md:746`:

> Generate wiki-links (`[[note-name]]`) between related fragments and add them
> to the appropriate frontmatter arrays.

It is a backticked syntax example in the canonical spec that `creek/scaffold.py:314`
deploys into every vault. Nothing an operator can do clears it.

## Why this is filed rather than fixed in #1344

#1344 widened `broken-links` from `01-Fragments` to the whole vault, and worked
around this by excluding `00-Creek-Meta/Ontology/` from the surveyed source set
(`creek.vault.links.iter_link_sources`). That is a deny-list, not a root-cause
fix: the next scaffold file that documents link syntax reintroduces the false
positive. `tests/test_lint_generated_index_pages.py::test_fresh_init_vault_lints_clean`
is the tripwire.

## Proposed fix

Make extraction code-span aware — strip fenced blocks and inline code before
matching — so a documented example is never reported as a used link.

## Risk / blast radius

This changes link extraction for more than the lint check:

- `creek/clean/hygiene.py` — `BrokenLinkScanner` **and** `OrphanScanner`
  (an orphan credited by a link inside a code fence would stop being credited)
- `creek/clean/filters/markdown.py:191`
- `creek/generate/state_tiers.py`
- `creek/purge/engine.py` — right-to-be-forgotten sweeps match `[[id]]` in bytes

A purge that stops matching a link inside a code fence is a **privacy
regression**, so the purge call sites must be audited separately and may need to
keep the current greedy matching deliberately.

Note it would **not** remove the need for the `Processing-Log`/`State` carve-out:
`creek/generate/state.py`'s drift section renders broken targets un-backticked.

## Acceptance criteria

[...]

PR #1462, opened 2026-08-12
fix(lint): stop reporting the index pages Creek just wrote, and survey the links it never read

### LINKED ISSUE #1344: fix(lint): orphan-compiled flags every generated index page (13/13 false positives) and broken-links surveys only 01-Fragments
**Severity:** High
**Remediation:** fix in place

## Problem

`_FRAGMENT_DIRS` in `creek/lint/checks/orphan_compiled.py:33` counts inbound
links from `01-Fragments` and `10-Liminal` only. But `IndexGenerator.generate_all()`
writes 12 Dataview-driven navigation pages **into the compiled dirs themselves**
(`02-Threads/Thread-Index.md`, `03-Eddies/Eddy-Map.md`, and 10×
`06-Frequencies/F*/F*-Index.md`) that nothing wiki-links, by design. And
`08-Decisions` briefs emit `- [[<thread-id>]]` / `- [[<praxis-id>]]`
(`creek/generate/decisions.py:867,892`) that the check never sees.
Family 0 (false-positive storm).

**Current state:** any vault that has run `creek index` gets 12 permanent,
unfixable `orphan-compiled` findings on every `creek lint`, forever.

## Evidence (corroboration)

**Signal 1 — reproducing artifact.** Vault seeded with
`IndexGenerator(...).generate_all()` plus one Praxis page linked from a Decision
brief:

```
13 orphan compiled page(s)
- `02-Threads/Thread-Index.md`
- `03-Eddies/Eddy-Map.md`
- `04-Praxis/praxis-1.md`
- `06-Frequencies/F1-Agency/F1-Index.md`   … (x10)
```

13/13 findings are false positives. `rg '\[\[' creek/generate/indexes.py` →
zero matches; nothing in `creek/` ever emits a wiki-link naming an index page.

**Signal 2 — reading/intent.** The check "emits suggestions" an operator is
meant to act on, and `creek/vault/links.py:14` states the standard this repo set
for itself after #887: *"A check wrong 99.2% of the time is worse than one that
does not exist, because its output still has to be read."* Not covered by #1223
(building the index once per run) or #1277 (ancestry).

**Related, same package (bundle):** `broken-links`
(`creek/lint/checks/broken_links.py:35`) walks
`(vault_path / "01-Fragments").rglob("*.md")` while `build_link_index` on the
very next line walks the whole vault — then reports
`"N broken link(s) across M file(s)"` as a whole-vault verdict.
Every emergence generator writes wiki-links outside that slice
(`generate/decisions.py:867,882,892`, `wavelength.py:1161`, `voice.py:1079`,
`tags.py:475`, `unnamed.py:654,701`, `paradox.py:763`, `lexicon.py:373`), and
that is precisely the population that goes stale when `creek purge` deletes a
fragment. `docs/lint.md:46` advertises it unqualified.

## Scope

The two checks' survey scope. Does not change the lint report format or the
`--check` registry.

## Tasks

1. Extend `_FRAGMENT_DIRS` (`orphan_compiled.py:33`) to count inbound links from
   every vault directory that emits wiki-links — or invert the check to exclude
   machine-generated index pages by their `type:` frontmatter.
2. Make `broken_links.py:35` survey the same set of files `build_link_index`
   already walks, so the reported verdict matches its stated scope.
3. Add tests: a vault after `IndexGenerator.generate_all()` yields zero
   `orphan-compiled` findings; a broken `[[…]]` inside `08-Decisions` is reported.

## Acceptance Criteria

- [ ] `orphan_compiled.run(v)` returns 0 findings on a vault that has just run
      `creek index`.
- [ ] A dangling wiki-link in `08-Decisions` / `05-Wavelength` / `07-Voice` is
      reported by `broken-links`.

## Files to Create/Modify

| File | Action |
|------|--------|
| `creek-tools/creek/lint/checks/orphan_compiled.py` | Modify |
| `creek-tools/creek/lint/checks/broken_links.py` | Modify |
| `creek-tools/tests/test_lint.py` | Modify |

Geoffe-Ga:
## Summary

`orphan-compiled` reported **13 findings on a vault that had just run `creek index`, and all thirteen were false**. `broken-links` surveyed `01-Fragments` alone while printing `N broken link(s) across M file(s)` as a whole-vault verdict.

### The issue's "or" is wrong — the two remedies fix disjoint subsets

Task 1 offered "extend `_FRAGMENT_DIRS` … **or** invert the check to exclude machine-generated index pages". I reproduced the 13 and attributed each:

| Findings | Cause | Which remedy fixes it |
|---|---|---|
| `02-Threads/Thread-Index.md`, `03-Eddies/Eddy-Map.md`, 10× `06-Frequencies/F*/F*-Index.md` | Dataview index notes `IndexGenerator` writes. Nothing in `creek/` ever emits a wiki-link naming one. | **Only exclusion.** There are no inbound links to find, so widening the survey cannot help. |
| `04-Praxis/praxis-1.md` | Linked by `- [[praxis-1]]` from `08-Decisions/brief-1.md` — a real inbound link the check never read. | **Only widening.** Its `type` is `praxis`, so exclusion cannot touch it. |

So both ship. Either alone leaves that vault dirty.

### Surveying the literal whole vault is unshippable

Task 2 asks `broken-links` to survey "the same set `build_link_index` already walks". Taken literally that is the whole vault, and I measured why it cannot ship. `creek lint` writes its own report to `00-Creek-Meta/Processing-Log/lint-<date>.md`, rendering each finding as ``- `src` → `[[target]]` ``. On a vault holding **one** genuine broken link, three successive runs reported:

```
pass 1: 1 broken link(s) across 1 file(s)
pass 2: 2 broken link(s) across 2 file(s)   <- + Processing-Log/lint-2026-08-12.md
pass 3: 3 broken link(s) across 2 file(s)   <- the artefact now quotes it twice
```

One extra finding per run, forever — the same Family-0 storm the issue is about, self-inflicted by the fix. `00-Creek-Meta/State/` is the same loop (`section_drift_warnings` quotes targets; `section_lint_summary` appends the whole report verbatim). And on a fresh 32-file `creek init` vault a whole-vault scan produces exactly one finding:

```
- `00-Creek-Meta/Ontology/creek_ontology_agent_prompt.md` → `[[note-name]]`
```

— the deployed spec's own backticked syntax example at line 746. Permanent, and nothing an operator can do clears it.

New `creek.vault.links.iter_link_sources` is therefore the whole vault **minus those three prefixes and nothing else**. `00-Creek-Meta/Tag-Garden.md` stays surveyed: it emits real `[[fragment-id]]` links, so carving out all of `00-Creek-Meta` would have been a weakening. The admission criterion for any fourth prefix is recorded in the docstring — *quotes the scanner's own output, or documents the link syntax rather than using it* — which deliberately keeps in scope generated pages that quote fragment **body** text, like `08-Decisions` briefs, where a dangling link is genuinely dangling and the duplication is bounded rather than compounding.

`links.py` now names the asymmetry a future reader would otherwise simplify away: **every page remains a valid link *target* (`build_link_index` is unchanged); the set surveyed as *sources* is a strict named subset.**

### The check must not become decoration

- **Exclusion requires an explicit declaration.** No `type` key, an unknown `type`, no frontmatter, or a header YAML cannot parse → still a candidate, still reported. Were the default the other way, any page could exempt itself by writing a broken header, and a silent check is the same end state as the storm it replaces. Four tests pin this.
- **Self-links no longer credit.** Widening the source set to include the compiled layer opened a hole that could not exist when sources were fragments-only.
- **The five `type` literals now live beside the frozenset the generators themselves consume**, so `GENERATED_INDEX_TYPES` and the frontmatter actually written cannot drift (`test_generated_index_types_matches_what_the_generators_write`).
- **End-to-end proof.** Fresh `creek init` + `generate_all()` → `0 orphan compiled page(s)` / `0 broken link(s) across 31 file(s)`. Add one unreferenced thread page and one dangling link in `08-Decisions` → exactly those two reported, nothing else.

### One bug found while promoting the header reader

`read_header_meta` publishes "malformed YAML yields `{}`". Catching `yaml.YAMLError` alone did not keep that promise: PyYAML's constructors call `int()` and `datetime.date()` on the scanned text and let `ValueError` through. `created: 2020-13-45` — a date the ingest parsers can read straight out of an export — propagated out of `build_link_index` and aborted **all of `creek lint` and `creek state`** over one file. Fixed with a failing test first; pre-existing on `main`, but this PR is what makes the promise public.

### Scope

The 14 files are the two checks plus what they mechanically forced and what they made untrue:

| File | Why |
|---|---|
| `clean/hygiene.py`, `tests/test_hygiene.py` | `broken_links.py` is a wrapper; the survey lives in `BrokenLinkScanner.scan`. `OrphanScanner` and `_list_fragment_files` are untouched — there the helper means "candidate orphan fragments", a different question. |
| `tests/test_state.py` | `state.py:1440` is the scanner's third caller and writes a committable artefact. One appended test pins that `## Drift warnings` still filters to `admitted_paths` (fragments only). No production change to `state.py`. |
| `README.md:131`, `docs/cleaning-and-purge.md`, `docs/cleaning-pipeline.md`, `cli.py` docstring | Four more copies of the same false claim as `docs/lint.md:46`: "Scan **fragments** for wiki-links…". |

`docs/lint.md` carries the full argument once; the others point at it (§1.2, no second copies).

## Test plan

- **`./scripts/check-all.sh` → exit 0**, 12/12 gates, coverage 94.64%, per-file gate 261 files ≥ 80%.
- Unit **10057 passed / 5 skipped** (all pre-existing env skips: Ollama, live API keys); integration **358 passed**; e2e **25 passed**. Zero skips in the new module.
- ruff re-verified with `--no-cache` (#1119): clean, 580 files formatted.
- 36 new tests, zero existing tests deleted or renamed (`git diff --numstat` shows `0` deletions in all three appended files; the single deleted line in `test_vault_links.py` is its import line, which I widened when hoisting 11 function-local imports).
- **Mutation battery, 6/6 killed**, every file restored and SHA-256 verified (snapshot/restore, never `git checkout`):

  | Mutation | Tests turned red |
  |---|---|
  | restore `_FRAGMENT_DIRS` | 5 |
  | empty `GENERATED_INDEX_TYPES` | 10 |
  | default a missing marker to *excluded* | 4 |
  | drop the self-link guard | 1 |
  | remove the carve-out (literal whole vault) | 9 |
  | widen the carve-out to all of `00-Creek-Meta` | 3 |
  | rename `State` in the deny list | 1 (the writer-constant drift guard) |

- Path-form robustness checked by hand: absolute, trailing-slash, dot-segment and relative vault paths all give identical results, so the self-link `Path` comparison across two `rglob` walks is sound.
- `pre-commit run --all-files` was **not** run — it silently edits unrelated files. `check-all.sh` is the gate and was substituted for it.

## Known limits, disclosed

- The deny list is a path prefix, so a symlink or copy of a lint report placed elsewhere in the vault would be surveyed. The real invariant is "do not survey a document that quotes findings", which a prefix only approximates; a marker in the report's own frontmatter would be exact, but the report format is out of scope per the issue.
- `docs/linking.md:183` names `broken-links` as the detector for orphaned re-minted eddy/thread pages. That is `orphan-compiled`'s job and the sentence looks wrong, but it was wrong before this change and this change does not alter its truth value, so I left it.

## Follow-ups filed

- **#1460** (P2) — link extraction is not code-span aware, which is the root cause behind the `Ontology` carve-out. Fixing it would let that prefix be dropped; it also touches `OrphanScanner` and the purge sweeps, where losing a match is a *privacy* regression, so it needs its own audit.
- **#1461** (P3) — `lint.SKILL.md:85` points agents at `Processing-Log/lint/YYYY-MM-DD.md`; the runner writes `Processing-Log/lint-<date>.md`.

Closes #1344

### DIFF
--- a/creek-tools/README.md
+++ b/creek-tools/README.md
@@ -128,7 +128,7 @@ Every command is also documented under [`docs/`](docs/) with end-to-end examples
 |---------|---------|-----|
 | `creek clean orphans`       | Identify fragments with zero links after N days. | [cleaning-and-purge](docs/cleaning-and-purge.md) |
 | `creek clean stale-reviews` | Find review queue items older than N days. | [cleaning-and-purge](docs/cleaning-and-purge.md) |
-| `creek clean broken-links`  | Scan fragments for wiki-links pointing to nonexistent files. | [cleaning-and-purge](docs/cleaning-and-purge.md) |
+| `creek clean broken-links`  | Scan the vault for wiki-links pointing to nonexistent files. | [cleaning-and-purge](docs/cleaning-and-purge.md) |
 | `creek clean duplicates`    | Run the dedup sweep and emit a review report. | [cleaning-and-purge](docs/cleaning-and-purge.md) |
 | `creek clean report`        | Summary statistics on vault health. | [cleaning-and-purge](docs/cleaning-and-purge.md) |
 
--- a/creek-tools/creek/clean/hygiene.py
+++ b/creek-tools/creek/clean/hygiene.py
@@ -23,7 +23,7 @@ from typing import TYPE_CHECKING
 import frontmatter
 from pydantic import BaseModel, Field
 
-from creek.vault.links import build_link_index
+from creek.vault.links import build_link_index, iter_link_sources
 
 logger = logging.getLogger(__name__)
 
@@ -508,7 +508,17 @@ class StaleReviewScanner:
 
 
 class BrokenLinkScanner:
-    """Scan fragments for wiki-links and relative links to nonexistent files.
+    """Scan the vault for wiki-links and relative links to nonexistent files.
+
+    The survey is :func:`creek.vault.links.iter_link_sources`: the whole vault
+    minus Creek's own report folders (``00-Creek-Meta/Processing-Log/``,
+    ``00-Creek-Meta/State/``, ``00-Creek-Meta/Ontology/``). Scanning
+    ``01-Fragments`` alone — the behaviour before #1344 — made a dangling link
+    on a thread, praxis or decision page invisible while ``creek lint``
+    published the count as a whole-vault verdict. Scanning the *literal* whole
+    vault is the other failure: the lint report renders every finding as
+    ``- `src` → `[[target]]` ``, so three successive runs over a vault holding
+    ONE genuine broken link reported 1, then 2, then 3.
 
     Wiki-links resolve through :func:`creek.vault.links.build_link_index`,
     which knows every name a page can be linked by — filename stem,
@@ -522,42 +532,46 @@ class BrokenLinkScanner:
         """Scan the vault for broken links.
 
         Checks wiki-links (``[[Target]]``) and relative markdown links
-        (``[text](path)``) in all fragment files.
+        (``[text](path)``) in every surveyed source file — the whole vault
+        except Creek's own report folders, which quote findings back and would
+        otherwise inflate the count run over run.
 
         Args:
             vault_path: Root of the Obsidian vault.
 
         Returns:
-            A :class:`BrokenLinkResult` with broken link details.
+            A :class:`BrokenLinkResult` whose ``total_files_scanned`` counts
+            the files actually surveyed, so the "across N file(s)" line
+            ``creek lint`` prints does not overstate its own coverage.
         """
-        fragment_files = _list_fragment_files(vault_path)
+        source_files = iter_link_sources(vault_path)
         link_index = build_link_index(vault_path)
 
         broken_links: dict[str, list[str]] = {}
         total_broken = 0
 
-        for frag_file in fragment_files:
-            file_broken = self._check_file(frag_file, link_index, vault_path)
+        for source_file in source_files:
+            file_broken = self._check_file(source_file, link_index, vault_path)
             if file_broken:
-                broken_links[str(frag_file)] = file_broken
+                broken_links[str(source_file)] = file_broken
                 total_broken += len(file_broken)
 
         return BrokenLinkResult(
             broken_links=broken_links,
-            total_files_scanned=len(fragment_files),
+            total_files_scanned=len(source_files),
             total_broken=total_broken,
         )
 
     def _check_file(
         self,
-        frag_file: Path,
+        source_file: Path,
         link_index: LinkIndex,
         vault_path: Path,
     ) -> list[str]:
         """Check a single file for broken links.
 
         Args:
-            frag_file: Path to the fragment file.
+            source_file: Path to the surveyed markdown file.
             link_index: Vault-wide name → page index. A wiki-link is broken
                 only when nothing in the vault answers to its target under
                 any of its names.
@@ -566,7 +580,7 @@ class BrokenLinkScanner:
         Returns:
             List of broken link targets found in this file.
         """
-        content = frag_file.read_text(encoding="utf-8", errors="replace")
+        content = source_file.read_text(encoding="utf-8", errors="replace")
         broken: list[str] = [
             f"[[{target}]]"
             for target in _extract_wikilinks(content)
@@ -576,7 +590,7 @@ class BrokenLinkScanner:
         broken.extend(
             target
             for target in _extract_relative_links(content)
-            if self._is_broken_local_link(target, frag_file, vault_path)
+            if self._is_broken_local_link(target, source_file, vault_path)
         )
 
         return broken
@@ -584,7 +598,7 @@ class BrokenLinkScanner:
     @staticmethod
     def _is_broken_local_link(
         target: str,
-        frag_file: Path,
+        source_file: Path,
         vault_path: Path,
     ) -> bool:
         """Check whether a relative target resolves to no existing file.
@@ -597,15 +611,15 @@ class BrokenLinkScanner:
 
         Args:
             target: A relative link target (no URL scheme).
-            frag_file: The fragment file containing the link.
+            source_file: The surveyed file containing the link.
             vault_path: Root of the vault for resolving relative paths.
 
         Returns:
             True if the target resolves to no existing file under either
-            the fragment's directory or the vault root.
+            the source file's directory or the vault root.
         """
         try:
-            resolved = (frag_file.parent / target).resolve()
+            resolved = (source_file.parent / target).resolve()
             vault_resolved = (vault_path / target).resolve()
             missing = not resolved.exists() and not vault_resolved.exists()
         except OSError:
--- a/creek-tools/creek/cli.py
+++ b/creek-tools/creek/cli.py
@@ -6294,7 +6294,12 @@ def clean_stale_reviews(
 def clean_broken_links(
     vault: Path | None = typer.Option(None, help="Obsidian vault path"),
 ) -> None:
-    """Scan fragments for wiki-links pointing to nonexistent files."""
+    """Scan the vault for wiki-links pointing to nonexistent files.
+
+    Surveys every markdown file except Creek's own report folders
+    (``00-Creek-Meta/Processing-Log/``, ``State/``, ``Ontology/``), which
+    quote findings back and would inflate the count on each rerun.
+    """
     from creek.clean.hygiene import BrokenLinkScanner
 
     vault_path = _resolve_vault(vault)
--- a/creek-tools/creek/generate/indexes.py
+++ b/creek-tools/creek/generate/indexes.py
@@ -22,7 +22,7 @@ never ran.
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Final
 
 from creek.models import Frequency
 from creek.scaffold import VAULT_TEMPLATE_DIR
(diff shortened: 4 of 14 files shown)


Geoffe-Ga:
## Summary
Fixes issue #1344: `orphan-compiled` was reporting Creek's own generated Dataview index pages as false-positive orphans, and `broken-links` silently scanned only `01-Fragments` while reporting its count as a whole-vault verdict. The PR introduces a shared `iter_link_sources` (whole vault minus three machine-written report directories), a `GENERATED_INDEX_TYPES` frozenset consumed by both the index generators and the orphan check, promotes `read_header_meta` to public API, and fixes a real bug where malformed YAML (e.g. `created: 2020-13-45`) could raise `ValueError` out of PyYAML's constructors and abort all of `creek lint`/`creek state`.

## Strengths
- Root-caused both defects instead of picking one of the issue's two suggested remedies — the PR correctly shows they fix disjoint subsets of the 13 false positives and ships both.
- `GENERATED_INDEX_TYPES` is defined once beside the generator methods that write those `type` values, and a dedicated test (`test_generated_index_types_matches_what_the_generators_write`) guards against drift.
- `_UNSURVEYED_PREFIXES` in `creek/vault/links.py` is guarded against silent drift from `_PROCESSING_LOG` (`creek/lint/runner.py`) and `_STATE_SUBPATH` (`creek/generate/state.py`) via `test_unsurveyed_prefixes_track_the_writers_own_constants` — verified these constants exist and match.
- The self-link exclusion added to `orphan_compiled._credited_pages` is correctly reasoned about and tested: widening the inbound-source survey to the compiled layer itself created a new hole (a page could credit itself) that didn't exist when sources were fragments-only.
- Path-component matching in `_is_unsurveyed` (tuple of components, not string prefix) correctly avoids swallowing similarly-named siblings like `State-Machine-Notes.md`, with a test proving it.
- The `read_header_meta` exception widening (`yaml.YAMLError, ValueError, OverflowError, RecursionError`) is well justified and scoped tightly around the `yaml.safe_load` call — it fixes a real, reachable production bug rather than papering over one.
- Documentation updates (README, `docs/lint.md`, `docs/cleaning-and-purge.md`, `docs/cleaning-pipeline.md`, `cli.py` docstring) are consistent with the new behavior and cross-reference each other rather than duplicating the argument, per CLAUDE.md §1.2 (DRY).

## Security Concerns
None found. No new I/O beyond existing header-only file reads; no new subprocess/network/user-input handling. 🟢 LOW / none.

## Problems
None found that block merge.

## Code Quality
- `iter_link_sources`, `build_link_index`, and `_stems_in` (for `_COMPILED_DIRS`) each perform independent `rglob("*.md")` walks over the vault per lint run. This is a pre-existing pattern rather than a regression introduced here, but on the 35k-file vault the module's own docstrings cite as the scaling target, consolidating these into a single walk could be a worthwhile follow-up — not blocking.
- Tests reach into private module attributes across files (`_STATE_SUBPATH`, `_PROCESSING_LOG`, `_UNSURVEYED_PREFIXES`) to build the drift guard. This is a deliberate and well-justified choice (it's the whole point of the guard) but is unusual test/module coupling worth being aware of if these constants are ever renamed.

## CLAUDE.md Adherence
✅ Coverage: PR description reports 94.64% aggregate, per-file gate passing across 261 files.
✅ Linting/formatting: ruff re-verified `--no-cache` per the #1119 cache-poisoning guard this repo requires.
✅ Types: new/changed signatures are fully typed; `TYPE_CHECKING`-only `LinkIndex` import is correctly enabled by the module's `from __future__ import annotations`.
✅ Patterns: constants centralized to avoid drift (§1.2 DRY), root-cause fix rather than a shortcut (§1.3), no `noqa`/`type: ignore` additions.
✅ Commits: Conventional Commits style (`fix(lint): ...`), descriptive body explaining the "why."

## Testing
Extensive and well-targeted:
- 36 new tests across `test_hygiene.py`, `test_lint_generated_index_pages.py` (new file), `test_state.py`, and `test_vault_links.py`.
- Covers the exact 13-false-positive repro from #1344, the 1→2→3 feedback-loop regression for `broken-links`, self-link exclusion, all four explicit-declaration-required edge cases (no `type`, unknown `type`, no frontmatter, malformed YAML), the drift guards for both `GENERATED_INDEX_TYPES` and `_UNSURVEYED_PREFIXES`, and a `test_fresh_init_vault_lints_clean` integration tripwire bounding the deny-list.
- `test_state.py` addition specifically confirms the `## Drift warnings` section stays fragment-scoped (per #969) even as the underlying scanner's survey widens — the right regression test for a section that must NOT inherit the wider scope.
Insufficient testing is not a concern here.

## Documentation
Thorough. `docs/lint.md` carries the canonical explanation once; `README.md`, `docs/cleaning-and-purge.md`, and `docs/cleaning-pipeline.md` point at it rather than duplicating (verified the `#check-names` anchor exists). Module and function docstrings (`links.py`, `orphan_compiled.py`, `broken_links.py`) clearly state the withheld-prefix rationale and measured evidence behind each one.



## Verdict: LGTM

No bugs, missing tests, or CLAUDE.md violations found; the change is exhaustively tested (including a drift guard tying `_UNSURVEYED_PREFIXES` to the writers' own constants, self-link exclusion, and malformed-YAML edge cases) and every cross-reference (docs anchors, `_PROCESSING_LOG`/`_STATE_SUBPATH` values, `GENERATED_INDEX_TYPES`) checks out against the current tree. This is mergeable as-is; the one efficiency note below is a genuine optional follow-up, not something that should hold up this PR.

Geoffe-Ga:
**CI**: 7/12 Green
**VERDICT**: LGTM
**Action**: pull comment 5269659694 to see in-depth feedback and continue iterating

Geoffe-Ga:
**CI**: 13/13 Green
**VERDICT**: LGTM

